PR #4, opened 2020-09-01
bindings/capi: Check invalid parameters in capi.

mihashco:
Added checks to capi bindings to avoid crash when API is used with invalid parameters.

### DIFF
--- a/src/bindings/capi/tvgCapi.cpp
+++ b/src/bindings/capi/tvgCapi.cpp
@@ -83,6 +83,7 @@ TVG_EXPORT Tvg_Canvas* tvg_swcanvas_create()
 
 TVG_EXPORT Tvg_Result tvg_canvas_destroy(Tvg_Canvas* canvas)
 {
+    if (!canvas) return TVG_RESULT_INVALID_ARGUMENT;
     delete(reinterpret_cast<SwCanvas*>(canvas));
     return TVG_RESULT_SUCCESS;
 }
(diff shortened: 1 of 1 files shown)


src/bindings/capi/tvgCapi.cpp
```diff
@@ -83,55 +83,64 @@ TVG_EXPORT Tvg_Canvas* tvg_swcanvas_create()
...
+    if (!canvas) return TVG_RESULT_INVALID_ARGUMENT;
     delete(reinterpret_cast<SwCanvas*>(canvas));
     return TVG_RESULT_SUCCESS;
 }
 
 
 TVG_EXPORT Tvg_Result tvg_swcanvas_set_target(Tvg_Canvas* canvas, uint32_t* buffer, uint32_t stride, uint32_t w, uint32_t h, uint32_t cs)
 {
+    if (!canvas || !buffer) return TVG_RESULT_INVALID_ARGUMENT;
     return (Tvg_Result) reinterpret_cast<SwCanvas*>(canvas)->target(buffer, stride, w, h, static_cast<SwCanvas::Colorspace>(cs));
```
JSUYA, line 95: The buffer is being checked inside target function.
If the buffer check is absolutely necessary in this function, I think the color_stop of tvg_gradient_color_stops should be checked for the same reason.
What do you think?

JSUYA (commented):
Hi :)
Please check my comment

src/bindings/capi/tvgCapi.cpp
mihashco: Buffer check is not necessary here (it not cause crash, so I will prepare fix without buffer check). Thank you

JSUYA (approved):
LGTM
Thank you for check :+1:

PR #7, opened 2020-09-01
svg_loader: fix vector memory leaks.

hermet:
vector is designed for c++ syntaxes,
it works properly when c++ memory allocator is applied,

Here svg_loader uses c style structures which allocated using malloc()/calloc().
That brings the memory broken of stl vectors.

So, we replaced it with our customized SvgVector to easily fix it.

### DIFF
--- a/src/loaders/svg/tvgSvgLoaderCommon.h
+++ b/src/loaders/svg/tvgSvgLoaderCommon.h
@@ -119,6 +119,35 @@ enum class SvgParserLengthType
 struct SvgNode;
 struct SvgStyleGradient;
 
+template<class T>
+struct SvgVector
+{
+    T* list;
+    uint32_t cnt;
+    uint32_t reserved;
+
+    void push(T element)
+    {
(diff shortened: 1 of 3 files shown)


src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -320,10 +349,10 @@ struct SvgParser
 
 struct SvgLoaderData
 {
-    vector<SvgNode *> stack;
+    SvgVector<SvgNode *> stack = {nullptr, 0, 0};
```
JSUYA, line 352: Is there any reason to initialize only on this line?
Don't initialize another SvgVector<T>?

JSUYA (approved):
LGTM :+1: 
However, I left a comment because I was curious.
Please check it.

src/loaders/svg/tvgSvgLoaderCommon.h
hermet: others are get calloc() subsequent initialization

PR #8, opened 2020-09-02
ThorVG/Utils: Convert svg to png

JSUYA:
Rasterazing the svg file and extracting it as a png file.
The png encoder uses the 'lodepng' library.
(https://lodev.org/lodepng/)

Usage:
   svg2png [svgFileName] [Resolution] [bgColor]

Examples:
    $ svg2png input.svg
    $ svg2png input.svg 200x200
    $ svg2png input.svg 200x200 ff00ff


meson.build
```diff
@@ -36,8 +36,8 @@ headers = [include_directories('inc'), include_directories('.')]
 subdir('inc')
 subdir('src')
 
-if get_option('example') == true
-    subdir('example')
+if get_option('util') == true
```
hermet, line 39: util => utility, 
This is not a mandatory feature, so please default is false

hermet (changes requested):
1. Please move to src/bin
2. Please disable in default build, since it's not a mandatory feature.
3. Please build build-script properly, remove all installed thorvg lib/headers and build this as if you install first time.

meson.build
JSUYA: I renamed the option name and moved it to src/bin.

hermet (approved):

PR #10, opened 2020-09-03
Fix svg loader's segfualt

JSUYA:
In several cases of svg loaders, segfaults have occurred.
most cases, it refers to wrong node.
These two commits are fixed to refer to the correct node.

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1458,8 +1458,6 @@ static void _cloneNode(SvgNode* from, SvgNode* parent)
     for (uint32_t i = 0; i < from->child.cnt; ++i, ++child) {
         _cloneNode(*child, newNode);
     }
-
-    _freeNode(newNode);
 }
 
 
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1458,8 +1458,6 @@ static void _cloneNode(SvgNode* from, SvgNode* parent)
     for (uint32_t i = 0; i < from->child.cnt; ++i, ++child) {
         _cloneNode(*child, newNode);
     }
-
-    _freeNode(newNode);
```
hermet, line 1462: Is there no case of null parent?
=> if (!parent) _freeNode(newNode);

hermet (changes requested):
Please check a comment.

hermet (approved):

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: It is not necessary. Becuase if parent is null, a leak occurs because the created node is not connected. so. If parent is null, it should be returned at line 1452. However, _cloneNode is only used in UseNode case. And parent is null means that createNode fails, so it is not necessary for the current logic.

hermet: then please make it return at 1452 
if (!from || !parent) return;

hermet (approved):

PR #25, opened 2020-09-09
svg_loader: fixing issues with a A/a command

mgrudzinska:
1. Cmd 'A' from an svg path should not be connected with any other commands via the control points. -> changes moved to #42

2. When the sweep flag is given as a float it draws an elliptic arc different from what the user intended.

Right now if large_arc is given as a float, the 'A' command won't be plotted at all (and nothing after it). _parseLong function called for the first time will read only the part of a float until the dot character. The second call of this function will return 'false', since the strtol function finds a dot as the first char in a given string. This is the desired behavior.

If sweep is given as a float, the svg path will be plotted until a new cmd will occur (a new letter in an svg path). It's because after the second _parseLong call there is a _parseNumber call, which knows how to deal with a number starting with a point. 
So for example:
A 3 3 3 0 1.0 4 4
will be interpreted as: 
A 3 3 3 0 1 0 4 
with one number (4) still left for further interpretation.
So, the whole 'A' command will be plotted, but not in a way, that was intended.

### DIFF
--- a/src/loaders/svg/tvgSvgPath.cpp
+++ b/src/loaders/svg/tvgSvgPath.cpp
@@ -44,12 +44,14 @@ static bool _parseNumber(char** content, float* number)
 }
 
 
-static bool _parseLong(char** content, int* number)
+static bool _parseFlag(char** content, int* number)
 {
     char* end = NULL;
-    *number = strtol(*content, &end, 10) ? 1 : 0;
-    //If the start of string is not number
-    if ((*content) == end) return false;
+    *number = strtol(*content, &end, 10);
+    //If the start of string is not number or a number was a float
+    if ((*content) == end || *end == '.') return false;
+    //If a flag has a different value than 0 or 1
+    if (*number != 0 && *number != 1) return false;
     *content = _skipComma(end);
     return true;
 }
@@ -463,8 +465,8 @@ static char* _nextCommand(char* path, char* cmd, float* arr, int* count)
         if (_parseNumber(&path, &arr[0])) {
             if (_parseNumber(&path, &arr[1])) {
                 if (_parseNumber(&path, &arr[2])) {
-                    if (_parseLong(&path, &large)) {
-                        if (_parseLong(&path, &sweep)) {
+                    if (_parseFlag(&path, &large)) {
+                        if (_parseFlag(&path, &sweep)) {
                             if (_parseNumber(&path, &arr[5])) {
                                 if (_parseNumber(&path, &arr[6])) {
                                     arr[3] = large;


src/loaders/svg/tvgSvgPath.cpp
```diff
@@ -44,12 +44,14 @@ static bool _parseNumber(char** content, float* number)
 }
 
 
-static bool _parseLong(char** content, int* number)
+static bool _parseFlag(char** content, int* number)
```
JSUYA, line 47: I agree that the argument can be a float type and we need to support it.
However, you can easily parse the float type using strof.
And I think you can do typecast it to int.
_parseNumberToInt(char**, int* n)
{
n = (int)strof( , );
...
}
What do you think?

src/loaders/svg/tvgSvgPath.cpp
```diff
@@ -44,12 +44,14 @@ static bool _parseNumber(char** content, float* number)
...
 {
     char* end = NULL;
-    *number = strtol(*content, &end, 10) ? 1 : 0;
-    //If the start of string is not number
-    if ((*content) == end) return false;
+    *number = strtol(*content, &end, 10);
+    //If the start of string is not number or a number was a float
+    if ((*content) == end || *end == '.') return false;
+    //If a flag has a different value than 0 or 1
+    if (*number != 0 && *number != 1) return false;
```
JSUYA, line 54: If it is not 0 or 1, skipComma does not work and I think it may affect the content.

src/loaders/svg/tvgSvgPath.cpp
```diff
@@ -435,7 +437,8 @@ static void _processCommand(vector<PathCommand>* cmds, vector<Point>* pts, char
         case 'a':
         case 'A': {
             _pathAppendArcTo(cmds, pts, cur, curCtl, arr[5], arr[6], arr[0], arr[1], arr[2], arr[3], arr[4]);
-            *cur = {arr[5] ,arr[6]};
+            *cur = *curCtl = {arr[5] ,arr[6]};
```
JSUYA, line 440: I think this fix is a different issue than the parsing issue. so I recommend splitting up with a new PR or new commit.

JSUYA (changes requested):
Thank you for your interest in this issue.
I think this patch can be changed in a simpler and better way than it is now.
Please check the comments.

src/loaders/svg/tvgSvgPath.cpp
mgrudzinska: Well, not exactly. In svg spec (w3.org), the info about large-arc and sweep are:
- these two are flags
- we know what to do if their values are '0' or '1'
- in chapter 9.5.1 we have info, that "Arbitrary numerical values are permitted for all elliptical arc parameters (**other than the boolean flags**), ..."

At first I thought that we could allow values 1.0 and 0.0, but after some consideration I think that the safest approach would be to check for the exact values and type for these flags. If in a path there will be a mistake, like a wrong order of arguments after 'A' cmd, we won't be able to catch this mistake using typecasting.

src/loaders/svg/tvgSvgPath.cpp
mgrudzinska: If the value is not 0 or 1, _parseFlag will return false and in a consequence _nextCommand will return NULL, which will break the while loop in svgPathToTvgPath and any further commands from the svg path (content) will not be analysed. And this is the expected behavior according to the spec.

src/loaders/svg/tvgSvgPath.cpp
mgrudzinska: done.
I'll change this commit after we will agree how it should look like.

src/loaders/svg/tvgSvgPath.cpp
JSUYA: Sorry I completely misunderstood your intentions.

JSUYA (commented):
Oh my mistake, I pressed merge before approve.
I agree to this change.

PR #42, opened 2020-09-10
Fixing issues with a A/a command

mgrudzinska:
Cmd 'A' from an svg path should not be connected with any other commands via the control points.

This change will be excluded from #25


JSUYA (approved):
LGTM

PR #43, opened 2020-09-10
shape: added duplicate api.

mihashco:
Changes:
1. New shape->duplicate(Shape *from) api.
2. New example: testDuplicate
3. Added capi binding for duploicate api
4. Added capi duplication test in testCapi.c

Description:
Added implementation of duplicate api. Duplicate copies content of Shape *from object to newly created shape. Parameters (stroke, path etc) stored so far in origin object will be replaced by new values. 

TODO:
Implement gradient properties duplication

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -210,6 +210,8 @@ class TVG_EXPORT Shape final : public Paint
 public:
     ~Shape();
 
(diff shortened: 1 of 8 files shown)


mihashco:
Invalid codestyle

PR #44, opened 2020-09-10
shape: added duplicate api.

mihashco:
Changes:
1. New shape->duplicate(Shape) api.
2. New example: testDuplicate
3. Added capi binding for duploicate api
4. Added capi duplication test in testCapi.c

Description:

Added implementation of duplicate api. For now it supports stroke
properties and shape properties (fill color, path) duplication.

TODO:
Implement gradient properties duplication


inc/thorvg.h
```diff
@@ -210,6 +210,8 @@ class TVG_EXPORT Shape final : public Paint
 public:
     ~Shape();
 
+    Result duplicate(std::unique_ptr<tvg::Shape>& shape);
```
hermet, line 213: Why not this?
std::unique_ptr<tvg::Shape> duplicate() const noexcept;

inc/thorvg.h
```diff
@@ -210,6 +210,8 @@ class TVG_EXPORT Shape final : public Paint
...
+    Result duplicate(Shape* from); //For CAPI binding
```
hermet, line 214: No, conceptually, It's not allowed.

hermet (changes requested):
Please check a comment.

inc/thorvg.h
mihashco: I was thought that it will be easier to integrate with EFL. Duplication there looks like example below, so 
If we call duplicate which creates new object, we have to replace pointers and delete old object, but of course 
I can do it. Please let me know if you want it 
```c
Efl_VG *new_shape = evas_vg_shape_add(...); 
evas_vg_shape_dup(new_shape, old_shape)
```

inc/thorvg.h
mihashco: Let me think about it for a while. I have to consider how to call first duplicate method in CAPI bindings.

inc/thorvg.h
hermet: I stick to use  std::unique_ptrtvg::Shape duplicate() const noexcept;   this style, 
it's better for current thorvg API concept wise. Can you please change it?

mihashco: Done

inc/thorvg_capi.h
```diff
@@ -126,6 +126,7 @@ TVG_EXPORT Tvg_Result tvg_paint_transform(Tvg_Paint* paint, const Tvg_Matrix* m)
 /* Shape API                                                            */
 /************************************************************************/
 TVG_EXPORT Tvg_Paint* tvg_shape_new();
+TVG_EXPORT Tvg_Paint* tvg_shape_duplicate(Tvg_Paint* paint);
```
hermet, line 129: Please change api to  => tvg_paint_duplicate(const Tvg_Paint* paint);

Right now, we assume this paint is shape, 
but we must support all paints types for duplicate()

src/lib/tvgShape.cpp
```diff
@@ -48,6 +48,34 @@ unique_ptr<Shape> Shape::gen() noexcept
     return unique_ptr<Shape>(new Shape);
 }
 
+unique_ptr<Shape> Shape::duplicate() const noexcept
+{
+    auto shape = Shape::gen();
```
hermet, line 53: I think you just copy method like this.
shape->pImpl->duplicate(IMPL);  //here duplicate() copies all properties.

Shape::Impl::duplicate(Shape::Impl* from)
{
   x = from->x;
   y = from->y;
   memcpy() and etc;
   flag = from->flag;
}

src/lib/tvgShapePath.h
```diff
@@ -137,6 +136,22 @@ struct ShapePath
 
         return true;
     }
+
+    void copyFrom(ShapePath *p)
```
hermet, line 140: move body to Shape::Impl::duplicate()

hermet (changes requested):
Please check a comment.

inc/thorvg_capi.h
mihashco: done

src/lib/tvgShapePath.h
mihashco: done

hermet (approved):

PR #53, opened 2020-09-15
test: Add svg viewer to compare efl-svg

JSUYA:
You can see svg files in the svgs folder as a list.
The left side is the Tvg result and the right is the svg output of EFL.
image


hermet:
I will remove test folder and move sources to examples as for API show-case.
That case, this verification tool is not compatible with examples. 
It will be better if you create an additional repository to build up this application with thorvg depdency.

JSUYA:
This app is not required, so close it.

PR #54, opened 2020-09-15
bindings/capi: Added path getters + test in testCapi.c

pkalota:
Changes:
 - add shape's path getter for C binding
 - add test for this getter in testCapi.c
 - change tvg_shape_linear_gradient_set and tvg_shape_linear_gradient_set to new api

### DIFF
--- a/src/bindings/capi/tvgCapi.cpp
+++ b/src/bindings/capi/tvgCapi.cpp
@@ -254,6 +254,23 @@ TVG_EXPORT Tvg_Result tvg_shape_append_path(Tvg_Paint* paint, const Tvg_Path_Com
     return (Tvg_Result) reinterpret_cast<Shape*>(paint)->appendPath((PathCommand*)cmds, cmdCnt, (Point*)pts, ptsCnt);
 }
 
+TVG_EXPORT Tvg_Result tvg_shape_get_path_coords(const Tvg_Paint* paint, const Tvg_Point** pts, uint32_t* cnt)
+{
+    if (!paint) return TVG_RESULT_INVALID_ARGUMENT;
+    auto ret = reinterpret_cast<const Shape*>(paint)->pathCoords((const Point**)pts);
+    if (cnt) *cnt = ret;
+    return TVG_RESULT_SUCCESS;
+}
+
+
+TVG_EXPORT Tvg_Result tvg_shape_get_path_commands(const Tvg_Paint* paint, const Tvg_Path_Command** cmds, uint32_t* cnt)
+{
+    if (!paint) return TVG_RESULT_INVALID_ARGUMENT;
+    auto ret = reinterpret_cast<const Shape*>(paint)->pathCommands((const PathCommand**)cmds);
+    if (cnt) *cnt = ret;
+    return TVG_RESULT_SUCCESS;
+}
+
 
 TVG_EXPORT Tvg_Result tvg_shape_set_stroke_width(Tvg_Paint* paint, float width)
 {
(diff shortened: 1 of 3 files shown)


src/bindings/capi/tvgCapi.cpp
```diff
@@ -255,6 +255,17 @@ TVG_EXPORT Tvg_Result tvg_shape_append_path(Tvg_Paint* paint, const Tvg_Path_Com
 }
 
 
+TVG_EXPORT Tvg_Result tvg_shape_get_path(const Tvg_Paint* paint, const Tvg_Path_Command** cmds, uint32_t* cmdCnt, const Tvg_Point** pts, uint32_t* ptsCnt)
```
mihashco, line 258: In my opinion it will be better to split tvg_shape_get_path to tvg_shape_get_path_commands and tvg_shape_get_path_cords. In some cases, for example if we want to check if two shapes are able to interpolate we need only path commands. In geometry_get api we need only path coordinates, so if we split tvg_shape_get_path as I mentioned earlier we can save some memory, but you should wait with fix for Mr. Hermet Park opinion.

hermet: Both are fine to me.
In this case please check nullptr for each arguments and return value only if they are available.

hermet (changes requested):
Please check a comment.

src/bindings/capi/tvgCapi.cpp
```diff
@@ -254,6 +254,21 @@ TVG_EXPORT Tvg_Result tvg_shape_append_path(Tvg_Paint* paint, const Tvg_Path_Com
     return (Tvg_Result) reinterpret_cast<Shape*>(paint)->appendPath((PathCommand*)cmds, cmdCnt, (Point*)pts, ptsCnt);
 }
 
+TVG_EXPORT Tvg_Result tvg_shape_get_path_points(const Tvg_Paint* paint, const Tvg_Point** pts, uint32_t* cnt)
+{
+    if (!paint || !pts || !cnt) return TVG_RESULT_INVALID_ARGUMENT;
```
hermet, line 259: Please suggest to return valid values.
auto ret = reinterpret_cast<const Shape*>(paint)->pathCoords((const Point**)pts);   //pts null check is inside performed.
if (cnt) *cnt = ret;

src/bindings/capi/tvgCapi.cpp
```diff
@@ -254,6 +254,21 @@ TVG_EXPORT Tvg_Result tvg_shape_append_path(Tvg_Paint* paint, const Tvg_Path_Com
...
+    *cnt = reinterpret_cast<const Shape*>(paint)->pathCoords((const Point**)pts);
+    return TVG_RESULT_SUCCESS;
+}
+
+
+TVG_EXPORT Tvg_Result tvg_shape_get_path_commands(const Tvg_Paint* paint, const Tvg_Path_Command** cmds, uint32_t* cnt)
```
hermet, line 265: same as points.

src/bindings/capi/tvgCapi.cpp
```diff
@@ -254,6 +254,21 @@ TVG_EXPORT Tvg_Result tvg_shape_append_path(Tvg_Paint* paint, const Tvg_Path_Com
     return (Tvg_Result) reinterpret_cast<Shape*>(paint)->appendPath((PathCommand*)cmds, cmdCnt, (Point*)pts, ptsCnt);
 }
 
+TVG_EXPORT Tvg_Result tvg_shape_get_path_points(const Tvg_Paint* paint, const Tvg_Point** pts, uint32_t* cnt)
```
hermet, line 257: path_points => path_coords

hermet (changes requested):
Please check comments.

hermet (approved):

PR #78, opened 2020-09-23
gradient: Fixed radial gradient setter.

mihashco:
Removed check for gradient radius. Because of check, x and y values
was ignored when radius equals 0 and api was not usable
in integration with external libs which sets gradient center and
radius in separeted functions.

### DIFF
--- a/src/lib/tvgRadialGradient.cpp
+++ b/src/lib/tvgRadialGradient.cpp
@@ -64,7 +64,7 @@ RadialGradient::~RadialGradient()
 
 Result RadialGradient::radial(float cx, float cy, float radius) noexcept
 {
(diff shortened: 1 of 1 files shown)


src/lib/tvgRadialGradient.cpp
```diff
@@ -64,8 +64,6 @@ RadialGradient::~RadialGradient()
 
 Result RadialGradient::radial(float cx, float cy, float radius) noexcept
 {
-    if (radius < FLT_EPSILON) return Result::InvalidArguments;
```
hermet, line 67: we don't allow minus value here.
so you can change like this:
if (radius < 0) return Result::InvalidArguments;

hermet (changes requested):

hermet (approved):

PR #82, opened 2020-09-25
optimization: cut off file dependencies.

mihashco:
we should avoid code insertion during file dependencies,
such as #include "xxx.h" which has implementations.

This could increase binary size, we can avoid it as possible.

Current patch improves binary size like this:

From: file(2059008) = text(120360) data(8096) bss(80) dec(128536)
To  : file(1921832) = text(118429) data(7872) bss(56) dec(126357)

More additional patches will come in to optmize binary size.


mihashco:
my mistake

PR #85, opened 2020-10-01
examples: testStrokeLine example expanded

mgrudzinska:
Added dashed lines to circles and arcs.

Right now there is a problem with dashed curves (fixed in #86 ) and arcs with pie argument = 1. (fixed in #87) 

dash_1

### DIFF
--- a/src/examples/testStrokeLine.cpp
+++ b/src/examples/testStrokeLine.cpp
@@ -9,7 +9,7 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     if (!canvas) return;
 
     //Test for Stroke Width
-    for (int i = 0; i < 10; ++i) {
+    for (int i = 0; i < 7; ++i) {
         auto shape = tvg::Shape::gen();
         shape->moveTo(50, 50 + (25 * i));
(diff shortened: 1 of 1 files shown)


src/examples/testStrokeLine.cpp
```diff
@@ -58,49 +58,108 @@ void tvgDrawCmds(tvg::Canvas* canvas)
...
 
-    float dashPattern3[2] = {10, 10};
-    shape6->stroke(dashPattern3, 2);
+    float dashPattern3[6] = {10, 10, 1, 8, 1, 10};
+    shape6->stroke(dashPattern3, 6);
     if (canvas->push(move(shape6)) != tvg::Result::Success) return;
+
+    //For a comparison with shapes 10-12
+    auto shape7 = tvg::Shape::gen();
+    shape7->appendArc(70, 500, 160, 10, 30, 1);
```
hermet, line 107: please use true/false explicitly. 
appendArc(...., bool pie);

hermet (changes requested):
Please check my comment.

hermet (approved):

PR #86, opened 2020-10-01
Bezier curve:  improvement of the curve splitting algorithm

mgrudzinska:
bezAt() function needed to be fixed.
Also the convergence in the while loop was sped up and
the break condition was changed.

These fixes the problem of plotting dashed strokes in circles and arcs.

### DIFF
--- a/src/lib/tvgBezier.cpp
+++ b/src/lib/tvgBezier.cpp
@@ -110,27 +110,33 @@ float bezAt(const Bezier& bz, float at)
 {
     auto len = bezLength(bz);
     auto biggest = 1.0f;
+    auto smallest = 0.0f;
+    auto t = 0.5f;
+
+    //just in case to prevent an infinite loop
+    if (at <= 0) return 0.0f; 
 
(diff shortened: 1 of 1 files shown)


src/lib/tvgBezier.cpp
```diff
@@ -110,27 +110,33 @@ float bezAt(const Bezier& bz, float at)
...
     while (true) {
         auto right = bz;
         Bezier left;
-        bezSplitLeft(right, at, left);
-        auto len2 = bezLength(left);
+        bezSplitLeft(right, t, left);
+        len = bezLength(left);
 
-        if (fabs(len2 - len) < FLT_EPSILON) break;
+        if (fabs(len - at) < 1e-4 || fabs(smallest - biggest) < FLT_EPSILON) {
```
hermet, line 127: Can we use any standard macro instead of 1e-4?
ie) FLT_EPSILON?

hermet (changes requested):
Please check my comment.

PR #87, opened 2020-10-01
shape: fixing 'pie' option

mgrudzinska:
For a dashed arc with pie = 1 the closing line was missing.

### DIFF
--- a/src/lib/tvgShape.cpp
+++ b/src/lib/tvgShape.cpp
@@ -170,7 +170,7 @@ Result Shape::appendArc(float cx, float cy, float radius, float startAngle, floa
     if (pie) {
         pImpl->path->moveTo(cx, cy);
         pImpl->path->lineTo(start.x + cx, start.y + cy);
-    }else {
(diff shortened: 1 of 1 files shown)


src/lib/tvgShape.cpp
```diff
@@ -205,7 +205,7 @@ Result Shape::appendArc(float cx, float cy, float radius, float startAngle, floa
     }
 
     if (pie) {
-        pImpl->path->moveTo(cx, cy);
```
hermet, line 208: As reviewed, It looks neither moveTo nor lineTo not necessary because close() will automatically works as line To start point.
Could you please double-check if it works fine only when call close()?

hermet (changes requested):
Please check my comment.

src/lib/tvgShape.cpp
mgrudzinska: sure, it works fine.

hermet (approved):

PR #89, opened 2020-10-05
gradient: added check for gradient start and end set for 0

MathewP1:
This will fix problem with gradient start and stop set for (0, 0) (0, 0). It filled shapes with random color lines then. Now it won't fill in this case at all.

### DIFF
--- a/src/lib/sw_engine/tvgSwFill.cpp
+++ b/src/lib/sw_engine/tvgSwFill.cpp
@@ -235,7 +235,6 @@ void fillFetchRadial(const SwFill* fill, uint32_t* dst, uint32_t y, uint32_t x, 
 void fillFetchLinear(const SwFill* fill, uint32_t* dst, uint32_t y, uint32_t x, uint32_t offset, uint32_t len)
 {
     if (fill->linear.len < FLT_EPSILON) return;
-
     //Rotation
     float rx = x + 0.5f;
     float ry = y + 0.5f;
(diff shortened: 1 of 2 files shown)


src/lib/sw_engine/tvgSwFill.cpp
```diff
@@ -234,8 +234,6 @@ void fillFetchRadial(const SwFill* fill, uint32_t* dst, uint32_t y, uint32_t x,
 
 void fillFetchLinear(const SwFill* fill, uint32_t* dst, uint32_t y, uint32_t x, uint32_t offset, uint32_t len)
 {
-    if (fill->linear.len < FLT_EPSILON) return;
```
hermet, line 237: I think you don't need to remove this condition.

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -133,6 +134,7 @@ static bool _rasterSolidRle(SwSurface* surface, SwRleData* rle, uint32_t color)
 static bool _rasterLinearGradientRect(SwSurface* surface, const SwBBox& region, const SwFill* fill)
 {
     if (!fill) return false;
+    if (fill->linear.len < FLT_EPSILON) return false;
```
hermet, line 137: Rather than adding exception all cases, you can skip drawing in this entry point one time: rasterGradientShape()

hermet (changes requested):
Please check my comments.

src/lib/sw_engine/tvgSwRaster.cpp
MathewP1: I know I can but then I have to check if shape->fill exists there, then if it's linear or radial, and then do correct check. I though it makes more sense to put the check in place where we know what kind of fill we have

hermet (approved):

PR #90, opened 2020-10-07
common: code refactoring

hermet:
renamed the enum CompMethod -> CompositionMethod

added FIXME comment
also clean up internals for less code.


src/lib/tvgPaint.h
```diff
@@ -186,12 +182,12 @@ namespace tvg
...
         }
 
-        bool composite(Paint* target, CompMethod compMethod)
+        bool composite(Paint* target, CompositeMethod method)
         {
-            this->compTarget = target;
-            this->compMethod = compMethod;
-            if (this->compTarget) return true;
-            return false;
+            if (!target && method != CompositeMethod::None) return false;
```
JSUYA, line 187: Is the condition correct?
Why do we need the method condition?
If we need to refer to method, I think it should be if (!target || method == CompositeMethod::None).

JSUYA (changes requested):
please check comment

src/lib/tvgCanvasImpl.h
```diff
@@ -78,15 +78,11 @@ struct Canvas::Impl
 
         //Update single paint node
         if (paint) {
-            if (!paint->pImpl->update(*renderer, nullptr, compList, nullptr, RenderUpdateFlag::None)) {
-                return Result::InsufficientCondition;
```
JSUYA, line 82: Is there any reason to delete the case where the update fails?
I think it's kind of maintaining compatibility.
I think it's better to keep passing edata as argument.

hermet: There is no actual compatibility issues, still apis are changeable. 
Also, this return values are useless until now, users can't do anything about this.

JSUYA: User can call update from (class paint)->push() and pass fail to user in case of task creation failure during update~> prepare process.
There is a problem with returning all cases to InsufficientCondition.
But I don't think there is any reason not to return an error.

hermet: When it's meaningful then we can return values at that time.

JSUYA (approved):

PR #100, opened 2020-10-09
dash pattern: handling dash patterns < 1

mgrudzinska:
For the dash pattern values (both, length and gap) < 1 the program crashed.
curLen - variable type changed from int to float.

- Description :
For the dash pattern values (both, length and gap) < 1, the program crashes, because it falls into the infinite while loop (the total length of the dashed outline is reduced in each step by a value of zero, so it never gets smaller and the while loop termination condition is never satisfied).

The curLen variable changed to be float type instead of int, as it holds the current value of the plotted pattern, which is a float type.

- Issue Tickets (if any) :

- Tests or Samples (if any) :
Example that crashes without this correction: SmallDashPatterns.cpp
https://github.com/mgrudzinska/thorvg/tree/mgrudzinska/curLen_ex_crash

- Screen Shots (if any) :

### DIFF
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -165,7 +165,7 @@ struct SwStroke
 struct SwDashStroke
 {
     SwOutline* outline;
-    int32_t curLen;
+    float curLen;
     int32_t curIdx;
     Point ptStart;
     Point ptCur;
--- a/src/lib/sw_engine/tvgSwShape.cpp
+++ b/src/lib/sw_engine/tvgSwShape.cpp
@@ -113,7 +113,6 @@ static void _outlineMoveTo(SwOutline& outline, const Point* to, const Matrix* tr
     _growOutlinePoint(outline, 1);
 
     outline.pts[outline.ptsCnt] = _transform(to, transform);
-
     outline.types[outline.ptsCnt] = SW_CURVE_TYPE_POINT;
 
     if (outline.ptsCnt > 0) {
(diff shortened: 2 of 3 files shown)


hermet:
@mgrudzinska looks like it won't need to apply dash from the beginning if pattern values are too small ( < FLT_EPSILON).
easily, +  exception handling in the api level.
what doe you think?

mgrudzinska:
I agree that the dash pattern validation should be done in the api.

src/lib/sw_engine/tvgSwShape.cpp
```diff
@@ -254,7 +254,7 @@ static void _dashLineTo(SwDashStroke& dash, const Point* to, const Matrix* trans
             Line left, right;
             _lineSplitAt(cur, dash.curLen, left, right);;
             dash.curIdx = (dash.curIdx + 1) % dash.cnt;
-            if (!dash.curOpGap) {
+            if (!dash.curOpGap && dash.curLen > FLT_EPSILON) {
```
mgrudzinska, line 257: For a pattern like (0,0,3,4) there is no need to add to the outline points for a zero length curve.

src/lib/tvgShape.cpp
```diff
@@ -349,7 +349,16 @@ uint32_t Shape::strokeDash(const float** dashPattern) const noexcept
 {
     if (!pImpl->stroke) return 0;
 
-    if (dashPattern) *dashPattern = pImpl->stroke->dashPattern;
+    if (dashPattern) {
```
hermet, line 352: No, I mean ignore input in the setting side....

hermet (changes requested):
Please check a comment.

src/lib/tvgShape.cpp
hermet: I mean there is no point that dash has 0 values in the list (insane)
Thus, we can return fail if there is zero values in the list when user call shape->strokeDash().
Then there is no need to handle up above sides.

hermet (changes requested):
Please check more comments.

src/lib/sw_engine/tvgSwCommon.h
```diff
@@ -165,7 +165,7 @@ struct SwStroke
 struct SwDashStroke
 {
     SwOutline* outline;
-    int32_t curLen;
+    float curLen;
```
hermet, line 168: no reason to change this neither...

hermet (changes requested):
Please check it.

PR #101, opened 2020-10-12
tvgSvgLoader: Add points copy of missing polygon/polyline

JSUYA:
- Description :
When using `<use>` node, do atrribute copy.
At that time, when target(`url`) is `polygon` or `polyline`,
points array is not copied, causing a problem in output.
So, add missing `memcpy()`

- Tests or Samples (if any) :
Test SVG Sample
```html
viewBox="0 0 64 64" enable-background="new 0 0 64 64" xml:space="preserve">
    <g opacity="0.5">
        <defs>
            <polygon id="test" opacity="0.5" points="41.8,14.5 22.2,14.5 22.2,22.8 41.8,40.7"/>
        </defs>
        <use xlink:href="#test"  overflow="visible"/>
    </g>
</svg>
```

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1487,13 +1487,13 @@ static void _copyAttr(SvgNode* to, SvgNode* from)
         }
         case SvgNodeType::Polygon: {
             to->node.polygon.pointsCount = from->node.polygon.pointsCount;
-            to->node.polygon.points = (float*)calloc(to->node.polygon.pointsCount, sizeof(float));
+            to->node.polygon.points = (float*)malloc(to->node.polygon.pointsCount * sizeof(float));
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1488,11 +1488,13 @@ static void _copyAttr(SvgNode* to, SvgNode* from)
         case SvgNodeType::Polygon: {
             to->node.polygon.pointsCount = from->node.polygon.pointsCount;
             to->node.polygon.points = (float*)calloc(to->node.polygon.pointsCount, sizeof(float));
```
hermet, line 1490: This case, memory reset is not necessary, please replace calloc() to malloc()

hermet (changes requested):
Please check comments.

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: I add a patch to fix this.

hermet (approved):

PR #105, opened 2020-10-14
thorvg viewer: fix scale problem

kimcinoo:
If viewer calls load() with a data which does not have size changes of both
viewbox and canvas, then load() does not set a scale value.
This makes incorrect result on viewer side.

So thorvg wasm should set a scale value, whenever load() is called.

### DIFF
--- a/src/wasm/thorvgwasm.cpp
+++ b/src/wasm/thorvgwasm.cpp
@@ -23,7 +23,6 @@ public:
 
     bool load(string data, int width, int height)
     {
-        uint32_t pw, ph;
         float w, h;
 
         mErrorMsg = "None";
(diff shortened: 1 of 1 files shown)


src/wasm/thorvgwasm.cpp
```diff
@@ -55,16 +54,10 @@ class __attribute__((visibility("default"))) ThorvgWasm
...
-        pw = mDefaultWidth;
-        ph = mDefaultHeight;
         mDefaultWidth = static_cast<uint32_t>(w);
         mDefaultHeight = static_cast<uint32_t>(h);
 
-        if (pw != mDefaultWidth || ph != mDefaultHeight) {
-            updateScale();
-        }
-
+        updateScale();
```
hermet, line 60: looks like updateSize() calls updateScale() internally.
Does it really necessary?

hermet (commented):
Please check a comment.

src/wasm/thorvgwasm.cpp
kimcinoo: updateSize calls updateScale only if canvas size is changed.
So this is necessary.

PR #107, opened 2020-10-14
thorvg logo: update svg data

kimcinoo:
The path for the blue background did not start from 0 0.
So there is 1px width horizontal line which does not have any data.
This patch make the path start from 0 0.


hermet:
ahh.. I noticed this issue and the design resource has been changed, applied to all sides.

PR #111, opened 2020-10-21
common canvas: Added canvas->remove API

mihashco:
Description :

Introduce new method: `canvas-remove(Paint)` used to remove paints from
vector of paints in `Canvas` class. Additonally CAPI for new method was
added. Api is used for safly remove Paint object from canvas. It allows
to use following api sequence without crash:

```c
tvg_canvas_remove(canvas, paint);
tvg_paint_del(paint);
tvg_canvas_draw(canvas);
```

@API Additions:
Result Canvas::remove(Paint *p)
tvg_canvas_remove(Tvg_Canvas *canvas, Tvg_Paint *p)

### DIFF
--- a/src/lib/tvgCanvasImpl.h
+++ b/src/lib/tvgCanvasImpl.h
@@ -23,6 +23,7 @@
 #define _TVG_CANVAS_IMPL_H_
 
 #include <vector>
+#include <algorithm>
 #include "tvgPaint.h"
 
 /************************************************************************/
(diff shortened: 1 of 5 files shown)


src/lib/tvgCanvasImpl.h
```diff
@@ -53,6 +54,15 @@ struct Canvas::Impl
         return update(p);
     }
 
+    Result remove(Paint *p)
+    {
+        if (!p) return Result::InvalidArguments;
+        std::vector<Paint*>::iterator it;
```
JSUYA, line 60: You may omit 'std::'

mihashco: Fixed.

hermet:
need to consider the concept...

PR #117, opened 2020-10-23
svg_loader SvgLoader: Prevent memory overflow for tagName

JSUYA:
- Description :
When copying tagName, if length of referenced string is longer
than general case, it is not used as tagName.


hermet:
Here is commit message guide:

=> svg_loader loader: xxxx

hermet (approved):

Issue #77, opened 2020-09-23
Rendering is graphically broken


hermet:
Clock rotation partially broken when it's rotating.
See)  examples/testStress
![Uploading shot-2020-09-23_15-01-44.jpg…]()

mgrudzinska:
PR #126

PR #126, opened 2020-10-29
sw_engine shape: improvement of the rectangle fast tracking algorithm

mgrudzinska:
- Description :
The algorithm erroneously treated some shapes (like isosceles trapezoids  and specifically arranged zero width parallelograms) as rectangles, witch causes the whole bbox to be filled - undesirable behavior.

- Issue Tickets (if any) : #77 

- Tests or Samples (if any) :
   a) Stress.cpp from examples 
   b) Svg.cpp for the image errorsStress.svg:
https://github.com/mgrudzinska/thorvg/tree/mgrudzinska/Stress_ex_error

- Screen Shots (if any) :
errorsStress.svg:

Before (wrong):
stressErr

After (correct):
stressCor

### DIFF
--- a/src/lib/sw_engine/tvgSwShape.cpp
+++ b/src/lib/sw_engine/tvgSwShape.cpp
@@ -430,13 +430,14 @@ bool _fastTrack(const SwOutline* outline)
     auto pt3 = outline->pts + 2;
     auto pt4 = outline->pts + 3;
 
-    auto min1 = pt1->y < pt3->y ? pt1 : pt3;
-    auto min2 = pt2->y < pt4->y ? pt2 : pt4;
-    if (min1->y != min2->y) return false;
+    SwPoint a;
+    SwPoint b;
+    a.x = pt1->x;
+    a.y = pt3->y;
+    b.x = pt3->x;
+    b.y = pt1->y;
 
-    SwCoord len1 = pow(pt1->x - pt3->x, 2) + pow(pt1->y - pt3->y, 2);
-    SwCoord len2 = pow(pt2->x - pt4->x, 2) + pow(pt2->y - pt4->y, 2);
-    if (len1 == len2) return true;
+    if ((*pt2 == a && *pt4 == b) || (*pt2 == b && *pt4 == a)) return true;
 
     return false;
 }


src/lib/sw_engine/tvgSwShape.cpp
```diff
@@ -430,10 +430,26 @@ bool _fastTrack(const SwOutline* outline)
     auto pt3 = outline->pts + 2;
     auto pt4 = outline->pts + 3;
 
-    auto min1 = pt1->y < pt3->y ? pt1 : pt3;
-    auto min2 = pt2->y < pt4->y ? pt2 : pt4;
-    if (min1->y != min2->y) return false;
+    if (*pt1 == *pt4 && *pt2 == *pt3) return false;
```
JSUYA, line 433: Is this condition necessary?
I think, In tvgSwShape's updateBBox, width and height are already being checked.
I think fastTrack only needs to check if it is rectangle.

JSUYA (commented):
Nice catch. please check my comment

src/lib/sw_engine/tvgSwShape.cpp
mgrudzinska: this condition prevents from filling the whole bbox when the rect has a zero width. 
Since the pictures in the Stress example were so small, the zero width rect occured sometimes and that was the reason why we saw 'the flashes'.

mgrudzinska: I misspelled 'zero width rect' when I meant 'zero width parallelogram' - there are, of course, parallelograms on the pictures in the description.
I also found an issue in my previous solution, so I decided to completely change the algorithm. 
Please have a look now.

JSUYA (commented):
Hello I am on vacation(~ November 13). So I couldn't actually run this change.

image
This change cannot check rectangle as above.

I think the most accurate way is to check if P1P2P3 and P3P4P1 are at right angles respectively. 
image
```
/**
 * Calculates the angle (in radians) between two vectors pointing outward from one center
 *
 * @param p0 first point
 * @param p1 second point
 * @param c center point
 */
function find_angle(p0,p1,c) {
    var p0c = sqrt(pow(c.x-p0.x,2)+
                        pow(c.y-p0.y,2)); // p0->c (b)   
    var p1c = sqrt(pow(c.x-p1.x,2)+
                        pow(c.y-p1.y,2)); // p1->c (a)
    var p0p1 = sqrt(pow(p1.x-p0.x,2)+
                         pow(p1.y-p0.y,2)); // p0->p1 (c)
    return acos((p1c*p1c+p0c*p0c-p0p1*p0p1)/(2*p1c*p0c));
}
```
If this method is expensive, I think it is a good idea to keep your first changes.
++ I think you can check whether fill area is zero or not in _checkValid().

ps. My review may not be accurate because I only looked at the code and checked it.

hermet (approved):

PR #149, opened 2020-11-26
common capi: Added scene clear API

mihashco:
Scene::clear() API allows users to remove shapes on their own, without
a crash in paint->dispose() or tvg_paint_del() methods. This case is
needed especially when thorvg is used to draw frames, when in one frame
we have scene with shape, and in next frames (future time stamps) user
deletes shapes

@API additions
```cpp
Result Scene::clear();
```
```c
Tvg_Result tvg_scene_clear(Tvg_Paint *scene);
```

Example:
```c
Tvg_Paint *scene = tvg_scene_new();
Tvg_Paint *shape = tvg_shape_new();

tvg_scene_push(scene, shape);
tvg_scene_clear();
//Now we can safelly free resources manually
tvg_paint_del(scene);
//Without tvg_scene_clear() memory allocatad for shape was double released
tvg_paint_del(shape);
```

### DIFF
--- a/src/lib/tvgScene.cpp
+++ b/src/lib/tvgScene.cpp
@@ -59,3 +59,11 @@ Result Scene::reserve(uint32_t size) noexcept
 
     return Result::Success;
 }
+
+
+Result Scene::clear() noexcept
+{
+    pImpl->paints.clear();
+
+    return Result::Success;
+}
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -303,6 +303,7 @@ public:
 
     Result push(std::unique_ptr<Paint> paint) noexcept;
     Result reserve(uint32_t size) noexcept;
+    Result clear() noexcept;
 
     static std::unique_ptr<Scene> gen() noexcept;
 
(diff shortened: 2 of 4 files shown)


src/lib/tvgSceneImpl.h
```diff
@@ -22,6 +22,7 @@
 #ifndef _TVG_SCENE_IMPL_H_
 #define _TVG_SCENE_IMPL_H_
 
+#include <thorvg.h>
```
hermet, line 25: Is it necessary?

src/lib/tvgScene.cpp
```diff
@@ -58,4 +58,10 @@ Result Scene::reserve(uint32_t size) noexcept
...
 
     return Result::Success;
-}
\ No newline at end of file
+}
+
+
+Result Scene::clear() noexcept
+{
+    return pImpl->clear();
```
hermet, line 66: Not necessary impl version here.

Let's do simply it.
pImpl->paint.clear();

hermet (changes requested):
Please check comments.

hermet (approved):

PR #166, opened 2020-12-04
sw_engine renderer: support scene opacity composition

hermet:
this is an additional enhancement of af8c278c5e2a47c21f44440cdb0a7aaf6f6d3baa

Now scene opacity composition is supported.

Also, this implementaion fixes an incorrect scene bounding box computation.

Plus, adding stroking feathering to shape bounding box size.

- Description :

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :

### DIFF
--- a/src/examples/Opacity.cpp
+++ b/src/examples/Opacity.cpp
@@ -34,6 +34,7 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     //Create a Scene 2
     auto scene2 = tvg::Scene::gen();
     scene2->opacity(127);              //Apply opacity to scene (0 - 255)
+    scene2->scale(1.2);
     scene2->reserve(2);
 
     //Star
--- a/src/examples/Opacity.cpp
+++ b/src/examples/Opacity.cpp
@@ -8,41 +8,58 @@ void tvgDrawCmds(tvg::Canvas* canvas)
 {
     if (!canvas) return;
 
+    //Create a Scene
+    auto scene = tvg::Scene::gen();
+    scene->opacity(175);              //Apply opacity to scene (0 - 255)
+    scene->reserve(2);
+
     //Prepare Circle
     auto shape1 = tvg::Shape::gen();
     shape1->appendCircle(400, 400, 250, 250);
     shape1->fill(255, 255, 0, 255);
-    canvas->push(move(shape1));
+    scene->push(move(shape1));
 
-    //Create a Scene
-    auto scene = tvg::Scene::gen();
-    scene->opacity(127);              //Apply opacity to scene (0 - 255)
-    scene->reserve(2);
-
-    //Star
+    //Round rectangle
     auto shape2 = tvg::Shape::gen();
-
-    //Appends Paths
-    shape2->moveTo(199, 34);
-    shape2->lineTo(253, 143);
-    shape2->lineTo(374, 160);
-    shape2->lineTo(287, 244);
-    shape2->lineTo(307, 365);
-    shape2->lineTo(199, 309);
-    shape2->lineTo(97, 365);
-    shape2->lineTo(112, 245);
-    shape2->lineTo(26, 161);
-    shape2->lineTo(146, 143);
-    shape2->close();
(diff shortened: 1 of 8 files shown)


src/lib/tvgSceneImpl.h
```diff
@@ -59,45 +66,50 @@ struct Scene::Impl
...
+            if (x2 < x + w) x2 = (x + w);
+            if (y < y1) y1 = y;
+            if (y2 < y + h) y2 = (y + h);
         }
 
-        if (px) *px = x;
-        if (py) *py = y;
-        if (pw) *pw = w;
-        if (ph) *ph = h;
+        if (px) *px = x1;
```
JSUYA, line 109: If size of paints is 0, px, py, pw, and ph are set to unintended values.

src/examples/Opacity.cpp
```diff
@@ -8,61 +8,78 @@ void tvgDrawCmds(tvg::Canvas* canvas)
...
+    shape3->opacity(127);
 
-    scene->push(move(shape2));
+    scene2->push(move(shape3));
 
     //Circle
-    auto shape3 = tvg::Shape::gen();
+    auto shape4 = tvg::Shape::gen();
 
     auto cx = 550.0f;
```
JSUYA, line 64: If  cx and cy is 50.0f, bounds x, y will be negative.
If so, rasterize will cause segfault.
Please check this

JSUYA (commented):
Please check my comments

src/lib/tvgSceneImpl.h
```diff
@@ -59,45 +66,50 @@ struct Scene::Impl
...
     {
-        //TODO: composition begin
-        //auto data = renderer.beginComp();
+        void* ctx = nullptr;
+
+        //Half translucent. This requires intermediate composition.
+        if (opacity < 255 && opacity > 0) {
+            float x, y, w, h;
+            bounds(&x, &y, &w, &h);
+            ctx = renderer.beginComposite(roundf(x), roundf(y), roundf(w), roundf(h));
```
JSUYA, line 75: The parameter of beginComposite is uint32_t.
If x and y are negative, i think it may cause problems during the typecasting process.

src/lib/tvgSceneImpl.h
hermet: will return false if paints.size is 0.

JSUYA (commented):
I tested it with a new patch.
I've checked a segfault when rendering an object outside canvas's bounds.
Please check this.
```CPP
diff --git a/src/examples/Opacity.cpp b/src/examples/Opacity.cpp
index f733c3d..f954816 100644
--- a/src/examples/Opacity.cpp
+++ b/src/examples/Opacity.cpp
@@ -61,8 +61,8 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     //Circle
 
-    auto cx = 550.0f;
-    auto cy = 550.0f;
+    auto cx = 1000.0f;
+    auto cy = 1000.0f;
     auto radius = 125.0f;
     auto halfRadius = radius * 0.552284f;


```

JSUYA:
And there is one more problem.
I think. transform data is not applied to surface (or bounds?).
```cpp
diff --git a/src/examples/Opacity.cpp b/src/examples/Opacity.cpp
index f733c3d..1d53bae 100644
--- a/src/examples/Opacity.cpp
+++ b/src/examples/Opacity.cpp
@@ -78,6 +78,8 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     shape4->opacity(200);
     scene2->push(move(shape4));
 
+    scene2->scale(1.2);
+
     //Draw the Scene onto the Canvas
     canvas->push(move(scene2));
 }
```

In this sample.

image
red circle is clliped. 
Perhaps  red circle is clipped from the original boundary.
Please check this

hermet:
@JSUYA  Crash is not related with this patch. it was potential problem.

JSUYA (approved):
LGTM

PR #167, opened 2020-12-04
common picture: revise resize method.

hermet:
this implementation handles the center-aligned scale.
Also removed viewbox usage since we have size() method.

- Description :

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :

### DIFF
--- a/src/lib/tvgPictureImpl.h
+++ b/src/lib/tvgPictureImpl.h
@@ -61,32 +61,34 @@ struct Picture::Impl
 
     void resize()
     {
-        uint32_t w = 0, h = 0;
-        if (this->w != 0 && this->h != 0) {
-            w = this->w;
-            h = this->h;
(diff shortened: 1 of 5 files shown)


src/lib/tvgPictureImpl.h
```diff
@@ -96,8 +98,15 @@ struct Picture::Impl
...
                 if (scene) {
                     paint = scene.release();
-                    resizing = true;
                     loader->close();
+
+                    if (w == 0 && h == 0) {
+                        w = loader->w;
+                        h = loader->h;
+                    } else {
+                        resize();
```
JSUYA, line 107: I think this call(resize) is unnecessary.
The resize function is called in update by the resizing flag.

hermet: it needs. 
size update()-> resize() -> render() -> reload()

PR #171, opened 2020-12-07
wasm: code refactoring.

hermet:
Use Picture::size() method instead of scale()

Picture newly supports size() method to resize image.
It's more convenient for users in its usage.

- Description :

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :

### DIFF
--- a/src/wasm/thorvgwasm.cpp
+++ b/src/wasm/thorvgwasm.cpp
@@ -28,8 +28,6 @@ public:
 
     bool load(string data, int width, int height)
     {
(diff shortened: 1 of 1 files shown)


hermet:
This patch must be applied with https://github.com/Samsung/thorvg/pull/169

kimcinoo:
I have verified with #169 and it works well. Looks good to me.

kimcinoo (approved):

PR #175, opened 2020-12-09
svg_loader: code refactoring

hermet:
Use custom vector data algorithm for optimizing binary size.
Plus remove tuple usage since it leads unnecessary instance copying.

binary size comparison:

[libthorvg.so] 1811400 >> 1785448
[text] 122914 >> 121255
[data] 7824 >> 7792
[dec] 130810 >> 129119

- Description :

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :

### DIFF
--- a/src/loaders/svg/tvgSvgPath.cpp
+++ b/src/loaders/svg/tvgSvgPath.cpp
@@ -59,7 +59,7 @@ static bool _parseFlag(char** content, int* number)
     return true;
 }
 
-void _pathAppendArcTo(vector<PathCommand>* cmds, vector<Point>* pts, Point* cur, Point* curCtl, float x, float y, float rx, float ry, float angle, bool largeArc, bool sweep)
+void _pathAppendArcTo(SvgVector<PathCommand>* cmds, SvgVector<Point>* pts, Point* cur, Point* curCtl, float x, float y, float rx, float ry, float angle, bool largeArc, bool sweep)
 {
     float cxp, cyp, cx, cy;
     float sx, sy;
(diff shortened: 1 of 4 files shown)


src/loaders/svg/tvgSvgPath.cpp
```diff
@@ -499,11 +499,8 @@ static char* _nextCommand(char* path, char* cmd, float* arr, int* count)
 }
 
 
-tuple<vector<PathCommand>, vector<Point>> svgPathToTvgPath(const char* svgPath)
+bool svgPathToTvgPath(const char* svgPath, SvgVector<PathCommand>& cmds, SvgVector<Point>& pts)
```
JSUYA, line 502: The return value from this function is always true. Failure cases are not yet checked. I think it is right to change it to void type.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -283,11 +283,14 @@ unique_ptr<Shape> _shapeBuildHelper(SvgNode* node, float vx, float vy, float vw,
...
 {
+    SvgVector<PathCommand> cmds;
+    SvgVector<Point> pts;
+
     switch (node->type) {
         case SvgNodeType::Path: {
             if (node->node.path.path) {
-                auto pathResult = svgPathToTvgPath(node->node.path.path->c_str());
-                shape->appendPath(get<0>(pathResult).data(), get<0>(pathResult).size(), get<1>(pathResult).data(), get<1>(pathResult).size());
+                if (svgPathToTvgPath(node->node.path.path->c_str(), cmds, pts))
```
JSUYA, line 292: I think it could check the results of cmds and pts.
svgPathToTvgPath(node->node.path.path->c_str(), cmds, pts));
if (cmds && pts)

Issue #180, opened 2020-12-09
Crash in _rasterSolidRle and _rasterTranslucentRle API


mihashco:
```
0x00007ffff74d3595 in _rasterSolidRle (surface=0x555555610c00, rle=0x7fffdc001040, color=4294836224) at ../src/lib/sw_engine/tvgSwRaster.cpp:252
252	                dst[i] = src + ALPHA_BLEND(dst[i], ialpha);
(gdb) bt
#0  0x00007ffff74d3595 in _rasterSolidRle (surface=0x555555610c00, rle=0x7fffdc001040, color=4294836224) at ../src/lib/sw_engine/tvgSwRaster.cpp:252
#1  0x00007ffff74d47d5 in rasterStroke (surface=0x555555610c00, shape=0x5555555df680, r=254 '\376', g=0 '\000', b=0 '\000', a=255 '\377') at ../src/lib/sw_engine/tvgSwRaster.cpp:481
#2  0x00007ffff74d53bf in tvg::SwRenderer::render (this=0x55555562acd0, shape=..., data=0x5555555df5e0) at ../src/lib/sw_engine/tvgSwRenderer.cpp:391
#3  0x00007ffff74c9e77 in tvg::Shape::Impl::render (this=0x5555555de7d0, renderer=...) at ../src/lib/tvgShapeImpl.h:223
#4  0x00007ffff74cac5b in tvg::PaintMethod<tvg::Shape::Impl>::render (this=0x55555562fab0, renderer=...) at ../src/lib/tvgPaint.h:225
#5  0x00007ffff74bef9e in tvg::Paint::Impl::render (this=0x55555562fa80, renderer=...) at ../src/lib/tvgPaint.h:169
#6  0x00007ffff74c719c in tvg::Scene::Impl::render (this=0x55555562d690, renderer=...) at ../src/lib/tvgSceneImpl.h:81
#7  0x00007ffff74c7baf in tvg::PaintMethod<tvg::Scene::Impl>::render (this=0x55555562d6c0, renderer=...) at ../src/lib/tvgPaint.h:225
#8  0x00007ffff74bef9e in tvg::Paint::Impl::render (this=0x55555562d660, renderer=...) at ../src/lib/tvgPaint.h:169
#9  0x00007ffff74bf49f in tvg::Canvas::Impl::draw (this=0x55555562ad00) at ../src/lib/tvgCanvasImpl.h:101
[...]

PR #181, opened 2020-12-09
sw_engine raster: Fixed crash in raster API

mihashco:
Crash occurs when surface is smaller than span. There is no need to render RLE outside of surface.

@Issues: 
#180

### DIFF
--- a/src/lib/tvgSwCanvas.cpp
+++ b/src/lib/tvgSwCanvas.cpp
@@ -66,6 +66,7 @@ Result SwCanvas::target(uint32_t* buffer, uint32_t stride, uint32_t w, uint32_t 
     if (!renderer) return Result::MemoryCorruption;
 
     if (!renderer->target(buffer, stride, w, h, cs)) return Result::InvalidArguments;
+    Canvas::pImpl->invalidate();
 
     return Result::Success;
 #endif
--- a/src/examples/Capi.cpp
+++ b/src/examples/Capi.cpp
@@ -9,13 +9,14 @@
 /* Capi Test Code                                                       */
 /************************************************************************/
 
-static uint32_t buffer[WIDTH * HEIGHT];
+static uint32_t* buffer = NULL;
+static Tvg_Canvas* canvas = NULL;
+static Eo* view = NULL;
 
 void testCapi()
 {
-    tvg_engine_init(TVG_ENGINE_SW | TVG_ENGINE_GL, 0);
 
-    Tvg_Canvas* canvas = tvg_swcanvas_create();
+    canvas = tvg_swcanvas_create();
     tvg_swcanvas_set_target(canvas, buffer, WIDTH, WIDTH, HEIGHT, TVG_COLORSPACE_ARGB8888);
 
     Tvg_Paint* shape = tvg_shape_new();
@@ -170,10 +171,6 @@ void testCapi()
     tvg_canvas_push(canvas, scene);
     tvg_canvas_draw(canvas);
     tvg_canvas_sync(canvas);
-
-    tvg_canvas_destroy(canvas);
-
-    tvg_engine_term(TVG_ENGINE_SW | TVG_ENGINE_GL);
 }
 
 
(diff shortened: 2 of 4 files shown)


hermet:
There must be something wrong earlier. This is not efficient if handling exception case every loop.
Bounding size should be aligned in... => rle construction or bounding box.
Need to investment for the proper approach.

mihashco:
@hermet Ok agree. I will try to find better solution.

src/examples/CapiResize.cpp
```diff
@@ -0,0 +1,125 @@
...
+    if ((w != WIDTH || h != HEIGHT) && (w != 0 && h != 0))
+    {
+      buffer = (uint32_t*) realloc(buffer, sizeof(uint32_t) * w * h);
+
+      tvg_swcanvas_set_target(canvas, buffer, w, w, h, TVG_COLORSPACE_ARGB8888);
+
+      tvg_canvas_clear(canvas, false);
+      tvg_scene_clear(scene);
+
+      tvg_canvas_push(canvas, scene);
```
hermet, line 65: scene is already pushed in the canvas...

src/examples/CapiResize.cpp
```diff
@@ -0,0 +1,125 @@
...
+    tvg_shape_append_rect(shape, 100, 0, 100, 100, 0, 0);
+    tvg_shape_set_stroke_color(shape, 255, 0, 0, 255);
+    tvg_shape_set_stroke_width(shape, 1);
+
+    buffer = (uint32_t*) malloc(sizeof(uint32_t) * WIDTH * HEIGHT);
+    tvg_swcanvas_set_target(canvas, buffer, WIDTH, WIDTH, HEIGHT, TVG_COLORSPACE_ARGB8888);
+
+    tvg_canvas_clear(canvas, false);
+    tvg_scene_clear(scene);
+    tvg_canvas_push(canvas, scene);
```
hermet, line 39: it's recommended to push the scene into canvas after paint is ready to update.
you can push the scene to canvas after you pushed all subsequent paints into the scene.
I found an issue that scene doesn't update any RenderUpdateFlag for this (need to fix it), but this is unlikely to occur any crash issues.

hermet (changes requested):

src/examples/CapiResize.cpp
hermet: Added one issue ticket... https://github.com/Samsung/thorvg/issues/200

hermet (changes requested):

src/examples/CapiResize.cpp
hermet: No... read again... scene doesn't have any issues. forget RenderUpdateFlag stuff here.

hermet (changes requested):

src/lib/tvgSwCanvas.cpp
```diff
@@ -67,6 +67,12 @@ Result SwCanvas::target(uint32_t* buffer, uint32_t stride, uint32_t w, uint32_t
 
     if (!renderer->target(buffer, stride, w, h, cs)) return Result::InvalidArguments;
 
+    Array<RenderData> clips;
```
hermet, line 70: how about this? this could supply both gl, sw engines. also more efficient than your version.

SwCanvas::target() {
   ...
   Canvas::pImpl->invalidate();
}

Canvas::Impl::invalidate()
{
   foreach(...) paint->pImpl->flags = RenderUpdate::All
}

src/examples/CapiResize.cpp
hermet: Can we merge this _resize_cb to Capi.cpp? (to not add new file)

You don't need to do nothing for paints here but just call reset target & redraw to guarantee the issue.

static void _resize_cb(..)
{
    int w = 0, h = 0;
    evas_object_geometry_get(obj, NULL, NULL, &w, &h);

    if ((w != WIDTH || h != HEIGHT) && (w != 0 && h != 0))
    {
      buffer = (uint32_t*) realloc(buffer, sizeof(uint32_t) * w * h);
      tvg_swcanvas_set_target(...);
      tvg_swcanvas_update(canvas);
      tvg_swcanvas_draw();
      tvg_swcanvas_sync();
      ...
}

inc/thorvg.h
```diff
@@ -28,6 +28,7 @@ protected: \
 
 #define _TVG_DECLARE_ACCESSOR() \
     friend Canvas; \
+    friend SwCanvas; \
```
hermet, line 31: --

inc/thorvg.h
```diff
@@ -43,6 +44,7 @@ class RenderMethod;
 class Scene;
 class Picture;
 class Canvas;
+class SwCanvas;
```
hermet, line 47: --

hermet (changes requested):
Please check a comment.

PR #184, opened 2020-12-14
svg_loader XmlParser: Print unsupported elements, attribute with log option

JSUYA:
- Description :
* Using printf is temporary. We are planning a proper way to print the log.
When parsing a Svg file, Loader print unsupported elements and attributes.

- Issue Tickets (if any) :
https://github.com/Samsung/thorvg/issues/152

- Tests or Samples (if any) :
[Enable log]
meson_option -> add 'debug' value in 'log' option

- Screen Shots (if any) :

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -28,6 +28,10 @@ if get_option('bindings').contains('capi') == true
     config_h.set10('THORVG_CAPI_BINDING_SUPPORT', true)
 endif
 
+if get_option('log') == true
+    config_h.set10('THORVG_LOG_ENABLED', true)
+endif
+
 configure_file(
     output: 'config.h',
     configuration: config_h
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -37,3 +37,8 @@ option('test',
    type: 'boolean',
    value: false,
    description: 'Enable building unit tests')
+
+option('log',
+    type: 'boolean',
+    value: false,
+    description: 'Enable log message')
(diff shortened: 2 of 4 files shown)


meson_options.txt
```diff
@@ -37,3 +37,9 @@ option('test',
    type: 'boolean',
    value: false,
    description: 'Enable building unit tests')
+
+option('log',
+   type: 'array',
+   choices: ['', 'info', 'debug', 'warrning', 'error'],
```
hermet, line 43: 1. we can go simply true/false,  not detailed option.

meson.build
```diff
@@ -28,6 +28,10 @@ if get_option('bindings').contains('capi') == true
     config_h.set10('THORVG_CAPI_BINDING_SUPPORT', true)
 endif
 
+if get_option('log').contains('debug') == true
+    config_h.set10('THORVG_LOG_DEBUG', true)
```
hermet, line 32: THORVG_LOG_ENABLED

hermet (changes requested):
Please check comments.

meson_options.txt
JSUYA: Changed boolean type

meson.build
JSUYA: Changed

hermet (approved):

hermet (approved):

PR #189, opened 2020-12-17
svg_loader SvgLoader: Print inefficient elements

JSUYA:
- Description :
Opacity is 0
both Fill.Opacity and Stroke Opacity are 0
point is 0 in Path
declared display="none"
width or height of bounds becomes 0

- Issue Tickets (if any) :
https://github.com/Samsung/thorvg/issues/153

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2225,9 +2225,49 @@ static void _styleInherit(SvgStyleProperty* child, SvgStyleProperty* parent)
 }
 
 
+#ifdef THORVG_LOG_ENABLED
+static void _inefficientNodeCheck(SvgNode* node){
+    if (!node->display) printf("SVG: Inefficient elements used [Display is none][Node Type : %s]\n", _nodeTypeToString(node->type).c_str());
+    if (node->style->opacity == 0) printf("SVG: Inefficient elements used [Opacity is zero][Node Type : %s]\n", _nodeTypeToString(node->type).c_str());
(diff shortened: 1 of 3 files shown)


src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -347,4 +347,33 @@ struct SvgLoaderData
     bool result = false;
 };
 
+#ifdef THORVG_LOG_ENABLED
+inline string nodeTypeToString(SvgNodeType type)
```
hermet, line 351: does it necessary to use inline? leave to compiler if possible
Also, if it uses only in _inefficientNodeCheck(), you can move it in one same place.

#ifdef THORVG_LOG_ENABLED
static string _nodeTypeToString()
{
...
}
static void _inefficientNodeCheck()
{
...
}
#endif

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2225,9 +2225,49 @@ static void _styleInherit(SvgStyleProperty* child, SvgStyleProperty* parent)
 }
 
 
+#ifdef THORVG_LOG_ENABLED
+void _inefficientNodeCheck(SvgNode* node){
```
hermet, line 2229: static void

hermet (changes requested):
Please check comments

src/loaders/svg/tvgSvgLoaderCommon.h
JSUYA: Removed inline
and _nodeTypeToString is also used in XmlParser.cpp. Therefore, it cannot be moved to tvgSvgLoade.

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: Changed

hermet (approved):

PR #217, opened 2021-01-21
sw_engine math: fixing matrix transformation

mgrudzinska:
- Description :
Unnecessary rounding during matrix transformation has been removed.
The problem occured when scaling a shape with a dashed stroke.

- Issue Tickets (if any) :

- Tests or Samples (if any) :
https://github.com/mgrudzinska/thorvg/tree/mgrudzinska/Scaled_dashed_stroke_example
ScaledShape.cpp

- Screen Shots (if any) :
before changes:
image

after changes:
image

### DIFF
--- a/src/lib/sw_engine/tvgSwMath.cpp
+++ b/src/lib/sw_engine/tvgSwMath.cpp
@@ -422,8 +422,8 @@ SwPoint mathTransform(const Point* to, const Matrix* transform)
 {
     if (!transform) return {TO_SWCOORD(to->x), TO_SWCOORD(to->y)};
 
-    auto tx = round(to->x * transform->e11 + to->y * transform->e12 + transform->e13);
-    auto ty = round(to->x * transform->e21 + to->y * transform->e22 + transform->e23);
+    auto tx = to->x * transform->e11 + to->y * transform->e12 + transform->e13;
+    auto ty = to->x * transform->e21 + to->y * transform->e22 + transform->e23;
 
     return {TO_SWCOORD(tx), TO_SWCOORD(ty)};
 }


JSUYA (approved):

hermet:
@mgrudzinska  This might occur side effects by dropping decimal values.
Need to verify with svg test cases precisely ...
especially (example/Svg2) . please make it sure 1 pixel swift or lack of 1 pixel filling.
Could you please confirm this?

mgrudzinska:
@hermet I double checked and I think, that we shouldn't use the round() function in this place.
The reason of that is illustrated on the picture below.

The red rectangles (almost completely covered) are plotted without any translation. I shifted them from the original positions by adding a constant to their coordinates directly in the path (for the upper one I added 0.49 to each coordinate and for the lower one 0.51).

The corresponding green rectangles I shifted using the translate() function: translate(0.49,0.49) for the upper one and translate(0.51,0.51) for the lower one. 

The expected result is that the green rectangles completely cover the red ones. But they doesn't. It is because of the round() function that is used when the translate() is used. For shapes without translation the values are redirected to the SW_COORD() function without any changes. And when we use translate() they are first rounded. 

rounded

hermet:
@mgrudzinska basically, here round() is necessary for more accurate rendering.
What do you think if user tries to move a point to (0.9, 0.9) 
will you show it at (0, 0) ? or (1, 1)? 
maybe red rectangle is incorrect than green one. (if it's an issue.)

Also, i can't reproduce that issue(Scaled_dashed_stroke_example) here without your correction.
Even I can't imaging how that round() occurs those rendering result...

hermet (changes requested):
please check a comment.

mgrudzinska:
> What do you think if user tries to move a point to (0.9, 0.9)
> will you show it at (0, 0) ? or (1, 1)?
but these coordinates must be converted into SWCOORD.

Suppose we have a point (0, 0) that we want to shift (translate) by (0.9, 0.9).
The mathTransform() (tvgSwMath.cpp) will be called, which will set tx = 0.9 and ty = 0.9. They will be rounded to tx = 1 and ty = 1. Then they will be converted to SWCOORD (multiplied by 64 and cast to signed long: 1 * 64 = 64 -> 64) and we get the point (64.64).

If we remove round(), then the value of 0.9 will be converted by TO_SWCOORD into 57 (0.9 * 64 = 57.6 -> 57). So finally we get a point (57.57), which was, I believe,  our intention. Am I wrong?

Please notice that I have not removed any round() from the raster functions. 


> Also, i can't reproduce that issue(Scaled_dashed_stroke_example) here without your correction.
The behavior does not change - the result from "ScaledShape.cpp" still looks like the first picture in this PR.

hermet:
> > What do you think if user tries to move a point to (0.9, 0.9)
> > will you show it at (0, 0) ? or (1, 1)?
> > but these coordinates must be converted into SWCOORD.
> 
> Suppose we have a point (0, 0) that we want to shift (translate) by (0.9, 0.9).
> The mathTransform() (tvgSwMath.cpp) will be called, which will set tx = 0.9 and ty = 0.9. They will be rounded to tx = 1 and ty = 1. Then they will be converted to SWCOORD (multiplied by 64 and cast to signed long: 1 * 64 = 64 -> 64) and we get the point (64.64).
> 
> If we remove round(), then the value of 0.9 will be converted by TO_SWCOORD into 57 (0.9 * 64 = 57.6 -> 57). So finally we get a point (57.57), which was, I believe, our intention. Am I wrong?
> 
> Please notice that I have not removed any round() from the raster functions.
> 
> > Also, i can't reproduce that issue(Scaled_dashed_stroke_example) here without your correction.
> > The behavior does not change - the result from "ScaledShape.cpp" still looks like the first picture in this PR.

@mgrudzinska I think you're right. thanks for your contribution!

PR #219, opened 2021-01-26
sw_engine renderer: rendering optimization of shapes without strokes

mgrudzinska:
For solid shapes without strokes, the value of 'a' was incorrectly assigned,
which resulted in unnecessary calls to raster functions.
The variable holding the value 'a' is now reset before being used again.

### DIFF
--- a/src/lib/sw_engine/tvgSwRenderer.cpp
+++ b/src/lib/sw_engine/tvgSwRenderer.cpp
@@ -318,15 +318,16 @@ bool SwRenderer::renderShape(RenderData data)
 
     if (auto fill = task->sdata->fill()) {
         rasterGradientShape(surface, &task->shape, fill->id());
-    } else{
+    } else {
         task->sdata->fillColor(&r, &g, &b, &a);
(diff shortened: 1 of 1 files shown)


src/lib/sw_engine/tvgSwRenderer.cpp
```diff
@@ -324,6 +324,7 @@ bool SwRenderer::renderShape(RenderData data)
         if (a > 0) rasterSolidShape(surface, &task->shape, r, g, b, a);
     }
 
+    a = 0;
```
JSUYA, line 327: If you want to determine whether or not a stroke is created, I think it is good to check the return value of strokeColor for the readability of the code.

How about this
```cpp
if (task->sdata->strokeColor(&r, &g, &b, &a) == Result::Success)
```

mgrudzinska: yours looks more elegant.
are we ok with an extra if statement?

mgrudzinska: changed according to your recommendation

PR #222, opened 2021-01-29
canvas, capi: expose force update flag.

mihashco:
Changes:
1. Force flag is exposed to capi bindings. It is necessary to improve of
stability of thorvg clients. In some cases (after buffer change) it is
necessary to re-render shapes.
2. Fixed issue with stroke rle reset

@API-Changes:
from: tvg_canvas_update(Tvg_Canvas* canvas);
to: tvg_canvas_update(Tvg_Canvas* canvas, bool force);

### DIFF
--- a/inc/thorvg_capi.h
+++ b/inc/thorvg_capi.h
@@ -398,7 +398,7 @@ TVG_EXPORT Tvg_Result tvg_canvas_clear(Tvg_Canvas* canvas, bool free);
 * - TVG_RESULT_SUCCESS: if ok.
 * - TVG_RESULT_INVALID_PARAMETERS: if canvas is invalid]
 */
-TVG_EXPORT Tvg_Result tvg_canvas_update(Tvg_Canvas* canvas);
+TVG_EXPORT Tvg_Result tvg_canvas_update(Tvg_Canvas* canvas, bool force);
 
 
 /*!
--- a/src/lib/sw_engine/tvgSwShape.cpp
+++ b/src/lib/sw_engine/tvgSwShape.cpp
@@ -465,6 +465,7 @@ void shapeDelOutline(SwShape* shape, uint32_t tid)
 void shapeReset(SwShape* shape)
 {
     rleReset(shape->rle);
+    rleReset(shape->strokeRle);
     shape->rect = false;
     _initBBox(shape->bbox);
 }
--- a/src/lib/tvgRender.h
+++ b/src/lib/tvgRender.h
@@ -44,7 +44,7 @@ struct Compositor {
     uint32_t        opacity;
 };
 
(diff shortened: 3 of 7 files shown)


JSUYA:
I am not against this code change. I recommend splitting these two problems into two commits.

src/lib/sw_engine/tvgSwShape.cpp
```diff
@@ -465,6 +465,7 @@ void shapeDelOutline(SwShape* shape, uint32_t tid)
 void shapeReset(SwShape* shape)
 {
     rleReset(shape->rle);
+    rleReset(shape->strokeRle);
```
hermet, line 468: you can raise a patch for this change with another one.

src/lib/tvgRender.h
```diff
@@ -44,7 +44,7 @@ struct Compositor {
     uint32_t        opacity;
 };
 
-enum RenderUpdateFlag {None = 0, Path = 1, Color = 2, Gradient = 4, Stroke = 8, Transform = 16, Image = 32, All = 64};
+enum RenderUpdateFlag {None = 0, Path = 1, Color = 2, Gradient = 4, Stroke = 8, Transform = 16, Image = 32, All = 63};
```
hermet, line 47: you can raise a patch for this change with another one.

inc/thorvg_capi.h
```diff
@@ -398,7 +398,7 @@ TVG_EXPORT Tvg_Result tvg_canvas_clear(Tvg_Canvas* canvas, bool free);
 * - TVG_RESULT_SUCCESS: if ok.
 * - TVG_RESULT_INVALID_PARAMETERS: if canvas is invalid]
 */
-TVG_EXPORT Tvg_Result tvg_canvas_update(Tvg_Canvas* canvas);
+TVG_EXPORT Tvg_Result tvg_canvas_update(Tvg_Canvas* canvas, bool force);
```
hermet, line 401: User have no idea what happens update() internally, it brings user wonder why force is necessary.
I think thorvg can conditionally make it in smarter way.

ex)
1. CanvasImple could have a boolean such as forceUpdateAll;
2. when target is changed, it set forceUpdateAll = true
3. canvas update() is called, override flag with forceUpdateAll update paints 
4. reset forceUpdateAll in draw() after drawing paints.

hermet (changes requested):
please check comments.

src/lib/tvgRender.h
mihashco: Done: please check #226

src/lib/sw_engine/tvgSwShape.cpp
mihashco: done. Please check #228

inc/thorvg_capi.h
mihashco: Please check #229

PR #223, opened 2021-02-01
common render: removed unnecessary assignments in update() fun

mgrudzinska:
The rotation part of the transformation matrix can be assigned,
it does not have to be multiplied by the identity matrix.
The values of e13 and e23 of the transformation matrix represents the translation,
so they don't need to be taken into account in the rotation part.

### DIFF
--- a/src/lib/tvgRender.cpp
+++ b/src/lib/tvgRender.cpp
@@ -68,8 +68,8 @@ bool RenderTransform::update()
     m.e33 = 1.0f;
 
     //scale
-    m.e11 *= scale;
-    m.e22 *= scale;
+    m.e11 = scale;
+    m.e22 = scale;
 
     //rotation
     if (fabsf(degree) > FLT_EPSILON) {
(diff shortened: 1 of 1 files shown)


JSUYA (approved):
LGTM

JSUYA (approved):
LGTM x2

hermet:
hmm.. this looks wrong to me. position must be transformed by rotation so that we can support accumulated transform during scene-graph... need to verify it further.

hermet (changes requested):
looks wrong to me. I will verify it later.

JSUYA:
> looks wrong to me. I will verify it later.

matrix m is initialized with an identity matrix.
This patch has reduced unnecessary computation.

example)
auto t11 = m.e11 * cosVal + m.e12 * sinVal;
auto t12 = m.e11 * -sinVal + m.e12 * cosVal;
...
m.e12 = t12;
->

m.e11 is 1. and m.e12 is 0.
So..
auto t11 = 1 * cosVal + 0 * sinVal;
auto t12 = 1 * -sinVal + 0 * cosVal;

Therefore, the results of both expressions are the same.

m.e12 = m.e11 * -sinVal;
m.e11 *= cosVal;

The matrix m and the cumulative transformation of this function are not related.
Am I misunderstanding?

hermet:
> > looks wrong to me. I will verify it later.
> 
> matrix m is initialized with an identity matrix.
> This patch has reduced unnecessary computation.
> 
> example)
> auto t11 = m.e11 * cosVal + m.e12 * sinVal;
> auto t12 = m.e11 * -sinVal + m.e12 * cosVal;
> ...
> m.e12 = t12;
> ->
> 
> m.e11 is 1. and m.e12 is 0.
> So..
> auto t11 = 1 * cosVal + 0 * sinVal;
> auto t12 = 1 * -sinVal + 0 * cosVal;
> 
> Therefore, the results of both expressions are the same.
> 
> m.e12 = m.e11 * -sinVal;
> m.e11 *= cosVal;
> 
> The matrix m and the cumulative transformation of this function are not related.
> Am I misunderstanding?

See how logic is different.

@ before
  auto t13 = m.e13 * cosVal + m.e23 * sinVal;
  m.e13 = t13;
  m.e13 += x;

@after 
  m.e13 = x;

hermet (approved):

hermet:
verified no any side effect.

PR #226, opened 2021-02-02
renderer: Fixed RenderUpdate::All flag fields.

mihashco:
- Description :
All flag should have all bits set to 1. If flag = 64 rendering task don't update anything, because all checked bits in flag are not set.


hermet (approved):

hermet:
good catch, thanks!

PR #227, opened 2021-02-02
sw_engine raster: removing unnecessary logical 'and' operation

mgrudzinska:
Shifting the 32-bit number by 24 bits leaves only 8 bits,
so there is no need to mask them with 0xFF.


hermet (approved):
thanks!

PR #229, opened 2021-02-03
sw canvas: Added forceUpdateAll flag in canvas.

mihashco:
Force update flag value is set after target change. It is necessary
to rerender shapes (gen its RLE) when target buffer size was modified.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -43,6 +43,7 @@ class RenderMethod;
 class Scene;
 class Picture;
 class Canvas;
+class SwClass;
 
 
 enum class TVG_EXPORT Result { Success = 0, InvalidArguments, InsufficientCondition, FailedAllocation, MemoryCorruption, NonSupport, Unknown };
@@ -152,6 +153,8 @@ public:
     virtual Result sync() noexcept;
 
     _TVG_DECLARE_PRIVATE(Canvas);
+
+    friend class SwCanvas;
 };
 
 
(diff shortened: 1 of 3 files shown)


inc/thorvg.h
```diff
@@ -43,6 +43,7 @@ class RenderMethod;
 class Scene;
 class Picture;
 class Canvas;
+class SwClass;
```
hermet, line 47: what is this?

inc/thorvg.h
```diff
@@ -152,6 +153,8 @@ class TVG_EXPORT Canvas
     virtual Result sync() noexcept;
 
     _TVG_DECLARE_PRIVATE(Canvas);
+
+    friend class SwCanvas;
```
hermet, line 157: SwCanvas is a derived, base class should not depend on it...

hermet (changes requested):

inc/thorvg.h
mihashco: This is needed for friendship

inc/thorvg.h
mihashco: without friendship I'm not able to switch flag in target API in SwClass. 


```c
Result SwCanvas::target(uint32_t* buffer, uint32_t stride, uint32_t w, uint32_t) {
//...
Canvas::pImpl->forceUpdateAll = true;
}
```

mihashco: If you don't like this solution, I can add virtual Canvas::target() method, but in this case it would be great to change also tvg_swcanvas_create() -> tvg_canvas_create() and tvg_swcanvas_set_target() to tvg_canvas_set_target() (this solution breaks backward compatibility). 

Maybe do you know better option ? I can consider other possibilities if you want

hermet:
We can fix by this => https://github.com/Samsung/thorvg/pull/256

PR #230, opened 2021-02-03
sw_engine common: improving the accuracy of calculations in the ALPHA…

mgrudzinska:
…_MULTIPLY() fun.

When multiplying the alpha values the division by 256 was performed (in form
of arithmetic shift >> 8), while the exact result required division by 255.
An expression was added to correct the result.

### DIFF
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -263,7 +263,7 @@ static inline uint32_t COLOR_INTERPOLATE(uint32_t c1, uint32_t a1, uint32_t c2, 
 
 static inline uint8_t ALPHA_MULTIPLY(uint32_t c, uint32_t a)
 {
-    return (c * a) >> 8;
+    return (c * a + (c * a >> 8) + 0x80) >> 8;
 }
 
 static inline SwCoord HALF_STROKE(float width)


src/lib/sw_engine/tvgSwCommon.h
```diff
@@ -263,7 +263,7 @@ static inline uint32_t COLOR_INTERPOLATE(uint32_t c1, uint32_t a1, uint32_t c2,
 
 static inline uint8_t ALPHA_MULTIPLY(uint32_t c, uint32_t a)
 {
-    return (c * a) >> 8;
+    return (c * a + (c * a >> 8) + 0x80) >> 8;
```
JSUYA, line 266: How about this solution? 
(https://stackoverflow.com/questions/45041273/how-to-correctly-multiply-two-colors-with-byte-components)
c = 255
a = 255
(255 * 255) >> 8 = 1111 1110
(255 * 255 + (255 * 255 >> 8) + 0x80) >> 8
=  (1111 1110 0000 0001 + 1111 1110 + 1000 0000) >> 8
= 1111 1111 0111 1111 >> 8
= 1111 1111
(a * b + 0xFF) >> 8 
= 1111 1110 0000 0001 + 1111 1111
= 1111 1111 0000 0000 >> 8
= 1111 1111

JSUYA (commented):
Please check my comment

src/lib/sw_engine/tvgSwCommon.h
mgrudzinska: it depends on how precise we want to be and at what cost.

- your solution has one less multiplication operation. I can possibly save the result of the multiplication in a variable and use it, but either way it's more expensive than yours.
- mine solution is accurate - there are about 20 cases where dividing on floats would be x.502 and my solution would return x, not x + 1
- result from your formula is 1 higher the the exact result in about 30% of the cases (out of all over 65k)
e.g. c = a = 1

mine:
(0x1 + 0x0 + 1000 0000) >> 8
= 1000 0001 >> 8
= 0

yours:
(0x1 + 1111 1111
= 0000 0001 0000 0000 >> 8
= 1

I don't see any significant difference in time consumption between our formulas. Do you?

JSUYA: Sorry i don't understand... In this case
>e.g. c = a = 1
>mine:
>(0x1 + 0x0 + 1000 0000) >> 8
>= 1000 0001 >> 8
>= 0
the result is 0, isn't it wrong? What we're expecting is c * a, isn't it more accurate that the result is 1?

mgrudzinska: but following this logic we have:
1) c=255 and a=1 -> c*a=255
2) c=255 and a=255 -> c*a=65025

we cannot think of it as simple multiplication, but rather as scaling, where the scaling factor is 'a/255'
so we have:
ad 1) ALPHA_MULTIPLY(255,1) -> 255 * (1/255) = 255*0 = 0
ad 2) ALPHA_MULTIPLY(255,255) -> 255 * (255/255) = 255*1 = 255

do you agree?

JSUYA: oh I got it wrong. When c = 1 a = 1, the result is 0. What we expect is 1/255 x 1/255.
you are right.

JSUYA (approved):

hermet:
@mgrudzinska We can compromise between trivial loss and performance. 
Could you please share the result  how's actual rendering is different?

hermet (changes requested):
That logic has been used many years for several products in Tizen. 
We can compromise the performance and trivial loss unless it practically affect the actual result.
Thus, it would be better If you bring any comparison image between (before & after)

hermet:
https://github.com/Samsung/thorvg/pull/319/files

PR #231, opened 2021-02-03
sw_engine common: improving the alpha blending algorithm

mgrudzinska:
Calculations accuracy in ALPHA_BLEND function has been
improved. Until now blending resulted in a slight hue change
(all color channels affected).

### DIFF
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -261,13 +261,8 @@ static inline SwCoord TO_SWCOORD(float val)
 
 static inline uint32_t ALPHA_BLEND(uint32_t c, uint32_t a)
 {
-    uint32_t ag = ((c >> 8) & 0x00ff00ff) * a;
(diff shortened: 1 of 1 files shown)


hermet:
@mgrudzinska This could affect eye-catchable performance drop.

Could you please compare test and share the result how it improve the image result? 

I don't like to apply this unless clear enhancement. (saving it for performance)

hermet (changes requested):
@mgrudzinska This could affect eye-catchable performance drop.

Could you please compare test and share the result how it improves the image result?

I don't like to apply this unless clear enhancement. (saving it for performance)

mgrudzinska:
I'm sorry it took so long.

Please check the image that you get from the code:
```

mgrudzinska:
@hermet please check now. This is the method you used in the ALPHA MULTIPLY function.
I don't see any difference between the exact calculation and this approximation, so in my opinion it is sufficient.

However, if you think this change is unnecessary, please close this PR.

hermet:
@mgrudzinska here is my approach => https://github.com/Samsung/thorvg/pull/386

hermet (approved):

PR #232, opened 2021-02-04
sw_engine raster: checking zero division while inverting the transfor…

mgrudzinska:
…orm matrix

In case the user has provided the transformation matrix, it is possible that
the inverse matrix does not exist.

### DIFF
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -46,12 +46,14 @@ static uint32_t _argbJoin(uint8_t r, uint8_t g, uint8_t b, uint8_t a)
 }
 
 
-static void _inverse(const Matrix* transform, Matrix* invM)
+static bool _inverse(const Matrix* transform, Matrix* invM)
 {
     // computes the inverse of a matrix m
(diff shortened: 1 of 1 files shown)


src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -46,13 +46,15 @@ static uint32_t _argbJoin(uint8_t r, uint8_t g, uint8_t b, uint8_t a)
...
 
-static void _inverse(const Matrix* transform, Matrix* invM)
+static bool _inverse(const Matrix* transform, Matrix* invM)
 {
     // computes the inverse of a matrix m
     auto det = transform->e11 * (transform->e22 * transform->e33 - transform->e32 * transform->e23) -
                transform->e12 * (transform->e21 * transform->e33 - transform->e23 * transform->e31) +
                transform->e13 * (transform->e21 * transform->e32 - transform->e22 * transform->e31);
 
+    if (abs(det) < FLT_EPSILON) return false;
```
JSUYA, line 56: det seems to be a float type. fabsf?

mgrudzinska: changed, thx

PR #233, opened 2021-02-04
common: comparison of floats using abs function

mgrudzinska:
The function abs() was used to compare the elements of the matrix
with other values instead of the operators == and !=.


JSUYA (approved):

hermet:
@mgrudzinska It's a trick for less logic. Considering common case, origin logic is enough to go.

hermet (changes requested):
@mgrudzinska Obviously, that logic is intended. It's a sort of trick for less logic. 
Unless any issues due to this, we don't need to apply this.

PR #235, opened 2021-02-04
common Scene: Fix translucent composition condition

JSUYA:
- Description :
If a scene has another scene as a child and
the number of children is 1, the composition of the child does not work.
Therefore, fix that composition works as long as the number of children is not 0.

- Tests or Samples (if any) :
```cpp
    auto rootScene = tvg::Scene::gen();
    rootScene->opacity(255);

    //Create a Scene
    auto childScene = tvg::Scene::gen();
    childScene->opacity(175);              //Apply opacity to scene (0 - 255)
    childScene->reserve(2);

    //Prepare Circle
    shape1->appendCircle(400, 400, 250, 250);
    shape1->fill(255, 255, 0, 255);
    childScene->push(move(shape1));

    rootScene->push(move(childScene));

    canvas->push(move(rootScene));

```

### DIFF
--- a/src/lib/tvgSceneImpl.h
+++ b/src/lib/tvgSceneImpl.h
@@ -65,7 +65,7 @@ struct Scene::Impl
         Compositor* cmp = nullptr;
 
         //Half translucent. This condition requires intermediate composition.
-        if ((opacity < 255 && opacity > 0) && (paints.count > 1)) {
+        if ((opacity < 255 && opacity > 0) && (paints.count > 0)) {
             uint32_t x, y, w, h;
             if (!bounds(renderer, &x, &y, &w, &h)) return false;
             cmp = renderer.target(x, y, w, h);


src/lib/tvgSceneImpl.h
```diff
@@ -65,7 +65,7 @@ struct Scene::Impl
         Compositor* cmp = nullptr;
 
         //Half translucent. This condition requires intermediate composition.
-        if ((opacity < 255 && opacity > 0) && (paints.count > 1)) {
+        if ((opacity < 255 && opacity > 0) && (paints.count > 0)) {
```
hermet, line 68: if the scene has just one simple share, composition is not required, 
it would be better if condition check whether child is another scene or not.

JSUYA: Is there a way to know if a child is a scene or another shape or picture?
Since we are using the `fno-rtti` option, we cannot use `dynamic_cast` and class `typeid`
So there seems to be no way to check the child's class type. is that right?
I don't think of a good way.
Another way is to create a type get function.
```
enum PaintType {Shape, Scene, Picture};
PaintType type(Paint*);
```
Please share any good idea.

hermet: @JSUYA I think we can put a type info(with accessing it) in the internal Paint member.

PR #236, opened 2021-02-05
sw_engine raster: fixing rasterization of an image with InvMask

mgrudzinska:
In case of the image rasterization with an inverse mask the opacity
was omitted and the alpha value instead of the inverse alpha value was used.

The result of the InvMasking:
expected (see comment below):
invMaskOK

current (wrong):
invMaskWrong


mgrudzinska:
translucent inv masking does not work correctly

mgrudzinska:
this fixes the issue for the solid inv mask (a=255)
I think that the transparent inv mask should work in a bit different way - needs to be done

patrykka:
In my opinion this fix works fine.

hermet (approved):
@mgrudzinska  great job!

Issue #206, opened 2021-01-10
Optimization Image Raster


hermet:
Image rasterization still unnecessary transform pixels.
We can avoid this precisely if check the transform matrix

for example, transform has only translation, we can just shift image origin position without transform.

see OPTIMIZE ME comments in tvgSwRaster.cpp rasterImage()

mgrudzinska:
https://github.com/Samsung/thorvg/pull/239

PR #239, opened 2021-02-08
sw_engine raster: image rastering optimization

mgrudzinska:
Adding functions for rastering the translated images (no rotation and scaling).

This pr contains 3 commits for readabilty.

I introduced changes reducing the number of multiplications performed in loops.
I changed the order of the functions, sticking to the rule (for each block): first identity, then for translated images and finally for general transformations.
In the last commit I introduced all functions for rastering the translated images.

- Issue Tickets (if any) : #206 

- Tests or Samples :  draft version of OptRasterTest - calls all raster functions for images
https://github.com/mgrudzinska/thorvg/tree/mgrudzinska/Test_image_raster_opr_trans

### DIFF
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -332,8 +332,8 @@ static bool _rasterTranslucentImageRle(SwSurface* surface, const SwRleData* rle,
     for (uint32_t i = 0; i < rle->size; ++i, ++span) {
         auto dst = &surface->buffer[span->y * surface->stride + span->x];
         auto src = img + span->x + span->y * w;    //TODO: need to use image's stride
+        auto alpha = ALPHA_MULTIPLY(span->coverage, opacity);
         for (uint32_t x = 0; x < span->len; ++x, ++dst, ++src) {
-            auto alpha = ALPHA_MULTIPLY(span->coverage, opacity);
             *src = ALPHA_BLEND(*src, alpha);
             *dst = *src + ALPHA_BLEND(*dst, 255 - surface->blender.alpha(*src));
         }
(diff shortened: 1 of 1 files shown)


src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -71,17 +71,21 @@ static bool _inverse(const Matrix* transform, Matrix* invM)
 }
 
 
-static bool _identify(const Matrix* transform)
+static int _identify(const Matrix* m)
```
hermet, line 74: This does more than identify() check.... bad.

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -871,35 +1056,43 @@ bool rasterClear(SwSurface* surface)
...
-            if (translucent) return _rasterTranslucentImage(surface, image->data, image->w, image->h, opacity, bbox);
-            else return _rasterImage(surface, image->data, image->w, image->h, bbox);
+        if (auto transformType = _identify(transform)) {
+            //Identity or NoTransformation
+            if (transformType == 1) {
+                if (translucent) return _rasterTranslucentImage(surface, image->data, image->w, image->h, opacity, bbox);
+                else return _rasterImage(surface, image->data, image->w, image->h, bbox);
+            }
+            //Translation
+            if (translucent) return _rasterTranslucentImage(surface, image->data, image->w, image->h, opacity, bbox, -transform->e13, -transform->e23);
```
hermet, line 1088: Why don't reuse the existing method? We can just offset the region xy...
_rasterTranslucentImage(SwSurface* surface, uint32_t *img, uint32_t w, uint32_t h, uint32_t opacity, const SwBBox& region);

hermet (changes requested):
too much function introduction... 
For translated image, I think we can reuse the exiting image drawing method...

hermet (changes requested):
Please don't come in the same patch thread unless it has the previous patch dependency...

hermet (changes requested):
Could you please separate PR this?

mgrudzinska:
@hermet please check #274  - I separated one of the commits.

src/lib/sw_engine/tvgSwRaster.cpp
mgrudzinska: @hermet I'm very sorry that it takes so long. I'll push the new patch today/tomorrow.

mgrudzinska: my problem is that I can't see how to use a region because both pictures and shapes/svgs may use the same rastering functions. For ex an svg with clipping and opacity will use _translucentImage, and a translucent picture (pixels) uses it as well.
For a picture I can easily use region to call the same function for the identity transform and the translation, but it's not what I want to do for an svg...

Am I missing something?

hermet:
https://github.com/Samsung/thorvg/pull/1053

Issue #26, opened 2020-09-10
Implement Stencil for Merge Shape with Stroking in Sw Engine


hermet:
These stroking shouldn't be visible in overlapped merging shape regions.

![b7689480-aa3e-11ea-9762-522092100cc1](https://user-images.githubusercontent.com/3711518/92690822-9769a980-f37c-11ea-8b34-acb0b67cf7ea.jpeg)

hermet:
FYI, Here is my brief approach for this:

1. Allocate a 1bite(or 1bit) (per pixel) stencil buffer
2. Fill shape body to stencil buffer first (without outline)
3. Back to the origin drawing routine
4. Check stencil buffer whether outline is inside of shapes when outline drawing turn.
5. decide to draw outline. (if stencil bit is true, skip drawing outline, otherwise drawing)

This routine only works when one shape have several shapes with outlining,

hermet:
we have the merging shapes function in https://github.com/thorvg/thorvg/commit/b26672a1f6c20718d8ba329c849bce465cc59365

PR #240, opened 2021-02-09
Overlapping strokes example

MathewP1:
Example of shapes with overlapping strokes for added stroke stencil functionality.
Issue: #26


hermet (changes requested):
Stencil is inproper name here. People have no idea how this works under the hood.
How about MergeShapes.cpp ?

hermet:
@MathewP1  ping?

PR #241, opened 2021-02-09
sw_engine: stencil for overlapping strokes

MathewP1:
Issue: #26 
Example: #240 

Implementation of stroke stencil. In case of multiple shapes inside one Shape, only the outline stroke is visible.

Before:
before_stencil
After:
after_stencil

### DIFF
--- a/src/examples/Stencil.cpp
+++ b/src/examples/Stencil.cpp
@@ -1,0 +1,153 @@
+#include "Common.h"
+
+/************************************************************************/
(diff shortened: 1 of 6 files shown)


MathewP1:
I've changed it to draft because I found some bugs and performance decrease. Working on both right now.

hermet (changes requested):
You have a separate patch for example here:
https://github.com/Samsung/thorvg/pull/240

Does it necessary to include example again?

PR #245, opened 2021-02-16
sw_engine SwRle: Fix RENDER_POOL_SIZE to 10000 byte

JSUYA:
- Description :
The static analyzer recommends using a stack memory of
single local variable as 10000 bytes.


hermet:
I think we can't change this value, instead need to use dynamic (reusable memory) buffer ...

hermet (changes requested):
I think we can't change this value, It might occur lack of memory to draw large amount of shapes.
Let's leave it unless any critical issue due to this.

PR #246, opened 2021-02-16
common Task: Prevent data race condition

JSUYA:
- Description :
The static analyzer warns of a data race condition
for the 'ready' variable. So use lock to prevent it.


hermet (changes requested):
This is not necessary. this prepare step will be performed on single thread.  (no any chances parallel)

Issue #247, opened 2021-02-17
crash in examples: Opacity, MultiCanvas, Masking, InvMasking, Capi, ClipPath when vector option is set to 'avx' 


mihashco:
```
#0  rasterRGBA32 (dst=dst@entry=0x7fffe9b95810, val=val@entry=0, offset=offset@entry=0, len=356)
    at /usr/lib/gcc/x86_64-linux-gnu/9/include/avxintrin.h:1321
#1  0x00007ffff7fb4026 in rasterClear (surface=surface@entry=0x555555a69f50) at ../src/lib/sw_engine/tvgSwRaster.cpp:828
#2  0x00007ffff7fb4e35 in tvg::SwRenderer::target (this=0x555555b102f0, x=32, y=<optimized out>, w=356, h=339)
    at ../src/lib/sw_engine/tvgSwRenderer.cpp:415
#3  0x00007ffff7fb508a in tvg::SwRenderer::renderShape (this=0x555555b102f0, data=0x555555b944c0)
    at ../src/lib/sw_engine/tvgSwRenderer.cpp:309
#4  0x00007ffff7fad030 in tvg::Paint::Impl::render (this=0x555555b93f40, renderer=...) at ../src/lib/tvgPaint.h:191
#5  0x00007ffff7faf1a6 in tvg::Scene::Impl::render (this=0x555555b10590, renderer=...) at ../src/lib/tvgSceneImpl.h:76
#6  0x00007ffff7fad030 in tvg::Paint::Impl::render (this=0x555555b10560, renderer=...) at ../src/lib/tvgPaint.h:191
#7  0x00007ffff7facc16 in tvg::Canvas::Impl::draw (this=<optimized out>) at ../src/lib/tvgCanvasImpl.h:101
```

PR #248, opened 2021-02-17
sw common: Fixed crash in rasterRGBA32

mihashco:
Description:
Crash was observed in examples when composite object was used.
It was caused because __m256i object was used on non aligned
memory to 32bit. Algorithm in this function was changed to use
unaligned __m256i_u object. Code was also simplified.

- Issue Tickets: #247 

- Tests or Samples: To verify check if crash occurs on examples: ClipPath, Opacity, Capi or Masking.


hermet:
Oh, this avx support is not completed...
Anyway, good to fix it.

hermet (approved):
@mihashco great thanks!

PR #249, opened 2021-02-18
examples: new AnimateMask example and svg images with mask

patrykka:
Added example with animated mask.
    Modified Svg example to display more files.
    Mask and mask-gradient svg images.

- Description :

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :


JSUYA:
sorry, I have updated the build ci script code. because there is a problem in the build ci.
Then there was a conflict in this PR. I added unnecessary commits to this PR as I resolved it.
please make the branch up-to-date, rebase it, and force push it again?

src/examples/AnimateMasking.cpp
```diff
@@ -0,0 +1,191 @@
...
+/* Drawing Commands                                                     */
+/************************************************************************/
+
+tvg::Shape *pMaskShape = nullptr;
+tvg::Shape *pMask = nullptr;
+
+static bool rendered = false;
+static int count = 0;
+static int frame = 0;
+static double t1, t2, t3, t4;
```
JSUYA, line 13: These values are unused.

JSUYA (approved):
Thanks for the nice example.
Overall, I agree with this patch. But please check 1 comment.

And some of added SVG's part are not displayed properly.
These problems are not why this patch shouldn't be merged.
It's just sharing in advance what we need to improve :)

Skeleton's left eye
image

gradient with mask
image

src/examples/AnimateMasking.cpp
patrykka: I removed those unused values, thanks.
Regarding problems in loading SVG, you are right - gradients with masks are not working and the same with loding example like Skeleton and its left eye.

hermet:
@patrykka Could you please share where bone.svg, human.svg image come from? We need to verify any license issues.

patrykka:
@hermet 
Direct links:
https://s3-us-west-2.amazonaws.com/s.cdpn.io/9473/a.svg
https://s3-us-west-2.amazonaws.com/s.cdpn.io/9473/b.svg

it is used in a lot of different public examples, but if you think about license problem I can make x-ray from cartman svg and put something different inside it.

hermet:
@patrykka there is no any license info, in this case we can not implicitly use it. Could you please replace the resource to others?

PR #258, opened 2021-02-25
common: optimize scene composition

hermet:
Scene could avoid composition if its children is only child (non scene)

This patch bring it conditionally check so as to avoid unnecesary expensive job.

@Issues: 254

- Description :

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :


JSUYA (approved):
LGTM

PR #267, opened 2021-02-28
common shape: added api for quadratic Bezier commands

mgrudzinska:
In order to increase the convenience of users the quadraticTo api
has been added, along with the corresponding tvg_shape_quadratic_to capi

- Issue Tickets (if any) : 264


JSUYA:
I don't know that it's good way to add this API.
because user can draw quadraticTo using cubicTo.
and because if we want to support quadraticTo, we need to consider supporting CustomPath(PathCommand::QuadraticTo ?) as well.

mgrudzinska:
the only reason I added this api was because of the user's request. if you think it is unnecessary let's close it

hermet:
@mgrudzinska  it's not quite necessary.. let's see it's really useful or not.

Issue #257, opened 2021-02-24
SvgLoader: svg image is not properly displayed


patrykka:
In Svg example, image:
![Screenshot](https://user-images.githubusercontent.com/5477159/109024175-c28f4d80-76bd-11eb-915f-3d388bc46a51.png)

is not properly displayed, it should be like this: 
![Screenshot2](https://user-images.githubusercontent.com/5477159/109024208-c91dc500-76bd-11eb-885c-2779a96d267e.png)

patrykka:
I tested this problem on different svg viewers and results are different.
To resolve this problem we can turn around one eye and nose in skull path or we can fix loader
@hermet @JSUYA  what do you think ?

patrykka:
one of editors shows it as:
![sc](https://user-images.githubusercontent.com/5477159/109559267-bdc20380-7ada-11eb-81ec-5371a370c510.png)


partial svg code:
``` html
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg version="1.1"
	xmlns="http://www.w3.org/2000/svg"
	xmlns:xlink="http://www.w3.org/1999/xlink"
	 width="400" height="400" viewBox="0 0 400 400" xml:space="preserve">
	<rect x="0" y="0" width="900" height="900" fill="grey"/>
	<g>
		<path fill-rule="evenodd" fill="#FFFFFF" opacity=".8" d="
		M290,260c1-11.75,17.5-2,18-14c0.5-13.75,9-5.5,9-27
		c0.5-54.25-48-58-48-58s-46.5-0.75-49,52c-0.5,19.25,8.5,24,7,34c-0.75,13.75,14.25-2,16,9c-1.25,4.25,1.29,2.02,4.87,2.03
		c1.25-1.91,1.7,0.01,2.63,0.01c1.32,0,2.38-2.04,4.25,0.01c1.79,0,3.37-2.43,5.63-0.01c4.87-2.29,4.4-0.02,6.62-0.03
		c4.12-1.63,4.74-0.04,7-0.08c2.03-0.03,3.5-2.18,5.75-0.11c2.25-1.57,3.46-0.1,4.88-0.17C287.92,257.5,290,260.29,290,260z
[...]

PR #269, opened 2021-03-02
svgloader: fillRule value passed to the shape

mgrudzinska:
The value was loadeed but not set to the shape.

- Issue Tickets (if any) : #257


JSUYA (approved):
LGTM :)

PR #270, opened 2021-03-04
svgloader: fixing SVG image display when viewBox size is not given

mgrudzinska:
When the viewBox is not given its dimensions should be determined
by the height and width parameters

example:
```
<svg height="80" width="80">
<path d="M 10 10 h 60 v 60 h -60 z" fill="none" stroke="#FF0000"stroke-width="6"/>
</svg>
```
current result:
SVG_view_bad

after changes:
SVG_view_ok


JSUYA (approved):
LGTM :)

PR #277, opened 2021-03-07
svg_loader: limiting the ploted area of svg to viewBox

mgrudzinska:
The 'viewBox' element given in the svg file determines
the coordinates of the plotted area.

- Tests or Samples:
<svg viewBox="0 0 200 200" height="200" width="200">
    <rect x="0" y="0" width="150" height="150" fill="#FF0000"/>
    <circle cx="150" cy="150" r="100" fill="#0000FF"/>
</svg>

- Screen Shots:

current:
view_bad

expected:
view_ok

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -363,7 +363,14 @@ unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, float vx, float vy, flo
                     scene->push(_sceneBuildHelper(*child, vx, vy, vw, vh));
                 } else {
                     auto shape = _shapeBuildHelper(*child, vx, vy, vw, vh);
-                    if (shape) scene->push(move(shape));
+                    // clipping the viewBox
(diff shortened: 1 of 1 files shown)


mgrudzinska:
I don't think this solution is optimal because I shouldn't make changes in the  tvgPictureImpl.h file.
I could create the clipPath node manually in the Svg Loader but I need to familiarize myself with it.

In order to include the viewBox element into drawing the svg image I considered also splitting Bezier curves and lines (cut them if they cross the viewBox area)  or constraining the shapes bbox . However, none of these solutions seem better.

Am I missing some obvious solution to this problem? 
I'd really appreciate your advice in this subject.

mgrudzinska:
this solution doesn't work well it there is a mask in the svg

JSUYA:
I made a pr on this issue in a different way. Please review this. https://github.com/Samsung/thorvg/pull/280

mgrudzinska:
I added the clip to all shapes (before it was added to the final scene and it didn;t work properly)

hermet (approved):

PR #278, opened 2021-03-07
svg_loader: ommiting xml entities in the svg file

mgrudzinska:
SVG files created using 'sketch' may contain xml entities,
which are not understandable by the svg loader.
For now only '&quot;' is ommitted.

- Tests or Samples :
<svg viewBox="0 0 500 300" width="500" height="300">
    <defs>
        <rect id="mask_path" x="50" y="100" width="400" height="100" />
    </defs>
    <rect fill="#FF0000" x="0" y="0" width="500" height="300" />
    <mask id="mask" fill="white">
        <use xlink:href="#mask_path" />
    </mask>
    <circle fill="#0000FF" mask="url(&quot;#mask&quot;)" cx="250" cy="150" r="200" />
</svg>

- Screen Shots:
current:
xml_bad

expected:
xml_ok

### DIFF
--- a/src/loaders/svg/tvgXmlParser.cpp
+++ b/src/loaders/svg/tvgXmlParser.cpp
@@ -207,11 +207,10 @@ bool simpleXmlParseAttributes(const char* buf, unsigned bufLength, simpleXMLAttr
     if (!buf || !func) return false;
 
     while (itr < itrEnd) {
-        const char* p = itr;
+        const char* p = _skipWhiteSpacesAndXmlEntities(itr, itrEnd);
         const char *key, *keyEnd, *value, *valueEnd;
         char* tval;
 
-        p = _skipWhiteSpacesAndXmlEntities(p, itrEnd);
         if (p == itrEnd) return true;
 
         key = p;
--- a/src/loaders/svg/tvgXmlParser.cpp
+++ b/src/loaders/svg/tvgXmlParser.cpp
@@ -89,6 +89,69 @@ static const char* _simpleXmlUnskipWhiteSpace(const char* itr, const char* itrSt
 }
 
 
+static const char* _simpleXmlSkipXmlEntities(const char* itr, const char* itrEnd)
+{
+    auto p = itr;
+    while (*itr == '&' && itr < itrEnd) {
+        for (int i = 0; i < NUMBER_OF_XML_ENTITIES; ++i) {
+            if (strncmp(itr, xmlEntity[i], xmlEntityLength[i]) == 0) {
+                itr += xmlEntityLength[i];
+                break;
+            }
+        }
+        if (itr == p) break;
+        p = itr;
+    }
+    return itr;
+}
+
+
+static const char* _simpleXmlUnskipXmlEntities(const char* itr, const char* itrStart)
+{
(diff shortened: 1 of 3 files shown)


JSUYA:
That's a good point. But in this case, only `&quot;` is resolved.

If all XML entities should be treated the same, I think it would be good if these symbols `&nbsp;` (1 space), `&lt;` (<), `&gt;` (>), `&amp;` (&), `&quot;` (“), `&#035;` (#), `&#039;` (‘) are handled together.

Also you could pre-process this code in tvgXmlParser.
Would you consider it to be handled by XmlParser?

hermet:
@mgrudzinska ping?

mgrudzinska:
@hermet  - I'll do this today

mgrudzinska:
@JSUYA could you please take a look? 
I did not finish, because it seems to me that this version is illegible and sensitive to errors. Perhaps I misunderstood what you meant?
I do not know where to expect the XML entities, so I tried to keep the solution as general as possible.
If you confirm that it should look like that I will add all other entities.

src/loaders/svg/tvgXmlParser.cpp
```diff
@@ -176,20 +240,28 @@ bool simpleXmlParseAttributes(const char* buf, unsigned bufLength, simpleXMLAttr
...
+        valueEnd = _unskipWhiteSpacesAndXmlEntities(valueEnd, value);
+
         memcpy(tmpBuf, key, keyEnd - key);
         tmpBuf[keyEnd - key] = '\0';
 
         tval = tmpBuf + (keyEnd - key) + 1;
-        memcpy(tval, value, valueEnd - value);
-        tval[valueEnd - value] = '\0';
+        int i = 0;
+        while (value < valueEnd) {
```
JSUYA, line 253: Why need this logic? 
Entities are already skipped from 245 line.

src/loaders/svg/tvgXmlParser.cpp
```diff
@@ -144,10 +207,11 @@ bool simpleXmlParseAttributes(const char* buf, unsigned bufLength, simpleXMLAttr
     if (!buf || !func) return false;
 
     while (itr < itrEnd) {
-        const char* p = _simpleXmlSkipWhiteSpace(itr, itrEnd);
+        const char* p = itr;
```
JSUYA, line 210: How about changing only function?
p = _skipWhiteSpacesAndXmlEntities(itr, itrEnd);
Are there any cases where this would cause problems?

JSUYA (commented):
Sorry for the late review.

src/loaders/svg/tvgXmlParser.cpp
mgrudzinska: we need this to remove all entities from the middle part of the value, like in the example that we had:
mask="url(&quot;#mask&quot;)"

src/loaders/svg/tvgXmlParser.cpp
mgrudzinska: it's a remnant of an earlier version of the changes :)
thx

JSUYA (approved):
LGTM

PR #280, opened 2021-03-08
common Picture: Calculate to area that to be drawn, based on the ViewBox

JSUYA:
- Description :
If the drawing target(paints) is outside the scope of the viewbox, there is no need to draw.
The <viewbox> provided by SVG has a feature to limit the area to be drawn.
Therefore, use renderer's composite so that the actual area to be drawn is drawn using
vx,vy,vw and vh values of the loader.

- Issue Tickets (if any) :
https://github.com/Samsung/thorvg/pull/277

- Tests or Samples (if any) :
[Test SVG]
```html
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 552 240" width="400" height="200" version="1.1">
  <circle  opacity="0.9" fill="#2c9691" fill-rule="nonzero"  cx="100" cy="100" r="300" />
</svg>
```

- Screen Shots (if any) :
[before]
image

[result]
image

### DIFF
--- a/src/lib/tvgPictureImpl.h
+++ b/src/lib/tvgPictureImpl.h
@@ -39,6 +39,7 @@ struct Picture::Impl
     void *rdata = nullptr;              //engine data
     float w = 0, h = 0;
     bool resizing = false;
+    Matrix transform = { 1, 0, 0 , 0, 1, 0, 0, 0, 1 };
 
     Impl(Picture* p) : picture(p)
     {
@@ -115,6 +116,8 @@ struct Picture::Impl
     {
         auto flag = reload();
 
+        if (transform) this->transform = (Matrix)(transform->m);
+
         if (pixels) rdata = renderer.prepare(*picture, rdata, transform, opacity, clips, static_cast<RenderUpdateFlag>(pFlag | flag));
         else if (paint) {
             if (resizing) resize();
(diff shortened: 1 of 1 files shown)


src/lib/tvgPictureImpl.h
```diff
@@ -123,10 +126,58 @@ struct Picture::Impl
...
         if (pixels) return renderer.renderImage(rdata);
-        else if (paint) return paint->pImpl->render(renderer);
+        else if (paint){
+            Compositor* cmp = nullptr;
+            uint32_t x = 0, y = 0, w = 0, h = 0;
+            calcTargetRegion(&x, &y, &w, &h);
+            if (x != loader->vx || y != loader->vy || w != loader->vw || h != loader->vh){
+                cmp = renderer.target(x, y, w, h);
+            }
+            if (cmp) renderer.beginComposite(cmp, CompositeMethod::None, 255);
```
hermet, line 175: we can find out a better optimal method for the clip out region such as ClipPath.

hermet (changes requested):
we can find out a better optimal method for the clip out region such as ClipPath.

JSUYA:
make a new patch
https://github.com/Samsung/thorvg/pull/292

PR #281, opened 2021-03-10
picture: enabling the use of size() for the duplicate image

mgrudzinska:
size() used the loader parameters and the duplicate image did not contain
this information, which caused a crash. Now the data is stored in the pImpl
of the picture.

- Tests or Samples :
```
        auto picture1 = tvg::Picture::gen();
        picture1->load(EXAMPLE_DIR"/tiger.svg");
        picture1->translate(370, 370);
        picture1->size(300, 300);

        auto picture2 = unique_ptr<tvg::Picture>(static_cast<tvg::Picture*>(picture1->duplicate()));
        picture2->translate(550, 550);
        picture2->size(200, 200);

        canvas->push(move(picture1));
        canvas->push(move(picture2));
```

PR #282, opened 2021-03-10
picture: enabling duplication of the picture with a raw image

mgrudzinska:
duplicate() function worked only for svg images. Now it is possible
to duplicate a Picture with a raw image.

contains changes from #281 

- Tests or Samples : #283

### DIFF
--- a/src/lib/tvgPictureImpl.h
+++ b/src/lib/tvgPictureImpl.h
@@ -39,6 +39,8 @@ struct Picture::Impl
     void *rdata = nullptr;              //engine data
     float w = 0, h = 0;
     bool resizing = false;
+    float vw = 0, vh = 0, vx = 0, vy = 0;  //viewbox parameters
+    bool preserveAspect = true;
 
     Impl(Picture* p) : picture(p)
     {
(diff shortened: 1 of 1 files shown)


src/lib/tvgPictureImpl.h
```diff
@@ -188,18 +199,32 @@ struct Picture::Impl
...
+        dup->w = w;
+        dup->h = h;
+        dup->vw = vw;
+        dup->vh = vh;
+        dup->vx = vx;
+        dup->vy = vy;
+        dup->preserveAspect = preserveAspect;
+
+        if (paint) dup->paint = paint->duplicate();
+        else if (pixels) dup->pixels = pixels;
```
mgrudzinska, line 227: is it ok? 

is it possible, that the main Picture will be deleted before the duplicated one? then I'll copy the content

src/lib/tvgPictureImpl.h
```diff
@@ -39,6 +39,8 @@ struct Picture::Impl
     void *rdata = nullptr;              //engine data
     float w = 0, h = 0;
     bool resizing = false;
+    float vw = 0, vh = 0, vx = 0, vy = 0;  //viewbox parameters
```
hermet, line 42: i think we can avoid this memory allocation.. guess duplicate could share the source's loader like flyweight.

ie) tvgPictureImpl.h

    Paint* duplicate()
    {
        ...
         dup->loader = loader;
        ...
    }

hermet (changes requested):
Please check a comment.

hermet:
please check this:
https://github.com/Samsung/thorvg/pull/301

PR #290, opened 2021-03-15
svg_loader: limiting the ploted area of svg to viewBox

mgrudzinska:
The 'viewBox' element given in the svg file determines
the coordinates of the plotted area. Thi final scene
clipped in the Svg loader.


mgrudzinska:
like #277 but correct. The final scene is clipped instead of shapes (when shapes were transformed so was the clip - that was not intended)

hermet:
@mgrudzinska guess this is conflicted by this? =>  https://github.com/Samsung/thorvg/pull/292

PR #293, opened 2021-03-15
sw_engine rleClipPath: Modify clippath spans creation size.

JSUYA:
- Description :
When updating rle for clipped area, realloc size should be larger than spans size.
so this patch prevents the problem while realloc.

- Issue Tickets (if any) :
https://github.com/Samsung/thorvg/issues/291

- Tests or Samples (if any) :
Test ./Stress.cpp

### DIFF
--- a/src/lib/sw_engine/tvgSwRle.cpp
+++ b/src/lib/sw_engine/tvgSwRle.cpp
@@ -923,8 +923,9 @@ void updateRleSpans(SwRleData *rle, const SwSpan* curSpans, uint32_t size)
     }
 
     if (!rle->spans || !curSpans) return;
+    rle->alloc = size * 2;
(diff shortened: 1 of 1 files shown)


src/lib/sw_engine/tvgSwRle.cpp
```diff
@@ -923,8 +923,9 @@ void updateRleSpans(SwRleData *rle, const SwSpan* curSpans, uint32_t size)
     }
 
     if (!rle->spans || !curSpans) return;
+    rle->alloc = size * 2;
```
hermet, line 926: why 2? it could increase unnecessary memory for shape drawings,  we need to examine the exact reason

hermet (changes requested):

hermet (approved):

PR #297, opened 2021-03-15
examples: revise StrokeGradient

hermet:
renamed GradientStroke & repositioned visuals.

- Description :

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :


src/examples/GradientStroke.cpp
```diff
@@ -26,23 +26,24 @@ void tvgDrawCmds(tvg::Canvas* canvas)
...
-    shape1->lineTo(50, 100);
     shape1->close();
 
     shape1->stroke(0, 255, 0, 255);
     shape1->stroke(20);
     shape1->stroke(tvg::StrokeJoin::Miter);
     shape1->stroke(tvg::StrokeCap::Butt);
+
     auto fillStroke1 = tvg::LinearGradient::gen();
     fillStroke1->linear(50, 50, 200, 200);
```
mgrudzinska, line 48: I'd shift the fill, so all 3 colors are visible, but as you wish
fillStroke1->linear(100, 100, 250, 250);

mgrudzinska: the same in line 53

hermet: Updated the code, do you satisfied with that? :)

mgrudzinska: that is the result that we get now:
ex_bad


The idea was to check whether a stroke and a fill have the same grad:
ex_ok

You changed the fill differently than I wanted, but it's up to you. In any case, the fill and the stroke's fill should be the same. That's why I wrote that the change also applies to line 53.

hermet:
@mgrudzinska  i hope the last change satisfy with you.

mgrudzinska:
more than happy ;)

PR #298, opened 2021-03-15
shape: stroke reset withdraw

mmaciola:
Using rlottie with efl this fixes stroke animation

### DIFF
--- a/src/lib/sw_engine/tvgSwShape.cpp
+++ b/src/lib/sw_engine/tvgSwShape.cpp
@@ -465,7 +465,6 @@ void shapeDelOutline(SwShape* shape, uint32_t tid)
 void shapeReset(SwShape* shape)
 {
     rleReset(shape->rle);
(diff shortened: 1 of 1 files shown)


hermet (changes requested):
I think you missed shapeResetStroke() ??

mmaciola:
I didn't mean to add `shapeResetStroke()` here.
As I know, `rleReset(shape->strokeRle);` was added in `shapeReset` to prevent seg fault when canvas was smaller than used area. Since adding that line some other prevention mechanisms was introduced but its existence is related to problems in efl now.

hermet:
@mmaciola  At least the code shouldn't be there but shapeResetStroke() could be called for resetting the stroke at some point.

As a reviewer, I can't verify proper solution because you didn't introduce any detailed info . -callstack,  reproduce sequence, etc- Thus this kind of patch get hardly reviewed.

See this guide:

PR #300, opened 2021-03-19
initializer: add reference counting for the engines initialization.

hermet:
Introduced the reference counting for the backend engines so that
tvg prevents unpaired engine initialization/termination calls by user mistake.

@Issues: 296

- Description :

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :

### DIFF
--- a/src/lib/tvgInitializer.cpp
+++ b/src/lib/tvgInitializer.cpp
@@ -35,7 +35,7 @@
 /************************************************************************/
 /* Internal Class Implementation                                        */
 /************************************************************************/
-static bool initialized = false;
+
 
 /************************************************************************/
 /* External Class Implementation                                        */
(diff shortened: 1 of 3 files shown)


src/lib/tvgInitializer.cpp
```diff
@@ -67,16 +65,12 @@ Result Initializer::init(CanvasEngine engine, uint32_t threads) noexcept
```
JSUYA, line 65: If user call initializer twice or more.., LoaderMgr::init and TaskScheduler::init are to be called regardless of count.
Is this no problem?
And no matter how many times init is called, the number of threads is fixed to the number initially specified.

JSUYA (commented):
Loa

src/lib/tvgInitializer.cpp
hermet: No problem. There is no correct policy about thread, we strict to keep the thread count at the first call of engine init.

Issue #243, opened 2021-02-15
If shape uses gradient and mask together, mask does not work normally.


JSUYA:
[code]
```cpp
diff --git a/src/examples/Masking.cpp b/src/examples/Masking.cpp
index 8d517ef..d63b1af 100644
--- a/src/examples/Masking.cpp
+++ b/src/examples/Masking.cpp
@@ -14,7 +14,18 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     //Solid Rectangle
     auto shape = tvg::Shape::gen();
     shape->appendRect(0, 0, 400, 400, 0, 0);
-    shape->fill(0, 0, 255, 255);
+    //LinearGradient
+    auto fill = tvg::LinearGradient::gen();
+    fill->linear(0, 0, 400, 400);
+
+    //Gradient Color Stops
+    tvg::Fill::ColorStop colorStops[2];
+    colorStops[0] = {0, 0, 0, 0, 255};
+    colorStops[1] = {1, 255, 255, 255, 255};
+
+    fill->colorStops(colorStops, 2);
+
+    shape->fill(move(fill));
 
     //Mask
     auto mask = tvg::Shape::gen();

```

[screen shot]
![image](https://user-images.githubusercontent.com/7413838/107931405-88d48d80-6fbf-11eb-8a3e-9bf0571c90ba.png)

patrykka:
there is same problem with mask+gradient when defined in svg file.
Visible in Svg example:
![image](https://user-images.githubusercontent.com/5477159/109015622-9cfe4600-76b5-11eb-9da3-3263f8116508.png)

patrykka:
@JSUYA @hermet 
Could you please share your thoughts how gradients + composition should be implemented?
My idea is to add some function like rasterGradientShape inside renderShape, what do you think?

PR #302, opened 2021-03-21
gradients: paint linerar gradients properly with mask composition method

patrykka:
@Issues: 243

- Description :
  After adding gradients to rectangle, mask wasnt visible as it should, described in #243, also some problems with invert masking + gradients was found here is fix also for that.

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :
After fix:

image

### DIFF
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -642,8 +642,32 @@ static bool _rasterLinearGradientRect(SwSurface* surface, const SwBBox& region, 
         }
     //Opaque Gradient
     } else {
-        for (uint32_t y = 0; y < h; ++y) {
-            fillFetchLinear(fill, buffer + y * surface->stride, region.min.y + y, region.min.x, 0, w);
+        if (surface->compositor) {
+            auto method = surface->compositor->method;
+            uint32_t *shape = fill->ctable;
+            auto cbuffer = surface->compositor->image.data + (region.min.y * surface->stride) + region.min.x;
+            auto sbuffer = shape + (region.min.y ) + region.min.x;
+            for (uint32_t y = 0; y < h; ++y) {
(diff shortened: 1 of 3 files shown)


src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -642,8 +642,29 @@ static bool _rasterLinearGradientRect(SwSurface* surface, const SwBBox& region,
...
-        for (uint32_t y = 0; y < h; ++y) {
-            fillFetchLinear(fill, buffer + y * surface->stride, region.min.y + y, region.min.x, 0, w);
+        if (surface->compositor) {
+            uint32_t *shape = fill->ctable;
+            auto cbuffer = surface->compositor->image.data + (region.min.y * surface->stride) + region.min.x;
+            auto sbuffer = shape + (region.min.y ) + region.min.x;
+            for (uint32_t y = 0; y < h; ++y) {
+                auto dst = &buffer[y * surface->stride];
+                auto cmp = &cbuffer[y * surface->stride];
+                auto src = &sbuffer[y];
```
hermet, line 652: I think we can locate the condition out of the for loop. What do you think? 
ex) 
if (AlphaMask)  for (x = 0; x < w; ++x, ++dst, ++cmp) ...
else if (InvAlphaMas)  for (x = 0; x < w; ++x, ++dst, ++cmp) ...

hermet (changes requested):
Could you please prepare any Examples (or append in the existing one) for checking ? Examples can be come with additional patch.

src/lib/sw_engine/tvgSwRaster.cpp
patrykka: Thanks, of course.
Changed

hermet (approved):

PR #308, opened 2021-03-30
capi: changing function name

mgrudzinska:
For uniformity of the functions naming the function name
tvg_shape_stroke_get_gradient changed to tvg_shape_get_stroke_gradient.


mgrudzinska:
Isn't it too late for such a change?
I need to use this api in Evas

mihashco:
For me it is ok. It is not used yet, so additional changes are not required

mihashco (approved):

PR #310, opened 2021-04-01
sw_engine: fixing linear gradient rastering for shapes with composition

mgrudzinska:
Masking and Inverse masking are for now rastered properly only for the linear
gradients without the opacity.

The if statement is pulled outside the for loop in order to improve the performance.

Please be aware that antialiasing still needs to be applied, as well as translucent gradients and radial gradients.

- Screen Shots :
mask
inv

### DIFF
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -616,7 +616,6 @@ static bool _rasterLinearGradientRect(SwSurface* surface, const SwBBox& region, 
 
     //Translucent Gradient
     if (fill->translucent) {
-
         auto tmpBuf = static_cast<uint32_t*>(alloca(surface->w * sizeof(uint32_t)));
         if (!tmpBuf) return false;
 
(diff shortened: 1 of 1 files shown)


src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -631,25 +630,36 @@ static bool _rasterLinearGradientRect(SwSurface* surface, const SwBBox& region,
...
-            uint32_t *shape = fill->ctable;
             auto cbuffer = surface->compositor->image.data + (region.min.y * surface->stride) + region.min.x;
-            auto sbuffer = shape + (region.min.y ) + region.min.x;
-            for (uint32_t y = 0; y < h; ++y) {
-                auto dst = &buffer[y * surface->stride];
-                auto cmp = &cbuffer[y * surface->stride];
-                auto src = &sbuffer[y];
-                if (method == CompositeMethod::AlphaMask) {
-                    for (uint32_t x = 0; x < w; ++x, ++dst, ++cmp) {
+            auto sbuffer = static_cast<uint32_t*>(alloca(w * sizeof(uint32_t)));
```
mgrudzinska, line 634: please double check whether the sbuffer size should be 'w' and not 'surface->w' as it was originally in l.727

hermet (approved):

PR #316, opened 2021-04-02
sw_engine: removed unnecessary function parameter

mgrudzinska:
In the fillFetchLinear function the offset parameter was removed.
The destination address may be shifted directly in the dst parameter,
it doesn't need to be passed separately.


mgrudzinska:
merged with #321

PR #317, opened 2021-04-02
sw_engine: unification of linear and radial gradients rastering functions

mgrudzinska:
The only difference between these functions was the use of the corresponding
fillFetchLinear/Radial function. Since the fillFetch function pointer was introduced,
the splitting was no longer needed. This significantly reduced LOC.

The introduced changes causes the rastering of the radial gradient with masking
to work.

before:
before

after:
after


mgrudzinska:
#316 should be merged first

@hermet Is that a good approach to use a function pointer or it is better to keep radial and linear functions separately? 
Please keep in mind that these rastering functions will be bigger after we develop the code for translucent gradient with masking.

mgrudzinska:
since I was warried about the performance I opened #318

Issue #209, opened 2021-01-11
Masking display incorrectly


JSUYA:
In Masking.cpp example.
When mask object has transparency, the result wrong display.
Alpha is applied to the stroke and fill of the target.

Test code
```cpp
diff --git a/src/examples/Masking.cpp b/src/examples/Masking.cpp
index 8d517ef..fc942d7 100644
--- a/src/examples/Masking.cpp
+++ b/src/examples/Masking.cpp
@@ -52,13 +52,14 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     star->lineTo(426, 161);
     star->lineTo(546, 143);
     star->close();
-    star->stroke(10);
+    star->stroke(30);
     star->stroke(255, 255, 255, 255);
 
     //Mask3
     auto mask3 = tvg::Shape::gen();
     mask3->appendCircle(600, 200, 125, 125);
-    mask3->fill(255, 255, 255, 255);
+    mask3->fill(255, 255, 255, 125);
+    //or mask3->opacity(125);
     star->composite(move(mask3), tvg::CompositeMethod::AlphaMask);
     if (canvas->push(move(star)) != tvg::Result::Success) return;

```
![image](https://user-images.githubusercontent.com/7413838/104146224-573f3400-540d-11eb-80b8-dea02830b8fd.png)

hermet:
Looks this is correct behavior as masking just doing as it's role.  masking should not change its source's behavior.

hermet:
Sorry, I misunderstood.

mgrudzinska:
TODO:
- Translucent linear gradient
- Opaque radial gradient
- Translucent radial gradient
- antialiasing in opaque linear gradient

mgrudzinska:
#326 should close the subject of the gradient shapes with masks

mgrudzinska:
@hermet @JSUYA @patrykka 

I have a question - how should the expected result of the following code look like?

```
        //-------------------------------------------
        auto shape1 = tvg::Shape::gen();
        shape1->appendRect(400, 0, 400, 400, 0, 0);
        shape1->fill(255, 0, 0, 255);

        //Alpha Mask
        auto mask1 = tvg::Shape::gen();
        mask1->appendCircle(600, 200, 180, 180);
        mask1->fill(0, 0, 0, 127);

        shape1->composite(move(mask1), tvg::CompositeMethod::AlphaMask);
        canvas->push(move(shape1));

        //-------------------------------------------
[...]

PR #318, opened 2021-04-04
sw_engine: rastering the opaque radial gradient with masking

mgrudzinska:
To be merged after #316 

In the radial gradient rastering functions, the part supporting
the (inverse) masking was added.

issue: #209 

before:
before

after:
after


mgrudzinska:
Please  merge after #316 

@hermet I closed #317 because I didn't fill comfortable with a function pointer used in the for loop of the rastering functions, since I was warried about the performance. I'd really appreciate your advice in this subject. Are my doubts unjustified?

mgrudzinska:
merged with #321

PR #319, opened 2021-04-05
sw_engine raster: ++ precise alpha multiplying

hermet:
previously alpha multiplying operation doesn't have perfect precision,
could loss 1 pixel since it divides 255 values by 256.

This improved operation comply with both precision & performance.

- Description :

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :

### DIFF
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -274,7 +274,7 @@ static inline uint32_t COLOR_INTERPOLATE(uint32_t c1, uint32_t a1, uint32_t c2, 
 
 static inline uint8_t ALPHA_MULTIPLY(uint32_t c, uint32_t a)
 {
-    return (c * a) >> 8;
+    return ((c * a + 0xff) >> 8);
 }
 
 static inline SwCoord HALF_STROKE(float width)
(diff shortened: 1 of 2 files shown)


src/lib/sw_engine/tvgSwCommon.h
```diff
@@ -274,7 +274,7 @@ static inline uint32_t COLOR_INTERPOLATE(uint32_t c1, uint32_t a1, uint32_t c2,
 
 static inline uint8_t ALPHA_MULTIPLY(uint32_t c, uint32_t a)
 {
-    return (c * a) >> 8;
+    return ((c * a + 0xff) >> 8);
```
mgrudzinska, line 277: we had this discussion already in #230 
result from this formula is 1 higher the the exact result in about 30% of the cases (out of all over 65k)
so it's not really the 'precise alpha multiplying', but the solution that I wanted to apply, has one additional multiplication. 

the question is how precise we want to be and what's the price that we are ready to pay.

hermet: @mgrudzinska i know. i decided this since it's better precise than current version.
Most cases it results in correct result,  only when values c= 0 a=0  results in 1.
but we don't allow alpha == 0 case in the rasterization.

mgrudzinska: not only for c=a=0, for example c=a=1 in this formula gives also 1 instead of 0, but I understand your point. you are right that it's better than the current version

mgrudzinska (approved):

src/lib/sw_engine/tvgSwCommon.h
hermet: c=a=1
1 * 1 + 255 / 256 = 1  (correct result.)

hermet: @mgrudzinska yeah see if it's actual issued later.

PR #320, opened 2021-04-05
sw_engine raster: refactoring the gradient rastering functions

mgrudzinska:
The gradient rastering functions have been splitted into translucent
and opaque.

The functions have been splitted because the translucent part still requires adding part regarding the composition. Keeping the opaque and translucent parts in one function makes it very long and illegible.

contains #316 and #318.


mgrudzinska:
merged with #321

PR #326, opened 2021-04-08
sw_engine: refactoring of the gradient rastering functions

mgrudzinska:
The gradient 'translucent' functions support cases with composition
as well as with a < 255.

before:
maskowanie_transl_gefore

after:
maskowanie_transl_after

### DIFF
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -596,7 +596,7 @@ static bool _rasterImage(SwSurface* surface, const uint32_t *img, uint32_t w, ui
 /* Gradient                                                             */
 /************************************************************************/
 
-static bool _rasterTranslucentLinearGradientRect(SwSurface* surface, const SwBBox& region, const SwFill* fill)
+static bool _translucentLinearGradientRect(SwSurface* surface, const SwBBox& region, const SwFill* fill)
 {
     if (fill->linear.len < FLT_EPSILON) return false;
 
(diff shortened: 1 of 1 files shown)


mgrudzinska:
I refactored the code to make it more like the rest of the code

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -596,348 +596,481 @@ static bool _rasterImage(SwSurface* surface, const uint32_t *img, uint32_t w, ui
 /* Gradient                                                             */
 /************************************************************************/
 
-static bool _rasterTranslucentLinearGradientRect(SwSurface* surface, const SwBBox& region, const SwFill* fill)
```
hermet, line 599: i think you don't need to change the function name.  please keep it convention.

hermet (changes requested):
If you need a review, please make it sure refactoring / enhancement / bugfix 
Please do not composite review things.

hermet:
@mgrudzinska atomic patches are welcomed.

mgrudzinska:
I meant that the refactoring from this pr leads to the enhancement :) 
I removed one label

src/lib/sw_engine/tvgSwRaster.cpp
mgrudzinska: I changed the name, because I also changed what this func does. I used the same convention as we are using right now. It looks horrible in github, so I'll summarize:

For now we have the external class implementation functions called by the rendering functions. 
Their names are like ```rasterSolidShape``` or ```rasterGradientShape```.

These functions are using the internal class impl functions, like:
- ```rasterSolidShape``` can call (fast track): ```_rasterTranslucentRect``` or ```_rasterSolidRect```
- ```rasterGradientShape``` can call (fast track): ```_rasterTranslucentLinearGradientRect``` and ```_rasterOpaqueLinearGradientRect```

Some of these functions, however, either do the rasterization or choose which function to call - this happens in case of translucent functions. So e.g.:
- ```_rasterTranslucentRect``` chooses between ```_translucentRectAlphaMask```, ```_translucentRectInvAlphaMask``` and ```translucentRect```
- ```_rasterTranslucentLinearGradientRect``` for now handles the cases with the opacity, but the cases with a composition are handled in the ```_rasterOpaqueLinearGradientRect```

This is not as it suppose to be. So after my changes, ```_rasterOpaqueLinearGradientRect``` handles only the cases without the opacity (I mean with op = 255) and without any composition, while ```_rasterTranslucentLinearGradientRect``` chooses between ```_translucentLinearGradientRectAlphaMask```, ```_translucentLinearGradientRectInvAlphaMask``` and ```_translucentLinearGradientRect```. So it's in accordance with the used naming convention

mgrudzinska:
#384 and #385 fix the some problems as this pr.
the only part left is the refactorization - will be done in this pr soon

hermet:
@mgrudzinska does still valid patch? what about this => https://github.com/Samsung/thorvg/pull/385  ?

mgrudzinska:
@hermet I still want to do a refactorization, so that the grad functions are written in the same convention as all  other. 
I'll do this today

mgrudzinska:
refactoring is done in #391 and #395.

PR #335, opened 2021-04-12
sw_engine: fixing overlapping masks

mgrudzinska:
The buffer to which the mask is rastered is only partially cleared.
If the object to which the mask is applied overlaps an area where
another mask was used, an erroneous image is generated. The buffer
clearing area has been increased to the size of the object to which
the mask is applied.

before:
mask_overlap

after:
mask_overlap_ok

### DIFF
--- a/src/lib/tvgPaint.cpp
+++ b/src/lib/tvgPaint.cpp
@@ -167,7 +167,7 @@ bool Paint::Impl::render(RenderMethod& renderer)
     /* Note: only ClipPath is processed in update() step.
         Create a composition image. */
     if (cmpTarget && cmpMethod != CompositeMethod::ClipPath) {
-        auto region = cmpTarget->pImpl->bounds(renderer);
+        auto region = smethod->bounds(renderer);
         if (region.w == 0 || region.h == 0) return false;
         cmp = renderer.target(region);
         renderer.beginComposite(cmp, CompositeMethod::None, 255);


mgrudzinska:
please be aware, that I didn't make any changes in the corresponding gl_engine functions

src/lib/sw_engine/tvgSwRenderer.cpp
```diff
@@ -395,7 +395,7 @@ bool SwRenderer::beginComposite(Compositor* cmp, CompositeMethod method, uint32_
 }
 
 
-Compositor* SwRenderer::target(const RenderRegion& region)
+Compositor* SwRenderer::target(const RenderRegion& region, const RenderRegion* clearRegion)
```
mgrudzinska, line 398: @hermet could you please advise a better way of doing this? 
I don't want to change the size of the cmp, since there is no point of rastering anything more than its bbox.
But I need to know at this point also the size of the whole object, that will be masked,  to clear the buffer correspondingly

src/lib/tvgPaint.cpp
```diff
@@ -169,7 +169,8 @@ bool Paint::Impl::render(RenderMethod& renderer)
     if (cmpTarget && cmpMethod != CompositeMethod::ClipPath) {
         auto region = cmpTarget->pImpl->bounds(renderer);
         if (region.w == 0 || region.h == 0) return false;
-        cmp = renderer.target(region);
+        auto clearRegion = smethod->bounds(renderer);
+        cmp = renderer.target(region, &clearRegion);
```
hermet, line 173: good catch, but you can easily fix this issue.
here region argument is designed for optimal rendering area.
So you can unifiy the region with the current shape's boundary rather than separate arguments.
ie)
        auto region = cmpTarget->pImpl->bounds(renderer);
        region.union(smethod->bounds(renderer));
        cmp = renderer.target(region);


    void union(const RenderRegion& rhs)
    {
        auto x1 = x + w;
        auto y1 = y + h;
        auto x2 = rhs.x + rhs.w;
        auto y2 = rhs.y + rhs.h;

        x = (x < rhs.x) ? x : rhs.x;
        y = (y < rhs.y) ? y : rhs.y;
        w = ((x1 < x2) ? x1 : x2) - x;
        h = ((y1 < y2) ? y1 : y2) - y;
    }

hermet (changes requested):
Please check a comment.

src/lib/tvgPaint.cpp
mgrudzinska: I do not see how it would work. Union returns an area that is bigger than the region, but smaller than the clearRegion.

I wanted to pass both these areas because when the mask is rastered, there is no point in considering a larger area than the region. But when the shape is rastered, the entire clearRegion will be used, so it has to be cleared.

hermet: This my typo. 
   w = ((x1 < x2) ? x1 : x2) - x;
    h = ((y1 < y2) ? y1 : y2) - y;
My intention of union is "region + smethod->bounds(renderer)"
   w = ((x1 > x2) ? x1 : x2) - x;
    h = ((y1 > y2) ? y1 : y2) - y;

mgrudzinska: than I don't think that we need to do this - method->bounds(renderer) is enough
we don't need to raster the mask, which doesn't cover with the shape, so the shape's region is enough

hermet: @mgrudzinska no, not enough, buffer should be cleared for mask + target region since the buffer can be reusable.

mgrudzinska: @hermet probably I was unclear.
I think that we can solve the problem using ```region = smethod->bounds(renderer)``` for masks and inv masks, but than we don't need #349. Below I'm trying to explain.

Lets consider the situation as below:
a1

1) alpha mask rect
If I use the union funct, as you proposed, than #349 doesn't make any sense any more. 
First the mask is rastered - the loop covers only the red region - it's ok.
Than the shape is rastered - and that's the problem. Now the intersect func from #349 intersects the shape region (region - blue) with the result of the union func (```surface->compositor->bbox``` - the black border). The later can be even bigger than the shape, but never smaller. So finally, during the shape rasterization, the loop covers the whole shapes area (blue), as it is now (before #349).
Additionally, in the target func from theSwRenderer.cpp we clear the whole area marked with a black border in the pict. We need to clear the region equal to the shape region. So that is what I meant by saying that I don't need the union func and the shapes bbox is enough - I can simply use ```region = smethod->bounds(renderer)```. For rect, it's enough to clear the mask bbox, but I don't now how to distinguish at this stage between the rect masks and the rle masks. That's why I introduced the second argument to the target func - the clearRagion. 
2) alpha mask rle and inv mask
I think that in these cases the only difference between the usage of the union funct vs ```region = smethod->bounds(renderer)```, is the area that is cleared in target funct. But as before - we need to clear only the area that corresponds to the shapes bbox.

hermet: @mgrudzinska 
1. let's forget #349 here since it has a different purpose.
2. Doesn't this way work?
    auto region = smethod->bounds(renderer);
    cmp = renderer.target(region);
    I think here region of target must be clearRegion == region

hermet (commented):
@mgrudzinska added comment.

src/lib/tvgPaint.cpp
hermet: @mgrudzinska too old history i wrote... confused.

hermet: @mgrudzinska 
1. let's forget #349  here. it has a different topic issue. (we can invalid it after this)
2. you're correct but your version makes me confuse since target() doesn't need to have several region info. I think correct approach is unique region info that is reffered for clearing buffer & drawing partial mask region.

mgrudzinska: yes, so this is exactly what I suggested "I think that we can solve the problem using region = smethod->bounds(renderer)". My only concern was about the perf, but it won't be worse than it's now, so I understand that you accept that.
I'll push a changes in a few mins

hermet (approved):
Great, Thanks!

PR #336, opened 2021-04-12
capi: svgPath parse and append to shape paint

mmaciola:
This is a binding for parsing and appending svgPath into paint.

I am open if you decide to move the function from tvgCapi.cpp into general cpp and create just short binding in `bindings/capi` or create separate binding for `svgPathToTvgPath`.

### DIFF
--- a/src/bindings/capi/tvgCapi.cpp
+++ b/src/bindings/capi/tvgCapi.cpp
@@ -24,6 +24,9 @@
 #include <thorvg.h>
 #include "thorvg_capi.h"
 
+#include "tvgSvgLoaderCommon.h"
+#include "tvgSvgPath.h"
+
 using namespace std;
 using namespace tvg;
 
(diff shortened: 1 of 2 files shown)


src/bindings/capi/tvgCapi.cpp
```diff
@@ -555,6 +558,23 @@ TVG_EXPORT Tvg_Result tvg_scene_clear(Tvg_Paint* scene)
     return (Tvg_Result) reinterpret_cast<Scene*>(scene)->clear();
 }
 
+/************************************************************************/
+/* SVG Loader                                                           */
+/************************************************************************/
+TVG_EXPORT Tvg_Result tvg_svg_path_append_to_shape(Tvg_Paint* paint, const char* svgPath)
```
mgrudzinska, line 564: could you please elaborate why can't we use the evas_vg__shape_append_svg_path from evas? 
I'm asking, since we had this discussion when we started the integration process. The conclusion was, that we will make the conversion on the evas site, and so we did.

hermet (changes requested):
this specific svg feature is not allowed in tvg.

mmaciola:
My misapprehension. If it was discussed earlier with the conclusion that it should not be implemented that way, I am okay. Closing

PR #337, opened 2021-04-13
sw_engine: fixing masking behavior

mgrudzinska:
Masking with the opacity shouldn't change the target itself.
The mask's opacity < 255 causes the masked part to appear.

before (left - Mask, right - invMask):
mask_notok

after (left - Mask, right - invMask):
mask_ok


mgrudzinska:
It works, but I'd like to think about the variables/functions names.
Is this way of solving this problem is the best one?

hermet:
@mgrudzinska  i will review this soon.

hermet:
@mgrudzinska I'm not sure what are you trying to fix. "before version" looks correct to me. what brought you "Masking with the opacity shouldn't change the target itself." ?

mgrudzinska:
@hermet, let me think about that.
Could you please tell me, how do you know how the inverse masking with opacity should work?

mgrudzinska:
@hermet I think that you are right and masking works as it should, but I'm not sure about inverse mask. 
Please check #209 - there are 3 red plots. I thought that the correct result is the 3 one. now I think that it should be the second one, but I can't find any precise info to verify this.

hermet:
just checked, it's still valid issue.

PR #342, opened 2021-04-13
loader svg: displaying SVG images without the specified viewBox attri…

mgrudzinska:
…bute

If no viewBox attribute is given, the height and width attributes are used
to establish the viewBox size. However, this information was not available
inside the scene builder, even though it was needed.

before:
v2_bad

after:
v2_ok

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2574,7 +2574,7 @@ void SvgLoader::run(unsigned tid)
         _updateComposite(loaderData.doc, loaderData.doc);
         if (defs) _updateComposite(loaderData.doc, defs);
     }
-    root = svgSceneBuild(loaderData.doc);
+    root = svgSceneBuild(loaderData.doc, vx, vy, vw, vh);
 };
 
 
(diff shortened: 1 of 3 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2598,11 +2598,17 @@ bool SvgLoader::header()
         //Override size
         if (loaderData.doc->node.doc.w > 0) {
             w = loaderData.doc->node.doc.w;
-            if (vw < FLT_EPSILON) vw = w;
+            if (vw < FLT_EPSILON) {
+                vw = w;
+                loaderData.doc->node.doc.vw = w;
+            }
```
JSUYA, line 2604: I recommend passing the viewBox(vx, vy, vw, vh) information as an argument of svgSceneBuild(node, vx, vy, vw, vh) when calling svgSceneBuild in SvgLoader::run() (2577 line).
The data in loaderData.doc->node.doc should remain the same as the parsed data.

JSUYA (changes requested):
Please check my comment

JSUYA (approved):
LGTM

PR #343, opened 2021-04-14
common_shape: Fixed possible memory leak.

mihashco:
Description:
If rendering engine integrated with ThorVG use tvg::canvas::clear with
free == false, then render data is not disposed. It can cause memory
leaks. In this commit check was added to verify if render data is still
available during shape destruction. In that case dispose of render data
is called.

Notes: 
Stability was checked on existing examples. No crash was observed.

### DIFF
--- a/src/lib/tvgShapeImpl.h
+++ b/src/lib/tvgShapeImpl.h
@@ -206,6 +206,7 @@ struct Shape::Impl
     RenderData rdata = nullptr;         //engine data
     Shape *shape = nullptr;
     uint32_t flag = RenderUpdateFlag::None;
+    RenderMethod *renderer = nullptr;
 
     Impl(Shape* s) : shape(s)
     {
@@ -213,8 +214,10 @@ struct Shape::Impl
 
     ~Impl()
     {
+        if (renderer && rdata) renderer->dispose(rdata);
         if (fill) delete(fill);
         if (stroke) delete(stroke);
+        rdata = nullptr;
     }
 
     bool dispose(RenderMethod& renderer)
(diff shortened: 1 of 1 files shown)


src/lib/tvgShapeImpl.h
```diff
@@ -206,15 +206,21 @@ struct Shape::Impl
...
     {
     }
 
     ~Impl()
     {
+        if (renderer && rdata) renderer->dispose(rdata);
         if (fill) delete(fill);
         if (stroke) delete(stroke);
+
+        renderer = nullptr;
```
hermet, line 221: these looks unnecessary.

src/lib/tvgShapeImpl.h
```diff
@@ -231,6 +237,7 @@ struct Shape::Impl
 
     void* update(RenderMethod& renderer, const RenderTransform* transform, uint32_t opacity, Array<RenderData>& clips, RenderUpdateFlag pFlag)
     {
+        this->renderer = &renderer;
```
hermet, line 240: i think we need to call dispose() immediately when paints are popped from canvas.

hermet (changes requested):

src/lib/tvgShapeImpl.h
mihashco: agree. I found simpler solution, please check it

mihashco:
Found crash with new solution

src/lib/tvgShapeImpl.h
mihashco: I left rdata = nullptr here to avoid crash

hermet: @mihashco i undestood, need to figure out a better apporoach. let me think.

hermet:
@mihashco https://github.com/Samsung/thorvg/pull/346/files Could you please have a look this?

PR #346, opened 2021-04-15
common: fix a potential memory leak after clear()

hermet:
Paints must clear canvas engine data if they were dismissed from the canvas,

Canvas::clear(free = false) must retain all the paints through the paints hierarchy
so that it keeps the consistency of all dangled paints lifecycle.

Previously, it just keeps the immediate paints lives of canvas, but not them of scene nor picture.

This patch changes a policy which was not considered seriously,

Now it keeps the all paints lives through the tree-hieararchy.

Instead, Scene will add clear() to manange its children lifecycle.

- Description :

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :

### DIFF
--- a/src/lib/tvgCanvasImpl.h
+++ b/src/lib/tvgCanvasImpl.h
@@ -59,11 +59,9 @@ struct Canvas::Impl
         if (!renderer || !renderer->clear()) return Result::InsufficientCondition;
 
         //free paints
-        if (free) {
-            for (auto paint = paints.data; paint < (paints.data + paints.count); ++paint) {
-                (*paint)->pImpl->dispose(*renderer);
-                delete(*paint);
-            }
+        for (auto paint = paints.data; paint < (paints.data + paints.count); ++paint) {
+            (*paint)->pImpl->dispose(*renderer);
+            if (free) delete(*paint);
         }
 
         paints.clear();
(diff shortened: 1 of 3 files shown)


src/lib/tvgCanvasImpl.h
```diff
@@ -59,11 +59,9 @@ struct Canvas::Impl
         if (!renderer || !renderer->clear()) return Result::InsufficientCondition;
 
         //free paints
-        if (free) {
-            for (auto paint = paints.data; paint < (paints.data + paints.count); ++paint) {
-                (*paint)->pImpl->dispose(*renderer);
-                delete(*paint);
-            }
+        for (auto paint = paints.data; paint < (paints.data + paints.count); ++paint) {
+            (*paint)->pImpl->dispose(*renderer);
```
mihashco, line 63: I think this part could cause crash in sequence: tvg_canvas_clear(canvas, false); tvg_paint_del(paint); We are not able to check easily if canvas was cleared and then ignore tvg_paint_del. Also we are not able to ignore tvg_paint_del() completely, because paint duplication cause memory leaks.

hermet: Can you explain what do you prefer the usage here? 
do you want this?
tvg_canvas_clear(canvas, false);
tvg_paint_del(paint);

hermet: see https://github.com/Samsung/thorvg/pull/382

PR #349, opened 2021-04-15
sw_engine raster: optimize alpha masking

hermet:
in case of alpha masking, out of mask region is invalid,
we can skip the masking composition step for the out of masking boundary
by intersecting the target/masking region before the rasterazation.

- Description :

- Issue Tickets (if any) :

- Tests or Samples (if any) :

- Screen Shots (if any) :

### DIFF
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -135,6 +135,14 @@ struct SwBBox
     {
         min.x = min.y = max.x = max.y = 0;
     }
+
+    void intersect(const SwBBox& rhs)
+    {
+        min.x = (min.x > rhs.min.x) ? min.x : rhs.min.x;
+        min.y = (min.y > rhs.min.y) ? min.y : rhs.min.y;
+        max.x = (max.x < rhs.max.x) ? max.x : rhs.max.x;
+        max.y = (max.y < rhs.max.y) ? max.y : rhs.max.y;
+    }
 };
 
 struct SwFill
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -167,7 +167,9 @@ static bool _rasterTranslucentRect(SwSurface* surface, const SwBBox& region, uin
 {
     if (surface->compositor) {
         if (surface->compositor->method == CompositeMethod::AlphaMask) {
-            return _translucentRectAlphaMask(surface, region, color);
+            auto intersected = region;
+            intersected.intersect(surface->compositor->bbox);
+            return _translucentRectAlphaMask(surface, intersected, color);
         }
         if (surface->compositor->method == CompositeMethod::InvAlphaMask) {
             return _translucentRectInvAlphaMask(surface, region, color);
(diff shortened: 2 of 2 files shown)


hermet:
@mgrudzinska this patch might fix https://github.com/Samsung/thorvg/pull/335 alpha masking issues but still that patch is necessary for inverse alpha masking (with my correction in feedback).

mgrudzinska:
@hermet, you are right, it works for rect with alpha masking. and what about rle?

For rle and for the invMasking we could use the union function that you proposed, but I don't understand what this function should do.  If I code it as you proposed it does not work properly - please check:
```
    //Solid Rectangle
    shape->appendRect(0, 0, 600, 600, 0, 0);
    shape->fill(0, 0, 255, 255);

    //Mask
    auto mask = tvg::Shape::gen();
    mask->appendCircle(400, 200, 125, 125);
    mask->fill(255, 0, 0, 155);
    shape->composite(move(mask), tvg::CompositeMethod::AlphaMask);
    canvas->push(move(shape));


    //Star
    auto star = tvg::Shape::gen();
    star->moveTo(199, 34);
    star->lineTo(253, 143);
    star->lineTo(374, 160);
    star->lineTo(287, 244);
    star->lineTo(307, 365);
    star->lineTo(199, 309);
    star->lineTo(97, 365);
    star->lineTo(112, 245);
    star->lineTo(26, 161);
    star->lineTo(146, 143);
    star->close();
    star->translate(300,0);
    star->fill(80, 80, 80, 255);
    
    //Mask3
    auto mask2 = tvg::Shape::gen();
    mask2->appendCircle(500, 200, 25, 25);
    mask2->fill(255, 255, 255, 255);
    star->composite(move(mask2), tvg::CompositeMethod::AlphaMask);
    canvas->push(move(star));
```
the result:
union

if I change it to return sum of these two regions, than I don't need it and I can simply write in tvgPaint.cpp:
auto region = smethod->bounds(renderer);
but than the whole region will be rastered and it is not optimal at all. That's why I wanted to have both regions in the target() fun.

hermet:
@mgrudzinska this patch doesn't fix any bugs, but optimize the alpha masking for rect. RLE is out of scope.

mgrudzinska:
@hermet I misunderstood your comment. I thought that this patch should be used instead of #335.
I refactorized the #326 so that exactly the same logic can be applied to _rasterTranslucentRadialGradientRect and _rasterTranslucentLinearGradientRect.

mgrudzinska (approved):

src/lib/sw_engine/tvgSwCommon.h
```diff
@@ -135,6 +135,14 @@ struct SwBBox
...
         min.x = min.y = max.x = max.y = 0;
     }
+
+    void intersect(const SwBBox& rhs)
+    {
+        min.x = (min.x > rhs.min.x) ? min.x : rhs.min.x;
+        min.y = (min.y > rhs.min.y) ? min.y : rhs.min.y;
+        max.x = (max.x < rhs.max.x) ? max.x : rhs.max.x;
+        max.y = (max.y < rhs.max.y) ? max.y : rhs.max.y;
+    }
```
mgrudzinska, line 145: in case of non overlapping boxes:

```if (min.x > max.x || min.y > max.y) min.x = min.y = max.x = max.y = 0;```

for now we have segf.

mgrudzinska (changes requested):

PR #350, opened 2021-04-15
loaders png: apply task for async loading.

hermet:
We assume async loading is better than synchronous.

Strangely, this doesn't prove any performance improvement...

### DIFF
--- a/src/loaders/png/tvgPngLoader.cpp
+++ b/src/loaders/png/tvgPngLoader.cpp
@@ -27,8 +27,8 @@ PngLoader::PngLoader()
 {
     image = static_cast<png_imagep>(calloc(1, sizeof(png_image)));
     image->version = PNG_IMAGE_VERSION;
-    image->opaque = NULL;
 }
+
 
 PngLoader::~PngLoader()
 {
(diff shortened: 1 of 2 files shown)


mgrudzinska:
@hermet Can I ask how did you check the perf? To be precise - how many png files did you load? 
I'd expect that in case of one file, there will be no advantage on using tasks, but when one tries to load many of files, then they can be loaded at the same time.

so my concerns are about the case with the svg loader. is that not exactly the same case? can you observe the improvement of the perf when you do the exact same test as for the png loader?

hermet:
> @hermet Can I ask how did you check the perf? To be precise - how many png files did you load?
> I'd expect that in case of one file, there will be no advantage on using tasks, but when one tries to load many of files, then they can be loaded at the same time.
> 
> so my concerns are about the case with the svg loader. is that not exactly the same case? can you observe the improvement of the perf when you do the exact same test as for the png loader?

@mgrudzinska of course, i don't use one image loading. i used 100 ~ 500 pngs loading at the same time for the performance test. :)

hermet:
@there is no any benefits here.

PR #351, opened 2021-04-16
capi: Added Inverted Alpha Mask composition enumeration.

mihashco:
Added missing composite method. Backend supports this compostion type.


inc/thorvg_capi.h
```diff
@@ -108,7 +108,8 @@ typedef enum {
 typedef enum {
     TVG_COMPOSITE_METHOD_NONE = 0,
     TVG_COMPOSITE_METHOD_CLIP_PATH,
-    TVG_COMPOSITE_METHOD_ALPHA_MASK
+    TVG_COMPOSITE_METHOD_ALPHA_MASK,
+    TVG_COMPOSITE_METHOD_INV_ALPHA_MASK,
```
hermet, line 112: What do you think? If we use full addressed naming? (not like c++ version)
TVG_COMPOSITE_METHOD_INVERSE_ALPHA_MASK

hermet (changes requested):
Please check a comment.

inc/thorvg_capi.h
mihashco: Ok, Updated

hermet (approved):

PR #357, opened 2021-04-25
common paint: changed result in composite API

mgrudzinska:
A possible error is due to erroneous arguments, so
InvalidArgument is return instead of InsufficientCondition

### DIFF
--- a/src/lib/tvgPaint.cpp
+++ b/src/lib/tvgPaint.cpp
@@ -298,7 +298,7 @@ Paint* Paint::duplicate() const noexcept
 Result Paint::composite(std::unique_ptr<Paint> target, CompositeMethod method) const noexcept
 {
     if (pImpl->composite(target.release(), method)) return Result::Success;
-    return Result::InsufficientCondition;
+    return Result::InvalidArguments;
 }
 
 


src/lib/tvgPaint.cpp
```diff
@@ -298,7 +298,7 @@ Paint* Paint::duplicate() const noexcept
 Result Paint::composite(std::unique_ptr<Paint> target, CompositeMethod method) const noexcept
 {
     if (pImpl->composite(target.release(), method)) return Result::Success;
-    return Result::InsufficientCondition;
```
hermet, line 301: which scenario would occur this ? any case?

"A possible error is due to erroneous arguments"

hermet (changes requested):
please check a comment.

src/lib/tvgPaint.cpp
mgrudzinska: e.g. ```composite(nullptr, CompositeMethod::AlphaMask)``` or ```composite(target, CompositeMethod::None)```
so in cases, when the args makes no sense.
in your opinion it should be InsufficientCondition instead InvalidArguments?

hermet: @mgrudzinska ok, you're correct.

hermet (approved):

PR #359, opened 2021-04-26
infra Actions: If no coding style error, no comment

JSUYA:
- Description :
and change action name

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1,4 +1,4 @@
-name: Continuous-integration
+name: Build Test
 
 on:
   pull_request:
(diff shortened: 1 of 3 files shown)


hermet:
@JSUYA  if there is no comment for verification, how user know the action is finished?

JSUYA:
Click `Show all checks` of `All check have Passed` at the bottom of the comment timeline and 
you can see  "Check Commit / coding-style (pull_request) Successful in 24s".
This job will almost always be successful unless the build fails. it also means that the checker worked fine.
I think `perfect` comment  is unnecessary.
Please tell me if you want to keep 'perfect' comment. i will modify the patch.

hermet:
@JSUYA make sense.

hermet (approved):

PR #360, opened 2021-04-26
infra Travis: Remove travis-ci build test

JSUYA:
- Description :
We don't need to build test with travis-ci.
The build test is already being worked in the github-action.
In addition, travis-ci causes unnecessary failures because the recent request count is not properly counted.


hermet:
we always appreciated unused stuff.

PR #361, opened 2021-04-26
doc common: adding the documentation for the C++ APIs

mgrudzinska:
Added the missing descriptions for Sw/Gl/Canvas and Initializer.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1,5 +1,5 @@
 /*!
-* \file thorvg.h
+* @file thorvg.h
 */
 
 
@@ -67,7 +67,8 @@ enum class TVG_EXPORT Result
 /**
  * @brief Enumeration specifying the values of the path commands accepted by ThorVG.
  *
- * Not to be confused with the path commands from the svg path element.
+ * Not to be confused with the path commands from the svg path element (like M, L, Q, H and many others).
+ * ThorVG interprets all of them and translates to the ones from the PathCommand values.
  */
 enum class TVG_EXPORT PathCommand
 {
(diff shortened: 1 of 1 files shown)


inc/thorvg.h
```diff
@@ -1062,29 +1097,32 @@ class TVG_EXPORT GlCanvas final : public Canvas
...
-     *
-     * @see ...
+     * @brief Initializes the ThorVG engine.
+     * @param[in] engine The engine type - the same as specified in the 'meson_options.txt' file.
+     * @param[in] threads The number of threads - zero is treated as one.
+     * @return Result::Success when succeeded, otherwise the returned value depends on the problem type:
+     * Result::InsufficientCondition - an internal error possibly with memory allocation,
+     * Result::InvalidArguments - if unknown engine type chosen,
+     * Result::NonSupport - in case the engine type in the 'meson_options.txt' differs from the one specified in the @p engine argument,
+     * Result::Unknown - other.
```
JSUYA, line 1114: Hi 
In doxygen syntax, you can use `@retval` to describe the return value.
ex)
```
     * @return Result::Success when succeeded, otherwise the returned value depends on the problem type:
     * @retval Result::InsufficientCondition An internal error possibly with memory allocation,
     * @retval Result::InvalidArguments If unknown engine type chosen,
     * @retval Result::NonSupport In case the engine type in the 'meson_options.txt' differs from the one specified in the @p engine argument,
     * @retval Result::Unknown other. 
```



before
image

after
image

inc/thorvg.h
```diff
@@ -1062,29 +1097,32 @@ class TVG_EXPORT GlCanvas final : public Canvas
  *
```
JSUYA, line 1097: Please change Engine -> Initializer

inc/thorvg.h
mgrudzinska: thx
Please check now. I used @retval in all cases where were more than 2 return values.
Also, in the case above, I used all ```@retval``` commands. I didn't mixed ```@return``` with ```@retval```.

mgrudzinska:
conflicts resolved

JSUYA (approved):
LGTM

hermet (approved):

PR #364, opened 2021-04-27
doc common: Use @retval to describe return values

JSUYA:
- Description :
retval is used when writing a description of the return value in doxygen syntax.

- Issue Tickets (if any) :
https://github.com/Samsung/thorvg/issues/27


mgrudzinska:
@JSUYA please check #361, I applied you comment to all the cases where there is more than 2 ret values.
I also think that mixing ```@return``` and ```@retval``` is not a good idea

PR #377, opened 2021-05-03
doc: linguistic changes in C++ API docs

mgrudzinska:
- the file description is added
- @defgroup requires 2 args - label and a name
- I changed '@return Result::Success when succeed' to '@return Result::Success' everywhere, where this is the only result that can be returned. I find it confusing as it was, since the question arises "and what if it does not succeed? "
- using @c before Result::Success doesn't look good - it is already marked as link, so do we have to change its font?
before:
docs_before
after:
docs_after

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1,6 +1,14 @@
 /*!
-* @file thorvg.h
-*/
+ * @file thorvg.h
+ *
+ * The main APIs enabling the TVG initialization, preparation of the canvas and provisioning of its content:
+ * - drawing shapes such as line, curve, arc, rectangle, circle or user-defined
+ * - drawing pictures - SVG, PNG, RAW
+ * - solid or gradient filling
+ * - continuous and dashed stroking
+ * - clipping and masking
+ * and finally drawing the canvas and TVG termination.
+ */
 
 
 #ifndef _THORVG_H_
@@ -50,7 +58,8 @@ class Picture;
 class Canvas;
 
 /**
- * @defgroup ThorVG
+ * @defgroup ThorVG ThorVG
+ * @brief ThorVG classes and enumerations providing C++ APIs.
  */
 
 /**@{*/
@@ -60,7 +69,7 @@ class Canvas;
  */
 enum class TVG_EXPORT Result
 {
-    Success = 0,           ///< The value returned in case of the correct request execution.
+    Success = 0,           ///< The value returned in case of a correct request execution.
     InvalidArguments,      ///< The value returned in the event of a problem with the arguments given to the API - e.g. empty paths or null pointers.
     InsufficientCondition, ///< The value returned in case the request cannot be processed - e.g. asking for properties of an object, which does not exist.
     FailedAllocation,      ///< The value returned in case of unsuccessful memory allocation.
@@ -77,10 +86,10 @@ enum class TVG_EXPORT Result
  */
 enum class TVG_EXPORT PathCommand
 {
-    Close = 0, ///< Ends the current sub-path and connects it with its initial point. This command expects 1 point: close position.
-    MoveTo,    ///< Sets a new initial point of the sub-path and a new current point. This command expects 1 point: moving position.
-    LineTo,    ///< Draws a line from the current point to the given point and sets a new value of the current point. This command expects 1 point: end position of the line.
-    CubicTo    ///< Draws a cubic Bezier curve from the current point to the given point using two given control points and sets a new value of the current point. This command expects 3 points: control 1, control 2, end.
+    Close = 0, ///< Ends the current sub-path and connects it with its initial point. This command doesn't expect any points.
+    MoveTo,    ///< Sets a new initial point of the sub-path and a new current point. This command expects 1 point: the starting position.
+    LineTo,    ///< Draws a line from the current point to the given point and sets a new value of the current point. This command expects 1 point: the end-position of the line.
+    CubicTo    ///< Draws a cubic Bezier curve from the current point to the given point using two given control points and sets a new value of the current point. This command expects 3 points: the 1st control-point, the 2nd control-point, the end-point of the curve.
 };
 
 /**
@@ -134,7 +143,7 @@ enum class TVG_EXPORT CompositeMethod
 };
 
 /**
- * @brief Enumeration specifying the engine type used for the graphics backend. For multiple backeneds, Bitwise operation is allowed.
+ * @brief Enumeration specifying the engine type used for the graphics backend. For multiple backeneds bitwise operation is allowed.
  */
 enum class TVG_EXPORT CanvasEngine
 {
@@ -170,12 +179,11 @@ struct Matrix
 /**
  * @class Paint
  *
- * @brief A abstract class for managing graphical elements.
+ * @brief An abstract class for managing graphical elements.
  *
- * Basically a graphical elements in TVG is composed as an object in a scene (Canvas).
- * Paint represents a graphical object, suggest common behaviors such as duplication, transformation and composition, etc.
- * TVG recommends user to regard a paint as a pack of volatile commands, they can prepare a paint then request Canvas to run it.
- *
+ * A graphical element in TVG is any object composed into a Canvas.
+ * Paint represents such a graphical object and its behaviors such as duplication, transformation and composition.
+ * TVG recommends the user to regard a paint as a set of volatile commands. They can prepare a Paint and then request a Canvas to run them.
  */
 class TVG_EXPORT Paint
 {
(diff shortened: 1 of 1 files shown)


inc/thorvg.h
```diff
@@ -1119,24 +1128,24 @@ class TVG_EXPORT SwCanvas final : public Canvas
      *
      * The buffer of a desirable size should be allocated and owned by the caller.
      *
-     * @param[in] buffer A pointer to a memory block of the size @p w x @p h, where the raster data are stored.
+     * @param[in] buffer A pointer to a memory block of the size @p stride x @p h, where the raster data are stored.
```
mgrudzinska, line 1131: in general the buffer size is stride x h, right?

hermet: @mgrudzinska correct.

inc/thorvg.h
```diff
@@ -233,19 +240,19 @@ class TVG_EXPORT Paint
      *
      * @param[in] o The opacity value in the range [0 ~ 255], where 0 is completely transparent and 255 is opaque.
      *
-     * @return @c Result::Success.
+     * @return Result::Success.
      *
-     * @note A half-translucent paint by opacity may require multiple render pass for composition, TVG recommends to avoid opacity change as possible.
+     * @note Setting the opacity with this API may require multiple renderings using a composition. It is recommended to avoid changing the opacity if possible.
```
hermet, line 245: "render pass" is more clear.

inc/thorvg.h
```diff
@@ -170,49 +179,47 @@ struct Matrix
 /**
  * @class Paint
  *
- * @brief A abstract class for managing graphical elements.
- *
- * Basically a graphical elements in TVG is composed as an object in a scene (Canvas).
- * Paint represents a graphical object, suggest common behaviors such as duplication, transformation and composition, etc.
- * TVG recommends user to regard a paint as a pack of volatile commands, they can prepare a paint then request Canvas to run it.
```
hermet, line 177: Do you need to remove the last sentence? It's a tip for tvg concept.

inc/thorvg.h
```diff
@@ -258,16 +265,16 @@ class TVG_EXPORT Paint
      * @param[out] w The width of the object.
      * @param[out] h The height of the object.
      *
-     * @return @c Result::Success when succeed, @c Result::InsufficientCondition otherwise.
+     * @return Result::Success when succeed, Result::InsufficientCondition otherwise.
      *
-     * @note bounding box doesn't indicate the rendering region in the result but primitive region of the object.
+     * @note The bounding box doesn't indicate the final rendered region. It's the rectangular bounding area of the object.
```
hermet, line 270: is there any clear words to express it instead of "rectangular bounding area" ?

inc/thorvg.h
```diff
@@ -317,9 +324,9 @@ class TVG_EXPORT Fill
      * @brief Sets the parameters of the colors of the gradient and their position.
      *
      * @param[in] colorStops An array of ColorStop data structure.
-     * @param[in] cnt The count of the @p colorStops array equal to the colors number used in the gradient.
+     * @param[in] cnt The size of the @p colorStops array equal to the colors number used in the gradient.
```
hermet, line 327: count is more clear. its parameter says "count"
size is ambiguous since  data size * data count

inc/thorvg.h
```diff
@@ -317,9 +324,9 @@ class TVG_EXPORT Fill
...
      *
-     * @return @c Result::Success when succeed.
```
hermet, line 322: We should not fix it, it always return Result::Success.

inc/thorvg.h
```diff
@@ -394,63 +401,64 @@ class TVG_EXPORT Canvas
      * @brief Passes drawing elements to the Canvas using Paint objects.
      *
      * Only pushed paints in the canvas will be drawing targets.
-     * They are retained by the canvas until you call Canvas::clear().
-     * If you know the count of push() in the advance, please call reserve().
+     * They are retained by the canvas until you call clear().
```
hermet, line 404: Canvas::clear()
Canvas::reserve()

inc/thorvg.h
```diff
@@ -394,63 +401,64 @@ class TVG_EXPORT Canvas
...
+     * If a @c nullptr is passed all paint objects retained by the Canvas are updated,
+     * otherwise only the paint to which the given @p paint points.
      *
      * @param[in] paint A pointer to the Paint object or @c nullptr.
      *
-     * @return @c Result::Success when succeed, @c Result::InsufficientCondition otherwise.
+     * @return Result::Success when succeed, Result::InsufficientCondition otherwise.
      *
-     * @note Updation behavior could be asynchronous by thread option.
+     * @note Update behavior can be asynchronous based on the assigned thread number.
```
hermet, line 441: number is ambiguous...  just async/sync. either thread or not.

inc/thorvg.h
```diff
@@ -394,63 +401,64 @@ class TVG_EXPORT Canvas
...
 
     /**
-     * @brief Request the canvas to draw paint objects.
+     * @brief Request the canvas to draw the Paint objects.
      *
-     * @return @c Result::Success when succeed, @c Result::InsufficientCondition otherwise.
+     * @return Result::Success when succeed, Result::InsufficientCondition otherwise.
      *
-     * @note Drawing could be asynchronous by thread option. To guarantee the done of drawing job, call sync().
+     * @note Drawing can be asynchronous based on the assigned thread number. To guarantee the drawing is done, call sync() afterwards.
```
hermet, line 450: same. thread number

inc/thorvg.h
```diff
@@ -477,31 +485,31 @@ class TVG_EXPORT LinearGradient final : public Fill
...
+     * the given points (@p x1, @p y1) and (@p x2, @p y2), respectively. Both lines are perpendicular to the line linking
+     * (@p x1, @p y1) and (@p x2, @p y2).
      *
      * @param[out] x1 The horizontal coordinate of the first point used to determine the gradient bounds.
      * @param[out] y1 The vertical coordinate of the first point used to determine the gradient bounds.
      * @param[out] x2 The horizontal coordinate of the second point used to determine the gradient bounds.
      * @param[out] y2 The vertical coordinate of the second point used to determine the gradient bounds.
      *
-     * @return @c Result::Success when succeed.
+     * @return Result::Success.
```
hermet, line 512: It's open case. please keep it "when succeed". it could return any error value.

inc/thorvg.h
```diff
@@ -1039,10 +1047,10 @@ class TVG_EXPORT Picture final : public Paint
  *
  * @brief A class to composite children paints.
  *
- * As the traditional graphics rendering method, TVG also enables scene-graph mechanism,
+ * As the traditional graphics rendering method, TVG also enables scene-graph mechanism.
  * This feature supports an array function for managing the multiple paints as one group paint.
  *
- * As a group, scene can be transformed, translucent, composited with other target paints,
+ * As a group, the scene can be transformed, made opaque and composited with other target paints,
```
hermet, line 1053: We intent to say "translucent." not opaque

hermet (changes requested):
good job. Please check comments.

inc/thorvg.h
hermet: all other changes are same.

inc/thorvg.h
mgrudzinska: done

inc/thorvg.h
mgrudzinska: "The Update behavior can be asynchronous if the assigned thread number is greater than zero.
* @see Initializer::init()"

if you prefer we can stay with 'thread option' although it is not clear to me, which option do we mean:
"The Update behavior can be asynchronous depending on the thread option."

inc/thorvg.h
mgrudzinska: done

inc/thorvg.h
hermet: @mgrudzinska  ""The Update behavior can be asynchronous if the assigned thread number is greater than zero." this one is much clear.

inc/thorvg.h
hermet: @mgrudzinska then it'd rather this:
The count of the @p colorStop equal ... 
I think size can't be applied this. it requires exact number of the element. not size.

inc/thorvg.h
hermet: @mgrudzinska great.

inc/thorvg.h
hermet: @mgrudzinska volatile means commands are just one single input that turns to invalid soon. So you suggestion can't satisfy with the meaning.

inc/thorvg.h
hermet: @mgrudzinska "render pass" "composition" are a sort of terms those are used among the graphics world.

hermet (changes requested):
Please check comments.

inc/thorvg.h
mgrudzinska: restored as it was originally: 
"Setting the opacity with this API may require multiple render pass for composition."

inc/thorvg.h
mgrudzinska: restored you version:
"TVG recommends user to regard a paint as a set of volatile commands. They can prepare a Paint and then request a Canvas to run them."

inc/thorvg.h
mgrudzinska: restored your version

hermet (approved):
very nice. thank you.

PR #378, opened 2021-05-05
capi: added check against null pointer before dereferencing

mgrudzinska:


src/bindings/capi/tvgCapi.cpp
```diff
@@ -528,7 +528,7 @@ TVG_EXPORT Tvg_Result tvg_gradient_set_spread(Tvg_Gradient* grad, const Tvg_Stro
 TVG_EXPORT Tvg_Result tvg_gradient_get_spread(Tvg_Gradient* grad, Tvg_Stroke_Fill* spread)
 {
     if (!grad) return TVG_RESULT_INVALID_ARGUMENT;
-    *spread = (Tvg_Stroke_Fill) reinterpret_cast<Fill*>(grad)->spread();
+    if (spread) *spread = (Tvg_Stroke_Fill) reinterpret_cast<Fill*>(grad)->spread();
```
hermet, line 531: I think it returns error when spread is valid.... -> TVG_RESULT_INVALID_ARGUMENT?

hermet (changes requested):
Please check a comment.

src/bindings/capi/tvgCapi.cpp
mgrudzinska: Please check now whether it is what you meant.
I also unified the file formatting - not sure whether I was allowed to do it is this pr...

src/bindings/capi/tvgCapi.cpp
```diff
@@ -460,95 +461,109 @@ TVG_EXPORT Tvg_Result tvg_picture_get_viewbox(const Tvg_Paint* paint, float* x,
...
 {
     if (!scene) return TVG_RESULT_INVALID_ARGUMENT;
     return (Tvg_Result) reinterpret_cast<Scene*>(scene)->reserve(size);
 }
 
+
 TVG_EXPORT Tvg_Result tvg_scene_push(Tvg_Paint* scene, Tvg_Paint* paint)
 {
     if (!scene || !paint) return TVG_RESULT_INVALID_ARGUMENT;
     return (Tvg_Result) reinterpret_cast<Scene*>(scene)->push(unique_ptr<Paint>((Paint*)paint));
```
mgrudzinska, line 563: I noticed that there are 4 places where we use c-type casting.
I'd use reinterpret_cast - what do you think?

hermet: @mgrudzinska  well... frankly i don't care. i just say there are few reason when you do code-refactoring.
1. Keep code consistency.
2. Keep less code
3. Increase Readibility
4. Any behavioral benefits.

hermet (approved):

PR #383, opened 2021-05-17
sw_engine: removing unnecessary variable from SwRadial struct

mgrudzinska:
Removing the 'inv2a' variable reduced the number of multiplications.

### DIFF
--- a/src/lib/sw_engine/tvgSwFill.cpp
+++ b/src/lib/sw_engine/tvgSwFill.cpp
@@ -156,7 +156,7 @@ bool _prepareRadial(SwFill* fill, const RadialGradient* radial, const Matrix* tr
     }
 
     fill->radial.a = radius * radius;
-    fill->radial.inv2a = pow(1 / (2 * fill->radial.a), 2);
+    fill->radial.inva = 1.0 / fill->radial.a;
 
     return true;
 }
@@ -210,12 +210,12 @@ void fillFetchRadial(const SwFill* fill, uint32_t* dst, uint32_t y, uint32_t x, 
     //Rotation
     auto rx = (x + 0.5f - fill->radial.cx) * fill->sy;
     auto ry = (y + 0.5f - fill->radial.cy) * fill->sx;
-    auto inv2a = fill->radial.inv2a;
     auto rxy = rx * rx + ry * ry;
     auto rxryPlus = 2 * rx;
-    auto det = (-4 * fill->radial.a * -rxy) * inv2a;
-    auto detDelta = (4 * fill->radial.a * (rxryPlus + 1.0f)) * inv2a;
-    auto detDelta2 = (4 * fill->radial.a * 2.0f) * inv2a;
+    auto inva = fill->radial.inva;
+    auto det = rxy * inva;
+    auto detDelta = (rxryPlus + 1.0f) * inva;
+    auto detDelta2 = 2.0f * inva;
 
    for (uint32_t i = 0 ; i < len ; ++i) {
         *dst = _pixel(fill, sqrt(det));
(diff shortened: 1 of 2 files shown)


src/lib/sw_engine/tvgSwFill.cpp
```diff
@@ -210,12 +209,11 @@ void fillFetchRadial(const SwFill* fill, uint32_t* dst, uint32_t y, uint32_t x,
...
     auto ry = (y + 0.5f - fill->radial.cy) * fill->sx;
-    auto inv2a = fill->radial.inv2a;
     auto rxy = rx * rx + ry * ry;
     auto rxryPlus = 2 * rx;
-    auto det = (-4 * fill->radial.a * -rxy) * inv2a;
-    auto detDelta = (4 * fill->radial.a * (rxryPlus + 1.0f)) * inv2a;
-    auto detDelta2 = (4 * fill->radial.a * 2.0f) * inv2a;
+    auto det = rxy / fill->radial.a;
+    auto detDelta = (rxryPlus + 1.0f) / fill->radial.a;
+    auto detDelta2 = 2.0f / fill->radial.a;
```
JSUYA, line 216: In general, Division is slower than multiplication.
What benefits can we get from this refactoring?

mgrudzinska: I hanged to multiplication. 

- i think the code is more readable without this variables
- we have one float less in the SwRadial  struct
- in total we have only 3 multiplications instead of 9:
    * 1 instead of 3 in 'det'
    *  1 instead of 3 in 'detDelta'
    *  1 instead of 3 'detDelat2'

do you think it's not worth it?

mgrudzinska: not to mention that we have now 2 LOC less ;)

hermet: @mgrudzinska Here question is which function is more serious.  sometimes we can give up readibility for performance.
so..
feching is called per pixels vs prepare is called just one time.
let's think.

src/lib/sw_engine/tvgSwFill.cpp
```diff
@@ -210,12 +209,12 @@ void fillFetchRadial(const SwFill* fill, uint32_t* dst, uint32_t y, uint32_t x,
     //Rotation
     auto rx = (x + 0.5f - fill->radial.cx) * fill->sy;
     auto ry = (y + 0.5f - fill->radial.cy) * fill->sx;
-    auto inv2a = fill->radial.inv2a;
     auto rxy = rx * rx + ry * ry;
     auto rxryPlus = 2 * rx;
-    auto det = (-4 * fill->radial.a * -rxy) * inv2a;
-    auto detDelta = (4 * fill->radial.a * (rxryPlus + 1.0f)) * inv2a;
-    auto detDelta2 = (4 * fill->radial.a * 2.0f) * inv2a;
+    auto inva = 1.0 / fill->radial.a;
```
hermet, line 214: Apart from the logic changes,

if we keep the inva in the data structure (as it does), we can avoid dividing operation per pixels.

hermet (changes requested):
Please check a comment.

src/lib/sw_engine/tvgSwFill.cpp
mgrudzinska: @hermet you are very right, sorry
so now the benefits are no 'pow' called and less multiplications. it's still something

src/lib/sw_engine/tvgSwFill.cpp
mgrudzinska: I restored it in the struct, but I used inva instead of inv2a. Am I missing something? do we need inv2a?

JSUYA (approved):
Cool! I think this change is good.

src/lib/sw_engine/tvgSwFill.cpp
hermet: @mgrudzinska neither i have no idea. what brought you to change to this code. since you introduced new code, i wish you can be sure by yourself first... :)

Issue #381, opened 2021-05-12
SVG fill="url(#…)" doesnt work if opacity set


mmaciola:
Svg images that use fill="url(#..)", such as `radialgrad1.svg` and `lineargrad1.svg`, won't display if opacity is set.
Example code:
```
auto svg = tvg::Picture::gen();
if (svg->load(EXAMPLE_DIR"/lineargrad1.svg") != tvg::Result::Success) return;
svg->opacity(100);

canvas->push(move(svg));
```

mgrudzinska:
this is solved in #326 

Because the opacity < 255 is set, the composition is applied, but currently there is no code to handle such a case (only alpha mask and inv mask are handled, since the 'if else' statements are not well handled).
In #326 the code is refactored - a function is added, that decides which secondary funct should be called, so all cases are handled properly.

mgrudzinska:
since #326 will be splitted into smaller patches, the problem is solved in #384

mmaciola:
After patch works ok

PR #384, opened 2021-05-18
sw_engine: fixing wrong 'if else' statement handling when gradient sh…

mgrudzinska:
…apes rastered

The cases when composition was applied were handled only for AlphaMask and
InvAlphaMask. When opacity value was to be < 255, there was no code
to handle this.

issue #381 

before:
before

after:
after


mgrudzinska:
I forgot about rle  in the first commit

mmaciola (approved):
Tested, looks ok. Gradient works properly after this patch

hermet (approved):

PR #385, opened 2021-05-18
sw_engine: gradient shapes with opacity < 255 rastered properly

mgrudzinska:
The cases with gradient shapes with composition are handled
in the same function as gradint shapes with opacity < 255.
Parts of the code from _rasterOpaque... grad functions moved to
_rasterTranslucent... grad functions.

before:
before

after:
after

This pr contains #384 

Please note, that the only change is that I moved some part of the code from one function into another.


hermet (approved):

hermet:
@mgrudzinska could you please rebase your patch ?

PR #386, opened 2021-05-21
sw_engine: rearrange the blending mechanism.

hermet:
Revised the blending mechanism by rearrange alpha value range from 0 to 256.
This is more clear than previous 255 alpha value range approach
because tvg uses bit-operation for faster results.
In the sequence, it divides alpha values by 256, it'd have chance to lose the value by 1
during 255 / 256. And then accumulated result brings the a bit different coloring.

### DIFF
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -117,7 +117,7 @@ struct SwSpan
 {
     int16_t x, y;
     uint16_t len;
-    uint8_t coverage;
+    uint16_t coverage;
 };
 
 struct SwRleData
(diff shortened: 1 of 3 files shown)


mgrudzinska:
@hermet In this branch I still see the difference in color, that I mentioned in #231, so I'm not sure how much do we gain by these changes.

It's a bit hard to read, because we have alpha and opacity, both in the range 0-255, but later we change the range only for the alpha to be 1-256 (_colorAlpha always returns values > 0). 

I think that the lower value of this range (1 instead of 0) is problematic. Lets take alpha mask - we need to call the ALPHA_BLEND(destination, 256-alpha). This alpha is always at least 1, so in result this acts as before, when it was 255-alpha.

Also please notice, that if the arguments of the ALPHA_MULTIPLY are: one in the range 0-255 and the second one 1-256, then the result, after dividing by 256, is still in the range 0-255, and when this value goes to the ALPHA_BLEND, nothing really changes and it works as before this patch.

hermet:
closed by this https://github.com/Samsung/thorvg/pull/231

PR #388, opened 2021-05-24
TVG Loader/Saver module: tvg helper file

mmaciola:
tvgTvgHelper.h is a base file used for .tvg saver/loader module

### DIFF
--- a/src/lib/tvgTvgHelper.h
+++ b/src/lib/tvgTvgHelper.h
@@ -1,0 +1,103 @@
+/*
+ * Copyright (c) 2020-2021 Samsung Electronics Co., Ltd. All rights reserved.
+
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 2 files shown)


src/lib/tvgTvgHelper.h
```diff
@@ -0,0 +1,82 @@
+#ifndef _TVG_STANDARD_HELPER_H_
```
JSUYA, line 1: Is there any reason this Helper header file should be in src/lib?
I think it should be in src/loaders/tvg if it's only used in tvgTvgLoader.
And if you move this header file, I think proper name of this file is tvgTvgLoaderHelper.h.

src/lib/tvgTvgHelper.h
```diff
@@ -0,0 +1,82 @@
...
+#define _TVG_STANDARD_HELPER_H_
+
+#include "tvgCommon.h"
```
JSUYA, line 4: Do we need this?

JSUYA (changes requested):
Please check my comment

src/lib/tvgTvgHelper.h
mgrudzinska: it is also used by the saver, but I need a moment to decide where the saver will be implemented

src/lib/tvgTvgHelper.h
```diff
@@ -0,0 +1,82 @@
...
+#define _read_tvg_float(dst, src) memcpy(dst, (src), sizeof(float))
+
+using TvgIndicator = uint8_t;
+using ByteCounter = uint32_t;
+using TvgFlag = uint8_t;
+#define TVG_INDICATOR_SIZE sizeof(TvgIndicator)
+#define BYTE_COUNTER_SIZE sizeof(ByteCounter)
+#define TVG_FLAG_SIZE sizeof(TvgFlag)
+
+struct tvg_block
```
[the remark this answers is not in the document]
mmaciola: renamed to tvgBlock

src/lib/tvgTvgHelper.h
mmaciola: removed include line

src/lib/tvgTvgHelper.h
mmaciola: Renamed `_TVG_STANDARD_HELPER_H_` into `_TVG_TVG_HELPER_H_` and added copyright header

JSUYA (approved):
I think this PR is appropriate.
When the review of other patches is complete, I will merge this.

hermet (changes requested):
@mmaciola Looks bad that this file make a coupling between tvg loader & common lib.
In point of tvg common lib view, It's not necessary. How about moving the loader side?
Also, commit message is too weak. at least you can mention any rules&policy to maintain these macro values.

hermet:
@mmaciola + question, is that right whenever new property is added, we should update the macro and loader code?

mmaciola:
This file will be used by saver module too; it must be accessible for both loader and saver.

When new features added, we will just add new definitions into this file. Existing definitions won't be changed. Loader and saver modules will be then changed accordingly to handle new features.

I changed commit message. Please write what you think

hermet (approved):

PR #399, opened 2021-05-27
svg_loader: the viewBox clipping composite layer added independently …

mgrudzinska:
…of the shapes bounds

For now the bounding box of all the shapes was established and if it was larger
than the viewBox of the SVG, the clipping layer was added. The bounds api
returns the rectangle that encloses the shapes before any transformations.
So comparing it with the viewBox doesn't make sense. The comparison is removed
and the clipping layer is always added.

Example illustrating the problem: The shape bounds and the viewBox are exactly the same,
so the clipping layer is not added. But the shape is later transformed and goes outside the viewBox,
so it should be clipped.

code:
```
<svg viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg">
<path d="M 0 0 h 200 v 200 h -200 z" fill="#F00" stroke-width="0" transform="translate(100,100)" />
</svg>
```

before:
before

after:
after

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -402,25 +402,18 @@ unique_ptr<Scene> svgSceneBuild(SvgNode* node, float vx, float vy, float vw, flo
 {
     if (!node || (node->type != SvgNodeType::Doc)) return nullptr;
 
-    unique_ptr<Scene> root;
     auto docNode = _sceneBuildHelper(node, vx, vy, vw, vh);
-    float x, y, w, h;
 
-    if (docNode->bounds(&x, &y, &w, &h) != Result::Success) return nullptr;
+    auto viewBoxClip = Shape::gen();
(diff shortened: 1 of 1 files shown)


JSUYA (approved):
Thank you for checking corner case.
Adding a clipping layer to every svg can obviously affect performance.
But I don't think there is any way now.
Because we don't know the boundary to which the transform is applied at the current step.

At least if we take action...
In my opinion... applying the composite layer is handled by load() of the picture, and considering the bounds of the docNode can be handled as optional.
Other than that, I don't have a good idea either.;;

+
Is this the problem you talked about yesterday?

JSUYA:
@hermet Any ideas?

mgrudzinska:
I thought that since we have all information about the shapes in the svg loader (paths and transformations), so we could find the real bbox. but it will take `a few calculations`, so I'm not sure whether it optimal.

hermet (approved):

hermet:
@mgrudzinska @JSUYA  if there is a reasonable reason to fix and no better ideas then we can accept this!

PR #400, opened 2021-05-28
common: added API to get the transformation matrix of the object

mgrudzinska:
I need this API in the saver module.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -232,9 +232,20 @@ public:
      * @param[in] m The 3x3 augmented matrix.
      *
      * @return Result::Success when succeed, Result::FailedAllocation otherwise.
-     *
      */
     Result transform(const Matrix& m) noexcept;
+
+    /**
+     * @brief Gets the matrix of the affine transformation of the object.
+     *
+     * The values of the matrix can be set by the transform() API, as well by the translate(),
+     * scale() and rotate(). In case no transformation was applied, the identity matrix is returned.
+     *
+     * @retval The augmented transformation matrix.
+     *
+     * @BETA_API
+     */
+    Matrix transform() noexcept;
 
     /**
      * @brief Sets the opacity of the object.
--- a/src/lib/tvgPaint.cpp
+++ b/src/lib/tvgPaint.cpp
@@ -281,6 +281,12 @@ Result Paint::transform(const Matrix& m) noexcept
     return Result::FailedAllocation;
 }
 
+Matrix Paint::transform() noexcept
+{
+    auto pTransform = pImpl->transform();
+    if (pTransform) return *pTransform;
+    return {1, 0, 0, 0, 1, 0, 0, 0, 1};
+}
 
 Result Paint::bounds(float* x, float* y, float* w, float* h) const noexcept
 {
--- a/src/lib/tvgPaint.h
+++ b/src/lib/tvgPaint.h
@@ -74,6 +74,15 @@ namespace tvg
             return true;
         }
 
+        Matrix* transform()
+        {
+            if (rTransform) {
+                rTransform->update();
+                return &rTransform->m;
+            }
+            return nullptr;
+        }
+
         bool bounds(float* x, float* y, float* w, float* h) const
         {
             return smethod->bounds(x, y, w, h);


inc/thorvg.h
```diff
@@ -232,10 +232,20 @@ class TVG_EXPORT Paint
...
+    /**
+     * @brief Gets the matrix of the affine transformation of the object.
+     *
+     * @param[out] m The 3x3 augmented matrix.
+     *
+     * @retval Result::Success when succeed.
+     * @retval Result::InvalidArguments If a @c nullptr is passed as the argument.
+     * @retval Result::InsufficientCondition In case no transformation was set.
+     */
+    Result transform(const Matrix** m) const noexcept;
```
hermet, line 247: the internal matrix data is very fragle.. (it can be freed or values are changed anytime) 
Thus I suggest either way:
1. Matrix transform();   
2. Result transform(Matrix& m);

The first one is the best as an api  for me but I'm not sure the exact situation of your scenario... Such If you're trying to avoid data copy for optimization point.

hermet (changes requested):
Please check a comment.

inc/thorvg.h
hermet: Also, you should not add const here because that paint data can be updated.

mgrudzinska: @hermet please check now, I returned the matrix by val.
yes, you're right, I didn't want to copy the data, but I think in this case it's ok

I removed const. btw - is the composite API in thorvg.h - is there `const` ok?

src/bindings/capi/tvgCapi.cpp
```diff
@@ -162,6 +162,17 @@ TVG_EXPORT Tvg_Result tvg_paint_transform(Tvg_Paint* paint, const Tvg_Matrix* m)
 }
 
 
+TVG_EXPORT Tvg_Result tvg_paint_get_transform(Tvg_Paint* paint, Tvg_Matrix* m)
+{
+    if (!paint || !m) return TVG_RESULT_INVALID_ARGUMENT;
+    tvg::Matrix mTranfs = reinterpret_cast<Paint*>(paint)->transform();
```
hermet, line 168: -- tvg prefix.

inc/thorvg.h
```diff
@@ -232,10 +232,19 @@ class TVG_EXPORT Paint
...
      */
     Result transform(const Matrix& m) noexcept;
 
+    /**
+     * @brief Gets the matrix of the affine transformation of the object.
+     *
+     * The values of the matrix can be set by the transform() API, as well by the translate(),
+     * scale() and rotate(). In case no transformation was applied, the identity matrix is returned.
+     *
+     * @retval The augmented transformation matrix.
```
hermet, line 244: Please add beta tags for new apis. Those will be turn out to official one when they are ready.
+@BETA_API

src/bindings/capi/tvgCapi.cpp
```diff
@@ -162,6 +162,17 @@ TVG_EXPORT Tvg_Result tvg_paint_transform(Tvg_Paint* paint, const Tvg_Matrix* m)
...
+    m->e11 = mTranfs.e11; m->e12 = mTranfs.e12; m->e13 = mTranfs.e13;
```
hermet, line 169: I think you can cast the matrix simply like this.
*m = (Tvg_Matrix) reinterpret_cast<Paint*>(paint)->transform();

src/lib/tvgPaint.cpp
```diff
@@ -281,6 +281,14 @@ Result Paint::transform(const Matrix& m) noexcept
     return Result::FailedAllocation;
 }
 
+Matrix Paint::transform() noexcept
+{
+    Matrix m = {1, 0, 0, 0, 1, 0, 0, 0, 1};
```
hermet, line 286: you have two memory allocation.
see below:
auto pTransform = pImpl->transform();
if (pTransform) return *pTransform;
else return {1, 0, 0, 0, 1, 0, 0, 0, 1};

hermet (changes requested):
@mgrudzinska next time please separate c / capi patches.

inc/thorvg.h
hermet: @mgrudzinska no, it's wrong :-@

hermet:
@mgrudzinska https://github.com/Samsung/thorvg/commit/2c37191584b18a532cd01e7986fc5f53ebeea80d

inc/thorvg.h
mgrudzinska: done

mgrudzinska:
I removed Capi into separate pr

hermet (approved):

hermet (approved):

PR #401, opened 2021-05-28
Support Blending Modes

mmaciola:
Patch introduces blending options such as Normal, Screen, Multiply, Overlay,
Darken, Lighten, ColorDodge, ColorBurn, HardLight, SoftLight, Difference,
Exclusion

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -151,6 +151,11 @@ enum class TVG_EXPORT CanvasEngine
     Gl = (1 << 2)  ///< OpenGL rasterizer.
 };
 
+/**
+ * @brief Enumeration specifying the blending mode used for paint blending.
+ */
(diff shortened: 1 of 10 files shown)


hermet:
@mmaciola I will check it. That would be better if we made a consensus for the interfaces prior to  implementate bodies.

mmaciola:
Updated commit message, optimized `BLEND_COLORS()` function and fixed for translucent rle

projectitis:
@hermet @mmaciola I would really like blending modes supported :) 

What is required to this work to make it suitable for ThorVG?  I am happy to contribute.

Hermet do you suggest a different interface? Please explain your idea.

hermet:
@projectitis Although I didn't review this yet, I guess this approach is good. I think we can work on this after v0.9 release?

hermet:
replace with this: https://github.com/thorvg/thorvg/pull/1500

PR #408, opened 2021-06-03
test: remove gtest & its infra.

hermet:
thorvg is going to use catch2 framework, this is a cleanup work
before introducing catch2 utc.

See Catch2:
https://github.com/catchorg/Catch2/tree/v2.x


JSUYA (approved):
good bye gtest

PR #411, opened 2021-06-04
svg_loader SceneBuilder: Remove incorrect calculation for gradient

JSUYA:
Calculation of objectBoundingBox case in gradient is incorrect code.
This causes the x,y coordinates of the gradient to be calculated incorrectly.
So remove it.

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -93,7 +93,6 @@ static unique_ptr<RadialGradient> _applyRadialGradientProperty(SvgStyleGradient*
 {
     Fill::ColorStop *stops;
     int stopCount = 0;
-    float gx, gy, gw, gh;
     int radius;
     float fillOpacity = 255.0f;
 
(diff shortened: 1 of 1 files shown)


hermet (commented):
@mgrudzinska leave it to your approval.

mgrudzinska:
@JSUYA I would also remove the analogous part of the code for the radial gradient.

could you please tell what was the original purpose of adding this part of the code? 
I thought that maybe it was some kind of optimization, but in such a case we would change the colors together with the bounds of the gradient, so it's not the case.

JSUYA:
> @JSUYA I would also remove the analogous part of the code for the radial gradient.
> 
> could you please tell what was the original purpose of adding this part of the code?
> I thought that maybe it was some kind of optimization, but in such a case we would change the colors together with the bounds of the gradient, so it's not the case.

As you may know, this parsing code is based on efl's svg parser.
The deleted code does not exist linear gradient(in the parser of the efl), but it exists in the radial gradient.
Perhaps it was added incorrectly in the process of implementing.


I checked the history of the code in the radial gradient.
We are doing enough calculations in the parser if the unit is user space.
When I did some tests, this code was calculating the results of cx and cy incorrectly.
So... now I think it's unnecessary.
(But I think there can be side effects or corner cases that I didn't think of...)

I added more patch that remove unnecessary code in radial gradient.
Please cross check these pathces.

mgrudzinska:
@JSUYA I tried to find some cases, in which there would be a problem when we remove this part of the code, but I couldn't find any. I found some cases of radial grad, for which the results are incorrect after this recalculations. 
So I think this part of the code should be removed.

mgrudzinska (approved):

PR #414, opened 2021-06-04
tests: add SwCanvas unit tests

hermet:


hermet:
This requires this patch: https://github.com/Samsung/thorvg/commit/eda022365f7e166a857cdb775acee437074e0849

test/testSwCanvas.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <thorvg.h>
+#include "catch.hpp"
+
+using namespace tvg;
+
+
+TEST_CASE("Missing Initialization", "[tvgSwCanvas]")
+{
+    auto canvas = SwCanvas::gen();
+    REQUIRE(canvas == nullptr);
```
mgrudzinska, line 32: @hermet does this line work correctly?

hermet: @mgrudzinska yes with above patches.

mgrudzinska: sorry, I just noticed...

test/testSwCanvas.cpp
```diff
@@ -0,0 +1,86 @@
...
+    auto canvas = SwCanvas::gen();
+    REQUIRE(canvas);
+
+    REQUIRE(canvas->mempool(SwCanvas::MempoolPolicy::Default) == Result::Success);
+    REQUIRE(canvas->mempool(SwCanvas::MempoolPolicy::Individual) == Result::Success);
+    REQUIRE(canvas->mempool(SwCanvas::MempoolPolicy::Shareable) == Result::Success);
+    REQUIRE(canvas->mempool(SwCanvas::MempoolPolicy::Default) == Result::Success);
+    REQUIRE(canvas->mempool(SwCanvas::MempoolPolicy::Individual) == Result::Success);
+    REQUIRE(canvas->mempool(SwCanvas::MempoolPolicy::Shareable) == Result::Success);
+
```
mgrudzinska, line 84: I understand, that it's not the place to check, what will happen if we try to change the mempool policy after pushing something into canvas, right? but if not now then when?

hermet: @mgrudzinska it will return InsufficientConditin as the doc say.

    /**
     * @brief Set sw engine memory pool behavior policy.
     *
     * Basically ThorVG draws a lot of shapes, it allocates/deallocates a few chunk of memory
     * while processing rendering. It internally uses one shared memory pool
     * which can be reused among the canvases in order to avoid memory overhead.
     *
     * Thus ThorVG suggests memory pool policy to satisfy user demands,
     * if it needs to guarantee the thread-safety of the internal data access.
     *
     * @param[in] policy Use the shared cache memory. The default value is @c true
     *
     * @retval Result::Success When succeed.
     * @retval Result::InsufficientCondition If the canvas has any paints.
     * @retval Result::NonSupport In case the software engine is not supported.
     *
     * @note When @c policy is set as @c MempoolPolicy::Individual, current instance of canvas uses its own individual
     *       memory data that is not shared with others. This is necessary when the canvas is accessed on a worker-thread.
     *
     * @warning It's not allowed after pushing any paints.
     *
     * @BETA_API
    */

mgrudzinska: yes, I know what will happen, I just asked whether we want to check that in the test, but I understand that we want.

hermet: @mgrudzinska good catch. all scenarios as we can imagine should be covered. unfortunately this patch is broken by my mistake but i will update with new one.

PR #415, opened 2021-06-04
common gradient: Fix possible crash in color setter

mihashco:
Memcpy is not allowed on nullptr. If colorStops is invalid colors are cleared.

### DIFF
--- a/src/lib/tvgFill.cpp
+++ b/src/lib/tvgFill.cpp
@@ -43,11 +43,13 @@ Fill::~Fill()
 
 Result Fill::colorStops(const ColorStop* colorStops, uint32_t cnt) noexcept
 {
+    if ((!colorStops && cnt > 0) || (colorStops && cnt == 0)) return Result::InvalidArguments;
+
     if (cnt == 0) {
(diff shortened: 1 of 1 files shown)


src/lib/tvgFill.cpp
```diff
@@ -43,11 +43,11 @@ Fill::~Fill()
 
 Result Fill::colorStops(const ColorStop* colorStops, uint32_t cnt) noexcept
 {
```
hermet, line 45: I think it must return immediately after checking invalidArguments.
ie)
if ((!colorStops && cnt > 0) || (colorStops && cnt ==0)) return invalid arguments;

hermet (changes requested):
Please have a look my comment.

src/lib/tvgFill.cpp
mihashco: colorStops(nullptr, 4) -> invalid argument
colorStops(ptr, 0) -> invalid argument 
colorStops(nullptr, 0) -> erease 

above is ok?

hermet (approved):

PR #416, opened 2021-06-04
utc capi: Added capi initializer, canvas, paint and shape utc

mmaciola:


test/capi/testBasic.cpp
```diff
@@ -0,0 +1,359 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include <thorvg_capi.h>
+#include "catch.hpp"
+
+TEST_CASE("Capi: Basic initialization", "[capiInitializer]")
```
JSUYA, line 26: Capi: Basic initialization <- 
I don't think "Capi" is needed. The second parameter already fully explains that this tc is capi.

test/capi/testBasic.cpp
```diff
@@ -0,0 +1,359 @@
...
+{
+    REQUIRE(tvg_engine_init(TVG_ENGINE_SW | TVG_ENGINE_GL, 0) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_engine_term(TVG_ENGINE_SW | TVG_ENGINE_GL) == TVG_RESULT_SUCCESS);
+}
+
+TEST_CASE("Capi: Canvas initialization", "[capiCanvas]")
```
JSUYA, line 32: I think it's a good idea to create a file for each test case group.
at least..
We classified CAPI into docs groups as shown below, so I think this classification can be referenced.
What do you think?

ThorVG_CAPI (BETA version) | ThorVG C language binding APIs
-- | --
Initializer | A module enabling initialization and termination of the TVG engines
SwCanvas | A module for rendering the graphical elements using the software engine
Paint | A module for managing graphical elements. It enables duplication, transformation and composition
Shape | A module for managing two-dimensional figures and their properties
Gradient | A module managing the gradient fill of objects
Picture | A module enabling to create and to load an image in one of the supported formats: svg, png and raw
Scene | A module managing the multiple paints as one group paint

JSUYA (commented):
Hi, Please check my comment

test/capi/testBasic.cpp
mmaciola: Removed "Capi" prefix

test/capi/testBasic.cpp
mmaciola: Refactored testBasic.c into four independent files

test/capi/testPaint.cpp
```diff
@@ -0,0 +1,113 @@
...
+    REQUIRE(paint != NULL);
+
+    REQUIRE(tvg_paint_transform(paint, &matrix_set) == TVG_RESULT_SUCCESS);
+
+    REQUIRE(tvg_paint_scale(paint, 2.0f) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_paint_scale(paint, 0.5f) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_paint_rotate(paint, 180.0f) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_paint_rotate(paint, 180.0f) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_paint_translate(paint, 10.0f, 10.0f) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_paint_translate(paint, -10.0f, -10.0f) == TVG_RESULT_SUCCESS);
```
mgrudzinska, line 41: I understand that the intention was to check what will happen when we call twice the same API (does it make sense to call is twice with the exact same value? - rotate) but I don't see how do we check whether the behavior is what we expected. no matter whether the second translate will be applied or not, the result will be success.

mgrudzinska: Also I wouldn't use the same values as the args, so in case something is mixed in the code, we can catch it

test/capi/testPaint.cpp
```diff
@@ -0,0 +1,113 @@
...
+    REQUIRE(tvg_paint_get_opacity(paint, &opacity_get) == TVG_RESULT_SUCCESS);
+    REQUIRE(opacity_get == opacity_get);
+
+    REQUIRE(tvg_paint_del(paint) == TVG_RESULT_SUCCESS);
+}
+
+TEST_CASE("Paint Bounds", "[capiPaintBounds]")
+{
+    Tvg_Paint *paint = NULL;
+    float x = 0, y = 10, w = 100, h = 100;
```
mgrudzinska, line 87: I wouldn't use the same values as the args, so in case something is mixed in the code, we can catch it

test/capi/testPaint.cpp
```diff
@@ -0,0 +1,113 @@
...
+    REQUIRE(tvg_paint_scale(paint, 2.0f) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_paint_scale(paint, 0.5f) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_paint_rotate(paint, 180.0f) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_paint_rotate(paint, 180.0f) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_paint_translate(paint, 10.0f, 10.0f) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_paint_translate(paint, -10.0f, -10.0f) == TVG_RESULT_SUCCESS);
+
+    REQUIRE(tvg_paint_get_transform(paint, &matrix_get) == TVG_RESULT_SUCCESS);
+
+    REQUIRE(matrix_get.e11 == matrix_set.e11);
```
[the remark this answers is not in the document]
mmaciola: Fixed transform, translate, scale and rotate tests

test/capi/testPaint.cpp
mmaciola: Changed `w` value

test/capi/testPaint.cpp
```diff
@@ -0,0 +1,177 @@
...
+    Tvg_Paint *paint = NULL;
+    Tvg_Matrix matrix_get;
+    float scale = 2.5f;
+
+    paint = tvg_shape_new();
+    REQUIRE(paint != NULL);
+
+    REQUIRE(tvg_paint_scale(paint, scale) == TVG_RESULT_SUCCESS);
+    REQUIRE(tvg_paint_get_transform(paint, &matrix_get) == TVG_RESULT_SUCCESS);
+    REQUIRE(matrix_get.e11 == scale);
```
mgrudzinska, line 84: I'm against comparing floats like ints. 
I know we do that sometimes in tvg, but it's because of performance reasons and even if the comparison won't work as we expect, the result still will be ok. and if the comparison won't work here, the test will not pass.
So I'd use fabs - as you did with a rotation sect

mgrudzinska: I'd change that everywhere you have floats

mmaciola: I am okay with direct comparison of floats there, for scale and translation, as these are directly set (without add or multiply operations) in `RenderTransform::update`.
Nevertheless I understand your point of view, I will change it to abs and eps comparison.

test/capi/testSwCanvas.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include <thorvg_capi.h>
+#include "catch.hpp"
+
+TEST_CASE("Canvas initialization", "[capiCanvas]")
+{
+    static uint32_t width = 200;
+    static uint32_t height = 200;
+
+    uint32_t * buffer = NULL;
```
hermet, line 31: please declare variables on the necessary line.
buffer = (uint32_t*) mlloac(... );

test/capi/testSwCanvas.cpp
```diff
@@ -0,0 +1,51 @@
...
+    Tvg_Canvas* canvas = NULL;
+
+    buffer = (uint32_t*) malloc(sizeof(uint32_t) * width * height);
+    REQUIRE(buffer != NULL);
```
hermet, line 35: REQUIRE(buffer);  keep it simpler.

test/capi/testSwCanvas.cpp
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (c) 2020-2021 Samsung Electronics Co., Ltd. All rights reserved.
```
hermet, line 2: 2021

test/capi/testShape.cpp
```diff
@@ -0,0 +1,237 @@
...
+TEST_CASE("Fill rule", "[capiFillRule]")
+{
+    Tvg_Paint *paint = NULL;
+    Tvg_Fill_Rule rule;
+    Tvg_Fill_Rule rule_get;
+
+    paint = tvg_shape_new();
+    REQUIRE(paint != NULL);
+
+    rule = TVG_FILL_RULE_EVEN_ODD;
```
hermet, line 230: you can use a constant immediately and reduce line.
All other parts as well.

test/capi/testShape.cpp
```diff
@@ -0,0 +1,237 @@
...
+    REQUIRE(a == a_get);
+
+    REQUIRE(tvg_paint_del(paint) == TVG_RESULT_SUCCESS);
+}
+
+TEST_CASE("Stroke dash", "[capiStrokeDash]")
+{
+    Tvg_Paint *paint = NULL;
+    float dashPattern[2] = {20, 10};
+    float * dashPattern_get;
```
hermet, line 141: please keep the coding convention.
float* 
Others as well.

hermet (changes requested):
Please check comments.
You need to carefully see the coding style in thorvg.

mmaciola:
@hermet Please find changes

hermet (approved):

hermet (approved):

PR #419, opened 2021-06-06
svg_loader: initializing uninitialized variables

mgrudzinska:
The 'loader->svgParse->global' variable was uninitialized when no viewBox attribute
was set. If gradient was applied, the division by zero occured and no gradient was drawn.

before:
before

after:
after


JSUYA (approved):
LGTM

PR #420, opened 2021-06-07
tests: add SwCanvas unit tests

hermet:


hermet:
@mgrudzinska come again

mgrudzinska:
@hermet give me sec, I played with the width and stride of the target and I got segf. I need to check whos fault is it - my or tvg :)

mgrudzinska:
@hermet if you are ok with #424 then one more case should be added here.

mgrudzinska:
what should happen in such a case? 
```canvas->mempool((SwCanvas::MempoolPolicy)4)```
now it we be valid call like for Shareable or Default.
btw - I don't see any difference between Shareable  and Default - do we need both of them?

hermet:
> what should happen in such a case?
> `canvas->mempool((SwCanvas::MempoolPolicy)4)`
> now it we be valid call like for Shareable or Default.
> btw - I don't see any difference between Shareable and Default - do we need both of them?

        Default = 0, ///< Default behavior that ThorVG is designed to.
        Shareable,   ///< Memory Pool is shared among the SwCanvases.
        Individual   ///< Allocate designated memory pool that is only used by current instance.

"Default" is used to leave to thorvg basic behavior. It's a blackbox(meaning, its changeable), If user have no idea of this mechanism then they can leave it to default behavior.

If user knows this mempool mechanism and explicitly use the specific one they can use the sharable or individual.

PR #440, opened 2021-06-14
draft: loaders: Async and sync loader

mmaciola:
This patch introduced synchronous loading possibility.

@API Changes:
Result Picture::load(const char* data, uint32_t size, bool async /*=true*/) noexcept

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -986,6 +986,7 @@ public:
 
     /**
      * @brief Loads a picture data from a memory block of a given size.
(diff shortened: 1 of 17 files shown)


hermet:
@mmaciola  i think you can consider an another approach for this like this 
 = > Result load(uint32_t* data, uint32_t w, uint32_t h, bool copy) noexcept;
Copying data instead of sync/async
In my opinion, the copying is better idea to, since async is dominated by engine option, not picture.

PR #449, opened 2021-06-16
svg_loader: adding check if stroke-dasharray attribute != none

mgrudzinska:
The stroke-dasharray value equal to "none" was causing a segfault
during a conversion to float. Fixed now.

sample code:
```
<svg viewBox="0 0 30 10" xmlns="http://www.w3.org/2000/svg">

  <line x1="0" y1="3" x2="30" y2="3" stroke="black"
          stroke-dasharray="none" />
</svg>
```


JSUYA (approved):
LGTM

PR #450, opened 2021-06-16
svg_loader: after reading an unsupported style attribute no other val…

mgrudzinska:
…ues were read

After finding an unsupported style attribute the log is printed (on request)
and processing of other values continues.

code:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
  <defs
     id="defs4">
    <linearGradient
       id="linearGradient1"
       x1="0"
       y1="0"
       x2="0.2"
       y2="0.2"
       spreadMethod="reflect">
      <stop
         style="test:test_value;stop-color:#ffff00;stop-opacity:0.3;"    <!-- PROBLEM -->
         offset="0"/>
      <stop
         style="stop-color:#0000ff;stop-opacity:1;"
         offset="1"/>
    </linearGradient>
  </defs>
  <rect x="0" y="0" width="100" height="100" fill="url(#linearGradient1)"/>
</svg>
```

before (the first color style is not loaded):
before

after:
after

### DIFF
--- a/src/loaders/svg/tvgXmlParser.cpp
+++ b/src/loaders/svg/tvgXmlParser.cpp
@@ -503,7 +503,14 @@ bool simpleXmlParseW3CAttribute(const char* buf, simpleXMLAttributeCb func, cons
         }
 
         if (key[0]) {
-            if (!func((void*)data, key, val)) return false;
+
+#ifdef THORVG_LOG_ENABLED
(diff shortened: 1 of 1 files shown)


mgrudzinska:
there was a typo in a log msg

JSUYA (approved):
LGTM

PR #451, opened 2021-06-16
svg_loader: fixing color parsing

mgrudzinska:
For colors given in a style attribute in the format "rgb(rrr, ggg, bbb)"
a parsing was incorect. Now fixed.

code:
```
<svg height="300" width="300">
  <path d="M 0 0 h 200 v 200 z" style="fill: rgb(255, 155, 55);"/>
</svg>
```

before:
before

after:
after


JSUYA (approved):
LGTM

Issue #40, opened 2020-09-10
Support Radial Gradient Transformation in SVG


JSUYA:
Radial gradient transformation is not yet supported.

hermet:
Come to think of it, thorvg already support Radial Gradient Transfomation.
see => testGradientTransform

Could you elaborate the issue case?

JSUYA:
This is SVG feature support.
GradientTransform for radial gradient in SVG Loader is not implemented yet.

mgrudzinska:
@JSUYA 
we have the following comment in the tvgSwFill.cpp:  "//FIXME; Scale + Rotation is not working properly"
Do you know something more about it? some examples?

I can not find a problem with that:
```
<svg viewBox="0 0 420 200" xmlns="http://www.w3.org/2000/svg">
  <radialGradient id="gradient1" gradientUnits="userSpaceOnUse"
      cx="100" cy="100" r="100" fx="100" fy="100">
    <stop offset="0%" stop-color="darkblue" />
    <stop offset="50%" stop-color="skyblue" />
    <stop offset="100%" stop-color="darkblue" />
  </radialGradient>
  <radialGradient id="gradient2" gradientUnits="userSpaceOnUse"
      cx="100" cy="100" r="100" fx="100" fy="100"
      gradientTransform="scale(2) translate(-50, 0)">
    <stop offset="0%" stop-color="darkblue" />
    <stop offset="50%" stop-color="skyblue" />
    <stop offset="100%" stop-color="darkblue" />
  </radialGradient>

  <rect x="0" y="0" width="200" height="200" fill="url(#gradient1)" />
  <rect x="0" y="0" width="200" height="200" fill="url(#gradient2)" style="transform: translate(220, 0) scale(0.5) rotate(45);" />
</svg>
```

the result from tvg is as expected:
![rad_trnsform](https://user-images.githubusercontent.com/67589014/122482684-52d90180-cfd1-11eb-8f71-4a6582be2fb9.PNG)

mgrudzinska:
@JSUYA  Is that the problem you meant?
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
  <defs
     id="defs4">
    <radialGradient
       id="radialGradient1"
       r="0.2"
       cx="0.3"
       cy="0.3"
       spreadMethod="reflect">
      <stop
         style="stop-color:#ff0000;stop-opacity:1;"
         offset="0"/>
      <stop
[...]

Issue #37, opened 2020-09-10
Support for applying the r value of radial gradient relative to the width and height


JSUYA:
```html
<?xml version="1.0" encoding="UTF-8"?>
<svg width="1000px" height="100px" viewBox="0 0 1000 100" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <!-- Generator: Sketch 58 (84663) - https://sketch.com -->
    <title>ime_bg_input_auto</title>
    <desc>Created with Sketch.</desc>
    <defs>
        <radialGradient cx="50%" cy="50%" r="10%"  id="radialGradient-3">
            <stop stop-color="#0000FF" offset="0%"></stop>
            <stop stop-color="#FF0000" offset="100%"></stop>
        </radialGradient>
    </defs>
    <g id="ime_bg_input_auto" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
        <rect id="Popup/Gradient-Copy-2" fill="url(#radialGradient-3)"  x="0" y="0" width="1000" height="100"></rect>
    </g>
</svg>
```

shape->rect(0,0,1000, 100)
grad->radial(500, 50 , 10);
// -> grad->radial(500, 50 , 0.1);
shape->fill(grad)

[Current]
![image](https://user-images.githubusercontent.com/7413838/92696838-6b9ef180-f385-11ea-83c6-774d5e35189f.png)

[To be]
![image](https://user-images.githubusercontent.com/7413838/92696632-19f66700-f385-11ea-8fdb-25b1bdc39c77.png)

JSUYA:
If this is fixed, further work is needed to make it supported in SVG as well.
https://github.com/Samsung/thorvg/issues/40#issuecomment-864021872

PR #460, opened 2021-06-17
svg_loader: radial gradient transformation support

mgrudzinska:
For now the gradient radius is scales like x-axis - it has to be changed
after issue #37 will be resolved.

issue #40 

code:
```
<svg viewBox="0 0 420 200">
  <radialGradient id="gradient1" gradientUnits="userSpaceOnUse"
      cx="100" cy="100" r="100" fx="100" fy="100"
      gradientTransform="scale(2) translate(-50, 0)">
    <stop offset="0%" stop-color="darkblue" />
    <stop offset="50%" stop-color="skyblue" />
    <stop offset="100%" stop-color="darkblue" />
  </radialGradient>

  <rect x="0" y="0" width="200" height="200" fill="url(#gradient1)" />
</svg>
```

before:
image

after:
image

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -114,8 +114,15 @@ static unique_ptr<RadialGradient> _applyRadialGradientProperty(SvgStyleGradient*
         g->radial->fy = g->radial->fy * rh + ry;
     }
 
-    //TODO: Radial gradient transformation is not yet supported.
-    //if (g->transform) {}
+    //TODO: Radial gradient transformation - all tests possible after rx/ry implementation
+    if (g->transform) {
+        auto cx = g->radial->cx * g->transform->e11 + g->radial->cy * g->transform->e12 + g->transform->e13;
+        g->radial->cy = g->radial->cx * g->transform->e21 + g->radial->cy * g->transform->e22 + g->transform->e23;
+        g->radial->cx = cx;
+
+        auto sx = sqrt(pow(g->transform->e11, 2) + pow(g->transform->e21, 2));
+        g->radial->r *= sx;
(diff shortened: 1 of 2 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -114,8 +114,15 @@ static unique_ptr<RadialGradient> _applyRadialGradientProperty(SvgStyleGradient*
...
 
-    //TODO: Radial gradient transformation is not yet supported.
-    //if (g->transform) {}
+    //TODO: Radial gradient transformation - all tests possible after rx/ry implementation
+    if (g->transform) {
+        auto cx = g->radial->cx * g->transform->e11 + g->radial->cy * g->transform->e12 + g->transform->e13;
+        g->radial->cy = g->radial->cx * g->transform->e21 + g->radial->cy * g->transform->e22 + g->transform->e23;
+        g->radial->cx = cx;
+
+        auto sx = sqrt(pow(g->transform->e11, 2) + pow(g->transform->e21, 2));
```
JSUYA, line 123: Can you explain this calculation?
Why g->transform->e21 ?

JSUYA (commented):
Please check my comment

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: for now it doesn't make any difference, we can use here e21 or e12.
it is because tvg doesn't support radial grad radius with respect to the shape width/height (#37)

after we solve #37 I think we should add one more step so we will be able to determine what was applied first - scaling or rotation, so we can properly find the  sx and sy values.

JSUYA (approved):
LGTM

PR #467, opened 2021-06-22
svg_loader XmlParser: Fix check the end of string

JSUYA:
While skipping the string, check the end of the string first.
Prevents an invalid read when parser really get to the end of the string.

issue: https://github.com/Samsung/thorvg/issues/461


JSUYA:
https://github.com/godotengine/godot/runs/2847035060?check_suite_focus=true#step:12:61

This patch to fix invaild reads found during godotengine's address sanitizer check.
I hope this patch will fix this issue ......

hermet (approved):

Issue #457, opened 2021-06-17
Assorted svg pack with multiple crashes


fire:
https://github.com/godotengine/godot/pull/49601#issuecomment-862334524

There is a zip.

## Crashes (not just these ones)
1. e-stop-019.svg
1. speed.svg
2. e-mask-022.svg
3. e-mask-023.svg
4. e-linearGradient-004.svg

mgrudzinska:
e-linearGradient-004.svg -> solved in #470

mgrudzinska:
e-stop-019.svg -> the final solution in #514

hermet:
i think all five issues are fixed. 
The last improvement patch => https://github.com/Samsung/thorvg/pull/539

fire:
Thanks.

I'll close this and when I test the zip again, I'll make new bug reports.

PR #475, opened 2021-06-22
common fill: handling the ColorStop offset values

mgrudzinska:
The ColorStop offset < 0 and > 1 should be treated as 0 and 1 respectively.
The offset value < than the previous offset value should be replaced
by the previous value - without this change segfault occurred.

specs:
https://svgwg.org/svg2-draft/pservers.html#StopNotes

We need this change to solve issue #457 (example e-stop-019.svg) - but there are also other problems with this svg file.

examples:
```
<svg viewBox="0 0 200 200">
        <stop offset="10" stop-color="blue"/>
        <stop offset="0" stop-color="red"/>
    </linearGradient>
    <rect x="20" y="20" width="160" height="160" fill="url(#grad)"/>
</svg>

<svg viewBox="0 0 200 200">
        <stop offset="0" stop-color="white"/>
        <stop offset=".5" stop-color="red"/>
        <stop offset="15" stop-color="white"/>
        <stop offset="1" stop-color="black"/>
    </linearGradient>
    <rect x="20" y="20" width="160" height="160" fill="url(#grad)"/>
</svg>


<svg viewBox="0 0 200 200">
        <stop offset="0" stop-color="white"/>
        <stop offset=".5" stop-color="red"/>
        <stop offset=".0" stop-color="white"/>
        <stop offset="1" stop-color="black"/>
    </linearGradient>
    <rect x="20" y="20" width="160" height="160" fill="url(#grad)"/>
</svg>


<svg viewBox="0 0 200 200">
        <stop offset="0" stop-color="yellow"/>
        <stop offset="0.5" stop-color="blue"/>
        <stop offset="2" stop-color="green"/>
    </linearGradient>
    <rect x="20" y="20" width="160" height="160" fill="url(#grad)"/>
</svg>

```

before (I prevented segfault to see the results, without it segf occurred):
before

after:
after


JSUYA:
Good point.
However, this problems are the Spec of SVG.
That's why I think this change should be handled by SVG Parser(tvgSvgLoader).
Of course we think tvgFill can conform to the SVG Spec, but we haven't agreed on that yet.

hermet:
@mgrudzinska  @JSUYA that is obvious wrong input. Could you please take it over in svg loader side?

src/lib/tvgFill.cpp
```diff
@@ -61,6 +61,13 @@ Result Fill::colorStops(const ColorStop* colorStops, uint32_t cnt) noexcept
     pImpl->cnt = cnt;
     memcpy(pImpl->colorStops, colorStops, cnt * sizeof(ColorStop));
 
+    if (pImpl->colorStops[0].offset < 0) pImpl->colorStops[0].offset = 0;
```
hermet, line 64: I'd like to suggest to add a comment here for this exception handling.
example:
//Correct offset range .... xxx

hermet (changes requested):

mgrudzinska:
@hermet - I'll handle this on the svg loader side, but I think this check should be done also in the tvgFill, so that the input from apis is checked as well.  are you ok with that?

hermet:
> @hermet - I'll handle this on the svg loader side, but I think this check should be done also in the tvgFill, so that the input from apis is checked as well. are you ok with that?

No, if so, I won't ask you to put in svg loader side. I think user should handle it by himself.

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -254,6 +254,7 @@ struct SvgStyleGradient
     SvgLinearGradient* linear;
     Matrix* transform;
     Array<Fill::ColorStop *> stops;
+    float prevOffset;
```
mgrudzinska, line 257: @JSUYA I added this so I don't need to call the ```colorStops``` getter every time (we need to know the previous offset value before we add a new 'stop').

mgrudzinska:
@hermet  @JSUYA 
I changed this patch as you wished, but I don't think it is a good idea to handle this only on the svg loader side (of course to handle this in the loader and tvgFill is also a bad idea, that's why I decided to do it only in the tvgFill). 

A user can call apis like this:
```
    colorStops[1] = {10, 255, 255, 255, 255};     // <- offset = 10
```
and it is perfectly good input, but the results from tvg will be different than it should be.
according to the specs:
"Gradient offset values less than 0 (or less than 0%) are rounded up to 0%. Gradient offset values greater than 1 (or greater than 100%) are rounded down to 100%."
so I think a user can expect, that the result will be the same as for the following calls:
```
    colorStops[1] = {1, 255, 255, 255, 255};     // <- offset = 1
```

long story short - I made the patch as you wanted.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2213,7 +2216,12 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content,
         stop->a = 255;
         simpleXmlParseAttributes(attrs, attrsLength, _attrParseStops, loader);
         if (loader->latestGradient) {
+            // check the offset corner cases - refer to: https://svgwg.org/svg2-draft/pservers.html#StopNotes
+            if (stop->offset < loader->latestGradient->prevOffset) stop->offset = loader->latestGradient->prevOffset;
+            else if (stop->offset > 1) stop->offset = 1;
```
JSUYA, line 2221: I don't think it's a good idea to modify the parsed data.
There are some case what parsed data is modified. But basically I think we should keep the parsed data.
When we create a tvg object using the parsed data, we can implement it according to the spec we think.

I mean...
You can apply your idea when creating colorStops of fill->Grad in In _applyLinearGradientProperty().

like this...?
```c
71     //Update the stops     
72     stopCount = g->stops.count;
73     if (stopCount > 0) {       
74         stops = (Fill::ColorStop*)calloc(stopCount, sizeof(Fill::ColorStop));
75         if (!stops) return fillGrad;      
76         float prevOffset = 0.0;           
77         for (uint32_t i = 0; i < g->stops.count; ++i) {
78             auto colorStop = g->stops.data[i];         
79             //Use premultiplied color
80             stops[i].r = colorStop->r;                            
81             stops[i].g = colorStop->g;
82             stops[i].b = colorStop->b;
83             stops[i].a = (colorStop->a * fillOpacity) / 255.0f;
84             stops[i].offset = colorStop->offset; 
85             if (colorStop->offset < prevOffset) stops[i].offset = prevOffset;
86             else if (colorStop->offset > 1) stops[i].offset = 1;
87             prevOffset = colorStop->offset;     
88         }
89         fillGrad->colorStops(stops, stopCount);
90         free(stops); 

```

mgrudzinska: yes, you've told me already once not to change the parsed data, I forgot. I'll remember in the future :)

JSUYA (approved):
LGTM

PR #476, opened 2021-06-22
svg_loader: correct parsing ColorStop offset values

mgrudzinska:
Values different from numbers and percentages should be ignored
and the default values should be applied (zeros).

We need this change to solve issue #457 (example e-stop-019.svg) - but there are also other problems with this svg file.

example:
```
<svg viewBox="0 0 200 200">
        <stop offset="10%" stop-color="white"/>
        <stop offset="20% " stop-color="red"/>
        <stop offset="30% k" stop-color="blue"/>
        <stop offset="40%" stop-color="yellow"/>
        <stop offset="0.5m" stop-color="red"/>
        <stop offset="0.6 " stop-color="green"/>
        <stop offset="70%m" stop-color="black"/>
        <stop offset="80%" stop-color="white"/>
    </linearGradient>
    <rect x="20" y="20" width="160" height="160" fill="url(#grad)"/>
</svg>          
```

The results after applying #475 
before:
before2

after:
after2

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -141,10 +141,17 @@ static float _gradientToFloat(const SvgParser* svgParse, const char* str, SvgPar
 static float _toOffset(const char* str)
 {
     char* end = nullptr;
+    auto strEnd = str + strlen(str);
 
     float parsedValue = svgUtilStrtof(str, &end);
 
-    if (strstr(str, "%")) parsedValue = parsedValue / 100.0;
+    end = _skipSpace(end, nullptr);
+    auto ptr = strstr(str, "%");
+
+    if (ptr) {
+        parsedValue = parsedValue / 100.0;
+        if (end != ptr || (end + 1) != strEnd) return 0;
+    } else if (end != strEnd) return 0;
 
     return parsedValue;
 }


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -141,10 +141,17 @@ static float _gradientToFloat(const SvgParser* svgParse, const char* str, SvgPar
...
 
     float parsedValue = svgUtilStrtof(str, &end);
 
-    if (strstr(str, "%")) parsedValue = parsedValue / 100.0;
+    end = _skipSpace(end, nullptr);
+    auto ptr = strstr(str, "%");
+
+    if (ptr) {
+        parsedValue = parsedValue / 100.0;
+        if (end != ptr || (end + 1) != strEnd) return 0;
```
mgrudzinska, line 153: ```end != ptr ```   ->   nothing can be before the % sign 
```(end + 1) != strEnd```  ->   nothing can be after the % sign (all spaces from the end were cleared earlier)

File: src/loaders/svg/tvgSvgLoader.cpp [hunk 2]
```diff
@@ -141,10 +141,17 @@ static float _gradientToFloat(const SvgParser* svgParse, const char* str, SvgPar
...
+    } else if (end != strEnd) return 0;
```
mgrudzinska, line 154: ```end != strEnd```   ->   in case no % sign was used, we need to check whether there is nothing left after the number

JSUYA (approved):
LGTM

PR #478, opened 2021-06-23
infra: add memory sanitize option in meson

hermet:
for optimial library, we removed some peripheral information in default,
this breaks asan environment.

This patch changes it to make it both.

if you want to use asan, please change the meson option.

default is disabled.

example)
$meson . build -Db_sanitize=address
$meson . build -Db_sanitize=address,undefined


hermet:
This resolves this https://github.com/Samsung/thorvg/issues/473

JSUYA (approved):
LGTM

PR #479, opened 2021-06-23
svg_loader SvgPath: Added corner case handling for flags of Arc

JSUYA:
It handles the case what the flag of the arc of the path is defined
without spaces or commas, such as 00, 01, 11 or 10.

issue: https://github.com/Samsung/thorvg/issues/455

### DIFF
--- a/src/loaders/svg/tvgSvgPath.cpp
+++ b/src/loaders/svg/tvgSvgPath.cpp
@@ -54,15 +54,14 @@ static bool _parseNumber(char** content, float* number)
 
 static bool _parseFlag(char** content, int* number)
 {
-    char* end = NULL;
-    *number = strtol(*content, &end, 10);
-    //If the start of string is not number or a number was a float
-    if ((*content) == end || *end == '.') return false;
-    //If a flag has a different value than 0 or 1
-    if (*number != 0 && *number != 1) return false;
-    *content = _skipComma(end);
+    if (*(*content) != '0' && *(*content) != '1') return false;
+    *number = *(*content) - '0';
+    *content = _skipComma(*content + 1);
+    if (*(*content) == '.') return false;
+
     return true;
 }
+
 
 void _pathAppendArcTo(Array<PathCommand>* cmds, Array<Point>* pts, Point* cur, Point* curCtl, float x, float y, float rx, float ry, float angle, bool largeArc, bool sweep)
 {


src/loaders/svg/tvgSvgPath.cpp
```diff
@@ -55,12 +55,22 @@ static bool _parseNumber(char** content, float* number)
 static bool _parseFlag(char** content, int* number)
 {
     char* end = NULL;
-    *number = strtol(*content, &end, 10);
-    //If the start of string is not number or a number was a float
-    if ((*content) == end || *end == '.') return false;
-    //If a flag has a different value than 0 or 1
-    if (*number != 0 && *number != 1) return false;
-    *content = _skipComma(end);
+    //NOTE: It handles the case what the flag of the arc of the path is defined
```
hermet, line 58: Itself is NOTE. do we really need to emphasize "I'm a note ?"

Suggest format:

/* It handles the case what the flag of the arc of the path is defined
    without spaces or commas, such as 00, 01, 11 or 10. */

JSUYA: I updated patch

hermet:
@mgrudzinska I'm leaving it to your review.

src/loaders/svg/tvgSvgPath.cpp
```diff
@@ -55,12 +55,22 @@ static bool _parseNumber(char** content, float* number)
...
+        *content = _skipComma(end);
+    }
+    else {
+        *number = strtol(*content, &end, 10);
+        //If the start of string is not number or a number was a float
+        if ((*content) == end || *end == '.') return false;
+        //If a flag has a different value than 0 or 1
+        if (*number != 0 && *number != 1) return false;
+        *content = _skipComma(end);
+    }
```
mgrudzinska, line 73: In my opinion if any of these two flags is something else than 0 or 1 it should be treated as an error.  I think that would be in agree with the specs. 
The thing is that in some online svg viewers (as w3school) it's not exactly like that. so the flags: ```0.0 1``` are considered to be wrong, but ```0 1.0``` are ok. So I'm a bit confused.

I think we should follow the specs, so I'd write this part like that:
```
static bool _parseFlag(char** content, int* number)
{
    if (*(*content) != '0' && *(*content) != '1') return false;
    *number = *(*content) - '0';
    *content = _skipComma(*content + 1);
    if (*(*content) == '.') return false;

    return true;
}
```
1. ```if (*(*content) != '0' && *(*content) != '1') return false;```  -> if there is something else than '0' or '1' false should be returned
2. ```if (*(*content) == '.') return false;```  -> if after the flag is a point it means that it was a float, so it's an error. In case it was some other NAN, it'll be handled in the next step, when the next value is loaded

What do you think? 
If you want to be in agree with w3school viewer (well, I don't know the algorithm they are using, I can check some cases to compare results), we can stay with you solution, but than I think we can remove the whole 'else' part and we don't really need the 'end' variable

JSUYA: Thank you for your comment.
I'm a little concerned about corner case.
But I think the code you suggested is simpler and better.
I updated patch.

hermet:
@JSUYA this patch has been reverted see: https://github.com/Samsung/thorvg/issues/555

JSUYA:
> @JSUYA this patch has been reverted see: #555

https://github.com/Samsung/thorvg/pull/562

PR #497, opened 2021-06-23
common: fix MSVC compatibility problem.

hermet:
clz (count leading zero) is builtin function
that is not available in MSVC.

This patch replace clz for MSVC env.

Thanks @fire for pointing out this.

Refers: https://stackoverflow.com/questions/355967/how-to-use-msvc-intrinsics-to-get-the-equivalent-of-this-gcc-code

### DIFF
--- a/src/lib/sw_engine/tvgSwMath.cpp
+++ b/src/lib/sw_engine/tvgSwMath.cpp
@@ -27,6 +27,21 @@
 /* Internal Class Implementation                                        */
 /************************************************************************/
 
+//clz: count leading zero’s
+#ifdef _MSC_VER
+    #include <intrin.h>
+    static uint32_t __inline _clz(uint32_t value)
+    {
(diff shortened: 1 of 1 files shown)


fire:
`thirdparty\thorvg\src\lib\sw_engine\tvgSwMath.cpp` has an extra `}`

src/lib/sw_engine/tvgSwMath.cpp
```diff
@@ -27,6 +27,21 @@
...
+//clz: count leading zero’s
+#ifdef _MSC_VER
+    #include <intrin.h>
+    static uint32_t __inline _clz(uint32_t value)
+    {
+        unsigned long leadingZero = 0;
+        if (_BitScanReverse(&leadingZero, value)) return 31 - leadingZero;
+        else return 32;
+    }
+}
```
fire, line 39: Location.

PR #503, opened 2021-06-24
svg2png: Clean up code

JSUYA:
Modify the code to fit coding style

### DIFF
--- a/src/bin/svg2png/svg2png.cpp
+++ b/src/bin/svg2png/svg2png.cpp
@@ -29,25 +29,25 @@
 using namespace std;
 
 
-struct PngBuilder {
-
+struct PngBuilder
+{
     void build(const std::string &fileName , const uint32_t width, const uint32_t height, uint32_t *buffer)
     {
         std::vector<unsigned char> image;
         image.resize(width * height * 4);
-        for(unsigned y = 0; y < height; y++)  {
-            for(unsigned x = 0; x < width; x++) {
+        for (unsigned y = 0; y < height; y++) {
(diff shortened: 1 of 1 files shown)


src/bin/svg2png/svg2png.cpp
```diff
@@ -30,30 +30,28 @@ using namespace std;
...
-        if(error) std::cout << "encoder error " << error << ": "<< lodepng_error_text(error) << std::endl;
+        if (error) std::cout << "encoder error " << error << ": "<< lodepng_error_text(error) << std::endl;
     }
 };
 
 
-struct App
-{
+struct App {
     void tvgRender(int w, int h)
```
hermet, line 55: origin code is correct in thorvg...

src/bin/svg2png/svg2png.cpp
```diff
@@ -160,7 +157,7 @@ struct App
         return 0;
     }
 
-private:
+ private:
     std::string basename(const std::string &str)
```
hermet, line 161: origin code is correct in thorvg...

hermet (changes requested):

src/bin/svg2png/svg2png.cpp
JSUYA: i updated patch

src/bin/svg2png/svg2png.cpp
JSUYA: i updated patch

hermet (approved):

PR #505, opened 2021-06-24
svg_loader: fix double-free memory.

hermet:
copy attribute must copy the url data,
otherwise, url memory can be atempted freeing twice.

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -879,7 +879,11 @@ static void _handleMaskAttr(TVG_UNUSED SvgLoaderData* loader, SvgNode* node, con
 #endif
     style->comp.method = CompositeMethod::AlphaMask;
     int len = strlen(value);
-    if (len >= 3 && !strncmp(value, "url", 3)) style->comp.url = _idFromUrl((const char*)(value + 3));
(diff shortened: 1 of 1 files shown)


mgrudzinska (approved):

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1596,6 +1600,7 @@ static void _copyAttr(SvgNode* to, const SvgNode* from)
     }
     //Copy style attribute;
     memcpy(to->style, from->style, sizeof(SvgStyleProperty));
+    if (from->style->fill.paint.url) to->style->fill.paint.url = new string(from->style->fill.paint.url->c_str());
```
JSUYA, line 1603: I think, a case for stroke should also be added.
```suggestion
    if (from->style->fill.paint.url) to->style->fill.paint.url = new string(from->style->fill.paint.url->c_str());
    if (from->style->stroke.paint.url) to->style->stroke.paint.url = new string(from->style->stroke.paint.url->c_str());
```

mgrudzinska: good point... I missed it

hermet: updated.

PR #510, opened 2021-06-24
tvg_loader: change values of flags

mgrudzinska:
It's a middle stage leading to removing flags from the tvgBinaryDesc.h file
and use the corresponding enums from the thorvg.h file.

In the saver module I'm using the enums and in the tvg loader the flags are used.
With this change the enums and flags are compatible. 
In the future I'd like to use enums in both cases.

### DIFF
--- a/src/lib/tvgBinaryDesc.h
+++ b/src/lib/tvgBinaryDesc.h
@@ -80,18 +80,18 @@ struct tvgBlock
 #define TVG_SHAPE_COLOR_INDICATOR   (TvgIndicator)0x43 // Shape has color
 
 #define TVG_SHAPE_FILLRULE_INDICATOR    (TvgIndicator)0x44 // Shape FillRule
-#define TVG_SHAPE_FILLRULE_WINDING_FLAG      (TvgFlag)0x01 // FillRule::Winding
-#define TVG_SHAPE_FILLRULE_EVENODD_FLAG      (TvgFlag)0x02 // FillRule::EvenOdd
+#define TVG_SHAPE_FILLRULE_WINDING_FLAG      (TvgFlag)0x00 // FillRule::Winding
+#define TVG_SHAPE_FILLRULE_EVENODD_FLAG      (TvgFlag)0x01 // FillRule::EvenOdd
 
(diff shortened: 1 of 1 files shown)


mgrudzinska:
@mmaciola  Could you please explain a bit more precise why you don't want to use enums from the thorvg.h?

mmaciola:
As we talked about before, designing the .tvg file, there were three major advantages using defined flags over enum from thorvg.h:
Using defined flags makes loading independent from changes in thorvg.h and makes .tvg clearly defined (compatibility if new value is added or enum is reordered).
I assumed that zero char '\0' would be the most common value in .tvg file, so I used flags enumaration starting from one to recognize an error if zero is found.
For loader, casting of value from tvg to enum requires range checking that may cause some hidden problems when new enum values added. Switch-case is easy readable.

Nevertheless, I am open to using enum values if code compactness is preferable over above @hermet

hermet:
@mmaciola I'm agree with @mmaciola about starting value from 1 not 0 and we don't need to discuss changing enum here. It's an another topic.

hermet (changes requested):
please check comments.

mgrudzinska:
@hermet I didn't want to change the values to zero and that's it. I wanted to remove them at all. this patch was just a middle stage to make the saver and the loader compatible.

just to be sure we understand each other:
I'll change parts of code like this:
```
flag = static_cast<TvgFlag>(shape->strokeJoin());
strokeDataByteCnt += writeMember(TVG_SHAPE_STROKE_JOIN_INDICATOR, TVG_FLAG_SIZE, &flag);
```
to look like this:
```
        switch (shape->strokeJoin()) {
            case StrokeJoin::Bevel: {
                flag= TVG_SHAPE_STROKE_JOIN_BEVEL_FLAG;
                break;
            }
            case StrokeJoin::Round: {
                flag= TVG_SHAPE_STROKE_JOIN_ROUND_FLAG;
                break;
            }
            case StrokeJoin::Miter: {
                flag= TVG_SHAPE_STROKE_JOIN_MITER_FLAG;
                break;
            }
        }
        strokeDataByteCnt += writeMember(TVG_SHAPE_STROKE_JOIN_INDICATOR, TVG_FLAG_SIZE, &flag);
```

hermet (approved):

PR #403, opened 2021-05-31
common paint: introducing basic iterators functionality

mgrudzinska:
The introduced Iterator class enables to access the children nodes of a given Paint.

PR #511, opened 2021-06-25
Mgrudzinska/saver iterator based

mgrudzinska:
Introducing the Saves class, that enables to save the paints in a binary format.

I'll change the description of the last commit.
It contains: #403, #508, #509 and #510.


src/lib/tvgSaverImpl.h
```diff
@@ -0,0 +1,494 @@
...
+#include "tvgPaint.h"
+#include "tvgBinaryDesc.h"
+#include <float.h>
+#include <math.h>
+#include <fstream>
+
+struct Saver::Impl
+{
+    Saver* saver;
+    char* buffer = nullptr;
```
mgrudzinska, line 34: @hermet - you asked me in a different pr to change the name buffer -> data.
Do you still want me to do this? I'd rather stay with the 'buffer' name. If I'll use 'data' I probably should change also the functions names.

mgrudzinska:
I changed the commit description

src/lib/tvgSaverImpl.h
hermet: @mgrudzinska if you want to keep it buffer, just leave it. we can rename it with sequential refactoring patches. I asked you to change the name for the sake of consistency among the loaders.  (though it's a sort of preference issue. no correct answer)

hermet (approved):

PR #512, opened 2021-06-25
tvg_saver: example

mgrudzinska:
the expected result:
tvg_saver


mgrudzinska:
I removed white spaces

hermet (changes requested):
very poor commit message, and please atomic patch as possible.

mgrudzinska:
I've removed changes from meson.build and changed the msg. please check now

hermet (approved):

Issue #430, opened 2021-06-10
SVG Gradient Issue


hermet:
"intertwingly.svg" in the thorvg examples doesn't show the gradient fill correctly. Need to investigate it.

mgrudzinska:
I solved it by removing the whole part with usePercentage from the grad, but I want to double check it, since now tvg supports only `userSpaceOnUse`. the other value could be `objectBoundingBox` and it has to be added (that's why clipping in the #429 doesn't work). So some recalculations, like with the usePercentage, has to be done and that's why I don't want to push now a patch where I removed them.

mgrudzinska:
it's partially solved in #514 - the rest of the differences in because tvg does not support fx and fy attributes

hermet:
look good enough.

PR #514, opened 2021-06-25
svg_loader: recalculating the default values of grads

mgrudzinska:
The default values of gradient properties were handled differently
than the values given by a user/in an svg file.

before:
1before

after:
1after

this helps in #430 and solves e-stop-019.svg from #457

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1091,8 +1091,14 @@ static SvgNode* _createSvgNode(SvgLoaderData* loader, SvgNode* parent, const cha
     doc->preserveAspect = true;
     simpleXmlParseAttributes(buf, bufLength, _attrParseSvgNode, loader);
 
-    if (loader->svgParse->global.w == 0) loader->svgParse->global.w = loader->svgParse->node->node.doc.w;
-    if (loader->svgParse->global.h == 0) loader->svgParse->global.h = loader->svgParse->node->node.doc.h;
+    if (loader->svgParse->global.w == 0) {
+        if (doc->w < FLT_EPSILON) loader->svgParse->global.w = 1;
+        else loader->svgParse->global.w = doc->w;
+    }
+    if (loader->svgParse->global.h == 0) {
+        if (doc->h < FLT_EPSILON) loader->svgParse->global.h = 1;
+        else loader->svgParse->global.h =doc->h;
+    }
 
     return loader->svgParse->node;
 }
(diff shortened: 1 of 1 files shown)


mgrudzinska:
@JSUYA I was wondering about the ``global`` variable. I think it's connected with nesting in the svg file - am I right? 
For now it looks a bit useless, but I didn't want to change the whole logic, since we may need it at some point.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1913,13 +1913,13 @@ static SvgStyleGradient* _createRadialGradient(SvgLoaderData* loader, const char
...
     /**
-    * Default values of gradient
+    * Default values of gradient transformed into global percentage
     */
-    grad->radial->cx = 0.5;
-    grad->radial->cy = 0.5;
-    grad->radial->fx = 0.5;
-    grad->radial->fy = 0.5;
-    grad->radial->r = 0.5;
+    grad->radial->cx = 0.5 / loader->svgParse->global.w;
```
hermet, line 1918: 1. Here it verifies w/h is greater than 0?

2. please append *f  for the float types. => 0.5f;

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1913,13 +1913,13 @@ static SvgStyleGradient* _createRadialGradient(SvgLoaderData* loader, const char
...
+    grad->radial->cy = 0.5 / loader->svgParse->global.h;
+    grad->radial->fx = 0.5 / loader->svgParse->global.w;
+    grad->radial->fy = 0.5 / loader->svgParse->global.h;
+    grad->radial->r = 0.5 / (sqrt(pow(loader->svgParse->global.h, 2) + pow(loader->svgParse->global.w, 2)) / sqrt(2.0));
```
hermet, line 1922: 2.0f

hermet (commented):
some minor nitpicks.

hermet:
leaving to @JSUYA verification.

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: it may be zero and then it's a problem for defaults but also for the values given by the user (there is also division by 0).
I'll think about this 'global' variable...

mgrudzinska:
@JSUYA I'd like to take care about ```global``` in a separate pr.
It's a different problem how to deal with zeros in a ```viewBox``` attrib. After solving it, it will be guarantied that global is != 0.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1913,13 +1913,13 @@ static SvgStyleGradient* _createRadialGradient(SvgLoaderData* loader, const char
...
     */
-    grad->radial->cx = 0.5;
-    grad->radial->cy = 0.5;
-    grad->radial->fx = 0.5;
-    grad->radial->fy = 0.5;
-    grad->radial->r = 0.5;
+    grad->radial->cx = 0.5f / loader->svgParse->global.w;
+    grad->radial->cy = 0.5f / loader->svgParse->global.h;
+    grad->radial->fx = 0.5f / loader->svgParse->global.w;
+    grad->radial->fy = 0.5f / loader->svgParse->global.h;
```
JSUYA, line 1921: If the viewbox and width and height are set to 0 in the svg file, global.w and h can be 0.
It's a wrong svg file but we need to make sure SvgLoader doesn't crash.
Even if you make a patch for global.w and h later, I think we need the logic to check zero now.

how about this?
```cpp
auto w = loader->svgParse->global.w  != 0 ? loader->svgParse->global.w : 1;
...
```

mgrudzinska: @JSUYA please give me some time to think about the best solution

hermet: Return immediately in the beginning of validation header. (if size is invalid)

mgrudzinska: it's just a temporary solution. For now I made sure, that the global variables are not zeros.
I don't want to add the check as @JSUYA suggested, since it should be done earlier, when the svg tag appears. The coordinate system is related to the attributes, that are present in this tag. And the global variables representing this coordinate system.

hermet: @mgrudzinska  I mentioned we can handle it in the header validation (of Svg Loader), It's an exceptional handling for safer program. Anyway no one reject this patch, and that topic is another one.

hermet (approved):

PR #516, opened 2021-06-25
svg_loader: negative attrs length fix

mmaciola:
After finding no attributes but white spaces, attrsLength could be negative
that resulted in segmentation fault in simpleXmlParseAttributes function.

Minimal svg that caused segmentation fault before this patch applied:
```
<?xml version="1.0" encoding="UTF-8"?>
<svg><g ></g></svg>
```

@Issue: https://github.com/Samsung/thorvg/issues/487

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2175,11 +2175,11 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content, 
     if (attrs) {
         //Find out the tag name starting from content till sz length
         sz = attrs - content;
-        attrsLength = length - sz;
         while ((sz > 0) && (isspace(content[sz - 1]))) sz--;
         if ((unsigned)sz >= sizeof(tagName)) return;
         strncpy(tagName, content, sz);
         tagName[sz] = '\0';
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2175,11 +2175,11 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content,
     if (attrs) {
         //Find out the tag name starting from content till sz length
         sz = attrs - content;
-        attrsLength = length - sz;
         while ((sz > 0) && (isspace(content[sz - 1]))) sz--;
         if ((unsigned)sz >= sizeof(tagName)) return;
         strncpy(tagName, content, sz);
         tagName[sz] = '\0';
+        attrsLength = length - sz;
     }
```
hermet, line 2184: @mmaciola  Looks it needs to return immediately if attrsLength is 0?

mmaciola: I am not sure about returning. The tag is ok and proper node should be created but no attrs parsing is needed

mmaciola:
Added similar change to `_svgLoaderParserForValidCheckXmlOpen`, fixing segmentation fault for `<svg ></svg>`

PR #520, opened 2021-06-26
sw_engine: removing a double application of the opacity to a grad

mgrudzinska:
The opacity is already applied and the fill-opacity and stroke-opacity
should be applied on the svg loader side - a separate PR.

code:
```
<svg height="400" width="400" viewBox="0 0 400 400" >
<defs>
<radialGradient id="grad" gradientUnits="userSpaceOnUse" cx="125" cy="125" r="25">
<stop offset="0.000" style="stop-color: rgb(255, 255, 255); stop-opacity: 1;"/>
<stop offset="1.000" style="stop-color: rgb(255, 255, 255); stop-opacity: 1;"/>
</radialGradient>
</defs>
<path d="M 0 0 h 400 v 400 h -400 z" fill="red" />
<path d="M 0,0 h 200 v 200 h -200 z" style="opacity: 0.5; fill: url(#grad);"/>
<path d="M 200,200 h 200 v 200 h -200 z" style="opacity: 0.5; fill: rgb(255, 255, 255);"/>
</svg>
```

before:
3before

after:
3after

### DIFF
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -313,8 +313,8 @@ void shapeResetStroke(SwShape* shape, const Shape* sdata, const Matrix* transfor
 bool shapeGenStrokeRle(SwShape* shape, const Shape* sdata, const Matrix* transform, const SwBBox& clipRegion, SwBBox& renderRegion, SwMpool* mpool, unsigned tid);
 void shapeFree(SwShape* shape);
 void shapeDelStroke(SwShape* shape);
-bool shapeGenFillColors(SwShape* shape, const Fill* fill, const Matrix* transform, SwSurface* surface, uint32_t opacity, bool ctable);
-bool shapeGenStrokeFillColors(SwShape* shape, const Fill* fill, const Matrix* transform, SwSurface* surface, uint32_t opacity, bool ctable);
+bool shapeGenFillColors(SwShape* shape, const Fill* fill, const Matrix* transform, SwSurface* surface, bool ctable);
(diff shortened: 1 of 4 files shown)


hermet:
@mgrudzinska Just skimmed over, this seems to bring a breakage or regression bugs since you are changing the rendering concept.  Actually, rendering engine shouldn't have any depedency to svg. If you'd like to point out the bugs or wrong behavior of rendering engine, I'd like to suggest you to show the wrong result from the correct tvg api usage, not svg concept.

mgrudzinska:
@hermet below the same example using apis - top left and bottom right squares should have the same color:

```
    shape->appendRect(0, 0, 800, 800, 0, 0);
    shape->fill(255, 0, 0, 255);
    if (canvas->push(move(shape)) != tvg::Result::Success) return;
    

    shape1->appendRect(0, 0, 400, 400, 0, 0);
    
    fill->linear(0, 0, 400, 400);
    colorStops[0] = {0, 0, 0, 255, 255};
    colorStops[1] = {1, 0, 0, 255, 255};
    
    shape1->fill(move(fill));
    shape1->opacity(100);
    if (canvas->push(move(shape1)) != tvg::Result::Success) return;
    

    shape2->appendRect(400, 400, 400, 400, 0, 0);
    shape2->fill(0, 0, 255, 255);
    shape2->opacity(100);
    if (canvas->push(move(shape2)) != tvg::Result::Success) return;
```

current result:
6before

hermet:
@mgrudzinska  here is the solution => https://github.com/Samsung/thorvg/pull/526

PR #522, opened 2021-06-27
svg_loader: currentColor - inheritance from a parent

mgrudzinska:
The 'currentColor' attribute value was not inherited. Fixed.

`curColorSet ` is used to distinguish between the two cases from the example below:

code:
```
<svg id="svg1" viewBox="0 0 200 200">
    <g color="green">
        <rect id="rect1" x="0" y="0" width="100" height="200"
              fill="currentColor" color="red"/>
    </g>
    <g color="green">
        <rect id="rect1" x="100" y="0" width="100" height="200"
              fill="currentColor"/>
    </g>
</svg>
```

before:
4before

after:
4after

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -780,6 +780,7 @@ static void _handlePaintAttr(SvgPaint* paint, const char* value)
 static void _handleColorAttr(TVG_UNUSED SvgLoaderData* loader, SvgNode* node, const char* value)
 {
     SvgStyleProperty* style = node->style;
+    style->curColorSet = true;
     _toColor(value, &style->r, &style->g, &style->b, nullptr);
 }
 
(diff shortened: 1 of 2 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -780,6 +780,7 @@ static void _handlePaintAttr(SvgPaint* paint, const char* value)
 static void _handleColorAttr(TVG_UNUSED SvgLoaderData* loader, SvgNode* node, const char* value)
 {
     SvgStyleProperty* style = node->style;
+    style->curColorSet = true;
```
JSUYA, line 783: I think `curColorSet` should be initialize `false` in `_createNode()`.
+) And the curColor variable is also missing initialization. I recommend to initialize them together.

mgrudzinska: Thx! 
I remembered about that, until I forgot... :D

JSUYA (approved):
LGTM

PR #526, opened 2021-06-29
sw_engine renderer: fix a broken composition bug

hermet:
tvg sw_engine tries to skip composition as far as its doable,
it tries composition only if the translucent shape with stroking is there.

There has been a condition bug that composition is applied unnecessarily
even if stroking is disabled. This resulted in broken blending
at gradient filling since it applied alpha values multiple times.

Issues:
https://github.com/Samsung/thorvg/issues/445


hermet:
https://github.com/Samsung/thorvg/pull/520

PR #528, opened 2021-06-29
svg_loader: allow multiple <defs> tags without data and memory leak

mmaciola:
If the svg file contained multiple defs tags, each subsequent tag
overwritten the previous one. This resulted in incorrect rendering of the
file and memory leaks.

Minimal svg that resulted in a memory leak:
```
<svg>
<defs></defs>
<defs></defs>
</svg>
```

Example svg file:
```
<defs id="defs01">
	<linearGradient id="Gradient01"
       x1="0" y1="0" x2="1" y2="1">
		<stop offset="10%" stop-color="#000" />
		<stop offset="90%" stop-color="#fff" />
	</linearGradient>
</defs>
<defs id="defs02">
	<linearGradient id="Gradient02"
       x1="0" y1="0" x2="1" y2="1">
		<stop offset="0" stop-color="#00bacc" />
		<stop offset="1" stop-color="#ccba00" />
	</linearGradient>
</defs>
<rect x="0" y="0" width="200" height="100" fill="url(#Gradient01)" />
<rect x="0" y="100" width="200" height="100" fill="url(#Gradient02)" />
</svg>
```

Before (104 bytes definitely lost and 328 bytes indirectly lost):
Screenshot before

After (no memory leak):
Screenshot after

@Issues: https://github.com/Samsung/thorvg/issues/491

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2200,9 +2200,14 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content, 
         }
 
         if (node->type == SvgNodeType::Defs) {
-            loader->doc->node.doc.defs = node;
-            loader->def = node;
-            if (!empty) loader->stack.push(node);
+            if (loader->def && loader->doc->node.doc.defs) {
+                free(node->style);
+                free(node);
+            } else {
+                loader->doc->node.doc.defs = node;
+                loader->def = node;
+                if (!empty) loader->stack.push(node);
+            }
         } else {
             loader->stack.push(node);
         }
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1068,9 +1068,8 @@ static SvgNode* _createNode(SvgNode* parent, SvgNodeType type)
 
 static SvgNode* _createDefsNode(TVG_UNUSED SvgLoaderData* loader, TVG_UNUSED SvgNode* parent, const char* buf, unsigned bufLength)
 {
+    if (loader->def && loader->doc->node.doc.defs) return nullptr;
     SvgNode* node = _createNode(nullptr, SvgNodeType::Defs);
-    if (!node) return nullptr;
-    simpleXmlParseAttributes(buf, bufLength, nullptr, node);
     return node;
 }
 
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2200,9 +2200,14 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content,
         }
 
         if (node->type == SvgNodeType::Defs) {
-            loader->doc->node.doc.defs = node;
-            loader->def = node;
-            if (!empty) loader->stack.push(node);
+            if (loader->def && loader->doc->node.doc.defs) {
+                free(node->style);
```
JSUYA, line 2204: Good point :)
but.. this node is pushed to parent in createNode. (probably parent is doc)
I don't think it's good to free nodes that have already been pushed into parent's child array.

I think it's good to check if defs is already declared when doing createDefsNode().

How about this?
```cpp
@@ -1066,8 +1066,9 @@ static SvgNode* _createNode(SvgNode* parent, SvgNodeType type)
 }
 
 
-static SvgNode* _createDefsNode(TVG_UNUSED SvgLoaderData* loader, TVG_UNUSED SvgNode* parent, const char* buf, unsigned bufLength)
+static SvgNode* _createDefsNode(SvgLoaderData* loader, TVG_UNUSED SvgNode* parent, const char* buf, unsigned bufLength)
 {
+    if (loader->def && loader->doc->node.doc.defs) return nullptr;
     SvgNode* node = _createNode(nullptr, SvgNodeType::Defs);
     if (!node) return nullptr;
     simpleXmlParseAttributes(buf, bufLength, nullptr, node);
@@ -2197,6 +2198,7 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content,
             if (loader->stack.count > 0) parent = loader->stack.data[loader->stack.count - 1];
             else parent = loader->doc;
             node = method(loader, parent, attrs, attrsLength);
+            if (!node) return;
         }
 
         if (node->type == SvgNodeType::Defs) {
```

JSUYA (commented):
Please check my comment :)

src/loaders/svg/tvgSvgLoader.cpp
mmaciola: Great idea!
I didn't cared about parent, as for _createDefsNode it is called _createNode(/*parent=*/ nullptr, /*type=*/ SvgNodeType::Defs); so parent is null and wont be pushed.
I think its great to abandon node creating instead of deleting it after creation. I will push a patch exactly like your idea.

JSUYA (approved):
LGTM :)

PR #529, opened 2021-06-29
svg_loader: fix the inifite loop bug when size has 'em' unit.

hermet:
svg may have 'em' unit for fonts, strtof() must handle this case to not
drop in the infinite loop.

Yet thorvg doesn't support font feature in svg,
it must multiply values with font size if those values have 'em' size.

@Issue: https://github.com/Samsung/thorvg/issues/488

### DIFF
--- a/src/loaders/svg/tvgSvgUtil.cpp
+++ b/src/loaders/svg/tvgSvgUtil.cpp
@@ -26,7 +26,6 @@
 #include <errno.h>
 #include "tvgSvgUtil.h"
 
-
 /************************************************************************/
 /* Internal Class Implementation                                        */
 /************************************************************************/
@@ -163,14 +162,21 @@ float svgUtilStrtof(const char *nPtr, char **endPtr)
         unsigned int expo_part;
         int minus_e;
 
-        iter++;
+        ++iter;
+
+        //Exception: svg may have 'em' unit for fonts. ex) 5em, 10.5em
+        if ((*iter == 'm') || (*iter == 'M')) {
+            //TODO: We don't support font em unit now, but has to multiply val * font size later...
+            a = iter + 1;
+            goto on_success;
+        }
 
         //signed or not
         minus_e = 1;
         if (*iter == '-')
         {
             minus_e = -1;
-            iter++;
+            ++iter;
         }
         else if (*iter == '+') iter++;
 


mmaciola (approved):

src/loaders/svg/tvgSvgUtil.cpp
```diff
@@ -163,14 +162,21 @@ float svgUtilStrtof(const char *nPtr, char **endPtr)
...
+            goto on_success;
+        }
 
         //signed or not
         minus_e = 1;
         if (*iter == '-')
         {
             minus_e = -1;
-            iter++;
+            ++iter;
```
mgrudzinska, line 179: out of curiosity - is there reason for which you have changed it? there are still other places where 'iter++' is used

hermet: for consistency. you can change it others.

hermet: @mgrudzinska I'm not just hang around to change the coding convention, when touch the code by the chance then it's the time to change violations near my code.

src/loaders/svg/tvgSvgUtil.cpp
```diff
@@ -163,14 +162,21 @@ float svgUtilStrtof(const char *nPtr, char **endPtr)
         unsigned int expo_part;
         int minus_e;
 
-        iter++;
+        ++iter;
+
+        //Exception: svg may have 'em' unit for fonts. ex) 5em, 10.5em
+        if ((*iter == 'm') || (*iter == 'M')) {
```
mgrudzinska, line 168: for now any unit indicator present after the number is returned from this funct. 
and after this patch the 'em' unit will be the only one that is "eaten" in this function. 
for consistency I'd rather add to the if-else block below an extra 'else' in which I'll return the "eaten" 'e'.

src/loaders/svg/tvgSvgUtil.cpp
mgrudzinska: ok, sure, I understand and I'll remember to use ++x

src/loaders/svg/tvgSvgUtil.cpp
hermet: @mgrudzinska sorry i can't catch your point.. so which line this condition supposed to be ?

mgrudzinska: sorry for not being precise
it's l.194 in the original file.
I'd add there:
```
else {
    a = iter - 1;
    goto on_success;
}
```
what do you think about such solution?

mgrudzinska: so if there is anything else than '+', '-" or a digit, we leave it

hermet: @mgrudzinska  This handles exponential part and it might not going there by  "else if (!isdigit(*(a - 1)))"

        //exponential part
        expo_part = 0;
        if (isdigit(*iter))
        {
            while (*iter == 0) iter++;

            for (; isdigit(*iter); iter++)
            {
                expo_part = expo_part * 10U + (unsigned int)(*iter - '0');
            }
        }
        else if (!isdigit(*(a - 1)))
        {
            a = nPtr;
            goto on_success;
        }
        else if (*iter == 0) goto on_success;


even does,  I think the purpose is different contextually . 

0. Exception Handling
1. signed or not
2. Exponential part

either way, no correct answer. it's trivial by me.

mgrudzinska: I'm not sure any more about the 'eating' that I mentioned before. Sorry.

mgrudzinska: So probably your conception was ok

Issue #515, opened 2021-06-25
InkScape and Firefox render the cube red


fire:
https://github.com/godotengine/godot/issues/41339

[icon_voxel_terrain.zip](https://github.com/Samsung/thorvg/files/6716904/icon_voxel_terrain.zip)

InkScape and Firefox render the cube red.

I believe ThorVG and NanoSVG render it blue and red.

hermet:
<path
       d="m8 1051.4-7-3v-8l7 3z"
       id="path4"
       style="fill:#fc9c9c;fill-opacity:0.99607843"
       fill="#68b6ff" />

Here style fill="fc9c9c" should override the color beyond the fill="68b6ff"

But tvg doesn't this...

mmaciola:
I checked the attached file. Blue and red rendering is the expected result: fill is set twice for each wall; once with `fill` attribute that sets bluish color and once with `style="fill:..;"` attribute that sets redish color. For right and top wall, style is set after fill (so style ovewrites fill and wall is redish). For left wall, it is fill attribute that is set last, so wall is properly rendered bluish.

Zylann:
Aren't styles supposed to take precedence over attributes? https://stackoverflow.com/questions/6525405/svg-use-attributes-or-css-to-style/36755226#36755226

mgrudzinska:
style has higher priority than fill, the cube should be red

hermet:
@fire fixed in v0.3.1

PR #530, opened 2021-06-29
svg_loader: set the higher priority of style attributes

mgrudzinska:
issue #515

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -717,7 +717,7 @@ static float _parseLength(const char* str, SvgLengthType* type)
 }
 
 
-static bool _parseStyleAttr(void* data, const char* key, const char* value);
+static bool _parseStyleAttr(void* data, const char* key, const char* value, bool style);
 
 
 static bool _attrParseSvgNode(void* data, const char* key, const char* value)
@@ -755,7 +755,7 @@ static bool _attrParseSvgNode(void* data, const char* key, const char* value)
     }
 #endif
     else {
-        return _parseStyleAttr(loader, key, value);
+        return _parseStyleAttr(loader, key, value, false);
     }
     return true;
 }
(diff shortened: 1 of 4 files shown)


mmaciola (approved):
Looks good, I like it. Narrow cases like `<path fill="white" fill="black" .. />` are also handled properly

mgrudzinska:
stop attributes are handled in the second commit

I need to read the specs before doing the same for other attribs of other tags. I'm not sure whether it should work  like in this pr or not.

src/loaders/svg/tvgXmlParser.h
```diff
@@ -46,10 +46,11 @@ enum class SimpleXMLType
 
 typedef bool (*simpleXMLCb)(void* data, SimpleXMLType type, const char* content, unsigned length);
 typedef bool (*simpleXMLAttributeCb)(void* data, const char* key, const char* value);
+typedef bool (*simpleXMLW3CAttributeCb)(void* data, const char* key, const char* value, bool style);
 
 bool simpleXmlParseAttributes(const char* buf, unsigned buflen, simpleXMLAttributeCb func, const void* data);
 bool simpleXmlParse(const char* buf, unsigned buflen, bool strip, simpleXMLCb func, const void* data);
-bool simpleXmlParseW3CAttribute(const char* buf, simpleXMLAttributeCb func, const void* data);
+bool simpleXmlParseW3CAttribute(const char* buf, simpleXMLW3CAttributeCb func, const void* data);
```
JSUYA, line 53: I don't think the `W3C` keyword has anything to do with the priority of the style attribute.
And I don't think we need to create a new function definition.

`_parseStyleAttr(void*, char*, char*)` calls `_parseStyleAttr(data, key, value, true);` and 
Calling each `_parseStyleAttr(... , false)`, we can avoid unnecessary definitions.

+) If we allow a little more code duplication, we can reduce the number of if-checks when looking for styleTags by making style=true and style=false functions respectively. It is optional anyway.

JSUYA (commented):
Please my comment :)

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -937,20 +940,31 @@ static bool _parseStyleAttr(void* data, const char* key, const char* value)
...
     sz = strlen(key);
     for (unsigned int i = 0; i < sizeof(styleTags) / sizeof(styleTags[0]); i++) {
         if (styleTags[i].sz - 1 == sz && !strncmp(styleTags[i].tag, key, sz)) {
-            styleTags[i].tagHandler(loader, node, value);
+            if (style) {
+                styleTags[i].tagHandler(loader, node, value);
+                node->style->flags = (SvgStyleFlags)((int)node->style->flags | (int)styleTags[i].flag);
+            } else if (!((int)node->style->flags && (int)styleTags[i].flag)) {
+                styleTags[i].tagHandler(loader, node, value);
+            }
```
mgrudzinska, line 953: @JSUYA If you think that it's a good idea I can duplicate this func. 
in the one for style=true we dont need 'if(style)', but the one for style=false still has to check 'if (!((int)node->style->flags && (int)styleTags[i].flag))'. 
I wouldn't duplicate this func, but if you wish, I'll do it. 

I did that for the '_attrParseStops' - I defined '_attrParseStopsStyle' for the case style=true. But in this case it was necessary since other way I'd have to modify the XmlParser.

JSUYA: ok I agree to your patch.

JSUYA (approved):

JSUYA:
Please rebase and merge this PR

Issue #521, opened 2021-06-26
SVG - stop-opacity cuts the color completely


mgrudzinska:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 140 140">
  <defs>
    <linearGradient
       id="grad" x1="0" y1="0" x2="0.2" y2="0.2" spreadMethod="reflect">
      <stop style="stop-color:#ff0000;stop-opacity:1;" offset="0"/>
      <stop style="stop-color:#0000ff;stop-opacity:0;" offset="1"/>
    </linearGradient>
  </defs>
  <rect x="0" y="0" width="100" height="100" fill="url(#grad)"/>
</svg>
```

tvg result:
![5before](https://user-images.githubusercontent.com/67589014/123713776-6d369900-d875-11eb-9151-e732c13a78b7.PNG)

expected result:
![5expected](https://user-images.githubusercontent.com/67589014/123713795-76276a80-d875-11eb-9501-f4d8b55163ce.PNG)

hermet:
@mgrudzinska can you explain this issue more detail? so that other volunteers can assign it.

mgrudzinska:
sorry, I edited the descr

mmaciola:
note it is thorvg issue, not just svg
```
void tvgDrawCmds(tvg::Canvas* canvas)
{
    if (!canvas) return;

    //Background
    auto shape = tvg::Shape::gen();
    shape->appendRect(0, 0, WIDTH, HEIGHT, 0, 0);
    shape->fill(255, 255, 255, 255);
    canvas->push(move(shape));

    //Shape
    auto shape1 = tvg::Shape::gen();
    shape1->appendRect(0, 0, WIDTH, HEIGHT, 0, 0);

    //LinearGradient
    tvg::Fill::ColorStop colorStops[2];
    colorStops[0] = {0, 255, 0, 0, 255};
    colorStops[1] = {1, 0, 0, 255, 25};

    auto fill = tvg::LinearGradient::gen();
    fill->linear(0, 0, WIDTH, HEIGHT);
    fill->colorStops(colorStops, 2);

    shape1->fill(move(fill));
    if (canvas->push(move(shape1)) != tvg::Result::Success) return;
}
```
![Screenshot from 2021-06-29 17-46-03](https://user-images.githubusercontent.com/71131832/123828638-feebe800-d901-11eb-94c6-3840503ab6c4.png)

mgrudzinska:
solved in #531

[...]

PR #531, opened 2021-06-29
sw_engine: stop-opacity value should be interpolated between stops

mgrudzinska:
The opacity value is interpolated as are all color channels.

example with apis:
```
    shape1->appendRect(0, 0, 400, 400, 0, 0);    

    fill->linear(0, 0, 400, 400);

    colorStops[0] = {0, 255, 0, 0, 255};
    colorStops[1] = {1, 0, 0, 255, 0};

    shape1->fill(move(fill));
    if (canvas->push(move(shape1)) != tvg::Result::Success) return;
```

svg xample (same result as from apis):
```
<svg viewBox="0 0 800 800">
  <defs>
      <stop style="stop-color:#0000ff;stop-opacity:1;" offset="0"/>
      <stop style="stop-color:#ff0000;stop-opacity:0;" offset="1"/>
    </linearGradient>
  </defs>
  <rect x="0" y="0" width="400" height="400" fill="#000"/>
  <rect x="0" y="0" width="400" height="400" fill="url(#grad)"/>
</svg>
```

before:
7before

after:
7after

issue #521

### DIFF
--- a/src/lib/sw_engine/tvgSwFill.cpp
+++ b/src/lib/sw_engine/tvgSwFill.cpp
@@ -49,11 +49,11 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface* 
     auto a = (pColors->a * opacity) / 255;
     if (a < 255) fill->translucent = true;
 
-    auto r = ALPHA_MULTIPLY(pColors->r, a);
-    auto g = ALPHA_MULTIPLY(pColors->g, a);
-    auto b = ALPHA_MULTIPLY(pColors->b, a);
-
+    auto r = pColors->r;
+    auto g = pColors->g;
+    auto b = pColors->b;
     auto rgba = surface->blender.join(r, g, b, a);
+
     auto inc = 1.0f / static_cast<float>(GRADIENT_STOP_SIZE);
     auto pos = 1.5f * inc;
     uint32_t i = 0;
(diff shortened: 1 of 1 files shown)


src/lib/sw_engine/tvgSwFill.cpp
```diff
@@ -70,31 +70,38 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface*
...
             auto dist = static_cast<int32_t>(256 * t);
             auto dist2 = 256 - dist;
-            fill->ctable[i] = COLOR_INTERPOLATE(rgba, dist2, rgba2, dist);
+
+            auto color = COLOR_INTERPOLATE(rgba, dist2, rgba2, dist);
+            uint8_t a = color >> 24;
+            uint8_t c1 = ALPHA_MULTIPLY((color >> 16) & 0xff, a);
+            uint8_t c2 = ALPHA_MULTIPLY((color >> 8) & 0xff, a);
+            uint8_t c3 = ALPHA_MULTIPLY(color & 0xff, a);
+            fill->ctable[i] = (a << 24 | c1 << 16 | c2 << 8 | c3);
```
mgrudzinska, line 88: at this point it is not important which channel is 'r', 'g', 'b'. I spilt the color into these channels, multiply by the aplha val and join the channels in the same way, that I split them.

src/lib/sw_engine/tvgSwFill.cpp
```diff
@@ -70,31 +70,38 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface*
...
         }
         rgba = rgba2;
     }
 
+    r = ALPHA_MULTIPLY(r, a);
+    g = ALPHA_MULTIPLY(g, a);
+    b = ALPHA_MULTIPLY(b, a);
+
     for (; i < GRADIENT_STOP_SIZE; ++i)
-        fill->ctable[i] = rgba;
```
hermet, line 94: I think you can pre-compute rgba and use it in the loop.

src/lib/sw_engine/tvgSwFill.cpp
```diff
@@ -70,31 +70,38 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface*
...
+        fill->ctable[i] = surface->blender.join(r, g, b, a);
 
     //Make sure the lat color stop is represented at the end of the table
-    fill->ctable[GRADIENT_STOP_SIZE - 1] = rgba;
+    fill->ctable[GRADIENT_STOP_SIZE - 1] = surface->blender.join(r, g, b, a);;
```
hermet, line 104: same rgba.

hermet (changes requested):
Please check comments.

src/lib/sw_engine/tvgSwFill.cpp
mgrudzinska: of course, but what's worse, I was using the wrong values (from the first stop instead from the last one) ! fixed now :)

src/lib/sw_engine/tvgSwFill.cpp
mgrudzinska: @mmaciola I can not do that, since then 'a' will be also multiplied

mmaciola: Are you sure that 'a' in color isn't 0xff there, if so it will be the same
`ALPHA_BLEND(color | 0xff000000, a);` might be also faster

mgrudzinska: you're right, done

src/lib/sw_engine/tvgSwFill.cpp
```diff
@@ -22,7 +22,7 @@
 #include <float.h>
 #include <math.h>
 #include "tvgSwCommon.h"
-
+#include <iostream>
```
hermet, line 25: is it necessary?

hermet (changes requested):

src/lib/sw_engine/tvgSwFill.cpp
mgrudzinska: such an easy pr and so many problems... sorry. I hope now it's ok

hermet (approved):

PR #539, opened 2021-06-30
svg_loader: fix a infinite loop error if compositions have circular d…

hermet:
…ependency.

Composition can be applied recursively if its children nodes have composition target to this one.
This can be occured by wrong svg description, and tvg prevents this exception case.

@Issue: https://github.com/Samsung/thorvg/issues/494

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -179,8 +179,19 @@ static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float
 {
     if (node->style->comp.method == CompositeMethod::None) return;
 
+    /* Do not drop in Circular Dependency.
+       Composition can be applied recursively if its children nodes have composition target to this one. */
+    if (node->style->comp.applying) {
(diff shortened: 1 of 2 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -179,9 +179,20 @@ static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float
 {
     if (node->style->comp.method == CompositeMethod::None) return;
 
+    /* Do not drop in Circular Dependency.
+       Composition can be applied recursively if its children nodes have composition target to this one. */
+    if (node->style->comp.applying) {
```
JSUYA, line 184: Missing initialization for this member.

hermet: @JSUYA  here? node->style = (SvgStyleProperty*)calloc(1, sizeof(SvgStyleProperty))

JSUYA: ah i forgot it.

JSUYA (approved):

PR #551, opened 2021-07-01
example: Svg3.cpp created- embeded <image> tag

mmaciola:
This patch introduces Svg3.cpp example file and 'image-embeded.svg' that are
testing embeded <image> tag.


src/examples/Svg3.cpp
```diff
@@ -0,0 +1,153 @@
+/*
```
hermet, line 1: I think you can add svg resource to images and test with Svg example.
This example code itself doesn't have any special showcase code.

hermet (changes requested):
Please have a look comment.

src/examples/Svg3.cpp
mmaciola: Ok

Issue #546, opened 2021-06-30
Memory leak when converting image (svg2png)


qarmin:
[e-mask-015.svg.zip](https://github.com/Samsung/thorvg/files/6743401/e-mask-015.svg.zip)

```
Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x7f01ad5f35a7 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:99
    #1 0x7f01ad321dba in _idFromUrl ../src/loaders/svg/tvgSvgLoader.cpp:282
    #2 0x7f01ad32536a in _handleMaskAttr ../src/loaders/svg/tvgSvgLoader.cpp:887
    #3 0x7f01ad32932c in _attrParseRectNode ../src/loaders/svg/tvgSvgLoader.cpp:1420
    #4 0x7f01ad34f450 in simpleXmlParseAttributes(char const*, unsigned int, bool (*)(void*, char const*, char const*), void const*) ../src/loaders/svg/tvgXmlParser.cpp:319
    #5 0x7f01ad334f7e in _createRectNode ../src/loaders/svg/tvgSvgLoader.cpp:1437
    #6 0x7f01ad33b29e in _svgLoaderParserXmlOpen ../src/loaders/svg/tvgSvgLoader.cpp:2212
    #7 0x7f01ad33ba52 in _svgLoaderParser ../src/loaders/svg/tvgSvgLoader.cpp:2257
    #8 0x7f01ad34f673 in simpleXmlParse(char const*, unsigned int, bool, bool (*)(void*, SimpleXMLType, char const*, unsigned int), void const*) ../src/loaders/svg/tvgXmlParser.cpp:425
    #9 0x7f01ad33cd8b in SvgLoader::run(unsigned int) ../src/loaders/svg/tvgSvgLoader.cpp:2615
    #10 0x7f01ad2ccf3e in tvg::Task::operator()(unsigned int) ../src/lib/tvgTaskScheduler.h:68
    #11 0x7f01ad2ccf3e in tvg::TaskSchedulerImpl::run(unsigned int) ../src/lib/tvgTaskScheduler.cpp:138
    #12 0x7f01acfbe693  (/lib/x86_64-linux-gnu/libstdc++.so.6+0xda693)

SUMMARY: AddressSanitizer: 32 byte(s) leaked in 1 allocation(s).
```

mmaciola:
Memory leak fixed in #556
[...]

PR #556, opened 2021-07-01
svg_loader: multiple composition memory leak

mmaciola:
Multiple composition are still not supported, but this patch fixes a
memory leak if it is tried to be applied.

@issue: #546


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -869,7 +869,11 @@ static void _handleClipPathAttr(TVG_UNUSED SvgLoaderData* loader, SvgNode* node,
 #endif
```
hermet, line 869: how about return here immediately?

hermet (changes requested):

JSUYA (commented):
I think you probably copied the logic of `_handleMaskAttr`. If you want to change that function return immediately, the mask case should be modified in the same way.

hermet (approved):

Issue #545, opened 2021-06-30
Memory leak when converting svg file (svg2png)


qarmin:
[e-linearGradient-009.svg.zip](https://github.com/Samsung/thorvg/files/6743390/e-linearGradient-009.svg.zip)

```
==40604==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 16 byte(s) in 2 object(s) allocated from:
    #0 0x7fbbf9957e17 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:154
    #1 0x7fbbf96a163b in _svgLoaderParserXmlOpen ../src/loaders/svg/tvgSvgLoader.cpp:2229
    #2 0x7fbbf96a1a52 in _svgLoaderParser ../src/loaders/svg/tvgSvgLoader.cpp:2257
    #3 0x7fbbf96b5673 in simpleXmlParse(char const*, unsigned int, bool, bool (*)(void*, SimpleXMLType, char const*, unsigned int), void const*) ../src/loaders/svg/tvgXmlParser.cpp:425
    #4 0x7fbbf96a2d8b in SvgLoader::run(unsigned int) ../src/loaders/svg/tvgSvgLoader.cpp:2615
    #5 0x7fbbf9632f3e in tvg::Task::operator()(unsigned int) ../src/lib/tvgTaskScheduler.h:68
    #6 0x7fbbf9632f3e in tvg::TaskSchedulerImpl::run(unsigned int) ../src/lib/tvgTaskScheduler.cpp:138
    #7 0x7fbbf9324693  (/lib/x86_64-linux-gnu/libstdc++.so.6+0xda693)

SUMMARY: AddressSanitizer: 16 byte(s) leaked in 2 allocation(s).
```

mmaciola:
The attached file looks improper or non-standard (`<stop>` tag is located inside `<rect>`). Exception handled and memory leak fixed in #557

PR #557, opened 2021-07-01
svg_loader: memory leak after improper stop tag

mmaciola:
`<stop>` tag should be always a child of a `<linearGradient>` or `<radialGradient>` element, but there were files with `<stop>` tag inside `<rect>`.
This patch fixes a memory leak if no gradient is defined before `<stop>` tag.

@issue: #545

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2234,7 +2234,7 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content, 
     } else if (!strcmp(tagName, "stop")) {
         if (!loader->latestGradient) {
 #ifdef THORVG_LOG_ENABLED
-            printf("SVG: Gradient stop but no gradient in the scope\n");
+            printf("SVG: Stop element is used outside of the Gradient element\n");
 #endif
             return;
         }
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2232,15 +2232,19 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content,
         }
         loader->latestGradient = gradient;
     } else if (!strcmp(tagName, "stop")) {
+        if (!loader->latestGradient) {
+#ifdef THORVG_LOG_ENABLED
+            printf("SVG: Gradient stop but no gradient in the scope\n");
```
hermet, line 2237: A little ambiguous to me. Could you please write it for more user-friendly?
The logs are tip information for users (designers & developers)

ex) Stop element is used outside of Gradient element.

hermet (changes requested):

src/loaders/svg/tvgSvgLoader.cpp
mmaciola: Ok, done

hermet:
@JSUYA please verify this.

hermet (approved):

PR #561, opened 2021-07-02
tests capi: Add capi Scene TC

jykeon:
issue: https://github.com/Samsung/thorvg/issues/45

### DIFF
--- a/test/capi/capiScene.cpp
+++ b/test/capi/capiScene.cpp
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All rights reserved.
+
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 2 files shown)


test/capi/capiScene.cpp
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+#include <thorvg_capi.h>
+#include "../catch.hpp"
+
+TEST_CASE("Scene Creation", "[capiScene]")
+{
+    Tvg_Paint* scene = tvg_scene_new();
+    REQUIRE(scene);
+
```
hermet, line 30: Please remove empty lines.

hermet (changes requested):

test/capi/capiScene.cpp
```diff
@@ -0,0 +1,64 @@
...
+TEST_CASE("Pushing Paints Into Scene", "[capiScene]")
+{
+    Tvg_Paint* scene = tvg_scene_new();
+    REQUIRE(scene);
+    Tvg_Paint* paint = tvg_picture_new();
+    REQUIRE(paint);
+    
+    REQUIRE(tvg_scene_push(scene, paint) == TVG_RESULT_SUCCESS);
+    
+
```
hermet, line 43: you must delete scene in the end of the function.
See:  tvg_paint_del(scene); 

Other TC case as well.

test/capi/capiScene.cpp
```diff
@@ -0,0 +1,64 @@
...
+    REQUIRE(tvg_scene_reserve(scene,100) == TVG_RESULT_SUCCESS):
+    
+}
+
+TEST_CASE("Scene Clear", "[capiScene]")
+{
+     Tvg_Paint* scene = tvg_scene_new();
+     REQUIRE(scene);
+     
+     REQUIRE(tvg_scene_clear(scene,1) == TVG_RESULT_SUCCESS) ;
```
hermet, line 60: 1 => true/false

hermet (changes requested):
I think API coverage per each TC is too weak. Rather than separate tests, collaborate all in one brings more strong test.

PR #563, opened 2021-07-02
test Picture: add Picture unit tests

hermet:


test/testPicture.cpp
```diff
@@ -0,0 +1,95 @@
...
+    REQUIRE(picture->load(nullptr, 100) == Result::InvalidArguments);
+    REQUIRE(picture->load(svg, 0) == Result::InvalidArguments);
+
+    //Positive cases
+    REQUIRE(picture->load(svg, strlen(svg)) == Result::Success);
+
+    float w, h;
+    REQUIRE(picture->size(&w, &h) == Result::Success);
+    REQUIRE(w == 1000.0f);
+    REQUIRE(h == 1000.0f);
```
JSUYA, line 61: Catch2 provides `Approx` class for floating point comparison. We can refer to this.

https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md#floating-point-comparisons

JSUYA (commented):
Please check my comment

test/testPicture.cpp
hermet: @JSUYA Requested size are returned without any computation. Must be exactly same to the input.

Issue #568, opened 2021-07-02
<use> element used before <defs> element


mmaciola:
Some svg software seems to include `<defs>` element at the end of the file (so after `<use>` element). In such scenario thorvg's svg_loader can't find and clone node, so such element won't be showed.

PR #569, opened 2021-07-02
svg_loader: postpone cloneNode()

mmaciola:
Some svg export software puts `<defs>` element at the end of the file.
If so, the `<defs>` element won't be found, when parsing `<use>`.
In such scenario, this patch postpone node cloning until the whole file
is parsed.

@issue: #568

### DIFF
--- a/src/loaders/svg/tvgSvgLoaderCommon.h
+++ b/src/loaders/svg/tvgSvgLoaderCommon.h
@@ -300,6 +300,12 @@ struct SvgParser
     } gradient;
 };
 
+struct SvgNodeIdPair
+{
+    SvgNode* node;
+    string *id;
+};
+
 struct SvgLoaderData
 {
     Array<SvgNode *> stack = {nullptr, 0, 0};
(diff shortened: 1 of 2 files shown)


src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -308,6 +314,7 @@ struct SvgLoaderData
     Array<SvgStyleGradient*> gradients;
     SvgStyleGradient* latestGradient = nullptr; //For stops
     SvgParser* svgParse = nullptr;
+    Array<SvgNodeIdPair*> cloneNodes;
```
hermet, line 317: It looks too fragmented memory.
How about  Array<SvgNodeIdPair>?

hermet (changes requested):
Please check a comment.

src/loaders/svg/tvgSvgLoaderCommon.h
mmaciola: Could you describe more?

JSUYA (approved):
I don't think there are any issues with this patch. I wait for hermet's approval.

src/loaders/svg/tvgSvgLoaderCommon.h
hermet: @mmaciola sorry for this, \< , \> these tags are ignored by this github viewer. see again the above.

mmaciola: ok. done

hermet (approved):

hermet:
@mmaciola thanks.

PR #570, opened 2021-07-02
png_loader: added open from memory

mmaciola:
Added PngLoader::open(const char* data, uint32_t size, bool copy)
Changed variables visibility to private

### DIFF
--- a/src/loaders/png/tvgPngLoader.cpp
+++ b/src/loaders/png/tvgPngLoader.cpp
@@ -51,6 +51,18 @@ bool PngLoader::open(const string& path)
     return true;
 }
 
+bool PngLoader::open(const char* data, uint32_t size, bool copy)
+{
+    image->opaque = NULL;
+
+    if (!png_image_begin_read_from_memory(image, data, size)) return false;
+
+    vw = w = image->width;
+    vh = h = image->height;
+
+    return true;
+}
+
 bool PngLoader::read()
 {
     png_bytep buffer;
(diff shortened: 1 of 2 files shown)


src/loaders/png/tvgPngLoader.cpp
```diff
@@ -51,6 +51,18 @@ bool PngLoader::open(const string& path)
     return true;
 }
 
+bool PngLoader::open(const char* data, uint32_t size, bool copy)
```
hermet, line 54: Did you verify read() works fine even after data is freed?
I think it needs to copy the data if copy == true.

hermet (changes requested):
Please check a comment.

src/loaders/png/tvgPngLoader.cpp
mmaciola: The change was tested. There is no need to copy as task aren't used for PngLoader.

hermet: @mmaciola  like this sequence?
1. picture->open()
2. free(data)
3. picture->push()

mmaciola: The test scenario is below. Works ok
```
    string path(EXAMPLE_DIR"/logo.png");

    //Open
    ifstream file(path);
    if (!file.is_open()) return ;
    auto size = sizeof(uint32_t) * (1000*1000);
    auto data = (char*)malloc(size);
    file.read(data, size);
    file.close();

    //Picture
    picture->load(data, size, false);

    //free and push
    free(data);
    canvas->push(move(picture));
```

hermet: @mmaciola good thanks!

Issue #542, opened 2021-06-30
Freeze when using svg2png tool with specific svg


qarmin:
[a-stroke-dasharray-004.svg.zip](https://github.com/Samsung/thorvg/files/6743379/a-stroke-dasharray-004.svg.zip)

GDB
```
Thread 1 "svg2png" received signal SIGINT, Interrupt.
__futex_abstimed_wait_common64 (cancel=true, private=0, abstime=0x0, clockid=0, expected=0, futex_word=0x612000000cc0) at ../sysdeps/nptl/futex-internal.c:74
74	../sysdeps/nptl/futex-internal.c: No such file or directory.
(gdb) backtrace
#0  __futex_abstimed_wait_common64 (cancel=true, private=0, abstime=0x0, clockid=0, expected=0, futex_word=0x612000000cc0)
    at ../sysdeps/nptl/futex-internal.c:74
#1  __GI___futex_abstimed_wait_cancelable64 (futex_word=futex_word@entry=0x612000000cc0, expected=expected@entry=0, clockid=clockid@entry=0, 
    abstime=abstime@entry=0x0, private=private@entry=0) at ../sysdeps/nptl/futex-internal.c:123
#2  0x00007ffff63df540 in __pthread_cond_wait_common (abstime=0x0, clockid=0, mutex=0x612000000c70, cond=0x612000000c98) at pthread_cond_wait.c:504
#3  __pthread_cond_wait (cond=0x612000000c98, mutex=0x612000000c70) at pthread_cond_wait.c:628
#4  0x00007ffff7048280 in std::condition_variable::wait(std::unique_lock<std::mutex>&) () from /lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff738f83d in tvg::Task::done (this=this@entry=0x612000000c68) at ../src/lib/tvgTaskScheduler.h:58
#6  0x00007ffff73bd84a in SvgLoader::scene (this=0x612000000c40) at ../src/loaders/svg/tvgSvgLoader.cpp:2743
#7  0x00007ffff7341083 in tvg::Picture::Impl::reload (this=this@entry=0x6060000004a0) at ../src/lib/tvgPictureImpl.h:100
#8  0x00007ffff734298b in tvg::Picture::Impl::update (pFlag=tvg::All, clips=..., opacity=255, transform=0x0, renderer=..., this=0x6060000004a0)
    at ../src/lib/tvgPictureImpl.h:119
[...]

PR #572, opened 2021-07-02
svg_loader: fix dead loop on non-digit char in stroke-dasharray

mmaciola:
Any non digit character in `stroke-dasharray` resulted in a dead looping

@issue: #542


JSUYA (approved):
LGTM

PR #580, opened 2021-07-05
sw_engine: stop-opacity value should be interpolated between stops

mgrudzinska:
The opacity value is interpolated as are all color channels.
It's similar to the reverted commit ac95433, but with two slight changes
(alpha_blend was missing in the ctable init and 'a' was used in an if
statement instead of 'a2').

like #531 but with 2 changes - marked in comments

### DIFF
--- a/src/lib/sw_engine/tvgSwFill.cpp
+++ b/src/lib/sw_engine/tvgSwFill.cpp
@@ -49,16 +49,16 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface* 
     auto a = (pColors->a * opacity) / 255;
     if (a < 255) fill->translucent = true;
 
-    auto r = ALPHA_MULTIPLY(pColors->r, a);
-    auto g = ALPHA_MULTIPLY(pColors->g, a);
-    auto b = ALPHA_MULTIPLY(pColors->b, a);
-
+    auto r = pColors->r;
+    auto g = pColors->g;
+    auto b = pColors->b;
     auto rgba = surface->blender.join(r, g, b, a);
+
     auto inc = 1.0f / static_cast<float>(GRADIENT_STOP_SIZE);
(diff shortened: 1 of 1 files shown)


src/lib/sw_engine/tvgSwFill.cpp
```diff
@@ -49,16 +49,16 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface*
...
+    auto g = pColors->g;
+    auto b = pColors->b;
     auto rgba = surface->blender.join(r, g, b, a);
+
     auto inc = 1.0f / static_cast<float>(GRADIENT_STOP_SIZE);
     auto pos = 1.5f * inc;
     uint32_t i = 0;
 
-    fill->ctable[i++] = rgba;
+    fill->ctable[i++] = ALPHA_BLEND(rgba | 0xff000000, a);
```
mgrudzinska, line 61: it was missing in #531

src/lib/sw_engine/tvgSwFill.cpp
```diff
@@ -70,30 +70,32 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface*
...
         auto next = curr + 1;
         auto delta = 1.0f / (next->offset - curr->offset);
-        a = (next->a * opacity) / 255;
-        if (!fill->translucent && a < 255) fill->translucent = true;
-
-        auto r = ALPHA_MULTIPLY(next->r, a);
-        auto g = ALPHA_MULTIPLY(next->g, a);
-        auto b = ALPHA_MULTIPLY(next->b, a);
+        auto a2 = (next->a * opacity) / 255;
+        if (!fill->translucent && a2 < 255) fill->translucent = true;
```
mgrudzinska, line 74: in #531 'a' was used instead of 'a2'

hermet (approved):

PR #581, opened 2021-07-06
svg_loader: Support percentage unit of dash-array

JSUYA:
Refers to the diagonal length of the viewport.
https://www.w3.org/TR/SVG2/coords.html#Units

issue: https://github.com/Samsung/thorvg/pull/572 , https://github.com/Samsung/thorvg/issues/542


mmaciola (approved):
LGTM

PR #582, opened 2021-07-06
draft: jpg_loader: introduced jpg decode using jpeg-compressor by richgel999

mmaciola:
This patch introduces an alternative jpg loader using jpeg-compressor by richgel999.
note: `jpgd.cpp` is modified version (changed `decompress_jpeg_image_from_stream`) for returning BGRA.


fire:
I appreciate you and your team took the time to do both implementations.

This will make Godot Engine's undecided switch to jpeg-turbo separate from the merge of thorvg as the svg viewer.

hermet:
we can revise this to support both version (libjpeg-turbo and this) later.

fire:
Did you mark what was changed in the jpg code?

mmaciola:
@fire uploaded the patch. I encapsulated the changed code of `jpgd.cpp` (`decompress_jpeg_image_from_stream`) in `#if 1` to mark the change

hermet:
https://github.com/Samsung/thorvg/issues/594

hermet:
Wondering richgel999's jpeg decoder is the best for us? there may be better github project source code we might able to apply ... Need to figure out further.

hermet:
@mmaciola I think soon we might need to revise this patch and apply to thorvg master properly... (choose either way static/dynamic linking from the meson)

PR #586, opened 2021-07-08
example: PictureJpg.cpp

mmaciola:
Created an example for testing jpg loading

Screenshot from 2021-07-08 16-05-07


hermet (approved):
great. thanks!

PR #589, opened 2021-07-09
jpg_loader JpgLoader: Prevent memory leak

JSUYA:
If the copied data in JpgLoader's data open is invalid,
it is processed without being cleared.
Therefore, data memory free is handled in the destructor.


hermet:
@mmaciola  hmm looks jpeg loader is messed up in the memory  usage...

hermet (approved):

PR #592, opened 2021-07-09
jpg_loader: decompress header on opening

mmaciola:
jpg header decompression on open()

### DIFF
--- a/src/loaders/jpg/tvgJpgLoader.cpp
+++ b/src/loaders/jpg/tvgJpgLoader.cpp
@@ -76,7 +76,6 @@ bool JpgLoader::open(const string& path)
 
     if (fread(data, size, 1, jpegFile) < 1) goto failure;
 
-    //decompress header
     if (tjDecompressHeader3(jpegDecompressor, data, size, &width, &height, &inSubsamp, &inColorspace) < 0) goto failure;
 
     ret = true;
@@ -97,7 +96,6 @@ bool JpgLoader::open(const char* data, uint32_t size, bool copy)
 {
     clear();
 
-    //decompress header
     if (tjDecompressHeader3(jpegDecompressor, (unsigned char *) data, size, &width, &height, &inSubsamp, &inColorspace) < 0) return false;
 
     if (copy) {
--- a/src/loaders/jpg/tvgJpgLoader.cpp
+++ b/src/loaders/jpg/tvgJpgLoader.cpp
@@ -76,7 +76,8 @@ bool JpgLoader::open(const string& path)
 
     if (fread(data, size, 1, jpegFile) < 1) goto failure;
 
-    //FIXME: Decompress header here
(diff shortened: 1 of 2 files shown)


src/loaders/jpg/tvgJpgLoader.h
```diff
@@ -47,6 +47,9 @@ class JpgLoader : public Loader
     unsigned char *image = nullptr;
     unsigned long size = 0;
     bool freeData = false;
+
+    int width, height;
```
hermet, line 51: It looks it doesn't need additional variables.
Base class tvgLoader has default size.
    float w = 0, h = 0;         //default image size

hermet (changes requested):
Please check a comment.

src/loaders/jpg/tvgJpgLoader.cpp
```diff
@@ -76,7 +76,8 @@ bool JpgLoader::open(const string& path)
 
     if (fread(data, size, 1, jpegFile) < 1) goto failure;
 
-    //FIXME: Decompress header here
+    //decompress header
```
hermet, line 79: unnecessary comment, its api name is already descriptive.

src/loaders/jpg/tvgJpgLoader.cpp
```diff
@@ -96,6 +97,9 @@ bool JpgLoader::open(const char* data, uint32_t size, bool copy)
 {
     clear();
 
+    //decompress header
```
hermet, line 100: -- comment.

hermet (commented):
more comments.

src/loaders/jpg/tvgJpgLoader.h
mmaciola: Sizes in tvgLoader are float. For turbojpeg sizes must be passed as integer pointers.
Additional variables could be removed, but it would look like this, that is not nice:
```
bool JpgLoader::open(const string& path) {
    ...
    int width, height;
    tjDecompressHeader3(..., &width, &height, ...);
    vw = w = width;
    vh = h = height;
    ...
}
bool JpgLoader::read() {
    ...
    int width = static_cast<int>(w);
    int height = static_cast<int>(h);
    image = (unsigned char *)tjAlloc(width * height * tjPixelSize[TJPF_BGRX]);
    ...
    tjDecompress2(..., width, 0, height, ...);
}
```

hermet: you can use local variables and copy it to members in that function.
ex)
int width, height;
tjDecompress2(..., width, 0, height, ...);
this->width = width;
this->height = height;

hermet: + vw, vh, w, h must be setup in the end.
there is no necessary to allocate more memory.
static casting is not cost actually.

PR #599, opened 2021-07-12
svg_loader SvgPath: Skip invalid path

JSUYA:
If the d attribute of SVG path is not parsed through _nextCommand(),
it is regarded as an invalid data.

issue: https://github.com/Samsung/thorvg/issues/595

### DIFF
--- a/src/loaders/svg/tvgSvgPath.cpp
+++ b/src/loaders/svg/tvgSvgPath.cpp
@@ -286,7 +286,7 @@ static int _numberCount(char cmd)
 }
 
 
-static void _processCommand(Array<PathCommand>* cmds, Array<Point>* pts, char cmd, float* arr, int count, Point* cur, Point* curCtl, Point* startPoint, bool *isQuadratic)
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgPath.cpp
```diff
@@ -525,6 +526,7 @@ bool svgPathToTvgPath(const char* svgPath, Array<PathCommand>& cmds, Array<Point
     while ((path[0] != '\0')) {
         path = _nextCommand(path, &cmd, numberArray, &numberCount);
         if (!path) break;
+        if (cmd == 0 || !strchr(commandList, cmd)) break;
         _processCommand(&cmds, &pts, cmd, numberArray, numberCount, &cur, &curCtl, &startPoint, &isQuadratic);
```
hermet, line 530: Why not (!processCommand(...)) break;
It looks processCommand() can return false when not matched any case.

hermet (changes requested):
Please check a command.

src/loaders/svg/tvgSvgPath.cpp
JSUYA: you are right. i updated patch. thank you

PR #603, opened 2021-07-13
svg_loader: image tag: force size equal to the viewbox

mmaciola:
For some svg files it is needed to force size equal the viewbox
for proper scaling

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -448,7 +448,6 @@ static unique_ptr<Picture> _imageBuildHelper(SvgNode* node, float vx, float vy, 
         }
 
         if (picture->load(href) != Result::Success) return nullptr;
-        picture->size(node->node.image.w, node->node.image.h);
     }
 
     float x, y, w, h;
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -459,6 +459,7 @@ static unique_ptr<Picture> _imageBuildHelper(SvgNode* node, float vx, float vy,
         auto syt = y * sy;
         Matrix m = {sx, 0, node->node.image.x - sxt, 0, sy, node->node.image.y - syt, 0, 0, 1};
         picture->transform(m);
+        picture->size(w, h);
```
JSUYA, line 462: For svg image loading, size set is called twice. (451 line)
Scaling and size set are also duplicated internally.
I don't think we need a size set of lines 462, 451 because we already do the proper transform.
What do you think?

JSUYA (commented):
Please check my comment

src/loaders/svg/tvgSvgSceneBuilder.cpp
mmaciola: Great point. I missed size is called twice. Removing both is enough. Thanks

JSUYA (approved):

Issue #604, opened 2021-07-13
Embeded svg image: possible race condition


mmaciola:
When loading embedded svg images, race condition sometimes happens on arc path processing (other paths seems ok).

Valid image:
![Screenshot from 2021-07-01 08-40-18](https://user-images.githubusercontent.com/71131832/125423643-cd611dd4-3b32-4f12-9269-86a417086bea.png)
Race condition examples:
![Screenshot from 2021-07-05 09-44-36](https://user-images.githubusercontent.com/71131832/125423606-c0063bb5-d466-4931-a738-b31b46397734.png)
![Screenshot from 2021-07-13 11-54-21](https://user-images.githubusercontent.com/71131832/125432705-dd9e6d8c-4578-4691-afe1-ecdfd4e61f1f.png)

mmaciola:
fixed

PR #605, opened 2021-07-13
svg_loader: race condition on arc processing

mmaciola:
Unneeded static keyword in _pathAppendArcTo function led to race conditioning
@issue: #604


hermet (approved):
stupid static here good catch!

PR #606, opened 2021-07-13
sw_renderer: changing the condition for a valid stroke

mgrudzinska:
The product of a stroke alpha value and the opacity should be > 0,
otherwise the stroke is invisible and there is no point to generate its rle.

### DIFF
--- a/src/lib/sw_engine/tvgSwRenderer.cpp
+++ b/src/lib/sw_engine/tvgSwRenderer.cpp
@@ -80,7 +80,7 @@ struct SwShapeTask : SwTask
         if (HALF_STROKE(strokeWidth) > 0) {
             sdata->strokeColor(nullptr, nullptr, nullptr, &strokeAlpha);
         }
-        bool validStroke = (strokeAlpha > 0) || sdata->strokeFill();
+        bool validStroke = (static_cast<uint32_t>(strokeAlpha * opacity / 255) > 0) || sdata->strokeFill();
         auto clipRegion = bbox;
 
         //invisible shape turned to visible by alpha.


src/lib/sw_engine/tvgSwRenderer.cpp
```diff
@@ -80,7 +80,7 @@ struct SwShapeTask : SwTask
         if (HALF_STROKE(strokeWidth) > 0) {
             sdata->strokeColor(nullptr, nullptr, nullptr, &strokeAlpha);
         }
-        bool validStroke = (strokeAlpha > 0) || sdata->strokeFill();
+        bool validStroke = (static_cast<uint32_t>(strokeAlpha * opacity / 255) > 0) || sdata->strokeFill();
```
hermet, line 83: any necessary for this?

hermet (changes requested):

src/lib/sw_engine/tvgSwRenderer.cpp
hermet: above strokeAlpha checks zero case and opacity as well.

mgrudzinska: lets take the shape->opacity(10) and shape->stroke(255, 0, 0, 20)
opacity = 10
alpha = 20
but still, this stroke won't be visible, since 20 * 10 / 255 = 0.8 -> 0
the same is done for a fill l.93-94

hermet:
@mgrudzinska I don't know why https://github.com/Samsung/thorvg/pull/540 includes this patch change also. So... I think your patches are a little hard to review because your patches include the duplication. So every PR I has to review the same stuff again and again that makes me confuse what your patches doing. 

When you write a patch, please separate the change point and write them based on "master" as possible. 

If a patch has to depend on other PR necessarily, please note the dependency order on PR comment so that reviewers can review the prior patch first. But a better method is not to raise a patch until the prior patch is merged.

mgrudzinska:
@hermet of course I'll try my best to do it the way that you've described. I'm sorry

hermet:
applied by this https://github.com/Samsung/thorvg/pull/540

PR #607, opened 2021-07-13
sw_renderer: refactoring in the run function of the shape

mgrudzinska:
Variables 'renderShape' and 'validStroke' renamed to 'visibleFill' and 'visibleStroke'.
The same variables used instead of 'filling' and 'stroking' variables. A code related to
a shape's fill moved from the part of the code responsible for a shape preparation.

### DIFF
--- a/src/lib/sw_engine/tvgSwRenderer.cpp
+++ b/src/lib/sw_engine/tvgSwRenderer.cpp
@@ -69,18 +69,14 @@ struct SwShapeTask : SwTask
     {
         if (opacity == 0) return;  //Invisible
 
-        /* Valid filling & stroking each increases the value by 1.
-           This value is referenced for compositing shape & stroking. */
-        bool stroking = false;
-        bool filling = false;
-
         //Valid Stroking?
         uint8_t strokeAlpha = 0;
         auto strokeWidth = sdata->strokeWidth();
         if (HALF_STROKE(strokeWidth) > 0) {
             sdata->strokeColor(nullptr, nullptr, nullptr, &strokeAlpha);
         }
-        bool validStroke = (strokeAlpha > 0) || sdata->strokeFill();
+        bool visibleStroke = (static_cast<uint32_t>(strokeAlpha * opacity / 255) > 0) || sdata->strokeFill();
+        bool visibleFill = false;
         auto clipRegion = bbox;
 
         //invisible shape turned to visible by alpha.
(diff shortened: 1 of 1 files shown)


src/lib/sw_engine/tvgSwRenderer.cpp
```diff
@@ -69,18 +69,14 @@ struct SwShapeTask : SwTask
...
-        bool stroking = false;
-        bool filling = false;
-
         //Valid Stroking?
         uint8_t strokeAlpha = 0;
         auto strokeWidth = sdata->strokeWidth();
         if (HALF_STROKE(strokeWidth) > 0) {
             sdata->strokeColor(nullptr, nullptr, nullptr, &strokeAlpha);
         }
-        bool validStroke = (strokeAlpha > 0) || sdata->strokeFill();
```
mgrudzinska, line 83: a stroke with the alpha value = 0 is still valid, but invisible

src/lib/sw_engine/tvgSwRenderer.cpp
```diff
@@ -69,18 +69,14 @@ struct SwShapeTask : SwTask
...
+        bool visibleStroke = (static_cast<uint32_t>(strokeAlpha * opacity / 255) > 0) || sdata->strokeFill();
```
mgrudzinska, line 78: this change is included in a separate pr #606

src/lib/sw_engine/tvgSwRenderer.cpp
```diff
@@ -92,41 +88,38 @@ struct SwShapeTask : SwTask
...
         if (flags & (RenderUpdateFlag::Gradient | RenderUpdateFlag::Transform | RenderUpdateFlag::Color)) {
-            auto fill = sdata->fill();
-            if (fill) {
-                auto ctable = (flags & RenderUpdateFlag::Gradient) ? true : false;
-                if (ctable) shapeResetFill(&shape);
-                if (!shapeGenFillColors(&shape, fill, transform, surface, opacity, ctable)) goto err;
-                filling = true;
-            } else {
-                shapeDelFill(&shape);
+            if (visibleFill) {
```
hermet, line 100: The reason to handle invisible case of fill to remove the resource if I remember correctly and it looks still valid.

hermet (changes requested):
Please check a comment.

src/lib/sw_engine/tvgSwRenderer.cpp
mgrudzinska: please check now - it's like the old version

I'd rather do it in a similar way as it is done for a stroke - if a fill is invisible, then we don't need the shape's rle:
```
        //Fill
        if (flags & (RenderUpdateFlag::Gradient | RenderUpdateFlag::Transform | RenderUpdateFlag::Color)) {
            if (visibleFill) {
                /* We assume that if stroke width is bigger than 2,
                   shape outline below stroke could be full covered by stroke drawing.
                   Thus it turns off antialising in that condition.
                   Also, it shouldn't be dash style. */
                auto antiAlias = (strokeAlpha == 255 && strokeWidth > 2 && sdata->strokeDash(nullptr) == 0) ? false : true;
                if (!shapeGenRle(&shape, sdata, antiAlias, clips.count > 0 ? true : false)) goto err;

                if (auto fill = sdata->fill()) {
                    auto ctable = (flags & RenderUpdateFlag::Gradient) ? true : false;
                    if (ctable) shapeResetFill(&shape);
                    if (!shapeGenFillColors(&shape, fill, transform, surface, opacity, ctable)) goto err;
                } else {
                    shapeDelFill(&shape);
                }
            } else {                                    // <-   the new part of the code
                rleFree(shape.rle);
                shape.rle = nullptr;
                shapeDelFill(&shape);
            }
        }
```
but since I'm not 100% sure, I left the old version...

hermet:
applied by this => https://github.com/Samsung/thorvg/pull/540

PR #612, opened 2021-07-14
test: testPicture added .jpg and .png loading tests

mmaciola:


test/testPicture.cpp
```diff
@@ -108,8 +107,75 @@ TEST_CASE("Load PNG file", "[tvgPicture]")
 
     float w, h;
     REQUIRE(picture->size(&w, &h) == Result::Success);
+
+    REQUIRE(w == 1000);
+    REQUIRE(h == 1000);
```
JSUYA, line 112: You can use Approx class for float point comparison.
(https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md#floating-point-comparisons)

mmaciola: I am not sure here. These float sizes are parsed back to int eg. for raw image data size. I would rather catch an error if even a slight change, as it might lead to an error. What do you think?

hermet: @mmaciola agree on this.

hermet (approved):

Issue #428, opened 2021-06-10
SVG Mask doesn't work.


hermet:
Try in the example:
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/mask

mgrudzinska:
currently TVG doesn't support gradient masks, so it doesn't matter what is it's color.
in the example a color defines the 'mask value' - the alpha value of a mask.

To get a 'mask value' from a color (sRGB) the luminance-to-alpha coefficients has to be applied:
```mask_value = (0.2125 * r + 0.7154 * g + 0.0721 * b ) * a / 255``` 
where `r, g, b` - non premultiplied, `a` in 0-255

I just need to think for a while where to apply this so it won't affect the performance.

hermet:
@mgrudzinska Sounds like tvg need to support CompositeMethod such as ColorMasking or AlphMatte/InvAlphaMatte or similar.

mgrudzinska:
@mmaciola 
https://www.w3.org/TR/SVG11/masking.html#Masking
https://www.w3.org/TR/SVG11/filters.html#feColorMatrixElement    <-    "luminanceToAlpha"

mgrudzinska:
after #620 is applied, #988 is also necessary to make the example work

hermet:
Confirmed it works now. :)

PR #620, opened 2021-07-16
compositeMethod: introduced ColorMask

mmaciola:
Introduced CompositeMethod::ColorMask that converts the source pixels to the
grayscale before alpha blending. Thanks to it, mask works more like typical
mask in graphics editor software.
Grayscale is calculated as an average value: A*(R+G+B)/3

@issue: #428

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -143,8 +143,9 @@ enum class TVG_EXPORT CompositeMethod
 {
     None = 0,     ///< No composition is applied.
     ClipPath,     ///< The intersection of the source and the target is determined and only the resulting pixels from the source are rendered.
-    AlphaMask,    ///< The pixels of the source and the target are alpha blended. As a result, only the part of the source, which intersects with the target is visible.
-    InvAlphaMask  ///< The pixels of the source and the complement to the target's pixels are alpha blended. As a result, only the part of the source which is not covered by the target is visible.
+    AlphaMask,    ///< The pixels of the source and the target are alpha blended. As a result, only the part of the source, which alpha intersects with the target is visible.
+    InvAlphaMask, ///< The pixels of the source and the complement to the target's pixels are alpha blended. As a result, only the part of the source which alpha is not covered by the target is visible.
+    LumaMask      ///< @BETA_API The source pixels are converted to the grayscale (luma value) and alpha blended with the target. As a result, only the part of the source, which intersects with the target is visible.
 };
 
 /**
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -235,6 +235,7 @@ struct SwImage
 struct SwBlender
 {
     uint32_t (*join)(uint8_t r, uint8_t g, uint8_t b, uint8_t a);
+    uint32_t (*lumaValue)(uint32_t c);
 };
 
 struct SwCompositor;
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -47,6 +47,18 @@ static inline uint32_t _ialpha(uint32_t c)
 }
 
 
+static inline uint32_t _abgrLumaValue(uint32_t c)
+{
+    return ((((c&0xff)*54) + (((c>>8)&0xff)*183) + (((c>>16)&0xff)*19))) >> 8; //0.2125*R + 0.7154*G + 0.0721*B
+}
+
+
+static inline uint32_t _argbLumaValue(uint32_t c)
+{
+    return ((((c&0xff)*19) + (((c>>8)&0xff)*183) + (((c>>16)&0xff)*54))) >> 8; //0.0721*B + 0.7154*G + 0.2125*R
+}
+
+
 static inline uint32_t _abgrJoin(uint8_t r, uint8_t g, uint8_t b, uint8_t a)
 {
     return (a << 24 | b << 16 | g << 8 | r);
@@ -139,7 +151,7 @@ static bool _rasterMaskedRect(SwSurface* surface, const SwBBox& region, uint32_t
     auto w = static_cast<uint32_t>(region.max.x - region.min.x);
     auto h = static_cast<uint32_t>(region.max.y - region.min.y);
 
-    auto cbuffer = surface->compositor->image.data + (region.min.y * surface->compositor->image.stride) + region.min.x;   //compositor buffer
+    auto cbuffer = surface->compositor->image.data + (region.min.y * surface->compositor->image.stride) + region.min.x; //compositor buffer
 
     for (uint32_t y = 0; y < h; ++y) {
         auto dst = &buffer[y * surface->stride];
@@ -173,6 +185,8 @@ static bool _rasterRect(SwSurface* surface, const SwBBox& region, uint32_t color
             return _rasterMaskedRect(surface, region, color, _alpha);
         } else if (surface->compositor->method == CompositeMethod::InvAlphaMask) {
             return _rasterMaskedRect(surface, region, color, _ialpha);
+        } else if (surface->compositor->method == CompositeMethod::LumaMask) {
+            return _rasterMaskedRect(surface, region, color, surface->blender.lumaValue);
         }
     } else {
         if (opacity == 255) {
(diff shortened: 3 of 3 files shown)


mmaciola:
Changed grayvalue calculation to the weighted method as (0.0721*B + 0.7154*G + 0.2125*R) * A

mgrudzinska:
@mmaciola sorry it takes so long...

how do you want to include this in the svg loader? there is a mask, that's all, no alpha mask or color mask

I thought that on an early stage of masking we could check the color values and convert them to the opacity value. Then we don't need all the changes here. but: 
it may be a wrong idea, since I see the comment from Hermet in #428 and it seems to be in agree with what you wrote :)

mmaciola:
Masking in svg loader needs some changes in terms of loading the mask into scene, not a shape, but in general, for new masking functionality, it is just changing the AlphaMask to ColorMask. 
ColorMask functionally will be useful for the whole thorvg, not just svg loader itself

mmaciola:
Rebased

mmaciola:
Todo: handle rgba colorspace

mmaciola:
Rebased

mgrudzinska (changes requested):
As I understand, we want to have two different kinds of masks, although I'm not sure what the use case for the alpha mask is, after we introduce the color mask. In svg files you have to define the fill of the mask, so for apis...?

Nevertheless please find my comments.

mgrudzinska:
right now we have 10 functions for the AlphaMask. 
The only difference for the ColorMask will be the line:
```auto tmp = ALPHA_BLEND(color, surface->blender.alpha(*cmp));```  ->  ```auto tmp = ALPHA_BLEND(color, _grayScale(*cmp));``` 
(I assumed that you will introduce the _grayScale function)
So I think that we should consider using funct pointers and pass the proper one to these 10 functions instead of copy them...

File: src/lib/sw_engine/tvgSwRaster.cpp [hunk 1]
```diff
@@ -176,6 +176,30 @@ static bool _translucentRectInvAlphaMask(SwSurface* surface, const SwBBox& regio
...
+
+    TVGLOG("SW_ENGINE", "Rectangle Color Mask Composition");
+
+    auto cbuffer = surface->compositor->image.data + (region.min.y * surface->stride) + region.min.x; //compositor buffer
+
+    for (uint32_t y = 0; y < h; ++y) {
+        auto dst = &buffer[y * surface->stride];
+        auto cmp = &cbuffer[y * surface->stride];
+        for (uint32_t x = 0; x < w; ++x) {
+            auto grayvalue = (((((*cmp)&0xff)*19) + ((((*cmp)>>8)&0xff)*183) + ((((*cmp)>>16)&0xff)*54)) * surface->blender.alpha(*cmp) + 0xffff) >> 16; //(0.0721*B + 0.7154*G + 0.2125*R) * A
```
mgrudzinska, line 194: rgb in cmp are already multiplied by alpha, so now you multiply it for the second time.

I'd intruduce a separate function like _grayScale that would return the value that you called grayvalue.

also coding convention:

- grayvalue -> grayValue
- spaces between numbers

hermet:
@mmaciola im really sorry for this but could you please rebase this patch?

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -47,6 +47,12 @@ static inline uint32_t _ialpha(uint32_t c)
 }
 
 
+static inline uint32_t _grayvalue(uint32_t c)
```
mgrudzinska, line 50: _grayvalue -> _grayValue

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -47,6 +47,12 @@ static inline uint32_t _ialpha(uint32_t c)
...
+{
+    return ((((c&0xff)*19) + (((c>>8)&0xff)*183) + (((c>>16)&0xff)*54)) * (c >> 24) + 0xffff) >> 16; //(0.0721*B + 0.7154*G + 0.2125*R) * A
+}
```
mgrudzinska, line 53: color is already alpha premultiplied, so you don't need to repeat that here

what if a color was build using _abgrJoin ?

mgrudzinska (changes requested):

src/lib/sw_engine/tvgSwRaster.cpp
mmaciola: thanks. fixed by introducing surface->blender.grayValue function

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -47,6 +47,18 @@ static inline uint32_t _ialpha(uint32_t c)
 }
 
 
+static inline uint32_t _abgrGrayValue(uint32_t c)
+{
+    return ((((c&0xff)*54) + (((c>>8)&0xff)*183) + (((c>>16)&0xff)*19)) + 0xff) >> 8; //0.2125*R + 0.7154*G + 0.0721*B
```
mgrudzinska, line 52: 0xff - I guess that 0xff is a factor similar to the one used in _multiplyAlpha? 

if so, then in this case it's not necessary. by using multiplication and bit shifting u are using approximation and the final 'gray value' is closer to `0.2125*R + 0.7154*G + 0.0721*B` without this factor (in 16% cases we get a value that is greater/smaller by 1, with the `0xff` factor in 92% of all cases the value is a bit different).

also I would add spaces here and in l.58

mgrudzinska (approved):
@hermet I left a comment, but since Michal is on vacation I'm ok with the current version and we can apply the small changes in a separate commit

inc/thorvg.h
```diff
@@ -143,8 +143,9 @@ enum class TVG_EXPORT CompositeMethod
 {
     None = 0,     ///< No composition is applied.
     ClipPath,     ///< The intersection of the source and the target is determined and only the resulting pixels from the source are rendered.
-    AlphaMask,    ///< The pixels of the source and the target are alpha blended. As a result, only the part of the source, which intersects with the target is visible.
-    InvAlphaMask  ///< The pixels of the source and the complement to the target's pixels are alpha blended. As a result, only the part of the source which is not covered by the target is visible.
+    AlphaMask,    ///< The pixels of the source and the target are alpha blended. As a result, only the part of the source, which alpha intersects with the target is visible.
+    InvAlphaMask, ///< The pixels of the source and the complement to the target's pixels are alpha blended. As a result, only the part of the source which alpha is not covered by the target is visible.
+    ColorMask     ///< The source pixels are converted to the grayscale and alpha blended with the target. As a result, only the part of the source, which intersects with the target is visible.
```
hermet, line 148: Please add @BETA_API tag to ColorMask

src/lib/sw_engine/tvgSwCommon.h
```diff
@@ -235,6 +235,7 @@ struct SwImage
 struct SwBlender
 {
     uint32_t (*join)(uint8_t r, uint8_t g, uint8_t b, uint8_t a);
+    uint32_t (*grayValue)(uint32_t c);
```
hermet, line 238: We don't need this function table in the blender.
tvg won't support multiple formats input.
So.. please call the inline function immediately such as _alpha(), _ialpha()

hermet (changes requested):
Thanks, please check few comments.

inc/thorvg.h
hermet: Suggestion: ColorMask =>  LumaMask
https://en.wikipedia.org/wiki/Luma_(video)

mmaciola: Renamed to LumaMask

src/lib/sw_engine/tvgSwCommon.h
mmaciola: As grayValue (lumaValue) uses color components, this function differs for ARGB and ABGR

src/lib/sw_engine/tvgSwRaster.cpp
hermet: @mmaciola could you please confirm this? (resolve it if it's done)

hermet (approved):
Great, thanks! One more, please double-check the @mgrudzinska 's comment.

src/lib/sw_engine/tvgSwRaster.cpp
mmaciola: Removed 0xff factor.
I attach the table with comparison.
Luma calc compare

PR #622, opened 2021-07-16
svg_loader: allow both clipPath and mask together

mmaciola:
`node->style->comp` changed to `node->style->clipPath` and `node->style->mask`.

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -176,38 +176,69 @@ static bool _appendChildShape(SvgNode* node, Shape* shape, float vx, float vy, f
 }
 
 
-static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float vy, float vw, float vh)
+static void _applyClipPathComposition(Paint* paint, const SvgNode* node, float vx, float vy, float vw, float vh)
 {
-    if (node->style->comp.method == CompositeMethod::None) return;
-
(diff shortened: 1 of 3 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -176,38 +176,69 @@ static bool _appendChildShape(SvgNode* node, Shape* shape, float vx, float vy, f
 }
 
 
-static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float vy, float vw, float vh)
+static void _applyClipPathComposition(Paint* paint, const SvgNode* node, float vx, float vy, float vw, float vh)
```
JSUYA, line 179: I think it can be handled with one methid.
like this
`static void _applyComposition(Paint* paint, SvgComposite& comp, Matrix* transform, CompositeMethod method, float vx, float vy, float vw, float vh)`
how about this?

JSUYA (commented):
Please check my comment

src/loaders/svg/tvgSvgSceneBuilder.cpp
mmaciola: Please find the changes

JSUYA: oh my opinion was to reduce duplicate code by passing method type and SvgComposite as parameters.
The patch is working good. so I'll merge this PR and refactor it later

JSUYA (approved):

PR #628, opened 2021-07-20
sw_engine Raster: Apply bilinear interpolation to images

JSUYA:
Apply bilinear interpolation when drawing images with transformation
If the mapped coordinate value is a floating point value,
bilinear interpolation is performed using adjacent pixel values.
Interpolation is not performed in cases when the color values to be interpolated
are the same or scale down case.

issue : https://github.com/Samsung/thorvg/issues/373

[left before right after]
image

image
(scale down case is same result)

image
(scale up x3)

Screenshot from 2021-07-21 10-56-39
(Translucent Image with clippath)

Screenshot from 2021-07-21 10-59-27
(Image with clippath)

Screenshot from 2021-07-21 11-04-25
(Translucent Image)

Screenshot from 2021-07-21 11-07-37_mask
(Translucent Image with mask)

Screenshot from 2021-07-21 11-09-58_invmask
(Translucent Image with inverse mask)

### DIFF
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -282,6 +282,21 @@ static inline uint32_t COLOR_INTERPOLATE(uint32_t c1, uint32_t a1, uint32_t c2, 
     return (c1 |= t);
 }
 
+//c1(x,y)--a1-- c2
+// |        |    |
+// |        |    |
+//a2---result---a2
+// |        |    |
+// |        |    |
+//c4 ------a1-- c3
+static inline uint32_t COLOR_BILINEAR_INTERPOLATE(uint32_t c1, uint32_t c2, uint32_t c3, uint32_t c4, uint32_t a1, uint32_t a2)
+{
+    return COLOR_INTERPOLATE(COLOR_INTERPOLATE(c1, 255 - a1, c2, a1),
+                             255 - a2,
+                             COLOR_INTERPOLATE(c4, 255 - a1, c3, a1),
+                             a2);
+}
+
 static inline uint32_t ALPHA_MULTIPLY(uint32_t c, uint32_t a)
 {
     return ((c * a + 0xff) >> 8);
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -282,18 +282,6 @@ static inline uint32_t COLOR_INTERPOLATE(uint32_t c1, uint32_t a1, uint32_t c2, 
     return (c1 |= t);
 }
 
-//c1(x,y)--a1-- c2
-// |        |    |
-// |        |    |
-//a2---result---a2
-// |        |    |
-// |        |    |
-//c4 ------a1-- c3
-static inline uint32_t COLOR_BILINEAR_INTERPOLATE(uint32_t c1, uint32_t c2, uint32_t c3, uint32_t c4, uint32_t a1, uint32_t a2)
-{
-    return COLOR_INTERPOLATE(COLOR_INTERPOLATE(c1, 255 - a1, c2, a1), 255 - a2, COLOR_INTERPOLATE(c4, 255 - a1, c3, a1), a2);
-}
-
 static inline uint32_t ALPHA_MULTIPLY(uint32_t c, uint32_t a)
 {
     return ((c * a + 0xff) >> 8);
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -291,10 +291,7 @@ static inline uint32_t COLOR_INTERPOLATE(uint32_t c1, uint32_t a1, uint32_t c2, 
 //c4 ------a1-- c3
 static inline uint32_t COLOR_BILINEAR_INTERPOLATE(uint32_t c1, uint32_t c2, uint32_t c3, uint32_t c4, uint32_t a1, uint32_t a2)
 {
-    return COLOR_INTERPOLATE(COLOR_INTERPOLATE(c1, 255 - a1, c2, a1),
-                             255 - a2,
-                             COLOR_INTERPOLATE(c4, 255 - a1, c3, a1),
-                             a2);
+    return COLOR_INTERPOLATE(COLOR_INTERPOLATE(c1, 255 - a1, c2, a1), 255 - a2, COLOR_INTERPOLATE(c4, 255 - a1, c3, a1), a2);
 }
 
 static inline uint32_t ALPHA_MULTIPLY(uint32_t c, uint32_t a)
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -106,10 +106,8 @@ static uint32_t _applyBilinearInterpolation(const uint32_t *img, uint32_t w, uin
     auto c3 = img[(rX + 1) + ((rY + 1) * w)];
     auto c4 = img[rX + ((rY + 1) * w)];
 
-    if (c1 == c2 && c1 == c3 && c1 == c4) {
-        return img[rX + (rY * w)];
-    }
-    return COLOR_BILINEAR_INTERPOLATE(c1, c2, c3, c4, dX, dY);
+    if (c1 == c2 && c1 == c3 && c1 == c4) return img[rX + (rY * w)];
+    return COLOR_INTERPOLATE(COLOR_INTERPOLATE(c1, 255 - dX, c2, dX), 255 - dY, COLOR_INTERPOLATE(c4, 255 - dX, c3, dX), dY);
 }
 
 
@@ -338,7 +336,26 @@ static bool _rasterTranslucentImageRle(SwSurface* surface, const SwRleData* rle,
 static bool _rasterTranslucentImageRle(SwSurface* surface, const SwRleData* rle, uint32_t *img, uint32_t w, uint32_t h, uint32_t opacity, const Matrix* invTransform)
 {
     auto span = rle->spans;
-    bool isDownScaling = (invTransform->e11 * invTransform->e11) + (invTransform->e21 * invTransform->e21) > 1 ? true : false;
+
+    for (uint32_t i = 0; i < rle->size; ++i, ++span) {
+        auto ey1 = span->y * invTransform->e12 + invTransform->e13;
+        auto ey2 = span->y * invTransform->e22 + invTransform->e23;
+        auto dst = &surface->buffer[span->y * surface->stride + span->x];
+        auto alpha = ALPHA_MULTIPLY(span->coverage, opacity);
+        for (uint32_t x = 0; x < span->len; ++x, ++dst) {
+            auto rX = static_cast<uint32_t>(roundf((span->x + x) * invTransform->e11 + ey1));
+            auto rY = static_cast<uint32_t>(roundf((span->x + x) * invTransform->e21 + ey2));
+            if (rX >= w || rY >= h) continue;
+            auto src = ALPHA_BLEND(img[rY * w + rX], alpha);     //TODO: need to use image's stride
+            *dst = src + ALPHA_BLEND(*dst, 255 - surface->blender.alpha(src));
+        }
+    }
+    return true;
+}
+
+static bool _rasterTranslucentUpScaleImageRle(SwSurface* surface, const SwRleData* rle, uint32_t *img, uint32_t w, uint32_t h, uint32_t opacity, const Matrix* invTransform)
+{
+    auto span = rle->spans;
     for (uint32_t i = 0; i < rle->size; ++i, ++span) {
         auto ey1 = span->y * invTransform->e12 + invTransform->e13;
         auto ey2 = span->y * invTransform->e22 + invTransform->e23;
@@ -351,11 +368,8 @@ static bool _rasterTranslucentImageRle(SwSurface* surface, const SwRleData* rle,
             auto rY = static_cast<uint32_t>(roundf(fY));
             if (rX >= w || rY >= h) continue;
             uint32_t src;
-            if (isDownScaling || rX == w - 1 || rY == h - 1) {
-                src = ALPHA_BLEND(img[rY * w + rX], alpha);     //TODO: need to use image's stride
-            } else {
-                src = ALPHA_BLEND(_applyBilinearInterpolation(img, w, h, fX, fY), alpha);     //TODO: need to use image's stride
-            }
+            if (rX == w - 1 || rY == h - 1) src = ALPHA_BLEND(img[rY * w + rX], alpha);     //TODO: need to use image's stride
+            else src = ALPHA_BLEND(_applyBilinearInterpolation(img, w, h, fX, fY), alpha);     //TODO: need to use image's stride
             *dst = src + ALPHA_BLEND(*dst, 255 - surface->blender.alpha(src));
         }
     }
@@ -382,7 +396,26 @@ static bool _rasterImageRle(SwSurface* surface, SwRleData* rle, uint32_t *img, u
 static bool _rasterImageRle(SwSurface* surface, SwRleData* rle, uint32_t *img, uint32_t w, uint32_t h, const Matrix* invTransform)
 {
     auto span = rle->spans;
-    bool isDownScaling = (invTransform->e11 * invTransform->e11) + (invTransform->e21 * invTransform->e21) > 1 ? true : false;
+
+    for (uint32_t i = 0; i < rle->size; ++i, ++span) {
+        auto ey1 = span->y * invTransform->e12 + invTransform->e13;
+        auto ey2 = span->y * invTransform->e22 + invTransform->e23;
+        auto dst = &surface->buffer[span->y * surface->stride + span->x];
+        for (uint32_t x = 0; x < span->len; ++x, ++dst) {
+            auto rX = static_cast<uint32_t>(roundf((span->x + x) * invTransform->e11 + ey1));
+            auto rY = static_cast<uint32_t>(roundf((span->x + x) * invTransform->e21 + ey2));
+            if (rX >= w || rY >= h) continue;
+            auto src = ALPHA_BLEND(img[rY * w + rX], span->coverage);    //TODO: need to use image's stride
+            *dst = src + ALPHA_BLEND(*dst, 255 - surface->blender.alpha(src));
+        }
+    }
+    return true;
+}
+
+
+static bool _rasterUpScaleImageRle(SwSurface* surface, SwRleData* rle, uint32_t *img, uint32_t w, uint32_t h, const Matrix* invTransform)
+{
+    auto span = rle->spans;
 
     for (uint32_t i = 0; i < rle->size; ++i, ++span) {
         auto ey1 = span->y * invTransform->e12 + invTransform->e13;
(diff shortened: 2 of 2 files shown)


hermet:
@mgrudzinska could you please have a look?

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -573,6 +573,66 @@ static bool _rasterImage(SwSurface* surface, uint32_t *img, uint32_t w, TVG_UNUS
     return true;
 }
 
+static uint32_t _applyBilinearInterpolation(const uint32_t *img, uint32_t w, uint32_t h, uint32_t rX, uint32_t rY, float fX, float fY)
+{
+    auto dX = fX - rX;
```
mgrudzinska, line 578: and what do you think of using a floor() function instead of passing rX and rY?
I didn't check the code below, since I wanted to answer fast, but now I will :)

```
auto rX = static_cast<uint32_t>(floor(fX));
auto rY = static_cast<uint32_t>(floor(fY));

auto dX = fX - rX;
auto dY = fY - rY;

//0.0039f is a condition for the interpolation ratio to be at least 1.
if (abs(dX) > 0.0039f && abs(dY) > 0.0039f) {
        float w1 = dX;
        float h1 = dY;
        uint32_t p1, p2, p3, p4;

        p1 = img[rX + rY  * w];
        p2 = img[rX + 1 + rY * w];
        p3 = img[rX + 1 + (rY + 1) * w];
        p4 = img[rX + (rY + 1) * w];

        if (p1 == p2 && p1 == p3 && p1 == p4) {
            return img[rX + (rY * w)];
        } else {
            auto c1 = COLOR_INTERPOLATE(p1, static_cast<uint32_t>((1 - h1) * 255), p4, static_cast<uint32_t>(h1 * 255));
            auto c2 = COLOR_INTERPOLATE(p2, static_cast<uint32_t>((1 - h1) * 255), p3, static_cast<uint32_t>(h1 * 255));
            return COLOR_INTERPOLATE(c1, static_cast<uint32_t>((1 - w1) * 255), c2, static_cast<uint32_t>(w1 * 255));
        }
}
return img[rX + (rY * w)];
```

mgrudzinska: what I wrote seems to be ok. 
I named the points as follows:
p1 -- p2
 |        |
p4 -- p3

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -573,6 +573,66 @@ static bool _rasterImage(SwSurface* surface, uint32_t *img, uint32_t w, TVG_UNUS
...
+    auto dY = fY - rY;
+
+    //0.0039f is a condition for the interpolation ratio to be at least 1.
+    if (abs(dX) > 0.0039f && abs(dY) > 0.0039f) {
```
mgrudzinska, line 582: also because of this condition there are some small artifacts. I think that in case e.x. dX < 0.0039f we could still interpolate in the y dimension. what do you think?

src/lib/sw_engine/tvgSwRaster.cpp
mmaciola: I like the above. I suggest removing w1 and h1 (use dX and dY directly).
In COLOR_INTERPOLATE I am a little suspicious that a1 + a2 won't sum up 255 (but 254). Maybe we can introduce function like:
```
static inline uint32_t COLOR_INTERPOLATE_2(uint32_t c1, uint32_t c2, uint8_t a)
{
    return COLOR_INTERPOLATE(c1, a, c2, 255-a);
}
```
It will reduce one float multiplication and casting and ensure a1+a2=255. What do you think?

mmaciola: Maybe we can also drop floor and use just `static_cast<uint32_t>(fX)`.
`floor` may be slow (we call it multiple times) and not vectorize as SIMD.

JSUYA: @mgrudzinska @mmaciola 
Based on review comments, I've modified the code to be simpler. As a result, unnecessary if-checks and calculations are reduced. Thank you for review.

src/lib/sw_engine/tvgSwRaster.cpp
JSUYA: I deleted this if-check.
It seems rather unnecessary to check every pixel for this condition.

mmaciola (approved):

src/lib/sw_engine/tvgSwCommon.h
```diff
@@ -278,6 +278,21 @@ static inline uint32_t COLOR_INTERPOLATE(uint32_t c1, uint32_t a1, uint32_t c2,
...
+//a2---result---a2
+// |        |    |
+// |        |    |
+//c4 ------a1-- c3
+static inline uint32_t COLOR_BILINEAR_INTERPOLATE(uint32_t c1, uint32_t c2, uint32_t c3, uint32_t c4, uint32_t a1, uint32_t a2)
+{
+    return COLOR_INTERPOLATE(COLOR_INTERPOLATE(c1, 255 - a1, c2, a1),
+                             255 - a2,
+                             COLOR_INTERPOLATE(c4, 255 - a1, c3, a1),
+                             a2);
```
mgrudzinska, line 293: I'm not sure about the format rules that we have - should we use one line? or can we split?
also the comment is a bit confusing for me, but as you wish :)

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -573,25 +627,30 @@ static bool _rasterImage(SwSurface* surface, uint32_t *img, uint32_t w, TVG_UNUS
...
-            auto rX = static_cast<uint32_t>(roundf(x * invTransform->e11 + ey1));
-            auto rY = static_cast<uint32_t>(roundf(x * invTransform->e21 + ey2));
+            auto fX = x * invTransform->e11 + ey1;
+            auto fY = x * invTransform->e21 + ey2;
+            auto rX = static_cast<uint32_t>(roundf(fX));
+            auto rY = static_cast<uint32_t>(roundf(fY));
             if (rX >= w || rY >= h) continue;
-            auto src = img[rX + (rY * w)];    //TODO: need to use image's stride
+            uint32_t src;
+            if (rX == w - 1 || rY == h - 1 || invTransform->e11 < 0 || invTransform->e22 < 0) {
```
mgrudzinska, line 643: I understand this part ```rX == w - 1 || rY == h - 1```, since it's obvious, but this one:
```invTransform->e11 < 0 || invTransform->e22 < 0``` ?
earlier you've made a comment, that it has something to do with downscaling, but it doesn't. not in this form. 
the result below is for the pict transformed as follows: 
```
        picture->translate(600, 0);
        picture->rotate(91);   // <- cos(91) < 0
```
pyt

on the right I removed the condition ```invTransform->e11 < 0 || invTransform->e22 < 0```

**Review by mgrudzinska — CHANGES_REQUESTED** [review]

File: src/lib/sw_engine/tvgSwRaster.cpp (hunk 4 above)
Comment (mgrudzinska) [reply src/lib/sw_engine/tvgSwRaster.cpp:643]: for sx = sy we could use ```e11^2 + e21^2 < 1``` condition to check whether it's downscaling

File: src/lib/sw_engine/tvgSwCommon.h (hunk 3 above)
Comment (JSUYA) [reply src/lib/sw_engine/tvgSwCommon.h:293]: I modified the code. One line is recommended. But... for readability, I think newlines can be allowed. (if the code line is toooooo long)

File: src/lib/sw_engine/tvgSwRaster.cpp [hunk 5]
```diff
@@ -573,25 +627,30 @@ static bool _rasterImage(SwSurface* surface, uint32_t *img, uint32_t w, TVG_UNUS
...
-            auto rX = static_cast<uint32_t>(roundf(x * invTransform->e11 + ey1));
-            auto rY = static_cast<uint32_t>(roundf(x * invTransform->e21 + ey2));
+            auto fX = x * invTransform->e11 + ey1;
+            auto fY = x * invTransform->e21 + ey2;
+            auto rX = static_cast<uint32_t>(roundf(fX));
+            auto rY = static_cast<uint32_t>(roundf(fY));
             if (rX >= w || rY >= h) continue;
-            auto src = img[rX + (rY * w)];    //TODO: need to use image's stride
+            uint32_t src;
+            if (rX == w - 1 || rY == h - 1 || (invTransform->e11 * invTransform->e11) + (invTransform->e21 * invTransform->e21) > 1) {
```
mgrudzinska, line 643: @JSUYA I'm a bit worried about the performance. 
for each pixel we do this check and for most of the cases the first two conditions are not fulfilled, so the "invTransform..." condition is checked. 
Do you think, that it would be a good idea to check, whether it's a downscaling or not, a bit earlier? this way we would duplicate some code, I know... 
So it's rather a question than a suggestion. 

If you think it's pointless I'll accept and you can merge (or wait for other reviewers)

JSUYA: Thanks for the review. Good point. I added a patch. If you have any better ideas please share me

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -325,17 +345,24 @@ static bool _rasterTranslucentImageRle(SwSurface* surface, const SwRleData* rle,
 static bool _rasterTranslucentImageRle(SwSurface* surface, const SwRleData* rle, uint32_t *img, uint32_t w, uint32_t h, uint32_t opacity, const Matrix* invTransform)
 {
     auto span = rle->spans;
-
+    bool isDownScaling = (invTransform->e11 * invTransform->e11) + (invTransform->e21 * invTransform->e21) > 1 ? true : false;
```
mgrudzinska, line 348: yes, I also thought about using a bool :)

and like below? 
I know that a huge amount of the code is duplicated. I'm not sure whether it's worth of it. 
what do you think? 

```
if (isDownScaling) {
    for (uint32_t i = 0; i < rle->size; ++i, ++span) {
        auto ey1 = span->y * invTransform->e12 + invTransform->e13;
        auto ey2 = span->y * invTransform->e22 + invTransform->e23;
        auto dst = &surface->buffer[span->y * surface->stride + span->x];
        auto alpha = ALPHA_MULTIPLY(span->coverage, opacity);
        for (uint32_t x = 0; x < span->len; ++x, ++dst) {
            auto rX = static_cast<uint32_t>(roundf((span->x + x) * invTransform->e11 + ey1));
            auto rY = static_cast<uint32_t>(roundf((span->x + x) * invTransform->e21 + ey2));
            if (rX >= w || rY >= h) continue;
            auto src = ALPHA_BLEND(img[rY * w + rX], alpha);     //TODO: need to use image's stride
            *dst = src + ALPHA_BLEND(*dst, 255 - surface->blender.alpha(src));
        }
    }
} else {
    for (uint32_t i = 0; i < rle->size; ++i, ++span) {
        auto ey1 = span->y * invTransform->e12 + invTransform->e13;
        auto ey2 = span->y * invTransform->e22 + invTransform->e23;
        auto dst = &surface->buffer[span->y * surface->stride + span->x];
        auto alpha = ALPHA_MULTIPLY(span->coverage, opacity);
        for (uint32_t x = 0; x < span->len; ++x, ++dst) {
            auto fX = (span->x + x) * invTransform->e11 + ey1;
            auto fY = (span->x + x) * invTransform->e21 + ey2;
            auto rX = static_cast<uint32_t>(fX);
            auto rY = static_cast<uint32_t>(fY);
            if (rX >= w || rY >= h) continue;
            uint32_t src;
            if (rX == w - 1 || rY == h - 1) {
                src = ALPHA_BLEND(img[rY * w + rX], alpha);     //TODO: need to use image's stride
            } else {
                src = ALPHA_BLEND(_applyBilinearInterpolation(img, w, h, fX, fY), alpha);     //TODO: need to use image's stride
            }
            *dst = src + ALPHA_BLEND(*dst, 255 - surface->blender.alpha(src));
        }
    }
}
```

also please notice, that in the `else` block I removed a floor function when rX and rY are calculated.

JSUYA: Thank you for your review. I have split the raster function as per your comment. So I added the appropriate functions.

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -373,6 +419,33 @@ static bool _rasterImageRle(SwSurface* surface, SwRleData* rle, uint32_t *img, u
 }
 
 
+static bool _rasterUpScaleImageRle(SwSurface* surface, SwRleData* rle, uint32_t *img, uint32_t w, uint32_t h, const Matrix* invTransform)
+{
+    auto span = rle->spans;
+    bool isDownScaling = (invTransform->e11 * invTransform->e11) + (invTransform->e21 * invTransform->e21) > 1 ? true : false;
```
mgrudzinska, line 425: remove

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -373,6 +419,33 @@ static bool _rasterImageRle(SwSurface* surface, SwRleData* rle, uint32_t *img, u
...
+        auto ey2 = span->y * invTransform->e22 + invTransform->e23;
+        auto dst = &surface->buffer[span->y * surface->stride + span->x];
+        for (uint32_t x = 0; x < span->len; ++x, ++dst) {
+            auto fX = (span->x + x) * invTransform->e11 + ey1;
+            auto fY = (span->x + x) * invTransform->e21 + ey2;
+            auto rX = static_cast<uint32_t>(roundf(fX));
+            auto rY = static_cast<uint32_t>(roundf(fY));
+            if (rX >= w || rY >= h) continue;
+            uint32_t src;
+            if (isDownScaling || rX == w - 1 || rY == h - 1) {
```
mgrudzinska, line 437: do you need to check `isDownScaling`?

mgrudzinska, line 437: do you need to check `isDownScaling`?

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -458,6 +532,114 @@ static bool _rasterTranslucentImage(SwSurface* surface, const uint32_t *img, uin
...
+    }
+    return true;
+}
+
+
+static bool _rasterTranslucentUpScaleImage(SwSurface* surface, const uint32_t *img, uint32_t w, uint32_t h, uint32_t opacity, const SwBBox& region, const Matrix* invTransform)
+{
+    if (surface->compositor) {
+        if (surface->compositor->method == CompositeMethod::AlphaMask) {
+            return _translucentImageAlphaMaskUpScale(surface, img, w, h, opacity, region, invTransform);
```
mgrudzinska, line 633: I think the naming convention would be `_translucentUpScaleImageAlphaMask`, but as you wish

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -1203,7 +1411,11 @@ bool rasterImage(SwSurface* surface, SwImage* image, const Matrix* transform, co
             if (translucent) return _rasterTranslucentImageRle(surface, image->rle, image->data, image->w, image->h, opacity);
             return _rasterImageRle(surface, image->rle, image->data, image->w, image->h);
         } else {
-            if (translucent) return _rasterTranslucentImageRle(surface, image->rle, image->data, image->w, image->h, opacity, &invTransform);
+            if (translucent) {
+                if (isUpScaling) return _rasterTranslucentImageRle(surface, image->rle, image->data, image->w, image->h, opacity, &invTransform);
+                return _rasterTranslucentUpScaleImageRle(surface, image->rle, image->data, image->w, image->h, opacity, &invTransform);
```
mgrudzinska, line 1416: the other way around:

```
if (isUpScaling) return _rasterTranslucentUpScaleImageRle(...)
return _rasterTranslucentImageRle(...)
```

mgrudzinska (changes requested):
@JSUYA please check the comments

src/lib/sw_engine/tvgSwRaster.cpp
JSUYA: Removed it.

src/lib/sw_engine/tvgSwRaster.cpp
JSUYA: Removed it.

JSUYA: Removed it.

src/lib/sw_engine/tvgSwRaster.cpp
JSUYA: Fixed. Thanks for checking in detail :)

src/lib/sw_engine/tvgSwRaster.cpp
JSUYA: Changed.

JSUYA:
Scale down case, bilinear Interpolation should be applied in a different way from the scale up case. This feature will make another PR.

mgrudzinska (approved):

hermet (changes requested):
Vertical scale up but horizontal scale down, which one will be better ScaleUp? ScaleDown?
Otherwise vise versa scenario?

I think the transformed image interpolation is not proper with this approach.
Only orthogonal scaled image does.

In the end, we need the polygon based texture mapping for support arbitrary transformed image interpolation....

Nevermind this right now, 
Please check nitpicks.

src/lib/sw_engine/tvgSwCommon.h
```diff
@@ -278,6 +278,18 @@ static inline uint32_t COLOR_INTERPOLATE(uint32_t c1, uint32_t a1, uint32_t c2,
...
 }
 
+//c1(x,y)--a1-- c2
+// |        |    |
+// |        |    |
+//a2---result---a2
+// |        |    |
+// |        |    |
+//c4 ------a1-- c3
+static inline uint32_t COLOR_BILINEAR_INTERPOLATE(uint32_t c1, uint32_t c2, uint32_t c3, uint32_t c4, uint32_t a1, uint32_t a2)
```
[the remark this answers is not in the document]
JSUYA: removed this function

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -336,6 +356,32 @@ static bool _rasterTranslucentImageRle(SwSurface* surface, const SwRleData* rle,
...
+        auto dst = &surface->buffer[span->y * surface->stride + span->x];
+        auto alpha = ALPHA_MULTIPLY(span->coverage, opacity);
+        for (uint32_t x = 0; x < span->len; ++x, ++dst) {
+            auto fX = (span->x + x) * invTransform->e11 + ey1;
+            auto fY = (span->x + x) * invTransform->e21 + ey2;
+            auto rX = static_cast<uint32_t>(roundf(fX));
+            auto rY = static_cast<uint32_t>(roundf(fY));
+            if (rX >= w || rY >= h) continue;
+            uint32_t src;
+            if (rX == w - 1 || rY == h - 1) {
```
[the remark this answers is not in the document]
JSUYA: modified

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -94,6 +94,26 @@ static bool _translucent(const SwSurface* surface, uint8_t a)
...
+
+    auto dX = static_cast<uint32_t>((fX - rX) * 255.0);
+    auto dY = static_cast<uint32_t>((fY - rY) * 255.0);
+
+    auto c1 = img[rX + (rY * w)];
+    auto c2 = img[(rX + 1) + (rY * w)];
+    auto c3 = img[(rX + 1) + ((rY + 1) * w)];
+    auto c4 = img[rX + ((rY + 1) * w)];
+
+    if (c1 == c2 && c1 == c3 && c1 == c4) {
```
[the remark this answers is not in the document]
JSUYA: modified

hermet (approved):
Great job!

JSUYA:
> Vertical scale up but horizontal scale down, which one will be better ScaleUp? ScaleDown?
> Otherwise vise versa scenario?
> 
> I think the transformed image interpolation is not proper with this approach.
> Only orthogonal scaled image does.
> 
> In the end, we need the polygon based texture mapping for support arbitrary transformed image interpolation....
> 
> Nevermind this right now,
> Please check nitpicks.

As you said, I didn't consider the case of scale up for x and y respectively.
I'll fix this by creating a new PR.

PR #636, opened 2021-07-22
apis: promote beta apis to the official ones.

hermet:
3 api candidates has been verified since it's tagged in beta,
we confirm that they are useful for tvg usages.

* Modified: 
removed composite() getter api

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -247,7 +247,7 @@ public:
      *
      * @retval The augmented transformation matrix.
      *
-     * @BETA_API
+     * @since 0.4
      */
     Matrix transform() noexcept;
 
@@ -1185,8 +1185,7 @@ public:
 
     /**
      * @brief Enumeration specifying the methods of Memory Pool behavior policy.
-     *
-     * @BETA_API
+     * @since 0.4
      */
     enum MempoolPolicy
     {
@@ -1236,7 +1235,7 @@ public:
      *
      * @warning It's not allowed after pushing any paints.
      *
-     * @BETA_API
+     * @since 0.4
     */
     Result mempool(MempoolPolicy policy) noexcept;
 


hermet:
@mmaciola @JSUYA @mihashco @mgrudzinska We're gonna release some apis, please have a look and leave any opinions if you have. (agree/reject/etc)

inc/thorvg.h
```diff
@@ -309,7 +309,7 @@ class TVG_EXPORT Paint
      *
      * @return The method used to composite the source object with the target.
      *
-     * @BETA_API
+     * @since 0.4
      */
     CompositeMethod composite(const Paint** target) const noexcept;
```
JSUYA, line 314: A few issues with composites remain.
I think it is right to open after these are resolved.
https://github.com/Samsung/thorvg/issues/524
https://github.com/Samsung/thorvg/issues/513
https://github.com/Samsung/thorvg/issues/598

hermet: @JSUYA  Though those are about the setter api which is already in the official, we don't need to rush the release. so removed this one in the candidate.

mihashco:
@hermet  agree, CAPI module is tested and usable. It is good idea to change its status.

inc/thorvg.h
```diff
@@ -1236,7 +1235,7 @@ class TVG_EXPORT SwCanvas final : public Canvas
      *
      * @warning It's not allowed after pushing any paints.
      *
-     * @BETA_API
```
mgrudzinska, line 1239: `    * @param[in] policy Use the shared cache memory. The default value is @c true`
It's confusing, since 'policy' is an enum, not a bool

` * @retval Result::InsufficientCondition If the canvas has any paints.`
any -> no

inc/thorvg.h
mgrudzinska: @hermet since we are on the subject: should I add composite getter to the capi?

mgrudzinska:
I commented on the docs, but besides that I agree with publication

inc/thorvg.h
hermet: @mgrudzinska good catch, could you please update the doc with a separate patch?  @c true is a typo, it can be changed to the default enum value.

PR #641, opened 2021-07-22
sw_engine clipping: fix segfault after scene used for ClipPath

mmaciola:
ClipPath doesn't work with scene. This is a temporary solution that fixes
segfault if scene is applied as clippath composition anyway.


mmaciola:
Created a full support patch, so closing this temporary one solution

PR #643, opened 2021-07-22
sw_engine clipping: allow scene for ClipPath

mmaciola:
This patch changes tvgSceneImpl::update behaviour to return an array of
paints, allowing to use scene for ClipPath.
@issue: https://github.com/Samsung/thorvg/issues/524


mmaciola:
Converted to draft. ClipPath of scene is an intersection, not an union as expected. Need further work.

PR #647, opened 2021-07-22
tvg_saver: removing saving the default values

mgrudzinska:
The default values (stroke cap/join, fill rule or color with alpha = 0)
were unnecessarily saved in the tvg format file.

### DIFF
--- a/src/savers/tvg/tvgTvgSaver.cpp
+++ b/src/savers/tvg/tvgTvgSaver.cpp
@@ -127,7 +127,7 @@ TvgBinCounter TvgSaver::writeTagProperty(TvgBinTag tag, TvgBinCounter cnt, const
     return growCnt;
 }
 
- 
+
 TvgBinCounter TvgSaver::serializePaint(const Paint* paint)
 {
     TvgBinCounter cnt = 0;
@@ -193,8 +193,8 @@ TvgBinCounter TvgSaver::serializeFill(const Fill* fill, TvgBinTag tag)
         cnt += writeTagProperty(TVG_TAG_FILL_LINEAR_GRADIENT, sizeof(args), args);
     }
 
-    auto flag = static_cast<TvgBinFlag>(fill->spread());
-    cnt += writeTagProperty(TVG_TAG_FILL_FILLSPREAD, SIZE(TvgBinFlag), &flag);
+    if (auto flag = static_cast<TvgBinFlag>(fill->spread()))
+        cnt += writeTagProperty(TVG_TAG_FILL_FILLSPREAD, SIZE(TvgBinFlag), &flag);
     cnt += writeTagProperty(TVG_TAG_FILL_COLORSTOPS, stopsCnt * sizeof(stops), stops);
 
     writeReservedCount(cnt);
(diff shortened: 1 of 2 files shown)


src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -203,31 +203,28 @@ TvgBinCounter TvgSaver::serializeFill(const Fill* fill, TvgBinTag tag)
 }
 
 
-TvgBinCounter TvgSaver::serializeStroke(const Shape* shape)
+TvgBinCounter TvgSaver::serializeStroke(const Shape* shape, const Fill* fill, uint8_t color[4])
 {
```
mgrudzinska, line 207: I didn't want to call `serializeStroke` if it not necessary.
In the svg the default value of the stroke width is 1, so we need to check the color alpha and the fill (l.287 and l.288)
so to prevent calling the same functions twice, I passed the fill and color values as paramters.

src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -203,31 +203,28 @@ TvgBinCounter TvgSaver::serializeFill(const Fill* fill, TvgBinTag tag)
...
     reserveCount();
 
+    //width
+    auto width = shape->strokeWidth();
+    auto cnt = writeTagProperty(TVG_TAG_SHAPE_STROKE_WIDTH, sizeof(width), &width);
+
     //cap
     auto flag = static_cast<TvgBinFlag>(shape->strokeCap());
-    auto cnt = writeTagProperty(TVG_TAG_SHAPE_STROKE_CAP, SIZE(TvgBinFlag), &flag);
+    if (flag != TVG_FLAG_SHAPE_STROKE_CAP_BUTT)
```
hermet, line 217: Shouldn't be the following like TVG_TAG_FILL_FILLSPREAD?
if (flag) cnt += writeTagProperty(TVG_TAG_SHAPE_STROKE_CAP, SIZE(TvgBinFlag), &flag);

src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -276,21 +273,29 @@ TvgBinCounter TvgSaver::serializeShape(const Shape* shape)
 {
     writeTag(TVG_TAG_CLASS_SHAPE);
     reserveCount();
+    TvgBinCounter cnt = 0;
 
     //fill rule
-    auto flag = (shape->fillRule() == FillRule::EvenOdd) ? TVG_FLAG_SHAPE_FILLRULE_EVENODD : TVG_FLAG_SHAPE_FILLRULE_WINDING;
-    auto cnt = writeTagProperty(TVG_TAG_SHAPE_FILLRULE, SIZE(TvgBinFlag), &flag);
+    if (shape->fillRule() == FillRule::EvenOdd) {
```
hermet, line 279: Fixed condition.... it needs a comment just in case if the FillRule is expanded... So that we can easily notice this to fix.

src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -203,31 +203,28 @@ TvgBinCounter TvgSaver::serializeFill(const Fill* fill, TvgBinTag tag)
...
+        cnt += writeTagProperty(TVG_TAG_SHAPE_STROKE_CAP, SIZE(TvgBinFlag), &flag);
 
     //join
     flag = static_cast<TvgBinFlag>(shape->strokeJoin());
-    cnt += writeTagProperty(TVG_TAG_SHAPE_STROKE_JOIN, SIZE(TvgBinFlag), &flag);
-
-    //width
-    auto width = shape->strokeWidth();
-    cnt += writeTagProperty(TVG_TAG_SHAPE_STROKE_WIDTH, sizeof(width), &width);
+    if (flag != TVG_FLAG_SHAPE_STROKE_JOIN_MITER)
```
hermet, line 222: Same as STROKE_CAP

src/savers/tvg/tvgTvgSaver.cpp
hermet: @mgrudzinska i don't much like this because it looks more complex... in the end, it's costless.

hermet (changes requested):
Please check comments.

src/savers/tvg/tvgTvgSaver.cpp
mgrudzinska: @hermet I couldn't do it like that, since there is a problem...
The default values of the cap and join are respectively:
- in the TVG apis: `Square` (enum = 0) and `Bevel` (enum = 0)
- in the SVG standard: 'Butt' (enum = 2) and `Miter` (enum = 2)    -> source: https://www.w3.org/TR/SVG2/painting.html

I think that it's a bit late to change the order of enums, so I had to write the if statement as I did.
what do you think?

src/savers/tvg/tvgTvgSaver.cpp
mgrudzinska: @hermet please check now. I call `strokeColor` and `strokeFill` before calling `serializeStroke` since otherwise for each shape from the svg file a stroke will be serialized, even if it's not visible.

src/savers/tvg/tvgTvgSaver.cpp
hermet: I can't understand why you matter the svg spec in tvg format...
It just need to save/restore the tvg values from/to tvg::paint.

hermet (changes requested):
Could you please explain why do you concern the svg spec here?

src/savers/tvg/tvgTvgSaver.cpp
mgrudzinska: I changed it, since it was wrong.

what I wanted, was to avoid saving the values, that are default in the svg format. 
so if we have a path `<path d="..." stoke-width="10" stroke="red" />` we know that this stroke will be plotted with cap=butt and join=miter.  but, since these are not the default values in ThorVG, we have to save them in the tvg file.

hermet (approved):

PR #649, opened 2021-07-23
svg_loader: code refactoring.

hermet:
clean up about logging before replacing it with TVGLOG()

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2484,49 +2484,50 @@ static void _styleInherit(SvgStyleProperty* child, const SvgStyleProperty* paren
 }
 
 
+static void _inefficientNodeCheck(TVG_UNUSED SvgNode* node){
 #ifdef THORVG_LOG_ENABLED
-static void _inefficientNodeCheck(SvgNode* node){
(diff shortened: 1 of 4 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2484,49 +2484,50 @@ static void _styleInherit(SvgStyleProperty* child, const SvgStyleProperty* paren
...
-}
 #endif
+}
+
 
 static void _updateStyle(SvgNode* node, SvgStyleProperty* parentStyle)
 {
     _styleInherit(node->style, parentStyle);
-#ifdef THORVG_LOG_ENABLED
     _inefficientNodeCheck(node);
```
JSUYA, line 2530: _inefficientNodeCheck and simpleXmlNodeTypeToString are methods used only when Log enabled.
We can avoid unnecessary function calls by wrapping the method implementation code and calling code with THORVG_LOG_ENABLED.

hermet: @JSUYA I believe that is costless since the function is empty. Even some compiler will remove that useless instructions. We'd rather to avoid #ifdef call in this case.

PR #650, opened 2021-07-23
common log: use the log macro to replace the print method easier.

hermet:
We can replace the system logger method by changing single line print source in common,
This also helps to remove the THORVG_LOG_ENABLED macro from each use-cases.

TVGLOG(): To print the hint & tip messages for users.
TVGERR(): To print the error message for debugging.

@Issues: https://github.com/Samsung/thorvg/issues/36

### DIFF
--- a/src/lib/tvgCommon.h
+++ b/src/lib/tvgCommon.h
@@ -28,6 +28,16 @@
 using namespace std;
 using namespace tvg;
 
+//for MSVC Compat
+#ifdef _MSC_VER
+    #define TVG_UNUSED
+    #define strncasecmp _strnicmp
+    #define strcasecmp _stricmp
+#else
+    #define TVG_UNUSED __attribute__ ((__unused__))
+#endif
+
+
 //TVG class identifier values
 #define TVG_CLASS_ID_UNDEFINED 0
 #define TVG_CLASS_ID_SHAPE     1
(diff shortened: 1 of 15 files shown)


src/lib/tvgCommon.h
```diff
@@ -38,14 +48,12 @@ using namespace tvg;
...
 enum class FileType { Tvg = 0, Svg, Raw, Png, Jpg, Unknown };
 
-//for MSVC Compat
-#ifdef _MSC_VER
-    #define TVG_UNUSED
-    #define strncasecmp _strnicmp
-    #define strcasecmp _stricmp
+#ifdef THORVG_LOG_ENABLED
+    #define TVGLOG(fmt, ...) fprintf(stdout, fmt "\n", ##__VA_ARGS__)  //Error Message for us to fix it
+    #define TVGERR(fmt, ...) fprintf(stderr, fmt "\n", ##__VA_ARGS__)  //Log Message for notifying user some useful hints
```
JSUYA, line 53: We use module name for most of our logs.
"module: ~~~"
So how about adding a module parameter to the LOG API? or wrapper?
ex)
TVG("~~~");
TVG("LOADER", "~~~");
TVG("SW_ENGINE", "~~~");

hermet: @JSUYA good suggestion. updated.

JSUYA (approved):
LGTM

PR #651, opened 2021-07-23
sw_engine: removed unused rleAlphaMask

mmaciola:


hermet:
Actually, it's good to use If that logic properly guided...

hermet (approved):

PR #660, opened 2021-07-26
common picture: remove viewbox() api.

hermet:
picture provides size() interface to return the image size,
viewbox() is conceptually not correct with the Picture.

Remove it under the beta api.


JSUYA (commented):
I'll check if other code is using the viewbox() API.

mmaciola (approved):
Looks ok. I am ok with removing x, y viewbox information.
`tvg_picture_get_viewbox` is used after svg loading in efl. When merged, I will create a patch for efl.

PR #663, opened 2021-07-27
sw_engine Raster: Improving image quality in image scale down

JSUYA:
An average value sampled based on adjacent pixels of the target pixel is used.

[Before / After / ubuntu image viewer ]
image

[Before / After ](opacity 200)
image

### DIFF
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -113,6 +113,26 @@ static uint32_t _applyBilinearInterpolation(const uint32_t *img, uint32_t w, uin
     return COLOR_INTERPOLATE(COLOR_INTERPOLATE(c1, 255 - dX, c2, dX), 255 - dY, COLOR_INTERPOLATE(c4, 255 - dX, c3, dX), dY);
 }
 
+static uint32_t _average2Nx2NPixel(SwSurface* surface, const uint32_t *img, uint32_t w, uint32_t h, uint32_t rX, uint32_t rY, uint32_t n)
+{
+    uint32_t c[4] = { 0 };
+    auto n2 = n * n;
+    auto source = img + rX - n + (rY - n) * w;
+    for (auto y = rY - n; y < rY + n; ++y) {
+        auto src = source;
+        for (auto x = rX - n; x < rX + n; ++x, ++src) {
+            c[0] += *src >> 24;
+            c[1] += (*src >> 16) & 0xff;
+            c[2] += (*src >> 8) & 0xff;
+            c[3] += *src & 0xff;
+        }
+        source += w;
+    }
+    for (auto i = 0; i < 4; ++i) {
+        c[i] = (c[i] >> 2) / n2;
+    }
+    return (c[0] << 24) | (c[1] << 16) | (c[2] << 8) | c[3];
+}
 
 /************************************************************************/
 /* Rect                                                                 */
@@ -332,6 +352,7 @@ static bool _rasterTranslucentImageRle(SwSurface* surface, const SwRleData* rle,
     return true;
 }
 
+
 static bool _rasterTranslucentUpScaleImageRle(SwSurface* surface, const SwRleData* rle, uint32_t *img, uint32_t w, uint32_t h, uint32_t opacity, const Matrix* invTransform)
 {
     auto span = rle->spans;
@@ -349,6 +370,30 @@ static bool _rasterTranslucentUpScaleImageRle(SwSurface* surface, const SwRleDat
             uint32_t src;
             if (rX == w - 1 || rY == h - 1) src = ALPHA_BLEND(img[rY * w + rX], alpha);     //TODO: need to use image's stride
             else src = ALPHA_BLEND(_applyBilinearInterpolation(img, w, h, fX, fY), alpha);     //TODO: need to use image's stride
+            *dst = src + ALPHA_BLEND(*dst, 255 - surface->blender.alpha(src));
+        }
+    }
+    return true;
+}
+
+
+static bool _rasterTranslucentDownScaleImageRle(SwSurface* surface, const SwRleData* rle, uint32_t *img, uint32_t w, uint32_t h, uint32_t opacity, const Matrix* invTransform, float scaling)
+{
+    uint32_t halfScaling = static_cast<uint32_t>(0.5f / scaling);
+    if (halfScaling == 0) halfScaling = 1;
+    auto span = rle->spans;
+    for (uint32_t i = 0; i < rle->size; ++i, ++span) {
+        auto ey1 = span->y * invTransform->e12 + invTransform->e13;
+        auto ey2 = span->y * invTransform->e22 + invTransform->e23;
+        auto dst = &surface->buffer[span->y * surface->stride + span->x];
+        auto alpha = ALPHA_MULTIPLY(span->coverage, opacity);
+        for (uint32_t x = 0; x < span->len; ++x, ++dst) {
+            auto rX = static_cast<uint32_t>(roundf(x * invTransform->e11 + ey1));
+            auto rY = static_cast<uint32_t>(roundf(x * invTransform->e21 + ey2));
+            if (rX >= w || rY >= h) continue;
+            uint32_t src;
+            if (rX < halfScaling || rY < halfScaling || rX >= w - halfScaling || rY >= h - halfScaling) src = ALPHA_BLEND(img[rY * w + rX], alpha);     //TODO: need to use image's stride
+            else src = ALPHA_BLEND(_average2Nx2NPixel(surface, img, w, h, rX, rY, halfScaling), alpha);     //TODO: need to use image's stride
             *dst = src + ALPHA_BLEND(*dst, 255 - surface->blender.alpha(src));
         }
     }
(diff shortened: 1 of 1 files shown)


hermet (commented):
I think image raster needs 3 scenarios.
1. Arbitrary Tranfsformed (Required texture mapping based algorithm)
2. Just Scaled Up/Down => Simple Bilinear (In that case do we really need up/down separate methods?...)
3. Original => Immediate Drawing.

I'm expecting scenario 2 which is the general use-case we can focus on in this enhancement...

hermet:
@JSUYA still draft?

JSUYA:
> I think image raster needs 3 scenarios.
> 
> 1. Arbitrary Tranfsformed (Required texture mapping based algorithm)
> 2. Just Scaled Up/Down => Simple Bilinear (In that case do we really need up/down separate methods?...)
> 3. Original => Immediate Drawing.
> 
> I'm expecting scenario 2 which is the general use-case we can focus on in this enhancement...

You want to apply bilinear interpolation to both cases "Up/Down scale",
but I can't figure out how to properly apply it in Down scale.
so.. In down scale, adjacent pixels are lost according to the scale ratio, so I averaged them.
I hope this approach is appropriate.

Please review again

hermet:
@mgrudzinska It needs more beholders, Could you please review this also?

mgrudzinska:
Of course, I'll review this patch

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -1387,10 +1582,12 @@ bool rasterImage(SwSurface* surface, SwImage* image, const Matrix* transform, co
             return _rasterImageRle(surface, image->rle, image->data, image->w, image->h);
         } else {
             if (translucent) {
-                if (isUpScaling) return _rasterTranslucentUpScaleImageRle(surface, image->rle, image->data, image->w, image->h, opacity, &invTransform);
+                if (scaling > 1.0f) return _rasterTranslucentUpScaleImageRle(surface, image->rle, image->data, image->w, image->h, opacity, &invTransform);
```
mgrudzinska, line 1585: I'd rather compare floats using epsilon, but as you wish. In the worst case scenario an unnecessary scaling will be done

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -113,6 +113,24 @@ static uint32_t _applyBilinearInterpolation(const uint32_t *img, uint32_t w, uin
...
+{
+    uint32_t c[4] = { 0 };
+    auto n2 = n * n;
+
+    for (auto x = rX - n; x < rX + n; ++x) {
+        for (auto y = rY - n; y < rY + n; ++y) {
+            c[0] += img[x + (y * w)] >> 24;
+            c[1] += (img[x + (y * w)] >> 16) & 0xff;
+            c[2] += (img[x + (y * w)] >> 8) & 0xff;
+            c[3] += img[x + (y * w)] & 0xff;
```
mgrudzinska, line 126: I'd write ```auto pix = img[x + (y * w)]``` and then use `pix` to avoid 4 multiplics

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -113,6 +113,24 @@ static uint32_t _applyBilinearInterpolation(const uint32_t *img, uint32_t w, uin
...
+        }
+    }
+    for (auto i = 0; i < 4; ++i) {
+        c[i] = (c[i] >> 2) / n2;
+    }
+    return (c[0] << 24) + (c[1] << 16) + (c[2] << 8) + c[3];
```
mgrudzinska, line 132: I'd use `|` instead of `+`

mgrudzinska:
if it's not very urgent, I'd like to read a bit about scaling before I finish my rev

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -1369,11 +1561,14 @@ bool rasterClear(SwSurface* surface)
 bool rasterImage(SwSurface* surface, SwImage* image, const Matrix* transform, const SwBBox& bbox, uint32_t opacity)
 {
     Matrix invTransform;
-    auto isUpScaling = false;
+    float scaling = 1.0f;
 
     if (transform) {
         if (!_inverse(transform, &invTransform)) return false;
-        isUpScaling = (transform->e11 * transform->e11) + (transform->e21 * transform->e21) > 1 ? true : false;
+        scaling = sqrt((transform->e11 * transform->e11) + (transform->e21 * transform->e21));
```
mgrudzinska, line 1568: since we know the scaling factor here, I'd calculate here `halfInvScaling = 0.5f/scaling` and pass this value to the down scaling functions as the `halfScaleRatio` value. In this way we don't need to call sqrt and multiplications. seams to be cheaper. what do u think?

mgrudzinska:
I've made some tests:

1) scaling << 1 - one can see a big improvement after this patch:
pr663_n2
pr663_n3

2) 1 > scaling > 0.5 - not much changes when compared to the current version of TVG, results from the viewer are better
we could consider some other algorithm in such cases
pr663_n4
pr663_n1

mgrudzinska (commented):
please check my comments

mgrudzinska:
I'd use bilinear interpolation for the cases when scaling is in the range 0.5-1. 
but I don't have args from literature for that. It seams to be reasonable though - we use 4 pixels in both algorithms. the only difference is, that in the one for the downsampling used in this patch, the weight for these 4 pixels are equal, and in the bilinear interpolation they are not. 
@JSUYA what do you think of such an approach?

JSUYA:
> I'd use bilinear interpolation for the cases when scaling is in the range 0.5-1.
> but I don't have args from literature for that. It seams to be reasonable though - we use 4 pixels in both algorithms. the only difference is, that in the one for the downsampling used in this patch, the weight for these 4 pixels are equal, and in the bilinear interpolation they are not.
> @JSUYA what do you think of such an approach?

wow thanks for your good idea.  As you said, when 0.5 - 1 down scaling, sampling algorithms doesn't make much sense.
I will improve to apply bilinear interpolation when the scale is 0.5 - 1.
It may take a little more time to change the code :)

JSUYA:
I did some tests. And as a result, I modified to use bilinear interpolation when scale is down(0.25) ~ up, and to use average pixel only in 0.25 or less.

70%
Screenshot from 2021-09-07 15-25-09

30%
Screenshot from 2021-09-07 15-19-29

10%
Screenshot from 2021-09-07 15-22-42

src/lib/sw_engine/tvgSwRaster.cpp
JSUYA: modified

src/lib/sw_engine/tvgSwRaster.cpp
JSUYA: modified

src/lib/sw_engine/tvgSwRaster.cpp
JSUYA: Fixed to use FLT_EPSILON.

src/lib/sw_engine/tvgSwRaster.cpp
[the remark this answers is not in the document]
JSUYA: I modified it to change halfScaleRatio code and use parameters.

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -416,6 +461,30 @@ static bool _rasterUpScaleImageRle(SwSurface* surface, SwRleData* rle, uint32_t
...
+    if (halfScaling == 0) halfScaling = 1;
+    auto span = rle->spans;
+
+    for (uint32_t i = 0; i < rle->size; ++i, ++span) {
+        auto ey1 = span->y * invTransform->e12 + invTransform->e13;
+        auto ey2 = span->y * invTransform->e22 + invTransform->e23;
+        auto dst = &surface->buffer[span->y * surface->stride + span->x];
+        for (uint32_t x = 0; x < span->len; ++x, ++dst) {
+            auto rX = static_cast<uint32_t>(roundf(x * invTransform->e11 + ey1));
+            auto rY = static_cast<uint32_t>(roundf(x * invTransform->e21 + ey2));
```
mgrudzinska, line 476: span->x is missing
```
auto rX = static_cast<uint32_t>(roundf((span->x + x) * invTransform->e11 + ey1));
auto rY = static_cast<uint32_t>(roundf((span->x + x) * invTransform->e21 + ey2));
```

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -601,6 +670,105 @@ static bool _rasterTranslucentUpScaleImage(SwSurface* surface, const uint32_t *i
...
+    if (halfScaling == 0) halfScaling = 1;
+    auto dbuffer = &surface->buffer[region.min.y * surface->stride + region.min.x];
+
+    for (auto y = region.min.y; y < region.max.y; ++y) {
+        auto dst = dbuffer;
+        auto ey1 = y * invTransform->e12 + invTransform->e13;
+        auto ey2 = y * invTransform->e22 + invTransform->e23;
+        for (auto x = region.min.x; x < region.max.x; ++x, ++dst) {
+            auto rX = static_cast<uint32_t>(roundf(x * invTransform->e11 + ey1));
+            auto rY = static_cast<uint32_t>(roundf(x * invTransform->e21 + ey2));
```
mgrudzinska, line 685: span->x is missing
```
auto rX = static_cast<uint32_t>(roundf((span->x + x) * invTransform->e11 + ey1));
auto rY = static_cast<uint32_t>(roundf((span->x + x) * invTransform->e21 + ey2));
```

mgrudzinska (changes requested):
@JSUYA I wouldn't go so far with the bilinear interp... this alg is not the best one for scaling down. it's better than nothing (so in our case for scaling 0.5-1). but in the range 0.25-0.5 it doesn't seem to be a better choice than the averagePixel alg. 

please compare the results for scaling = 0.249 (left) and 0.25 (right):

27_0249_0250_p4
27_0249_0250_t1

the drop of the image quality when switching from 0.249 -> 0.25 seems to be huge (you have to enlarge all the images to see it)

when I changed the 'the switching factor' (0.25) to be 0.5, the difference is not any more so huge (0.49 left, 0.5 right):

27_049_050_p4
27_049_050_t1
 
I'd set 0.5 as the limit and consider to introduce some other alg for 0.5-1 in the future. But if you think differently, we can keep the 0.25, it's not so important probably :)

src/lib/sw_engine/tvgSwRaster.cpp
JSUYA: I modified

JSUYA:
@mgrudzinska Thank you for check. 
I thought that the effect was small because the value of the halfScaling(average2Nx2NPixel()) referenced when the downscale value was 0.25f ~ 0.5f is 1. (at least in my resource)
I found the sample image you used (google image search) and tested it.
I changed condition to 0.5f.
If you have any good ideas to improve this, please share to me :)

mgrudzinska (approved):

mgrudzinska:
for me it's ready to merge, unless @hermet wants to take a look

hermet (approved):

PR #669, opened 2021-07-28
common picture: correct non-working size() behavior.

hermet:
Pixel-based image picture doesn't work at size() method.
Obvisouly, we missed to implement it properly.

This patch correct it.

@Issue: https://github.com/Samsung/thorvg/issues/656

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1048,7 +1048,7 @@ public:
      *
      * @BETA_API
      */
-    const uint32_t* data() const noexcept;
+    const uint32_t* data(uint32_t* w, uint32_t* h) const noexcept;
 
     /**
      * Must remove it!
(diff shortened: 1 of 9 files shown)


inc/thorvg.h
```diff
@@ -1048,7 +1048,7 @@ class TVG_EXPORT Picture final : public Paint
      *
      * @BETA_API
      */
-    const uint32_t* data() const noexcept;
+    const uint32_t* data(uint32_t* w, uint32_t* h) const noexcept;
```
JSUYA, line 1051: Same role as `size(uint32_t*, uint32_t*)`. Do we really need it?

hermet: @JSUYA not same. size() is current(changed) size,  and this one is the original size.

Issue #655, opened 2021-07-26
Alpha area of png image is wrong


JSUYA:
![image](https://user-images.githubusercontent.com/7413838/126945706-eed350a0-226d-4fe5-a126-68128e234621.png)

[icon_play.zip](https://github.com/Samsung/thorvg/files/6876199/icon_play.zip)

PicturePng.cpp
```diff
diff --git a/src/examples/PicturePng.cpp b/src/examples/PicturePng.cpp
index 9dd6a03..8603fd7 100644
--- a/src/examples/PicturePng.cpp
+++ b/src/examples/PicturePng.cpp
@@ -29,16 +29,23 @@
 void tvgDrawCmds(tvg::Canvas* canvas)
 {
     if (!canvas) return;
+    //Background
+    auto shape = tvg::Shape::gen();
+    shape->appendRect(0, 0, WIDTH, HEIGHT, 0, 0);    //x, y, w, h, rx, ry
+    shape->fill(255, 255, 255, 255);                 //r, g, b, a
+
+    if (canvas->push(move(shape)) != tvg::Result::Success) return;
 
     for (int i = 0; i < 5; ++i) {
         auto picture = tvg::Picture::gen();
-        if (picture->load(EXAMPLE_DIR"/logo.png") != tvg::Result::Success) {
+        if (picture->load(EXAMPLE_DIR"/icon_play.png") != tvg::Result::Success) {
              cout << "PNG is not supported, Did you enable PNG Loader??" << endl;
              return;
         }
         picture->translate(i* 150, i * 150);
         picture->rotate(30 * i);
         if (canvas->push(move(picture)) != tvg::Result::Success) return;
     }
 }

```

![image](https://user-images.githubusercontent.com/7413838/126945892-87a3e367-97da-4c1f-b10d-1eb232486d6d.png)

mmaciola:
Validated the buffer is loaded properly. After long a research, I found missing a alpha blending for images in tvgSwRaster.cpp
I will create a patch

JSUYA:
![image](https://user-images.githubusercontent.com/7413838/145947539-c7fc2501-9a98-4fe0-8f54-7442c476b88e.png)

I put the background in the PicturePng demo and tested it.
edges(?) are output differently than we expected.
@hermet Could you please check this?

```
diff --git a/src/examples/PicturePng.cpp b/src/examples/PicturePng.cpp
index 1cccf69..14331fb 100644
--- a/src/examples/PicturePng.cpp
[...]

PR #670, opened 2021-07-28
sw_engine: fix missing alpha blending for images

mmaciola:
There were a missing alpha blending for images that resulted in flipping to the next byte after adding dst pixels.
@issue: #655

### DIFF
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -552,7 +552,7 @@ static bool _rasterImage(SwSurface* surface, uint32_t *img, uint32_t w, TVG_UNUS
         auto dst = dbuffer;
         auto src = sbuffer;
         for (auto x = region.min.x; x < region.max.x; x++, dst++, src++) {
-            *dst = *src + ALPHA_BLEND(*dst, 255 - surface->blender.alpha(*src));
+            *dst = ALPHA_BLEND(*src, surface->blender.alpha(*src)) + ALPHA_BLEND(*dst, 255 - surface->blender.alpha(*src));
         }
         dbuffer += surface->stride;
         sbuffer += w;    //TODO: need to use image's stride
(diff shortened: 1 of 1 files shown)


hermet:
hmm it looks image source is not applied in pre-multiplied alpha...  how do we figure it out from the source.

JSUYA:
Are there no cases that alpha(opacity) is applied as a duplicate?
Need to some more check..

mmaciola:
No problems regarding the opacity and duplication were found
```
    picture->load("/home/mmaciola_local/Downloads/icon_play.png");
    picture->size(200, 200);

    auto picture2 = unique_ptr<tvg::Picture>(static_cast<tvg::Picture*>(picture->duplicate()));
    picture2->translate(200, 0);
    picture2->opacity(128);

    auto picture3 = unique_ptr<tvg::Picture>(static_cast<tvg::Picture*>(picture2->duplicate()));
    picture3->translate(400, 0);
    picture3->opacity(255);

    if (canvas->push(move(picture)) != tvg::Result::Success) return;
    if (canvas->push(move(picture2)) != tvg::Result::Success) return;
    if (canvas->push(move(picture3)) != tvg::Result::Success) return;
```
Screenshot from 2021-08-06 16-54-10

JSUYA (commented):
Sorry for the very late review.
This alpha addition calculation should also be applied to other image rasterize code like _rasterImageRle, translucentImage etc.
(+ recently added Upscale raster function)

JSUYA:
hmm.. as far as I understand.. for this code to work, all images we use must be non-premultiplied color. 

According to libpng's manual(http://www.libpng.org/pub/png/libpng-manual.txt), it seems that we can get a pemultiplied image by using png_set_alpha_mode(..., PNG_ALPHA_PREMULTIPLIED,...) before loading the image. (I may have misunderstood :( .. ) 
But it looks like we need to read using png_structp rather than simplified API (png_image and etc...) for this operation.

@hermet What do you think?

hermet:
We need a unified method from the loader to pass the pixel data in one uniform.
Currently, sw raster engine is supposed to do accept only premultiplied values to avoid additional blending computation.
So I think imaeg loaders(which has alpha channel) needs to prepare premultiplied pixels. 
So... yes. (@JSUYA good catch) we can try use that option whether it's available.

fyi, raster engine might need an option to deal both premultiplied & non-premultipled pixels in the end but not this stage.

@mmaciola please check @JSUYA's suggestion.

hermet (changes requested):
Please check the conversation.

mmaciola:
Thanks @JSUYA for suggestion. I have to investigate `png_set_alpha_mode` and the possibility of using it when using a simplified API of libpng

PR #673, opened 2021-07-30
example Picture: improve the test coverage.

hermet:
In this patch, it replaced the test images with Lenna.
and fix the wrong jpeg file size figuring.


test/testPicture.cpp
```diff
@@ -118,7 +118,7 @@ TEST_CASE("Load PNG file from data", "[tvgPicture]")
     REQUIRE(picture);
 
     //Open file
-    ifstream file(TEST_DIR"/logo.png");
+    ifstream file(TEST_DIR"/test.png");
```
JSUYA, line 121: This test will failed. Lenna size is 512 x 512

JSUYA (commented):
please check unit test result

JSUYA (approved):
LGTM

PR #676, opened 2021-08-01
loader: Colorspace of the PNG image is not considered

patrykka:


patrykka:
I am not sure if this is proper solution for this problem.
Libpng doesn't allow to apply color space after image load, so here is some limitation. Additionaly information about used color space is only in canvas, is there any other way to load in into Picutre object?

mmaciola:
The issue is global: for none of the loaders (jpg, png, raw and svg embeded images) it works now.
Setting RGBA/BGRA on `read()` function is ok, but we should pass `surface->cs` internally and do not left it for user to handle.
I think we can postpone `read()` function, calling it after pushing into the canvas and getting `Colorspace` from `RenderMethod`.
I think we could move `Colorspace` from `SwCanvas` to `Canvas`.

patrykka:
> The issue is global: for none of the loaders (jpg, png, raw and svg embeded images) it works now.
> Setting RGBA/BGRA on `read()` function is ok, but we should pass `surface->cs` internally and do not left it for user to handle.
> I think we can postpone `read()` function, calling it after pushing into the canvas and getting `Colorspace` from `RenderMethod`.
> I think we could move `Colorspace` from `SwCanvas` to `Canvas`.

Thank you for suggestion.
I will postpone read and make it general for all loaders.

hermet:
https://github.com/Samsung/thorvg/pull/1104

PR #677, opened 2021-08-03
loader: removed loader finding for LoaderMgr::loader(data, w, h, copy)

mmaciola:
LoaderMgr::loader(data, w, h, copy) is dedicated for raw images only.
No need to try other loaders.


JSUYA (approved):
looks good to me

hermet (approved):
good catch thanks!

Issue #571, opened 2021-07-02
Pass filetype to picture::load(data, size, copy)


mmaciola:
When we pass data into picture::load, loaders are tried one by one until success.
It is often known what type of data are passed and it would be beneficial (in terms of loading time and memory usage) to pass filetype into picture::load and try the expected loader first.

hermet:
Good Issue.

Here is my brief idea.

picture::load(data, size, string mimeType, copy) ;

if mimeType is given with a specific format name such as "jpeg", "jpg", "png", etc",
it could first try with the mapping table (format - registered loader) first.

PR #678, opened 2021-08-03
loaders: Pass mimetype to picture::load

mmaciola:
Added mimetype attribute to enfaster loading using a proper loader.

@Changed api: Picture::load(const char* data, uint32_t size, const std::string& mimeType, bool copy = false)
@issue: #571

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1005,6 +1005,7 @@ public:
      *
      * @param[in] data A pointer to a memory location where the content of the picture file is stored.
      * @param[in] size The size in bytes of the memory occupied by the @p data.
+     * @param[in] mimetype Mimetype of data. If null or unknown loaders will be tried one by one.
      * @param[in] copy Decides whether the data should be copied into the engine local buffer.
      *
      * @retval Result::Success When succeed.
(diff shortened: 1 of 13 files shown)


JSUYA (commented):
How about keeping original API together?
`Result load(const char* data, uint32_t size, bool copy = false)`
I don't think it is essential.

inc/thorvg.h
```diff
@@ -1005,6 +1005,7 @@ class TVG_EXPORT Picture final : public Paint
      *
      * @param[in] data A pointer to a memory location where the content of the picture file is stored.
      * @param[in] size The size in bytes of the memory occupied by the @p data.
+     * @param[in] mimetype Mimetype of data. If null or unknown loaders will be tried one by one.
```
hermet, line 1008: We can provide better explicit description.

Mimetype of data => Mimetype of data such as "jpg", "svg", "png", etc.

If null => If empty string or @c nullptr

inc/thorvg.h
```diff
@@ -1016,7 +1017,7 @@ class TVG_EXPORT Picture final : public Paint
      *
      * @warning: you have responsibility to release the @p data memory if the @p copy is true
      */
-    Result load(const char* data, uint32_t size, bool copy = false) noexcept;
```
hermet, line 1019: This api must be retained for the compatibility. You can deprecate it or just leave it now.

hermet (changes requested):
Please check comments.

mmaciola:
@hermet @JSUYA fixed according to the comments.
Original api `Result load(const char* data, uint32_t size, bool copy = false)` kept with deprecated note

hermet (approved):

PR #687, opened 2021-08-05
tvg_saver: optimize binary format.

hermet:
Skip to save transform data by accumulating them through the scene tree,
and then applying the final transform to the path points.

Assume that each transform have 36 bytes, it could be increased linear to paints node count
if every paints has transform in the worst case.

Fudamentally, this save their memory and only remains to Bitmap Pictures,
also helps to reduce the loading/rendering workloads since
it doesn't need to perform any transform jobs after converting.


hermet:
Question. Do we really need this?? in the worst case it's good but well.... in the practical scenario...

PR #688, opened 2021-08-05
JPG loader: print error string on failure

mmaciola:
Added error string printing on jpg image loading failure.
The error message help developer find the corrupted jpg file.
Error message is not printed for open from char* data as there the
loaders are tried on by one.

### DIFF
--- a/src/loaders/jpg/tvgJpgLoader.cpp
+++ b/src/loaders/jpg/tvgJpgLoader.cpp
@@ -78,7 +78,7 @@ bool JpgLoader::open(const string& path)
 
     int width, height, subSample, colorSpace;
     if (tjDecompressHeader3(jpegDecompressor, data, size, &width, &height, &subSample, &colorSpace) < 0) {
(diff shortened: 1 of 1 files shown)


src/loaders/jpg/tvgJpgLoader.cpp
```diff
@@ -77,7 +77,13 @@ bool JpgLoader::open(const string& path)
     if (fread(data, size, 1, jpegFile) < 1) goto failure;
 
     int width, height, subSample, colorSpace;
-    if (tjDecompressHeader3(jpegDecompressor, data, size, &width, &height, &subSample, &colorSpace) < 0) goto failure;
+    if (tjDecompressHeader3(jpegDecompressor, data, size, &width, &height, &subSample, &colorSpace) < 0) {
+#ifdef THORVG_LOG_ENABLED
```
hermet, line 81: Please use TVGERR()

hermet (changes requested):

hermet (approved):

Issue #459, opened 2021-06-17
Provide command line tool to open and parse multiple files  


qarmin:
Currently I don't see method to be able to test if loading different files will cause memory leaks or crash.

The best way, I think, is to create a command line app which will allow to open multiple files at once e.g.

in folder /home/user/sierra there is 1000 svg files, so command
```
thor-opener /home/user/sierra
```
will open and parse one by one files.

Output should be like this(of course when compiling with sanitizer support):
```
Opening file blabla.svg
Opening file blabla2.svg
Address sanitizer Usage after free:
.....
.....
.....
```
so it will show that `blabla2.svg` cause problems

hermet:
@qarmin Do you satisfy if we provide that kind of sample in thorvg repo?

qarmin:
Yeah, it will be great, for now I tried to use svg2png, but sadly it doesn't compile with address sanitizer

hermet:
i think we can improve the svg2png to allow all svg files in the folder.
example ...
$svg2png *.svg  400x400 ff00ff
$svg2png ./res/*.svg  400x400 ff00ff

hermet:
@JSUYA good idea?

PR #698, opened 2021-08-09
svg2png: redesigned

mmaciola:
Svg2png fully redesigned.
Introduced wildcards and multiple files testing. Created naming and colorspace flags.

```
Usage: 
   svg2png [svgFileName] [Resolution] [bgColor] [flags]

Flags:
    -d Disable looking in subdirectories for wildcard
    -c Use ABGR8888 colorspace. Default is ARGB8888.
    -s Don't save png file. Useful for testing memory leaks.

Examples:
    $ svg2png input.svg
    $ svg2png input.svg 200x200
    $ svg2png input.svg 200x200 ff00ff
    $ svg2png *.svg
    $ svg2png *.svg 200x200 #ff0000 -d -c
    $ svg2png *.svg -d -c -s
```

@Issue: #459

### DIFF
--- a/src/bin/svg2png/svg2png.cpp
+++ b/src/bin/svg2png/svg2png.cpp
@@ -25,180 +25,358 @@
 #include <thorvg.h>
 #include <vector>
 #include "lodepng.h"
+#include <dirent.h>
+#include <unistd.h>
+#include <limits.h>
 
 using namespace std;
 
+#define FLAG_PARSE_DIRECTORY (1)
+#define FLAG_PARSE_RECURSIVE (1 << 1)
+#define FLAG_USE_ABGR8888_COLORSPACE (1 << 2)
+#define FLAG_DO_NOT_SAVE_PNG (1 << 3)
 
 struct PngBuilder
 {
-    void build(const std::string &fileName , const uint32_t width, const uint32_t height, uint32_t *buffer)
+    void build(const string& fileName, const uint32_t width, const uint32_t height, uint32_t* buffer, bool altColorspace)
     {
-        std::vector<unsigned char> image;
-        image.resize(width * height * 4);
-        for (unsigned y = 0; y < height; y++) {
-            for (unsigned x = 0; x < width; x++) {
-                uint32_t n = buffer[ y * width + x ];
-                image[4 * width * y + 4 * x + 0] = (n >> 16) & 0xff;
-                image[4 * width * y + 4 * x + 1] = (n >> 8) & 0xff;
-                image[4 * width * y + 4 * x + 2] = n & 0xff;
-                image[4 * width * y + 4 * x + 3] = (n >> 24) & 0xff;
+        unsigned error = 0;
+        if (!altColorspace) {
+            //Used ARGB8888 so have to move pixels now
+            vector<unsigned char> image;
+            image.resize(width * height * 4);
+            for (unsigned y = 0; y < height; y++) {
+                for (unsigned x = 0; x < width; x++) {
+                    uint32_t n = buffer[y * width + x];
+                    image[4 * width * y + 4 * x + 0] = (n >> 16) & 0xff;
+                    image[4 * width * y + 4 * x + 1] = (n >> 8) & 0xff;
+                    image[4 * width * y + 4 * x + 2] = n & 0xff;
+                    image[4 * width * y + 4 * x + 3] = (n >> 24) & 0xff;
+                }
             }
+            error = lodepng::encode(fileName, image, width, height);
+
+        } else {
+            //Used ABGR8888 so no need to move pixels
+            error = lodepng::encode(fileName, (const unsigned char*) buffer, width, height);
         }
-        unsigned error = lodepng::encode(fileName, image, width, height);
 
         //if there's an error, display it
-        if (error) std::cout << "encoder error " << error << ": "<< lodepng_error_text(error) << std::endl;
+        if (error) cout << "encoder error " << error << ": " << lodepng_error_text(error) << endl;
     }
 };
 
-
-struct App
+struct Renderer
 {
-    void tvgRender(int w, int h)
+public:
+    int render(const char* path, int w, int h, const string& dst, uint32_t bgColor, bool altColorspace, bool dontSavePng)
     {
+        //Canvas
+        if (!canvas) createCanvas();
+        if (!canvas) {
+            cout << "Error: Canvas failure" << endl;
+            return 1;
+        }
+
+        //Picture
+        auto picture = tvg::Picture::gen();
+        if (picture->load(path) != tvg::Result::Success) {
+            cout << "Error: Couldn't load image " << path << endl;
+            return 1;
+        }
+
+        if (w == 0 || h == 0) {
+            float fw, fh;
+            picture->size(&fw, &fh);
+            w = static_cast<uint32_t>(fw);
+            h = static_cast<uint32_t>(fh);
+        } else {
+            picture->size(w, h);
+        }
+
+        //Buffer
+        createBuffer(w, h);
+        if (!buffer) {
+            cout << "Error: Buffer failure" << endl;
+            return 1;
+        }
+
+        if (canvas->target(buffer, w, w, h, altColorspace ? tvg::SwCanvas::ABGR8888 : tvg::SwCanvas::ARGB8888) != tvg::Result::Success) {
+            cout << "Error: Canvas target failure" << endl;
+            return 1;
+        }
+
+        //Background color if needed
+        if (bgColor != 0xffffffff) {
+            uint8_t r = (uint8_t)((bgColor & 0xff0000) >> 16);
+            uint8_t g = (uint8_t)((bgColor & 0x00ff00) >> 8);
+            uint8_t b = (uint8_t)((bgColor & 0x0000ff));
+
+            auto shape = tvg::Shape::gen();
+            shape->appendRect(0, 0, w, h, 0, 0);
+            shape->fill(r, g, b, 255);
+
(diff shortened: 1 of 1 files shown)


src/bin/svg2png/svg2png.cpp
```diff
@@ -25,180 +25,392 @@
...
-                width = static_cast<uint32_t>(fw);
-                height = static_cast<uint32_t>(fh);
-            //Otherwise, transform size to keep aspect ratio
-            } else {
-                width = w;
-                height = h;
-                picture->size(w, h);
-            }
+        //Buffer
+        _createBuffer(w, h);
```
hermet, line 96: Some has "_" prefix some doesn't... :-1:  you can remove it for the methods.

src/bin/svg2png/svg2png.cpp
```diff
@@ -25,180 +25,392 @@
...
-        if (argc > 1) path = argv[1];
-        if (argc > 2) {
-            char tmp[20];
-            char *x = strstr(argv[2], "x");
-            if (x) {
-                snprintf(tmp, x - argv[2] + 1, "%s", argv[2]);
-                *w = atoi(tmp);
-                snprintf(tmp, sizeof(tmp), "%s", x + 1);
-                *h = atoi(tmp);
+        const char* path
```
hermet, line 184: path = nullptr;

src/bin/svg2png/svg2png.cpp
```diff
@@ -25,180 +25,392 @@
...
+        { nullptr };
+
+        for (int i = 0, arg = 0; i < argc; i++) {
+            const char* p = argv[i];
+            if (*p == '-') {
+                //flags
+                if (p[1] == 'd') {
+                    //disable looking in subdirectories for wildcard
+                    flags |= FLAG_DISABLE_SUBDIRS;
+                } else if (p[1] == 'c') {
```
hermet, line 194: Question: which case user need to select it? I don't see any point of it.

hermet (changes requested):
please check comments.

qarmin:
It is possible now to not save png file?
When testing svg files(e.g. for memory leaks) I got several thousands of useless png files which I had to delete manually.

mmaciola:
@qarmin Sure. Created -s for not saving the png file.

src/bin/svg2png/svg2png.cpp
mmaciola: For bugs testing. I think there may be some problems found using this flag.

hermet: this has to be a useful tool as the svg2png, not a testing app. :(

src/bin/svg2png/svg2png.cpp
```diff
@@ -25,180 +25,397 @@
...
+                    image[4 * width * y + 4 * x + 2] = n & 0xff;
+                    image[4 * width * y + 4 * x + 3] = (n >> 24) & 0xff;
+                }
             }
+            error = lodepng::encode(fileName, image, width, height);
+
+        } else {
+            //Used ABGR8888 so no need to move pixels
+            error = lodepng::encode(fileName, (const unsigned char*) buffer, width, height);
         }
```
JSUYA, line 61: i have a question..
If the user changes the colorspace using -c, shouldn't the color of the output also change?

JSUYA (commented):
wildcard and -s option is good idea :+1:

hermet:
@mmaciola i think this app must have the clear purpose that "converting svg to png".  not the testing/profiling. It looks going to a wrong way.

src/bin/svg2png/svg2png.cpp
hermet: This is a user app, frankly, i don't agree on this, leave to @JSUYA s decision.

src/bin/svg2png/svg2png.cpp
```diff
@@ -25,180 +25,397 @@
...
+                    case 0:
+                    {
+                        break;
+                    }
+                    case 1:
+                    {
+                        path = p;
+                        break;
+                    }
+                    case 2:
```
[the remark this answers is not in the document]
mmaciola: It's interesting. It happened on my machine as well, but only for a single * and only without quotation marks. For *.svg and "*" or "*.svg" it doesn't.
Please find my new idea in post below. It would allow for multiple files and be would resistant to such a behavior.

src/bin/svg2png/svg2png.cpp
mmaciola: With and without -c flag it should look the same.
In the primary version of svg2png, ARGB8888 colorspace was used. As lodepng needs ABRG, color pixels was moved from ARGB to ABGR before calling `lodepng::encode`. I added -c that uses ABGR8888 at the beginning, so there is no need for moving pixels.
It can be tested with something that doesnt work correctly now, eg. png images. Png image can be embed in svg image and than colors in generated image will be changed.

hermet:
> @hermet @JSUYA What do you think about such a working mechanism:
> 
> ```
> Usage: 
>    svg2png [svgFileName] [-r resolution] [-b bgColor] [flags]
> 
> Flags:
>     -r set output image resolution
>     -b set output image background color
>     -d parse whole directory. If no directory specified, parse working directory. If single file specified, will parse parent directory.
>     -dr recursive. Same as -d, but include subdirectories.
>     -c Use ABGR8888 colorspace. Default is ARGB8888.
>     -s Don't save png file. Useful for testing memory leaks.
> 
> Examples:
>     $ svg2png input.svg
>     $ svg2png input.svg -r 200x200
>     $ svg2png input.svg -r 200x200 -b ff00ff
>     $ svg2png input1.svg input2.svg -r 200x200 -b ff00ff
>     $ svg2png . -d
>     $ svg2png -dr
>     $ svg2png * -d -c -s
> ```

'-d': Unnecessary. You can check whether the input is a directory or a single file. then process it.
'-dr': Unnecessary. You can recursive it in default. If user don't like it, they can prepare a separate folder.
'-c':  Use ABGR8888 colorspace. Default is ARGB8888.  => What user supposed to do for this? This is not good .
'-s Don't save png file. Useful for testing memory leaks.  => Svg2Png but no Png? hmm... removing generated pngs by testers is not a big deal.

Please keep it in mind, Simple is better. That's my opinion.

hermet:
[31/37] Compiling C++ object 'src/bin/svg2png/65c8cf4@@svg2png@exe/svg2png.cpp.o'
../src/bin/svg2png/svg2png.cpp: In member function ‘int App::setup(int, char**)’:
../src/bin/svg2png/svg2png.cpp:229:49: warning: argument 1 null where non-null expected [-Wnonnull]
                     DIR* dir = opendir(real_path);
                                                 ^
In file included from ../src/bin/svg2png/svg2png.cpp:28:0:
/usr/include/dirent.h:134:13: note: in a call to function ‘DIR* opendir(const char*)’ declared here
 extern DIR *opendir (const char *__name) __nonnull ((1));

src/bin/svg2png/svg2png.cpp
```diff
@@ -25,180 +25,324 @@
...
-                uint32_t n = buffer[ y * width + x ];
+                uint32_t n = buffer[y * width + x];
                 image[4 * width * y + 4 * x + 0] = (n >> 16) & 0xff;
                 image[4 * width * y + 4 * x + 1] = (n >> 8) & 0xff;
                 image[4 * width * y + 4 * x + 2] = n & 0xff;
                 image[4 * width * y + 4 * x + 3] = (n >> 24) & 0xff;
             }
         }
-        unsigned error = lodepng::encode(fileName, image, width, height);
+        error = lodepng::encode(fileName, image, width, height);
```
hermet, line 51: please declare it when it's necessary.
auto error = lodepng::encode(fileName, image, width, height);

src/bin/svg2png/svg2png.cpp
```diff
@@ -25,180 +25,324 @@
...
+        //Build Png
+        PngBuilder builder;
+        builder.build(dst, w, h, buffer);
 
-            //Terminate ThorVG Engine
-            tvg::Initializer::term(tvg::CanvasEngine::Sw);
+        cout << "Generated PNG file: " << dst << endl;
 
-            free(buffer);
+        //Clear canvas
```
hermet, line 122: useless comment.

src/bin/svg2png/svg2png.cpp
```diff
@@ -25,180 +25,324 @@
...
-            }
-        }
-        if (argc > 3) bgColor = strtol(argv[3], NULL, 16);
-
-        if (!path) return help();
+        //Terminate ThorVG Engine
+        tvg::Initializer::term(tvg::CanvasEngine::Sw);
 
-        std::array<char, 5000> memory;
+        //Free buffer
```
hermet, line 133: --useless comment.

src/bin/svg2png/svg2png.cpp
```diff
@@ -25,180 +25,324 @@
...
     }
 
-    bool svgFile() {
-        std::string extn = ".svg";
-        if (fileName.size() <= extn.size() || fileName.substr(fileName.size() - extn.size()) != extn)
-            return false;
+private:
+    unique_ptr<tvg::SwCanvas> canvas = nullptr;
+    uint32_t* buffer = nullptr;
+    uint32_t buffer_size = 0;
```
hermet, line 169: bufferSize

hermet (changes requested):
Please double check the compiler warnings & some more nitpicks.

hermet (approved):

hermet:
@mmaciola it will be better if svg2tvg is enhanced as well.

PR #701, opened 2021-08-09
tvg_saver: skip the invisible paint as possible.

hermet:


hermet:
@mmaciola  @mgrudzinska if you have any opinions, please leave it here...

mmaciola (approved):

mihashco (approved):

src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -412,7 +412,7 @@ TvgBinCounter TvgSaver::serializeComposite(const Paint* cmpTarget, CompositeMeth
     auto flag = static_cast<TvgBinFlag>(cmpMethod);
     auto cnt = writeTagProperty(TVG_TAG_PAINT_CMP_METHOD, SIZE(TvgBinFlag), &flag);
 
-    cnt += serialize(cmpTarget, nullptr);
+    cnt += serialize(cmpTarget, nullptr, true);
```
mgrudzinska, line 415: @hermet I just wanted to approve :)
I wondered, whether it is worth to check, whether the compTarget is the InvAlphaMask and if so call  ```cnt += serialize(cmpTarget, nullptr);``` but I don't think it's a very common case and we'd need to add an additional if

PR #707, opened 2021-08-10
tvg_saver: optimize saving data.

hermet:
This optimizes binary size by skipping the scene if it has the only child.

though the reduced size is too trivial size, (avg 0.4% as far as I checked our example svgs),
we can reduce the loading job & runtime memory.


src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -200,12 +200,59 @@ TvgBinCounter TvgSaver::serializePaint(const Paint* paint)
...
+}
+
+
 TvgBinCounter TvgSaver::serializeScene(const Scene* scene, const Matrix* transform)
 {
+    auto it = this->iterator(scene);
+    if (it->count() == 0) return 0;
+
+    //Case - Only Child: Skip saving this scene.
+    if (it->count() == 1) {
```
mmaciola, line 234: We may lose the matrix transformation and opacity of a scene. We should check these are default before skipping.

hermet: here given transform is the scene's transform and opacity is to multiplied in the serializeChild() inside.
So it takes cover.

mmaciola: you are right. If so, it's absolutely ok

src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -200,12 +200,59 @@ TvgBinCounter TvgSaver::serializePaint(const Paint* paint)
...
+    //If the parent & the only child have composition, we can't skip the parent....
+    if (compMethod != CompositeMethod::None && child->composite(nullptr) != CompositeMethod::None) return 0;
+
+    //propagate opacity
+    uint32_t opacity = parent->opacity();
+
+    if (opacity < 255) {
+        uint32_t tmp = (child->opacity() * opacity);
+        if (tmp > 0) tmp /= 255;
+        const_cast<Paint*>(child)->opacity(tmp);
```
mgrudzinska, line 218: tmp = 0 if child->opacity() = 0, in such a case there is no point to calculate tmp and to set it as a child opacity.
what do you think about the code below? :
```
    if (opacity < 255) {
        if (auto tmp = child->opacity()) {
            tmp = (tmp * opacity) / 255;
            const_cast<Paint*>(child)->opacity(tmp);
        }
    }
```

hermet: Either way, it needs to check opacity of parent to not dividing by zero.
it's ignorable...  I'd prefer simpler code than increase complexity  + 1

mgrudzinska: @hermet you divide by 255, not by 0. 
even if tmp = 0 you can still divide it by 255. it will remain zero.
the thing is, that there is no point of doing anything if the child opacity is already zero.

hermet: Yes anyway, divide 0 will be an issue. so your code is not proper.
and handling exceptions will code dirty and complex even we do there is no gain as you expected.

hermet:
@mgrudzinska  Sometimes we can select the clean code(pros) if the cons is ignoreable. In that case, the discussion is no point, just time-consuming. Rather than that, we can focus on any side effects, wrong coding convention, suspicious mistakes, potential bugs, wrong behavior concepts and obvious performance drop by your experience.

PR #709, opened 2021-08-10
common: RenderRegion x, y protect against negative value in unsigned int

### LINKED ISSUE #704: SVG does not display if negative viewbox
If svg image's viewbox x, y is negative (eg. viewBox="-1 0 100 100") and no `picture->size` is called, images won't display.

Snippet:
```
    if (!canvas) return;

    //Background
    shape->appendRect(0, 0, WIDTH, HEIGHT, 0, 0);
    shape->fill(255, 255, 255, 255);

    if (canvas->push(move(shape)) != tvg::Result::Success) return;

    if (picture->load("images/alphachannel.svg") != tvg::Result::Success) return;

    canvas->push(move(picture));
```

alphachannel.svg:
```
<svg id="svg1" viewBox="-1 0 100 100" xmlns="http://www.w3.org/2000/svg">
  <circle r="32" cx="35" cy="65" fill="#F00" opacity="0.5"/>
  <circle r="32" cx="65" cy="65" fill="#0F0" opacity="0.5"/>
  <circle r="32" cx="50" cy="35" fill="#00F" opacity="0.5"/>
</svg>
```

mmaciola:
This change protects against negative value in unsigned int of
RenderRegion.x/y. This fixes a problem of invisible paint if ClipPath
bounds was negative.

@issue: fixes #704

### DIFF
--- a/src/lib/tvgPaint.cpp
+++ b/src/lib/tvgPaint.cpp
@@ -77,6 +77,9 @@ static bool _clipPathFastTrack(Paint* cmpTarget, const RenderTransform* pTransfo
             y2 = y2 * pTransform->m.e22 + pTransform->m.e23;
         }
 
+        if (x1 < 0.0f) x1 = 0.0f;
+        if (y1 < 0.0f) y1 = 0.0f;
+
         viewport.x = static_cast<uint32_t>(x1);
         viewport.y = static_cast<uint32_t>(y1);
         viewport.w = static_cast<uint32_t>(roundf(x2 - x1 + 0.5f));
--- a/src/lib/tvgPictureImpl.h
+++ b/src/lib/tvgPictureImpl.h
@@ -89,7 +89,7 @@ struct Picture::Impl
                 if (auto p = loader->paint()) {
                     paint = p.release();
                     loader->close();
-                    if (w != loader->w && h != loader->h) {
+                    if (w != loader->w || h != loader->h) {
                         loader->resize(paint, w, h);
                         resizing = false;
                     }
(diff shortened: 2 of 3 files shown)


src/lib/tvgPaint.cpp
```diff
@@ -77,6 +77,9 @@ static bool _clipPathFastTrack(Paint* cmpTarget, const RenderTransform* pTransfo
             y2 = y2 * pTransform->m.e22 + pTransform->m.e23;
         }
 
+        if (x1 < 0.0f) x1 = 0.0f;
+        if (y1 < 0.0f) y1 = 0.0f;
```
mgrudzinska, line 81: I think that the problem is a bit deeper. for now we don't use the viewBox attribute exactly as we should. please check the w3.org for more details. the viewBox can scale the svg (viewBox + width & height attributes) or shift it (x and y values of the viewBox attribute != 0). it doesn't work this way in tvg for now. 

So the problem is not only for x,y < 0, but for values > 0 we don't get the correct result as well.
By overriding the x1 and y1 values by 0 we lose some information that we didn't use yet.

mgrudzinska (changes requested):
please check a comment

src/lib/tvgPaint.cpp
mmaciola: For this patch we don't lose information. The region is just for a fastTrack and it concerns ClipPath in general, not only SVG.
The issue that you mention should be taken separately. Is there some visible difference between tvg and validated viewer? Could you post such an svg file and create an issue for it?

mgrudzinska: @mmaciola
"Could you post such an svg file and create an issue for it?"   ->   first thing tomorrow :)

Issue #266, opened 2021-02-28
one closed shape within the path/shape implies closing all other


mgrudzinska:
if one shape is closed within the path/shape all other will be closed as well

example path:
<path d=\" M 100 100 h 100 v 100 Z M 300 100 h 100 v 100\" fill=\"none\" stroke-width=\"8\" stroke=\"#FF0000\"/>

expected result:
![close_ok](https://user-images.githubusercontent.com/67589014/109437155-ba1f7580-7a23-11eb-95eb-7ed1a1b1554b.PNG)

current result from tvg:
![close_bad](https://user-images.githubusercontent.com/67589014/109437167-c3a8dd80-7a23-11eb-88a1-80bad0307899.PNG)

PR #710, opened 2021-08-11
tvg_saver: optimize tvg format.

hermet:
So this optimization stragtegy is to merging shapes.
If two shapes have the same layer, having save properties except the paths,
we can integrate two shapes to one, this helps to build up a simpler
scene-tree, reduce the runtime memory, helps for faster processing for rendering.

As far as I checked tiger.svg, it removes 142 shape nodes,
decreased the binary size: 60537 -> 54568.

Overall, avg 4% binary size can be reduced by this patch.


mgrudzinska:
@hermet because of  issue #266, if one of the shapes was closed before the merge, after it all of them will be closed :
before merge on the left, after on the right
merging

hermet:
> @hermet because of issue #266, if one of the shapes was closed before the merge, after it all of them will be closed :
> before merge on the left, after on the right
> merging
> although I think it's not very common problem

Yeah. we actually have to deal with that issue... aside from this patch.

PR #720, opened 2021-08-17
picture: fix reloading images

### LINKED ISSUE #719: Can't change the loaded image
Once loaded, image can't be changed. Snippet (it's a real scenario that happens using C-api):
```
    if (!canvas) return;

    auto picture = tvg::Picture::gen().release();
    picture->load(EXAMPLE_DIR"/test.jpg");
    picture->size(400, 400);
    if (canvas->push(unique_ptr<tvg::Picture>(picture)) != tvg::Result::Success) return;

    picture->load(EXAMPLE_DIR"/test2.jpg");
```

mmaciola:
Calling picture->load after it was already once called resulted in
segmentation fault or memory leak (depending on whether the vector (svg, tvg)
or raster (jpg, png, raw) file was loaded).
This patch adds buffers disposing before loading new image.
To dispose already drawn elements, renderer is kept.

@issue: fixes #719

### DIFF
--- a/src/lib/tvgPictureImpl.h
+++ b/src/lib/tvgPictureImpl.h
@@ -61,9 +61,9 @@ struct Picture::Impl
 {
     shared_ptr<LoadModule> loader = nullptr;
     Paint* paint = nullptr;
-    uint32_t *pixels = nullptr;
-    Picture *picture = nullptr;
-    void *rdata = nullptr;              //engine data
+    uint32_t* pixels = nullptr;
+    Picture* picture = nullptr;
+    void* rdata = nullptr;            //engine data
     float w = 0, h = 0;
     bool resizing = false;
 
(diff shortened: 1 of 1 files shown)


src/lib/tvgPictureImpl.h
```diff
@@ -205,6 +219,13 @@ struct Picture::Impl
         if (!loader->read()) return Result::Unknown;
         w = loader->w;
         h = loader->h;
+        if (renderer) dispose(*renderer);
```
hermet, line 222: Dismiss the renderer... a bad case in the point of performance wise.
That would be better if we can prevent this scenario ealier from the app.

mmaciola: Ok. Removed disposing to just checking image is already loaded. If so, return InsufficientCondition.

hermet (approved):
for safety, we can do this right now but reloading(after sync) will be necessary in the end.

PR #724, opened 2021-08-19
svg: fix opacity percentage value parsing

mmaciola:
Percentage values wasn't handled by _toOpacity() function. Other incorrect values was handled wrongly. Now fixed.

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -163,11 +163,13 @@ static float _toOffset(const char* str)
 static int _toOpacity(const char* str)
 {
     char* end = nullptr;
-    int a = 0;
     float opacity = svgUtilStrtof(str, &end);
 
-    if (end && (*end == '\0')) a = lrint(opacity * 255);
-    return a;
+    if (end) {
+        if (end[0] == '%' && end[1] == '\0') return lrint(opacity * 2.55f);
+        else if (*end == '\0') return lrint(opacity * 255);
+    }
+    return 255;
 }
 
 


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -163,11 +163,13 @@ static float _toOffset(const char* str)
...
-    int a = 0;
     float opacity = svgUtilStrtof(str, &end);
 
-    if (end && (*end == '\0')) a = lrint(opacity * 255);
-    return a;
+    if (end) {
+        if (end[0] == '%' && end[1] == '\0') return lrint(opacity * 2.55f);
+        else if (*end == '\0') return lrint(opacity * 255);
+    }
+    return 255;
```
JSUYA, line 172: This return is a different behavior than before, but I think it's right.

JSUYA (approved):
LGTM

JSUYA (commented):
Could you please check one thing?
Opacity is usually expressed as a value between 0.0 and 1.0
Is it in the spec(opacity, fill-opcaity, stroke-opacity, etc...) to use percentage?
We need to check the spec.

mmaciola:
Documentation of [SVGTiny12](https://www.w3.org/TR/SVGTiny12/painting.html) actually specify percentage value for none property. However we accept percentage for various properties in other functions eg. [offset for gradient stop](https://www.w3.org/TR/SVGTiny12/painting.html#StopElement) and `_toOffset` function. In W3 in explicit it is defined just in [SVG2](https://www.w3.org/TR/SVG2/painting.html#FillOpacity).

PR #730, opened 2021-08-19
Wasm: allow iterator and add functions for layers revising

mmaciola:
Divided into three separate commits for easier review:
Commit 1: moves iterator functionality from SaveModule into separate IterratorModule
Commit 2: gives full thorvg dependencies for wasm executable and adds IterratorModule inheritance. These changes provides possibility to browse layers of the loaded image
Commit 3: adds functions for layers revising. New functions in thorvgwasm.cpp:
     `layers()` - that return a list of paints in a picture
     `bounds()` - that returns a bounds of a given paint
     `setOpacity()` - that sets the opacity of a given paint


hermet:
Do you have a concrete idea to utilize this function? We have no exact idea what feature and scenario ThorVG Viewer will provide.

hermet (approved):
At least the changes looks good to me.

hermet (approved):

PR #732, opened 2021-08-20
saver: added saving into a buffer

mmaciola:
This patch adds saving into a buffer functionality into a Saver.
Main purpose of this api now is adding export functionality into ThorVG Viewer

@API Additions:
Result save(std::unique_ptr<Paint> paint, const std::string& mimeType, uint8_t** buffer, uint32_t* size);

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1391,6 +1391,24 @@ public:
     Result save(std::unique_ptr<Paint> paint, const std::string& path) noexcept;
 
     /**
+     * @brief Export the given @p paint data to the given @p buffer.
+     * If export success, buffer size is returned in @p size.
+     *
+     * @param[in] paint The paint to be saved with all its associated properties.
+     * @param[in] mimeType Mimetype of a data, eg. "tvg".
+     * @param[out] buffer An empty buffer, in which the paint data is to be saved.
+     * @param[out] size Size of a buffer.
+     *
+     * @retval Result::Success When succeed.
+     * @retval Result::Unknown Others.
+     *
+     * @note Saving into a buffer is synchronous. No need to call sync().
+     *
+     * @BETA_API
+     */
+    Result save(std::unique_ptr<Paint> paint, const std::string& mimeType, uint8_t** buffer, uint32_t* size) noexcept;
+
+    /**
      * @brief Guarantees that the saving task is finished.
      *
      * The behavior of the saver will work on a sync/async basis, depending on the threading setting of the Initializer.
--- a/src/lib/tvgSaver.cpp
+++ b/src/lib/tvgSaver.cpp
@@ -74,6 +74,7 @@ static SaveModule* _find(FileType type)
 
 static SaveModule* _find(const string& path)
 {
+    //find extension. If dot is not found, will compare whole path string.
     auto ext = path.substr(path.find_last_of(".") + 1);
     if (!ext.compare("tvg")) {
         return _find(FileType::Tvg);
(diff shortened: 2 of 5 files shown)


inc/thorvg.h
```diff
@@ -1390,6 +1390,24 @@ class TVG_EXPORT Saver
...
+     *
+     * @param[in] paint The paint to be saved with all its associated properties.
+     * @param[in] mimeType Mimetype of a data, eg. "tvg".
+     * @param[out] buffer An empty buffer, in which the paint data is to be saved.
+     * @param[out] size Size of a buffer.
+     *
+     * @retval Result::Success When succeed.
+     * @retval Result::Unknown Others.
+     *
+     * @note Saving into a buffer is synchronous. No need to call sync().
```
hermet, line 1405: There is no reason to guarantee the sync behavior.
Whatever its behavior does, user could call sync() after save(), it's better for consistency (like other save methods)

inc/thorvg.h
```diff
@@ -1390,6 +1390,24 @@ class TVG_EXPORT Saver
...
+     *
+     * @BETA_API
+     */
+    Result save(std::unique_ptr<Paint> paint, const std::string& mimeType, uint8_t** buffer, uint32_t* size) noexcept;
```
hermet, line 1409: Consistency wise:
uint32_t save(std::unique_ptr<Paint> paint, const std::string& mimeType, uint8_t** buffer) noexcept;

hermet: Now we have a question, very important before this API.
What is the best guide to prepare the buffer size by user side?

src/lib/tvgSaver.cpp
```diff
@@ -117,6 +118,26 @@ Result Saver::save(std::unique_ptr<Paint> paint, const string& path) noexcept
 }
 
 
+Result Saver::save(std::unique_ptr<Paint> paint, const std::string& mimeType, uint8_t** buffer, uint32_t* size) noexcept
+{
+    //Already on saving an other resource.
```
hermet, line 123: I think user could save multiple times.
ie) to multiple files and to memory.
for later enhancement, you can leave TODO: comment here.

hermet (changes requested):

inc/thorvg.h
mmaciola: I think we can't. Buffer is allocated and reallocated by saver and user must have it's size in return. I dont see it

inc/thorvg.h
mmaciola: Do you see some scenario developer may use it asynchronous?
There is no mechanism of callback on finish and I see users (developers) will always call save() and sync() consecutively.
When saving into a file developer may not care further about the saved file so async is great, but for saving into a buffer next step is always passing data somewhere and async is useless if no callback.

hermet: Yes possibly, user can Save to buffer earlier than use it later.
there is no reason to block the case.

inc/thorvg.h
hermet: Ok, then please keep this api: 
uint32_t save(std::unique_ptr paint, const std::string& mimeType, uint8_t** buffer) noexcept;

btw, I'm wondering why can't use Save to file? (instead of this)

mmaciola: Good intuition: I don't use saving to file as there were rumors that FS api does not work correctly yet. I tested with actual emscripten version and it works ok. Created PR #748. The downside of FS is increase in size of thorvg-wasm.js 
As you wish, we can close this pr or convert to draft for creating saving into a buffer api later

mmaciola:
Closing as feature no more needed now

PR #733, opened 2021-08-20
tvg_loader: protect incorrect image size

mmaciola:
It may happen that tvg file is corrupted. This patch protects negative
image size that may happen, eg. when undesirable utf8 encoding applied.

### DIFF
--- a/src/loaders/tvg/tvgTvgLoader.cpp
+++ b/src/loaders/tvg/tvgTvgLoader.cpp
@@ -21,6 +21,7 @@
  */
 #include <memory.h>
 #include <fstream>
+#include <float.h>
 #include "tvgLoader.h"
 #include "tvgTvgLoader.h"
 
@@ -70,6 +71,8 @@ bool TvgLoader::readHeader()
     ptr += SIZE(float);
     READ_FLOAT(&h, ptr);
     ptr += SIZE(float);
+    
+    if (w < FLT_EPSILON || h < FLT_EPSILON) return false;
 
     //Decide the proper Tvg Binary Interpreter based on the current file version
     if (this->version >= 0) interpreter = new TvgBinInterpreter;


hermet:
Saver doesn't allow 0 size view data also we use the file signature at least.

I think this doesn't so helpful, if the file is corrupted.
Rather than it must be prevented by checksum way if we need a stronger way,

mmaciola:
I actually had two problems regarding tvg loader developing WASM:
1. readHeader() doesn't check if ptr exceeded size applied in load(data, size). Size was incorrect as there was strlen used instead size(). We can check such scenario (size exceed after header validation) and inform developer with LOG message on error.
2. String applied to wasm is automatically converted to UTF8: every char above 0x7F is converted into two up to six chars. There is no such character in header frame, so the problem affect reading main part. Sometimes float number of size included such and then value was mostly converted to negative. We could include such character in header frame (eg. in reserved part) and validate if [0xff 0x00] is [0xff 0x00].
What do you think?

hermet:
> I actually had two problems regarding tvg loader developing WASM:
> 
>     1. readHeader() doesn't check if ptr exceeded size applied in load(data, size). Size was incorrect as there was strlen used instead size(). We can check such scenario (size exceed after header validation) and inform developer with LOG message on error.
> 
>     2. String applied to wasm is automatically converted to UTF8: every char above 0x7F is converted into two up to six chars. There is no such character in header frame, so the problem affect reading main part. Sometimes float number of size included such and then value was mostly converted to negative. We could include such character in header frame (eg. in reserved part) and validate if [0xff 0x00] is [0xff 0x00].
>        What do you think?

1) I agree on adding preventing exceed of the buffer size by checking ptr <= size in the decoding interation.
2) Sorry I can't catch up your question. "String applied to wasm" ? in this context, what is the String source?

mmaciola:
2. It's a feature of emscripten. Just for information: If there is a data buffer in javascript and it won't be converted to eg. Int8Array in explicit way, it will be treated as text string and converted to UTF8 when calling cpp function with this buffer used as an argument. I don't know if any user may face similar problem, but it may be anyway protected

hermet:
>     1. It's a feature of emscripten. Just for information: If there is a data buffer in javascript and it won't be converted to eg. Int8Array in explicit way, it will be treated as text string and converted to UTF8 when calling cpp function with this buffer used as an argument. I don't know if any user may face similar problem, but it may be anyway protected

Ok... yet i don't understand how this patch helps for it.

You mentioned "0x7F is converted into two up to six chars" 
If 0x7F data is stored then what data we are expecting the value is converted to? and how your patch helps that?

hermet:
@mmaciola please check this patch. it could fix your problem.

PR #736, opened 2021-08-23
sw_engine SwShape: Refer to number of points, when create outline

JSUYA:
When appendPath is called, the size of the pts array is estimated based on the cmd list.
If cmd array is greater than pts array, when referencing pts in cmd,
it access to out of array of pts.
So refer to ptsCnt together with cmdCnt.


JSUYA:
Because outline.pts created using realloc() in _growOutlinePoint() is not initialized to {0, 0}, it is expected that it will eventually cause the same problem.

PR #737, opened 2021-08-24
sw_engine SwShape: Fix incorrect Point reference in Command

JSUYA:
When appendPath is called, the size of the pts array is estimated based on the cmd list.
If cmd array is greater than pts array, when referencing pts in cmd,
it access to out of array of pts.
If cmd needs to reference a point that is out of pts size, use 0, 0.


hermet:
Please have a look -> https://github.com/Samsung/thorvg/pull/738

PR #738, opened 2021-08-24
sw_engine shape: prevent crash when wrong pair of commands & points.

hermet:
add an exception handling,
if the given points count is not matched with expected points count from commands.

@Issue: https://github.com/Samsung/thorvg/issues/735


JSUYA (commented):
This way may not reliably cause a crash.

However, it is a problem caused by incorrect input of appendPath, and the user cannot check whether there is a problem from the Result of appendPath().
And there doesn't seem to be a way to tell the FW(the upper layer) the wrong pair.
That's why I don't think it's the right way.

If you want to keep this method, please apply it to _genDashOutline() as well.

PR #743, opened 2021-08-24
tvg_loader: make sure the pointer does not exceed the size

mmaciola:


src/loaders/tvg/tvgTvgLoader.cpp
```diff
@@ -91,6 +91,9 @@ bool TvgLoader::readHeader()
 
     ptr += TVG_HEADER_COMPRESS_SIZE;
 
+    //Make sure the pointer does not exceed the size now
+    if (ptr > this->data + this->size) return false;
```
hermet, line 95: I think we can return immediately in the beginning of readHeader()
if (size < TVG_HEADER_SIZE) return false;

if the wrong size is given, it can be less than 1.

hermet (commented):
Could you please check my comment?

src/loaders/tvg/tvgTvgLoader.cpp
mmaciola: ok done

PR #744, opened 2021-08-24
sw_engine avx: addad ALPHA_BLEND_AVX and avxRasterTranslucentRect fun…

mgrudzinska:
…ctions

Implemented using 128-bit registers.
avxRasterTranslucentRect is around 5 times faster than cRasterTranslucentRect (i7-8700 CPU - Coffe Lake)

### DIFF
--- a/src/lib/sw_engine/tvgSwRasterAvx.h
+++ b/src/lib/sw_engine/tvgSwRasterAvx.h
@@ -24,28 +24,107 @@
 
 #include <immintrin.h>
 
+#define N_32BITS_IN_128REG 4
+
+static inline __m128i ALPHA_BLEND(__m128i c, __m128i a)
+{
+    //1. set the masks for the A/G and R/B channels
+    auto AG = _mm_set1_epi32(0xff00ff00);
+    auto RB = _mm_set1_epi32(0x00ff00ff);
+
+    //2. mask the alpha vector - originally quartet [a, a, a, a]
+    auto aAG = _mm_and_si128(a, AG);
+    auto aRB = _mm_and_si128(a, RB);
+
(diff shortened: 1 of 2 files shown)


hermet:
@mgrudzinska do you have any test result of performance enhancement?

mgrudzinska:
@hermet yes, I'll post it later and let you know, since I would appreciate you comment on this

src/lib/sw_engine/tvgSwRasterAvx.h
```diff
@@ -24,6 +24,29 @@
 
 #include <immintrin.h>
 
+#define N_32BITS_IN_128REG 4
+
+static inline __m128i ALPHA_BLEND_AVX(__m128i c, __m128i a)
```
hermet, line 29: "AVX" in the naming  is not quite necessary since this implementation is in the AVX local space.

hermet:
> @hermet yes, I'll post it later and let you know, since I would appreciate you comment on this

coding convention wise, well no problem to me.

just, it would be better if the basic info is described in the commit message.
and please see avxRasterRGBA32() code, how I wrote the proper comments for the readers.
avx is a bit far from the default programming grammar, it would be nicer adding more kind comments.

src/lib/sw_engine/tvgSwRasterAvx.h
mgrudzinska: I tried to use the naming convention from the neon version of the file. I'll change it

hermet:
@mgrudzinska Great, I didn't know -03 generates avx simd automatically, absolutely we need a method to detect the architecture and have a decision which size of simd we use. I expect you can make it.

hermet (approved):

Issue #741, opened 2021-08-24
Size() doesn't work for tvg loaded scene


mmaciola:
Set size doesn't work for .tvg.

```
    auto picture = tvg::Picture::gen();
    if (picture->load(EXAMPLE_DIR"/test.tvg") != tvg::Result::Success) {
        cout << "Couldn't load svg image!" << endl;
        return;
    }

    picture->size(WIDTH / 2, HEIGHT / 2);
    canvas->push(move(picture));
```

PR #745, opened 2021-08-25
common picture: quick fix broken bounds() interface.

hermet:
picture must return the boundary info - 0, 0, w, h
We assume that it has a designated picture size.

Aside from this issue,
bounds() api must be reviewed, its behavior is quite in a trouble...
unless the result is not transformed, its information is useless...

@issue: #741


src/examples/TvgSaver.cpp
```diff
@@ -108,15 +108,20 @@ void exportTvg()
     free(data);
 
     //nested paints
```
hermet, line 110: Please ignore the examples code change. 
The logics doesn't changed anything.

PR #740, opened 2021-08-24
sw_engine: code refactoring.

hermet:
simplify neon code and fix the wrong range access.

PR #747, opened 2021-08-25
sw_engine neon: fix the iterations number of 'neon calls'

mgrudzinska:
In the case, when the dst memory was not aligned,
we went out of the dst range in the 'neon loop'.

### DIFF
--- a/src/lib/sw_engine/tvgSwRasterNeon.h
+++ b/src/lib/sw_engine/tvgSwRasterNeon.h
@@ -99,23 +99,25 @@ static inline bool neonRasterTranslucentRect(SwSurface* surface, const SwBBox& r
     auto vColor = static_cast<uint8x8_t>(vdup_n_u32(color));
     auto vIalpha = static_cast<uint8x8_t>(vdup_n_u8((uint8_t) ialpha));
     uint8x8_t* vDst = nullptr;
+    uint32_t align;
 
     for (uint32_t y = 0; y < h; ++y) {
-
         auto dst = &buffer[y * surface->stride];
 
         if ((((uint32_t) dst) & 0x7) != 0) {
             //fill not aligned byte
             *dst = color + ALPHA_BLEND(*dst, ialpha);
             vDst = (uint8x8_t*) (dst + 1);
+            align = 1;
         } else {
             vDst = (uint8x8_t*) dst;
+            align = 0;
         }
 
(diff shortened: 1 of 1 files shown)


src/lib/sw_engine/tvgSwRasterNeon.h
```diff
@@ -101,28 +101,28 @@ static inline bool neonRasterTranslucentRect(SwSurface* surface, const SwBBox& r
...
-        uint32_t left = w % 2;
-
-        if (align) *dst = color + ALPHA_BLEND(*dst, ialpha);
+        uint32_t iterations = (w - align) / 2;
+        uint32_t left = (w - align) % 2;
 
         for (uint32_t x = 0; x < iterations; ++x) vDst[x] = vadd_u8(vColor, ALPHA_BLEND_NEON(vDst[x], vIalpha));
-        
+
         if (left) dst[w] = color + ALPHA_BLEND(dst[w], ialpha);
```
mgrudzinska, line 123: out of the range - solved in #740

src/lib/sw_engine/tvgSwRasterNeon.h
```diff
@@ -101,28 +101,28 @@ static inline bool neonRasterTranslucentRect(SwSurface* surface, const SwBBox& r
...
-            vDst = (uint8x8_t*) dst;
+            align = false;
         }
+        vDst = (uint8x8_t*)dst;
 
-        uint32_t iterations = w / 2;
-        uint32_t left = w % 2;
-
-        if (align) *dst = color + ALPHA_BLEND(*dst, ialpha);
+        uint32_t iterations = (w - align) / 2;
```
hermet, line 118: looks bad  int - boolean.

hermet (changes requested):

src/lib/sw_engine/tvgSwRasterNeon.h
mgrudzinska: jap, I think that too :D
I wrote that since you did the same in #740 in neonRasterTranslucentRle
Please merge yours when it's ready and then I'll change this one.

hermet: it's bad showcase. you'd rather -1 for it. (even I didn't know it)

hermet (approved):

PR #750, opened 2021-08-25
tvg format: added validation flag (0x80) always set

mmaciola:
Added single validation flag (always set, 0x80) that may help against the problems described in thread [#733](https://github.com/Samsung/thorvg/pull/733)

### DIFF
--- a/src/loaders/tvg/tvgTvgLoader.cpp
+++ b/src/loaders/tvg/tvgTvgLoader.cpp
@@ -70,6 +70,7 @@ bool TvgLoader::readHeader()
     ptr += SIZE(float);
 
     //4. Reserved
+    if (!(*ptr & TVG_HEAD_FLAG_VALIDATION)) return false;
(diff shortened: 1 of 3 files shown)


src/loaders/tvg/tvgTvgLoader.cpp
```diff
@@ -70,6 +70,7 @@ bool TvgLoader::readHeader()
     ptr += SIZE(float);
 
     //4. Reserved
+    if (!(*ptr & TVG_HEAD_FLAG_VALIDATION)) return false;
     if (*ptr & TVG_HEAD_FLAG_COMPRESSED) compressed = true;
```
hermet, line 74: We already put the signature of the file (the minimum checking for the file type), this kind of similar validation flag is just duplicated in its purpose, not prevent the error case 100%. If you think validation is necessary, then we should introduce the perfect validation check such as CRC(Cyclic Redundancy Check)

hermet (changes requested):

PR #753, opened 2021-08-27
sw_engine avx: the not aligned memory separated in the rasterRGBA32 func

mgrudzinska:
The aligned and not aligned memory parts handled separately.
This gives around 10% of the performance increase (tested on i7-8700 CPU)

### DIFF
--- a/src/lib/sw_engine/tvgSwRasterAvx.h
+++ b/src/lib/sw_engine/tvgSwRasterAvx.h
@@ -25,6 +25,7 @@
 #include <immintrin.h>
 
 #define N_32BITS_IN_128REG 4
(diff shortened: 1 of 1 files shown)


hermet:
@mgrudzinska Could you please suggest any issue buffer position/size?

mgrudzinska:
@hermet I forced a push, since the conflicts were solved in a wrong way.

I tested the performance on 400x400 rect and 4000x4000. For small sizes, like 40x40, the results seams to be worse and maybe it's not a good idea to apply this patch. 
But the measurements are unstable - I didn't use any stabilizer... so I need to repeat the tests to be sure the numbers.

hermet (approved):

PR #759, opened 2021-08-30
docs: minor linguistic changes

mgrudzinska:


README.md
```diff
@@ -286,9 +286,9 @@ Examples:
     $ svg2png input1.svg input2.svg -r 200x200 -b ff00ff
     $ svg2png . -r 200x200
 ```
-Multiple files or directories are accepted for `svgFileName`. If a directory is applied, subdirectories are also searched. Looking for `.svg` file extension.
+Multiple files or directories can be passed as the `SVG files` parameter. If a directory is passed, its subdirectories are also searched for `.svg` file extension.
```
mmaciola, line 289: ` its subdirectories are also searched for .svg file extension` - It is another meaning: every file is looked for .svg extension, not only files in the directory and its subdirectories.
If want to use svg2png, files must have proper .svg extension. If pass a single file but without .svg ext, it will print an error. If passed from directory, none .svg files will be skipped.

mmaciola (changes requested):

README.md
mgrudzinska: @mmaciola another than what? can you please suggest how it should sound?
I passed a directory, all its subdirs were searched. as it is in the description.

mgrudzinska: also '#ffffff' format is not supported

mmaciola: Please move the sentence about the .svg extension to a different sentence than the sentence about directories

mgrudzinska: @mmaciola please check now

PR #760, opened 2021-08-30
svg2png: fixing the args interpretation

mgrudzinska:
The background color was misinterpreted and its value was treated
as an svg file name (warning was returned).
The color format can not be passed in a "#ffffff" format, since
the '#' sign is used in bash to comment a line and nothiing after
it is loaded.

```./svg2png ../../../../src/examples/images/bojo.svg -b ff0000```

before:
12before

```
Generated PNG file: /home/m.grudzinska/thorvg/src/examples/images/bojo.png
Warning: File "ff0000" is invalid.
```

after:
12after

```
./svg2png ../../../../src/examples/images/bojo.svg -b ff0000
Generated PNG file: /home/m.grudzinska/thorvg/src/examples/images/bojo.png
```

### DIFF
--- a/src/bin/svg2png/svg2png.cpp
+++ b/src/bin/svg2png/svg2png.cpp
@@ -177,7 +177,7 @@ public:
             const char* p = argv[i];
             if (*p == '-') {
                 //flags
-                const char* p_arg = (i + 1 < argc) ? argv[i] : nullptr;
+                const char* p_arg = (i + 1 < argc) ? argv[++i] : nullptr;
                 if (p[1] == 'r') {
                     //image resolution
                     if (!p_arg) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/bin/svg2png/svg2png.cpp [hunk 1]
```diff
@@ -198,12 +198,11 @@ struct App
                 } else if (p[1] == 'b') {
                     //image background color
                     if (!p_arg) {
-                        cout << "Error: Missing background color attribute. Expected eg. -b #fa7410." << endl;
+                        cout << "Error: Missing background color attribute. Expected eg. -b fa7410." << endl;
                         return 1;
                     }
 
-                    if (*p_arg == '#') ++p_arg;
```
mmaciola, line 205: I prefer to leave it. The rest is ok.

mmaciola (changes requested):

src/bin/svg2png/svg2png.cpp
mgrudzinska: do you have some arguments for that? 
bash interprets the # sign in a  given way and its not the way we expect here

PR #766, opened 2021-09-02
sw_engine: handling clips from outside the rendering region

mgrudzinska:
Clips from outside the rendering region were ignored resulting
in rendering shapes, which should be completely clipped.
In case of the clip fast track, the bounding box width/height
overflow occured and shape, that should be clipped, could be
visible (not always).

before:
15before

after (black screen, as expected):
15after

code:
```
    if (!canvas) return;

    auto star1 = tvg::Shape::gen();
    star1->appendRect(400, 400, 300, 300, 30, 30);
    star1->fill(255, 255, 0, 255);
    star1->stroke(255 ,0, 0, 255);
    star1->stroke(10);

    auto clipStar = tvg::Shape::gen();
    clipStar->appendRect(-450, 430, 110, 110, 0, 0);
    clipStar->fill(255, 255, 255, 255);

    star1->composite(move(clipStar), tvg::CompositeMethod::ClipPath);
    canvas->push(move(star1));


    string path(EXAMPLE_DIR"/rawimage_200x300.raw");

    ifstream file(path);
    if (!file.is_open()) return ;
    auto data = (uint32_t*)malloc(sizeof(uint32_t) * (200*300));
    file.read(reinterpret_cast<char *>(data), sizeof (uint32_t) * 200 * 300);
    file.close();

    if (picture->load(data, 200, 300, true) != tvg::Result::Success) return;
    free(data);

    auto clipPath = tvg::Shape::gen();
    clipPath->appendRect(-19, 30, 10, 100, 10, 10);
    clipPath->fill(255, 255, 255, 255); // clip object must have alpha.

    picture->composite(move(clipPath), tvg::CompositeMethod::ClipPath);
    canvas->push(move(picture));


    shape1->appendRect(300, 0, 400, 400, 0, 0);
    shape1->fill(255, 0, 255, 255);
    shape1->stroke(255, 0, 255, 255);
    shape1->stroke(5);

    auto clipShape1 = tvg::Shape::gen();
    clipShape1->appendRect(-19, 30, 10, 100, 10, 10);
    clipShape1->fill(255, 0, 0, 255);

    shape1->composite(move(clipShape1), tvg::CompositeMethod::ClipPath);
    canvas->push(move(shape1));
```

### DIFF
--- a/src/lib/sw_engine/tvgSwRenderer.cpp
+++ b/src/lib/sw_engine/tvgSwRenderer.cpp
@@ -143,11 +143,13 @@ struct SwShapeTask : SwTask
             if (shape.rle) {
                 if (clipper->rect) rleClipRect(shape.rle, &clipper->bbox);
                 else if (clipper->rle) rleClipPath(shape.rle, clipper->rle);
+                else goto err;
             }
             //Clip stroke rle
             if (shape.strokeRle) {
                 if (clipper->rect) rleClipRect(shape.strokeRle, &clipper->bbox);
                 else if (clipper->rle) rleClipPath(shape.strokeRle, clipper->rle);
+                else goto err;
             }
         }
         goto end;
@@ -183,7 +185,7 @@ struct SwImageTask : SwTask
             imageReset(&image);
 
             image.data = const_cast<uint32_t*>(pdata->data(&image.w, &image.h));
-            if (!image.data || image.w == 0 || image.h == 0) goto end;            
+            if (!image.data || image.w == 0 || image.h == 0) goto end;
 
             if (!imagePrepare(&image, transform, clipRegion, bbox, mpool, tid)) goto end;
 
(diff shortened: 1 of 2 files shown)


src/lib/sw_engine/tvgSwRenderer.cpp
```diff
@@ -143,11 +143,21 @@ struct SwShapeTask : SwTask
             if (shape.rle) {
                 if (clipper->rect) rleClipRect(shape.rle, &clipper->bbox);
                 else if (clipper->rle) rleClipPath(shape.rle, clipper->rle);
+                else {
```
hermet, line 146: this looks not good to access internal rle data in renderer.
probably.... -> goto err?
also, in this case it's not meaning to iterate for loop, you can break it immediately.

hermet (changes requested):
please check my comments.

hermet (approved):

PR #769, opened 2021-09-02
utc: loading and saving .tvg

mmaciola:
Added tests of loading and saving .tvg.
Created testSavers.cpp. Created files tag.tvg and test.tvg. Minor change
in tag.svg.


hermet (approved):
good, thanks!

PR #772, opened 2021-09-05
tvg_loader: code refactoring & optimization loading.

hermet:
Skip the reassigning values by the flags,
We assume those property values are reserved,
leave to apis that handle the invalids if the exceptions.


mgrudzinska (approved):
is ok

mmaciola (approved):

Issue #783, opened 2021-09-08
tvg format issues


hermet:
we have 3 issues observed in tvg format.
1. picture alignment : in case of svg, we manually reset the position by checking the viewbox, in case of tvg that part is missed. this issue might have a relation with https://github.com/Samsung/thorvg/issues/746
2. texts scaling (samsung & glaxy logo size issue).
3. one picture is not visible (in the middle of the screenshot).  need to figure out which one is. 

![shot-2021-09-08_14-00-05](https://user-images.githubusercontent.com/3711518/132449198-b23c2743-7de0-4e83-9a2f-b4cb2572673d.jpg)

mgrudzinska:
ad 3. gradient_stroke.tvg was missing, #784 fixes it

mgrudzinska:
4. the tiger image looks a bit different after saving into tvg format

tvg:
![22before](https://user-images.githubusercontent.com/67589014/132671369-b891758c-c1fa-4646-9363-735b7a289e9d.PNG)

svg:
![22after](https://user-images.githubusercontent.com/67589014/132671395-aefba80e-a156-4772-a2fa-8f85ec33b6f2.PNG)

mgrudzinska:
5. gallardo
![23galard](https://user-images.githubusercontent.com/67589014/132677354-8a47d983-7cf1-4f02-94cd-f58024bd8aef.PNG)

6. code:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
  <path d="M57,11c40-22,42-2,35,12c8-27-15-20-30-11z" fill="#47b" fill-rule='evenodd'/>
  <path d="M36,56h56c4-60-83-60-86-6c13-16,26-26,36-30l-29,53c20,23,64,26,79-12h-30c0,14-26,12-25-5zM37,43c0-17,26-17,26,0zM39,89c-10,7-42,15-26-16l29-52c-15,6-36,40-37,48c-12,35,14,37,37,20" fill="#47b" fill-rule='evenodd'/>
</svg>
```
![24msie](https://user-images.githubusercontent.com/67589014/132680952-35c44b5c-b421-4a27-9495-aa52eec617ab.PNG)

7. part of the original code:
```
<svg xmlns="http://www.w3.org/2000/svg" id="Penrose_Staircase" viewBox="0 0 372 283">
  <desc>A Penrose Staircase (Optical Illusion)</desc>
    <g id="Layer_1">
[...]

PR #784, opened 2021-09-08
tvg_saver: fix the stroke width scaling factor

mgrudzinska:
To get the scaling factor from the transformation matrix we have
to use two of its elements.

this patch fixes the gradient_stroke.tvg problem from 
issue: #783

### DIFF
--- a/src/savers/tvg/tvgTvgSaver.cpp
+++ b/src/savers/tvg/tvgTvgSaver.cpp
@@ -451,7 +451,7 @@ TvgBinCounter TvgSaver::serializeStroke(const Shape* shape, const Matrix* pTrans
 
     //width
     auto width = shape->strokeWidth();
-    if (preTransform) width *= pTransform->e11;  //we know x/y scaling factors are same.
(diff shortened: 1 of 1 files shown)


src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -451,7 +451,7 @@ TvgBinCounter TvgSaver::serializeStroke(const Shape* shape, const Matrix* pTrans
 
     //width
     auto width = shape->strokeWidth();
-    if (preTransform) width *= pTransform->e11;  //we know x/y scaling factors are same.
+    if (preTransform) width *= sqrt(pow(pTransform->e11, 2) + pow(pTransform->e21, 2));  //we know x/y scaling factors are same.
```
hermet, line 454: no i think pre-condition is missed at some points, because here xy shouldn't be different...

mgrudzinska: I understand that the scaling in both, x and y axis, are the same
but element e11 contains informations about the scaling and the rotation, that why we can not use only this value

hermet: aha got it!

hermet (approved):

PR #785, opened 2021-09-08
sw_engine image: fix the clipped image issue.

hermet:
there was a wrong condition introduced the bug that image was not updated,
because transformation is not re-applied after the first created the internal image data.

@Issues: https://github.com/Samsung/thorvg/issues/751


mmaciola (approved):
Tested and works correctly. Approve
https://mmaciola.github.io/thorvg.viewer.2/?s=https://raw.githubusercontent.com/Samsung/thorvg/master/src/examples/images/test.tvg

PR #792, opened 2021-09-09
common: proper scaling of the tvg image

mgrudzinska:
The svg image size was stored in the 'vw' and 'vh'
members of the loader. For all other image formats, this
information was stored in 'w' and 'h' variables.
As a result the tvg file didn't contain the proper
information about an image size.

To check whether the image has to be resized, we compared w and h with loader->w and loader->h, but then we used loader->vw and loader->vh to scale the image.

It was responsible for the #783 pt 2 issue.

### DIFF
--- a/src/lib/tvgLoadModule.h
+++ b/src/lib/tvgLoadModule.h
@@ -30,11 +30,7 @@ namespace tvg
 class LoadModule
 {
 public:
-    //default view box, if any.
-    float vx = 0;
-    float vy = 0;
-    float vw = 0;
-    float vh = 0;
+    float x = 0, y = 0;         //default top left corner of the image
     float w = 0, h = 0;         //default image size
     bool preserveAspect = true; //keep aspect ratio by default.
 
--- a/src/loaders/svg/tvgSvgLoader.h
+++ b/src/loaders/svg/tvgSvgLoader.h
@@ -31,6 +31,7 @@ public:
     string filePath;
     const char* content = nullptr;
     uint32_t size = 0;
+    float width = 0.0f, height = 0.0f; //svg tag attributes
 
     SvgLoaderData loaderData;
     unique_ptr<Scene> root;
--- a/src/lib/tvgPictureImpl.h
+++ b/src/lib/tvgPictureImpl.h
@@ -152,10 +152,10 @@ struct Picture::Impl
     bool viewbox(float* x, float* y, float* w, float* h) const
     {
         if (!loader) return false;
-        if (x) *x = loader->vx;
-        if (y) *y = loader->vy;
-        if (w) *w = loader->vw;
-        if (h) *h = loader->vh;
+        if (x) *x = loader->x;
+        if (y) *y = loader->y;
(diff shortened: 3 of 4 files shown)


src/loaders/svg/tvgSvgLoader.h
```diff
@@ -31,6 +31,7 @@ class SvgLoader : public LoadModule, public Task
     string filePath;
     const char* content = nullptr;
     uint32_t size = 0;
+    float width = 0.0f, height = 0.0f; //svg tag attributes
```
mgrudzinska, line 34: I kept for now (or rather moved from the LoadModule) 'width' and 'height' members. 
ThorVG doesn't support scaling of the image with these values. Maybe after we fix it, we decide to remove these two.

src/lib/tvgLoadModule.h
```diff
@@ -30,11 +30,7 @@ namespace tvg
 class LoadModule
 {
 public:
-    //default view box, if any.
-    float vx = 0;
```
hermet, line 34: Please fix the only wrong convention. otherwise, there is no point this kind of changes.

I think you can just move "w/h" to svg side instead of width/height.

hermet (changes requested):
the root change is good to me but Please check a comment.

src/lib/tvgLoadModule.h
mgrudzinska: in the loader we have:
- 'vx, vy, vw, vh' as the viewbox parameters
- and 'w, h' that are the width and height for all types of images, except the svgs (so I don;t want to move these two from the loadModule to the svg loader)
For the svg file, the real width and height of the image are the width and height from the 'viewbox' attribute. at least it is how I understand it... 


For now there are still some issues regarding this subject, but it's a different story I think:
```
<!DOCTYPE html>
<html>

<head>
	<title>SVG viewBox Attribute</title>
	<style type="text/css">
		svg {
			border: 1px solid #aaa;
		}
	</style>
</head>

<body>

	<svg width="200" height="200">
		<circle cx="50" cy="50" r="45" stroke="#000" stroke-width="3" fill="none"/>
	</svg>
    
	<svg>
		<circle cx="50" cy="50" r="45" stroke="#000" stroke-width="3" fill="none"/>
	</svg>
	
	<svg width="200" height="200" viewBox="0 0 100 100">
		<circle cx="50" cy="50" r="45" stroke="#000" stroke-width="3" fill="none"/>
	</svg>
 
        <svg viewBox="0 0 100 100">
		<circle cx="50" cy="50" r="45" stroke="#000" stroke-width="3" fill="none"/>
	</svg>
    
       <svg viewBox="0 0 200 200">
		<circle cx="50" cy="50" r="45" stroke="#000" stroke-width="3" fill="none"/>
	</svg>
</body>

</html>
```

mgrudzinska:
I'll think one more time about width/height/viewbox...

hermet:
https://github.com/Samsung/thorvg/pull/801

mgrudzinska:
solved in #805

PR #793, opened 2021-09-09
common picture: changed x and y values returned by the bounds api

mgrudzinska:
For an svg file, the bounds are established by the viewbox values.

For shapes we establish not only the width and height as the bounds, but also the starting points.
For raw pictures or pngs, the starting points are zeros, but for an svg it can be any value.
This change could help us with removing in the future the viewbox api.

### DIFF
--- a/src/lib/tvgPictureImpl.h
+++ b/src/lib/tvgPictureImpl.h
@@ -169,10 +169,10 @@ struct Picture::Impl
 
     bool bounds(float* x, float* y, float* w, float* h) const
     {
-        if (x) *x = 0;
-        if (y) *y = 0;
-        if (w) *w = this->w;
-        if (h) *h = this->h;
+        if (x) *x = loader->vx;
+        if (y) *y = loader->vy;
+        if (w) *w = loader->vw;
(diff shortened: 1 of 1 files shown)


mgrudzinska:
I'll change the test tomorrow, after you'll comment that this change makes sense to you

src/lib/tvgPictureImpl.h
```diff
@@ -169,10 +169,10 @@ struct Picture::Impl
 
     bool bounds(float* x, float* y, float* w, float* h) const
     {
-        if (x) *x = 0;
-        if (y) *y = 0;
-        if (w) *w = this->w;
-        if (h) *h = this->h;
+        if (x) *x = loader->vx;
```
hermet, line 172: viewbox is only available to svg.
we can't assume these values are always valid.

I know... current bounds() api has an issue. 
This is also regarded to this issue. https://github.com/Samsung/thorvg/issues/746

Need to think about fundamental solution...

hermet (changes requested):

mgrudzinska:
#805 is enough for now

Issue #746, opened 2021-08-25
view size is not concrete in tvg saver.


hermet:
Currently tvg saver writes the view size using paint's bounds() api.
but Scene doesn't get the transformed children geometry, the returned boundary info is incorrect.
Thus picture->size() won't be scaled we expected.

hermet:
one more problem... tvg needs to add the origin of the shapes as well?
INPUT:
scene
 - shape [rect: 100, 100, 200, 200]

save scene to tvg
currenet tvg size => 200x 200
expected tvg size => 300 x 300

mgrudzinska:
isn;t this solved by #805  ?

PR #794, opened 2021-09-09
tvg saver: centering tvg images

mgrudzinska:
This change resolves issue with scaling images, that had
negative width or height values in the viewbox svg attribute.
Works after bounds api is fixed (separate commit).

the result can be seen after #793 is applied.
issue :#746


hermet:
the issue was fixed by https://github.com/Samsung/thorvg/pull/805

mgrudzinska:
#805 is enough for now

PR #798, opened 2021-09-11
tvg_saver: proper serialization of a grad fill

mgrudzinska:
In the case when a gradinet fill is present, the transformation matrix
can't be preapplied.

issue #783 pt 5 (gallardo.svg)


mgrudzinska:
let's wait with merging, since for now the svg result is also wrong, so there's no point to compare tvg result with the wrong one

hermet (approved):

PR #800, opened 2021-09-14
common Picutre: Call resize when image loadded.

JSUYA:
If the size is not set in the picture,
SvgLoader and SvgSceneBuilder build the scene based on the viewbox.
An appropriate size set is required for the width and height set afterwards,
but it is inappropriate to transform the root at the loader level.
This is because the paint object of the picture is the root.

Therefore, when load() is finished, resizing can be called according
to the width, height, and size of the viewbox.

issue: https://github.com/Samsung/thorvg/issues/799

### DIFF
--- a/src/lib/tvgPictureImpl.h
+++ b/src/lib/tvgPictureImpl.h
@@ -173,7 +173,7 @@ struct Picture::Impl
         if (y) *y = 0;
         if (w) *w = this->w;
         if (h) *h = this->h;
- 
+
         return true;
     }
 
@@ -197,6 +197,7 @@ struct Picture::Impl
         if (!loader->read()) return Result::Unknown;
         w = loader->w;
         h = loader->h;
+        if (w != loader->vw || h != loader->vh) resizing = true;
         return Result::Success;
     }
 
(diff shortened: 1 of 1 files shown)


hermet (changes requested):
Could you please double-check  uint32_t reload() method?
If it's necessary, this condition must be there 
*since viewbox is only used vector image.

mgrudzinska:
@JSUYA could you please elaborate on why we can not set an additional transformation matrix in the svg loader? 
we clip the image already there and it's ok, so I don't understand why we can't set an extra transform there.

JSUYA:
> @JSUYA could you please elaborate on why we can not set an additional transformation matrix in the svg loader?
> we clip the image already there and it's ok, so I don't understand why we can't set an extra transform there.

I didn't think about using SvgLoader's `reisze()` method.
After SceneBuild, I tried to set the ratio of `viewbox`, and `width,height` to the transform of paint(root), and I thought that there would be a problem if it was overwritten after calling the `resize()` function.
Actually, that was wrong and I decided to close this PR.
https://github.com/Samsung/thorvg/pull/801 I think this patch is a more reasonable.

Issue #799, opened 2021-09-14
svg default size is incorrect


hermet:
When loading "samsung-7.svg" in the examples,
We expect its default size 2500 x 2500 but it's scaled down to (192.755997 192.755997) which is viewbox size.

```
auto pic = tvg::Picture::gen();
pic->load(EXAMPLE_DIR"/samsung-7.svg");
canvas->push(move(pic));
```

```
<svg xmlns="http://www.w3.org/2000/svg" width="2500" height="2500" viewBox="0 0 192.756 192.756">
```

It looks the paths of svg is scaled down to 192x192 in default.

PR #801, opened 2021-09-14
svg_loader: fix the wrong paint size.

hermet:
we expect that image loader returns the vector image in the default size.
but currently it doesn't if the viewbox and default image size of svg is not matched,

Thus, update the paint with the the default size at the end of the loading.

This also fixes the tvg sizing issue which image size and actual paint size are not matched.

@Issue: https://github.com/Samsung/thorvg/issues/783
@Issue: https://github.com/Samsung/thorvg/issues/799


mgrudzinska:
@hermet please check #805

I think resizing should be done before the image is clipped.
also I think that resizing doesn't need to compare the size with viewbox values. 
as the next step I'd like to remove viewbox api - as we already once did. 

anyway, please fill free to edit your pr and close mine, it was a bit easier to push a patch than explain

hermet:
@mgrudzinska there is no reason keep my patch if your version is necessary. Just wondering your patch will also help for these two issues. 
@issue: #783
@issue: #799

Issue #791, opened 2021-09-09
Target buffer in unpremultiplicated form


mmaciola:
Target buffer is now returned in a premultiplicated form.
Typical use scenario of a thorvg is that the background color is opaque and the target buffer is directly displayed on a screen. In this scenario, there is no need to think about the buffer premultiplication - for opaque background, premultiplicated and not premultiplicated buffer is the same.

However, there are use cases where background is not opaque - transparent or semitransparent - and the user would like to save the output buffer from thorvg into png or use with another renderer. In this scenario there is a need for unpremultiplicated buffer.

I am thinking that there should be unpremultiplication option

PR #802, opened 2021-09-14
common: Unmultiplicated colorspace

mmaciola:
This patch introduces lowercase colorspaces (abgr8888 and argb8888) whose
colors are un-alpha-premultiplied. Unmultiplicated colors are especially
needed for wasm thorvg viewer and svg2png / tvg2png.
Only C version now.

@issue: #791

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1196,8 +1196,10 @@ public:
      */
     enum Colorspace
     {
-        ABGR8888 = 0, ///< The channels are joined in the order: alpha, blue, green, red.
-        ARGB8888      ///< The channels are joined in the order: alpha, red, green, blue.
+        ABGR8888 = 0,      ///< The channels are joined in the order: alpha, blue, green, red. Colors are alpha-premultiplied.
+        ARGB8888,          ///< The channels are joined in the order: alpha, red, green, blue. Colors are alpha-premultiplied.
+        ABGR8888_STRAIGHT, ///< @BETA_API The channels are joined in the order: alpha, blue, green, red. Colors are un-alpha-premultiplied.
+        ARGB8888_STRAIGHT, ///< @BETA_API The channels are joined in the order: alpha, red, green, blue. Colors are un-alpha-premultiplied.
     };
 
     /**
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -1245,10 +1245,10 @@ void rasterRGBA32(uint32_t *dst, uint32_t val, uint32_t offset, int32_t len)
 
 bool rasterCompositor(SwSurface* surface)
 {
-    if (surface->cs == SwCanvas::ABGR8888) {
+    if (surface->cs == SwCanvas::ABGR8888 || surface->cs == SwCanvas::ABGR8888_STRAIGHT) {
         surface->blender.alpha = _colorAlpha;
         surface->blender.join = _abgrJoin;
-    } else if (surface->cs == SwCanvas::ARGB8888) {
+    } else if (surface->cs == SwCanvas::ARGB8888 || surface->cs == SwCanvas::ARGB8888_STRAIGHT) {
         surface->blender.alpha = _colorAlpha;
         surface->blender.join = _argbJoin;
     } else {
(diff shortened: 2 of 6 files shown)


src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -1361,6 +1361,31 @@ bool rasterClear(SwSurface* surface)
...
+            } else if (a == 0) {
+                buffer[x] = 0x00ffffff;
+            } else {
+                uint16_t r = ((buffer[x] >> 8) & 0xff00) / a;
+                uint16_t g = ((buffer[x]) & 0xff00) / a;
+                uint16_t b = ((buffer[x] << 8) & 0xff00) / a;
+                if (r > 0xff) r = 0xff;
+                if (g > 0xff) g = 0xff;
+                if (b > 0xff) b = 0xff;
+                buffer[x] = (a << 24) | (r << 16) | (g << 8) | (b);
```
JSUYA, line 1382: um... I'm not sure if it's appropriate to use  ALPHA_BLEND. But I think it can be expressed like this.
 ALPHA_BLEND(buffer[x], surface->blender.alpha(buffer[x])) 
Or maybe you could just use ALPHA_MULTIPLY for each rgb. What do you think?

JSUYA (commented):
I think.. It would be better to add an option to premultiplied/none-premultiplied rather than lowercase the colorspace for target() api (use default parameter?).
@hermet What do you think?

src/lib/sw_engine/tvgSwRaster.cpp
mmaciola: For unpremultiplication, the color values are divided, not multiplied. I am not sure it can be optimized.

hermet:
Thinking from the origin...

The best way is not to generate pre-multipled values in the first rendering sequence...
This looks a workaround or application perspective approach because users can do this with a utility function. such a unpremultiply(Buffer);

To make it sure, back to the origin, svg2png is necessary this... ? Can you please elaborate it?

mmaciola:
Un-pre-multiplicated buffer is needed. `lodepng::encode` that is used in svg2png expects not pre-multipliacted buffer. Try with edited rotlines.svg:
`svg2png src/examples/images/rotlines.svg -r 512x512`

rotlines.svg:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
  <g transform="rotate(-45 50 50)">
    <rect x="35%" width="10%" height="100%" fill="black"/>
    <rect x="45%" width="10%" height="100%" fill="red" opacity="50%"/>
    <rect x="55%" width="10%" height="100%" fill="white"/>
  </g>
</svg>
```

rotlines

mgrudzinska:
@mmaciola I slightly edited this svg (two red lines one with the opacity and one without it)
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
  <g transform="rotate(-45 50 50)">
    <rect x="35%" width="10%" height="100%" fill="black"/>
    <rect x="45%" width="10%" height="100%" fill="red" opacity="50%"/>
    <rect x="55%" width="10%" height="100%" fill="red"/>
  </g>
</svg>
```
and the result is as follows:
rr

isn't it correct?

mmaciola:
Yes, the image is correct, but it is not an image produced by svg2png. 
Image produced by svg2png:
rotlines

mgrudzinska:
before this patch:
old
after:
rotlines

mmaciola:
Before patch: 
rotlines_pre

After patch:
rotlines_post

inc/thorvg.h
```diff
@@ -1196,8 +1196,10 @@ class TVG_EXPORT SwCanvas final : public Canvas
      */
     enum Colorspace
     {
-        ABGR8888 = 0, ///< The channels are joined in the order: alpha, blue, green, red.
-        ARGB8888      ///< The channels are joined in the order: alpha, red, green, blue.
+        ABGR8888 = 0, ///< The channels are joined in the order: alpha, blue, green, red. Colors are alpha-premultiplied.
+        ARGB8888,     ///< The channels are joined in the order: alpha, red, green, blue. Colors are alpha-premultiplied.
+        abgr8888,     ///< The channels are joined in the order: alpha, blue, green, red. Colors are un-alpha-premultiplied. Alpha unmultiplication is an extra step, may be slower than premultiplied version.
```
hermet, line 1201: I think "Alpha unmultiplication is an extra step, may be slower than premultiplied version." this is not the proper description for users. They don't need to mind of it or, we should not support this option. So please remove them.

hermet: I would like to suggest 2 options.
ABGR8888_STRAIGHT    @BETA_API
ARGB8888_STRAIGHT    @BETA_API

or  SwCanvas::premultiplied(on/off);      @BETA_API


What do you think?

hermet (changes requested):
Please check my comments.

hermet (approved):

hermet:
@mmaciola thanks.

PR #805, opened 2021-09-15
svg_loader: proper svg resizing

mgrudzinska:
Solves the problem of 'vx' and 'vy' < 0 and cases with 'width'
and 'height' values different than width and height from the 'viewBox'
attribute.

after:
26after_2

code (in 3 variations: viewBox + width/height, only viewBox, only widht/height):
```
<svg viewBox="-20 -20  80 80" width="200" height="100">
<path d="M -100 -100 h 200 v 200 h -100 z" fill="black" />
<path d="M -10 -10 h 10 v 10 h -10 z" fill="blue" />
<path fill="#FFB400" d="M 35 35 h 10 v 10 h -10 z"/>
</svg>
```

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -25,6 +25,7 @@
 #include "tvgSvgSceneBuilder.h"
 #include "tvgSvgPath.h"
 #include "tvgSvgUtil.h"
+#include <float.h>
 
 static bool _appendShape(SvgNode* node, Shape* shape, float vx, float vy, float vw, float vh);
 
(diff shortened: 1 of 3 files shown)


mgrudzinska:
We have to distinguish between width or height values equal to zero and cases when these values were not provided. (same with viewbox). it's a separate problem from this pr.

hermet:
@mgrudzinska I'm waiting for @JSUYA 's review. It would be better if you address the issue number in the commit message. if you are trying to fix any one.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -516,14 +517,45 @@ static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, float vx, float
...
+            auto tvw = vw * sx;
+            auto tvh = vh * sy;
+            if (tvw > tvh) tvy -= (h - tvh) * 0.5f;
+            else tvx -= (w - tvw) * 0.5f;
+            Matrix m = {sx, 0, -tvx, 0, sy, -tvy, 0, 0, 1};
+            docNode->transform(m);
+        }
+    } else if (vx < 0 || vy < 0) {
+        docNode->translate(-vx, -vy);
+    }
```
JSUYA, line 555: How about extract method of this resize code?

JSUYA (approved):
I did some tests on this PR. Checked the sample code and https://github.com/Samsung/thorvg/issues/799 to fix it.

PR #806, opened 2021-09-15
tvg_saver: do not merge if the shapes have outlines.

hermet:
Merged shapes with outlines expects the layered,
unfortunately tvg can't cover the situation.

@Issues: https://github.com/Samsung/thorvg/issues/783


src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -108,6 +108,10 @@ static bool _merge(Shape* from, Shape* to)
     if (r != r2 || g != g2 || b != b2 || a != a2) return false;
 
     if (fabs(from->strokeWidth() - to->strokeWidth()) > FLT_EPSILON) return false;
+
+    //OPTIMIZE: Yet we can't merge outlining shapes unless we can support merging shapes feature.
+    if (from->strokeWidth() > 0 || to->strokeWidth()) return false;
```
mgrudzinska, line 113: you've already checked that from->strokeWidth() and  to->strokeWidth() are the same, so now it's enough to check whether one of them is > 0

hermet: A = 0
B = 255

if (A > 0) return false;

mgrudzinska: in l.100:
```
if (fabs(A-B) > 0) return false;
```
so A == B in l. 113

hermet: it's trivial. each line doesn't need to depend on the other condition.

hermet: I hope you remember that I mentioned you before this kind of topic.

mgrudzinska: I'm not sure whether I got your point, but if you think that the code is more clear like that, than you can merge it. 
it solves the penrose stairs issue :)

PR #808, opened 2021-09-16
prepare to release tvg format feature.

hermet:


mgrudzinska:
@hermet and the "BETA_API" notation next to the saver class (thorvg.h)?

hermet:
> @hermet and the "BETA_API" notation next to the saver class (thorvg.h)?

yes. we will.

mmaciola (approved):

PR #812, opened 2021-09-17
sw_engine: proper gradient transformation

mgrudzinska:
In the case of any shape transformation, the gradient fill
was not correct.

before:
28before

after:
28after

code:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 220 120">
  <defs
     id="defs4">
    <linearGradient
       id="linearGradient1"
       x1="0"
       y1="0"
       x2="4"
       y2="10"
       gradientUnits="userSpaceOnUse" 
       spreadMethod="reflect">
      <stop
         style="stop-color:#ff0000;stop-opacity:1;"
         offset="0"/>
      <stop
         style="stop-color:#0000ff;stop-opacity:1;"
         offset="1"/>
    </linearGradient>
  </defs>

  <g transform="matrix(1.2, 1.4, 1.21, 0.7, -20, -74)">
    <rect x="0" y="0" width="20" height="50" fill="url(#linearGradient1)" transform="matrix(1.814, 3.59, 1.04, -2.45, 40, 50)"/>
  </g>
</svg>
```


mgrudzinska:
How it was done in EFL (also in nanoSVG):
1. function like prepareLinear: `dx`, `dy`, `offset` values were calculated (let's call it vector `V`)
2. function like fillFetchLinear: `x` and `y` values (let's call it vector `X`) were transformed using the inverse of the transformation matrix (let's denote it `A`) and the final values of `t` and `inc` were established
so we can write it as follows:
```
t = V *  (A * X)
```
`V` is transposed, but it's not important here. 
`V` was calculated in the prepareLinear funct, and `A * X` was calculated in fillFetchLinear. The final multiplication `V * AX` was also done here. It was slow, since fillFetchLinear is called often.

Now, in ThorVG I've changed it a bit:
1. prepareLinear: `dx`, `dy`, `offset` values are transformed using the `A` matrix (same as above) 
2. fillFetchLinear: the final values of `t` and `inc` are established here
it can be written as follows:
```
t = (V *  A) * X
```
So we do the transformation in the prepareLinear function (vector and matrix multiplication: `V * A`) and only the last step is done in fillFetchLinear: `VA * X`.

src/lib/sw_engine/tvgSwCommon.h
```diff
@@ -347,6 +347,7 @@ void mpoolRetOutline(SwMpool* mpool, unsigned idx);
 SwOutline* mpoolReqStrokeOutline(SwMpool* mpool, unsigned idx);
 void mpoolRetStrokeOutline(SwMpool* mpool, unsigned idx);
 
+bool inverse(const Matrix* transform, Matrix* invM);
```
hermet, line 350: bool mathInverse(const Matrix* transform, Matrix* invM);

see tvgSwMath.cpp

src/lib/sw_engine/tvgSwFill.cpp
```diff
@@ -123,7 +114,21 @@ bool _prepareLinear(SwFill* fill, const LinearGradient* linear, const Matrix* tr
 
     fill->linear.dx /= fill->linear.len;
     fill->linear.dy /= fill->linear.len;
-    fill->linear.offset = -fill->linear.dx * x1 -fill->linear.dy * y1;
+    fill->linear.offset = -fill->linear.dx * x1 - fill->linear.dy * y1;
+
+    if (transform) {
+        Matrix invTransform;
+        if (!inverse(transform, &invTransform)) return false;
```
hermet, line 121: if here it needs the inversed matrix, what about getting it at fillGenColorTable() ? 
propbably radialGradient might also need it?

hermet (changes requested):
Please check comments.

PR #817, opened 2021-09-20
common: preventing msvc compiler warnings

mgrudzinska:
msvc warings regarding conversion int->float are still present

### DIFF
--- a/src/lib/sw_engine/tvgSwStroke.cpp
+++ b/src/lib/sw_engine/tvgSwStroke.cpp
@@ -40,8 +40,8 @@ static inline SwFixed SIDE_TO_ROTATE(const int32_t s)
 
 static inline void SCALE(const SwStroke& stroke, SwPoint& pt)
 {
-    pt.x *= stroke.sx;
-    pt.y *= stroke.sy;
+    pt.x = static_cast<SwCoord>(pt.x * stroke.sx);
+    pt.y = static_cast<SwCoord>(pt.y * stroke.sy);
 }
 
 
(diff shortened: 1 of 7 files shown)


src/lib/sw_engine/tvgSwStroke.cpp
```diff
@@ -40,8 +40,8 @@ static inline SwFixed SIDE_TO_ROTATE(const int32_t s)
 
 static inline void SCALE(const SwStroke& stroke, SwPoint& pt)
 {
-    pt.x *= stroke.sx;
-    pt.y *= stroke.sy;
+    pt.x = static_cast<SwCoord>(pt.x * stroke.sx);
+    pt.y = static_cast<SwCoord>(pt.y * stroke.sy);
```
mgrudzinska, line 44: I didn't want to cast before multiplication

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -80,7 +80,7 @@ static unique_ptr<LinearGradient> _applyLinearGradientProperty(SvgStyleGradient*
             stops[i].r = colorStop->r;
             stops[i].g = colorStop->g;
             stops[i].b = colorStop->b;
-            stops[i].a = (colorStop->a * opacity) / 255.0f;
+            stops[i].a = static_cast<uint8_t>((colorStop->a * opacity) / 255.0f);
```
mmaciola, line 83: We dont need a float operation (int * int / int)

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -146,7 +146,7 @@ static unique_ptr<RadialGradient> _applyRadialGradientProperty(SvgStyleGradient*
...
```
mmaciola, line 149: Same as above

mmaciola (changes requested):
Please see comments. Other changes are great

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: I've changed 255.0f -> 255, but I've kept casting, since opacity is int

mmaciola (approved):

PR #818, opened 2021-09-20
common: change args type in the internal size() implementation

mgrudzinska:
The API expected floats, whereas the called impl function expected
ints. The values were saved as floats, so the float->int->float conversion
is not needed.


mmaciola (approved):
Looks ok

hermet (approved):

mahammedmahieousman:
mgrudzinska:mgrudzinska/size

mahammedmahieousman:
_ousman_

Issue #813, opened 2021-09-18
Meson build not using clang (Windows x64)


projectitis:
Hi all,

I am trying to compile ThorVG for clang on windows using the following command(s):
`CC=clang CCX=clang++ meson build`

I get this error compiling sanitycheck (full log below):
 `LINK : fatal error LNK1104: cannot open file 'MSVCRT.lib'`

1) It appears that meson is not actually using clang - any idea how to force it to on windows?
2) If I can't, any idea how I can make the linker aware of MSVCRT? It's on my PATH, but it needs to be included in the linker library path.

Full error log:
````
Build started at 2021-09-18T22:55:16.926103
Main binary: C:\Program Files\Meson\meson.exe
Build Options: 
Python system: Windows
The Meson build system
Version: 0.59.1
Source dir: D:\Projects\thorvg
Build dir: D:\Projects\thorvg\build
Build type: native build
Project name: thorvg
Project version: 0.4.99
Sanity testing C++ compiler: cl
Is cross compiler: False.
Sanity check compiler command line: cl sanitycheckcpp.cc /Fesanitycheckcpp.exe /MD /nologo /showIncludes /link
Sanity check compile stdout:
sanitycheckcpp.cc
LINK : fatal error LNK1104: cannot open file 'MSVCRT.lib'

-----
Sanity check compile stderr:

-----

meson.build:1:0: ERROR: Compiler cl can not compile programs.
````

Thanks!

**Comment by mgrudzinska** [comment]
Hi, 

I've just tried to build ThorVG on Win using clang. 
First I've installed LLVM:
https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.1
and (using Visual Studio Installer) the C++ Clang tools:

Unfortunately I didn't reproduce your issue, that seems to be connected strictly with VS (maybe this will help? https://stackoverflow.com/questions/927574/compiling-c-program-causes-fatal-error-lnk1104)

Nevertheless, there was also a problem with pthread lib.  I solved it in #815 but let's wait for merge.
There are still some warning, but I was able to build the lib.

**Comment by projectitis** [comment]
Thank you for such a quick response!  I have the same toolset (Clang installed via VS).
Could I ask which commands you are using to compile thorvg with clang successfully?

**Comment by mgrudzinska** [comment]
Hi
I had to install some other compilers to see how to switch between them and surprisingly it doesn't seem to be a problem (in my case at least :) )
In a command prompt I've typed:
```
set CC=clang
[...]

PR #819, opened 2021-09-20
sw_engine avx: removing not portable variable type

mgrudzinska:
__m256i_u is an internal gcc/clang type, which can not be used by msvc.
To ensure the portability of the lib a proper intrinsic is used.

The question, whether the aligned and unaligned memory should be handled separately (like in #753) is still open. 

present in #813


mmaciola (approved):
Looks ok

hermet (approved):

PR #821, opened 2021-09-21
svg loader: converting path to the absolute one

mgrudzinska:
The image element of the svg file should point to an absolute image path.

image-embeded.svg

before:
30before

after:
30after


mmaciola:
I'm not sure about that. I have to check if there is a real problem with paths, but even if, I think it should be using a path ot the current working directory, not a path of meson compilation.

mgrudzinska:
please notice, that if a user reads the svg file using ifstream and then loads the data using load(data, size, mimeType), than the info about path is lost. I think we can live with that, because the other option is to add an additional arg (to the load api) with path, but then using this method of file loading doesn;t make any sense.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -442,7 +442,7 @@ static bool _isValidImageMimeTypeAndEncoding(const char** href, const char** mim
 }
 
 
-static unique_ptr<Picture> _imageBuildHelper(SvgNode* node, float vx, float vy, float vw, float vh)
+static unique_ptr<Picture> _imageBuildHelper(SvgNode* node, float vx, float vy, float vw, float vh, string svgPath)
```
hermet, line 445: we can avoid memory copy in the parameter,
probably =>
const string& svgPath;
or
const char* svgPath;

hermet (changes requested):
Please check a comment.

src/examples/images/image-embeded.svg
```diff
@@ -7,6 +7,6 @@
...
 	<image href="data:image/svg+xml;base64,PHN2ZyB4bWxucz0naHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmcnIHZpZXdCb3g9JzAgMCAxMDAgMTAwJz48Y2lyY2xlIGN4PSc1MCcgY3k9JzUwJyByPSc0OCcgZmlsbD0nbm9uZScgc3Ryb2tlPScjMDAwJy8+PHBhdGggZD0nTTUwLDJhNDgsNDggMCAxIDEgMCw5NmEyNCAyNCAwIDEgMSAwLTQ4YTI0IDI0IDAgMSAwIDAtNDgnLz48Y2lyY2xlIGN4PSc1MCcgY3k9JzI2JyByPSc2Jy8+PGNpcmNsZSBjeD0nNTAnIGN5PSc3NCcgcj0nNicgZmlsbD0nI0ZGRicvPjwvc3ZnPg0K" height="100" width="100" x="0" y="100" />
-	<image href="src/examples/images/yinyang.svg" height="100" width="100" x="100" y="100" />
-	<image href="src/examples/images/logo.png" height="80" width="80" x="60" y="60" clip-path="url(#circleView)" />
-</svg>
\ No newline at end of file
+	<image href="yinyang.svg" height="100" width="100" x="100" y="100" />
```
mgrudzinska, line 10: either an absolute path or a relative one, this is neither of them

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: of course, changed

JSUYA (approved):
LGTM

PR #826, opened 2021-09-22
svg_loader: fix potential mem corruption in _idFromUrl function

### LINKED ISSUE #822: Long url name can corrupt memory
`_idFromUrl` function should be fixed: long text can corrupt memory

mmaciola:
This patch fixes some potential memory corruptions in _idFromUrl function
when name (url) is longer than 50 chars or is incorrectly terminated.

issue: Closes #822


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -272,24 +272,19 @@ _parseDashArray(SvgLoaderData* loader, const char *str, SvgDash* dash)
...
+    if ((*url) == '\'' || (*url) == '\"') ++url;
     if ((*url) == '#') ++url;
 
-    while ((*url) != ')') {
-        tmp[i++] = *url;
-        ++url;
-    }
-    tmp[i] = '\0';
+    int i = 0;
+    while (url[i] > ' ' && url[i] != ')' && url[i] != '\'' && url[i] != '\"') ++i;
```
mgrudzinska, line 285: I don't think it's possible to find a `"` sign in the url arg, since we are using `"` to determine the beginning and the end of a tag/value etc.
and `url[i] > ' '` - url can not contain spaces, what is the scenario for that?

mmaciola: That's a fine catch with a quote sign. Will change it.
url must not contain any whitespace and `url[i] > ' '` should stop searching on '\0', space, tab, new line, etc

mgrudzinska: if it contains a white space etc, doesn't in mean that it's invalid ?

PR #830, opened 2021-09-23
sw_engine: code refactoring.

hermet:
rasterization functions are getting bigger.
split implementation vector-based and image-based for easier maintenance.


hermet:
This increases the symbol size 14k bytes.... :(

PR #831, opened 2021-09-23
apis: release official apis.

hermet:
tvg picture is going to be released, we need the Saver as well.
Also, two stable periperal apis are released in v0.5

@APIs:
Result Picture::load(const char* data, uint32_t size, const std::string& mimeType, bool copy = false) noexcept;
CompositeMethod Paint::composite(const Paint** target) const noexcept;
Result Saver::save(std::unique_ptr<Paint> paint, const std::string& path, bool compress = true) noexcept;
Result Saver::sync() noexcept;
static std::unique_ptr<Saver> Saver::gen() noexcept;

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -315,7 +315,7 @@ public:
      *
      * @return The method used to composite the source object with the target.
      *
-     * @BETA_API
+     * @since 0.5
      */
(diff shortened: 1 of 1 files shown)


hermet:
@mihashco @mmaciola @JSUYA @mgrudzinska  yes/no?

mgrudzinska (approved):

PR #834, opened 2021-09-23
test: added test for tvg_picture_get_viewbox function

bchrescionko:
Added lines for testing tvg_picture_get_viewbox function in test/capi/capiPicture.cpp

### DIFF
--- a/test/capi/capiPicture.cpp
+++ b/test/capi/capiPicture.cpp
@@ -47,7 +47,7 @@ TEST_CASE("Load Svg file in Picture", "[capiPicture]")
 
 TEST_CASE("Load Svg Data in Picture", "[capiPicture]")
 {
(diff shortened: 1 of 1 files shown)


mgrudzinska:
@hermet, since the viewbox api is in beta, I assume the corresponding capi also will be marked as such. so do we want to include it into tests?

hermet:
@mgrudzinska there is no point to keep the deprecated apis test, but for the beta apis, it's not recommended neither. But I think we can apply it because we have no concrete plans to promoting/dropping it.

test/capi/capiPicture.cpp
```diff
@@ -66,6 +66,12 @@ TEST_CASE("Load Svg Data in Picture", "[capiPicture]")
     REQUIRE(w == Approx(1000).epsilon(0.0000001));
     REQUIRE(h == Approx(1000).epsilon(0.0000001));
 
+    //Verify Position
+    float x, y;
+    REQUIRE(tvg_picture_get_viewbox(picture, &x, &y, &w, &h) == TVG_RESULT_SUCCESS);
+    REQUIRE(x == Approx(0).epsilon(0.0000001));
+    REQUIRE(y == Approx(0).epsilon(0.0000001));
```
hermet, line 73: btw, why do you only check x/y, not w/h?

hermet (commented):
please check a comment.

test/capi/capiPicture.cpp
bchrescionko: Becouse it was check with with tvg_picture_get_size in line 65, so I didn't if I had to check it again

hermet: Unit test is to verify the apis work correctly or not.
In point of API perspective view, tvg_picture_get_size() and tvg_picture_get_viewbox() have no relationship with each others.
That means it needs test possible scenarios for tvg_picture_get_viewbox() return values.

Thus, please check all return values.

hermet (changes requested):
Please check my comments.

test/capi/capiPicture.cpp
bchrescionko: Ok, I wll check it

hermet (approved):

Issue #29, opened 2020-09-10
cpu: AVX optimization


hermet:
Still there are a lot of rasterzing parts, not have SIMD operation.
We need a better fine-tuned software raster engine.

We can consider to apply AVX to some rasterizing methods in tvgSwRaster.cpp

such as:

_translucentRect()
_translucentRectAlphaMask()
_translucentRectInvAlphaMask()
_rasterTranslucentRect()
...
_rasterRadialGradientRle()

Some might not fit to SIMD (this case we can skip to apply) but Some would be nicer than basic.

* Submitting individual patches for each methods will be better.

hermet:
https://www.codeproject.com/Articles/874396/Crunching-Numbers-with-AVX-and-AVX
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#techs=AVX,AVX2&expand=4950
https://github.com/usamadar/SIMD/blob/master/AVXTest/avxtest.cpp
https://m.blog.naver.com/PostView.nhn?blogId=programsite&logNo=150117854865&proxyReferer=https:%2F%2Fwww.google.com%2F

mgrudzinska:
we encountered a problem related to compiling the ThorVG lib with the `-mavx` flag and running it on the host, that does not support avx. 

it can be reproduced on a pc using any program with a linked lib. The lib has to be compiled with `-mavx` (the lib itself doesn't need to call any avx intrinsics). I built the shared lib as follows:
`g++ -mavx -c -o lib.o lib.cpp 
gcc -shared -o liblib.so lib.o`
the program is compiled without `-mavx` flag:
`g++ main.cpp -o main -Ldirectory_to_the_lib -llib`
(LD_LIBRARY_PATH has to be updated)
Since we need to test this on an architecture,  that doesn't support AVX, the Intel Software Development Emulator can be used:
https://software.intel.com/content/www/us/en/develop/articles/intel-software-development-emulator.html
after the installation we can run the main (we pretend to run it on Nehalem microarch): 
`sde -nhm -- ./main`
and it will crash.
the problem is, that the assembly instructions with the vex prefixes will be used, which are not valid on the Nehalem microarchitecture.

**So, the conclusion is:**
we cannot compile ThorVG with `-mavx` flag. Only the avx module can be compiled like that. During the runtime we have to decide, whether the host supports avx or not, so whether the avx module can or cannot be used.

mgrudzinska:
just as an info for now:
[...]

PR #835, opened 2021-09-24
infra: determining simd extension during runtime

mgrudzinska:
THORVG_SIMD_SUPPORT will be used in the near future to choose the proper raster functions.
issue #29


hermet (changes requested):
I think simd mantenance logic must be in sw_engine side. 
because it totally depends it. What do you think?

mgrudzinska:
@hermet could you please be a bit more specific? all of the changes are already made in the sw_engine

vtorri:
`HOST_SYSTEM_WINDOWS` is useless. use `_WIN32`, it is defined for all Windows compiler (msvc, gcc, clang, clang-cl, intel-cl, borland, etc...). It's simpler and reduce meson code

vtorri:
note that iirc, mpeg123 code has a runtime detection of cpu architectures

vtorri:
according to mpg213 author : 
```
This is in the src/libmpg123/getcpuflags* files. Some assembly to get
registers, then flags to check in those.
```

PR #839, opened 2021-09-28
infra: Add window build test in CI

JSUYA:
Add build test for normal build

https://github.com/Samsung/thorvg/issues/600

### DIFF
--- a/src/lib/tvgInitializer.cpp
+++ b/src/lib/tvgInitializer.cpp
@@ -23,6 +23,10 @@
 #include "tvgTaskScheduler.h"
 #include "tvgLoader.h"
 
+#ifdef _WIN32
+    #include <cstring>
+#endif
+
 #ifdef THORVG_SW_RASTER_SUPPORT
     #include "tvgSwRenderer.h"
 #endif
--- a/.github/workflows/build_win.yml
+++ b/.github/workflows/build_win.yml
@@ -1,0 +1,30 @@
+name: Build(Window)
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ilammy/msvc-dev-cmd@v1
+      with:
+        submodules: true
+    - name: Install Packages
+      run: |
+        pip install meson==0.58.0 ninja
+    - name: Build
+      run: |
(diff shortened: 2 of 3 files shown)


src/lib/tvgCommon.h
```diff
@@ -25,6 +25,10 @@
 #include "config.h"
 #include "thorvg.h"
 
+#ifdef _BUILD_WIN_
+#include <cstring>
```
hermet, line 29: I think we can isolate this include to the necessary files.
Could you please suggest the error position?

hermet (changes requested):

src/lib/tvgCommon.h
mgrudzinska: @hermet @JSUYA before I noticed this pr I prepared mine: https://github.com/Samsung/thorvg/pull/840

I dodn;t want to include string.h in tvgInitializer.cpp so I changed the code a bit.
I would appreciate your comment on such an approach.

JSUYA: I deleted the code change for the build fix.
I hope that build fail will be fixed in a separate patch(add cstring) or #840.

mgrudzinska (changes requested):
cstring should be also included in the svg loader (tvgSvgLoaderCommon and tvgSvgUtil)

mgrudzinska:
I've just got an error on windows using c++ (msys) compiler - it seams that string.h should be also included in tvgTvgSaver.cpp (strdup requires it)

src/lib/tvgInitializer.cpp
```diff
@@ -23,6 +23,10 @@
 #include "tvgTaskScheduler.h"
 #include "tvgLoader.h"
 
+#ifdef _MSC_VER
```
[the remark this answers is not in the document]
JSUYA: I changed to _WIN32

JSUYA:
> I've just got an error on windows using c++ (msys) compiler - it seams that string.h should be also included in tvgTvgSaver.cpp (strdup requires it)

This PR is for normal build test in MSVC(`ilammy/msvc-dev-cmd@v1`) environment.
tvg Saver is not yet a default option.
I will consider it when adding the build test environment later.

mgrudzinska:
@JSUYA the svg loader is the def opt and it requires cstring, but I understand it will be in a sep pr.

mgrudzinska (approved):

PR #840, opened 2021-09-28
common: fixing warnings on windows

mgrudzinska:
cstring was not included

### DIFF
--- a/src/lib/tvgInitializer.cpp
+++ b/src/lib/tvgInitializer.cpp
@@ -43,25 +43,31 @@ static uint16_t _version = 0;
 static bool _buildVersionInfo()
 {
     auto SRC = THORVG_VERSION_STRING;   //ex) 0.3.99
-    auto p = SRC;
     const char* x;
+    auto p = SRC;
+
+    x = p;
+    while (*x != '\0') ++x;
+    auto len = x - SRC;
 
     char major[3];
-    x = strchr(p, '.');
-    if (!x) return false;
+    x = p;
+    while (*x != '.' && (x - SRC < len - 1)) x++;
+    if (x - SRC == len - 1) return false;
     memcpy(major, p, x - p);
     major[x - p] = '\0';
     p = x + 1;
 
     char minor[3];
-    x = strchr(p, '.');
-    if (!x) return false;
+    x = p;
+    while (*x != '.' && (x - SRC < len - 1)) x++;
+    if (x - SRC == len - 1) return false;
     memcpy(minor, p, x - p);
     minor[x - p] = '\0';
     p = x + 1;
 
     char micro[3];
-    x = SRC + strlen(THORVG_VERSION_STRING);
+    x = SRC + len;
     memcpy(micro, p, x - p);
     micro[x - p] = '\0';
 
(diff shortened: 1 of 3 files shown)


src/lib/tvgInitializer.cpp
```diff
@@ -43,25 +43,31 @@ static uint16_t _version = 0;
...
-    auto p = SRC;
     const char* x;
+    auto p = SRC;
+
+    x = p;
+    while (*x != '\0') ++x;
+    auto len = x - SRC;
 
     char major[3];
-    x = strchr(p, '.');
```
mgrudzinska, line 50: I could include string.h instead of these changes.. should I?

mgrudzinska:
#839

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -24,6 +24,7 @@
 
 #include "tvgCommon.h"
 #include "tvgArray.h"
+#include <string.h>
```
JSUYA, line 27: According to your pr, we don't use str functions anymore, do we need this header?

src/loaders/svg/tvgSvgUtil.cpp
```diff
@@ -263,7 +263,9 @@ string svgUtilURLDecode(const char *src)
 {
     if (!src) return nullptr;
 
-    auto length = strlen(src);
+    const char* x = src;
```
JSUYA, line 266: How about `auto` instead of const char*?

src/loaders/svg/tvgSvgUtil.cpp
```diff
@@ -290,7 +292,10 @@ string svgUtilBase64Decode(const char *src)
...
```
JSUYA, line 295: How about `auto` instead of const char*?

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: we are using functions from cstring in two places. 
One is the 'src/lib/tvgInitializer.cpp' and I changed a bit the code, so we don;t need to include 'cstring'
The other one is svg loader - in all files. also all files but 'tvgSvgUtil' have 'tvgSvgLoaderCommon' included. So I've included 'cstring' in the common file, and I've changed 'tvgSvgUtil' so it doesn't need 'cstring' any more

mgrudzinska: @JSUYA but I included `string.h` no matter on what platform we compile the lib. I can't find any info why it's necessary only on windows...

src/loaders/svg/tvgSvgUtil.cpp
mgrudzinska: sorry, I forgot about that... changed

src/loaders/svg/tvgSvgUtil.cpp
mgrudzinska: sorry, I forgot about that... changed

hermet (changes requested):
Why not including string...? I think there's no much gain, but code complexity is increased.
Any specific reason?

JSUYA:
> Why not including string...? I think there's no much gain, but code complexity is increased. Any specific reason?

I was concerned that problems related to str function would continue to occur depending on the build environment.
So I thought not using it might be a better solution.
I did some tests and modified the github action script and build environment appropriately. (https://github.com/Samsung/thorvg/pull/839)

@mgrudzinska I am sorry to bother you.

mgrudzinska:
one more time - #839 :)

Issue #844, opened 2021-09-28
Variable Length Arrays (VLA) are not portable


projectitis:
Although many compilers support variable length arrays (VLA) - sometimes as extensions - they are not part of the C++ standard. MSVC fails to compile them. The alternative is to create the array using new/delete. Here is a [little info](https://stackoverflow.com/questions/1887097/why-arent-variable-length-arrays-part-of-the-c-standard) about them.

I'm not sure how to generate a patch, but below is the commit replacing 2 VLAS in `src/loaders/tvg/tvgTvgBinInterpreter.cpp` on my fork: https://github.com/DeriveSDK/thorvg/commit/6562e77e25ac3e9126b4f8a5663b9ae951b11858

I have also noticed VLAs used in a couple of places in the examples. If you verify that this patch is ok (or suggest an alternative if not) I will create PRs for any others I come across. 

An alternative, of course, is not to support MSVC, which seems to be a bit backwards compared to many other c++ compilers!

Thanks for reminding me, @mgrudzinska :) This was raised and split from #824

mgrudzinska:
you have to fork ThorVG, create a new branch (from the current/updated master branch), make changes that you think are necessary, commit them and push. Then on github it will be easy, since github will ask u whether you want to create a pull request. 
Please let us know if u need some help with that. 
Compiling on linux with the 'pedentic' flag gives a waring regarding this topic, and since u find it, it'd be cool if you'd be the author of pr, that fixes it :))

mgrudzinska:
![s1](https://user-images.githubusercontent.com/67589014/135172729-82fca67d-cc90-41e3-be66-8421b82a2992.PNG)
![s2](https://user-images.githubusercontent.com/67589014/135172736-f30ff8ce-1077-430f-a81f-a61a086c3670.PNG)

mgrudzinska:
u can do it from your current repo :)

projectitis:
Yes, that's I would normally do,. but you asked for a patch and I misunderstood - sorry!

projectitis:
Code design question: Is ThorVG generally trying to avoid compiler-conditional code? E.g.
````
if MSVC
  Do it this way
else
  Do it this way
````
I'd like to know if you would prefer a single (non-conditional) solution that is portable, or a conditional one?
The reason is that I can use _malloca (to allocate stack memory) on MSVC, but it's not portable either :)

mgrudzinska:
[...]

PR #845, opened 2021-09-29
portability: Changes to support MSVC

projectitis:
Recommend the following changes to support building on MSVC:
- Define capi export visibility for WIN32. Equivalent to `__attribute__ ((visibility ("default")))` is `__declspec(dllexport)`
- Replace variable length arrays (not supported by MSVC) with stack allocation via `alloca` (note: same code used as already found in `tvgXmlParser`)
- Changed designated initlializers in arrays of structs (`test/capi/capi[Linear/Radial]Gradient` to avoid `error C7555: use of designated initializers requires at least '/std:c++latest'` since default for thorvg is c++14

Happy for any feedback and further changes you recommend!
Relates to #844

### DIFF
--- a/src/bindings/capi/thorvg_capi.h
+++ b/src/bindings/capi/thorvg_capi.h
@@ -27,11 +27,14 @@
 #endif
 
 #ifdef TVG_BUILD
-    #define TVG_EXPORT __attribute__ ((visibility ("default")))
+    #ifdef _WIN32
+        #define TVG_EXPORT __declspec(dllexport)
+    #else
+        #define TVG_EXPORT __attribute__ ((visibility ("default")))
+    #endif
 #else
     #define TVG_EXPORT
 #endif
-
 
 #ifdef __cplusplus
 extern "C" {
--- a/src/loaders/tvg/tvgTvgBinInterpreter.cpp
+++ b/src/loaders/tvg/tvgTvgBinInterpreter.cpp
@@ -20,6 +20,13 @@
  * SOFTWARE.
  */
 #include <memory.h>
+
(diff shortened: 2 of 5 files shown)


src/bindings/capi/thorvg_capi.h
```diff
@@ -27,12 +27,15 @@
 #endif
 
 #ifdef TVG_BUILD
-    #define TVG_EXPORT __attribute__ ((visibility ("default")))
+    #ifdef _WIN32
```
JSUYA, line 30: In `inc/thorvg.h` we are using `_MSC_VER` instead of `_WIN32`.
How about using `_MSC_VER`?

JSUYA (approved):
+1 :)

src/bindings/capi/thorvg_capi.h
projectitis: That would be fine :) I copied the `_WIN32` example from an existing file `tvgXmlParser` but I am happy to use whatever is suggested.

hermet (approved):
thanks!

src/bindings/capi/thorvg_capi.h
projectitis: @hermet Do you think `#ifdef _WIN32` should be changed to `#ifdef _MSC_VER`?
On windows:
- gcc supports both  `__attribute__ ((visibility ("default")))` and `__declspec(dllexport)`
- clang supports both `__attribute__ ((visibility ("default")))` and `__declspec(dllexport)`
- MSVC only supports `__declspec(dllexport)`
So they both work fine. What is your preference?

PR #846, opened 2021-09-29
Updated readme with optional dependencies

projectitis:
This is just a suggestion. Feel free to edit or ignore as you see fit!
There may be others I have missed. I also didn't know where to link EGL and GLESv2.

As a user, it is helpful to know these possible dependencies rather than finding out after I change an option that I need additional libraries :)


JSUYA (approved):
+1 :)

hermet (approved):

PR #848, opened 2021-09-29
capi: Release official CAPI's.

mihashco:
Changes:
- Removed beta api tag from CAPI module
- For some APIs, the tag has not been changed due to main API status
- Moved some documentation changes to have similar descriptions in main
API and CAPI
- Removed deprecation warnings.

### DIFF
--- a/src/bindings/capi/thorvg_capi.h
+++ b/src/bindings/capi/thorvg_capi.h
@@ -13,7 +13,6 @@
 * - composition: blending, masking, path clipping
 * - pictures: SVG, PNG, bitmap
 *
-* @BETA_API
 */
 
 #ifndef __THORVG_CAPI_H__
@@ -38,7 +37,7 @@ extern "C" {
 #endif
 
 /**
-* \defgroup ThorVG_CAPI ThorVG_CAPI (BETA version)
+* \defgroup ThorVG_CAPI ThorVG_CAPI
 * \brief ThorVG C language binding APIs.
 *
 * \{
@@ -216,6 +215,10 @@ typedef struct
 
 /**
  * \brief A data structure representing a three-dimensional matrix.
+ *
+ * The elements e11, e12, e21 and e22 represent the rotation matrix, including the scaling factor.
+ * The elements e13 and e23 determine the translation of the object along the x and y-axis, respectively.
+ * The elements e31 and e32 are set to 0, e33 is set to 1.
  */
 typedef struct
 {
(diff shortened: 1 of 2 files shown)


src/bindings/capi/tvgCapi.cpp
```diff
@@ -194,7 +194,7 @@ TVG_EXPORT Tvg_Result tvg_paint_get_opacity(Tvg_Paint* paint, uint8_t* opacity)
 TVG_EXPORT Tvg_Result tvg_paint_get_bounds(const Tvg_Paint* paint, float* x, float* y, float* w, float* h)
```
hermet, line 194: I think we need to replace this to  tvg_paint_get_bounds(const Tvg_Paint* paint, float* x, float* y, float* w, float* h, bool transformed); as same to c++ version.
Could you please make the patch? (separate patch is welcomed)

hermet (commented):
@mihashco please have one comment.

hermet:
waiting for @mgrudzinska  @mmaciola 's verification.

src/bindings/capi/tvgCapi.cpp
mihashco: I thought about it, the problem is that another library uses this API so such a change would break backwards compatibility. You would have to bump up the version, so I think a separate commit would be a better solution.

src/bindings/capi/thorvg_capi.h
```diff
@@ -586,7 +597,7 @@ TVG_EXPORT Tvg_Result tvg_canvas_update_paint(Tvg_Canvas* canvas, Tvg_Paint* pai
 
 
 /*!
-* \brief The function start rendering process.
+* \brief Request the canvas to draw the Paint objects.
```
mgrudzinska, line 600: Paint -> Tvg_Paint

src/bindings/capi/thorvg_capi.h
```diff
@@ -1151,7 +1173,7 @@ TVG_EXPORT Tvg_Result tvg_shape_set_stroke_linear_gradient(Tvg_Paint* paint, Tvg
 
 /*!
 * \brief Sets the radial gradient fill of the stroke for all of the figures from the path.
-*
+*f
```
mgrudzinska, line 1176: ?

src/bindings/capi/thorvg_capi.h
```diff
@@ -1671,7 +1693,7 @@ TVG_EXPORT Tvg_Result tvg_picture_load_raw(Tvg_Paint* paint, uint32_t *data, uin
 
 
 /*!
-* \brief The function loads data into given paint object. (BETA version)
+* \brief The function loads data into given paint object.
```
mgrudzinska, line 1696: `The function loads data into the given paint object.`

mgrudzinska (changes requested):
I have also some comment to the parts of the file that weren't changed in this pr, but I think I should prepare a separate one

mmaciola (approved):

src/bindings/capi/tvgCapi.cpp
mihashco: Agree

src/bindings/capi/thorvg_capi.h
```diff
@@ -317,6 +324,9 @@ TVG_EXPORT Tvg_Result tvg_engine_term(unsigned engine_method);
 /*!
 * \brief Creates a Canvas object.
 *
+* \note A Canvas behavior depends on the raster engine though the final content of the buffer is expected to be identical.
+* \warning The Paint objects belonging to one Canvas can't be shared among multiple Canvases.
```
mgrudzinska, line 328: I'd move these 2 sentences to the Canvas group (l.307).

mgrudzinska (approved):
only one comment, but not so important probably

PR #849, opened 2021-09-29
sw_engine: building tvgSwRasterAvx as a library

mgrudzinska:
it's a draft, since in `src/lib/sw_engine/tvgSwRaster.cpp` the changes are made only for testing purposes. 
I want to use function pointers, but not to each raster function, rather only to the external ones. but to do this we need to implement more avx functions.

issue #29

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -40,6 +40,10 @@ if get_option('savers').contains('tvg') == true
     config_h.set10('THORVG_TVG_SAVER_SUPPORT', true)
 endif
 
+if host_machine.system() == 'windows'
+    config_h.set10('HOST_SYSTEM_WINDOWS', true)
+endif
+
 simd_type = 'none'
 
 if get_option('vector') == true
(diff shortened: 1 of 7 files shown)


mgrudzinska:
@hermet I'd appreciate your comment on the solution - is it ok to use tvgSwRasterAvx as a library? I also use ifdefs to include it instead of inside tvgSwRasterAvx.h/cpp files - is that correct?

hermet:
> @hermet I'd appreciate your comment on the solution - is it ok to use tvgSwRasterAvx as a library? I also use ifdefs to include it instead of inside tvgSwRasterAvx.h/cpp files - is that correct?

if you mean that dynamic linking, it's not good. Back to the origin does it necessary? Can't we decide it to call or not by the toggling variable?

mgrudzinska:
I'll come back to this pr after more functions will be implemented in the avx version

Issue #850, opened 2021-09-30
SVG loader: svg image is incorrectly clipped


mmaciola:
[SVG image from this link](https://upload.wikimedia.org/wikipedia/commons/b/b0/NewTux.svg) is incorrectly clipped
Viewer link: https://samsung.github.io/thorvg.viewer/?s=https%3A%2F%2Fupload.wikimedia.org%2Fwikipedia%2Fcommons%2Fb%2Fb0%2FNewTux.svg

Snippet:
```
    auto picture = tvg::Picture::gen();
    if (picture->load("/.../NewTux.svg") != tvg::Result::Success) {
         cout << "PNG is not supported. Did you enable PNG Loader?" << endl;
         return;
    }
    picture->size(WIDTH, HEIGHT);
    if (canvas->push(move(picture)) != tvg::Result::Success) return;
```
![Screenshot from 2021-09-30 10-05-24](https://user-images.githubusercontent.com/71131832/135413188-b36afbbc-2a65-4d39-befb-bb26a42a9c86.png)

PR #851, opened 2021-09-30
svg_loader: proper handling width/height units

mgrudzinska:
Units of the svg width and height tags were incorrectly assigned.
issue #850

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -697,6 +697,8 @@ error:
     }
 
 
+/*
+// TODO - remove?
 static constexpr struct
 {
     const char* tag;
@@ -712,7 +714,6 @@ static constexpr struct
     LENGTH_DEF(in, SvgLengthType::In)
 };
 
-
 static float _parseLength(const char* str, SvgLengthType* type)
 {
     float value;
@@ -725,7 +726,7 @@ static float _parseLength(const char* str, SvgLengthType* type)
     value = svgUtilStrtof(str, nullptr);
     return value;
 }
-
+*/
 
 static bool _parseStyleAttr(void* data, const char* key, const char* value);
 static bool _parseStyleAttr(void* data, const char* key, const char* value, bool style);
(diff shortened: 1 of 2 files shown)


JSUYA (commented):
Please see the `_toFloat()` function.
maybe you can use it like this.
doc->w = _toFloat(loader->svgParse, value, SvgParserLengthType::Horizontal);
doc->h = _toFloat(loader->svgParse, value, SvgParserLengthType::Vertical);

mmaciola:
Maybe use `end` from svgUtilStrtof:

```
static float _parseLength(const char* str, SvgLengthType* type)
{
    char* end = nullptr;
    float value = svgUtilStrtof(str, &end);

    *type = SvgLengthType::Px;
    if (*end != '\0') {
        for (unsigned int i = 0; i < sizeof(lengthTags) / sizeof(lengthTags[0]); i++) {
            if (!strcmp(lengthTags[i].tag, end)) *type = lengthTags[i].type;
        }
        if (*type != SvgLengthType::Px) _handleType(&value, type);
    }
    return value;
}
```

mgrudzinska:
@mmaciola @JSUYA after this change we don;t need neither SvgLengthType nor _parseLength. cout you please confirm?

I didn;t remove them, since I'm not sure how the percentage values of the width and height should work. for now global is zero, is nothing will be ploted in such a case.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -762,7 +762,7 @@ static bool _attrParseSvgNode(void* data, const char* key, const char* value)
     }
 #ifdef THORVG_LOG_ENABLED
     else if (!strcmp(key, "x") || !strcmp(key, "y")) {
-        if (0.0f == _parseLength(value, &type)) TVGLOG("SVG", "Unsupported attributes used [Elements type: Svg][Attribute: %s][Value: %s]", key, value);
```
JSUYA, line 765: Please keep this code like this.
`if (0.0f != _parseLength(value, &type)) TVGLOG("SVG", "Unsupported attributes used [Elements type: Svg][Attribute: %s][Value: %s]", key, value);`

`==` is my fault.
The reason this condition is needed is because of the VOC(?) that, as I remember.
It generates unnecessary warnings when x and y are zero, even though the image doesn't actually change.
`x = 0`, `y = 0`, which are generated by default in certain tools, appear to have a problem even though there is no effect.

JSUYA (approved):
I agree to this change.
Please see my comments before merging

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: @JSUYA could you please take a look one more time and decide, whether are we ok with the commented parts? I didn't want to remove them (yet).

JSUYA (approved):
LGTM :)

mmaciola (approved):

PR #854, opened 2021-09-30
docs: TVG logo svg file added to the docs directory

mgrudzinska:


hermet (changes requested):
please check a comment.

docs/small_logo.svg
```diff
@@ -0,0 +1 @@
...preserveAspectRatio="none" height="62" width="62" viewBox="0 0 1000 1000" xmlns="http://www.w3.org/2000/svg"><path d="M-.01986229-.01483051 999.12603.51483051 999.95363 999.91554.07944915 999.94865Z" fill="#fff" stroke-width="3.910218"/><g fill="#252f35"><g stroke-width="3.864492"><path d="M256.61221 100.51736H752.8963V386.99554H256.61221Z"/><path d="M201.875 100.51736H238.366478V386.99554H201.875Z"/><path d="M771.14203 100.51736H807.633508V386.99554H771.14203Z"/></g><path d="M420.82388 380H588.68467V422.805317H420.82388Z" stroke-width="3.227"/><path d="m420.82403 440.7101v63.94623l167.86079 25.5782V440.7101Z"/><path d="M420.82403 523.07258V673.47362L588.68482 612.59701V548.13942Z"/></g><g fill="#222f35"><path d="M420.82403 691.37851 588.68482 630.5019 589 834H421Z"/><path d="m420.82403 852.52249h167.86079v28.64782H420.82403v-28.64782 0 0"/><path d="m439.06977 879.17031c0 0-14.90282 8.49429-18.24574 15.8161-4.3792 9.59153 0 31.63185 0 31.63185h167.86079c0 0 4.3792-22.04032 0-31.63185-3.34292-7.32181-18.24574-15.8161-18.24574-15.8161z"/></g><g fill="#fff"><path d="m280 140h15v55l8 10 8-10v-55h15v60l-23 25-23-25z"/><path d="m335 140v80h45v-50h-25v10h10v30h-15v-57h18v-13z"/></g></svg>
```
hermet, line 1: there was a reason to put the small_logo.svg in res folder. actually, generated html doesn't work with the relative path svg, if your patch works, you can remove the small_logo.svg in res folder instead. Could you please double-check it?

mgrudzinska: I'm not sure whether I understand. Do you want to be able to call `doxygen some_path/Doxygen` ?
if that's the case then we have 2 problems with paths: with logo and with the awesome_css.

after this patch I can produce valid html by calling doxygen from the docs directory. It didn;t work before since I don;t have access to your home dir:
```
- PROJECT_LOGO           = /home/hermet/Projects/thorvg/res/small_logo.svg
+ PROJECT_LOGO           = small_logo.svg
```

mgrudzinska: I removed small_logo.svg from res, I think it's better when all files used in docs are in the docs directory

hermet: @mgrudzinska Ok, Could you please change the icon stuff only here? (as the patch title mentioned) You don't need to change the Doxyfile up to your system version in this change. Because the file can be modified to the last system used.

hermet (changes requested):

PR #859, opened 2021-10-01
infra: add all option to enable all loader formats.

hermet:
just for developer convenience (including me)

ex) $meson . build -Dloaders="all" ...

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -20,19 +20,25 @@ if get_option('engines').contains('gl') == true
     config_h.set10('THORVG_GL_RASTER_SUPPORT', true)
 endif
 
-if get_option('loaders').contains('svg') == true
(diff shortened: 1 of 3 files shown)


mgrudzinska (changes requested):
1. meson.build in src/loaders has to be edited as well
2. I'd keep `''` as an option (no loaders)

in general I find this change very very useful :)

hermet:
>     1. meson.build in src/loaders has to be edited as well
> 
>     2. I'd keep `''` as an option (no loaders)
> 
> 
> in general I find this change very very useful :)

ok, makes sense! thanks.

mgrudzinska (approved):

mmaciola (approved):

PR #860, opened 2021-10-01
capi: replaced bounds() api with the latest.

hermet:
The next api of c++ version has been deprecated

Tvg_Result tvg_paint_get_bounds(const Tvg_Paint* paint, float* x, float* y, float* w, float* h);

instead, we introduce the next one under the beta.

Tvg_Result tvg_paint_get_bounds(const Tvg_Paint* paint, float* x, float* y, float* w, float* h, bool transformed);

### DIFF
--- a/src/bindings/capi/thorvg_capi.h
+++ b/src/bindings/capi/thorvg_capi.h
@@ -813,21 +813,21 @@ TVG_EXPORT Tvg_Paint* tvg_paint_duplicate(Tvg_Paint* paint);
 
 
 /*!
-* \brief Gets the bounding box of the Tvg_Paint object before any transformation. (BETA_API)
+* \brief Gets the axis-aligned bounding box of the Tvg_Paint object. (BETA_API)
 *
 * \param[in] paint The Tvg_Paint object of which to get the bounds.
 * \param[out] x The x coordinate of the upper left corner of the object.
 * \param[out] y The y coordinate of the upper left corner of the object.
 * \param[out] w The width of the object.
 * \param[out] h The height of the object.
-* \param[in] transformed if @c true, apply the transformation of the paint.
+* \param[in] transformed If @c true, the transformation of the paint is taken into account, otherwise it isn't.
 *
 * \return Tvg_Result enumeration.
 * \retval TVG_RESULT_SUCCESS Succeed.
 * \retval TVG_RESULT_INVALID_ARGUMENT An invalid Tvg_Paint pointer.
 * \retval TVG_RESULT_INSUFFICIENT_CONDITION Other errors.
 *
-* \note Transformation of an object changes the returned values.
+* \note The bounding box doesn't indicate the actual drawing region. It's the smallest rectangle that encloses the object.
 */
 TVG_EXPORT Tvg_Result tvg_paint_get_bounds(const Tvg_Paint* paint, float* x, float* y, float* w, float* h, bool transformed);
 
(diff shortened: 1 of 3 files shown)


mihashco:
@hermet. Probably old API is used in capiPaint.cpp. You plan to modify test in separated commit ?

hermet:
@mihashco yep, i added in this patch thread.

src/bindings/capi/thorvg_capi.h
```diff
@@ -813,13 +813,14 @@ TVG_EXPORT Tvg_Paint* tvg_paint_duplicate(Tvg_Paint* paint);
 
 
 /*!
-* \brief Gets the bounding box of the Tvg_Paint object before any transformation.
+* \brief Gets the bounding box of the Tvg_Paint object before any transformation. (BETA_API)
```
mgrudzinska, line 816: #868 :
``` 
* \brief Gets the axis-aligned bounding box of the paint object. (BETA_API)
*
* In case \p transform is \c true, all object's transformations are applied first, and then the bounding box is established. Otherwise, the bounding box is determined before any transformations.
```

src/bindings/capi/thorvg_capi.h
```diff
@@ -813,13 +813,14 @@ TVG_EXPORT Tvg_Paint* tvg_paint_duplicate(Tvg_Paint* paint);
...
 *
 * \param[in] paint The Tvg_Paint object of which to get the bounds.
 * \param[out] x The x coordinate of the upper left corner of the object.
 * \param[out] y The y coordinate of the upper left corner of the object.
 * \param[out] w The width of the object.
 * \param[out] h The height of the object.
+* \param[in] transformed if @c true, apply the transformation of the paint.
```
mgrudzinska, line 823: #868 :
```
* \param[in] transformed If \c true, the paint's transformations are taken into account, otherwise they aren't.
```

src/bindings/capi/thorvg_capi.h
```diff
@@ -813,22 +813,23 @@ TVG_EXPORT Tvg_Paint* tvg_paint_duplicate(Tvg_Paint* paint);
 
 
 /*!
-* \brief Gets the bounding box of the Tvg_Paint object before any transformation.
+* \brief Gets the axis-aligned bounding box of The Tvg_Paint object. (BETA_API)
```
mgrudzinska, line 816: The -> the

src/bindings/capi/thorvg_capi.h
```diff
@@ -813,22 +813,23 @@ TVG_EXPORT Tvg_Paint* tvg_paint_duplicate(Tvg_Paint* paint);
...
 *
 * \param[in] paint The Tvg_Paint object of which to get the bounds.
 * \param[out] x The x coordinate of the upper left corner of the object.
 * \param[out] y The y coordinate of the upper left corner of the object.
 * \param[out] w The width of the object.
 * \param[out] h The height of the object.
+* \param[in] transformed if @c true, the transformation of the paint is taken into account, otherwise it isn't.
```
mgrudzinska, line 823: if -> If

src/bindings/capi/thorvg_capi.h
mgrudzinska: I'd add somewhere the more detailed description, but I see, that u don't :) so as you wish

mgrudzinska (approved):

PR #861, opened 2021-10-01
loaders: prepare for static jpg/png loaders.

hermet:
This patch has 2 purposes,

1. revise the loaders infrastructure to support both static/external linking loaders.
2. add a template for static jpg/png loaders after external jpg/png.

Our default loaders prefer static linking, external loaders are only available
when dependent libraries on the build system are found.

You might wonder why we have the external loaders together,
they might be faster than static loaders since the popular libraries are likely to be well maintained,
fine-grained optimized.

Thus in this patch, meson tries to apply the external loaders first
then see if the dependencies were found or not,
if it's failed, it turns to the default static loaders.

Next this patch, we need the contribution for actual static jpg/png loaders implementation.

@Issue: https://github.com/Samsung/thorvg/issues/594

### DIFF
--- a/src/loaders/external_jpg/meson.build
+++ b/src/loaders/external_jpg/meson.build
@@ -1,0 +1,18 @@
+source_file = [
+   'tvgJpgLoader.h',
+   'tvgJpgLoader.cpp',
+]
+
+jpg_dep = dependency('libturbojpeg', required: false)
+
+if not jpg_dep.found()
+	jpg_dep = cc.find_library('turbojpeg', required: false)
+endif
+
+if jpg_dep.found()
+    subloader_dep += [declare_dependency(
+        include_directories : include_directories('.'),
+        dependencies : jpg_dep,
+        sources : source_file
+    )]
+endif
--- a/src/loaders/meson.build
+++ b/src/loaders/meson.build
@@ -1,24 +1,30 @@
 subloader_dep = []
 
+if get_option('loaders').contains('tvg') == true
+    subdir('tvg')
+endif
+
 if get_option('loaders').contains('svg') == true
     subdir('svg')
 endif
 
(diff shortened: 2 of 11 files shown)


src/loaders/meson.build
```diff
@@ -1,24 +1,28 @@
...
 endif
 
-if get_option('loaders').contains('tvg') == true
-    subdir('tvg')
-endif
+subdir('external_jpg')
 
-if get_option('loaders').contains('jpg') == true
+if not jpg_dep.found()
     subdir('jpg')
```
JSUYA, line 20: I think the external_jpg folder and the jpg folder should have an `or` relationship.

```
if not jpg_dep.found()
  subdir('jpg')
else
  subdir('external_jpg')
endif
```

src/loaders/external_jpg/meson.build
```diff
@@ -0,0 +1,18 @@
+source_file = [
+   'tvgJpgLoader.h',
+   'tvgJpgLoader.cpp',
+]
+
+jpg_dep = dependency('libturbojpeg', required: false)
+
+if not jpg_dep.found()
+	jpg_dep = cc.find_library('turbojpeg', required: false)
+endif
```
JSUYA, line 10: I think this dependency check can be handled by the parent build script.

hermet: This dependency must be addressed here so that external_jpg can be isolated itself.
That's why I left it. Otherwise parent <-> external_jpg have the strong coupling.

src/loaders/meson.build
hermet: Probably, but we don't have dynamic linking mechanism yet.
Right now, we do support either module. it's enough to so far.

Issue #863, opened 2021-10-01
gl_engine: Use of attribute 'fallthrough' not supported until c++17


projectitis:
The file `gl_engine/tvgGlGeometry.cpp` contains the attribute `__attribute__ ((fallthrough));`.  This attribute tells the compiler that the fallthough in the case statement (not including `break`) is intentional.

There are two issues:

1. Fallthrough is only defined from c++17 onwards (default for ThorVG is c++14)
2. MSVC also does not recognise `__attribute__ ((fallthrough));` which must be replaced with `[[fallthrough]]`

projectitis:
If you think it is a good idea, could someone please add a new `label` for `portability`? This covers any changes to ensure code works across different compilers and different platforms.

hermet:
@projectitis added the label, thanks!

PR #864, opened 2021-10-01
common, gl_engine: portable fallthrough attribute

projectitis:
Define a portable version of  `__attribute__ ((fallthrough))` to `tvgCommon.h` and replaced it's use in `gl_engine/tvgGlGeometry.cpp`.

The attribute is only available from c++17 onward and has a different syntax in MSVC.

See #863

Could somebody please test this change on non-windows systems, please?


hermet (approved):
LGTM!

PR #866, opened 2021-10-03
all: fixing clang warnings

mgrudzinska:
fopen->fopen_s, strdup -> _strdup, strncpy -> strncpy_s
__declspec(dllexport) -> __attribute__ ((visibility ("default")))

I didn;t want to define _CRT_NONSTDC_NO_DEPRECATE to eliminate some warnings.

e1
e2
e3
e4

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -18,7 +18,7 @@
 #include <string>
 
 #ifdef TVG_BUILD
-    #ifdef _MSC_VER
+    #if defined(_MSC_VER) && !defined(__clang__)
         #define TVG_EXPORT __declspec(dllexport)
         #define TVG_DEPRECATED __declspec(deprecated)
     #else
@@ -306,7 +306,7 @@ public:
      * @return Result::Success when succeed, Result::InsufficientCondition otherwise.
      *
      * @note The bounding box doesn't indicate the actual drawing region. It's the smallest rectangle that encloses the object.
-     * 
+     *
      * @BETA_API
      */
     Result bounds(float* x, float* y, float* w, float* h, bool transformed) const noexcept;
--- a/src/savers/tvg/tvgTvgSaver.cpp
+++ b/src/savers/tvg/tvgTvgSaver.cpp
@@ -31,6 +31,20 @@
     #include <alloca.h>
 #endif
 
+static FILE* _fopen(const char* filename, const char* mode)
+{
+#ifdef __clang__
+    FILE *fp;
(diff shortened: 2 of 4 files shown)


projectitis (approved):
MSVC has no problem with these changes :)

inc/thorvg.h
```diff
@@ -18,7 +18,7 @@
 #include <string>
 
 #ifdef TVG_BUILD
-    #ifdef _MSC_VER
+    #if defined(_MSC_VER) && !defined(__clang__)
```
JSUYA, line 21: Is it ok to leave `_MSC_VER` in `./src/lib/sw_engine/tvgSwMath.cpp` without modification?

projectitis: It looks like clang supports `__builtin_clz`, so I would recommend to also change `_MSC_VER` in that file to `#if defined(_MSC_VER) && !defined(__clang__)`

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -25,6 +25,11 @@
 #include "tvgCommon.h"
 #include "tvgArray.h"
 
+#ifdef __clang__
```
hermet, line 28: This can be commonly used. I'd like to ask you to put this in tvgCommon.h

src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -181,8 +184,14 @@ bool TvgSaver::saveEncoding(const std::string& path)
     memcpy(uncompressed, &compressedSizeBits, TVG_HEADER_COMPRESSED_SIZE_BITS);
 
     //Good optimization, flush to file.
+#ifdef __clang__
```
hermet, line 187: Not negative at this but it would be better to split to a function for removing duplicates.

static FILE* _fopen(...)
{
#ifdef __clang__
#else
#endif
}

hermet (changes requested):

inc/thorvg.h
mgrudzinska: done

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: done

src/savers/tvg/tvgTvgSaver.cpp
mgrudzinska: done

hermet (approved):

PR #869, opened 2021-10-04
docs: c and c++ apis docs improved

mgrudzinska:


mihashco (approved):

inc/thorvg.h
```diff
@@ -1144,8 +1145,8 @@ class TVG_EXPORT Scene final : public Paint
     /**
      * @brief Passes drawing elements to the Scene using Paint objects.
      *
-     * Only pushed paints in the scene will be drawing targets.
-     * They are retained by the scene until you call Scene::clear().
+     * Only the paints pushed into the scene will be drawing targets.
+     * The paints are retained by the scene until the Scene::clear() is called.
```
hermet, line 1149: hmm... is it proper to use "the" for the method()?
https://www.cairographics.org/manual/cairo-cairo-scaled-font-t.html

hermet (commented):
@mgrudzinska could you please check a comment?

inc/thorvg.h
mgrudzinska: well... it sounds ok here. I can remove if you wish, it'll be still correct. 
I'm not a native :)
I changed here "Only the paints pushed into the scene will be drawing targets." -> "Only the paints pushed into the scene will be the drawn targets."

mgrudzinska: I also changed "in the advance" -> "in advance"

hermet: Question: Are they wrong sentence? 

Cairo: "as they would be drawn by cairo_show_glyphs() "
Direct3D: "A handle to a DLL that implements a software rasterizer. If DriverType is D3D_DRIVER_TYPE_SOFTWARE, Software must not be NULL. Get the handle by calling LoadLibrary, LoadLibraryEx , or GetModuleHandle."

Simpler code and docs are better for users. if it doesn't violate the (programming) grammar.

mgrudzinska: as I said, I'm not a native. fixed

PR #878, opened 2021-10-06
infra : add build option

JSUYA:


JSUYA:
I wanted to include unit tests. However, the test all pass could not be done due to a file load (probably a permission problem) problem.

PR #880, opened 2021-10-06
examples, tests: replace backslashes with forward slashes for portability

projectitis:
Replace backslashes with forward slashes in `EXAMPLE_DIR` and `TEST_DIR` (meson).
These path defines had backslashes in windows, which were acting as escape characters in strings.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -6,8 +6,8 @@ project('thorvg',
 
 config_h = configuration_data()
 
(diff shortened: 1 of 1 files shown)


JSUYA (approved):
Currently, the meson version used in the build script of the github action is 0.45.
This will be improved later.

I have checked that this change works well for the current version.
And https://github.com/Samsung/thorvg/pull/878 also needs this PR. :)

PR #886, opened 2021-10-07
vs2019: Add the preset visual studio project

JSUYA:
This is a compilable vs project to generate library for the fan of visual studio.

Still, there are a lot of compile warnings, it works to generate working dll and lib.


vtorri:
Hmm, meson can create a VS solution. I use this with meson 0.63.0:

```
rm -rf builddir_vs

meson setup builddir_vs \
      --backend vs \
      --vsenv \
      --prefix=/opt/thor_64 \
      --buildtype=debug \
      --default-library shared \
      -Dvector=true
```

note that --vsenv works correctly with a recent meson.

About __max, what about using [std::max](https://en.cppreference.com/w/cpp/algorithm/max) ?

hermet:
> Hmm, meson can create a VS solution. I use this with meson 0.63.0:
> 
> ```
> rm -rf builddir_vs
> 
> meson setup builddir_vs \
>       --backend vs \
>       --vsenv \
>       --prefix=/opt/thor_64 \
>       --buildtype=debug \
>       --default-library shared \
>       -Dvector=true
> ```
> 
> note that --vsenv works correctly with a recent meson.
> 
> About __max, what about using [std::max](https://en.cppreference.com/w/cpp/algorithm/max) ?

@vtorri Good. tip! and any issues on __max() ?

vtorri:
about __max, no issue afaics. It's just that std::max removes a #ifdef

hermet:
> about __max, no issue afaics. It's just that std::max removes a #ifdef

std::max makes sense to me.

JSUYA:
> Hmm, meson can create a VS solution. I use this with meson 0.63.0:

wow Thank you for good information. I don't think this PR is needed anymore. :)

> About __max, what about using [std::max](https://en.cppreference.com/w/cpp/algorithm/max) ?

I think this should be handled in a new PR, not this pr :)

PR #893, opened 2021-10-10
svg_loader: define the scaling factors for units changing

mgrudzinska:
The values in absolute units are calculated based on the pixel unit.
Till now the default value of 90dpi was used. The necessary constants
are kept, but the new ones for 96dpi are defined as well and used by default.

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -64,6 +64,17 @@
 /* Internal Class Implementation                                        */
 /************************************************************************/
 
+/*
+ * According to: https://www.w3.org/TR/SVG2/coords.html#Units
+ * and: https://www.w3.org/TR/css-values-4/#absolute-lengths
+ */
+#define PX_PER_IN 96        //1 in = 96 px
+#define PX_PER_PC 16        //1 pc = 1/6 in  -> PX_PER_IN/6
+#define PX_PER_PT 1.333333f //1 pt = 1/72 in -> PX_PER_IN/72
+#define PX_PER_MM 3.779528f //1 in = 25.4 mm -> PX_PER_IN/25.4
+#define PX_PER_CM 37.79528f //1 in = 2.54 cm -> PX_PER_IN/2.54
+
+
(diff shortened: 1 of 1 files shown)


JSUYA:
Neat code :) I just have questions. 
Are there any cases that 90DPI needed? what cases is it used?
If it is necessary, shouldn't it be defined in the common header? (PX_PER_IN)

hermet:
@mgrudzinska ping?

mgrudzinska:
> Neat code :) I just have questions. Are there any cases that 90DPI needed? what cases is it used? If it is necessary, shouldn't it be defined in the common header? (PX_PER_IN)

As I understand it, the values of 90 or 72 were used in CSS1 and CSS2, but now the units are related to each other by the fixed ratio and any other values used to switch between the units can be considered as obsolete. So since CSS defines 1 inch to be equal to 96 pixels, I modified the code and kept only the 96 value.
Nevertheless please read the text below, maybe I misunderstood something.


From CSS1 (https://www.w3.org/TR/CSS1/#units):
"Pixel units, as used in the last rule, are relative to the resolution of the canvas, i.e. most often a computer display. If the pixel density of the output device is very different from that of a typical computer display, the UA should rescale pixel values. The suggested reference pixel is the visual angle of one pixel on a device with a pixel density of 90dpi and a distance from the reader of an arm's length. For a nominal arm's length of 28 inches, the visual angle is about 0.0227 degrees."

But since CSS3 (the most recent version: https://www.w3.org/TR/css-values-4/#absolute-lengths):
"Note: This definition of the pixel unit and the physical units differs from the earlier editions of CSS1 and CSS2. In particular, in previous versions of CSS the pixel unit and the physical units were not related by a fixed ratio: the physical units were always tied to their physical measurements while the pixel unit would vary to most closely match the reference pixel. (This unfortunate change was made because too much existing content relies on the assumption of 96dpi, and breaking that assumption broke the content.)"

JSUYA (approved):
Thanks for the detailed explanation.
We are based on the SVG2 specification. maybe? :) 
So I think it's right to change it to 96 PPI.
I've found a few cases where they're using units, but I don't think it's a problem right now.

PR #897, opened 2021-10-11
common: new api for a grad transformation

mgrudzinska:
The new api allows to transform the gradient fill.

we need that to properly transform gradient used in svg.


hermet:
@mgrudzinska could you please add its TC as well?

PR #918, opened 2021-10-14
common: defining macro with the overloaded new operator

mgrudzinska:
New throws a bad_alloc exception. To get a nullptr we have to call
its overloaded version new(std::nothrow).

I've changed new -> NEW_NOTHROW only in some places to show how it will look. 
If we agree to apply this approach I'll change it everywhere of course.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -16,6 +16,9 @@
 
 #include <memory>
 #include <string>
+#include <new>
+
+#define NEWT new(std::nothrow)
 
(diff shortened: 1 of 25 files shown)


hermet:
@mgrudzinska ok, i expected the new returns the nullptr. it says it could only work on the old compilers. I think we can apply your suggetion but i recommend you to use simpler MACRO NAME such as "NEW" or "TNEW", we have to use this everywhere everytime mandatory.

PR #919, opened 2021-10-15
capi examples: fixed build issue.

mihashco:
Initialize API now use Tvg_Engine instead of integer.


mgrudzinska:
oh, I think it should be fixed inside thorvg_capi... It's because of the change that I made...

PR #920, opened 2021-10-15
loader png: support static png loader 

hermet:
Copied only necessary decoding functions from the lodepng opensource.
See: https://lodev.org/lodepng/

additional changes:
- disabled crc for the optimal size.
- converted the format bgr -> rgb for our png example.

Still we don't have a concrete idea for the image formats,
we should fix the format converting between bgra <-> rgba later.

@Issue: https://github.com/Samsung/thorvg/issues/594


hermet:
@mmaciola about colorspace issues. I know, we don't have the correct policy yet. Actually, we need to build up it from the infrastructure. So this case, I just only build it for our basic example working, not others else.

mmaciola:
@hermet I understand your approach. We can merge it as a base and resolve these issues in a separate pr

PR #926, opened 2021-10-21
sw_engine: Fixed rendering region calculation

mihashco:
Changes: 

Removed unnecessary shift in max value of render region. In verification picture all letters should have 12px width. 
Before change: "P" letters has 11 px width nad "l" letter has 13 px witdth (tested on 800x800 viewport). After change width of the letters is the same.

Before:
invalid
After: 
valid

Reference (inkscape)
reference


hermet:
@mihashco hmm.. just noticed  P round edge is worse than before.... :(

PR #932, opened 2021-10-22
common: Introduced Accessor for traversing the scene-tree. 

hermet:
Basically, this Accessor is a utility to debug the Scene structure,
by traversing the scene-tree by users.

You can search specific nodes to read the property information,
figure out the structure of the scene tree and its size.

We actually don't recommend you to touch the property unless you really
know the each paint's position and role because it's not visible, difficult to
understand its anatomy.

Also, You must underatnd that modifying the nodes of the scene will be going
well with both the art-design structure and the prorgram logic.

In this first version, Accessor only supports for the Picture class.

@example:
```
picture->load("test.svg");

//The callback function from lambda expression.
//This function will be called for every paint nodes of the tree.
{
    if (paint->identifier() == Shape::identifier()) {
        auto shape = (tvg::Shape*) paint;
        //override properties?
        uint8_t r, g, b, a;
        shape->fillColor(&r, &g, &b, &a);
        shape->fill(r / 2, g / 2, b / 2, a);
    }

    //You can return false, to stop traversing immediately.
    return true;
};

auto accessor = tvg::Accessor::gen();

picture = accessor->access(move(picture), f);
...
```
@issue: https://github.com/Samsung/thorvg/issues/693


fire:
Hi! Can you rebase this on 0.6.0?

hermet:
> Hi! Can you rebase this on 0.6.0?
> 
> Edited:
> 
> I also have to make a lot of Shape* nonconst to be able to modify the data.

Yes. i intended to put it actually, because modifying the internals is not the standard. That let users to know unwilling jobs they are doing... but at least we can give a chance and someone like you can utilize(hack) it. if you don't agree we can discuss it further.

hermet:
@fire rebase is done!

fire:
> Yes. i intended to put it actually, because modifying the internals is not the standard. That let users to know unwilling jobs they are doing... but at least we can give a chance and someone like you can utilize(hack) it. if you don't agree we can discuss it further.

So you wanting to remove const?

hermet:
> > Yes. i intended to put it actually, because modifying the internals is not the standard. That let users to know unwilling jobs they are doing... but at least we can give a chance and someone like you can utilize(hack) it. if you don't agree we can discuss it further.
> 
> So you [are] wanting to remove const?

@fire  no, not actually. but later when spec is concrete and stable.

fire:
I have trouble getting 0.6 to work, rendering ordinary svgsare failing and I don’t know where to start.

hermet:
@fire  Ah this patch looks weird.... bad, Something has been corrupted. need to track back.

hermet:
@fire please have a look at this patch https://github.com/Samsung/thorvg/pull/1127

PR #972, opened 2021-10-31
capi: init and term apis args

mgrudzinska:
These two apis expected Tvg_Engine enum as an arg - bitwise
operation is allowed. The enum -> int conversion is implicit, but
not the other way. Since the bitwise OR performed on enums returns
the int type, the explicit conversion (int -> enum) was necessary.
This is fixed by changing the arg type to be int.

### DIFF
--- a/src/examples/Capi.cpp
+++ b/src/examples/Capi.cpp
@@ -100,7 +100,7 @@ void testCapi()
     tvg_canvas_push(canvas, shape2);
 
 
-//////3. Radial gradient shape with a radial dashed stroke
+//////3. Radial gradient shape with a radial gradient stroke
     //Set a shape
     Tvg_Paint* shape3 = tvg_shape_new();
     tvg_shape_append_rect(shape3, 550.0f, 20.0f, 100.0f, 50.0f, 0.0f, 0.0f);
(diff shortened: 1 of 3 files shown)


src/examples/Capi.cpp
```diff
@@ -269,7 +269,7 @@ void resize_cb(void *data, Evas *e, Evas_Object *obj, void *event_info)
 int main(int argc, char **argv)
 {
     elm_init(argc, argv);
-    tvg_engine_init(Tvg_Engine(TVG_ENGINE_SW | TVG_ENGINE_GL), 0);
```
hermet, line 272: @mihashco  @mgrudzinska btw any issues at this? our thorvg capi example has no problems.

mgrudzinska: @hermet because we did the explicit casting:
``` tvg_engine_init(Tvg_Engine(TVG_ENGINE_SW | TVG_ENGINE_GL), 0);```

mgrudzinska: we can not call this funct like this ``` tvg_engine_init(TVG_ENGINE_SW | TVG_ENGINE_GL, 0);```
do you think is that ok that the user has to remember about the casting ?

hermet: Well.. though it's not standard, I don't much see the trouble here. They can easily understand & apply the casting when it's necessary, otherwise they will copy from our examples. If you think this is a matter, then bit operation for the enum is another issue as well. So... please choose, leave it with good example code or revert your old patch: https://github.com/Samsung/thorvg/pull/914/files  :)

hermet (changes requested):
please check my coment.

src/examples/Capi.cpp
mgrudzinska: @hermet if it's ok with you I think we can keep it as it is now.

PR #975, opened 2021-11-01
sw_engine: var/funct renaming

mgrudzinska:
Changed names:
shape->rect -> shape->fastTrack
_factTrack -> _axisAlignedRect

The check whether a shape is an axis aligned rectangle should be done
only if the shape has no clips.

(a shape can be a rect but still its `fastTrack` can be false)

### DIFF
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -215,7 +215,7 @@ struct SwShape
     SwRleData*   strokeRle = nullptr;
     SwBBox       bbox;           //Keep it boundary without stroke region. Using for optimal filling.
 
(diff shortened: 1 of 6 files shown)


hermet (changes requested):
Thanks for your effort but please fix a bug, then coming again with a separate code refactoring... (or vice versa)
This kind of code refactoring change is too digression. (Meaning that, hard to catch the core part by the reviewers... Not good habit.)

mgrudzinska:
@hermet please check #977, here I'll do only the refactoring, after #977 is merged

hermet:
> @hermet please check #977, here I'll do only the refactoring, after #977 is merged

#977 has been applied. thanks.

hermet (approved):
thanks.

PR #979, opened 2021-11-02
infra meson: Add -Dstatic_library option

JSUYA:
Add an option for use in environments where
the use of static libraries must be enforced.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -47,3 +47,8 @@ option('log',
     type: 'boolean',
     value: false,
     description: 'Enable log message')
+
+option('static',
+    type: 'boolean',
+    value: false,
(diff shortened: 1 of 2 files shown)


meson_options.txt
```diff
@@ -47,3 +47,8 @@ option('log',
     type: 'boolean',
     value: false,
     description: 'Enable log message')
+
+option('static_library',
```
hermet, line 51: I'd like to suggest a simpler option name here.
static_library => static

hermet (commented):

meson_options.txt
JSUYA: I updated patch. Thank you

hermet (approved):
thanks

Issue #429, opened 2021-06-10
SVG ClipPath doesn't work


hermet:
Try in the example:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/clip-path

mgrudzinska:
1) ```clipPathUnits``` added in #985 

2) ```clip-path="circle() fill-box"``` format is not supported in tvg for now

hermet:
Current result:

![thorvg](https://user-images.githubusercontent.com/3711518/230284329-e8bf95af-a60e-4f63-8da5-47eb1c9746e1.png)

mgrudzinska:
the expected result (the css styling isn't supported everywhere): 
<img width="215" alt="Zrzut ekranu 2023-04-7 o 02 18 51" src="https://user-images.githubusercontent.com/67589014/230516108-448ad540-5209-4793-aed0-cc9200e466b4.png">

the top left will work after we implement #661 and #985 - both very old and has to be changed

hermet:
@mgrudzinska  Hmm... clippath can be simply composed using a single shape. Probably, svg loader could achieve it without scene-clippath. it's better for performance.

hermet:
the result checked on the latest:

![image](https://user-images.githubusercontent.com/3711518/236726248-7fd5a15b-e71c-4a7d-b493-e17d63deedc1.png)

mgrudzinska:
after #1547 :

<img width="515" alt="Zrzut ekranu 2023-07-30 o 11 40 08" src="https://github.com/thorvg/thorvg/assets/67589014/0edb4d7a-f6df-4537-bef3-dea53a551e3e">

On Chrome, the results remain the same. The rest of these samples include CSS styling that isn't widely supported. Therefore, once we reach a consensus on #1547, we can close this issue. If necessary, we can open a new one to address the extension of CSS support.

PR #661, opened 2021-07-26
sw_engine clipping: allow scene for ClipPath

mmaciola:
This patch changes the sw_engine, introducing SwSceneTask and allowing to use
scene for ClipPath. Union region of rle is being created.

@issue: #524

PR #985, opened 2021-11-02
svg_loader: clipPathUnits introduced

mgrudzinska:
The clipPathUnits can have one of two possible values:
userSpaceOnUse (the defalut one) and objectBoundingBox.

before:
44before

after:
44after

examples:
[clipEX.zip](https://github.com/Samsung/thorvg/files/7461444/clipEX.zip)

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -47,14 +47,12 @@
  * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
  * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
-
-#include <math.h>
 #include <string>
 #include "tvgSvgLoaderCommon.h"
 #include "tvgSvgSceneBuilder.h"
 #include "tvgSvgPath.h"
 #include "tvgSvgUtil.h"
-#include <float.h>
+#include "tvgMath.h"
 
 static bool _appendShape(SvgNode* node, Shape* shape, float vx, float vy, float vw, float vh, const string& svgPath);
 static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, float vx, float vy, float vw, float vh, const string& svgPath, bool mask);
@@ -228,12 +226,36 @@ static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float
         if (compNode && compNode->child.count > 0) {
             node->style->clipPath.applying = true;
 
+            auto isTransform = (!compNode->node.comp.userSpace ? true : false);
+            Matrix finalTransform = {1, 0, 0, 0, 1, 0, 0, 0, 1};
+            if (isTransform) finalTransform = {vw, 0, vx, 0, vh, vy, 0, 0, 1};
+
+            if (node->transform) {
+                if (isTransform) finalTransform = mathMultiply(node->transform, &finalTransform);
+                else {
+                    finalTransform = *node->transform;
+                    isTransform = true;
+                }
+            }
+ 
+            // Although svg viewers (like w3school) display different results than ThorVG, 
+            // no justification for this behaviour was found in the SVG specs. 
+            // The approach below seams to be in compliance with the SVG standard for both,
+            // the case of the objectBoundingBox and of the userSpaceOnUse.
+            if (compNode->transform) {
+                if (isTransform) finalTransform = mathMultiply(&finalTransform, compNode->transform);
+                else {
+                    finalTransform = *compNode->transform;
+                    isTransform = true;
+                }
+            }
+
             auto comp = Shape::gen();
             comp->fill(255, 255, 255, 255);
-            if (node->transform) comp->transform(*node->transform);
+            if (isTransform) comp->transform(finalTransform);
 
             auto child = compNode->child.data;
-            auto valid = false; //Composite only when valid shapes are existed
+            auto valid = false; //Composite only when valid shapes exist
 
             for (uint32_t i = 0; i < compNode->child.count; ++i, ++child) {
                 if (_appendChildShape(*child, comp.get(), vx, vy, vw, vh, svgPath)) valid = true;
@@ -356,6 +378,15 @@ static void _applyProperty(SvgNode* node, Shape* vg, float vx, float vy, float v
         vg->stroke(style->stroke.paint.color.r, style->stroke.paint.color.g, style->stroke.paint.color.b, style->stroke.opacity);
     }
 
+    if (node->style->clipPath.node && !node->style->clipPath.node->node.comp.userSpace) {
+        vg->bounds(&vx, &vy, &vw, &vh, false);
+        if (auto strokeW = vg->strokeWidth()) {
+            vx += 0.5f * strokeW;
+            vy += 0.5f * strokeW;
+            vw -= strokeW;
+            vh -= strokeW;
+        }
+    }
     _applyComposition(vg, node, vx, vy, vw, vh, svgPath);
 }
 
@@ -539,6 +570,9 @@ static unique_ptr<Picture> _imageBuildHelper(SvgNode* node, float vx, float vy, 
         picture->transform(m);
     }
 
+    if (node->style->clipPath.node && !node->style->clipPath.node->node.comp.userSpace) {
+        picture->bounds(&vx, &vy, &vw, &vh, false); //TODO: bounds without any strokes
+    }
     _applyComposition(picture.get(), node, vx, vy, vw, vh, svgPath);
     return picture;
 }
@@ -578,6 +612,9 @@ static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, float vx, float 
                     auto shape = _shapeBuildHelper(*child, vx, vy, vw, vh, svgPath);
                     if (shape) scene->push(move(shape));
                 }
+            }
+            if (node->style->clipPath.node && !node->style->clipPath.node->node.comp.userSpace) {
+                scene->bounds(&vx, &vy, &vw, &vh, false); //TODO: bounds without any strokes
             }
             _applyComposition(scene.get(), node, vx, vy, vw, vh, svgPath);
             scene->opacity(node->style->opacity);


src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -350,6 +350,7 @@ struct SvgNode
         SvgImageNode image;
     } node;
     bool display;
+    bool objectBBox = false;
```
mgrudzinska, line 353: I'd need that for `mask` too. Not sure about patterns and filters

JSUYA: I think we need to create a structure for clippath because the clippathUnit you add now is not a common property of SvgNode.
for example
`struct SvgClipPathNode` ?

JSUYA: and
We are using userSpace as member variable name related to `userSpaceOnUse | objectBoundingBox` attribute. (Gradient case)
So, I think `bool userSpace` is better.
what do you think?

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -363,6 +376,15 @@ static void _applyProperty(SvgNode* node, Shape* vg, float vx, float vy, float v
...
     }
 
+    if (node->style->clipPath.node && node->style->clipPath.node->objectBBox) {
+        vg->bounds(&vx, &vy, &vw, &vh, false);
+        if (auto strokeW = vg->strokeWidth()) {
+            vx += 0.5f * strokeW;
+            vy += 0.5f * strokeW;
+            vw -= strokeW;
+            vh -= strokeW;
+        }
```
JSUYA, line 386: This code started to duplicate. It seems to be possible to extract by method.

JSUYA (commented):
Thank you for your work. Please check my comment. :)

mgrudzinska:
@JSUYA thanks for the rev
please take a look on #992 as an introduction to this commit (I'll change this one accordingly)

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: as I marked in some places, I'd need some other version of bounds api, so this part is only a temporary solution

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: @JSUYA for the gradient `objectBoundingBox` is the default units value. so `userSpace == true` only on request.
For clip and mask, the default is `userSpaceOnUse`, so by the analogy, I decided to use a different name, so that `objectBBox == true` only on request.

do you think it's less clear? if so I can change it

mgrudzinska:
@JSUYA @hermet please notice the comment that I made in the code:

"Although svg viewers (like w3school) display different results than ThorVG, no justification for this behaviour was found in the SVG specs. The approach below seams to be in compliance with the SVG standard for both,  the case of the objectBoundingBox and of the userSpaceOnUse."

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: @JSUYA I'll take care of that in a separate pr, since I need to solve the problem with bounds without any strokes for the scene

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -356,6 +378,15 @@ static void _applyProperty(SvgNode* node, Shape* vg, float vx, float vy, float v
         vg->stroke(style->stroke.paint.color.r, style->stroke.paint.color.g, style->stroke.paint.color.b, style->stroke.opacity);
     }
 
+    if (node->style->clipPath.node && !node->style->clipPath.node->node.comp.userSpace) {
+        vg->bounds(&vx, &vy, &vw, &vh, false);
+        if (auto strokeW = vg->strokeWidth()) {
```
JSUYA, line 383:
```
<clipPath id="myClipOBB" clipPathUnits="objectBoundingBox" transform="matrix(1 0 0 1 -10 -10)">
<circle cx="0.5" cy="0.5" r="0.15"/>
</clipPath>

<clipPath id="myClipUS" transform="matrix(1 0 0 1 0 0)">
<circle cx="50" cy="50" r="15" />
</clipPath>


<rect x="25" y="25" width="50" height="50"  fill="red" clip-path="url(#myClipUS)"  fill-opacity="0.5" transform="matrix(1 0 0 1 0 0)" /> 

<rect x="0" y="0" width="100" height="100" fill="blue" clip-path="url(#myClipOBB)" fill-opacity="0.5" transform="matrix(1 0 0 1 10 10)" />
</svg>


```
Even if stroke is not declared, the default value is 1 .
This doesn't seem to be the intended result.

JSUYA (commented):
Hi @mgrudzinska 
I've modified your sample a bit and tested it.

```
<clipPath id="myClipOBB" clipPathUnits="objectBoundingBox" transform="matrix(1 0 0 1 -10 -10)">
<circle cx="0.5" cy="0.5" r="0.15"/>
</clipPath>

<clipPath id="myClipUS" transform="matrix(1 0 0 1 0 0)">
<circle cx="50" cy="50" r="15" />
</clipPath>

<rect x="0" y="0" width="100" height="100"  fill="red" clip-path="url(#myClipUS)"  fill-opacity="0.5" transform="matrix(1 0 0 1 0 0)" /> 
<rect x="0" y="0" width="100" height="100" fill="blue" clip-path="url(#myClipOBB)" fill-opacity="0.5" transform="matrix(1 0 0 1 10 10)" />
</svg>

```

The circles of the displayed result(chrome browser) have the same position and size.
But in our result, no blue circle is output.
As you said, w3school (or chrome) could be wrong.
But when I checked the final matrix value, there seems to be a problem with the translate value(dx, dy).

```
~/dev/os/thorvg (master) 15:43:40 $ git diff src/loaders/
diff --git a/src/loaders/svg/tvgSvgSceneBuilder.cpp b/src/loaders/svg/tvgSvgSceneBuilder.cpp
index bab096e1..f89ebc8c 100644
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -243,7 +243,15 @@ static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float
             // The approach below seams to be in compliance with the SVG standard for both,
             // the case of the objectBoundingBox and of the userSpaceOnUse.
             if (compNode->transform) {
-                if (isTransform) finalTransform = mathMultiply(&finalTransform, compNode->transform);
+                if (isTransform){
+                    finalTransform = mathMultiply(&finalTransform, compNode->transform);
+
+                    printf("%f %f %f \n%f %f %f\n%f %f %f\n", finalTransform.e11, finalTransform.e12, finalTransform.e13,
+                                                              finalTransform.e21, finalTransform.e22, finalTransform.e23,
+                                                              finalTransform.e31, finalTransform.e32, finalTransform.e33);
+                    printf("\n\n\n");
+
+                }
                 else {
                     finalTransform = *compNode->transform;
                     isTransform = true;
~/dev/os/thorvg (master) 15:43:42 $ 
~/dev/os/thorvg (master) 15:43:42 $ 
~/dev/os/thorvg (master) 15:43:42 $ 
~/dev/os/thorvg (master) 15:43:42 $ ./build/src/examples/Svg
tvg engine: software
SVG: //home/junsu/dev/os/thorvg/src/examples/images/bojo.svg
1.000000 0.000000 0.000000 
0.000000 1.000000 0.000000
0.000000 0.000000 1.000000



100.000000 0.000000 -990.000000 
0.000000 100.000000 -990.000000
0.000000 0.000000 1.000000

```

I can't give you any good comment on how to solve this problem as I haven't been find an exact guide either.
(But at least the result of this example of chrome seems to be correct.)
I think it's also a way to put the problem on hold until we find the way....

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: @JSUYA thank you very much for the detailed rev.
please check my comments.

1. >Even if stroke is not declared, the default value is 1 .
This doesn't seem to be the intended result.

It's because the bounds() api is not the one that I really need here... but please lets keep it here as it is and I'll prepare a separate pr with some internal implementation of bounds for the svg loader. 
as you wrote, since the def value of the stroke width = 1, so this value is taken into account in the bounds() api, so I need to subtract this value here -> added in bounds, subtracted here. So the result is as it should be, but the way to get it should be improved :)


2. >The circles of the displayed result(chrome browser) have the same position and size.
But in our result, no blue circle is output.
As you said, w3school (or chrome) could be wrong.
But when I checked the final matrix value, there seems to be a problem with the translate value(dx, dy).

the result from tvg is as I'd expect. the `transform="matrix(1 0 0 1 -10 -10)"` is the same as `transform="translate(-10 -10)"`. When `objectBoundingBox` is used, these values are treated as they are percentages of the object's bbox. so for the rect from the example, we get: 100 * (-10) = -1000, and since you also transformed the path (+10), we got -1000 + 10 = -990. 

It works exactly the same for gradients (please notice the transformation matrix for the LinearGradient1):

JSUYA: Ok
I understand that there is a problem with the transform of clippath in chrome and w3school.
I have one more question.

```
<clipPath id="myClipOBB" clipPathUnits="objectBoundingBox" >
<circle cx="0.5" cy="0.5" r="0.3" transform="matrix(1 0 0 1 0.1 0.1)"/>
</clipPath>

<clipPath id="myClipUS">
<circle cx="50" cy="50" r="30" transform="matrix(1 0 0 1 10 10)"/>
</clipPath>


<rect x="0" y="0" width="100" height="100"  fill="red" clip-path="url(#myClipUS)"  fill-opacity="0.5" transform="matrix(1 0 0 1 0 0)" /> 

<rect x="0" y="0" width="100" height="100" fill="blue" clip-path="url(#myClipOBB)" fill-opacity="0.5" transform="matrix(1 0 0 1 0 0)" />
</svg>

```

In this example, I think the transform applied to circle should be affected by objectBoundingBox.
In chrome, both circles are output the same. (This behavior is the same behavior we intend.)
but in thovg, the output is wrong.
( transform="matrix(1 0 0 1 0 0)"  output is also wrong)
It seems that the children of clippath(Scene) should be affected by OBB in the same way.
I know it's getting more and more complicated, too.

What do you think?

JSUYA: and about the stroke..
If you use style->stroke.paint.none on line 349, you could be able to check whether or not there is a stroke.
What do you think?

src/loaders/svg/tvgSvgSceneBuilder.cpp
hermet: Doesn't this simple work?
```
auto stroke = vg->strokeWidth(); 
vg->stroke(0);  //excluding stroke width.
vg->bounds(&vx, &vy, &vw, &vh, false);
vg->stroke(stroke);
```

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: @JSUYA I'm so sorry that from the very beginning this patch is not perfect...
I was wrong about saying that clip shouldn't be a scene... It actually has to a scene, since otherwise we can not use different transformations for the shapes.

So I was wrong here:
https://github.com/Samsung/thorvg/pull/988#discussion_r742362843
clipPath is not a container element, but still it has to be a scene in our implementation.

I'll come back to this pr after #661 is merged

hermet:
@mgrudzinska it would be nicer if you update the patch.

mgrudzinska:
> @mgrudzinska it would be nicer if you update the patch, let's close #429

@hermet before 0.9 is released? please give me some timelines

hermet:
> > @mgrudzinska it would be nicer if you update the patch, let's close #429
> 
> @hermet before 0.9 is released? please give me some timelines. I can do it asap, but I need to know how urgent this is

Of course not 0.9. it's up to you after it.

hermet:
> > @mgrudzinska it would be nicer if you update the patch, let's close #429
> 
> @hermet before 0.9 is released? please give me some timelines. I can do it asap, but I need to know how urgent this is

Of course not 0.9. it's up to you after it.

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: > Ok I understand that there is a problem with the transform of clippath in chrome and w3school. I have one more question.
> 
> ```
> <svg id="svg1" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
> <clipPath id="myClipOBB" clipPathUnits="objectBoundingBox" >
> <circle cx="0.5" cy="0.5" r="0.3" transform="matrix(1 0 0 1 0.1 0.1)"/>
> </clipPath>
> 
> <clipPath id="myClipUS">
> <circle cx="50" cy="50" r="30" transform="matrix(1 0 0 1 10 10)"/>
> </clipPath>
> 
> 
> <rect x="0" y="0" width="100" height="100"  fill="red" clip-path="url(#myClipUS)"  fill-opacity="0.5" transform="matrix(1 0 0 1 0 0)" /> 
> 
> <rect x="0" y="0" width="100" height="100" fill="blue" clip-path="url(#myClipOBB)" fill-opacity="0.5" transform="matrix(1 0 0 1 0 0)" />
> </svg>
> ```
> 
> In this example, I think the transform applied to circle should be affected by objectBoundingBox. In chrome, both circles are output the same. (This behavior is the same behavior we intend.) but in thovg, the output is wrong. ( transform="matrix(1 0 0 1 0 0)" output is also wrong) It seems that the children of clippath(Scene) should be affected by OBB in the same way. I know it's getting more and more complicated, too.
> 
> What do you think?

solved in #1532

mgrudzinska:
replaced by #1547

PR #987, opened 2021-11-02
WASM: Use ARGB8888. Convert into ABGR and unmultiply after sync()

mmaciola:
Wasm needs ABGR STRAIGHT buffer.
As there is a problem using ABGR for loaders, it is better to use ARGB8888.
This patch makes wasm using ARGB8888 and relocate pixels and unmultiply after sync().


hermet:
@mmaciola looks good to me, until thorvg stable its feasibility, it would be better to handle it in user-side.

hermet (approved):

PR #988, opened 2021-11-03
svg_loader: mask node loaded as a scene

mgrudzinska:
Since the mask color is important, the mask can not be saved
as a shape. To preserve the color information, each shape
is saved separately and pushed into the final scene.

NOTE:
the presented results are drawn after pr #620 is applied

before:
45before

after:
45after

code:
```
<svg viewBox="-10 -10 120 120">
  <mask id="myMask">
    <rect x="0" y="0" width="100" height="100" fill="white" />
    <path d="M10,35 A20,20,0,0,1,50,35 A20,20,0,0,1,90,35 Q90,65,50,95 Q10,65,10,35 Z" fill="black" />
  </mask>

  <polygon points="-10,110 110,110 110,-10" fill="orange" />
  <circle cx="50" cy="50" r="50" mask="url(#myMask)" />
</svg>
```
------------------------

before:
45before2

after:
45after2

code:
```
<svg viewBox="-10 -10 120 120">
  <mask id="myMask">
  <g fill="red">
    <rect x="0" y="0" width="50" height="20" />
    <g fill="white">
      <rect x="50" y="0" width="100" height="100" />
    </g>
  </g>
  </mask>

  <circle cx="50" cy="50" r="50" mask="url(#myMask)" fill="yellow"/>
</svg>
```

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -56,7 +56,7 @@
 #include "tvgSvgUtil.h"
 #include <float.h>
 
-static bool _appendShape(SvgNode* node, Shape* shape, float vx, float vy, float vw, float vh);
+static bool _appendShape(SvgNode* node, Shape* shape, float vx, float vy, float vw, float vh, const string& svgPath);
 static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, float vx, float vy, float vw, float vh, const string& svgPath, bool mask);
 
 /************************************************************************/
@@ -199,16 +199,16 @@ static unique_ptr<RadialGradient> _applyRadialGradientProperty(SvgStyleGradient*
 }
 
 
-static bool _appendChildShape(SvgNode* node, Shape* shape, float vx, float vy, float vw, float vh)
+static bool _appendChildShape(SvgNode* node, Shape* shape, float vx, float vy, float vw, float vh, const string& svgPath)
 {
     auto valid = false;
 
-    if (_appendShape(node, shape, vx, vy, vw, vh)) valid = true;
+    if (_appendShape(node, shape, vx, vy, vw, vh, svgPath)) valid = true;
 
     if (node->child.count > 0) {
         auto child = node->child.data;
         for (uint32_t i = 0; i < node->child.count; ++i, ++child) {
-            if (_appendChildShape(*child, shape, vx, vy, vw, vh)) valid = true;
+            if (_appendChildShape(*child, shape, vx, vy, vw, vh, svgPath)) valid = true;
         }
     }
 
@@ -216,7 +216,7 @@ static bool _appendChildShape(SvgNode* node, Shape* shape, float vx, float vy, f
 }
 
 
-static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float vy, float vw, float vh)
+static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float vy, float vw, float vh, const string& svgPath)
 {
     /* ClipPath */
     /* Do not drop in Circular Dependency for ClipPath.
@@ -236,7 +236,7 @@ static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float
             auto valid = false; //Composite only when valid shapes are existed
 
             for (uint32_t i = 0; i < compNode->child.count; ++i, ++child) {
-                if (_appendChildShape(*child, comp.get(), vx, vy, vw, vh)) valid = true;
+                if (_appendChildShape(*child, comp.get(), vx, vy, vw, vh, svgPath)) valid = true;
             }
 
             if (valid) paint->composite(move(comp), CompositeMethod::ClipPath);
@@ -255,7 +255,7 @@ static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float
         if (compNode && compNode->child.count > 0) {
             node->style->mask.applying = true;
 
-            auto comp = _sceneBuildHelper(compNode, vx, vy, vw, vh, "", true);
+            auto comp = _sceneBuildHelper(compNode, vx, vy, vw, vh, svgPath, true);
             if (comp) {
                 if (node->transform) comp->transform(*node->transform);
                 paint->composite(move(comp), CompositeMethod::AlphaMask);
@@ -267,7 +267,7 @@ static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float
 }
 
 
-static void _applyProperty(SvgNode* node, Shape* vg, float vx, float vy, float vw, float vh)
+static void _applyProperty(SvgNode* node, Shape* vg, float vx, float vy, float vw, float vh, const string& svgPath)
 {
     SvgStyleProperty* style = node->style;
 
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -199,6 +199,27 @@ static unique_ptr<RadialGradient> _applyRadialGradientProperty(SvgStyleGradient*
 }
 
 
+static bool _appendScene(SvgNode* node, Scene* scene, float vx, float vy, float vw, float vh)
```
JSUYA, line 202: I hope that this operation also considers the clip-path.
(Of course I know that this should be solved first.(https://github.com/Samsung/thorvg/issues/524)......)

JSUYA (commented):
I know that the current logic only adds the last declared node to the composition. So this pr is necessary.
But I don't think this PR has anything to do with ColorMask.
A  ColorMask patch does not have to be preceded if the scene composition is properly applied.

```html
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"  viewBox="-10 -10 120 120">
  <mask id="myMask">
    <rect x="0" y="0" width="50" height="50" fill="red" />
    <rect x="50" y="50" width="50" height="50" fill="red" />
  </mask>

  <polygon points="-10,110 110,110 110,-10" fill="orange" />
  <circle cx="50" cy="50" r="50" mask="url(#myMask)" fill="red"/>
</svg>
```
image

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: no, it does not, since the clipPath is not a container element (https://www.w3.org/TR/SVG11/intro.html#TermContainerElement):

"container element - An element which can have graphics elements and other container elements as child elements. Specifically: ‘a’, ‘defs’, ‘glyph’, ‘g’, ‘marker’, ‘mask’, ‘missing-glyph’, ‘pattern’, ‘svg’, ‘switch’ and ‘symbol’."

so according to the svg standard a clip can not be a scene (on the svg_loader site, as I understand, the api can work as we want, so #661 can still be applied)

mgrudzinska:
> I know that the current logic only adds the last declared node to the composition. So this pr is necessary. But I don't think this PR has anything to do with ColorMask. A ColorMask patch does not have to be preceded if the scene composition is properly applied.

I've changed the description. what I meant is, that the result of the example that you presented is not correct, because a ColorMask is not supported yet. but even if it was, a slight modification of this example:
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"  viewBox="-10 -10 120 120">
  <mask id="myMask">
    <rect x="0" y="0" width="50" height="50" fill="red" />
    <rect x="50" y="50" width="50" height="50" fill="blue" />  <!-- red changed to blue -->
  </mask>

  <polygon points="-10,110 110,110 110,-10" fill="orange" />
  <circle cx="50" cy="50" r="50" mask="url(#myMask)" fill="red"/>
</svg>
```
still won't work correctly without this commit.

mgrudzinska:
@JSUYA I applied your comments, thank you very much for them.
I still have to find out whether svgPath is necessary in masking (so whether an image can be a mask). So I've changed to draft.

mgrudzinska:
> @JSUYA I applied your comments, thank you very much for them. I still have to find out whether svgPath is necessary in masking (so whether an image can be a mask). So I've changed to draft.

yes, an image can be the mask node

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -564,9 +557,9 @@ static unique_ptr<Scene> _useBuildHelper(const SvgNode* node, float vx, float vy
 }
 
 
-static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, float vx, float vy, float vw, float vh, const string& svgPath)
+static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, float vx, float vy, float vw, float vh, const string& svgPath, bool mask)
 {
-    if (_isGroupType(node->type)) {
+    if (_isGroupType(node->type) || mask) {
```
JSUYA, line 562: How about adding `SvgNodeType::Mask` to the if condition of _isGroupType?
I haven't tested that method yet. Could you please check it?

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -577,16 +570,16 @@ static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, float vx, float
...
                         scene->push(_useBuildHelper(*child, vx, vy, vw, vh, svgPath));
                     else
-                        scene->push(_sceneBuildHelper(*child, vx, vy, vw, vh, svgPath));
+                        scene->push(_sceneBuildHelper(*child, vx, vy, vw, vh, svgPath, false));
                 } else if ((*child)->type == SvgNodeType::Image) {
                     auto image = _imageBuildHelper(*child, vx, vy, vw, vh, svgPath);
                     if (image) scene->push(move(image));
-                } else {
-                    auto shape = _shapeBuildHelper(*child, vx, vy, vw, vh);
+                } else if ((*child)->type != SvgNodeType::Mask) {
```
JSUYA, line 577: I don't think this change is necessary if MaskType is added to isGroup.

src/loaders/svg/tvgSvgSceneBuilder.cpp
JSUYA: Oh I didn't know this spec at all. Thanks for URL :)

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: It won't work. I need to go inside this 'if condition' for the 'SvgNodeType::Mask' only if the _sceneBuildHelper was called from `applyComposite`.

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: If Mask is added to the isGroup, I'll jump into the l.576 (569) and it's not what I want. 
I added this condition as the _shapeBuildHelper won't be called, but it's not necessary since the result of _shapeBuildHelper for the mask will be false - it's just for the performance

mgrudzinska:
@JSUYA please read my comments

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -255,26 +255,19 @@ static void _applyComposition(Paint* paint, const SvgNode* node, float vx, float
...
-            if (node->transform) comp->transform(*node->transform);
-
-            auto child = compNode->child.data;
-            auto valid = false; //Composite only when valid shapes are existed
-
-            for (uint32_t i = 0; i < compNode->child.count; ++i, ++child) {
-                if (_appendChildShape(*child, comp.get(), vx, vy, vw, vh)) valid = true;
+            auto comp = _sceneBuildHelper(compNode, vx, vy, vw, vh, svgPath, true);
+            if (comp) {
+                if (node->transform) comp->transform(*node->transform);
```
JSUYA, line 260: (Sorry for not reviewing all at once.)
comp(scene) has already a transform matrix when there is transform information in the mask at line 564.
This code will overwrite the transform information.
You'll probably have to use matrix multiply.

ex)
```html
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"  viewbox="-10 -10 120 120"> 
  <mask id="mymask" transform="translate(50 0)">
    <rect x="0" y="0" width="50" height="50" fill="red" />
    <rect x="50" y="50" width="50" height="50" fill="blue" />  <!-- red changed to blue -->
  </mask>

  <polygon points="-10,110 110,110 110,-10" fill="orange" />
  <circle cx="50" cy="50" r="50" mask="url(#mymask)" transform="translate(-25 0)" fill="red"/>
</svg>
```

image

src/loaders/svg/tvgSvgSceneBuilder.cpp
JSUYA: Ok i got it. thank you for check :)

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: @JSUYA thank you for such a detailed rev!
I changed the code, but in a different way than you asked, since according to the standard a mask can not contain transform attribute (and a clipPath can :) https://www.w3.org/TR/SVG11/masking.html)
so in this case overwriting was good :D
so if we build a scene that is a mask, the transformation is skipped now

mgrudzinska: so the result from the image viewer is incorrect :)

JSUYA: oh thank you for check.
I often use this site to check the specification. https://developer.mozilla.org/en-US/docs/Web/SVG/Element/mask
There is a transform in global attribute in this url, so I thought we should support transform.
anyway, ubuntu image viewer is incorrect. :)

JSUYA (approved):
:+1:

PR #989, opened 2021-11-03
apis: promote beta apis to the official ones.

hermet:
these apis are good to open in the next release.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -306,8 +306,6 @@ public:
      * @return Result::Success when succeed, Result::InsufficientCondition otherwise.
      *
      * @note The bounding box doesn't indicate the actual drawing region. It's the smallest rectangle that encloses the object.
-     *
-     * @BETA_API
      */
     Result bounds(float* x, float* y, float* w, float* h, bool transformed) const noexcept;
 
@@ -410,8 +408,6 @@ public:
      * @param[in] m The 3x3 augmented matrix.
      *
      * @return Result::Success when succeed, Result::FailedAllocation otherwise.
-     *
-     * @BETA_API
      */
     Result transform(const Matrix& m) noexcept;
 
@@ -437,8 +433,6 @@ public:
      * In case no transformation was applied, the identity matrix is returned.
      *
      * @retval The augmented transformation matrix.
-     *
-     * @BETA_API
      */
     Matrix transform() const noexcept;
 
--- a/src/bindings/capi/thorvg_capi.h
+++ b/src/bindings/capi/thorvg_capi.h
@@ -839,7 +839,7 @@ TVG_EXPORT Tvg_Paint* tvg_paint_duplicate(Tvg_Paint* paint);
 
 
 /*!
-* \brief Gets the axis-aligned bounding box of the Tvg_Paint object. (BETA_API)
+* \brief Gets the axis-aligned bounding box of the Tvg_Paint object.
 *
 * \param[in] paint The Tvg_Paint object of which to get the bounds.
 * \param[out] x The x coordinate of the upper left corner of the object.
@@ -1694,7 +1694,7 @@ TVG_EXPORT Tvg_Result tvg_gradient_get_spread(const Tvg_Gradient* grad, Tvg_Stro
 
 
 /*!
-* \brief Sets the matrix of the affine transformation for the gradient object. (BETA_API)
+* \brief Sets the matrix of the affine transformation for the gradient object.
 *
 * The augmented matrix of the transformation is expected to be given.
 *
@@ -1710,7 +1710,7 @@ TVG_EXPORT Tvg_Result tvg_gradient_set_transform(Tvg_Gradient* grad, const Tvg_M
 
 
 /*!
-* \brief Gets the matrix of the affine transformation of the gradient object. (BETA_API)
+* \brief Gets the matrix of the affine transformation of the gradient object.
 *
 * In case no transformation was applied, the identity matrix is set.
 *


hermet:
@JSUYA  @mihashco  @mmaciola  @mgrudzinska  Please leave feedbacks if you have any. Thanks.

mgrudzinska:
transform apis are ok, but the bounds... 
in the svg loader I need almost the same api, but without taking a stroke into account... like this one:
https://github.com/Samsung/thorvg/commit/96c58db223fe658b3910083e1e215ed8ae81f27f

It's necessary for introducing clipPathUnits and maskContentUnits

hermet:
> transform apis are ok, but the bounds... in the svg loader I need almost the same api, but without taking a stroke into account... like this one: [96c58db](https://github.com/Samsung/thorvg/commit/96c58db223fe658b3910083e1e215ed8ae81f27f)
> 
> It's necessary for introducing clipPathUnits and maskContentUnits

@mgrudzinska 
I think we can regard it with a different topic because Shape::bounds() must be considered, not Paint::bounds(). For instance, if you use the bounds without stroke for the Picture or Scene, what is the result user expect?

Plus, I'm wondering now, is this a common feature for users? => good to go, otherwise, we don't need to to provide it unless actual user comes out.

JSUYA:
> > transform apis are ok, but the bounds... in the svg loader I need almost the same api, but without taking a stroke into account... like this one: [96c58db](https://github.com/Samsung/thorvg/commit/96c58db223fe658b3910083e1e215ed8ae81f27f)
> > It's necessary for introducing clipPathUnits and maskContentUnits
> 
> @mgrudzinska I think we can regard it with a different topic because Shape::bounds() must be considered, not Paint::bounds(). For instance, if you use the bounds without stroke for the Picture or Scene, what is the result user expect?
> 
> Plus, I'm wondering now, is this a common feature for users? => good to go, otherwise, we don't need to to provide it unless actual user comes out.

I'm not sure User needs the transformed bounds API.
We recommend that in this case we provide bounds and matrix(or they saved it) and the user does the calculations himself.

+
About stroke with bounds is for a similar reason.

And~ stroke with bounds is difficult to provide the intended exact size by simply calculating the stroke width size.
(ex If the stroke with the `miter(join option)` is very sharp(?), the pixel will be out of half the width.)

mgrudzinska:
> @mgrudzinska I think we can regard it with a different topic because Shape::bounds() must be considered, not Paint::bounds(). For instance, if you use the bounds without stroke for the Picture or Scene, what is the result user expect?

That's what I need - the bounds without the stroke for a scene - if a scene has the shapes and these shapes have thestrokes, I need to know the bounds without the strokes.

hermet:
> > > transform apis are ok, but the bounds... in the svg loader I need almost the same api, but without taking a stroke into account... like this one: [96c58db](https://github.com/Samsung/thorvg/commit/96c58db223fe658b3910083e1e215ed8ae81f27f)
> > > It's necessary for introducing clipPathUnits and maskContentUnits
> > 
> > 
> > @mgrudzinska I think we can regard it with a different topic because Shape::bounds() must be considered, not Paint::bounds(). For instance, if you use the bounds without stroke for the Picture or Scene, what is the result user expect?
> > Plus, I'm wondering now, is this a common feature for users? => good to go, otherwise, we don't need to to provide it unless actual user comes out.
> 
> I'm not sure User needs the transformed bounds API. We recommend that in this case we provide bounds and matrix(or they saved it) and the user does the calculations himself.

@JSUYA  Untransformed bounds is less useful actually ( I can't see any useful scenario), if you're thinking the picking coordinates, they needs the transformed. If they do it themselves, they need to handle the shapes across the tree. too paintful.

hermet:
> > @mgrudzinska I think we can regard it with a different topic because Shape::bounds() must be considered, not Paint::bounds(). For instance, if you use the bounds without stroke for the Picture or Scene, what is the result user expect?
> 
> That's what I need - the bounds without the stroke for a scene - if a scene has the shapes and these shapes have the strokes, I need to know the bounds without the strokes.

So im asking is your scenario is general? or very special case.

In some points, user could disable the stroke if they needs the bounds without stroke. very simple & intuitive.

JSUYA:
> transform apis are ok, but the bounds... in the svg loader I need almost the same api, but without taking a stroke into account... like this one: [96c58db](https://github.com/Samsung/thorvg/commit/96c58db223fe658b3910083e1e215ed8ae81f27f)
> 
> It's necessary for introducing clipPathUnits and maskContentUnits

As for my opinion on this...
About bounds api with stroke width is difficult to provide the intended exact size by simply calculating the stroke width size.
If the stroke with the `miter(join option)` is very sharp(?), the pixel will be out of half the width.

mgrudzinska:
> As for my opinion on this... About bounds api with stroke width is difficult to provide the intended exact size by simply calculating the stroke width size. If the stroke with the `miter(join option)` is very sharp(?), the pixel will be out of half the width.

@JSUYA it is so true! 
code:
```
shape1->moveTo(10, 4);
shape1->cubicTo(6,7, 6,7, 2,3);
shape1->cubicTo(4,10, 10,7, 10,10);
shape1->stroke(255, 255, 0, 255); 
shape1->stroke(0.25);
shape1->stroke(tvg::StrokeJoin::Miter);
shape1->scale(100);
```
result (blue rect shows the bounds after the transformation)
bounds1

mgrudzinska:
> > That's what I need - the bounds without the stroke for a scene - if a scene has the shapes and these shapes have the strokes, I need to know the bounds without the strokes.
> 
> @mgrudzinska So i'm asking is your scenario is general? or very special case.

It's a general rule in the svg specs - not to use a stroke when bounds are established :) but ok, I got your point. I think we have all the information we need in the svg loader, so probably I can live without such an api, unless we decide to change the api because the topic raised by @JSUYA 

pict->bounds returns the picture size and the pict->size() api is not considered as a transformation - we should probably address this in the docs (I thought about 'size' as it is a scaling, so a transformation)

hermet:
> > > That's what I need - the bounds without the stroke for a scene - if a scene has the shapes and these shapes have the strokes, I need to know the bounds without the strokes.
> > 
> > 
> > @mgrudzinska So i'm asking is your scenario is general? or very special case.
> 
> It's a general rule in the svg specs - not to use a stroke when bounds are established :) but ok, I got your point. I think we have all the information we need in the svg loader, so probably I can live without such an api, unless we decide to change the api because the topic raised by @JSUYA
> 
> pict->bounds returns the picture size and the pict->size() api is not considered as a transformation - we should probably address this in the docs (I thought about 'size' as it is a scaling, so a transformation)

@mgrudzinska  We provide the stroke region when stroke is set. if you don't need it, you can set stroke width = 0  and then call bounds(). 

And you can update the picture::size() doc. picture size is a different factors from the geometry of the paint.

PR #992, opened 2021-11-03
svg_loader: SvgCompositeNode introduced

mgrudzinska:
Clips and masks require to implement some additional variables,
so the new struct has been introduced. Implementation of the usage
of these additional variables will be the subject of a separate
commit.

### DIFF
--- a/src/loaders/svg/tvgSvgLoaderCommon.h
+++ b/src/loaders/svg/tvgSvgLoaderCommon.h
@@ -215,6 +215,11 @@ struct SvgPolygonNode
     float* points;
 };
 
+struct SvgCompositeNode
+{
+    bool userSpace;
+};
+
 struct SvgLinearGradient
 {
     float x1;
@@ -225,7 +230,6 @@ struct SvgLinearGradient
     bool isY1Percentage;
     bool isX2Percentage;
     bool isY2Percentage;
-
 };
 
 struct SvgRadialGradient
@@ -348,6 +352,7 @@ struct SvgNode
         SvgPathNode path;
         SvgLineNode line;
         SvgImageNode image;
+        SvgCompositeNode comp;
     } node;
     bool display;
     ~SvgNode();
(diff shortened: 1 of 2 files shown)


src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -353,6 +354,11 @@ struct SvgNode
     ~SvgNode();
 };
 
+struct SvgCompositeNode : SvgNode
+{
+    bool objectBBox = false;
```
mgrudzinska, line 359: @JSUYA for the gradient `objectBoundingBox` is the default units value. so `userSpace == true` only on request.
For clip and mask, the default is `userSpaceOnUse`, so by the analogy, I decided to use a different name, so that `objectBBox == true` only on request.

do you think it's less clear? if so I can change it

JSUYA: if you give the meaning of default value, you are right.
I just thought that it would be better for if the unit of gradient and unit of clip-path were declared with similar names.
If someday we need to support pattern and support patternUnit, its name will be objectBBbox.
+
We already use bbox in engine code, but we decided not to use abbreviations, so I don't know if it's better to keep BBox. hahahaha....

JSUYA (commented):
Thank you for your work. But the new struct I was talking about wasn't this.
My opinion was to declare something like SvgDocNode or SvgUseNode and include it in the union of SvgNode. Sorry for not explaining clearly.

I think inheritance is appropriate, but I don't think it's the case with the current SvgNode structure.

What do you think?

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: changed

mgrudzinska:
@JSUYA thanks for your patience
please check it one more time...

hermet (approved):

PR #994, opened 2021-11-05
common Array: Add limit size check

JSUYA:
If user have many of memory space or in certain environment,
realloc may not return NULL even after allocating as much memory as MAX.
Therefore, input more than the size allowed by the size argument's type is restricted.

### DIFF
--- a/src/lib/tvgArray.h
+++ b/src/lib/tvgArray.h
@@ -45,6 +45,7 @@ struct Array
 
     bool reserve(uint32_t size)
     {
+        if (size >= UINT32_MAX) return false;
         if (size > reserved) {
             reserved = size;
             data = static_cast<T*>(realloc(data, sizeof(T) * reserved));


hermet (approved):

src/lib/tvgArray.h
```diff
@@ -45,6 +45,7 @@ struct Array
 
     bool reserve(uint32_t size)
     {
+        if (size >= UINT32_MAX) return false;
```
hermet, line 48: Question: curious about, how this size can be over the UINT32_MAX?...
you mean, if the size == UINT32_MAX it will be issued?

JSUYA: @hermet 
you are right. ` >= UINT32_MAX ` is wrong.
Actually, the problem is that there is case where realloc doesn't return NULL in OOM situation.
When I run the test and see the memory usage in real time with the system monitor app, realloc returns a non-NULL value when the memory is not enough.

I've searched several times for this reason, but haven't been able to find a definitive answer.
So, since our tvg::Array is mainly used for tvg objects, I thought setting a limit constant would be an alternative.


What do you think about modify `if (size == UNIT32_MAX) return false;`  and add that the limit of the reserve() API is UINT32_MAX to the description,

hermet: We can't guarantee the OOM situation. leave it to be (crashed) fine by me. 
That's our current our policy.

hermet: @JSUYA  please have a look at this patch.

PR #1006, opened 2021-11-06
tvg_saver: fix memory leaks

mgrudzinska:
In the case save() api is called, memory is released in all
scenarios but Result::MemoryCorruption - consistent with other apis
and less confiusing for the user.

### DIFF
--- a/src/savers/tvg/tvgTvgSaver.cpp
+++ b/src/savers/tvg/tvgTvgSaver.cpp
@@ -748,9 +748,6 @@ bool TvgSaver::save(Paint* paint, const string& path, bool compress)
 {
     close();
 
-    this->path = strdup(path.c_str());
-    if (!this->path) return false;
-
     float x, y;
     x = y = 0;
     paint->bounds(&x, &y, &vsize[0], &vsize[1], false);
@@ -763,6 +760,9 @@ bool TvgSaver::save(Paint* paint, const string& path, bool compress)
         TVGLOG("TVG_SAVER", "Saving paint(%p) has zero view size.", paint);
         return false;
     }
+
+    this->path = strdup(path.c_str());
+    if (!this->path) return false;
 
     this->paint = paint;
     this->compress = compress;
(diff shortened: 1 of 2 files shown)


test/capi/capiSavers.cpp
```diff
@@ -69,8 +69,6 @@ TEST_CASE("Save a paint into a file", "[capiSaver]")
     REQUIRE(tvg_saver_save(saver, paint3, TEST_DIR"/test.tvg", false) == TVG_RESULT_INSUFFICIENT_CONDITION);
 
     REQUIRE(tvg_saver_del(saver) == TVG_RESULT_SUCCESS);
-
-    tvg_paint_del(paint_empty);
```
mgrudzinska, line 73: after #1004 this is not needed (in the first version of this pr I also included the changes that are present in 1005, since I missed it)

src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -763,6 +760,9 @@ bool TvgSaver::save(Paint* paint, const string& path, bool compress)
         return false;
     }
 
+    this->path = strdup(path.c_str());
```
hermet, line 763: any reason to change this?

hermet (commented):

src/savers/tvg/tvgTvgSaver.cpp
mgrudzinska: if it stays where it was and the if condition in l.761 is true, who will released this memory?

hermet: ok, good catch thanks.

test/capi/capiSavers.cpp
hermet: Could you please remove this change in this PR? the patch was reverted already. Others looks good to me.

hermet (changes requested):
Please check a comment.

mgrudzinska:
@hermet please check now

hermet (approved):
great, thanks!

PR #1009, opened 2021-11-07
common: composition target released

mgrudzinska:
In the case when a composition method was set to none, the comp target
was not released. Fixed


JSUYA (approved):
I think it is right to delete the target(in return false) because the ownership of the target is passed when composite.

JSUYA:
https://github.com/Samsung/thorvg/pull/1017  This patch solves it.

PR #1010, opened 2021-11-07
sw_engine: fix memory leak

mgrudzinska:
After update the same task was pushed into the tasks array.
When one of them was released, the second pointer was invalid.


hermet:
@mgrudzinska any sample or examples you referenced?

mgrudzinska:
@hermet capiScene.cpp, test case "Scene reusing paints"

mgrudzinska:
I used meson option `-Db_sanitize=address`

hermet:
@mgrudzinska please have a look at this https://github.com/Samsung/thorvg/pull/1019

PR #1011, opened 2021-11-08
common Array: Add limit size check

JSUYA:
If user have many of memory space or in certain environment,
realloc may not return NULL even after allocating as much memory as MAX.
Therefore, input more than the size allowed by the size argument's type is restricted.

### DIFF
--- a/src/lib/tvgArray.h
+++ b/src/lib/tvgArray.h
@@ -45,6 +45,7 @@ struct Array
 
     bool reserve(uint32_t size)
     {
+        if (size == UINT32_MAX) return false;
         if (size > reserved) {
             reserved = size;
             data = static_cast<T*>(realloc(data, sizeof(T) * reserved));


JSUYA:
+ After https://github.com/Samsung/thorvg/pull/994 was closed, I did a force push (modify&rebase) and it did not reopen, so I created a new PR.

My explanation of OOM is wrong. This situation is not OOM.
Just my PC is using 32GB RAM and when I tested it, it succeeded in creating memory(UINT32_MAX * sizen(T)).

In environments with very large memory, this test will fail.
Therefore, we need to delete too big size test or limit the array.

hermet:
@JSUYA This patch couldn't help the issue because there won't be any guarantee of the exact size. If your assuming is right how to protect if the realloc() is failed at (UINT32_MAX -  alpha) ?   I believe realloc() will return nullptr at failure so this patch https://github.com/Samsung/thorvg/pull/1002 will resolve the problem. please double check it.

JSUYA:
> @JSUYA This patch couldn't help the issue because there won't be any guarantee of the exact size. If your assuming is right how to protect if the realloc() is failed at (UINT32_MAX - alpha) ? I believe realloc() will return nullptr at failure so this patch #1002 will resolve the problem. please double check it.

#1002 controls what happens when realloc returns NULL. #1002 is not related to this problem.
The problem is the case what realloc doesn't return NULL.
We wrote a test case expecting UINT32_MAX to return null.
On large machines, if you have enough memory, the test fails because it doesn't return null. That's why I said we should either delete the tc or set a limit.

hermet:
@JSUYA 
1. So... does this patch prevent your issue? .. how to guarantee it fails at UINT32_MAX?
2. realloc() will return null pointer when it's failed. so i think you assume is incorrect. see doc. https://linux.die.net/man/3/realloc If you argue out of standard, please ignore it.

hermet (changes requested):
Please check the comment.

JSUYA:
> @JSUYA
> 
> 1. So... does this patch prevent your issue? .. how to guarantee it fails at UINT32_MAX?
> 2. realloc() will return null pointer when it's failed. so i think you assume is incorrect. see doc. https://linux.die.net/man/3/realloc If you argue out of standard, please ignore it.

@hermet 
```
    //Too Big Size
    REQUIRE(scene->reserve(UINT32_MAX) == Result::FailedAllocation);

```
This is our tc. right?
If realloc succeeds UNIT32_MAX <- 34GB, is this TC a failure? is it a pass?
I'm not saying that realloc failed to allocate memory and didn't return NULL.
Because the Too Big Size we set may not be big on some machines, I am talking about arbitrary limit or tc deletion.

hermet:
> > @JSUYA
> > 
> > 1. So... does this patch prevent your issue? .. how to guarantee it fails at UINT32_MAX?
> > 2. realloc() will return null pointer when it's failed. so i think you assume is incorrect. see doc. https://linux.die.net/man/3/realloc If you argue out of standard, please ignore it.
> 
> @hermet
> 
> ```
>     //Too Big Size
>     REQUIRE(scene->reserve(UINT32_MAX) == Result::FailedAllocation);
> ```
> 
> If realloc succeeds UNIT32_MAX <- 34GB, is this TC a failure? is it a pass? I'm not saying that realloc failed to allocate memory and didn't return NULL. Because the Too Big Size we set may not be big on some machines, I am talking about arbitrary limit or tc deletion.

@JSUYA  Ok, I think that TC is wrong since it depends on the machine, (not guaranteed) so it must be removed.

JSUYA:
> > > @JSUYA
> > > 
> > > 1. So... does this patch prevent your issue? .. how to guarantee it fails at UINT32_MAX?
> > > 2. realloc() will return null pointer when it's failed. so i think you assume is incorrect. see doc. https://linux.die.net/man/3/realloc If you argue out of standard, please ignore it.
> > 
> > 
> > @hermet
> > ```
> >     //Too Big Size
> >     REQUIRE(scene->reserve(UINT32_MAX) == Result::FailedAllocation);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If realloc succeeds UNIT32_MAX <- 34GB, is this TC a failure? is it a pass? I'm not saying that realloc failed to allocate memory and didn't return NULL. Because the Too Big Size we set may not be big on some machines, I am talking about arbitrary limit or tc deletion.
> 
> @JSUYA Ok, I think that TC is wrong since it depends on the machine, (not guaranteed) so it must be removed.

Thank you for check. I will make a new patch :)

PR #1012, opened 2021-11-08
test Array: Remove too big size test

JSUYA:
On large memory machines, if machine have enough memory,
tests failed because it doesn't return null.


mgrudzinska:
@junsu testInitializer -> `REQUIRE(Initializer::init(CanvasEngine::Sw, -1) == Result::FailedAllocation);` 
 I think it's the same issue

hermet (approved):
thanks

hermet:
@mgrudzinska please see this: https://github.com/Samsung/thorvg/pull/1016

PR #1015, opened 2021-11-09
sw_engine: unification of mask and inv mask functions

mgrudzinska:
The function pointers used to pass the proper blending method - mask
or inverse mask.


mgrudzinska:
@hermet @JSUYA 
I checked all the cases using the very simple example (attached).
The performance is a bit better (~3-4%, so it may be not statistically significant... ). Nevertheless, it doesn't seam to be worse. 
I think the subtraction (255-alpha) changed to (~alpha) could be efficient

hermet (approved):
great, thanks!

PR #1022, opened 2021-11-09
sw_engine: missing implementations added

mgrudzinska:
Raster func for (inv)masked clipped images were missing .

before:
47before

after:
47after

example (calls all new funcs):
[Masking_missing_func.zip](https://github.com/Samsung/thorvg/files/7508282/Masking_missing_func.zip)


mgrudzinska:
builds after #1015

hermet:
@mgrudzinska could you please PR example as well?

mgrudzinska:
I didn't close it on purpose...  #1023

PR #1028, opened 2021-11-10
sw_engine: missing implementations added

mgrudzinska:
Rastering functions for masked images that are not rectangular
(clipped) were implemented.


mgrudzinska:
like #1022 but with the new refactoring - I didn't manage to do this correctly in #1022 :)

hermet:
@mgrudzinska I think you can write more descriptive commit message than "missing implement added"

for example)
Support rendering "translucentImageRleMask", ...

mgrudzinska:
@hermet sorry for that. please check now, would you prefer if I list the functions names in the commit msg?

hermet (approved):
thanks for your contribution!

PR #1029, opened 2021-11-10
example: MaskedClipPath added

mgrudzinska:


mgrudzinska:
@hermet I added the example as you asked, but I'm not sure whether we really need it. 
I used it to call all the new functions (#1028) and check them, but as I see we don;t have such a convention to prepare the examples so that they draw all possibilities.

hermet:
@mgrudzinska respect your opinion. thanks!

PR #1045, opened 2021-11-12
infra bot: Add Memcheck bot (valgrind)

JSUYA:
If the unit test is successful, valgrind memory check is executed based on the test.
If a leak occurs in the test result, the bot notifies the PR as a comment.
This notify may not be directly related to the created current PR.

(Asan is on hold because it is not well tested in the github action CI environment.)

(test example)
Screenshot from 2021-11-12 13-47-47

issue: https://github.com/Samsung/thorvg/issues/1032

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -63,7 +63,9 @@ jobs:
     - name: Install Packages
       run: |
         sudo apt-get update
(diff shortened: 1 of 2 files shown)


hermet (approved):
good job thanks!

PR #1046, opened 2021-11-12
common paint: fix invalid memory access in unit test

hermet:
casting the paint to shape is not allowed if the compositor target
is not shape, here it concretely checking the the type before casting....


JSUYA (approved):
LGTM

PR #1054, opened 2021-11-16
sw_engine image: improve the image scaling quality

hermet:
revise the image scaler.


hermet:
This is the prototype of the texture-mapping algorithm for the image-scaler. With this approach, we can apply the scale factor x/y separately. Yet we don't have any requests of this xy scale factors, just leave this patch under the draft.

PR #1061, opened 2021-11-17
example boundary2: added image boundary check sample. 

hermet:


mgrudzinska (commented):
I'm not a big fun of this example, as well as the Boundary.cpp. They are more like tests than examples for the user. 
So I'd merge them into one example - you could either replace some of the shapes with pictures or add them on the top.

But it's your call, the example is correct.

hermet:
> I'm not a big fun of this example, as well as the Boundary.cpp. They are more like tests than examples for the user. So I'd merge them into one example - you could either replace some of the shapes with pictures or add them on the top.
> 
> But it's your call, the example is correct.

@mgrudzinska I know your point. This is to verify the visual result of the ranging correction. we can't verify with the our api tests suite. yet examples is the best place we can put them in.

PR #1062, opened 2021-11-17
extension Lottie: Support Lottie animation

JSUYA:
Support Lottie animation.

issue: https://github.com/Samsung/thorvg/issues/692

ex)
image


hermet:
@mgrudzinska  @mmaciola  it would be better if you guys could review this as well.

mmaciola (changes requested):
I see some general logic problem: in the example, each time we want to move a single frame forward, lottie file is reopened and reparsed to load a single frame. It takes so much time, that on my Virtual Machine the window of the example doesn't show up at all. It is smooth when I limit displaying to 10 json files, above it starts glitchy and above 25, it can't show up.
Could you add some logic to store parsed lottie, reuse pictures in the example and add a function to rerender a given frame?

JSUYA:
> I see some general logic problem: in the example, each time we want to move a single frame forward, lottie file is reopened and reparsed to load a single frame. It takes so much time, that on my Virtual Machine the window of the example doesn't show up at all. It is smooth when I limit displaying to 10 json files, above it starts glitchy and above 25, it can't show up. Could you add some logic to store parsed lottie, reuse pictures in the example and add a function to rerender a given frame?

Thank you for your review.
You are right. In this case, Parsing did every time it is opened.
We haven't discussed the interface in detail yet.
I will make an improved prototype by applying patch #1020 based on your comments.

JSUYA:
Not everything is perfect, but I think the overall structure is ready to be reviewed.
@hermet @mmaciola  Could you please review this PR again?
```
src/extension/
├── lottie
│   ├── inc
│   │   ├── meson.build
│   │   └── thorvg_lottie.h
│   ├── lib
│   │   ├── meson.build
│   │   ├── tvgLottie.cpp
│   │   └── tvgLottieImpl.h
│   ├── loader
│   │   ├── external_lottie
│   │   │   ├── meson.build
│   │   │   ├── tvgLottieLoader.cpp
│   │   │   └── tvgLottieLoader.h
│   │   └── meson.build
│   └── meson.build
└── meson.build

```

JSUYA:
A few problems remain, so I'm changing it to draft.

JSUYA:
um.. I had not consider the reuse of the parsed data(json) in the lottie loader.
This caused performance problem.(Reloaded json data every frame.)
And some json was not drawn normally, I couldn't find the reason.

I couldn't start again this work due to another tasks. Sorry for late

PR #1064, opened 2021-11-17
all: fix compilation errors on non-windows clang

mmaciola:
Thorvg couldn't be compiled on macos as non-portable microsoft calls used.
Changed definitions checking to Visual Studio only.


mgrudzinska (approved):
checked on win (clang, gcc, msvc)

PR #1065, opened 2021-11-17
sw_engine raster: optimize the image raster

hermet:
these solid images shouldn't apply the alpha blending...


mgrudzinska (approved):
of course :)
surface->blender.ialpha(src) in all these cases is 0

PR #1067, opened 2021-11-18
sw_engine texmap: introduced texture mapping polygon drawing.

hermet:
Introduced the texture-mapping algorithm for the image-scaler.
With this approach, we can support any arbitrary forms of the transformed shapes
while appling the scale factors - x/y separately.

Also this is more efficient when the transformed image is far from the rectangular form
because our current normal image rasterizer based on the given drawing region whereas
our new approach is span-based.

Thus, we applied this mechanism only for the transformed images.


hermet:
This is the prototype of the texture-mapping algorithm for the image-scaler. With this approach, we can apply the scale factor x/y separately. Yet we don't have any requests of this xy scale factors, just leave this patch under the draft.

PR #1078, opened 2021-11-22
tvg_saver TvgBinInterpreter: Prevent misaligned memory access

JSUYA:
When parsing a binary stored as a char type,
interpreter can access the misaligned memory while accessing it with a pointer.
To prevent that, pass the array copied to memcpy as tvg Object.

issue: https://github.com/Samsung/thorvg/issues/1074


hermet (approved):

hermet:
@JSUYA thanks!

PR #1093, opened 2021-11-25
Mgrudzinska/svg loader 2

mgrudzinska:
there are 2 commits:
- in the first one I'm using Box instead of vx, vy, vw, vh
- in the second one I introduced function that uses bounds - I'll need it in some other places after clipPathUnits are introduced, so it's to prevent copying the same code

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -56,12 +56,19 @@
 #include "tvgSvgPath.h"
 #include "tvgSvgUtil.h"
 
-static bool _appendShape(SvgNode* node, Shape* shape, float vx, float vy, float vw, float vh, const string& svgPath);
-static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, float vx, float vy, float vw, float vh, const string& svgPath, bool mask);
-
 /************************************************************************/
 /* Internal Class Implementation                                        */
 /************************************************************************/
+
+struct Box
+{
+    float x, y, w, h;
+};
+
+
+static bool _appendShape(SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath);
+static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox, const string& svgPath, bool mask);
+
 
 static inline bool _isGroupType(SvgNodeType type)
 {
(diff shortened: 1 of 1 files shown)


mgrudzinska:
@JSUYA please take a look on the new funct - as u mentioned earlier, a stroke width is 1 by default. So if it is already in the shape, bounds() will take it into account, so I need to subtract it. Do you agree?

@hermet I remember that u suggested to do this like that:
```
s = vg->strokeWidth();
vg->stroke(0);
vg->bounds(...);
vg->stroke(s);
```
I don't want to change the shape, although I know that I revert the change 2 steps later. if for some reason it's a better approach, I'll apply it of course, but I don't see any advantage at this point

hermet:
> a stroke width is 1 by default.

@mgrudzinska  strange why stroke width is 1 in default. isn't it must be zero? only the reason, it's easy and simple to code. nothing more.

hermet:
@mgrudzinska btw, the commit message is a bit bad... 

svg_loader: Box struct introduced 
svg_loader: an additional internal function introduced

Nobody can't understand why you did this from your abstraction.... rather than describe the code itself, it would be better if you write the purpose of the patch. :)

JSUYA:
> @JSUYA please take a look on the new funct - as u mentioned earlier, a stroke width is 1 by default. So if it is already in the shape, bounds() will take it into account, so I need to subtract it. Do you agree?
>

Basically, stroke is not drawn unless defined "width"
Therefore, I think it is correct to assume that the default value of width is 0.
so. I don't think it's necessary to consider the stroke width in bounds when the stroke is not defined.

mgrudzinska:
> > a stroke width is 1 by default.
> 
> @mgrudzinska strange why stroke width is 1 in default... isn't it zero?

in the svg loader it has to be 1, since that's the svg standard. 
in the api we set the default width to be zero and it's also correct, so when the user calls bounds() he gets the result without any stroke (if it was not directly declared).

but since I need to use bounds() inside the svg loader I had to subtract the default stroke width.
consider 2 svgs:
```
<svg viewBox="0 0 100 100" >
<rect x="50" y="50" width="10" height="10" fill="red" stroke="blue"/>
</svg>
```
and 
```
<svg viewBox="0 0 100 100" >
<rect x="50" y="50" width="10" height="10" fill="red">
</svg>
```
One has a stroke and the other one doesn't, but bounds() called from inside the svg loader returns exactly the same result. It's ok, but that's why I need to make this subtraction.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -56,20 +56,45 @@
...
 static inline bool _isGroupType(SvgNodeType type)
 {
     if (type == SvgNodeType::Doc || type == SvgNodeType::G || type == SvgNodeType::Use || type == SvgNodeType::ClipPath) return true;
     return false;
 }
 
 
+//According to: https://www.w3.org/TR/SVG11/coords.html#ObjectBoundingBoxUnits (the last paragraph)
+//a stroke width should be ignored for bounding box calculations
+Box _boundingBox(const Shape* shape)
```
hermet, line 82: +static  ?
static Box _boundingBox(..)

hermet (approved):
please check a comment and submit by yourself 
thanks!

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: done, thx

PR #1097, opened 2021-11-26
sw_engine raster: improve the transformed rle image rasterizer.

hermet:
replace the transformed rle rgba image with the texmap raster.

this patch removes the several scattered transformed image rasterizer,
reuse the unified one texmap method instead.

yay, it's much clean and optimal.


hermet:
yet need to turn off the clipper shapes' anti-aliasing.

PR #1101, opened 2021-11-29
 sw_engine raster: fix the regression bug of the image simple scaling rendering

hermet:
Gave up the optimization this logic.

It brings too many digressed code, not good for maintenance...


JSUYA (approved):
I have checked that the test that caused the problem is fixed.

PR #1102, opened 2021-11-30
sw_engine SwRaster: Initialize uninitialized local variable

JSUYA:


src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -641,7 +641,8 @@ static bool _rasterUpScaledRleRGBAImage(SwSurface* surface, const SwImage* image
 static bool _transformedRleRGBAImage(SwSurface* surface, const SwImage* image, const Matrix* transform, uint32_t opacity)
 {
     Matrix itransform;
-    if (transform && !mathInverse(transform, &itransform)) return false;
```
hermet, line 644: @p transform can be nullptr.... need to check it as it was before.

Also, we can ignore the initialization of the matrix here.

Otherwise...
if (transform) mathInverse()
else mathIdentity()

hermet (changes requested):

src/lib/sw_engine/tvgSwRaster.cpp
JSUYA: I'm sorry I misread transform and itransform.

hermet (approved):

PR #1105, opened 2021-12-01
sw_engine raster: remove unnecessary functions.

hermet:
These alpha/inverse alpha blender table is not useful so far,
we remove them since it just decrease the performance by by-pass addressing.


JSUYA (changes requested):
A build error occurs when the `vector` option is true. (`tvgSwRasterNeon.h`, `tvgSwRasterAvx.h` )

hermet:
@JSUYA fixed

JSUYA (approved):

PR #1111, opened 2021-12-14
png_loaders: fix the wrong premultiplying operations.

hermet:
It should not modify the alpha channel value while premultiplying
that turned out with the wrong visual result.

@Issue: https://github.com/Samsung/thorvg/issues/655


JSUYA (approved):
LGTM

PR #1113, opened 2021-12-14
sw_engine texmap: implemented fast span-edge anti-aliasing

hermet:
This Anti-Aliasing mechanism is originated from Hermet Park's idea.
To understand this AA logic, you can refer this page:
www.hermet.pe.kr/122 (hermetpark@gmail.com)

@Issue: https://github.com/Samsung/thorvg/issues/161


hermet:
before
after

PR #1114, opened 2021-12-14
common: new api introduced to get the information about preserveAspec…

mgrudzinska:
…tRatio

The getter returns the preserveAspectRatio attribute of the loaded svg file.
True is returned in case th aspect ratio is preseved, false otherwise.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1184,6 +1184,15 @@ public:
     Result viewbox(float* x, float* y, float* w, float* h) const noexcept;
 
     /**
+     * @brief Gets the information about the preserveAspectRatio attribute of the loaded SVG picture.
+     *
+     * @warning Please do not use it, this API is not official one. It could be modified in the next version.
+     *
+     * @BETA_API
+     */
+    Result aspectRatio(bool* preserved) const noexcept;
+
+    /**
      * @brief Creates a new Picture object.
      *
      * @return A new Picture object.
--- a/src/bindings/capi/thorvg_capi.h
+++ b/src/bindings/capi/thorvg_capi.h
@@ -1856,6 +1856,14 @@ TVG_EXPORT Tvg_Result tvg_picture_get_size(const Tvg_Paint* paint, float* w, flo
 TVG_EXPORT Tvg_Result tvg_picture_get_viewbox(const Tvg_Paint* paint, float* x, float* y, float* w, float* h);
 
 
+/*!
+* \brief Gets the information about the preserveAspectRatio attribute of the loaded SVG picture. (BETA_API)
+*
+* \warning Please do not use it, this API is not official one. It can be modified in the next version.
+*/
+TVG_EXPORT Tvg_Result tvg_picture_get_aspect_ratio(const Tvg_Paint* paint, bool* preserved);
+
+
 /** \} */   // end defgroup ThorVGCapi_Picture
 
 
--- a/src/lib/tvgPictureImpl.h
+++ b/src/lib/tvgPictureImpl.h
@@ -156,6 +156,13 @@ struct Picture::Impl
         return true;
     }
 
+    bool aspectRatio(bool* preserved) const
+    {
+        if (!loader) return false;
+        if (preserved) *preserved = loader->preserveAspect;
+        return true;
+    }
+
     bool size(float w, float h)
     {
         this->w = w;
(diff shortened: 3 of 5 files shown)


mgrudzinska:
@hermet please check 268012 on gerrit - that's why I'd need this capi

inc/thorvg.h
```diff
@@ -1183,6 +1183,15 @@ class TVG_EXPORT Picture final : public Paint
...
     Result viewbox(float* x, float* y, float* w, float* h) const noexcept;
 
+    /**
+     * @brief Gets the information about the preserveAspectRatio attribute of the loaded SVG picture.
+     *
+     * @warning Please do not use it, this API is not official one. It could be modified in the next version.
+     *
+     * @BETA_API
+     */
+    Result preserveAspectRatio(bool* preserve) const noexcept;
```
hermet, line 1193: suggest:
bool aspectRatio() const noexcept;

src/bindings/capi/thorvg_capi.h
```diff
@@ -1856,6 +1856,14 @@ TVG_EXPORT Tvg_Result tvg_picture_get_size(const Tvg_Paint* paint, float* w, flo
 TVG_EXPORT Tvg_Result tvg_picture_get_viewbox(const Tvg_Paint* paint, float* x, float* y, float* w, float* h);
 
 
+/*!
+* \brief Gets the information about the preserveAspectRatio attribute of the loaded SVG picture. (BETA_API)
+*
+* \warning Please do not use it, this API is not official one. It can be modified in the next version.
+*/
+TVG_EXPORT Tvg_Result tvg_picture_get_preserve_aspect_ratio(const Tvg_Paint* paint, bool* preserve);
```
hermet, line 1864: suggest:
tvg_picture_get_aspect_ratio(const Tvg_Paint* paint, bool* preserve)

src/lib/tvgPictureImpl.h
```diff
@@ -156,6 +156,13 @@ struct Picture::Impl
         return true;
     }
 
+    bool preserveAspectRatio(bool* preserve) const
+    {
+        if (!loader) return false;
```
hermet, line 161: default value is true. (not false)

hermet (changes requested):
Please have a look my comments.
Thanks

src/lib/tvgPictureImpl.h
mgrudzinska: I know :) but  it's the value indicating success or failure of the api, it's not the info about the aspect ratio preservation

inc/thorvg.h
mgrudzinska: I've chosen the longer name, since this api returns only the info, whether the aspect ratio is preserved or not. It doesn't return the value of the preserveAspectRatio attribute.

In svg files the possible preserveAspectRatio values are 'xMinYMin meet', 'xMidYMid slice', ..., 'none' 
In tvg we support 'none' and only one of the rest - so we force uniform scaling, without going into details regarding xMin/Max etc. 

that's why I thought it may be confusing if the api is called 'aspectRatio', but I changed the name as you suggested

hermet: @mgrudzinska Hi, two reasons I suggested.
1. API naming consistency. You can check other getter apis.
2. Picture is the abstracted class. This API shouldn't be designed only for SVG.
3. Actually, I don't  know which scenario of this apis is useful so far (It would be better if you mention the *REAL*l user scenario where it's necessary) but In the end I think we might need the AspectRatio setter/getter options with an universal way.

src/lib/tvgPictureImpl.h
hermet: For consistency, c++ getter apis doesn't use return type but we prefer the immediate value return.
If you think it's a bit useful or necessary, instead we can put TVGERR() in that case with the useful info message.
for instance)
TVGERR("Picture", "Loader is invalid, Please load the file first....");

inc/thorvg.h
mgrudzinska: @hermet please take a look at #1117 
it's more general case, so if in the future we'd like to support other preserveAspectRatio attributes, than the api will be still valid. the version from this pr is very limited.
if #1117 is not what you meant by 'an universal way' please let me know, I'll try to change the prs accordingly.

The real use case for this api is, that I need it in efl :) otherwise it's probably as useful as the viewbox api is.

hermet: @mgrudzinska  why efl needs this?

mgrudzinska:
I solved the problem in efl without this api

PR #1115, opened 2021-12-15
sw_engine SwShape: Prevent null access

JSUYA:
strokeOutline returned strokeExportOutline() is the address of mpool->strokeOutline[idx].
Assuming this value is null, mpoolRetStrokeOutline on line 617 will access mpool->strokeOutline[idx].

Logically, the logic of the mpool* functions does not occur in this case.
However, in the case of the strokeOutline pointer of shapeGenStrokeRle,
static analyzer may occur a warning depending on the implementation of mpool.

### DIFF
--- a/src/lib/sw_engine/tvgSwShape.cpp
+++ b/src/lib/sw_engine/tvgSwShape.cpp
@@ -594,10 +594,6 @@ bool shapeGenStrokeRle(SwShape* shape, const Shape* sdata, const Matrix* transfo
     }
 
     strokeOutline = strokeExportOutline(shape->stroke, mpool, tid);
-    if (!strokeOutline) {
-        ret = false;
-        goto fail;
-    }
 
     if (!mathUpdateOutlineBBox(strokeOutline, clipRegion, renderRegion, false)) {
         ret = false;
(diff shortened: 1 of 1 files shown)


src/lib/sw_engine/tvgSwShape.cpp
```diff
@@ -614,7 +614,7 @@ bool shapeGenStrokeRle(SwShape* shape, const Shape* sdata, const Matrix* transfo
         if (shapeOutline->closed) free(shapeOutline->closed);
         free(shapeOutline);
     }
-    mpoolRetStrokeOutline(mpool, tid);
+    if (strokeOutline) mpoolRetStrokeOutline(mpool, tid);
```
hermet, line 617: I think we can remove the null-check exception at 597 line that was used for direct memory allocation before introducing memory pool... now null-checking is useless.

hermet (changes requested):
Good change but could you please check my comment?

src/lib/sw_engine/tvgSwShape.cpp
JSUYA: I know what your opinion means. but..
The case only the strokeOutline of shapeGenStrokeRle was considered, I thought that the static analyzer would warn that line 602 was used without null check.
umm..
I will update that remove nullcheck at 597 line.

hermet (approved):
thanks

PR #1117, opened 2021-12-16
common: aspectRatio api introduced

mgrudzinska:
The new api gets the information about the preserveAspectRatio
attribute of the loaded svg file. Since the preserveAspectRatio
contains two fields, two enums are returned:
AspectRatioAlign and AspectRatioMethod.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -148,6 +148,25 @@ enum class TVG_EXPORT CompositeMethod
 };
 
 /**
+ * @brief Enumeration indicating whether uniform scaling has to be forced and if so, how the graphic should be aligned.
+ */
+enum class TVG_EXPORT AspectRatioAlign
+{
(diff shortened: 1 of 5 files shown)


mgrudzinska:
@hermet I prepared this pr to replace #1114 with the more universal api. please let me know if it's not what you meant.

mgrudzinska:
@JSUYA please take a look first on the discussion in #1114

JSUYA (commented):
thank you for your work
Hermet also mentioned, preserveAspectRatio's getter is SVG specific.

preserveAspectRatio is a property that is determined inside the SVG file.
And do enough resize in SVGLoader(SvgSceneBuilder).
Therefore, even if the user gets AspectRatio, there is no need to use it unless resizing is required.

I think that there will be few cases of using it even if it is provided.

+
However, if we want to set the size of the picture and provide an image that stretches to fit the size, I think it can be provided as a setter for png and jpg.

mgrudzinska:
I solved the problem in efl without this api

Issue #1118, opened 2021-12-17
Crash when rasterizing svg image


JSUYA:
SVG file : [Godot.Planner.2021-11-06.zip](https://github.com/Samsung/thorvg/files/7731527/Godot.Planner.2021-11-06.zip)

After patch 06cef2a710f2bb27900eb509c4ae38a034c742ea, crash occurs when rasterizing image.
 

[BT]

```
SVG: //home/junsu/dev/os/thorvg/src/examples/images/Godot.Planner.2021-11-06.svg

Thread 1 "Svg" received signal SIGBUS, Bus error.
_interpDownScaler (img=<optimized out>, w=<optimized out>, rX=rX@entry=4294967247, rY=rY@entry=38, n=29, h=<optimized out>) at ../src/lib/sw_engine/tvgSwRaster.cpp:115
warning: Source file is more recent than executable.
115	            c[2] += (*p >> 8) & 0xff;
(gdb) bt
#0  _interpDownScaler (img=<optimized out>, w=<optimized out>, rX=rX@entry=4294967247, rY=rY@entry=38, n=29, h=<optimized out>) at ../src/lib/sw_engine/tvgSwRaster.cpp:115
#1  0x00007ffff7bc00e3 in _rasterScaledRGBAImage (halfScale=29, region=..., itransform=0x7fffffffcaac, image=0x555555da65a8, surface=0x555555dad4e0)
    at ../src/lib/sw_engine/tvgSwRaster.cpp:779
#2  _scaledRGBAImage (opacity=<optimized out>, region=..., transform=<optimized out>, image=0x555555da65a8, surface=0x555555dad4e0) at ../src/lib/sw_engine/tvgSwRaster.cpp:821
#3  _rasterRGBAImage (opacity=<optimized out>, region=..., transform=<optimized out>, image=0x555555da65a8, surface=0x555555dad4e0) at ../src/lib/sw_engine/tvgSwRaster.cpp:959
#4  rasterImage (surface=0x555555dad4e0, image=0x555555da65a8, transform=<optimized out>, bbox=..., opacity=<optimized out>) at ../src/lib/sw_engine/tvgSwRaster.cpp:1479
#5  0x00007ffff7bb4ba3 in tvg::Paint::Impl::render (this=0x7fffd413cf70, renderer=...) at ../src/lib/tvgPaint.cpp:178
[...]

PR #1119, opened 2021-12-17
sw_engine raster: ++safety of the scaled image rasterization

hermet:
Prevent out of buffer boundary access.

@Issues: https://github.com/Samsung/thorvg/issues/1118

### DIFF
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -82,8 +82,8 @@ static inline uint32_t _halfScale(float scale)
 //Bilinear Interpolation
 static uint32_t _interpUpScaler(const uint32_t *img, uint32_t w, uint32_t h, float sx, float sy)
 {
-    auto rx = static_cast<uint32_t>(sx);
-    auto ry = static_cast<uint32_t>(sy);
+    auto rx = (uint32_t)(sx);
+    auto ry = (uint32_t)(sy);
     auto rx2 = rx + 1;
     if (rx2 >= w) rx2 = w - 1;
     auto ry2 = ry + 1;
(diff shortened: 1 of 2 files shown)


hermet:
Need additional patch to fix the #1118

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -740,9 +748,11 @@ static bool _rasterScaledTranslucentRGBAImage(SwSurface* surface, const SwImage*
     if (image->scale < DOWN_SCALE_TOLERANCE) {
         for (auto y = region.min.y; y < region.max.y; ++y, dbuffer += surface->stride) {
             auto sy = static_cast<uint32_t>(y * itransform->e22 + itransform->e23);
+            if (sy >= image->h) continue;
             auto dst = dbuffer;
             for (auto x = region.min.x; x < region.max.x; ++x, ++dst) {
                 auto sx = static_cast<uint32_t>(x * itransform->e11 + itransform->e13);
```
JSUYA, line 754: I think the problem occurs when the result of (x * itransform->e11 + itransform->e13) is negative.
(or memory problem ?)
Wouldn't the result of this expression be checked first(0 > result || w < result ?)  and then cast?

JSUYA (commented):
please check my comment

src/lib/sw_engine/tvgSwRaster.cpp
hermet: we use "unsigned" so negative is not happening.

JSUYA: As far as I know there is a problem casting float to "unsigned" int.
please refer to https://embeddeduse.com/2013/08/25/casting-a-negative-float-to-an-unsigned-int/

hermet: @JSUYA  updated thanks.

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -392,27 +395,27 @@ static bool _rasterScaledTranslucentRleRGBAImage(SwSurface* surface, const SwIma
     //Center (Down-Scaled)
     if (image->scale < DOWN_SCALE_TOLERANCE) {
         for (uint32_t i = 0; i < image->rle->size; ++i, ++span) {
-            auto sy = static_cast<uint32_t>(span->y * itransform->e22 + itransform->e23);
+            auto sy = (uint32_t)(span->y * itransform->e22 + itransform->e23);
```
JSUYA, line 398: Wouldn't the logic of this function be changed like other functions?

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -429,42 +432,42 @@ static bool _rasterScaledRleRGBAImage(SwSurface* surface, const SwImage* image,
...
```
JSUYA, line 435: this function too.

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -660,13 +663,15 @@ static bool _rasterScaledMaskedTranslucentRGBAImage(SwSurface* surface, const Sw
     // Down-Scaled
     if (image->scale < DOWN_SCALE_TOLERANCE) {
         for (auto y = region.min.y; y < region.max.y; ++y) {
+            auto sy = (uint32_t)(y * itransform->e22 + itransform->e23);
```
JSUYA, line 666: this function too

JSUYA (commented):
Looks like some functions haven't been fixed yet. could you please check again?

JSUYA (approved):

PR #1122, opened 2021-12-18
Hide the thorvg LZW implementation.

fire:
Use an anonymous namespace so that I can compile on Godot Engine.

Godot uses Dictionary() too.


hermet (approved):
thanks!

PR #1123, opened 2021-12-21
example: add script to execute all

kimcinoo:
This patch will save your time to check example.

  ./all.sh [interval time between examples, default is 1 second]

### DIFF
--- a/src/examples/all.sh
+++ b/src/examples/all.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+RED='\033[31m'
+GREEN='\033[32m'
+NC='\033[0m'
+
+INTERVAL=${1:-1}
+EXAMPLES=`find . -executable -type f | sort | uniq`
+for EXAMPLE in $EXAMPLES
+do
+	if [[ $EXAMPLE == *.sh ]]; then
+		continue
+	fi
+
+	echo -e "Execute: "${GREEN}$EXAMPLE${NC}" for "$INTERVAL" second(s)"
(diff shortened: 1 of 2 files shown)


hermet (changes requested):
@kimcinoo Great, Thanks for your contribution but I tested, nothing executed...

It looks following script should be fixed because executables won't have "executable" field name In my laptop.

for example, checking "TvgSaver" with your script.

EXAMPLES=`ls ./ | xargs -IFILE file FILE | grep executable`

TvgSaver: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=821db9f4e4684d284a06b1284df8b9ad7497b850, for GNU/Linux 3.2.0, with debug_info, not stripped

Probably this can be a better solution?
`"find <dir> -executable -type f"`

kimcinoo:
Oops, for me it gives
```
./TvgSaver: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=a2ea70e09cf7c337521cd6553b1fd91412610cba, not stripped
```
It seems that the previous commands depend on how to build the TvgSaver.
Your suggestion is neat and general. Very nice!

hermet (approved):
excellent! thanks!

Issue #619, opened 2021-07-16
svg: Path that works normally doesn't work as mask


mmaciola:
```
<svg width="1000" height="1000" viewBox="0 0 1000 1000" xmlns="http://www.w3.org/2000/svg">
	<g fill="red" mask="url(#vg)">
		<rect x="0" y="0" width="1000" height="1000" />
	</g>
	<defs>
		<mask id="vg">
			<rect fill="white" x="0" y="0" width="1000" height="1000" />
			<path fill="black" d="m280 140h15v55l8 10 8-10v-55h15v60l-23 25-23-25z"/>
			<path fill="black" d="m335 140v80h45v-50h-25v10h10v30h-15v-57h18v-13z"/>
		</mask>
	</defs>
</svg>
```
Should be:
![logo_test](https://user-images.githubusercontent.com/71131832/125918690-43cbe46c-2880-4ff5-a9e3-93517dff9488.png)
Is:
![Screenshot from 2021-07-16 10-36-56](https://user-images.githubusercontent.com/71131832/125918755-a729ff0a-80a8-4c04-a0d2-917574336cf2.png)

mgrudzinska:
on the current version of TVG, after #620 is merged it should be solved

mgrudzinska:
@hermet all we need is pr #620 - @mmaciola is the author. there are some changes that have to be done in this pr.
Is that an urgent issue? 
as far as I know Michal will come back to work after new year

mmaciola:
@hermet @mgrudzinska I am officially off till the end of December, but I will be near the computer anyway.
I have just updated the PR #620 and #621

mgrudzinska:
#1124 is also necessary

hermet:
Confirmed it works now :)

PR #1124, opened 2021-12-21
svg_loader: LumaMask used

mgrudzinska:
For masks with colors other than white the ColorMask should be used
instead of the AlphaMask. For the perf reasons the ColorMask is not used
when it's not necessary.

issue #619
 
it builds after #620 is merged

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -67,7 +67,7 @@ struct Box
 
 
 static bool _appendShape(SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath);
-static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox, const string& svgPath, bool mask);
+static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox, const string& svgPath, bool mask, bool* isMaskWhite = nullptr);
 
 
 static inline bool _isGroupType(SvgNodeType type)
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -575,15 +576,22 @@ static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox,
...
                     if ((*child)->type == SvgNodeType::Use)
-                        scene->push(_useBuildHelper(*child, vBox, svgPath));
+                        scene->push(_useBuildHelper(*child, vBox, svgPath, compMethod));
                     else
-                        scene->push(_sceneBuildHelper(*child, vBox, svgPath, false));
+                        scene->push(_sceneBuildHelper(*child, vBox, svgPath, false, compMethod));
                 } else if ((*child)->type == SvgNodeType::Image) {
                     auto image = _imageBuildHelper(*child, vBox, svgPath);
                     if (image) scene->push(move(image));
                 } else if ((*child)->type != SvgNodeType::Mask) {
```
hermet, line 585: It looks it would be better to have Luma or Alpha Mask types?

JSUYA (approved):
LGTM

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: @hermet I was sure, that I've made a comment here...
please check #1130, but nevertheless I think, because of the performance reasons, It's a good idea to check, whether the mask has a color other than white. if no, there is no reason to use a LumaMask. 

and I don't think, that we can check this earlier... when we parse the file, but then the goal is to parse it, not to check some conditions.

I'll merge a bit later my PRs that are already accepted and rebase this one and ask you to take a look one more time

hermet (approved):

hermet:
@mgrudzinska sorry for conflict, could you please rebase and then apply it?

mgrudzinska:
@hermet not a problem, done :)
I've changed the arg type from CompositeMethod to bool, I think it's easier to read this way. everything else is as it was.

PR #1128, opened 2021-12-31
svg_loader: composite node splitted on mask and clip nodes

mgrudzinska:
The SvgCompositeNode is replaced by the SvgMaskNode and SvgClipNode.
This is needed for using Luma/AlphaMask in the svg loader and in the future,
when we introduce other mask's features.


mgrudzinska:
@junsu suggested such a solution a while ago (https://github.com/Samsung/thorvg/pull/992#pullrequestreview-798436217)... sorry for not applying it back then :)

hermet (approved):

PR #1129, opened 2021-12-31
svg_loader: postponed nodes properly cloned

mgrudzinska:
In a case when no defs tag was used, no child node was find.
Also the node attributes should not override those set in the ancestor.

before:
02_before

after:
02_after

code:
```
<svg viewBox="0 0 30 30" xmlns="http://www.w3.org/2000/svg">

<use href="#myCircle1" x="10" fill="blue"/>
<use href="#myCircle1" x="20" fill="white" stroke="red"/>

<use href="#myCircle2" x="10" y="10" fill="blue"/>
<use href="#myCircle2" x="20" y="10" fill="white" stroke="red"/>

<circle id="myCircle1" cx="5" cy="5" r="4" stroke="blue"/>

<defs>
<circle id="myCircle2" cx="5" cy="5" r="2" stroke="green"/>
</defs>

</svg>
```

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1745,6 +1745,112 @@ error_grad_alloc:
 }
 
 
+static void _styleInherit(SvgStyleProperty* child, const SvgStyleProperty* parent)
+{
+    if (parent == nullptr) return;
+    //Inherit the property of parent if not present in child.
+    if (!child->curColorSet) {
+        child->color = parent->color;
+        child->curColorSet = parent->curColorSet;
+    }
+    //Fill
+    if (!((int)child->fill.flags & (int)SvgFillFlags::Paint)) {
+        child->fill.paint.color = parent->fill.paint.color;
+        child->fill.paint.none = parent->fill.paint.none;
+        child->fill.paint.curColor = parent->fill.paint.curColor;
+        if (parent->fill.paint.url) child->fill.paint.url = _copyId(parent->fill.paint.url);
+    }
+    if (!((int)child->fill.flags & (int)SvgFillFlags::Opacity)) {
+        child->fill.opacity = parent->fill.opacity;
+    }
+    if (!((int)child->fill.flags & (int)SvgFillFlags::FillRule)) {
+        child->fill.fillRule = parent->fill.fillRule;
+    }
+    //Stroke
+    if (!((int)child->stroke.flags & (int)SvgStrokeFlags::Paint)) {
+        child->stroke.paint.color = parent->stroke.paint.color;
+        child->stroke.paint.none = parent->stroke.paint.none;
+        child->stroke.paint.curColor = parent->stroke.paint.curColor;
+        child->stroke.paint.url = parent->stroke.paint.url ? _copyId(parent->stroke.paint.url) : nullptr;
+    }
+    if (!((int)child->stroke.flags & (int)SvgStrokeFlags::Opacity)) {
+        child->stroke.opacity = parent->stroke.opacity;
+    }
+    if (!((int)child->stroke.flags & (int)SvgStrokeFlags::Width)) {
+        child->stroke.width = parent->stroke.width;
+    }
+    if (!((int)child->stroke.flags & (int)SvgStrokeFlags::Dash)) {
+        if (parent->stroke.dash.array.count > 0) {
+            child->stroke.dash.array.clear();
+            child->stroke.dash.array.reserve(parent->stroke.dash.array.count);
+            for (uint32_t i = 0; i < parent->stroke.dash.array.count; ++i) {
+                child->stroke.dash.array.push(parent->stroke.dash.array.data[i]);
+            }
+        }
+    }
+    if (!((int)child->stroke.flags & (int)SvgStrokeFlags::Cap)) {
+        child->stroke.cap = parent->stroke.cap;
+    }
+    if (!((int)child->stroke.flags & (int)SvgStrokeFlags::Join)) {
+        child->stroke.join = parent->stroke.join;
+    }
+}
+
+
+static void _styleCopy(SvgStyleProperty* to, const SvgStyleProperty* from)
+{
+    if (from == nullptr) return;
+    //Copy the properties of 'from' only if they were explicitly set (not the default ones).
+    if (from->curColorSet) {
+        to->color = from->color;
+        to->curColorSet = true;
+    }
+    //Fill
+    to->fill.flags = (SvgFillFlags)((int)to->fill.flags | (int)from->fill.flags);
+    if (((int)from->fill.flags & (int)SvgFillFlags::Paint)) {
+        to->fill.paint.color = from->fill.paint.color;
+        to->fill.paint.none = from->fill.paint.none;
+        to->fill.paint.curColor = from->fill.paint.curColor;
+        if (from->fill.paint.url) to->fill.paint.url = _copyId(from->fill.paint.url);
+    }
+    if (((int)from->fill.flags & (int)SvgFillFlags::Opacity)) {
+        to->fill.opacity = from->fill.opacity;
+    }
+    if (((int)from->fill.flags & (int)SvgFillFlags::FillRule)) {
+        to->fill.fillRule = from->fill.fillRule;
+    }
+    //Stroke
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1727,6 +1727,110 @@ static SvgStyleGradient* _cloneGradient(SvgStyleGradient* from)
 }
 
 
+static void _styleInherit(SvgStyleProperty* child, const SvgStyleProperty* parent)
```
mgrudzinska, line 1730: I moved this func from below, without any changes.

Of course, I can unify the `_styleInherit` and the `_styleCopy` functions, by adding a bool as an additional arg:
```
static void _styleInherit(SvgStyleProperty* style1, const SvgStyleProperty* style2, bool b)
```
with the comparisons inside looking like:
```
if (((int)style1->fill.flags & (int)SvgFillFlags::Paint) == b) {
```

but please notice how hard it became to read and understand. 
If `b` is `false` then this comparison behaves as in  `_styleInherit`. 
if `b` is `true` then it's like  `_styleCopy`, but I have to call it `_styleCopy(from->style, to->style);` (instead of `_styleCopy(to->style, from->style)`), that's why I'd have to change the args names parent/child or from/to to style1/style2.

JSUYA: I know there is a lot of duplicate code. As you said, the code(`bool b...`) you presented is difficult to understand and complex.
Because inherit and copy have different if conditions, I think it's okay to have separate method.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1727,6 +1727,110 @@ static SvgStyleGradient* _cloneGradient(SvgStyleGradient* from)
...
+{
+    if (from == nullptr) return;
+    //Copy the properties of 'from' only if they were explicitly set (not the default ones).
+    //Fill
+    if (((int)from->fill.flags & (int)SvgFillFlags::Paint)) {
+        to->fill.paint.color = from->fill.paint.color;
+        to->fill.paint.none = from->fill.paint.none;
+        to->fill.paint.curColor = from->fill.paint.curColor;
+        if (from->fill.paint.url) to->fill.paint.url = _copyId(from->fill.paint.url);
+    } else if (to->fill.paint.curColor && !to->curColorSet) {
```
JSUYA, line 1792: Is this conditional correct? Shouldn't you check the color value of from?

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1727,6 +1727,110 @@ static SvgStyleGradient* _cloneGradient(SvgStyleGradient* from)
...
+    if (((int)from->fill.flags & (int)SvgFillFlags::FillRule)) {
+        to->fill.fillRule = from->fill.fillRule;
+    }
+    //Stroke
+    if (((int)from->stroke.flags & (int)SvgStrokeFlags::Paint)) {
+        to->stroke.paint.color = from->stroke.paint.color;
+        to->stroke.paint.none = from->stroke.paint.none;
+        to->stroke.paint.curColor = from->stroke.paint.curColor;
+        to->stroke.paint.url = from->stroke.paint.url ? _copyId(from->stroke.paint.url) : nullptr;
+    } else if (to->stroke.paint.curColor && !to->curColorSet) {
```
JSUYA, line 1807: Same as 1792 line.

JSUYA (approved):
Please check my comment.
If there is no problem, please merge. I agree with this patch.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1727,6 +1727,112 @@ static SvgStyleGradient* _cloneGradient(SvgStyleGradient* from)
...
+static void _styleCopy(SvgStyleProperty* to, const SvgStyleProperty* from)
+{
+    if (from == nullptr) return;
+    //Copy the properties of 'from' only if they were explicitly set (not the default ones).
+    if (from->curColorSet) {
+        to->color = from->color;
+        to->curColorSet = true;
+    }
+    //Fill
+    to->fill.flags = (SvgFillFlags)((int)to->fill.flags | (int)from->fill.flags);
```
mgrudzinska, line 1791: flags has to be copied as well

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1727,6 +1727,112 @@ static SvgStyleGradient* _cloneGradient(SvgStyleGradient* from)
...
+        child->stroke.join = parent->stroke.join;
+    }
+}
+
+
+static void _styleCopy(SvgStyleProperty* to, const SvgStyleProperty* from)
+{
+    if (from == nullptr) return;
+    //Copy the properties of 'from' only if they were explicitly set (not the default ones).
+    if (from->curColorSet) {
```
mgrudzinska, line 1786: @JSUYA thanks for your comment. I missed this part.
I removed the code `else if (to->fill.paint.curColor && !to->curColorSet)` (for a stroke as well) and I added this `if` block.

I think the `style->color` value should be copied if it was set (so if `curColorSet` is true) regardless of the `curColor` value.

we have two variables related to the current color attribute:
- curColorSet - which is `true` if the color value was set `color="red"`
- curColor - it's `true` if we set `fill="currentColor"` 

As far as I understand, the `curColor` is not very useful for now. I don't want to remove the 'curColor' for now, since it may be useful later, when we take care of cases with multiple `fill` attribute set for a given element. I need to read specs first, to know how to handle such cases - or of course someone else can fix it

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1727,6 +1727,112 @@ static SvgStyleGradient* _cloneGradient(SvgStyleGradient* from)
...
+        if (from->fill.paint.url) to->fill.paint.url = _copyId(from->fill.paint.url);
+    }
+    if (((int)from->fill.flags & (int)SvgFillFlags::Opacity)) {
+        to->fill.opacity = from->fill.opacity;
+    }
+    if (((int)from->fill.flags & (int)SvgFillFlags::FillRule)) {
+        to->fill.fillRule = from->fill.fillRule;
+    }
+    //Stroke
+    to->stroke.flags = (SvgStrokeFlags)((int)to->stroke.flags | (int)from->stroke.flags);
```
mgrudzinska, line 1805: flags has to be copied as well

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1735,7 +1841,8 @@ static void _copyAttr(SvgNode* to, const SvgNode* from)
         if (to->transform) *to->transform = *from->transform;
     }
     //Copy style attribute
-    *to->style = *from->style;
+    _styleCopy(to->style, from->style);
+    to->style->flags = (SvgStyleFlags)((int)to->style->flags | (int)from->style->flags);
```
mgrudzinska, line 1845: flags has to be copied as well

mgrudzinska:
@JSUYA after your comment I realized, that I had to copy not only the attributes values, but the flags as well. 
I also changed the code related to the currentColor

Since these may be not very obvious changes, I'd like to ask you to take a look one more time - sorry for that.
I marked all the lines that I've changed, since I forced the push...

mgrudzinska:
to test the changes I used the svg: (but after #1136 and #1137 are merged)
```
<svg viewBox="0 0 30 30" xmlns="http://www.w3.org/2000/svg">
<use href="#myCircle1" x="10" fill="red"/>
<use href="#myCircle2" x="10" fill="red"/>

<defs>
<linearGradient id="linearGradient1" x1="0" y1="0" x2="0.2" y2="0.2" spreadMethod="reflect">
<stop style="stop-color:#ff0000;stop-opacity:1;" offset="0"/>
<stop style="stop-color:#0000ff;stop-opacity:1;" offset="1"/>
</linearGradient>

<g id="myCircle1" fill="url(#linearGradient1)" color="green">
<circle cx="5" cy="5" r="4" stroke="currentColor"/>
</g>
</defs>
<g id="myCircle2" stroke="url(#linearGradient1)" color="green">
<g color="red">
<circle cx="5" cy="10" r="4" fill="currentColor"/>
</g>
</g>

<use href="#myCircle1" fill="red"/>
<use href="#myCircle2" x="15" fill="red"/>
</svg>
```

mgrudzinska:
just rebasing

JSUYA (approved):
I checked.
LGTM :)

PR #1132, opened 2022-01-05
sw_engine SwRaster: Initialize uninitialized transform value

JSUYA:


src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -499,6 +499,7 @@ static bool _rasterScaledRleRGBAImage(SwSurface* surface, const SwImage* image,
 static bool _scaledRleRGBAImage(SwSurface* surface, const SwImage* image, const Matrix* transform, const SwBBox& region, uint32_t opacity)
 {
     Matrix itransform;
+    mathIdentity(&itransform);
     if (transform && !mathInverse(transform, &itransform)) return false;
```
hermet, line 503: Actually, that is intended code when @p transform is nullptr but for the maintenance we can apply it.

Just one thing.

 if (transform && !mathInverse(transform, &itransform)) return false;
else mathIdentify(&itransform);

hermet (commented):
Please check a comment.

src/lib/sw_engine/tvgSwRaster.cpp
JSUYA: Thank you for review. I modified code.

hermet (approved):

PR #1142, opened 2022-01-09
svg_loader: css style node introduced

mgrudzinska:
For now it is assumed that only one style element is defined
in an svg file, although it can be easily changed if needed.
The style node will be used to define the style applied to a node
of a given type or in a case when a class attrib was used.

### DIFF
--- a/src/loaders/svg/tvgSvgLoaderCommon.h
+++ b/src/loaders/svg/tvgSvgLoaderCommon.h
@@ -51,6 +51,7 @@ enum class SvgNodeType
     Video,
     ClipPath,
     Mask,
+    CssStyle,
     Unknown
 };
 
@@ -233,6 +234,10 @@ struct SvgMaskNode
     bool userSpace;
 };
 
+struct SvgCssStyleNode
+{
+};
+
 struct SvgLinearGradient
 {
     float x1;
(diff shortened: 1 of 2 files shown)


src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -233,6 +234,10 @@ struct SvgMaskNode
     bool userSpace;
 };
 
+struct SvgCssStyleNode
+{
```
mgrudzinska, line 238: empty for now, attribs like `title`, `type` can be introduced, although I have a problem with finding out how to use the `title` attrib. that's why I assume for now, that only one style element is used in the svg file...

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -401,6 +407,7 @@ struct SvgLoaderData
     Array<SvgNode *> stack = {nullptr, 0, 0};
     SvgNode* doc = nullptr;
     SvgNode* def = nullptr;
+    SvgNode* cssStyle = nullptr;
```
mgrudzinska, line 410: if there will be more than one style element in the file, I'd like to use this var to know which one was the last one - like for gradients.

mgrudzinska:
will be present in a cumulative css pr

PR #1143, opened 2022-01-09
svg_loader: class attribute added

mgrudzinska:
It's the first step in introducing the css inline style sheets into tvg.
For now the class attribute is set only for SvgNodes (not for grads),
it's not used yet.


mgrudzinska:
will be present in a cumulative css pr

PR #1144, opened 2022-01-09
svg_loader: parsing of the css internal style sheets implemented

mgrudzinska:
The CData parser and the CSS style parser implementations added.
Both enables to read the css style from an svg file in a case, when
CDATA marker was used (it's not obligatory, so the future patches
should enables to read the scc style even if the marker was not used).


mgrudzinska:
I'll push all changes in a one cumulative css pr

PR #1159, opened 2022-01-18
build: Add missing <cstring> includes for MinGW compatibility

akien-mga:
Fixes downstream issue in Godot:
- https://github.com/godotengine/godot/issues/56894

This could also go in `lib/tvgCommon.h` or `include/thorvg.h` to be available everywhere IINM. For now I chose to put it where it's used like for other standard library headers like `math.h`, but that means other missing includes might be introduced if your pipelines don't test against MinGW (as apparently it's fine without for MSVC).

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -51,6 +51,7 @@
 
 #define _USE_MATH_DEFINES       //Math Constants are not defined in Standard C/C++.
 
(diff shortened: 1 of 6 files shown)


hermet (approved):

hermet:
@akien-mga thanks, could you please add yourself here as well? 
[AUTHORS](https://github.com/Samsung/thorvg/blob/master/AUTHORS)

Issue #702, opened 2021-08-09
Support CSS Style in SVG?


hermet:
Still there are demands of this....

mgrudzinska:
I'll start with inline start sheets. I think that the external ones are not so demanded for now (?)

mgrudzinska:
#1142 
#1143 
#1144
I want to introduce the very first version, so files like this:
```
<svg viewBox="0 0 300 300">
<defs>
    <style>
    <![CDATA[
        circle.myCircle {
            fill: red;
            stroke: blue;
        }
        circle {
            fill: green;
            stroke-width: 10;
            stroke: yellow
        }
        svg {
             fill: pink;
        }
    ]]>
</style>
</defs>

<circle cx="100" cy="50" r="50" class="myCircle"/>
<circle cx="100" cy="100" r="50" class="myCircle" style="fill:gray;"/>
<circle cx="100" cy="150" r="50" />
<rect x="0" y="0" width="40" height="40"/>
</svg>
```
will be properly drawn.

assumptions:
- only one class name used as a reference style node
- only one style element in the svg file (easy to change, but I don't know how to use 'title' attrib...)
- CDATA marker is used

in the near future:
- set a style using id
- parse even if CDATA marker was not used

JSUYA:
Wow good :)
I think we can verify this using the e-style-00*.svg sample in the SVG.zip sample discussed here.
(https://github.com/godotengine/godot/pull/49601)
Even if we don't satisfy all the svg files, it will help you set our goals.

I've read your 3 patches. I expect there will create a lot of patches.
How about implementing with one Draft PR? And you can split them into whatever size you think is appropriate or review them all at once, even if it takes time.

mgrudzinska:
#1160 
I'll check this pr one more time and make some comments to make it easier to review

DONE:
- parsing targets in formats: tag, tag.name, .name
- with/without CDATA marker
- works for style defined on the beginning/end of the file
- proper precedence applied

TODO:
- gradient support
- clipath andmasks support
- introduce global targets like svg, g, ...
- other targets formats, like id
- check in the svg2 standard: geometric properties, how to use title (for now only one <style> per svg file) 



```
<svg  viewBox="0 0 300 300">
 <rect x="0" y="0" width="50" height="50" class="moj" />
 <circle cx="50" cy="50" r="20"  stroke="black" stroke-width="3" class="moj" />
 
 <rect x="100" y="0" width="50" height="50" class="moj" style="fill:red"/>
[...]

PR #1160, opened 2022-01-18
introducing css internal style sheets

mgrudzinska:
svg_loader: a basic implementation of the css internal style sheets

Support for class and type selectors (for the graphics elements) introduced.
The geometric attributes are not supported in the style attribute (feature added in SVG2 standard).

issue #702 
https://github.com/Samsung/thorvg/wiki/CSS-styling

---------
summary for a reviewer:
- tvgXmlParser.cpp - simpleXmlParseCSSAttribute() - parsing targets in formats: tag {}, tag.name {}, .name {}
- all methods _create...Node() has one additional arg - a pointer to a function. till now the simpleXmlParseAttributes() was used inside _create...Node() to determine the attributes values. Since the same functions (_create...Node()) are used to create the nodes with css style and the format of the data is different, I need to use there simpleXmlParseW3CAttribute() instead.
- proper precedence (from the highest):
  1. inline style (`style="fill: red; stroke: blue"`)
  2. css style for a proper target and with a name (`<circle ... class="name" />`)
  3. css style for all targets, with a name (class attribute used, but the type of the node is not checked) 
  4. css style for a proper target (`<circle .../>`   <- no class used, style applied to all circles)
  5. attributes (`fill="red" stroke="green"`)
- to apply the style defined at the end of the svg file, the postponed mechanism applied (similar to the one used for the 'use' nodes)

+ comments in issue #702

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -798,7 +798,7 @@ static bool _attrParseSvgNode(void* data, const char* key, const char* value)
     } else if (!strcmp(key, "preserveAspectRatio")) {
         if (!strcmp(value, "none")) doc->preserveAspect = false;
     } else if (!strcmp(key, "style")) {
-        return simpleXmlParseW3CAttribute(value, 0, _parseStyleAttr, loader);
+        return simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
     }
 #ifdef THORVG_LOG_ENABLED
     else if ((!strcmp(key, "x") || !strcmp(key, "y")) && fabsf(svgUtilStrtof(value, nullptr)) > FLT_EPSILON) {
@@ -951,13 +951,13 @@ static void _handleDisplayAttr(TVG_UNUSED SvgLoaderData* loader, SvgNode* node, 
 }
 
 
-static SvgNode* _findCssStyleNode(const SvgNode* cssStyle, const char* title)
+static SvgNode* _findCssStyleNode(const SvgNode* cssStyle, const char* title, SvgNodeType type)
 {
     if (!cssStyle) return nullptr;
 
     auto child = cssStyle->child.data;
     for (uint32_t i = 0; i < cssStyle->child.count; ++i, ++child) {
-        if (((*child)->id) && !strcmp((*child)->id, title)) return (*child);
+        if ((*child)->type == type && ((*child)->id) && !strcmp((*child)->id, title)) return (*child);
     }
     return nullptr;
 
@@ -973,8 +973,8 @@ static void _handleCssClassAttr(SvgLoaderData* loader, SvgNode* node, const char
     *cssClass = _copyId(value);
 
     //TODO: works only if style was defined before it is used
-    if (auto cssNode = _findCssStyleNode(loader->cssStyle, *cssClass)) {
-        //TODO: check SVG2 stndard - should the geometric properties be copied?
+    if (auto cssNode = _findCssStyleNode(loader->cssStyle, *cssClass, node->type)) {
+        //TODO: check SVG2 standard - should the geometric properties be copied?
         _copyAttr(node, cssNode, false);
     }
 }
@@ -1054,7 +1054,7 @@ static bool _attrParseGNode(void* data, const char* key, const char* value)
     SvgNode* node = loader->svgParse->node;
 
     if (!strcmp(key, "style")) {
-        return simpleXmlParseW3CAttribute(value, 0, _parseStyleAttr, loader);
+        return simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
     } else if (!strcmp(key, "transform")) {
         node->transform = _parseTransformationMatrix(value);
     } else if (!strcmp(key, "id")) {
@@ -1083,7 +1083,7 @@ static bool _attrParseClipPathNode(void* data, const char* key, const char* valu
     SvgClipNode* clip = &(node->node.clip);
 
     if (!strcmp(key, "style")) {
-        return simpleXmlParseW3CAttribute(value, 0, _parseStyleAttr, loader);
+        return simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
     } else if (!strcmp(key, "transform")) {
         node->transform = _parseTransformationMatrix(value);
     } else if (!strcmp(key, "id")) {
@@ -1107,7 +1107,7 @@ static bool _attrParseMaskNode(void* data, const char* key, const char* value)
     SvgMaskNode* mask = &(node->node.mask);
 
     if (!strcmp(key, "style")) {
-        return simpleXmlParseW3CAttribute(value, 0, _parseStyleAttr, loader);
+        return simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
     } else if (!strcmp(key, "transform")) {
         node->transform = _parseTransformationMatrix(value);
     } else if (!strcmp(key, "id")) {
@@ -1286,7 +1286,7 @@ static bool _attrParsePathNode(void* data, const char* key, const char* value)
         //Temporary: need to copy
         path->path = _copyId(value);
     } else if (!strcmp(key, "style")) {
-        return simpleXmlParseW3CAttribute(value, 0, _parseStyleAttr, loader);
+        return simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
     } else if (!strcmp(key, "clip-path")) {
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
@@ -1348,7 +1348,7 @@ static bool _attrParseCircleNode(void* data, const char* key, const char* value)
     }
 
     if (!strcmp(key, "style")) {
-        return simpleXmlParseW3CAttribute(value, 0, _parseStyleAttr, loader);
+        return simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
     } else if (!strcmp(key, "clip-path")) {
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
@@ -1415,7 +1415,7 @@ static bool _attrParseEllipseNode(void* data, const char* key, const char* value
     } else if (!strcmp(key, "class")) {
         _handleCssClassAttr(loader, node, value);
     } else if (!strcmp(key, "style")) {
-        return simpleXmlParseW3CAttribute(value, 0, _parseStyleAttr, loader);
+        return simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
     } else if (!strcmp(key, "clip-path")) {
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
@@ -1489,7 +1489,7 @@ static bool _attrParsePolygonNode(void* data, const char* key, const char* value
     if (!strcmp(key, "points")) {
         return _attrParsePolygonPoints(value, &polygon->points, &polygon->pointsCount);
     } else if (!strcmp(key, "style")) {
-        return simpleXmlParseW3CAttribute(value, 0, _parseStyleAttr, loader);
+        return simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
     } else if (!strcmp(key, "clip-path")) {
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
@@ -1576,7 +1576,7 @@ static bool _attrParseRectNode(void* data, const char* key, const char* value)
     } else if (!strcmp(key, "class")) {
         _handleCssClassAttr(loader, node, value);
     } else if (!strcmp(key, "style")) {
-        ret = simpleXmlParseW3CAttribute(value, 0, _parseStyleAttr, loader);
+        ret = simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
     } else if (!strcmp(key, "clip-path")) {
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
@@ -1641,7 +1641,7 @@ static bool _attrParseLineNode(void* data, const char* key, const char* value)
     } else if (!strcmp(key, "class")) {
         _handleCssClassAttr(loader, node, value);
     } else if (!strcmp(key, "style")) {
-        return simpleXmlParseW3CAttribute(value, 0, _parseStyleAttr, loader);
+        return simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
     } else if (!strcmp(key, "clip-path")) {
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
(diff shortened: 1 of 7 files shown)


src/loaders/svg/tvgXmlParser.cpp
```diff
@@ -467,6 +467,12 @@ bool simpleXmlParseW3CAttribute(const char* buf, simpleXMLAttributeCb func, cons
     do {
         char* sep = (char*)strchr(buf, ':');
         next = (char*)strchr(buf, ';');
+        if (sep >= end)
+        {
+            next = nullptr;
+            sep = nullptr;
+        }
+        if (next >= end) next = nullptr;
```
mgrudzinska, line 475: till now this func got data like:
`fill: red; stroke: green`
since I'd like to use the same func to read the css style data like 
`fill: red; stroke: green }
circle.red { fill: red; }
I have to explicitly check, whether I didn't go outside the scope of one style (circle.red defines a separate style)

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -401,12 +408,15 @@ struct SvgLoaderData
...
     SvgNode* def = nullptr;
+    SvgNode* cssStyle = nullptr;
     Array<SvgStyleGradient*> gradients;
     SvgStyleGradient* latestGradient = nullptr; //For stops
     SvgParser* svgParse = nullptr;
     Array<SvgNodeIdPair> cloneNodes;
+    Array<SvgNodeIdPair> cloneCssStyleNodes;
     int level = 0;
     bool result = false;
+    bool style = false; //TODO: find a better sollution?
```
mgrudzinska, line 419: when parsing data or cdata, I need to know what was the previous tag. if it was 'style' I can assume that the data describe the style.  otherwise tvg ignores it.

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -401,12 +408,15 @@ struct SvgLoaderData
     Array<SvgNode *> stack = {nullptr, 0, 0};
     SvgNode* doc = nullptr;
     SvgNode* def = nullptr;
+    SvgNode* cssStyle = nullptr;
```
mgrudzinska, line 411: the parent node with styles - for now only one is allowed. for two <style> tags in one svg file the styles won;t be correctly applied. I can change it and use an array, but I don't know how to use title tag, to switch between styles...

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2750,13 +2993,47 @@ static void _updateComposite(SvgNode* node, SvgNode* root)
 }
 
 
+static void _updateCssStyle(SvgNode* doc, SvgNode* cssStyle)
```
mgrudzinska, line 2996: the last update - all nodes has to be checked, not only the ones with class attrib (e.x. style set for all circles will be applied here).

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2586,6 +2788,44 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content,
...
+    const char* attrs = nullptr;
+    unsigned int attrsLength = 0;
+
+    FactoryMethod method;
+    GradientFactoryMethod gradientMethod;
+    SvgNode *node = nullptr;
+
+    while (auto next = simpleXmlParseCSSAttribute(content, length, &tag, &name, &attrs, &attrsLength)) {
+        if ((method = _findGroupFactory(tag))) {
+            //TODO - node->id ??? add additional var for svgnode?
```
mgrudzinska, line 2804: this 'TODO' I'll check when I'll apply support for other target formats

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1904,17 +2100,15 @@ static void _copyAttr(SvgNode* to, const SvgNode* from)
...
+            to->node.polygon.points = (float*)malloc(to->node.polygon.pointsCount * sizeof(float));
+            memcpy(to->node.polygon.points, from->node.polygon.points, to->node.polygon.pointsCount * sizeof(float));
             break;
         }
         case SvgNodeType::Polyline: {
-            if ((to->node.polyline.pointsCount = from->node.polyline.pointsCount)) {
-                to->node.polyline.points = (float*)malloc(to->node.polyline.pointsCount * sizeof(float));
-                memcpy(to->node.polyline.points, from->node.polyline.points, to->node.polyline.pointsCount * sizeof(float));
-            }
+            to->node.polyline.pointsCount = from->node.polyline.pointsCount;
```
mgrudzinska, line 2109: regression - need to change it !!

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1904,17 +2100,15 @@ static void _copyAttr(SvgNode* to, const SvgNode* from)
             break;
         }
         case SvgNodeType::Polygon: {
-            if ((to->node.polygon.pointsCount = from->node.polygon.pointsCount)) {
-                to->node.polygon.points = (float*)malloc(to->node.polygon.pointsCount * sizeof(float));
-                memcpy(to->node.polygon.points, from->node.polygon.points, to->node.polygon.pointsCount * sizeof(float));
-            }
+            to->node.polygon.pointsCount = from->node.polygon.pointsCount;
```
mgrudzinska, line 2103: regression - need to change it !!

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -946,6 +948,154 @@ static void _handleDisplayAttr(TVG_UNUSED SvgLoaderData* loader, SvgNode* node,
...
+    for (uint32_t i = 0; i < cssStyle->child.count; ++i, ++child) {
+        if ((*child)->type == SvgNodeType::CssStyle) {
+            if ((title && (*child)->id && !strcmp((*child)->id, title))) return (*child);
+        }
+    }
+    return nullptr;
+}
+
+
+static void _cssStyleCopy(SvgStyleProperty* to, const SvgStyleProperty* from)
```
mgrudzinska, line 979: the value is copied only if it wasn't set using style attribute (style="fill:red").
the proper style flag is set, but since the style attribute has the highest priority, it will be able to overwrite the values.

mgrudzinska:
@hermet @JSUYA 
I'll solve the windows issue and add a general description so we can merge it as one commit, but the rest is ready. I tried to describe what and why I've changed. 
I'd appreciate any comment or question.

JSUYA (commented):
Overall I think it's good.
I leave a tiny comment, please check it
+ please add an example to test this CSS style

src/loaders/svg/tvgXmlParser.cpp
```diff
@@ -449,7 +457,7 @@ bool simpleXmlParse(const char* buf, unsigned bufLength, bool strip, simpleXMLCb
 }
 
 
-bool simpleXmlParseW3CAttribute(const char* buf, simpleXMLAttributeCb func, const void* data)
+bool simpleXmlParseW3CAttribute(const char* buf, unsigned buflen, simpleXMLAttributeCb func, const void* data)
```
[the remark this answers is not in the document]
mgrudzinska: yes, I do need this now.

till now `simpleXmlParseW3CAttribute()` got args like "fill:red; stroke:blue\0" - there was the '\0' sign at the end

and now it's no longer the case.
let say we have such content in our svg file:
```
  <style>
    rect.mine {
        fill: blue;
        stroke: #00ff00;
    }
    circle.mine {
        fill: blue;
        stroke: #00ff00;
    }
  </style>
```
to read this, in the funct '_svgLoaderParserXmlCssStyle()' I'm calling in a while loop 'simpleXmlParseCSSAttribute()'.
it gets as one of its argument, the 'content' the following (in the first iteration) :
```
    rect.mine {
        fill: blue;
        stroke: #00ff00;
    }
    circle.mine {
        fill: blue;
        stroke: #00ff00;
    }
```
inside 'simpleXmlParseCSSAttribute()' the 'tag' and 'name' are allocated (`tag=rect` and `name=mine`), but I do not allocate memory for "fill:blue;stroke:#00ff00;", instead I return a pointer `attrs` pointing on it and the length `attrsLength`.
so if I try to print the whole `attrs` variable, I'll get:
```
fill: blue;
        stroke: #00ff00;
    }
    circle.mine {
        fill: blue;
        stroke: #00ff00;
    }
```
`attrs` are passed to `simpleXmlParseW3CAttribute()`, but I don;t want this function to read all of that, only the first part, till the first `}` sign occurrence. That's why I need to pass also 'attrsLength'.

Do you think that I should allocate the memory for `attrib`? Then I won't need `attrsLength` any more, but I thought that I don;t need a separate memory to store this variable.

src/loaders/svg/tvgXmlParser.cpp
```diff
@@ -467,6 +475,12 @@ bool simpleXmlParseW3CAttribute(const char* buf, simpleXMLAttributeCb func, cons
     do {
         char* sep = (char*)strchr(buf, ':');
         next = (char*)strchr(buf, ';');
+        if (sep >= end)
+        {
```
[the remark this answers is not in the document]
mgrudzinska: done

mgrudzinska:
@JSUYA example added (#1175), style fixed. I also answered regarding the additional arg in `simpleXmlParseW3CAttribute`

JSUYA (approved):
LGTM.
@hermet any ideas?

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2644,6 +2672,46 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content,
...
+    GradientFactoryMethod gradientMethod;
+    SvgNode *node = nullptr;
+
+    const char *buf = content;
+    unsigned buflen = length;
+
+
+    while (auto next = simpleXmlParseCSSAttribute(buf, buflen, &tag, &name, &attrs, &attrsLength)) {
+        if ((method = _findGroupFactory(tag))) {
+            //TODO - node->id ??? add additional var for svgnode?
```
hermet, line 2806: @mgrudzinska tip: that will be better if you add TVGLOG() for those missing tag items so that we can figure out problems later when unsupported svg data comes.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -75,7 +75,8 @@
 #define PX_PER_CM 37.79528f //1 in = 2.54 cm -> PX_PER_IN/2.54
 
 
-typedef SvgNode* (*FactoryMethod)(SvgLoaderData* loader, SvgNode* parent, const char* buf, unsigned bufLength);
+typedef bool (*parseAttributes)(const char*, unsigned, simpleXMLAttributeCb, const void*);
```
hermet, line 78: I think you can add parameter names here in order to clarify the function specification.

src/loaders/svg/tvgXmlParser.cpp
```diff
@@ -449,7 +449,7 @@ bool simpleXmlParse(const char* buf, unsigned bufLength, bool strip, simpleXMLCb
 }
 
 
-bool simpleXmlParseW3CAttribute(const char* buf, simpleXMLAttributeCb func, const void* data)
+bool simpleXmlParseW3CAttribute(const char* buf, TVG_UNUSED unsigned buflen, simpleXMLAttributeCb func, const void* data)
```
hermet: Suggestion the convention
buflen => bufLength
```
 static SvgNode* _createCssStyleNode(SvgLoaderData* loader, SvgNode* parent, const char* buf, unsigned bufLength, parseAttributes func)
```

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2137,9 +2144,9 @@ static void _cloneNode(SvgNode* from, SvgNode* parent)
 }
 
 
-static void _postponeCloneNode(SvgLoaderData* loader, SvgNode *node, char* id)
+static void _postponeCloneNode(Array<SvgNodeIdPair>* nodes, SvgNode *node, char* id)
```
hermet, line 2149: Suggestion: ```Array <SvgNodeIdPair>& nodes```

hermet (approved):
Overall looks fine. just added some nit-picks.
Plus, I just suggest two things.

1. CSS style is an additional spec of the SVG. I think it would better if we can separate the css related implementation from the SVG Loader file (ie. tvgSvgStyle.cpp)
2. We can maintain the list of the css - spec tvg are supporting. it's good point since it's just begun. It would be helpful if you write the wiki for supported css style spec/tag table.

Maybe we can release the css feature in the next major version up release.
Overall good job!  Thanks.

mgrudzinska:
@hermet I'll try to apply your comments tmr, or at least some of them (I'm not sure whether I see how to separate the css part)

"Maybe we can release the css feature in the next major version up" - do you mean the 1st of February ?

hermet:
@mgrudzinska The date of the next release needs our consensus. We can release the next major version later/after Feb.  at least, we must confirm the CSS feature in this release. (This doesn't mean that we have to cover the full spec.)

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: please check the example:
```
<svg viewBox="0 0 200 200">
    <defs>
        <style>
        <![CDATA[
            rect.mineStyle {fill: blue}
        ]]>
        </style> 
    </defs>

    <![CDATA[ 
        rect {stroke-width: 15; stroke: red}
    ]]>

    <rect x="0" y="0" width="50" height="50" class="mineStyle" />
</svg>
```

The second CDATA block should be skipped. 
While it is read, `loader->cssStyle != nullptr` and `loader->style = false`., so checking the `loader->cssStyle` value won't be enough.
Although I pushed a commit, since the `loader->style` value was reset in a wrong place. fixed now

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: done

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: done

src/loaders/svg/tvgXmlParser.cpp
mgrudzinska: changed also in tvgXmlParser.h

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: may I ask why? so that in the future I'll use the preferred method right away

I didn't apply this suggestion, since this function was defined earlier and it is directly connected with this pr (I needed to predefine it), though I can do it in a separate pr in all places (clonePostponedNodec, stylePostponedNodes, postponeNodeClone)

mgrudzinska:
@hermet 
please check: 
https://github.com/Samsung/thorvg/wiki/CSS-styling
I'll expand the most important section - internal css. I didn't have time these days, I'll do it on weekend. Also, I hope to implement on weekend the features marked as `tbd`, but we'll see :)

I didn't move css functs into a separate file - I'll do it on weekend

@hermet @JSUYA thank you for your revs

mgrudzinska:
@JSUYA @hermet please double check last two commits, especially whether I didn't miss anything while moving the `_updateStyle(loaderData.doc, nullptr)` function call. 

I didn;t move to the new file the following functs: `_svgLoaderParserXmlCssStyle`, `_createCssStyleNode'`and `_attrParseCssStyleNode`. I think they should stay in the `tvgSvgLoader.cpp` where all other functs like them are. Please let me know if I'm wrong

Issue #1161, opened 2022-01-18
Stack overflow when converting image


qarmin:
```
/home/rafal/test/thorvg/build/src/bin/svg2png/svg2png /home/rafal/Desktop/SV/file0.svg -r 1x1
```
```
=================================================================
==77425==ERROR: AddressSanitizer: stack-overflow on address 0x7f25d13ffff8 (pc 0x7f25d66df8dc bp 0x000000000000 sp 0x7f25d1400000 T1)
    #0 0x7f25d66df8dc in __sanitizer::StackDepotBase<__sanitizer::StackDepotNode, 1, 20>::Put(__sanitizer::StackTrace, bool*) ../../../../src/libsanitizer/sanitizer_common/sanitizer_stackdepotbase.h:109
    #1 0x7f25d66df36b in __sanitizer::StackDepotPut(__sanitizer::StackTrace) ../../../../src/libsanitizer/sanitizer_common/sanitizer_stackdepot.cpp:98
    #2 0x7f25d663650a in __asan::Allocator::Allocate(unsigned long, unsigned long, __sanitizer::BufferedStackTrace*, __asan::AllocType, bool) ../../../../src/libsanitizer/asan/asan_allocator.cpp:573
    #3 0x7f25d6632d47 in __asan::asan_realloc(void*, unsigned long, __sanitizer::BufferedStackTrace*) ../../../../src/libsanitizer/asan/asan_allocator.cpp:1000
    #4 0x7f25d66bebf4 in __interceptor_realloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:165
    #5 0x7f25d6385869 in tvg::Array<float>::reserve(unsigned int) ../src/lib/tvgArray.h:56
    #6 0x7f25d638d566 in _styleInherit ../src/loaders/svg/tvgSvgLoader.cpp:1785
................
    #481 0x7f25d639353c in _cloneNode ../src/loaders/svg/tvgSvgLoader.cpp:1948
    #482 0x7f25d639353c in _cloneNode ../src/loaders/svg/tvgSvgLoader.cpp:1935
    #483 0x7f25d639353c in _cloneNode ../src/loaders/svg/tvgSvgLoader.cpp:1948
    #484 0x7f25d639353c in _cloneNode ../src/loaders/svg/tvgSvgLoader.cpp:1935
    #485 0x7f25d639353c in _cloneNode ../src/loaders/svg/tvgSvgLoader.cpp:1948
[...]

PR #1164, opened 2022-01-18
svg_loader: preventing cyclic node cloning

mgrudzinska:
If a parent node is a child of the node from which
properties are copied, a cyclic call appears. Fixed

issue #1161

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1932,7 +1932,17 @@ static void _copyAttr(SvgNode* to, const SvgNode* from)
 }
 
 
-static void _cloneNode(SvgNode* from, SvgNode* parent)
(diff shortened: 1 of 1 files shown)


mgrudzinska:
it's not a global solution, I need to change it

mgrudzinska:
svg with a cycle:
```
<svg height="100" width="100">
  <mask id="d">
  <g>
    <use href="#d"/>
  </g>
  </mask>
  <circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red" />
</svg>
```

before this pr caused stackoverflow

mgrudzinska:
> it's not a global solution, I need to change it

now it is

hermet:
@mgrudzinska this is my suggestion: https://github.com/Samsung/thorvg/pull/1168

Issue #1162, opened 2022-01-18
Different color of object after converting it to png


qarmin:
![Screenshot from 2022-01-18 17-31-48](https://user-images.githubusercontent.com/41945903/149978606-12c25c71-6c90-4af7-bbb9-8a032f336158.png)

[AAA.zip](https://github.com/Samsung/thorvg/files/7890765/AAA.zip)

qarmin:
Here entire object is missing
[90.zip](https://github.com/Samsung/thorvg/files/7890800/90.zip)
![Screenshot from 2022-01-18 17-36-06](https://user-images.githubusercontent.com/41945903/149979300-fedd7aee-21fb-4dcb-b804-94c9caab72b2.png)

mgrudzinska:
The problem with 90.svg is solved in #1165.

54.svg and 60.svg - both are using css styling, which will be supported in tvg very soon (i checked both files in pr #1160 and they look correct)

PR #1165, opened 2022-01-18
svg_loader: group tags misinterpreted as parent nodes

mgrudzinska:
Group tags in a format <tag .../> finds no application
to the rest of the svg file and should be skipped.

issue #1162 - the 90.svg file

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2538,6 +2538,7 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content, 
 
     if ((method = _findGroupFactory(tagName))) {
         //Group
+        if (empty) return;
         if (!loader->doc) {
             if (strcmp(tagName, "svg")) return; //Not a valid svg document
             node = method(loader, nullptr, attrs, attrsLength);


mgrudzinska:
@hermet please confirm.
The group tags are `defs`, `g`, `svg`, `mask `and `clipPath`. None of these make any difference to the rest of the file if they are given in the format `<tag .../>`, so if they have no children nodes.

mgrudzinska:
all of these are not correct right now in tvg (without this pr):

transformation from the first `g` tag shouldn't be applied:
```
<svg viewBox="0 0 100 100" >
  <g transform='translate(40,40)'/>
  <g transform='translate(50,50)'>
    <path d='m 0 0 h 100 v 100 z' fill="red"/>
  </g>
</svg>
```

mask should be ignored:
```
<svg viewBox="0 0 100 100" >
  <mask id="myMask"/>
  <g transform='translate(50,50)'>
    <path d='m 0 0 h 100 v 100 z' fill="red"/>
  </g>
</svg>
```

not a valid svg:
```
<svg viewBox="0 0 100 100" />
    <path d='m 0 0 h 100 v 100 z' fill="red"/>
</svg>
```

JSUYA (approved):
LGTM

hermet (approved):

PR #1168, opened 2022-01-20
svg_loader: ++robustness

hermet:
Prevent recursive calls by counting just in case.

The size is arbitrary value, we can adjust it experimentally.

@Issue: https://github.com/Samsung/thorvg/issues/1161

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1934,8 +1934,15 @@ static void _copyAttr(SvgNode* to, const SvgNode* from)
 }
 
 
-static void _cloneNode(SvgNode* from, SvgNode* parent)
+static void _cloneNode(SvgNode* from, SvgNode* parent, int depth)
 {
+    /* Exception handling: Prevent invalid SVG data input.
+       The size is the arbitrary value, we need an experimental size. */
+    if (depth == 8192) {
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1934,8 +1934,15 @@ static void _copyAttr(SvgNode* to, const SvgNode* from)
 }
 
 
-static void _cloneNode(SvgNode* from, SvgNode* parent)
+static void _cloneNode(SvgNode* from, SvgNode* parent, int depth)
 {
+    /* Exception handling: Prevent invalid SVG data input.
+       The size is the arbitrary value, we needs a experimental size of it. */
+    if (depth == 8192) {
+        TVGERR("SVG", "Infinite recursive call = %d! A current SVG data may be wrongly written!\n", depth);
```
mgrudzinska, line 1942: my suggestion, but as u wish:
Infinite recursive call - stopped after %d calls! Svg file may be incorrectly formatted.

mgrudzinska (approved):

src/loaders/svg/tvgSvgLoader.cpp
hermet: @mgrudzinska  applied with your suggestion. thanks.

PR #1170, opened 2022-01-20
sw_engine: float casted on the unsigned int type

mgrudzinska:
uint32_t -> int32_t


mgrudzinska:
I think it's an undef behavior and not a portability issue

mgrudzinska:
I know that a similar problem is in the raw loader, but I'd like to do some checks before fixing it there

hermet (approved):

PR #1173, opened 2022-01-24
loaders jpg: fix all memory leaks.

hermet:
These were detected by asan with PictureJpg example,

fixed them all.


mgrudzinska (approved):

src/loaders/jpg/tvgJpgLoader.cpp
```diff
@@ -128,5 +129,9 @@ unique_ptr<Surface> JpgLoader::bitmap()
 
 void JpgLoader::run(unsigned tid)
 {
+    if (image) {
```
mgrudzinska, line 132: same question as for png - how can I have here image != nullptr?

PR #1174, opened 2022-01-24
loader png: fix all memory leaks.

hermet:
These were detected by asan with PicturePng example,

fixed them all.

### DIFF
--- a/src/loaders/png/tvgPngLoader.cpp
+++ b/src/loaders/png/tvgPngLoader.cpp
@@ -72,6 +72,7 @@ PngLoader::PngLoader()
 PngLoader::~PngLoader()
 {
     if (freeData) free(data);
+    free(image);
 }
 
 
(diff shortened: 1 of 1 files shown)


src/loaders/png/tvgPngLoader.cpp
```diff
@@ -180,10 +181,14 @@ unique_ptr<Surface> PngLoader::bitmap()
 
 void PngLoader::run(unsigned tid)
 {
+    if (image) {
```
mgrudzinska, line 184: just a question - how can we achieve `image != null` here?

hermet: if user try to read the data twice? 
open() -> read() -> open() -> read();

hermet: Null checking is not quite necessary. Either case it's trivial, you can refactor the code if you want. :-)

PR #1175, opened 2022-01-25
example: cssStyle.svg add

mgrudzinska:
For now it's a simple file used to check the styling precedence
and overall correctness.
After the css stylling will be fully introduced a more complex file
can be used.

current result;
07_css_after_before

after #1160 (the expected result):
08_css_ex


JSUYA (approved):
LGTM

hermet (approved):

PR #1180, opened 2022-02-04
loader png external: fix potential memory leak if read fails

mmaciola:


mgrudzinska (approved):
image will be freed internally in libpng, approved

PR #1191, opened 2022-02-16
jpeg_loader: resolve the asan bug report.

hermet:
fix the invalid negative shift operations.

@Issue: https://github.com/Samsung/thorvg/issues/1172

### DIFF
--- a/src/loaders/jpg/tvgJpgd.cpp
+++ b/src/loaders/jpg/tvgJpgd.cpp
@@ -382,8 +382,8 @@ struct Row
         const int tmp2 = z1 + MULTIPLY(z3, - FIX_1_847759065);
         const int tmp3 = z1 + MULTIPLY(z2, FIX_0_765366865);
 
-        const int tmp0 = (ACCESS_COL(0) + ACCESS_COL(4)) << CONST_BITS;
-        const int tmp1 = (ACCESS_COL(0) - ACCESS_COL(4)) << CONST_BITS;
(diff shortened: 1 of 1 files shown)


src/loaders/jpg/tvgJpgd.cpp
```diff
@@ -2607,7 +2607,7 @@ void jpeg_decoder::decode_block_ac_refine(jpeg_decoder *pD, int component_id, in
 {
     int s, k, r;
     int p1 = 1 << pD->m_successive_low;
-    int m1 = (-1) << pD->m_successive_low;
+    int m1 = static_cast<unsigned int>(-1) << pD->m_successive_low;
```
mgrudzinska, line 2610: @hermet 
don't u think that it's not correct now?
The thing is, that left shifting of a negative val is undef beh... but I don;t think that a simple casting is a solution in this case

```
  int32_t i = -1;
  uint32_t ui = static_cast<uint32_t>(i);

  cout << (i << 2) << " " << (ui << 2) << endl;
```

hermet: Your sample code is not identical with this case.
I think you must test like this instead.
```
int32_t i = -1;
printf("%d %d\n", i << 2, ((uint_32t) i) << 2);
```

Anyway, the old version is not correct. 
This change keeps the original logic while avoiding the unexpected result by the compilers.

PR #1194, opened 2022-02-21
svg_loader: handle the exception properly.

hermet:
viewBox doesn't expect the missing attributes,
it won't have any default values.

So we can decide the fault when the values are missed.

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1143,13 +1143,8 @@ static bool _attrParseSymbolNode(void* data, const char* key, const char* value)
     SvgSymbolNode* symbol = &(node->node.symbol);
 
     if (!strcmp(key, "viewBox")) {
-        if (_parseNumber(&value, &symbol->vx)) {
-            if (_parseNumber(&value, &symbol->vy)) {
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1143,13 +1143,8 @@ static bool _attrParseSymbolNode(void* data, const char* key, const char* value)
...
     if (!strcmp(key, "viewBox")) {
-        if (_parseNumber(&value, &symbol->vx)) {
-            if (_parseNumber(&value, &symbol->vy)) {
-                if (_parseNumber(&value, &symbol->vw)) {
-                    _parseNumber(&value, &symbol->vh);
-                }
-            }
-        }
+        if (!_parseNumber(&value, &symbol->vx) || !_parseNumber(&value, &symbol->vy)) return false;
+        if (!_parseNumber(&value, &symbol->vw) || !_parseNumber(&value, &symbol->vh)) return false;
```
mgrudzinska, line 1147: I'd rather use one `if` for all 4 calls

hermet: If it doesn't break tvg coding convention, there is no point.

PR #1195, opened 2022-02-21
initializer: thread safety

mmaciola:
If multiple threads was using thorvg, there was a high probability of
wrong initialization that meant TaskSchedulerImpl inst was null and
run() wasn't being called after TaskScheduler::request().
Adding mutex fixes it.

### DIFF
--- a/src/lib/tvgInitializer.cpp
+++ b/src/lib/tvgInitializer.cpp
@@ -22,6 +22,7 @@
 #include "tvgCommon.h"
 #include "tvgTaskScheduler.h"
 #include "tvgLoader.h"
+#include <mutex>
 
 #ifdef _WIN32
     #include <cstring>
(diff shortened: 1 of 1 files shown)


mmaciola:
I am attaching the test code: [MultithreadVectorImageTest.zip](https://github.com/Samsung/thorvg/files/8109251/MultithreadVectorImageTest.zip).
Fail doesn't occur every time, please run multiple times.
Before patch you will find print:
```

Vector image render multithread test

Rasterize: Draw fail 2
Rasterize: Draw fail 2
Rasterize: Draw fail 2
```

After patch, fail no more occurs.

hermet:
@mmaciola This is actually out of the policy. Other APIs aren't stable on multi-threading. 
I assume the user uses multi-threading.

1. The user must init/term the thorvg on the main thread, This case, the user can call thorvg APIs on the designated threads of each Canvas instance.
2. Supports the thread-safety only for the infrastructure such as Initializer. (your approach)
3. Leaves the locking to users.

@mmaciola Have you considered the above options?

hermet (changes requested):

Issue #1181, opened 2022-02-04
default value of aspect ratio is "xMinYMid meet"


mgrudzinska:
in an svg file the default value of the `preserveAspectRatio` attribute is `xMidYMid meet`
TVG supports only `none` value for this attrib. Anything else is interpreted as the default value, which seems to be `xMinYMid meet` instead of `xMidYMid`

mgrudzinska:
I think that the def value is '"xMinYMin meet"'.  Either way it should be changed

hermet:
@mgrudzinska could you please close this if this is resolved by the commits?

mgrudzinska:
@hermet I took me some time to respond :D
The issue is not solved, I try to solve it in the next week

mgrudzinska:
solved in #1249

PR #1196, opened 2022-02-23
svg_loader: preserveAspectRatio fix

mgrudzinska:
The default value should be xMidYMid, now it is.

issue #1181

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -705,17 +705,13 @@ unique_ptr<Scene> svgSceneBuild(SvgNode* node, float vx, float vy, float vw, flo
             auto tvy = vy * scale;
             auto tvw = vw * scale;
             auto tvh = vh * scale;
-            if (vw > vh) tvy -= (h - tvh) * 0.5f;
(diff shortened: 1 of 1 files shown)


JSUYA (approved):
SVG samples are looks fine.
But we don't have a test for `else` case (preserveAspect is false).
I think, If there is no problem, you can do merge this. Could you please check?

mgrudzinska:
> SVG samples are looks fine. But we don't have a test for `else` case (preserveAspect is false). I think, If there is no problem, you can do merge this. Could you please check?

the changes in the 'else' segment didn't change anything, since these lines:
```
            if (tvw > tvh) tvy -= (h - tvh) * 0.5f;
            else tvx -= (w - tvw) * 0.5f;
```
never have changed `tvy` or `tvx` values. if 'preserveAspect' is false, `h = tvh` and `w = tvw`


example to test the `if` segment:
```
<svg viewBox="-10 0 120 100" width="240" height="100">
   <path d="M -100 -100 h 300 v 300 h -300 z" fill="green" />
   <path d="M -100 -100 h 300 v 300 z" fill="blue" />
   <circle cx="50" cy="50" r="60" fill="red" />
</svg>
```

PR #1198, opened 2022-02-23
svg_loader: use node fixed

mgrudzinska:
By mistake the use node was improperly transformed and/or displayed
for a reference node other than a symbol node.


JSUYA (approved):
LGTM

PR #1199, opened 2022-02-23
svg_loader: width and height attribs of the use element applied

mgrudzinska:


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2123,6 +2123,10 @@ static bool _attrParseUseNode(void* data, const char* key, const char* value)
     for (unsigned int i = 0; i < sizeof(useTags) / sizeof(useTags[0]); i++) {
         if (useTags[i].sz - 1 == sz && !strncmp(useTags[i].tag, key, sz)) {
             *((float*)(array + useTags[i].offset)) = _toFloat(loader->svgParse, value, useTags[i].type);
+
+            if (useTags[i].offset == offsetof(SvgRectNode, w)) use->isWidthSet = true;
+            else if (useTags[i].offset == offsetof(SvgRectNode, h)) use->isHeightSet = true;
```
JSUYA, line 2128: It seems that SvgUseNode should be used, not SvgRectNode.
The same goes for lines 2107~2110.

JSUYA (changes requested):

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: thx! changed

JSUYA (approved):
LGTM :)

PR #1206, opened 2022-03-16
build: Add another `<cstring>` include for MinGW compatibility

akien-mga:
Same problem as #1159.

If you're interested, I could look into adding a MinGW build to your Windows CI pipeline.


hermet (approved):

hermet:
@akien-mga  great thanks!

Issue #1007, opened 2021-11-07
Scaling aliasing


fire:
Hi, thanks for getting raster image import working.

I noticed that the web viewer seems to have corrupted the display.

![image](https://user-images.githubusercontent.com/32321/140634250-4788d0ad-7d97-4c77-8f4d-4b4fb7e6cbb3.png)

[Godot Planner 2021-11-06.svg.txt](https://github.com/Samsung/thorvg/files/7491810/Godot.Planner.2021-11-06.svg.txt)

fire:
Firefox control screenshot. 
![image](https://user-images.githubusercontent.com/32321/140634338-66694e62-7405-44c9-8d2f-501fcdfad3d1.png)

JSUYA:
@mmaciola It seems to be a colorspace problem in png loader&png library. Could you please check?

JSUYA:
@fire  Thank you for your report. It is different from the issue you raised, but please refer to it
The reason why the arrow of the path is not displayed in our output is that we do not yet support the marker(`<marker>`,  `marker-end`)  :)
![image](https://user-images.githubusercontent.com/7413838/140681140-8e57bcd1-b12f-4f8a-906c-396198dbb578.png)

hermet:
It looks the image is not re-scaled with the higher resolution...

fire:
Good catch on the arrow ends. I didn’t notice. The main issue is the aliasing.

hermet:
@fire ok fundamentally, it's not about thorvg engine issue but resolution problem of the viewer. those fonts are too small at 1152x1152 and current viewer is available up to that resolution. if it allows to increase the resolution higher, the fonts will be looked same with the firefox. if the firefox renders that svg on the lower resolution like thorvg viewer the font drawing will be looked same ugly.

hermet:
@mmaciola do you have any opinion to support higher resolution in viewer side ? for example, ui layout is resized following to the selected resolution.

mmaciola:
I just added custom resolution patch for thorvg viewer: Drag and drop the image, click resolution and type desired one, press enter.

https://user-images.githubusercontent.com/71131832/140932284-b16a3f0f-5e06-4b26-8cd9-1c36855c7d45.mov

Regarding the colorspace of the top right image: it is a global problem of static png loader (not only viewer). LodePng is desired for RGB colorspace and it was only partly changed for BGR. If any non standard png, colorspace may be incorrect (rgb used).

hermet:
[...]

PR #1217, opened 2022-05-01
png_loader tvgPngLoader: Revert change to channel order

projectitis:
Revert an earlier change to store internal PNG channels as BGR (changed back to RGB).

- Reverts #1041
- Fixes #1112 (can be closed without merging)
- Reopens #1007 (needs to be fixed a different way)

PNGs now render correctly. I have tested this with 4 kinds of PNG, but not with grayscale PNGs yet.

However, please be aware that it doesn't matter if you target the buffer with ColorSpace ARGB8888 or ABGR8888, the PNG still appears the same both ways. i.e. the Picture renderer needs to be updated to swap the channels. This is likely the same bug as #1007 

```
// Changing the ColorSpace does not change the PNG channel order during drawing.
// Both of these result in the same drawing output.
canvas->target(buffer, width, width, height, tvg::SwCanvas::ARGB8888); // RGB
canvas->target(buffer, width, width, height, tvg::SwCanvas::ABGR8888); // BGR
```

### DIFF
--- a/src/loaders/png/tvgLodePng.cpp
+++ b/src/loaders/png/tvgLodePng.cpp
@@ -2090,17 +2090,7 @@ static unsigned unfilterScanline(unsigned char* recon, const unsigned char* scan
     size_t i;
     switch (filterType) {
         case 0: {
-            if (bytewidth == 4) {
-                for (i = 0; i < length; i += 4) {
-                    //RGBA -> BGRA
-                    recon[i + 0] = scanline[i + 2];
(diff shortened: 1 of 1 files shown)


JSUYA (commented):
https://github.com/Samsung/thorvg/pull/1222#discussion_r946278152 I mentioned here too 
This patch have problem with png color space representation.

left - As a result of applying this patch
right - chrome browser

hermet:
replaced with https://github.com/thorvg/thorvg/pull/1597

PR #1219, opened 2022-05-04
svg_loader: Fix invaild '<' nested check

JSUYA:
When Type is DOCTYPE, Child Entities start with '<'.
This condition is valid when general Elements (svg, g, path etc) are used.
Add a Doctype check to if condition.
(There can be various cases related to '<' nested case.
But for now, I only add Doctype considering the side effect.)

### DIFF
--- a/src/loaders/svg/tvgXmlParser.cpp
+++ b/src/loaders/svg/tvgXmlParser.cpp
@@ -388,7 +388,7 @@ bool simpleXmlParse(const char* buf, unsigned bufLength, bool strip, simpleXMLCb
 
             if (p) {
                 //Invalid case: '<' nested
-                if (*p == '<') return false;
+                if (*p == '<' && type != SimpleXMLType::Doctype) return false;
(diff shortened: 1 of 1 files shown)


JSUYA:
case)
```html
<?xml version="1.0" encoding="utf-8"?>
<!-- Generator: Adobe Illustrator 12.0.1, SVG Export Plug-In . SVG Version: 6.00 Build 51448)  -->
<!DOCTYPE svg [
	<!ENTITY ns_svg "http://www.w3.org/2000/svg">
	<!ENTITY ns_xlink "http://www.w3.org/1999/xlink">
]>
<svg  version="1.0" id="Layer_1" xmlns="&ns_svg;" xmlns:xlink="&ns_xlink;" width="33" height="33" viewBox="0 0 33 33"
	 style="overflow:visible;enable-background:new 0 0 33 33;" xml:space="preserve">

```

hermet (approved):

Issue #1218, opened 2022-05-03
Picture: Implement image meshes


projectitis:
The next version of the Rive renderer requires support for image meshes. ThorVG currently has support for a texture map (2 triangles) but would need to allow mapping via an arbitrary mesh.

projectitis:
@hermet @mgrudzinska @JSUYA @mmaciola I have added code for the low-level drawing of triangle meshes, based existing code in `tvgSwRasterTexmap.cpp`. Note that this is theoretical until I make/modify a Paint that supports it and can run and test it :)
Refer to [this commit on my fork](https://github.com/DeriveSDK/thorvg/commit/74f3af2191c635ae7941e42f5a9921f4daac0021).

My question is - how should the Paint look?  We have a few different options - what is your preference? (also feel free to suggest any alternative)

Option 1: Build it into picture
```
auto picture = tvg::Picture::gen();
picture->load(EXAMPLE_DIR"/test.png");
Polygon triangles[4] = { /* vertex data here (points + UVs) */ }
picture->mesh(&triangles, 4); // Define a mesh made of 4 triangles
canvas->push(move(picture));
```

Option 2: Extend Picture, but same thing:
```
// class ImageMesh extends Picture
auto mesh = tvg::Mesh::gen();
mesh ->load(EXAMPLE_DIR"/test.png");
Polygon triangles[4] = { /* vertex data here (point + UVs) */ }
mesh ->mesh(&triangles, 4);
canvas->push(move(mesh));
```

Option 3: Mesh which can share picture (this is my preference)
```
auto picture = tvg::Picture::gen();
picture->load(EXAMPLE_DIR"/test.png");

// class Mesh extends Paint
auto mesh1 = tvg::Mesh::gen();
mesh1->texture( picture.get() );
mesh1->mesh({ /* vertex data here (points + UVs) */ }, 7 );
canvas->push(move(mesh1));

auto mesh2 = tvg::Mesh::gen();
mesh2->texture( picture.get() );
mesh2->mesh({ /* vertex data here (points + UVs) */ }, 18 );
mesh2->mesh({ /* vertex data here (points + UVs) */ }, 3, true ); // Append more triangles to the mesh
canvas->push(move(mesh2));
```

Thoughts?

JSUYA:
I think option 1's approach is appropriate. :)  👍👍   @hermet ?

  +)  I'm not sure if mesh is a function that can only be used in picture class.
Does the same apply to SVG images?

[...]

PR #1222, opened 2022-05-11
Feature/image mesh

projectitis:
This is a draft PR to show my work in progress for image mesh support.
Any feedback and suggestions welcome.
Relates to #1218 

It is currently implemented as a feature of the Picture object:
```
auto picture = Picture::gen();
picture->load("texture.png");
Polygon triangles[4] = { /* vertex data here (points + UVs) */ }
picture->mesh(&triangles, 4); // Define a mesh made of 4 triangles
canvas->push(move(picture));
```

Note: This branch also contains various other fixes, like the PNG channel-order bug, and better windows support.


projectitis:
Hi @hermet @JSUYA - 

I have this working now, but there are currently two issues.
1) There is a seam between triangles
2) The clipping is not happening correctly

If you have any idea about how to fix these issues, or where they might be occurring, please let me know where to look!

Here is a video showing the issue:
https://user-images.githubusercontent.com/32183808/167791956-cc03826b-e9b2-4a2e-b46f-31b95fdb46c5.mp4

The code to create this is:
```
auto picture = Picture::gen();
picture->load("texture.png"); // 200 x 200 image (if different, change the UVs in the triangle data)
Polygon triangles[2];
triangles[0] = {{
	{
		{0,0}, // pt
		{0,0} // uv
	},
	{
		{200,-50},
		{200,0}
	},
	{
		{0,200},
		{0,200}
	}
}};
triangles[1] = { {
	{
		{200,-50},
		{200,0}
	},
	{
		{0,200},
		{0,200}
	},
	{
		{200,250},
		{200,200}
	}
} };
picture->mesh(&triangles, 2);
canvas->push(move(picture));
```

projectitis:
Hi @hermet @JSUYA - I have fixed (removed) the seams between the triangles by combining all triangles into one AA span area instead of treating each triangle as it's own one. I suspect that this will cause slower rendering for some shaped meshes because the area might contain many transparent/undrawn pixels. It's better to have no seams, though.

However, still the following issues.
1) The clipping area is still wrong
2) While rotating the mesh, at a certain point an exception is thrown trying to free the AA spans:
```
    free(aaSpans->lines);
    free(aaSpans); // exception thrown on this line
```
I believe that they are both related to each other because it always happens when the clipping area shrinks to it's smallest size.

Demo:
https://user-images.githubusercontent.com/32183808/167963654-0551f13e-70b2-401a-8614-d9f482dc06c0.mp4

projectitis:
`imagePrepare ---> _genOutline` is the problem with the clipping area.
But it does not fix the exception that occurs.

projectitis:
`imagePrepare ---> _genOutline` is the problem with the clipping area.
It was still assuming a rectangular clipping area of a standard image.
I have put in a quick fix (that probably can be improved a lot).
But fixing that does not fix the exception.

hermet:
> Hi @hermet @JSUYA -
> 
> I have this working now, but there are currently two issues.
> 
> 1. There is a seam between triangles
> 2. The clipping is not happening correctly
> 
> If you have any idea about how to fix these issues, or where they might be occurring, please let me know where to look!
> 
> Here is a video showing the issue: https://user-images.githubusercontent.com/32183808/167791956-cc03826b-e9b2-4a2e-b46f-31b95fdb46c5.mp4
> 
> The code to create this is:
> 
> ```
> auto picture = Picture::gen();
> picture->load("texture.png"); // 200 x 200 image (if different, change the UVs in the triangle data)
> Polygon triangles[2];
> triangles[0] = {{
> 	{
> 		{0,0}, // pt
> 		{0,0} // uv
> 	},
> 	{
> 		{200,-50},
> 		{200,0}
> 	},
> 	{
> 		{0,200},
> 		{0,200}
> 	}
> }};
> triangles[1] = { {
> 	{
> 		{200,-50},
> 		{200,0}
> 	},
> 	{
> 		{0,200},
> 		{0,200}
> 	},
> 	{
> 		{200,250},
> 		{200,200}
> 	}
> } };
> picture->mesh(&triangles, 2);
> canvas->push(move(picture));
> ```


Current AA logic is only adaptable to one triangle, it never consider the mesh types. However, you gonna make it, very impressive. :)

Clipping wise, I believe you can figure out and make it work. I believe it won't be difficult because its basic features are already working there. Otherwise you can ignore the clipping feature . We can revisit and enhance it step by step later. Please try to do best, I will revisit you code soon and point out the issue you need helps.  👍

projectitis:
@hermet I am back on this. Made some progress today. This is an image mesh made in the latest version of the Rive editor and rendered using the Rive-Tizen viewer with image mesh implemented!

p.s. That is my new puppy, Ginny.

image

image

hermet:
Good job, looks impressive. Please tell when the patch is ready. I will help for this. Once this patch is applied, I believe you can drive the feature development with dev-access. Also I wanted to talk about Derive SDK you're probably working for. It would be helpful for both of us If you describe your project mission or vision. :-)

projectitis:
@hermet thank you. It is actually the same code as now (no changes).  I am trying to fix the failing test. Then I will re-open the PR.

There is a bit of detail in the Derive SDK description - what information specifically should I add to that?

projectitis:
It would be great if somebody could help with a failing test :) 
* The tests on `master` all pass
* The picture tests on this branch have some failing tests (trying to access a NULL matrix)

But I cannot see the difference between the two branches that causes this 🤔 

This is the test that fails:
```
// testPicture.cppm line 325
TEST_CASE("Load JPG file and render", "[tvgPicture]")
{
    REQUIRE(Initializer::init(CanvasEngine::Sw, 0) == Result::Success);

    REQUIRE(canvas);

    uint32_t buffer[100*100];
    REQUIRE(canvas->target(buffer, 100, 100, 100, SwCanvas::Colorspace::ABGR8888) == Result::Success);

    auto picture = Picture::gen();
    REQUIRE(picture);

    REQUIRE(picture->load(TEST_DIR"/test.jpg") == Result::Success);

    REQUIRE(canvas->push(move(picture)) == Result::Success); // XXXXXXXXXXXXXXXXX Exception here

    REQUIRE(Initializer::term(CanvasEngine::Sw) == Result::Success);
}
```

This is the test result:
```
../test/testPicture.cpp(338): FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal
```

This is the call stack when the exception occurs:
```
tvgUnitTests.exe!_onlyShifted(const tvg::Matrix * m) Line 31
tvgUnitTests.exe!imagePrepare(SwImage * image, tvg::Polygon * triangles, unsigned int count, const tvg::Matrix * transform, const SwBBox & clipRegion, SwBBox & renderRegion, SwMpool * mpool, unsigned int tid) Line 123
tvgUnitTests.exe!SwImageTask::run(unsigned int tid) Line 196
tvgUnitTests.exe!tvg::TaskSchedulerImpl::request(tvg::Task * task) Line 156
tvgUnitTests.exe!tvg::TaskScheduler::request(tvg::Task * task) Line 185
tvgUnitTests.exe!tvg::SwRenderer::prepareCommon(SwTask * task, const tvg::RenderTransform * transform, unsigned int opacity, const tvg::Array<void *> & clips, tvg::RenderUpdateFlag flags) Line 623
tvgUnitTests.exe!tvg::SwRenderer::prepare(tvg::Surface * image, tvg::Polygon * triangles, unsigned int triangleCnt, void * data, const tvg::RenderTransform * transform, unsigned int opacity, tvg::Array<void *> & clips, tvg::RenderUpdateFlag flags) Line 640
tvgUnitTests.exe!tvg::Picture::Impl::update(tvg::RenderMethod & renderer, const tvg::RenderTransform * pTransform, unsigned int opacity, tvg::Array<void *> & clips, tvg::RenderUpdateFlag pFlag) Line 134
tvgUnitTests.exe!tvg::PaintMethod<tvg::Picture::Impl>::update(tvg::RenderMethod & renderer, const tvg::RenderTransform * transform, unsigned int opacity, tvg::Array<void *> & clips, tvg::RenderUpdateFlag renderFlag) Line 180
tvgUnitTests.exe!tvg::Paint::Impl::update(tvg::RenderMethod & renderer, const tvg::RenderTransform * pTransform, unsigned int opacity, tvg::Array<void *> & clips, unsigned int pFlag) Line 242
tvgUnitTests.exe!tvg::Canvas::Impl::update(tvg::Paint * paint, bool force) Line 97
tvgUnitTests.exe!tvg::Canvas::Impl::push(std::unique_ptr<tvg::Paint,std::default_delete<tvg::Paint>> paint) Line 57
tvgUnitTests.exe!tvg::Canvas::push(std::unique_ptr<tvg::Paint,std::default_delete<tvg::Paint>> paint) Line 48
```

The exception occurs in `_onlyShifted` because the matrix is NULL.
```
static inline bool _onlyShifted(const Matrix* m)
{
    if (mathEqual(m->e11, 1.0f) && mathEqual(m->e22, 1.0f) && mathZero(m->e12) && mathZero(m->e21)) return true; // <<< m is NULL
    return false;
}
```

Any ideas?

projectitis:
@hermet Ready for review!

inc/thorvg.h
```diff
@@ -1177,6 +1201,49 @@ class TVG_EXPORT Picture final : public Paint
...
+     * 
+     * @param[in] triangles An array of Polygon objects (triangles) that make up the mesh. Must be at least 1.
+     * @param[in] count The number of Polygon objects (triangles) provided
+     * 
+     * @retval Result::Success When succeed.
+     * @retval Result::NonSupport If there was an error
+     * 
+     * @note The Polygons are copied internally, so modifying them after calling Mesh::mesh has no affect.
+     */
+    Result mesh(const Polygon* triangles, const uint32_t count) noexcept;
```
JSUYA, line 1218: How about keeping only appendMesh?
According to the description of appendMesh, there is no need to use this.
@hermet, what do you think?

inc/thorvg.h
```diff
@@ -1177,6 +1201,49 @@ class TVG_EXPORT Picture final : public Paint
...
+     * If a mesh is provided, the transform property of the Picture will apply to the triangle mesh, and the
+     * image data will be used as the texture.
+     * 
+     * @param[in] triangles An array of Polygon objects (triangles) that make up the mesh. Must be at least 1.
+     * @param[in] count The number of Polygon objects (triangles) provided
+     * 
+     * @retval Result::Success When succeed.
+     * @retval Result::NonSupport If there was an error
+     * 
+     * @note The Polygons are copied internally, so modifying them after calling Mesh::mesh has no affect.
```
JSUYA, line 1232: Among the image types provided by Picture, tvg, svg are not currently supported. Could you please add an explanation for this?

inc/thorvg.h
```diff
@@ -1177,6 +1201,49 @@ class TVG_EXPORT Picture final : public Paint
...
+     * @brief Adds additional triangles to the mesh. It is safe to call this even if there is no mesh yet.
+     * 
+     * If a mesh is provided, the transform property of the Picture will apply to the triangle mesh, and the
+     * image data will be used as the texture.
+     * 
+     * @param[in] triangles An array of Polygon objects (triangles) that make up the mesh. Must be at least 1.
+     * @param[in] count The number of Polygon objects (triangles) provided
+     * 
+     * @retval Result::Success When succeed.
+     * @retval Result::NonSupport If there was an error
```
JSUYA, line 1230: Looking at the code, it seems that `Result::NonSupport` does not returned.
Could you please add an explanation for returning `Result::InsufficientCondition`?

src/lib/tvgPictureImpl.h
```diff
@@ -220,6 +259,24 @@ struct Picture::Impl
...
     }
 
+    bool mesh(const Polygon* triangles, const uint32_t count, const bool append)
+    {
+        if (!triangles || !count) return false;
+
+        if (append && this->triangles){
+            uint32_t newTriangleCnt = this->triangleCnt + count;
+            this->triangles = (Polygon*)realloc(this->triangles, sizeof(Polygon) * newTriangleCnt);
+            for (int i=0; i<count; i++) this->triangles[this->triangleCnt + i] = triangles[i];
```
JSUYA, line 269: Could you please add space?
`i<count` -> `i < count`

src/loaders/png/tvgLodePng.cpp
```diff
@@ -2090,17 +2086,7 @@ static unsigned unfilterScanline(unsigned char* recon, const unsigned char* scan
...
-                    //RGBA -> BGRA
-                    recon[i + 0] = scanline[i + 2];
-                    recon[i + 1] = scanline[i + 1];
-                    recon[i + 2] = scanline[i + 0];
-                    recon[i + 3] = scanline[i + 3];
-                }
-            } else {
-                for (i = 0; i != length; ++i) recon[i] = scanline[i];
-            }
+            for (i = 0; i < length; ++i) recon[i] = scanline[i];
```
JSUYA, line 2089: When this change takes effect, https://github.com/Samsung/thorvg/issues/1007#issuecomment-964153604 occurs again. Could you please check?

(https://github.com/Samsung/thorvg/issues/1007#issue-1046668771 Please check the diagram box color on the right of the SVG file attached here)

src/lib/tvgPictureImpl.h
JSUYA, line 269: Please remove warning message
```
../src/lib/tvgPictureImpl.h: In member function ‘bool tvg::Picture::Impl::mesh(const tvg::Polygon*, uint32_t, bool)’:
../src/lib/tvgPictureImpl.h:269:28: warning: comparison of integer expressions of different signedness: ‘int’ and ‘const uint32_t’ {aka ‘const unsigned int’} [-Wsign-compare]
  269 |             for (int i=0; i<count; i++) this->triangles[this->triangleCnt + i] = triangles[i];
      |                           ~^~~~~~
../src/lib/tvgPictureImpl.h:275:28: warning: comparison of integer expressions of different signedness: ‘int’ and ‘const uint32_t’ {aka ‘const unsigned int’} [-Wsign-compare]
  275 |             for (int i=0; i<count; i++) this->triangles[i] = triangles[i];
```

JSUYA (commented):
wow awesome work. Rive example is very impressive. Thank you for your work.
I left a few brief comments. Could you please check?

(I'm so glad we're back in thorvg after a long time )

inc/thorvg.h
projectitis: I will change this to a single method that matches the Impl:

`Result mesh(const Polygon* triangles, const uint32_t count, bool append = false) noexcept;`

Is that ok?

inc/thorvg.h
projectitis: Thank you for picking this up! I will change this to `Result::InvalidArguments` and add an explanation.

src/loaders/png/tvgLodePng.cpp
projectitis: I'll revert this change in this branch, and we can work on the channel order bug in the other branch.

inc/thorvg.h
projectitis: Done

inc/thorvg.h
projectitis: Done

inc/thorvg.h
projectitis: Done

src/lib/tvgPictureImpl.h
projectitis: Done

src/loaders/png/tvgLodePng.cpp
projectitis: Reverted

src/lib/tvgPictureImpl.h
projectitis: Done

projectitis:
@JSUYA Thank you for the review. I have made the suggested changes. Please do let me know if you have other suggestions.

I did not know if it is your policy for the reviewer to resolve the threads, or for the contributor to do so. I have left them for your to resolve. Please let me know.

inc/thorvg.h
```diff
@@ -184,6 +184,30 @@ struct Matrix
 };
 
 
+/**
+ * @brief A data structure representing a texture mesh vertex
+ * 
+ * @param pt The vertex coordinate
+ * @param uv The texture coordinate in the range (0..width, 0..height)
```
hermet, line 191: Could you please add ?  Once the feature is fully stable we will add the release tag.
@BETA_API

inc/thorvg.h
```diff
@@ -184,6 +184,30 @@ struct Matrix
...
+{
+   Point pt;
+   Point uv;
+};
+
+
+/**
+ * @brief A data structure representing a triange in a texture mesh
+ * 
+ * @param vertex The three vertices that make up the polygon
```
hermet, line 203: @BETA_API

inc/thorvg.h
hermet: In my opinion, the uv range would be better to use in the normalized range like (0 ~ 1). 
Otherwise, the user must get the size of the image first. We can skip that usage if it's possible.

inc/thorvg.h
```diff
@@ -1177,6 +1201,44 @@ class TVG_EXPORT Picture final : public Paint
...
+     * @param[in] count The number of Polygon objects (triangles) provided
+     * @param[in] append If false, will replace the mesh completely with the provided mesh. If true, will append the mesh.
+     * 
+     * @retval Result::Success When succeed.
+     * @retval Result::InvalidArguments If triangles is null, or if the count is 0
+     * 
+     * @note The Polygons are copied internally, so modifying them after calling Mesh::mesh has no affect.
+     * @note Only raster image types are supported at this time (png, jpg).
+     */
+    Result mesh(const Polygon* triangles, const uint32_t count, bool append = false) noexcept;
```
hermet, line 1229: @BETA_API

inc/thorvg.h
```diff
@@ -1177,6 +1201,44 @@ class TVG_EXPORT Picture final : public Paint
...
+    /**
+     * @brief Return the number of triangles in the mesh, and optionally get a pointer to the array of triangles in the mesh.
+     * 
+     * @param[out] triangles Optional. A pointer to the array of Polygons used by this mesh
+     * 
+     * @return uint32_t The number of polygons in the array
+     * 
+     * @note Modifying the triangles returned by this method will modify them directly within the mesh
+     */
+    uint32_t mesh(const Polygon** triangles) const noexcept;
```
hermet, line 1240: @BETA_API

inc/thorvg.h
```diff
@@ -1177,6 +1201,44 @@ class TVG_EXPORT Picture final : public Paint
...
+     * the UV is not between 0 and 1, it is between 0 and the image width/height. i.e.:
+     * 0 <= uv.x <= Picture.width
+     * 0 <= uv.y <= Picture.height
+     * 
+     * Only raster image types are supported at this time (png, jpg). Vector types like svg and tvg do not support
+     * mesh deformation. However, if required you should be able to render a vector image to a raster image and then apply a mesh.
+     * 
+     * @param[in] triangles An array of Polygon objects (triangles) that make up the mesh. Must be at least 1.
+     * @param[in] count The number of Polygon objects (triangles) provided
+     * @param[in] append If false, will replace the mesh completely with the provided mesh. If true, will append the mesh.
```
hermet, line 1221: Do it really need the append function?? I'm worrying if it increases the feature-complexity without any practical benefits.

inc/thorvg.h
```diff
@@ -1177,6 +1201,44 @@ class TVG_EXPORT Picture final : public Paint
...
+     * 
+     * @retval Result::Success When succeed.
+     * @retval Result::InvalidArguments If triangles is null, or if the count is 0
```
hermet, line 1224: I wonder if the user wants to remove the mesh from the Picture instance.  For instance they can remove the existing mesh by setting Polygon = nullptr with count = 0;

src/lib/sw_engine/tvgSwImage.cpp
```diff
@@ -51,10 +51,50 @@ static bool _genOutline(SwImage* image, const Matrix* transform, SwMpool* mpool,
...
+            else if (triangles[i].vertex[2].pt.x > max.x) max.x = triangles[i].vertex[2].pt.x;
+            if (triangles[i].vertex[2].pt.y < min.y) min.y = triangles[i].vertex[2].pt.y;
+            else if (triangles[i].vertex[2].pt.y > max.y) max.y = triangles[i].vertex[2].pt.y;
+        }
+
+        to[0] = {min.x,min.y};
+        to[1] = {max.x,min.y};
+        to[2] = {max.x,max.y};
+        to[3] = {min.x,max.y};
+    }
```
hermet, line 89: coding convention
if () {
  ...
} else {
  ...
}

src/lib/sw_engine/tvgSwCommon.h
```diff
@@ -321,7 +321,7 @@ bool strokeParseOutline(SwStroke* stroke, const SwOutline& outline);
 SwOutline* strokeExportOutline(SwStroke* stroke, SwMpool* mpool, unsigned tid);
 void strokeFree(SwStroke* stroke);
 
-bool imagePrepare(SwImage* image, const Matrix* transform, const SwBBox& clipRegion, SwBBox& renderRegion, SwMpool* mpool, unsigned tid);
+bool imagePrepare(SwImage* image, Polygon* triangles, uint32_t count, const Matrix* transform, const SwBBox& clipRegion, SwBBox& renderRegion, SwMpool* mpool, unsigned tid);
```
hermet, line 324: suggest more clear name:
i.e) count -> trianglesCnt
Other parameter names in functions as well.

src/lib/sw_engine/tvgSwRasterTexmap.h
```diff
@@ -600,3 +587,54 @@ static bool _rasterTexmapPolygon(SwSurface* surface, const SwImage* image, const
...
+      Should provide two Polygons, one for each triangle.
+      // TODO: region?
+*/
+static bool _rasterTexmapPolygonMesh(SwSurface* surface, const SwImage* image, const Polygon* triangles, const uint32_t count, const Matrix* transform, const SwBBox* region, uint32_t opacity, uint32_t (*blendMethod)(uint32_t))
+{
+    //Exceptions: No dedicated drawing area?
+    if (!region && image->rle->size == 0) return false;
+
+    // Step polygons once to transform
+    Polygon* transformedTris = (Polygon*)malloc(sizeof(Polygon) * count);
```
hermet, line 611: suggestion: Polygon* => auto

src/lib/sw_engine/tvgSwRasterTexmap.h
```diff
@@ -600,3 +587,54 @@ static bool _rasterTexmapPolygon(SwSurface* surface, const SwImage* image, const
...
+        if (transformedTris[i].vertex[1].pt.y < ys) ys = transformedTris[i].vertex[1].pt.y;
+        else if (transformedTris[i].vertex[1].pt.y > ye) ye = transformedTris[i].vertex[1].pt.y;
+        if (transformedTris[i].vertex[2].pt.y < ys) ys = transformedTris[i].vertex[2].pt.y;
+        else if (transformedTris[i].vertex[2].pt.y > ye) ye = transformedTris[i].vertex[2].pt.y;
+    }
+
+    // Get AA spans and step polygons again to draw
+    auto aaSpans = _AASpans(ys, ye, image, region);
+    if (aaSpans) {
+        for (uint32_t i=0; i<count; i++) {
```
hermet, line 630: fix convention
i=0   =>   i = 0
i<count   =>   i < count

src/lib/sw_engine/tvgSwRasterTexmap.h
```diff
@@ -600,3 +587,54 @@ static bool _rasterTexmapPolygon(SwSurface* surface, const SwImage* image, const
...
+            Polygon polygon = transformedTris[i];
```
hermet, line 631: Probably we can avoid the data copy?
_rasterPolygonImage(surface, image, region, opacity, transformedTris[i], blendMethod, aaSpans);

src/lib/tvgPicture.cpp
```diff
@@ -119,3 +119,17 @@ const uint32_t* Picture::data(uint32_t* w, uint32_t* h) const noexcept
     if (pImpl->surface) return pImpl->surface->buffer;
     else return nullptr;
 }
+
+
+Result Picture::mesh(const Polygon* triangles, const uint32_t count, bool append) noexcept
+{
+    if (!triangles || !count) return Result::InvalidArguments;
```
hermet, line 126: suggestion: !count -> count == 0

src/lib/tvgPictureImpl.h
```diff
@@ -166,10 +173,42 @@ struct Picture::Impl
...
+                if (triangles[i].vertex[2].pt.y < min.y) min.y = triangles[i].vertex[2].pt.y;
+                else if (triangles[i].vertex[2].pt.y > max.y) max.y = triangles[i].vertex[2].pt.y;
+            }
+
+            if (x) *x = min.x;
+            if (y) *y = min.y;
+            if (w) *w = max.x - min.x;
+            if (h) *h = max.y - min.y;
+
+        }
```
hermet, line 203: convention
if (xx) {
} else {
}

src/lib/tvgPictureImpl.h
```diff
@@ -220,6 +259,22 @@ struct Picture::Impl
         return Result::Success;
     }
 
+    Result mesh(const Polygon* triangles, const uint32_t count, const bool append)
+    {
+        if (append && this->triangles){
```
hermet, line 264: convention:
if (xx) {

src/lib/tvgPictureImpl.h
```diff
@@ -220,6 +259,22 @@ struct Picture::Impl
...
+            uint32_t newTriangleCnt = this->triangleCnt + count;
+            this->triangles = (Polygon*)realloc(this->triangles, sizeof(Polygon) * newTriangleCnt);
+            for (uint32_t i = 0; i < count; i++) this->triangles[this->triangleCnt + i] = triangles[i];
+            this->triangleCnt = newTriangleCnt;
+        }
```
hermet, line 269: convention:
if () {
} else {
}

src/lib/sw_engine/tvgSwRenderer.cpp
```diff
@@ -181,6 +181,8 @@ struct SwShapeTask : SwTask
 struct SwImageTask : SwTask
 {
     SwImage image;
+    Polygon* triangles;
+    uint32_t count;
```
hermet, line 185: Suggest clear name here. triangleCnt

hermet (changes requested):
Overall good to go. Here some nit-picks I'd like to ask you check them.
If you have any opinions against them please comments. :)

hermet:
One more, the commits are too messy here. If you clean up with the minimum necessary patch set. It would be much better.
Otherwise, I will squash them into one commit.

inc/thorvg.h
projectitis: @hermet I originally [matched skia](https://github.com/rive-app/rive-cpp/blob/master/skia/renderer/src/skia_factory.cpp#L170), which uses 0 - width/height, but I agree that 0 - 1 is better and I will change it.

inc/thorvg.h
projectitis: @hermet I agree. Will change to be set only.

inc/thorvg.h
projectitis: @hermet Agree. Will remove mesh if triangles==nullptr or count is 0.

src/lib/sw_engine/tvgSwCommon.h
projectitis: @hermet I agree - I actually just used count because I saw it used in other functions, like `Result colorStops(const ColorStop* colorStops, uint32_t cnt) noexcept;` but triangleCount is better.

projectitis:
@hermet Thank you for the review. I have addressed all your comments. The biggest changes are:

1. Polygon (triangle) now expects normalized UV coordinates. In order to do this, the renderer converts them to image coordinates immediately before rendering. This is necessary because the user can load a new image (with same mesh) which might have a different size.
2. Picture.mesh will now clear the mesh if called with nullptr or count==0

Aside from that, updates are argument names, linting and comments (including @BETA_API tag).

I am happy to receive more suggestions.

Please squash commits when this PR is ready to merge.

JSUYA (approved):
Overall I think it's good. Thanks again for your work.
I left comments, please check it.
and I approved. Wait for @hermet's approval

src/lib/tvgPictureImpl.h
```diff
@@ -220,6 +258,20 @@ struct Picture::Impl
...
+            for (uint32_t i = 0; i < triangleCount; i++) this->triangles[i] = triangles[i];
+        } else {
+            free(this->triangles);
+            this->triangles = nullptr;
+            this->triangleCount = 0;
+        }
+        return true;
+    }
+
     Paint* duplicate()
```
hermet, line 275: I missed one point, duplicate() must copy the interal member data of the Picture. So do polygon data as well.
Could you please copy mesh data here as well?

hermet (changes requested):
Please check comments one more, Also could you please rebase the patch so that we can merge patch without any conflicts?

src/lib/tvgPicture.cpp
projectitis: No longer valid (function has changed)

projectitis:
@hermet Thanks for remembering the duplicate() function. I also double-checked all comments and fixed some additional linting. The only difference now that I can see from your comments is `triangleCnt` vs `triangleCount`.

I squashed commits and rebased to master.

Hopefully this is all good to go once the checks have passed.

hermet (approved):

hermet:
applied with this: 
https://github.com/Samsung/thorvg/commit/3dd65dfed00849f0bd9d0bb0ade177fa961cd7a5

thanks!

PR #1225, opened 2022-05-23
Fix building with GCC 13.

marxin:
Fixes the following error:
`../src/lib/tvgLzw.h:27:5: error: ‘uint8_t’ does not name a type`


hermet (approved):
thanks!

PR #1226, opened 2022-05-23
Format code files with dos2unix, ensure newline at EOF

akien-mga:
Just a codebase quality suggestion, feel free to accept or reject (or ask for changes).

I ran those two commands on the `.cpp` and `.h` files:
```bash
find . -type f -not -path '*/\.git/*' \( -name "*.cpp" -o -name "*.h" \) -exec dos2unix {} \;
find . -type f -not -path '*/\.git/*' \( -name "*.cpp" -o -name "*.h" \) -exec sed -i -e '$a\' {} \;
```

`dos2unix` makes the file use LF line endings (Unix style) instead of CRLF (DOS style). It also removes BOMs (Byte-Order-Mark). This was already the case for the majority of files and usually the standard for cross-platform projects.

The `sed` commands adds a newline at the end of file if there isn't, also to match the Unix convention. This helps avoid diff such as this one (provided that newlines at EOF don't get removed by Windows contributors):
image


akien-mga:
Converting to draft for now as I need to test this more - I updated Godot to thorvg `master` with these changes and our SVG loading broke, I'm looking into why (there's almost no change in thorvg itself so I'm a bit puzzled - so who knows, it might be due to BOM removal in this PR or simlar): https://github.com/godotengine/godot/pull/61326

hermet (approved):
thanks!

PR #1227, opened 2022-05-23
Cleanup FreeBSD preprocessor checks (if vs ifdef)

akien-mga:
Should be equivalent but checking if the value is defined (`#ifdef`) is cleaner
than checking if it is set to something different from 0 (`#if`).


hermet (approved):

hermet:
thanks!

PR #1223, opened 2022-05-12
Fix alloca on FreeBSD

mundusnine:
Fix compile errors on FreeBSD.
- alloca is defined in stdlib.h on [freeBSD](https://www.freebsd.org/cgi/man.cgi?alloca)
Issue in [godot](https://cirrus-ci.com/task/5485473287110656?logs=build#L657)

PR #1228, opened 2022-06-11
fix OpenBSD compile issue (alloca strikes again)

omar-polo:
This is a similar, slightly more generic way to fix the issue (partially) fixed in #1223 and #1224.  I've found this issue trying to build Godot on OpenBSD.

Linux seems the only system AFAICS to provide `alloca.h`.  All the BSDs have the declaration in `stdlib.h`.

Truth be told, we could just as well use `stdlib.h` on linux too (at least when building on glibc) but [the alloca manpage for linux](https://man.voidlinux.org/alloca#Notes_on_the_GNU_version) lists some caveats that I don't know if apply in this case (and I don't have a linux machine at hand to properly test.)

### DIFF
--- a/src/lib/sw_engine/tvgSwRaster.cpp
+++ b/src/lib/sw_engine/tvgSwRaster.cpp
@@ -22,10 +22,10 @@
 
 #ifdef _WIN32
     #include <malloc.h>
-#elif defined(__FreeBSD__)
-    #include <stdlib.h>
-#else
+#elif defined(__linux__)
     #include <alloca.h>
+#else
+    #include <stdlib.h>
 #endif
 
 #include "tvgMath.h"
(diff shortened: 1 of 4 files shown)


hermet:
@mundusnine hello, could you please help us to review this patch on FreeBSD? whether it's compilable or not. at least this works on my linux machine.

hermet:
@omar-polo sorry for late review, does this compilable on FreeBSD as well? I have no that machine, I can't test it...

omar-polo:
@hermet no problem :)

I can't test right now on FreeBSD too, but it shouldn't change anything there.  Instead of including `stdlib.h` only on FreeBSD and `alloca.h` otherwise, this pr switches only linux to use `alloca.h` and the other systems to use `stdlib.h` (WIN32 keeps using `malloc.h`).  So, on FreeBSD nothing should change, since both before and after only `stdlib.h` is included.  (it fixes the build on the others BSDs though!)

I'll try to compile-test in on a FreeBSD machine sometimes this week however.

Thanks!

hermet:
@omar-polo ah i see. i think it looks fine to me. thanks :)

hermet (approved):

omar-polo:
Thanks!

PR #1235, opened 2022-08-19
Windows: set TVG_EXPORT to dllexport also for mingw

vtorri:
In addition, remove TVG_EXPORT from enum class. They are types
not symbols

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -18,7 +18,7 @@
 #include <string>
 
 #ifdef TVG_BUILD
-    #if defined(_MSC_VER) && !defined(__clang__)
+    #if defined(_WIN32) && !defined(__clang__)
         #define TVG_EXPORT __declspec(dllexport)
         #define TVG_DEPRECATED __declspec(deprecated)
     #else
@@ -74,7 +74,7 @@ class Accessor;
 /**
  * @brief Enumeration specifying the result from the APIs.
  */
-enum class TVG_EXPORT Result
+enum class Result
 {
     Success = 0,           ///< The value returned in case of a correct request execution.
     InvalidArguments,      ///< The value returned in the event of a problem with the arguments given to the API - e.g. empty paths or null pointers.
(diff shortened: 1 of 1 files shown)


hermet:
@projectitis It would be better if you can verify on your window machine.

projectitis:
@hermet I can confirm this still works (build + tests) fine with both MSVC (cl, link) and LLVM clang (clang-cl, lld-link) on windows. 

I don't usually use mingw, but I installed mingw with msys and was able to build and test with clang++ as well.

I have not been able to test with gcc. I don't know how to configure the project for gcc. setting CXX=gcc just results in clang++ being selected.

But I am happy to verify this PR based on these results.

vtorri:
@projectitis If you follow the steps at msys2.org, once it is installed, in the start menu you can select the terminal with  predefined environment for gcc 32 bits, gcc 64 bits,clang 32 bits, clang 64bits, for ARM too, etc...
For example, in the start menu, under MSYS2 64bit, you have
 * `MSYS2 MinGW 32-bit`   <--- for gcc 32 bits toolchain
 * `MSYS2 MinGW 64-bit`   <--- for gcc 64 bits toolchain
 * `MSYS2 MSYS`  <-- for development of MSYS2 binaries, do not select this one
 * etc...

if you select `MSYS2 MinGW 64-bit`, a terminal is opened, then:
 * `uname -a` reports `MINGW64_NT-10.0-19043 DESKTOP-R5I7FJ7 3.3.5-341.x86_64 2022-07-04 21:33 UTC x86_64 Msys`
 * `gcc --version` reports `gcc.exe (Rev3, Built by MSYS2 project) 12.1.0` etc...

so you normally don't have to set any environment variables for the compiler.

vtorri:
Note that I have also checked with VS 2022 Community with cl.

you can also see the public symbols of the DLL with [dependency walker](https://dependencywalker.com/)

projectitis:
@vtorri Thanks for the tips. That was a very painless experience :) U have been able to compile with gcc. There are lots of warnings unrelated to your PR that can be addressed separately, but as far as your changes go they work fine with all compilers I have tried.

projectitis (approved):

hermet (approved):
thanks everyone.

PR #1245, opened 2022-08-20
Draft - Paint: Blending modes implementation

projectitis:
@hermet @mmaciola I started from mmaciola's implementation, but made some changes.

The biggest change is that the blending mode is now added to the surface as a function pointer.

This implementation should be more performant because the switch statement has been removed for every pixel being drawn.  The blending function pointer is only set one time before the paint is rendered.

This is just a draft implementation - it builds and runs, but the resulting blending is not correct.

I would really appreciate your help and pointers on what I have done incorrectly!

Please refer to #401

### DIFF
--- a/src/lib/sw_engine/tvgSwCommon.h
+++ b/src/lib/sw_engine/tvgSwCommon.h
@@ -236,6 +236,7 @@ struct SwBlender
 {
     uint32_t (*join)(uint8_t r, uint8_t g, uint8_t b, uint8_t a);
     uint32_t (*lumaValue)(uint32_t c);
+    uint32_t (*blend)(uint32_t src, uint32_t dst, uint8_t alpha, uint8_t ialpha, uint32_t src_blended);
 };
 
 struct SwCompositor;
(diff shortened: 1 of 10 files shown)


src/lib/sw_engine/tvgSwCommon.h
```diff
@@ -347,6 +370,19 @@ void mpoolRetOutline(SwMpool* mpool, unsigned idx);
...
+uint32_t blendMultiply(uint32_t src, uint32_t dst, uint8_t alpha, uint8_t ialpha, uint32_t src_blended);
+uint32_t blendOverlay(uint32_t src, uint32_t dst, uint8_t alpha, uint8_t ialpha, uint32_t src_blended);
+uint32_t blendDarken(uint32_t src, uint32_t dst, uint8_t alpha, uint8_t ialpha, uint32_t src_blended);
+uint32_t blendLighten(uint32_t src, uint32_t dst, uint8_t alpha, uint8_t ialpha, uint32_t src_blended);
+uint32_t blendColorDodge(uint32_t src, uint32_t dst, uint8_t alpha, uint8_t ialpha, uint32_t src_blended);
+uint32_t blendColorBurn(uint32_t src, uint32_t dst, uint8_t alpha, uint8_t ialpha, uint32_t src_blended);
+uint32_t blendHardLight(uint32_t src, uint32_t dst, uint8_t alpha, uint8_t ialpha, uint32_t src_blended);
+uint32_t blendSoftLight(uint32_t src, uint32_t dst, uint8_t alpha, uint8_t ialpha, uint32_t src_blended);
+uint32_t blendDifference(uint32_t src, uint32_t dst, uint8_t alpha, uint8_t ialpha, uint32_t src_blended);
+uint32_t blendExclusion(uint32_t src, uint32_t dst, uint8_t alpha, uint8_t ialpha, uint32_t src_blended);
```
projectitis, line 384: These are the blending functions that can be set on the surface->blender

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -167,7 +167,7 @@ static bool _rasterMaskedRect(SwSurface* surface, const SwBBox& region, uint32_t
         auto cmp = &cbuffer[y * surface->stride];
         for (uint32_t x = 0; x < w; ++x, ++dst, ++cmp) {
             auto tmp = ALPHA_BLEND(color, blendMethod(*cmp));
-            *dst = tmp + ALPHA_BLEND(*dst, _ialpha(tmp));
+            *dst = surface->blender.blend(color, *dst, _alpha(tmp), _ialpha(tmp), tmp);
```
projectitis, line 170: This is the file where I need the most help :)

src/lib/sw_engine/tvgSwRenderer.cpp
```diff
@@ -580,6 +580,51 @@ bool SwRenderer::dispose(RenderData data)
...
+        case BlendingMode::Difference:
+            this->surface->blender.blend = &blendDifference;
+            break;
+        case BlendingMode::Exclusion:
+            this->surface->blender.blend = &blendExclusion;
+            break;
+    }
+}
+
+
```
projectitis, line 627: This is where the blending function is assigned to the surface.

src/lib/sw_engine/tvgSwRaster.cpp
hermet: Could you please elaborate this "help" comment?

hermet:
replace with this: https://github.com/thorvg/thorvg/pull/1500

PR #1246, opened 2022-08-20
fix compilation of svg2png on Windows

vtorri:


vtorri:
@JSUYA any news on this ?

JSUYA:
> @JSUYA any news on this ?

Sorry, I'm late. I haven't tested this code yet. (I am currently unable to test windows build...)

Could you explain further why this change is needed?
Any problems using opendir on WIN32? I don't know much about this.

vtorri:
@JSUYA easy:
 * with mingw : opendir() exists but all the DT_* enums are not implemented. This implementation is lacking a lot of stuff
 * with Visual Studio, it does not exist :-)

to test a Windows build, several solutions:
 * on Windows, use MSYS2 + mingw-w64 (same than a POSIX env like linux)
 * on Windows, install Visual Studio, run meson 0.63 with Visual Studio backend and compile with Visual Studio
 * on Linux, cross-sompile to Windows and use Wine

vtorri:
or better, use CI for this. You should enable tools and examples with msys2 and Visual Studio. Then you will see if it compiles or not. If it does not compile, fix the code :-) And you can also run commands in CI, so it is easy to test svg2png within CI

vtorri:
I can improve the code by using (line 333 of the commit) `FindFirstFileEx((path + "\\*.svg")` as it will select only files with the svg extension, and not all the files.

src/bin/svg2png/svg2png.cpp
```diff
@@ -299,15 +325,42 @@ struct App
...
     {
         //open directory
-        if (!dir) {
-            dir = opendir(path.c_str());
-            if (!dir) {
-                cout << "Couldn't open directory \"" << path.c_str() << "\"." << endl;
-                return 1;
+#ifdef _WIN32
+        WIN32_FIND_DATA fd;
+        HANDLE h = FindFirstFileEx((path + "\\*").c_str(), FindExInfoBasic, &fd, FindExSearchNameMatch, NULL, 0);
```
JSUYA, line 333: just I have a question because I'm curious.
Is there any reason to use `FindFirstFileEx`? (instead of `FindFirstFile`)

JSUYA (commented):
> I can improve the code by using (line 333 of the commit) `FindFirstFileEx((path + "\\*.svg")` as it will select only files with the svg extension, and not all the files.

Line 279 is already checking the extension.

@hermet  Would you like to review it together?
build setup is not easy on me...

src/bin/svg2png/svg2png.cpp
vtorri: yes: by default, FindFirstFile() search for files with also the old DOS file names (8 chars for file name + 3 for extension). Nowadays, there is no such limitation on the files. To remove this search, I use FindFirstFileEx() with the option FindExInfoBasic. Hence the search is faster. See this [link](https://docs.microsoft.com/en-us/windows/win32/api/minwinbase/ne-minwinbase-findex_info_levels)

vtorri:
> > I can improve the code by using (line 333 of the commit) `FindFirstFileEx((path + "\\*.svg")` as it will select only files with the svg extension, and not all the files.
> 
> Line 279 is already checking the extension.

anyway, the search will be faster.

src/bin/svg2png/svg2png.cpp
vtorri: notealso that i benchmarked this some years ago to verify this. Indeed it is faster.

hermet:
@vtorri please tell me if you need a write-access. I believe you can drive the window port of thorvg.

hermet (approved):

hermet:
@vtorri it would be better if you describe the commit message more detail. anyway thanks! :)

Issue #1239, opened 2022-08-19
SVGs with no viewBox and width&height attribs


mgrudzinska:
I need to read what the expected behavior is, but for sure something has to be rendered (there is not for now)
pointed out in https://github.com/godotengine/godot/pull/64530

bruvzg:
Some extra info:

Here are a few examples of the SVG code extracted from fonts, it's either missing `viewBox` completely or it's not enclosing the image:

`🍏` from Twitter Color Emoji font:

```svg
<svg xmlns="http://www.w3.org/2000/svg" id="glyph318"><g transform="translate(0 -6.75) translate(0,-1638.4) scale(56.888888888888886)"><path fill="#77B255" d="M24 7c-3 0-3 1-6 1s-3-1-6-1c-4 0-9 2-9 9 0 11 6 20 10 20 3 0 3-1 5-1s2 1 5 1c4 0 10-9 10-20 0-7.001-5-9-9-9z" /><path fill="#3E721D" d="M17.311 7.88s-1.775-4.674-6.58-6.06c-3.843-1.108-6.318.26-6.318.26s3.012 3.991 5.895 4.822c2.882.83 7.003.978 7.003.978z" /><path fill="#662113" d="M18 10a1 1 0 0 1-1-1c0-3.441 1.2-6.615 3.293-8.707a.999.999 0 1 1 1.414 1.414C19.986 3.427 19 6.085 19 9a1 1 0 0 1-1 1z" /></g></svg>
```

<img width="200" alt="Screenshot 2022-08-20 at 10 20 18" src="https://user-images.githubusercontent.com/7645683/185733991-f0a49ce8-401c-47cb-9de9-04e9b5dba66c.png">


`🍏` from EmojiOne Color font:

```svg
<?xml version='1.0' encoding='utf8'?>
[...]

PR #1248, opened 2022-08-21
svg_loader: handling svgs without any viewPort and viewBox information

mgrudzinska:
issue #1239

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -787,20 +787,27 @@ static bool _attrParseSvgNode(void* data, const char* key, const char* value)
 
     if (!strcmp(key, "width")) {
         doc->w = _toFloat(loader->svgParse, value, SvgParserLengthType::Horizontal);
+        doc->bboxFlag |= BBoxFlag::Width;
     } else if (!strcmp(key, "height")) {
         doc->h = _toFloat(loader->svgParse, value, SvgParserLengthType::Vertical);
+        doc->bboxFlag |= BBoxFlag::Height;
     } else if (!strcmp(key, "viewBox")) {
         if (_parseNumber(&value, &doc->vx)) {
             if (_parseNumber(&value, &doc->vy)) {
                 if (_parseNumber(&value, &doc->vw)) {
-                    _parseNumber(&value, &doc->vh);
-                    loader->svgParse->global.h = (uint32_t)doc->vh;
+                    if (_parseNumber(&value, &doc->vh)) {
+                        doc->bboxFlag |= BBoxFlag::ViewboxH;
+                        loader->svgParse->global.h = (uint32_t)doc->vh;
+                    }
+                    doc->bboxFlag |= BBoxFlag::ViewboxW;
+                    loader->svgParse->global.w = (uint32_t)doc->vw;
                 }
-                loader->svgParse->global.w = (uint32_t)doc->vw;
+                doc->bboxFlag |= BBoxFlag::ViewboxY;
+                loader->svgParse->global.y = (int)doc->vy;
             }
-            loader->svgParse->global.y = (int)doc->vy;
+            doc->bboxFlag |= BBoxFlag::ViewboxX;
+            loader->svgParse->global.x = (int)doc->vx;
         }
-        loader->svgParse->global.x = (int)doc->vx;
     } else if (!strcmp(key, "preserveAspectRatio")) {
         if (!strcmp(value, "none")) doc->preserveAspect = false;
     } else if (!strcmp(key, "style")) {
@@ -814,7 +821,6 @@ static bool _attrParseSvgNode(void* data, const char* key, const char* value)
     }
     return true;
 }
-
 
 //https://www.w3.org/TR/SVGTiny12/painting.html#SpecifyingPaint
 static void _handlePaintAttr(SvgPaint* paint, const char* value)
@@ -1249,6 +1255,7 @@ static SvgNode* _createSvgNode(SvgLoaderData* loader, SvgNode* parent, const cha
     loader->svgParse->global.h = 0;
 
     doc->preserveAspect = true;
+    doc->bboxFlag = BBoxFlag::None;
     func(buf, bufLength, _attrParseSvgNode, loader);
 
     if (loader->svgParse->global.w == 0) {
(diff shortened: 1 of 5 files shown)


mgrudzinska:
I'd appreciate any comment on that. It's a strange case - no viewBox and no viewPort info...
I read the svg standard, but still I'm not sure how to handle the situation, when width/eight are given in percentages. Each online viewer that I used shows different results....

1) from https://www.w3.org/TR/SVG2/geometry.html#Sizing:
"The value auto for width and height on the svg element is treated as 100%."

2) from (for css) https://svgwg.org/specs/integration/#svg-css-sizing:
"When SVG element is the top level element, the [width](https://svgwg.org/svg2-draft/geometry.html#Sizing) and [height](https://svgwg.org/svg2-draft/geometry.html#Sizing) are set to '100%'
To resolve 'auto' value on ‘[svg](https://svgwg.org/svg2-draft/struct.html#SVGElement)’ element if the ‘[viewBox](https://svgwg.org/svg2-draft/coords.html#ViewBoxAttribute)’ attribute is not specified:

 * Use author specified [width](https://svgwg.org/svg2-draft/geometry.html#Sizing) and [height](https://svgwg.org/svg2-draft/geometry.html#Sizing) ‘[svg](https://svgwg.org/svg2-draft/struct.html#SVGElement)’ element attributes.
 * If any of the sizing attributes are missing, resolve the missing ‘[svg](https://svgwg.org/svg2-draft/struct.html#SVGElement)’ element [width](https://svgwg.org/svg2-draft/geometry.html#Sizing) to '300px' and missing [height](https://svgwg.org/svg2-draft/geometry.html#Sizing) to '150px' (using [CSS 2.1 replaced elements size calculation](https://www.w3.org/TR/CSS21/visudet.html#inline-replaced-width)).
To resolve 'auto' value on ‘[svg](https://svgwg.org/svg2-draft/struct.html#SVGElement)’ element if the ‘[viewBox]"

3) from https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/width#svg
<html>
<body>


Default value | auto (treated as 100%)
-- | --



</body>
</html>

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -145,6 +150,20 @@ enum class SvgParserLengthType
...
 
+enum BBoxFlag
+{
+    None = 0,
+    Width = 1,
+    Height = 2,
+    ViewboxX = 4,
+    ViewboxY = 8,
+    ViewboxW = 16,
+    ViewboxH = 32,
```
JSUYA, line 161: How about displaying as 0x1, 0x2 referring to `enum class SvgStyleFlags`?

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3161,6 +3176,14 @@ bool SvgLoader::header()
         }
 
         preserveAspect = loaderData.doc->node.doc.preserveAspect;
+
+        if ((loaderData.doc->node.doc.bboxFlag & BBoxFlag::All) == BBoxFlag::All ||
+            (loaderData.doc->node.doc.bboxFlag & BBoxFlag::Viewbox) == BBoxFlag::Viewbox ||
+            (loaderData.doc->node.doc.bboxFlag & BBoxFlag::WidthHeight) == BBoxFlag::WidthHeight) {
+            if (vw <= 0 || vh <= 0)
+                return false;
```
JSUYA, line 3184: The bool return of header() determines whether the svg image load failed. (valid, invalid)
If vw and vh are 0, it will be invisible, but I don't think it's a load failure. What do you think?
If you don't necessarily have to return false, how about printing a LOG?

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3161,6 +3176,14 @@ bool SvgLoader::header()
         }
 
         preserveAspect = loaderData.doc->node.doc.preserveAspect;
+
+        if ((loaderData.doc->node.doc.bboxFlag & BBoxFlag::All) == BBoxFlag::All ||
+            (loaderData.doc->node.doc.bboxFlag & BBoxFlag::Viewbox) == BBoxFlag::Viewbox ||
+            (loaderData.doc->node.doc.bboxFlag & BBoxFlag::WidthHeight) == BBoxFlag::WidthHeight) {
```
JSUYA, line 3182: I haven't read all the svg specs, but I think we should consider the case where only one of width or height is declared.

```html
<?xml version='1.0' encoding='utf8'?>
<svg xmlns="http://www.w3.org/2000/svg" id="glyph310" width="64"><g transform="translate(0 -9.6) translate(-0.0 -104.8576) scale(2.048)"><path d="M31.4 4.9c1.1 2.5-.5 10.9-.5 10.9s-7.5-4.4-8.6-6.9c-1.1-2.5 0-5.4 2.5-6.5 2.5-1.1 5.5 0 6.6 2.5" fill="#64892f" /><path d="M33.3 16.3h-2.6c0-4.3 2.4-11.6 12-11.6v2.6c-8.7 0-9.4 8.2-9.4 9" fill="#947151" /><path d="M32 15.8C24.7 8.4 3 8.1 3 29.3 3 42.7 18.8 62 25.7 62c3.1 0 4.2-2 6.1-2 1.8 0 2.4 1.9 6.4 1.9C45.1 62 61 42.8 61 29.3 61 8.1 39.3 8.4 32 15.8z" fill="#8cc63e" /></g></svg>
```
(tested in chrome)

I don't think `BBoxFlag::WidthHeight` or `BBoxFlag::Viewbox` are necessary in that sense.

JSUYA (commented):
long time no see :)

oh.. I didn't understand 100% of this spec right now. It seems to me that this PR is not very different from the spec.
I don't think there is a problem because the viewbox or width and height are replaced with bounds and percentage unit refers to it. Can you explain the problem a bit more?

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: yes, you are right, I'll change it

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: from [specs ](https://www.w3.org/TR/SVG2/coords.html#ViewBoxAttribute) regarding the viewBox attrib: "A negative value for <width> or <height> is an error and invalidates the ‘viewBox’ attribute. A value of zero disables rendering of the element."
so for values '<0' 'return false' is correct. 
I've included the '=0' case, to stop calling all of the following functions, but you are right, I'll change it

mgrudzinska:
> long time no see :)
> 
yes, nice to hear from you after such a long time :) 
and thx for such a quick rev!

mgrudzinska:
moved to #1318

mgrudzinska:
I'll close this pr after all comments are handled

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: I was wrong, negative values invalidate the 'viewbox' attrib and not the whole svg

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: done

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: done in #1318

mgrudzinska:
all rev comments applied in #1318

Issue #954, opened 2021-10-27
sw_engine: stroke borders are miscalculated when transformation is applied


mgrudzinska:
current result:
![41before](https://user-images.githubusercontent.com/67589014/138977576-e294edae-9a18-4863-b426-1c1c47c102de.PNG)

expected:
<img width="611" alt="Zrzut ekranu 2021-10-27 o 02 01 25" src="https://user-images.githubusercontent.com/67589014/138977972-b55ba350-a346-4fd7-ace6-49170e0fbaf4.png">

mgrudzinska:
check:
```
<svg xmlns="http://www.w3.org/2000/svg" height="206.25000" width="406.25000">
  <defs id="defs3">
    <linearGradient id="linearGradient2143">
      <stop id="stop2145" offset="0.00000000" style="stop-color: rgb(0, 0, 0); stop-opacity: 1;"/>
      <stop id="stop2147" offset="1.0000000" style="stop-color: rgb(255, 255, 255); stop-opacity: 1;"/>
    </linearGradient>

    <linearGradient inkscape:collect="always" xlink:href="#linearGradient2143" id="linearGradient2857" gradientUnits="userSpaceOnUse" gradientTransform="matrix(1.56589, 0, 0, 0.732797, 331.41, -147.922)" x1="-327.88300" y1="220.74342" x2="-327.78314" y2="233.92133"/>
  </defs>

  <path d="M 0 0 h 400 v 400 h -400 z" fill="blue"/>
[...]

PR #1249, opened 2022-08-21
svg_loader: preserveAspectRatio attrib handled according to the svg standard

mgrudzinska:
issue #1181 

sample file to test the bug:
```
<svg viewBox="0 0 15 15" xmlns="http://www.w3.org/2000/svg">
  <symbol id="myDot" width="10" height="5" viewBox="0.5 -1 4 4" preserveAspectRatio="none">
    <circle cx="0.1" cy="1" r="2" />
  </symbol>

  <use href="#myDot" x="0" y="5" height="4" style="opacity:0.8" />
  <use href="#myDot" x="5" y="5" width="5" style="opacity:0.6" />
</svg>
```

before:
before

after:
after

to test the aspect ratio:
```
<svg id="svg1" width="200" height="100" viewBox="0 0 200 200" preserveAspectRatio="xMinYMin slice"
     xmlns="http://www.w3.org/2000/svg">
    <rect id="rect1" x="70" y="20" width="60" height="60" fill="green"/>
    <rect id="frame" x="1" y="1" width="198" height="98" fill="none" stroke="black"/>
</svg>

<svg id="svg1" width="200" height="100" viewBox="0 0 200 200" preserveAspectRatio="none"
     xmlns="http://www.w3.org/2000/svg">
    <rect id="rect1" x="70" y="20" width="60" height="60" fill="green"/>
    <rect id="frame" x="1" y="1" width="198" height="98" fill="none" stroke="black"/>
</svg>

<svg id="svg1" width="200" height="100" viewBox="0 0 200 200" preserveAspectRatio="xMinYMin"
     xmlns="http://www.w3.org/2000/svg">
    <rect id="rect1" x="70" y="20" width="60" height="60" fill="green"/>
    <rect id="frame" x="1" y="1" width="198" height="98" fill="none" stroke="black"/>
</svg>

<svg id="svg1" width="200" height="100" viewBox="0 0 200 200" preserveAspectRatio="xMidYMid  slice"
     xmlns="http://www.w3.org/2000/svg">
    <rect id="rect1" x="70" y="20" width="60" height="60" fill="green"/>
    <rect id="frame" x="1" y="1" width="198" height="98" fill="none" stroke="black"/>
</svg>

<svg id="svg1" width="200" height="100" viewBox="-20 -20 200 200" preserveAspectRatio="xMaxYMax slice"
     xmlns="http://www.w3.org/2000/svg">
    <rect id="rect1" x="70" y="20" width="60" height="60" fill="green"/>
    <rect id="frame" x="1" y="1" width="198" height="98" fill="none" stroke="black"/>
</svg>

<svg id="svg1" width="200" height="100" viewBox="0 0 200 200" preserveAspectRatio="xMidYMid"
     xmlns="http://www.w3.org/2000/svg">
    <rect id="rect1" x="70" y="20" width="60" height="60" fill="green"/>
    <rect id="frame" x="1" y="1" width="198" height="98" fill="none" stroke="black"/>
</svg>

<svg id="svg1" width="200" height="100" viewBox="-50 -100 200 200" preserveAspectRatio="xMinYMin"
     xmlns="http://www.w3.org/2000/svg">
    <rect id="rect1" x="70" y="20" width="60" height="60" fill="green"/>
    <rect id="frame" x="1" y="1" width="198" height="98" fill="none" stroke="black"/>
</svg>

<svg id="svg1" width="200" height="100" viewBox="0 0 200 200" preserveAspectRatio="xMaxYMax"
     xmlns="http://www.w3.org/2000/svg">
    <rect id="rect1" x="70" y="20" width="60" height="60" fill="green"/>
    <rect id="frame" x="1" y="1" width="198" height="98" fill="none" stroke="black"/>
</svg>

<svg id="svg1" width="200" height="100" viewBox="-10 -90 200 200" preserveAspectRatio="xMinYMin slice"
     xmlns="http://www.w3.org/2000/svg">
    <rect id="rect1" x="70" y="20" width="60" height="60" fill="green"/>
    <rect id="frame" x="1" y="1" width="198" height="98" fill="none" stroke="black"/>
</svg>
```

before:
before_asp

after:
after_asp

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -115,8 +115,52 @@ static bool _parseNumber(const char** content, float* number)
     if ((*content) == end) return false;
     //Skip comma if any
     *content = _skipComma(end);
+
     return true;
 }
+
+
+static constexpr struct
+{
+    AspectRatioAlign align;
+    const char* tag;
+} alignTags[] = {
+    { AspectRatioAlign::XMinYMin, "xMinYMin" },
+    { AspectRatioAlign::XMidYMin, "xMidYMin" },
+    { AspectRatioAlign::XMaxYMin, "xMaxYMin" },
+    { AspectRatioAlign::XMinYMid, "xMinYMid" },
+    { AspectRatioAlign::XMidYMid, "xMidYMid" },
+    { AspectRatioAlign::XMaxYMid, "xMaxYMid" },
+    { AspectRatioAlign::XMinYMax, "xMinYMax" },
+    { AspectRatioAlign::XMidYMax, "xMidYMax" },
+    { AspectRatioAlign::XMaxYMax, "xMaxYMax" },
+};
+
+
+static bool _parseAspectRatio(const char** content, AspectRatioAlign* align, AspectRatioMeetOrSlice* meetOrSlice)
+{
+    if (!strcmp(*content, "none")) {
+        *align = AspectRatioAlign::None;
+        return true;
+    }
+
+    for (unsigned int i = 0; i < sizeof(alignTags) / sizeof(alignTags[0]); i++) {
+        if (!strncmp(*content, alignTags[i].tag, 8)) {
+            *align = alignTags[i].align;
+            *content += 8;
+            *content = _skipSpace(*content, nullptr);
+            break;
+        }
+    }
+
+    if (!strcmp(*content, "meet"))
+        *meetOrSlice = AspectRatioMeetOrSlice::Meet;
+    else if (!strcmp(*content, "slice"))
+        *meetOrSlice = AspectRatioMeetOrSlice::Slice;
+
+    return true;
+}
+
 
 /**
  * According to https://www.w3.org/TR/SVG/coords.html#Units
@@ -802,7 +846,7 @@ static bool _attrParseSvgNode(void* data, const char* key, const char* value)
         }
         loader->svgParse->global.x = (int)doc->vx;
     } else if (!strcmp(key, "preserveAspectRatio")) {
-        if (!strcmp(value, "none")) doc->preserveAspect = false;
+        _parseAspectRatio(&value, &doc->align, &doc->meetOrSlice);
     } else if (!strcmp(key, "style")) {
         return simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
 #ifdef THORVG_LOG_ENABLED
@@ -1156,7 +1200,7 @@ static bool _attrParseSymbolNode(void* data, const char* key, const char* value)
         symbol->h = _toFloat(loader->svgParse, value, SvgParserLengthType::Vertical);
         symbol->hasHeight = true;
     } else if (!strcmp(key, "preserveAspectRatio")) {
-        if (!strcmp(value, "none")) symbol->preserveAspect = false;
+        _parseAspectRatio(&value, &symbol->align, &symbol->meetOrSlice);
     } else if (!strcmp(key, "overflow")) {
         if (!strcmp(value, "visible")) symbol->overflowVisible = true;
     } else {
@@ -1248,7 +1292,8 @@ static SvgNode* _createSvgNode(SvgLoaderData* loader, SvgNode* parent, const cha
     loader->svgParse->global.w = 0;
     loader->svgParse->global.h = 0;
 
-    doc->preserveAspect = true;
+    doc->align = AspectRatioAlign::XMidYMid;
+    doc->meetOrSlice = AspectRatioMeetOrSlice::Meet;
     func(buf, bufLength, _attrParseSvgNode, loader);
 
     if (loader->svgParse->global.w == 0) {
@@ -1309,7 +1354,8 @@ static SvgNode* _createSymbolNode(SvgLoaderData* loader, SvgNode* parent, const 
     if (!loader->svgParse->node) return nullptr;
 
     loader->svgParse->node->display = false;
-    loader->svgParse->node->node.symbol.preserveAspect = true;
+    loader->svgParse->node->node.symbol.align = AspectRatioAlign::XMidYMid;
+    loader->svgParse->node->node.symbol.meetOrSlice = AspectRatioMeetOrSlice::Meet;
     loader->svgParse->node->node.symbol.overflowVisible = false;
 
     loader->svgParse->node->node.symbol.hasViewBox = false;
@@ -3127,7 +3173,7 @@ void SvgLoader::run(unsigned tid)
 
         _updateStyle(loaderData.doc, nullptr);
     }
-    root = svgSceneBuild(loaderData.doc, vx, vy, vw, vh, w, h, preserveAspect, svgPath);
+    root = svgSceneBuild(loaderData.doc, vx, vy, vw, vh, w, h, align, meetOrSlice, svgPath);
 }
 
 
@@ -3160,7 +3206,8 @@ bool SvgLoader::header()
             if (vh < FLT_EPSILON) vh = h;
         }
 
-        preserveAspect = loaderData.doc->node.doc.preserveAspect;
+        align = loaderData.doc->node.doc.align;
+        meetOrSlice = loaderData.doc->node.doc.meetOrSlice;
     } else {
         TVGLOG("SVG", "No SVG File. There is no <svg/>");
         return false;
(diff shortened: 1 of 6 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3216,30 +3263,19 @@ bool SvgLoader::resize(Paint* paint, float w, float h)
...
-        if (tw > th) ty -= (h - th) * 0.5f;
-        else tx -= (w - tw) * 0.5f;
-        paint->translate(-tx, -ty);
-    } else {
-        //Align
-        auto tx = 0.0f;
-        auto ty = 0.0f;
-        auto tw = this->w * sx;
-        auto th = this->h * sy;
-        if (tw > th) ty -= (h - th) * 0.5f;
```
mgrudzinska, line 3237: @JSUYA I can't find any explanation why we had this condition.. do u have some idea?

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -585,20 +659,8 @@ static unique_ptr<Scene> _useBuildHelper(const SvgNode* node, const Box& vBox, c
...
-                if (sx < sy) sy = sx;
-                else sx = sy;
-            }
-
-            auto tvx = symbol.vx * sx;
-            auto tvy = symbol.vy * sy;
-            auto tvw = vw * sx;
-            auto tvh = vh * sy;
-            tvy -= (symbol.h - tvh) * 0.5f;
-            tvx -= (symbol.w - tvw) * 0.5f;
```
mgrudzinska, line 600: here was a bug - these should be width and height instead of symbol.w and symbol.h

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: It is expected that the code added for center-aligned scale in 20de2bfc153252531b8186972f2d1a36de240474

mgrudzinska: I see, but it was an old logic, when we kinda mixed in this function the general concept of resizing and the transformations between the viewPort and the viewBox.
 
Now, I think, the logic should be that the preserveAspectRatio is take into account only while the scene is built. After that, we have a picture with some width and height. and if we want to squeeze this picture, we should be able to do it, regardless of the svg's aspect ratio. if we want to keep the proportion, we can use scale instead of resize.  

@hermet could you please comment on that? 
I'd keep only these two lines from this function:
```
        Matrix m = {sx, 0, 0, 0, sy, 0, 0, 0, 1};
        paint->transform(m);
```

src/loaders/svg/tvgSvgLoader.h
```diff
@@ -32,6 +32,8 @@ class SvgLoader : public LoadModule, public Task
     string svgPath = "";
     const char* content = nullptr;
     uint32_t size = 0;
+    AspectRatioAlign align = AspectRatioAlign::XMidYMid;
+    AspectRatioMeetOrSlice meetOrSlice = AspectRatioMeetOrSlice::Meet;
```
JSUYA, line 36: Is this public member really necessary?
doc's align and meetOrslice are not changed, and the loaderData can be accessed wherever they are used.

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: `xMidYMid` is set to default, so we can change it according to your plan.

JSUYA (approved):
Overall I think it's good.
I left a few small comments. Please merge after check.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -115,9 +115,53 @@ static bool _parseNumber(const char** content, float* number)
...
+    { AspectRatioAlign::XMinYMid, "xMinYMid" },
+    { AspectRatioAlign::XMidYMid, "xMidYMid" },
+    { AspectRatioAlign::XMaxYMid, "xMaxYMid" },
+    { AspectRatioAlign::XMinYMax, "xMinYMax" },
+    { AspectRatioAlign::XMidYMax, "xMidYMax" },
+    { AspectRatioAlign::XMaxYMax, "xMaxYMax" },
+};
+
+
+static bool _parseAspectRatio(const char** content, AspectRatioAlign* align, AspectRatioMeetOrSlice* meetOrSlice)
```
hermet, line 140: Suggestion:  AspectRatioAlign _parseAspectRatio(const char** content, AspectRatioMeetOrSlice* meetOrSlice)
Here return value is useless?

src/loaders/svg/tvgSvgLoader.cpp
hermet: @mgrudzinska I think we can fix it if it's necessary but please double-check our examples whether the result is different after this.

mgrudzinska: thx for your comments, the #1258 shows, that the resizing shouldn't have anything to do with the aspect ratio. I've changed it

src/loaders/svg/tvgSvgLoader.h
mgrudzinska: changed

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: 1. I'm not a fun of returning one value and the other one changing by passing it by ref/pointer. Now both are returned in the same way.
2. I think that this function should take care of all the text passed as the preserveAspectRatio value. so if in some svg file we get ```preserveAspectRatio="none some_additional_text"``` the return value should be false. So for now it's not usefull, but in the future, we could use the returned bool. That's why I didn;t use void.

mgrudzinska:
@hermet I answered on your comments, if you agree please merge, otherwise please leave some comment, I'll apply your suggestions.

hermet:
> @hermet I answered on your comments, if you agree please merge, otherwise please leave some comment, I'll apply your suggestions.

@mgrudzinska  thanks, so this may occur the compatibility issue... we need to check this considerably.
Do you have any suggestion to fix the broken result in examples? at least, we need to think about a guide-line to fix them user side...

mgrudzinska:
@hermet the edited part of the Svg.cpp
```
    //image scaling preserving its aspect ratio 
    float scale;
    float shiftX = 0.0f, shiftY = 0.0f;
    float w, h;
    picture->size(&w, &h);
    if (w > h) {
        scale = SIZE / w;
        shiftY = (SIZE - h * scale) * 0.5f;
    } else {
        scale = SIZE / h;
        shiftX = (SIZE - w * scale) * 0.5f;
    }

    picture->scale(scale);
    picture->translate((count % NUM_PER_ROW) * SIZE + shiftX, (count / NUM_PER_ROW) * (HEIGHT / NUM_PER_COL) + shiftY);
```

it's up to the user, whether to preserve the image's aspect ratio or whether to center the image. If we want that in our example, we can apply the above. 
Please let me know whether it's what you meant.
If so I need to apply the changes to the resize() function in the tvg loader.

hermet:
@fire @qarmin it would be better if someone could double-check this patch doesn't break the godot project.

qarmin:
I only see a little problem with this image - [BB.zip](https://github.com/Samsung/thorvg/files/9464887/BB.zip) (500px x 500px)
Screenshot from 2022-08-31 22-42-56

mgrudzinska:
@qarmin and how would you expect this image will look like?

the results you showed are exactly as I expected, after this pr I mean

mgrudzinska:
this image is not squared and you've forced it to be, that's why it's stretched. 
so I'd say, that you have to use scale instead of size, if you want to preserve it's aspect ratio. The problem is that there is not such an option when the svg2png is used?

qarmin:
In rsvg, when I forced certain size I got different result:

Screenshot from 2022-08-31 23-22-47

Screenshot from 2022-08-31 23-25-35

[BB.zip](https://github.com/Samsung/thorvg/files/9465216/BB.zip)
[CC.zip](https://github.com/Samsung/thorvg/files/9465220/CC.zip)

mgrudzinska:
@qarmin ok, now I see the problem :) thx!
I don;t think it has something to do with this pr, but I'll investigate and confirm

mgrudzinska:
@qarmin this problem is caused by some other reasons: issue #954 

we don't need to scale this svg file to observe this issue:
```
<svg height="16" viewBox="0 0 16 16" width="16" xmlns="http://www.w3.org/2000/svg">
<path d="m2 2 4 6-4 6" fill="blue" stroke="#ff0000"  stroke-width="0.3" transform="matrix(2 0 0 1 0 0)"/>
</svg>
```

@hermet so beside the fact, that maybe at some point it'd be good to take care of #954, I think that the examples showed by qarmin confirm, that this pr is correct

qarmin (approved):

src/loaders/svg/tvgSvgLoader.cpp
hermet: In the future when? I believe we can replace this in the future *when it's really necessary*. that's our policy.

Issue #1257, opened 2022-08-26
Memory leaks when converting svg to png


qarmin:
Example files
[AAA.zip](https://github.com/Samsung/thorvg/files/9435682/AAA.zip)

Memory leaks:
[Memory leaks.txt](https://github.com/Samsung/thorvg/files/9435683/Memory.leaks.txt)

Example leak
```
Direct leak of 11 byte(s) in 1 object(s) allocated from:
    #0 0x7f4ea328c9a7 in __interceptor_strdup ../../../../src/libsanitizer/asan/asan_interceptors.cpp:454
    #1 0x7f4ea2f91c06 in _styleCopy ../src/loaders/svg/tvgSvgLoader.cpp:1945
    #2 0x7f4ea2f91c06 in _copyAttr ../src/loaders/svg/tvgSvgLoader.cpp:1993
    #3 0x7f4ea2f91c06 in _cloneNode ../src/loaders/svg/tvgSvgLoader.cpp:2082
    #4 0x7f4ea2f91c06 in _cloneNode ../src/loaders/svg/tvgSvgLoader.cpp:2066
    #5 0x7f4ea2f94ef0 in _attrParseUseNode ../src/loaders/svg/tvgSvgLoader.cpp:2146
    #6 0x7f4ea2fb492d in simpleXmlParseAttributes(char const*, unsigned int, bool (*)(void*, char const*, char const*), void const*) ../src/loaders/svg/tvgXmlParser.cpp:361
    #7 0x7f4ea2f911ad in _createUseNode ../src/loaders/svg/tvgSvgLoader.cpp:2171
    #8 0x7f4ea2f9735f in _svgLoaderParserXmlOpen ../src/loaders/svg/tvgSvgLoader.cpp:2714
    #9 0x7f4ea2f97c8d in _svgLoaderParser ../src/loaders/svg/tvgSvgLoader.cpp:2792
    #10 0x7f4ea2fb4f29 in simpleXmlParse(char const*, unsigned int, bool, bool (*)(void*, SimpleXMLType, char const*, unsigned int), void const*) ../src/loaders/svg/tvgXmlParser.cpp:429
    #11 0x7f4ea2f9994e in SvgLoader::run(unsigned int) ../src/loaders/svg/tvgSvgLoader.cpp:3112
    #12 0x7f4ea2f00fdd in tvg::Task::operator()(unsigned int) ../src/lib/tvgTaskScheduler.h:68
[...]

PR #1260, opened 2022-08-28
svg_loader: only the first css style node is interpreted

mgrudzinska:
Since the css id selector is not supported in TVG, only the first
style node is taken into account.

issue #1257 (for ex ABCD_10191.svg)


JSUYA (approved):
LGTM

PR #1263, opened 2022-08-30
svg_loader: prevent mem leaks

mgrudzinska:
A necessary check added before strdup function is called.

svgs with mem leaks:
```
<svg viewBox="0 0 10 10" xmlns="http://www.w3.org/2000/svg"
  <path d="M 0 0 h 100 v 100 z" fill="red" d="M 0 0 h 100 v 100 h -100 z"/>
</svg>
```

```
  
  <clipPath id="myClip1">
    <circle cx="40" cy="35" r="75" />
  </clipPath>

  <clipPath id="myClip2">
    <circle cx="40" cy="35" r="55" />
  </clipPath>

 <rect x="5" y="90" width="50" height="50" class="mineStyle" clip-path="url(#myClip1)"/>


 <defs>
  <style>
    rect.mineStyle {
        fill: blue;
        stroke: #3399ff;
        clip-path: url(#myClip2);
    }
  </style>
 </defs>

</svg>
```

### DIFF
--- a/src/loaders/svg/tvgSvgCssStyle.cpp
+++ b/src/loaders/svg/tvgSvgCssStyle.cpp
@@ -128,8 +128,14 @@ void cssCopyStyleAttr(SvgNode* to, const SvgNode* from)
     //Copy style attribute
     _copyStyle(to->style, from->style);
 
-    if (from->style->clipPath.url) to->style->clipPath.url = strdup(from->style->clipPath.url);
-    if (from->style->mask.url) to->style->mask.url = strdup(from->style->mask.url);
+    if (from->style->clipPath.url) {
+        if (to->style->clipPath.url) free(to->style->clipPath.url);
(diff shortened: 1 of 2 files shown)


vtorri:
patch might be simplified as free(NULL) is a no-op

hermet (approved):

hermet:
there are too many duplicated code. maybe we can introduce the macro like this.
example:

SAFE_STRDUP(dst, src)
{
  if (src)
  {
    if (dst) free(dst)
    dst = strdup(x)
  } 
}

vtorri:
> there are too many duplicated code. maybe we can introduce the macro like this. example:
> 
> `SAFE_STRDUP(dst, src) { if (src) { if (dst) free(dst) dst = strdup(x) } }`

`if (dst)` is useless, just free

Issue #1255, opened 2022-08-26
Hundreads of possible broken SVG files


qarmin:
This files were found programmatically, so not all this images are invalid.

To check if png created from svg is valid I:
- Created png file from svg with inkscape (`inkscape --export-type=png -h 500`)
- Created png file from with svg2png (`svg2png -r 500x500`)
- Compared png files with each other

This is not ideal comparison, because for some files inkscape generate invalid png files.

Even if files looks in file browser preview similar, then I suggest to look at them closer, because probably they differ in shadows or other less visible things.

[SVG.zip](https://github.com/Samsung/thorvg/files/9434108/SVG.zip)
![Screenshot from 2022-08-26 16-25-58](https://user-images.githubusercontent.com/41945903/186930569-0c46657c-9054-42e0-9eb4-a539b6eccbe4.png)

mgrudzinska:
ABCD_297.svg:
before nothing was drawn, now (#1266) we have this - still not exactly as it should look like:
![aftt](https://user-images.githubusercontent.com/67589014/187802298-c2611743-1fbe-4e57-9819-6dfa86a46054.PNG)

so we are one step (from hundreds) closer to the perfect lib :D

qarmin:
New list with svg files that are not square and without any text - [NotText.zip](https://github.com/Samsung/thorvg/files/9479820/NotText.zip)

mgrudzinska:
why without ```<image>``` ? 
I'll take a look on each one of them later, but it seams that not all of them are really broken:
KOSZ_8786 from tvg:
![d](https://user-images.githubusercontent.com/67589014/232527804-469c2211-9769-4fa7-a12b-dd2f702b9155.png)

qarmin:
In most of the time I use compiled version of thorvg without images support(without any reason), later I may enable it to find more differences

KOSZ_8786 example is strange, because I used automatic generation for this image with command `svg2png file.svg -r 400x400` and now everything works fine - most of the time).

Looks that this files only draws partially due #1370 (In 1000 tested same files I got 3 AddressSanitizer crashes and also 3 same broken files were produced)

qarmin:
Broken files(png files which looks different than with rsvg) and problematic files(files which cause crashes or cannot be parsed by this tool) will now be automatically found by CI once per week - https://github.com/qarmin/SVG-regression-finder/actions

[...]

PR #1266, opened 2022-08-31
svg_loader: deeper search for postponed nodes

mgrudzinska:
Till now the proper node was searched only among children,
now all the nodes are checked.

sample svg (+ ABCD_297.svg from #1255):
```
<svg viewBox="0 0 50 50" xmlns="http://www.w3.org/2000/svg">
  <defs>
    <g stroke="blue" stroke-width="2" id="gID">
    <circle id="myCircle" cx="0" cy="0" r="5" />
    </g>
  </defs>

  <use x="5" y="5" href="#myCircle" fill="green" />
  <use x="5" y="15" href="#gID" fill="red" />
</svg>
```

before:
bef

after:
aft

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1803,19 +1803,10 @@ static SvgNode* _getDefsNode(SvgNode* node)
 }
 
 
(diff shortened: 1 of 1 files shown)


JSUYA (approved):
LGTM

PR #1274, opened 2022-09-30
gl_engine: fixed the signature of the prepare() method in GlRenderer

vpxyz:
The implementation of the virtual prepare() method is different from that of RenderMethod, so the compilation fails. This modification solves the problem.


vpxyz:
Hi,
gl_engine compilation fails due to an incorrect signature in the GlRenderer class. This fix the compilation.
Ciao

hermet (approved):
thanks!

Issue #1275, opened 2022-11-17
Change the func type of Accessor::access() to std::function


JSUYA:
Currently, the `func` parameter type of the access() api is a function pointer.
```
std::unique_ptr<Picture> access(std::unique_ptr<Picture> picture, bool(*func)(const Paint* paint)) noexcept;
```
There are limited ways to pass user data to the func function. So to support this we can add `void* userData` or use std::function.
Existing APIs should be deprecated and new APIs should be added.
ex)
```
std::unique_ptr<Picture> access(std::unique_ptr<Picture> picture, const std::function<bool(const Paint* paint)>) noexcept;
```

hermet:
yes i think std::function sounds reasonable.

JSUYA:
Added new beta api. User can use std::function with this api.
`std::unique_ptr<Picture> set(std::unique_ptr<Picture> picture, std::function<bool(const Paint* paint)> func) noexcept;`

PR #1277, opened 2022-11-18
common Accessor: Add access API using std::function

JSUYA:
It can pass user data to the callback function.

https://github.com/Samsung/thorvg/issues/1275

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1615,6 +1615,19 @@ public:
     std::unique_ptr<Picture> access(std::unique_ptr<Picture> picture, bool(*func)(const Paint* paint)) noexcept;
 
     /**
+     * @brief Access the Picture scene stree nodes.
+     *
+     * @param[in] picture The picture node to traverse the internal scene-tree.
+     * @param[in] func The callback function calling for every paint nodes of the Picture.
+     * @param[in] data Data will be passed to callback function.
+     *
+     * @return Return the given @p picture instance.
+     *
+     * @note The bitmap based picture might not have the scene-tree.
+     */
+    std::unique_ptr<Picture> access(std::unique_ptr<Picture> picture, bool(*func)(const Paint* paint, void* data), void* data) noexcept;
+
+    /**
      * @brief Creates a new Accessor object.
      *
      * @return A new Accessor object.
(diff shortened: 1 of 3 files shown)


JSUYA:
I wanted to add an api that uses std::function.
ex)
```cpp
    std::unique_ptr<Picture> access(std::unique_ptr<Picture> picture, std::function<bool(const Paint* paint)> func) noexcept;

```
However, this causes an overloading error with an already opened API that use a function pointer.

inc/thorvg.h
```diff
@@ -1614,6 +1614,19 @@ class TVG_EXPORT Accessor final
...
+     * @brief Access the Picture scene stree nodes.
+     *
+     * @param[in] picture The picture node to traverse the internal scene-tree.
+     * @param[in] func The callback function calling for every paint nodes of the Picture.
+     * @param[in] data Data will be passed to callback function.
+     *
+     * @return Return the given @p picture instance.
+     *
+     * @note The bitmap based picture might not have the scene-tree.
+     */
```
hermet, line 1627: could you please add @BETA_API tag ?

hermet (approved):
looks good to me. please check a comment.

hermet:
@JSUYA btw, you suggested this https://github.com/Samsung/thorvg/issues/1275, i agree on this idea. dosen't need to use this type?

JSUYA:
> @JSUYA btw, you suggested this #1275, i agree on this idea. dosen't need to use this type?


As mentioned here(https://github.com/Samsung/thorvg/pull/1277#issuecomment-1319692487), there is an overload issue.

For these two APIs, Example's Accessor.cpp code makes an error.
    std::unique_ptr<Picture> access(std::unique_ptr<Picture> picture, bool(*func)(const Paint* paint)) noexcept;
    std::unique_ptr<Picture> access(std::unique_ptr<Picture> picture, std::function<bool(const Paint* paint)> func) noexcept;

[Accessor example]
```
```
[error]
```
../src/examples/Accessor.cpp:73:5: error: call of overloaded ‘access(std::remove_reference<std::unique_ptr<tvg::Picture>&>::type, tvgDrawCmds(tvg::Canvas*)::<lambda(const tvg::Paint*)>)’ is ambiguous
```

If we allow changing the type of the API that has already been opened, we can use https://github.com/Samsung/thorvg/pull/1280.

hermet:
@JSUYA how about naming the new method "Accessor::set()" then we can deprecate the previous one.

hermet:
great thanks!

PR #1279, opened 2022-11-20
[svg2png] size limitation while converting the file

mgrudzinska:
In case the svg file size is too large, a heap overflow occurred when conversting to png. To prevent this a size limitation has been added - the resolution of the resulting png file cannot be higher than 8k (7680 x 4320).

### DIFF
--- a/src/bin/svg2png/svg2png.cpp
+++ b/src/bin/svg2png/svg2png.cpp
@@ -37,6 +37,10 @@
     #include <sys/stat.h>
 #endif
 
+#define WIDTH_8K 7680
+#define HEIGHT_8K 4320
+#define SIZE_8K 33177600 //WIDTH_8K x HEIGHT_8K
+
 using namespace std;
 
 struct PngBuilder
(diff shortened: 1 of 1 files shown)


JSUYA (approved):
LGTM
+) Could you please add a log message for the user to know? 
And I think it would be good to show this info in help() as well.

src/bin/svg2png/svg2png.cpp
```diff
@@ -94,6 +97,17 @@ struct Renderer
             picture->size(&fw, &fh);
             w = static_cast<uint32_t>(fw);
             h = static_cast<uint32_t>(fh);
+
+            if (w > WIDTH_MAX || h > HEIGHT_MAX) {
```
hermet, line 101: how about if (w * h < 8K size)??

mgrudzinska:
thx for comments, I'll apply changes tmr

src/bin/svg2png/svg2png.cpp
mgrudzinska: done

PR #1280, opened 2022-11-21
common Accessor: Change parameter type of access api to std::funciton

JSUYA:
There are limited ways to pass user data to the func function.
So to support this we change parameter type to use std::function.


hermet:
@JSUYA  what do you think?
image

JSUYA:
thank you for review. I will close this pr. please refer to https://github.com/Samsung/thorvg/pull/1277.

Issue #1276, opened 2022-11-17
Supports the user to find the element's ID of SVG 


JSUYA:
We support finding `Paint` created via SVG using `Accessor`.
However, there is no way to check which of the elements declared by paint in svg.

In addition to `identifier()`, we can add `tag` or `id` of string type to `Paint` and add it for user to set and get.
Or we can add a way to provide `id` list for svg or tvg in picture.

ex svg)
```
<svg id="svg1" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
  <circle id = "1" r="32" cx="35" cy="65" fill="#F00" opacity="0.5"/>
  <circle id = "2" r="32" cx="65" cy="65" fill="#0F0" opacity="0.5"/>
  <circle id = "3" r="32" cx="50" cy="35" fill="#00F" opacity="0.5"/>
</svg>
```

hermet:
tvg paint hierarchy won't be exactly identical with svg's data structure.. seems quite fragile. Need to think a concrete solution...

hermet:
Let's close this and raise a issue when the user comes up with a requirement.

PR #1285, opened 2022-11-29
common Paint: Add name getter/setter API

JSUYA:
Adds an API that can set/get string type name to tvgPaint.
This can be used when the user needs to traverse the scene tree or find a specific instance.
+) The SVG's "id" attribute is set to name when SVG is loaded.

Issue: 
- https://github.com/thorvg/thorvg/issues/1276
- https://github.com/thorvg/thorvg/issues/2055

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -427,6 +427,28 @@ public:
      */
     uint32_t identifier() const noexcept;
 
+    /**
+     * @brief Sets the name for the current instance.
+     *
+     * The name can be up to 127 characters long.
+     *
+     * @note The "id" attribute of the SVG format is set to name, and save/load is not yet supported in the case of the TVG format.
+     *
+     * @return Result::Success when succeed, Result::InvalidArguments in case the name length exceeded 128 characters.
+     *
+     * @BETA_API
+     */
+    Result name(std::string name) const noexcept;
+
+    /**
+     * @brief Returns the name set for the current instance.
+     *
+     * @return The name set on the current instance when succeed, otherwise an empty string.
+     *
+     * @BETA_API
+     */
+    std::string name() const noexcept;
+
     _TVG_DECLARE_PRIVATE(Paint);
 };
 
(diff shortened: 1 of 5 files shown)


JSUYA:
This is a simple solution for https://github.com/Samsung/thorvg/issues/1276.
I think the size(cost) of tvgPaint can up unnecessarily. and it can not be used universally except in special cases.
What do you think?

JSUYA:
Rebased PR

JSUYA:
@hermet Do you have any comments for updates? At the time this patch was written, it wasn't a necessary thing. There were comments related to hashing, but I can't pay attention to them for a long time. (If hashing, it can be used more widly, but for now, I think it is over-spec. It can be sufficiently improved in the future.)
Or someone can take this task.

hermet:
@JSUYA I agree with @kariem2k's opinion; ThorVG could save memory in general cases if it has a name with a single reference id. Thus, the usage would be something like this
```
bool paint->identify(string);   //Internally encode the string to a hash ID and compare its value.
```

Yet my question is whether `tvg::Paint` needs to return the name as a string or not. If so, how can it be done without any cost. On consideration.

JSUYA:
> Yet my question is whether `tvg::Paint` needs to return the name as a string or not. If so, how can it be done without any cost. On consideration.

It is not necessary to return it as a string. We can use `if(Accessor::hashToString(paint->id()) == "ID") `.

hermet:
> Accessor::hashToString

If ThorVG should support value recovery, then I guess hashing wouldn't be beneficial.

hermet:
@JSUYA Please have a look at this. I'm not sure about this feature for SVG since there have been no practical requirements so far. It will just make things difficult for the SVG loader without any purpose.

https://github.com/thorvg/thorvg/pull/2334/commits/a0b1e502c25b01daba7b9e0876dc1e5b0c23782c
https://github.com/thorvg/thorvg/pull/2334/commits/380b6d596ee17fd68c047949ccac7ccd4bb1c852

JSUYA:
> @JSUYA Please have a look at this. I'm not sure about this feature for SVG since there have been no practical requirements so far. It will just make things difficult for the SVG loader without any purpose.
> 
> [a0b1e50](https://github.com/thorvg/thorvg/commit/a0b1e502c25b01daba7b9e0876dc1e5b0c23782c) [380b6d5](https://github.com/thorvg/thorvg/commit/380b6d596ee17fd68c047949ccac7ccd4bb1c852)

Oh I've been waiting for your opinion since i think this discussion is not conclusive. I checked the contents of the `hermet/type` branch.
In my memory, this issue was requested by someone. https://github.com/thorvg/thorvg/issues/1276 But I don't remember
I agree that it is not a necessary feature now.

I'm closing this PR. If a feature related to this(string type return) is needed, it can be discussed again in a new place other than this PR.

PR #1286, opened 2022-11-29
docs : Add vcpkg Installation Document

rinechran:
Thorvg can be installed with vcpkg

I added it to the readme.


hermet (approved):
thanks!

PR #1291, opened 2022-12-12
svg_loader: Change SVG viewbox variable type from int to float

JSUYA:
The viewbox and size(width, height) defined in SVG can be float type.
This prevents matrix calculation errors caused by this.


JSUYA:
In https://github.com/Samsung/thorvg/issues/1210#issuecomment-1345796361 fix `2.png`'s error.

hermet (approved):
no doubt on this, probably we need to double-check any regressions just in case.

Issue #1192, opened 2022-02-17
two issues with composition


mgrudzinska:
in the example below there are 2 issues:
- skewing: for `PROBLEM1 != 0` and   `PROBLEM2 == 0` the `_compFastTrack` returns `true`, which is not the case (for `PROBLEM1 == 0` and   `PROBLEM2 != 0` the `_compFastTrack` returns `false` as it should)   ->   inside `_compFastTrack` funct the  `mathRightAngle` checks the e21 value, but the e12 val is never checked
- composition for cases when `_compFastTrack` return false - doesn't work 

example:
```
    float PROBLEM1 = 0.1f;
    float PROBLEM2 = 0.1f;

    auto scene = tvg::Scene::gen();
    tvg::Matrix m = {1, PROBLEM1, -50, PROBLEM2, 1, 0, 0, 0, 1};

    auto star1 = tvg::Shape::gen();
    tvgDrawStar(star1.get());
    star1->fill(255, 255, 0, 255);
    scene->push(move(star1));

    auto clip = tvg::Shape::gen();
    clip->appendRect(100, 100, 100, 100, 0, 0);
    clip->transform(m);
    scene->composite(move(clip), tvg::CompositeMethod::ClipPath);

    canvas->push(move(scene));

    // shapes to compare
    clip = tvg::Shape::gen();
    clip->fill(0, 0, 255, 55);
    clip->appendRect(100, 100, 100, 100, 0, 0);
    clip->transform(m);
    canvas->push(move(clip));

    star1 = tvg::Shape::gen();
    tvgDrawStar(star1.get());
    star1->stroke(255 ,0, 0, 55);
    star1->stroke(10);
    canvas->push(move(star1));

```

mgrudzinska:
solved

PR #1296, opened 2022-12-27
common: ignoring color/alpha/opacity of a clip object

mgrudzinska:
According to the svg specs clip's fill and opacity should be ignored. Till now setting the alpha/opacity value to zero resulted in the shape's rendering abort.

part of the #1192 issue

### DIFF
--- a/src/lib/sw_engine/tvgSwRenderer.cpp
+++ b/src/lib/sw_engine/tvgSwRenderer.cpp
@@ -77,7 +77,9 @@ struct SwShapeTask : SwTask
 
     void run(unsigned tid) override
     {
-        if (opacity == 0) return;  //Invisible
+        auto compMethod = CompositeMethod::None;
+        auto usedAsClip = (sdata->composite(nullptr, &compMethod) == Result::Success) && (compMethod == CompositeMethod::ClipPath);
+        if (opacity == 0 && !usedAsClip) return;  //Invisible
 
         uint8_t strokeAlpha = 0;
         auto visibleStroke = false;
@@ -99,7 +101,7 @@ struct SwShapeTask : SwTask
             sdata->fillColor(nullptr, nullptr, nullptr, &alpha);
             alpha = static_cast<uint8_t>(static_cast<uint32_t>(alpha) * opacity / 255);
             visibleFill = (alpha > 0 || sdata->fill());
-            if (visibleFill || visibleStroke) {
+            if (visibleFill || visibleStroke || usedAsClip) {
                 shapeReset(&shape);
                 if (!shapePrepare(&shape, sdata, transform, clipRegion, bbox, mpool, tid, clips.count > 0 ? true : false)) goto err;
             }
@@ -111,7 +113,7 @@ struct SwShapeTask : SwTask
 
         //Fill
         if (flags & (RenderUpdateFlag::Gradient | RenderUpdateFlag::Transform | RenderUpdateFlag::Color)) {
-            if (visibleFill) {
+            if (visibleFill || usedAsClip) {
                 /* We assume that if stroke width is bigger than 2,
                    shape outline below stroke could be full covered by stroke drawing.
                    Thus it turns off antialising in that condition.
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -364,6 +364,20 @@ public:
     CompositeMethod composite(const Paint** target) const noexcept;
 
     /**
+     * @brief Gets the composition source object and the composition method.
+     *
+     * @param[out] source The paint of the composition source object.
+     * @param[out] method The method used to composite the source object with the target.
+     *
+     * @return Result::Success when the paint object used as a composition target, Result::InsufficientCondition otherwise.
+     *
+     * @warning Please do not use it, this API is not official one. It could be modified in the next version.
+     *
+     * @BETA_API
+     */
+    Result composite(const Paint** source, CompositeMethod* method) const noexcept;
+
+    /**
      * @brief Return the unique id value of the paint instance.
      *
      * This method can be called for checking the current concrete instance type.
--- a/src/examples/ClipPath.cpp
+++ b/src/examples/ClipPath.cpp
@@ -63,9 +63,9 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     //Move Star1
     star1->translate(-10, -10);
 
+    // color/alpha/opacity are ignored for a clip object - no need to set them
     auto clipStar = tvg::Shape::gen();
     clipStar->appendCircle(200, 230, 110, 110);
-    clipStar->fill(255, 255, 255, 255); // clip object must have alpha.
     clipStar->translate(10, 10);
 
     star1->composite(move(clipStar), tvg::CompositeMethod::ClipPath);
@@ -80,9 +80,9 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     //Move Star2
     star2->translate(10, 40);
 
+    // color/alpha/opacity are ignored for a clip object - no need to set them
     auto clip = tvg::Shape::gen();
     clip->appendCircle(200, 230, 130, 130);
-    clip->fill(255, 255, 255, 255); // clip object must have alpha.
     clip->translate(10, 10);
 
     scene->push(move(star1));
@@ -110,9 +110,9 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     star3->stroke(10);
     star3->translate(400, 0);
 
+    // color/alpha/opacity are ignored for a clip object - no need to set them
     auto clipRect = tvg::Shape::gen();
     clipRect->appendRect(500, 120, 200, 200, 0, 0);          //x, y, w, h, rx, ry
-    clipRect->fill(255, 255, 255, 255); // clip object must have alpha.
     clipRect->translate(20, 20);
 
     //Clipping scene to rect(shape)
@@ -127,10 +127,10 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     picture->scale(3);
     picture->translate(50, 400);
 
+    // color/alpha/opacity are ignored for a clip object - no need to set them
     auto clipPath = tvg::Shape::gen();
     clipPath->appendCircle(200, 510, 50, 50);          //x, y, w, h, rx, ry
     clipPath->appendCircle(200, 650, 50, 50);          //x, y, w, h, rx, ry
-    clipPath->fill(255, 255, 255, 255); // clip object must have alpha.
     clipPath->translate(20, 20);
 
     //Clipping picture to path
@@ -143,9 +143,9 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     shape1->appendRect(500, 420, 100, 100, 20, 20);
     shape1->fill(255, 0, 255, 160);
 
+    // color/alpha/opacity are ignored for a clip object - no need to set them
     auto clipShape = tvg::Shape::gen();
     clipShape->appendRect(600, 420, 100, 100, 0, 0);
-    clipShape->fill(255, 0, 255, 150);
 
     //Clipping shape1 to clipShape
     shape1->composite(move(clipShape), tvg::CompositeMethod::ClipPath);
--- a/src/lib/tvgPaint.cpp
+++ b/src/lib/tvgPaint.cpp
@@ -385,6 +385,19 @@ CompositeMethod Paint::composite(const Paint** target) const noexcept
 }
 
 
+Result Paint::composite(const Paint** source, CompositeMethod* method) const noexcept
+{
+    if (source) *source = pImpl->compSource;
+    auto met = (pImpl->compSource && pImpl->compSource->pImpl->compData ?
+                pImpl->compSource->pImpl->compData->method : CompositeMethod::None);
+    if (method) *method = met;
+
+    if (pImpl->compSource != nullptr && met != CompositeMethod::None)
+        return Result::Success;
+    return Result::InsufficientCondition;
+}
+
+
 Result Paint::opacity(uint8_t o) noexcept
 {
     if (pImpl->opacity == o) return Result::Success;
--- a/src/lib/tvgPaint.h
+++ b/src/lib/tvgPaint.h
@@ -63,6 +63,7 @@ namespace tvg
         StrategyMethod* smethod = nullptr;
         RenderTransform* rTransform = nullptr;
         Composite* compData = nullptr;
+        Paint* compSource = nullptr;
         uint32_t renderFlag = RenderUpdateFlag::None;
         uint32_t ctxFlag = ContextFlag::Invalid;
         uint32_t id;
@@ -137,6 +138,7 @@ namespace tvg
                 if (!target && method == CompositeMethod::None) return true;
                 compData = static_cast<Composite*>(calloc(1, sizeof(Composite)));
             }
+            target->pImpl->compSource = source;
             compData->target = target;
             compData->source = source;
             compData->method = method;


mgrudzinska:
@hermet also I'd like to remove 'source' from the Composite struct, but I'd wait for a rev in this pr.

hermet:
Quick question. looks compatibility break. thorvg renderer hasn't been designed only for svg spec. couldn't be handled in svg loader?

mgrudzinska:
I can try to handle this in the svg loader, but the apis behavior is now a bit counter intuitive

try this example:
'''
    // 1. fastTrack logic used
    clip->appendRect(0, 0, 100, 100, 0, 0);
    // clip->fill(255, 0, 0, 0);     // this line doesn't change anything
    circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
    canvas->push(move(circle));

    // 2. fastTrack is not applied in this case - nothing is drawn 
    circle = tvg::Shape::gen();
    circle->appendCircle(200, 200, 100, 100);
    circle->fill(0, 0, 255, 255);

    clip = tvg::Shape::gen();
    clip->appendRect(100, 100, 100, 100, 1, 1);
    // clip->fill(255, 0, 0, 0);     // this line doesn't change anything
    circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
    canvas->push(move(circle));
'''

current result:
Zrzut ekranu 2022-12-27 o 23 00 34

expected result:
Zrzut ekranu 2022-12-27 o 23 00 02


Do you think we can live with that?

hermet:
@mgrudzinska ok, thanks. I will review your changes.

src/lib/sw_engine/tvgSwRenderer.cpp
```diff
@@ -76,7 +76,9 @@ struct SwShapeTask : SwTask
 
     void run(unsigned tid) override
     {
-        if (opacity == 0) return;  //Invisible
+        auto compMethod = CompositeMethod::None;
+        auto usedAsClip = sdata->composite(nullptr, &compMethod) && (compMethod == CompositeMethod::ClipPath);
```
JSUYA, line 80: According to the code, it seems important to find out whether it is used as a clip in the SwRenderer.
I think this could be handled internally, but I can't come up with a good way.
um... I think we can check them because we update the clips first in the "prepare" step.
maybe like this?
```diif
--- a/src/lib/sw_engine/tvgSwRenderer.cpp
+++ b/src/lib/sw_engine/tvgSwRenderer.cpp
@@ -73,11 +73,12 @@ struct SwShapeTask : SwTask
     SwShape shape;
     const Shape* sdata = nullptr;
     bool cmpStroking = false;
+    bool usedAsClip = false;
 
     void run(unsigned tid) override
     {
         auto compMethod = CompositeMethod::None;
-        auto usedAsClip = sdata->composite(nullptr, &compMethod) && (compMethod == CompositeMethod::ClipPath);
+        //auto usedAsClip = sdata->composite(nullptr, &compMethod) && (compMethod == CompositeMethod::ClipPath);
         if (opacity == 0 && !usedAsClip) return;  //Invisible
 
         uint8_t strokeAlpha = 0;
@@ -593,6 +594,7 @@ void* SwRenderer::prepareCommon(SwTask* task, const RenderTransform* transform,
     if (clips.count > 0) {
         //Guarantee composition targets get ready.
         for (auto clip = clips.data; clip < (clips.data + clips.count); ++clip) {
+            static_cast<SwShapeTask*>(*clip)->usedAsClip = true;
             static_cast<SwShapeTask*>(*clip)->done();

```

If the composite() getter API could be used elsewhere, this might be better.

JSUYA (commented):
SVG's ClipPath does not check opacity and color. So the purpose of this PR is right.
(But it is affected by clippath's `display` attribute. We need to fix this on another PR.... hahaha...)

mgrudzinska:
@JSUYA thx for your rev

1. > SVG's ClipPath does not check opacity and color. So the purpose of this PR is right. (But it is affected by clippath's `display` attribute. We need to fix this on another PR.... hahaha...)

The 'display' attrib is handled int the svg loader and it seams to be handled correctly. pleas let me know if something has to be modified. 

2. regarding bool -> Result - applied

3. the comment about moving the check to the prepareCommon. I'll double check, but as far I saw it didn;t work. I'll come back with this part and leave a comment

mgrudzinska:
@JSUYA regarding the 3rd point - it won't work, since the clips are handled before the targets. setting the 'usedAsClip' value while handling the target is too late already.

JSUYA:
> @JSUYA regarding the 3rd point - it won't work, since the clips are handled before the targets. setting the 'usedAsClip' value while handling the target is too late already.

Hi this is my test code.
I just wanted to think about how to handle it without a separate API.
( `static_cast<SwShapeTask*>(*clip)->usedAsClip = true;` um I don't like this part pretty much... )

[demo code]
```cpp
    auto circle = tvg::Shape::gen();
    circle->appendCircle(200, 200, 100, 100);
    circle->fill(0, 0, 255, 255);

    clip->appendRect(100, 100, 100, 100, 1, 1);
    circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
    canvas->push(move(circle));
```

[swengine diff]
```diff cpp
diff --git a/src/lib/sw_engine/tvgSwRenderer.cpp b/src/lib/sw_engine/tvgSwRenderer.cpp
index 03bc695..681e746 100644
--- a/src/lib/sw_engine/tvgSwRenderer.cpp
+++ b/src/lib/sw_engine/tvgSwRenderer.cpp
@@ -73,6 +73,7 @@ struct SwShapeTask : SwTask
     SwShape shape;
     const Shape* sdata = nullptr;
     bool cmpStroking = false;
+    bool usedAsClip = false;

     void run(unsigned tid) override
     {
@@ -98,7 +99,7 @@ struct SwShapeTask : SwTask
             sdata->fillColor(nullptr, nullptr, nullptr, &alpha);
             alpha = static_cast<uint8_t>(static_cast<uint32_t>(alpha) * opacity / 255);
             visibleFill = (alpha > 0 || sdata->fill());
-            if (visibleFill || visibleStroke) {
+            if (visibleFill || visibleStroke || usedAsClip) {
                 shapeReset(&shape);
                 if (!shapePrepare(&shape, sdata, transform, clipRegion, bbox, mpool, tid, clips.count > 0 ? true : false)) goto err;
             }
@@ -110,7 +111,7 @@ struct SwShapeTask : SwTask

         //Fill
         if (flags & (RenderUpdateFlag::Gradient | RenderUpdateFlag::Transform | RenderUpdateFlag::Color)) {
-            if (visibleFill) {
+            if (visibleFill || usedAsClip) {
                 /* We assume that if stroke width is bigger than 2,
                    shape outline below stroke could be full covered by stroke drawing.
                    Thus it turns off antialising in that condition.
@@ -591,6 +592,7 @@ void* SwRenderer::prepareCommon(SwTask* task, const RenderTransform* transform,
     if (clips.count > 0) {
         //Guarantee composition targets get ready.
         for (auto clip = clips.data; clip < (clips.data + clips.count); ++clip) {
+            static_cast<SwShapeTask*>(*clip)->usedAsClip = true;
             static_cast<SwShapeTask*>(*clip)->done();
         }
         task->clips = clips;

```

mgrudzinska:
```
    if (picture->load(EXAMPLE_DIR"/cartman.svg") != tvg::Result::Success) return;
    picture->scale(30);

    auto clipPath = tvg::Shape::gen();
    clipPath->appendCircle(200, 510, 50, 50);
    //clipPath->fill(255, 255, 255, 255);

    picture->composite(move(clipPath), tvg::CompositeMethod::ClipPath);

    canvas->push(move(picture));
```

mgrudzinska:
@JSUYA please change the example to be using only one thread. 

in tvgPaint.cpp the Paint::Impl::update() function - if a paint has a composition it is handled first, that's why your proposition  won't work in a general case (it may work if there is more threads). Do you agree? 

I understand that the api I've introduced is not a very elegant solution. We can think on some other one.

JSUYA (approved):
Oh I didn't think of a one thread case.
I tried to find a way to handle it by modifying the internal API, but I couldn't find a good solution because the structure was complicated for each case.

+) There is already a BETA_API tag in Docs, but I recommend adding a warning message just like any other beta apis.

```
@warning Please do not use it, this API is not official one. It could be modified in the next version.
```

mgrudzinska:
> +) There is already a BETA_API tag in Docs, but I recommend adding a warning message just like any other beta apis.
> 
> ```
> @warning Please do not use it, this API is not official one. It could be modified in the next version.
> ```

done

hermet:
@mgrudzinska @JSUYA does it hard to call fill(255 255 255 255) / opacity(255) when it's going to call compoiste() in svgloader side?

hermet:
Good job. pleaes let me think further at these points.
1. Breaking compatibility is the last chance if we have no any options. (Does this inevitable?)
2. Applying this change without adding a new api.

hermet:
@mgrudzinska Hi, I tested your sample code(https://github.com/Samsung/thorvg/issues/1192) and they seems working fine without your patch. I don't know why you picked up this as a issue. Also this one looks working well as well.

```
    auto circle = tvg::Shape::gen();
    circle->appendCircle(100, 100, 100, 100);
    circle->fill(255, 0, 0, 255);

    clip->appendRect(0, 0, 100, 100, 0, 0);
    clip->fill(255, 255, 255, 255);
    circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
    canvas->push(move(circle));    

    circle = tvg::Shape::gen();
    circle->appendCircle(200, 200, 100, 100);
    circle->fill(0, 0, 255, 255);

    clip = tvg::Shape::gen();
    clip->appendRect(100, 100, 100, 100, 1, 1);
    clip->fill(255, 255, 255, 255); 
    circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
    canvas->push(move(circle));   
```
shot-2023-01-15_17-35-38

hermet (commented):
Please have a look at my nit-picks.

mgrudzinska:
> @mgrudzinska Hi, I tested your sample code(#1192) and they seems working fine without your patch. I don't know why you raised up this with the issue. 

No, it doesn't. 
Current result:
Zrzut ekranu 2023-01-15 o 16 22 26
The expected result:
Zrzut ekranu 2023-01-15 o 16 26 05


> Also this following code looks working well too.
> 
> ```
>     auto circle = tvg::Shape::gen();
>     circle->appendCircle(100, 100, 100, 100);
>     circle->fill(255, 0, 0, 255);
> 
>     auto clip = tvg::Shape::gen();
>     clip->appendRect(0, 0, 100, 100, 0, 0);
>     clip->fill(255, 255, 255, 255);
>     circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
>     canvas->push(move(circle));    
> 
>     circle = tvg::Shape::gen();
>     circle->appendCircle(200, 200, 100, 100);
>     circle->fill(0, 0, 255, 255);
> 
>     clip = tvg::Shape::gen();
>     clip->appendRect(100, 100, 100, 100, 1, 1);
>     clip->fill(255, 255, 255, 255); 
>     circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
>     canvas->push(move(circle));   
> ```
> 
> shot-2023-01-15_17-35-38
> 
> Please let me know if I'm missing.

Sure it works, since you set the clip's alpha value to 255. that's the problem. If you remove this line, the default alpha value of a shape is 0. In such a case your 'clip' won't be handled, unless it's rectangle - then different logic is applied and the result is as expected.

hermet:
> > @mgrudzinska Hi, I tested your sample code(#1192) and they seems working fine without your patch. I don't know why you raised up this with the issue.
> 
> No, it doesn't. Current result: Zrzut ekranu 2023-01-15 o 16 22 26 The expected result: Zrzut ekranu 2023-01-15 o 16 26 05
> 
> > Also this following code looks working well too.
> > ```
> >     auto circle = tvg::Shape::gen();
> >     circle->appendCircle(100, 100, 100, 100);
> >     circle->fill(255, 0, 0, 255);
> > 
> >     auto clip = tvg::Shape::gen();
> >     clip->appendRect(0, 0, 100, 100, 0, 0);
> >     clip->fill(255, 255, 255, 255);
> >     circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
> >     canvas->push(move(circle));    
> > 
> >     circle = tvg::Shape::gen();
> >     circle->appendCircle(200, 200, 100, 100);
> >     circle->fill(0, 0, 255, 255);
> > 
> >     clip = tvg::Shape::gen();
> >     clip->appendRect(100, 100, 100, 100, 1, 1);
> >     clip->fill(255, 255, 255, 255); 
> >     circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
> >     canvas->push(move(circle));   
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > shot-2023-01-15_17-35-38
> > Please let me know if I'm missing.
> 
> Sure it works, since you set the clip's alpha value to 255. that's the problem. If you remove this line, the default alpha value of a shape is 0. In such a case your 'clip' won't be handled, unless it's rectangle - then different logic is applied and the result is as expected.

So what is the bug exactly when you miss the 255 ? currently the clippers are supposed to have alpha=255. without it we don't guarantee any correct result. our examples guides it already.

mgrudzinska:
> So what is the bug exactly when you miss the 255 ? currently the clippers are supposed to have alpha=255. without it we don't guarantee any correct result. our examples guides it already.

since the behavior is different for the 'fasttrack' cases and for the rest, I think it's very counter intuitive. when I opened #1192 I didn't know what was the exact reason of the problem. 

If you think that it works as it should, please close this pr and the #1192 issue as well (the other part ot the problem is already solved and merged)

hermet:
> > So what is the bug exactly when you miss the 255 ? currently the clippers are supposed to have alpha=255. without it we don't guarantee any correct result. our examples guides it already.
> 
> since the behavior is different for the 'fasttrack' cases and for the rest, I think it's very counter intuitive. when I opened #1192 I didn't know what was the exact reason of the problem.
> 
> If you think that it works as it should, please close this pr and the #1192 issue as well (the other part ot the problem is already solved and merged)

So far If i understood correctly, this issue is regarded with clippers which doesn't set alpha/opacity properly. Then I would like to say, wrong usage doesn't guarantee any expected results. It's a kind of like that you free() twice for a memory, sometimes it can be crashed sometimes it doesn't. Rather than considered a  issue, I have looked this as an usage enhancement, not fixing issues. So PR is still valid.

JSUYA:
```
// #1
    auto circle = tvg::Shape::gen();
    circle->appendCircle(100, 100, 100, 100);
    circle->fill(255, 0, 0, 255);

    clip->appendRect(0, 0, 100, 100, 0, 0);
    //clip->fill(255, 255, 255, 255);
    circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
    canvas->push(move(circle));    

// #2
    circle = tvg::Shape::gen();
    circle->appendCircle(200, 200, 100, 100);
    circle->fill(0, 0, 255, 255);

    clip = tvg::Shape::gen();
    clip->appendRect(100, 100, 100, 100, 1, 1);
    //clip->fill(255, 255, 255, 255); 
    circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
    canvas->push(move(circle));   
```

These two examples have the same content but different results. (#2 is not displayed, #2 is rounded rectangle)
I also think it's very counter intuitive.

If we have to choose between, I think clippath should only use the path information.
If the user needs visibility of the source, he can use Mask.
We say in the docs that it is the result of `rendered` pixels,
but in fact properties such as the thickness of the path do not affect it.

This is just my opinion, so I think we just need to define it clearly.
(Both #1 and #2 must be visible or neither must be invisible.)

hermet:
> ```
> // #1
>     auto circle = tvg::Shape::gen();
>     circle->appendCircle(100, 100, 100, 100);
>     circle->fill(255, 0, 0, 255);
> 
>     auto clip = tvg::Shape::gen();
>     clip->appendRect(0, 0, 100, 100, 0, 0);
>     //clip->fill(255, 255, 255, 255);
>     circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
>     canvas->push(move(circle));    
> 
> // #2
>     circle = tvg::Shape::gen();
>     circle->appendCircle(200, 200, 100, 100);
>     circle->fill(0, 0, 255, 255);
> 
>     clip = tvg::Shape::gen();
>     clip->appendRect(100, 100, 100, 100, 1, 1);
>     //clip->fill(255, 255, 255, 255); 
>     circle->composite(move(clip), tvg::CompositeMethod::ClipPath);
>     canvas->push(move(circle));   
> ```
> 
> These two examples have the same content but different results. (#2 is not displayed, #2 is rounded rectangle) I also think it's very counter intuitive.
> 
> If we have to choose between, I think clippath should only use the path information. If the user needs visibility of the source, he can use Mask. We say in the docs that it is the result of `rendered` pixels, but in fact properties such as the thickness of the path do not affect it.
> 
> This is just my opinion, so I think we just need to define it clearly. (Both #1 and #2 must be visible or neither must be invisible.)

Ok, first of all, to be clarfied, this is not about a bug nor issue, this is about a specification problem. 

Current our clipper requires the alpha=255. it's our guidance and specification. if a user writes that above code, at the end he must call the fill alpha=255 whatever we fix the current result. Without it, we can't guarantee the result. Specifically, it's out of our ability. Thinking about the svg which are written wrongly. the result wont' be same among the svg engines even we can't expect the result how it look like. the code without aligning the api spec, will be fragile and unstable, no fw/engines perfectly gurantee the compatibility for the unexpected usages. That's the fact.

Let's imagine we don't display it if the clipper have no alpha, then someone might think it's wrong result in the meantime someone might think it's good. So this is a perspective problem. there is no perfect answer. 

I agreen on enhancing our api usage. it's good to go, that's why I think this PR is still valid but if we argue it's a bug/issue for the above usage, no I will say go fix the user code. that's the current solution.

Here are my thought in my mind:
 
1. we can put Error log when clipper has no alpha/opacity.
2. we set alpha/opacity forcely at the canvas side when the clipper has no values of them.
3. stick to apply this PR approach.

mgrudzinska:
@hermet @JSUYA 
no matter what we decide here, to fulfill the svg standard, I've prepared #1308 - dealing with the clip's opacity inside the svg loader

mgrudzinska:
And what about #1309?
regardless of this one, I think we still need #1308

hermet (approved):

hermet:
@mgrudzinska "api: composite return type changed" this patch shouldn't be separated. It must be merged into the first patch.

PR #1297, opened 2022-12-27
paint: fastTrack should return false for the skew transformations

mgrudzinska:
part of the #1192 issue


hermet (approved):
For the next time, I'd like to ask you to add the issue link in the commit message for tracking the change history later like:

@Issue: https://github.com/Samsung/thorvg/issues/1192

Or, you can check other commit messages how they handle 
Or our guideline:

This would be helpful when we summarize the release note from the commit history.

Thanks.

PR #1298, opened 2023-01-01
scene: fixing nested masking

mgrudzinska:
For any type of masking composition may be required.

@Issue: https://github.com/Samsung/thorvg/issues/1295

### DIFF
--- a/src/lib/tvgSceneImpl.h
+++ b/src/lib/tvgSceneImpl.h
@@ -60,6 +60,11 @@ struct Scene::Impl
     Array<Paint*> paints;
     uint8_t opacity;                     //for composition
     RenderMethod* renderer = nullptr;    //keep it for explicit clear
+    Scene* scene = nullptr;
+
+    Impl(Scene* s) : scene(s)
+    {
+    }
 
     ~Impl()
     {
@@ -81,8 +86,14 @@ struct Scene::Impl
 
     bool needComposition(uint32_t opacity)
     {
+        if (opacity == 0 || paints.count == 0) return false;
+
+        //Masking may require composition (even if opacity == 255)
+        auto compMethod = scene->composite(nullptr);
+        if (compMethod != CompositeMethod::None && compMethod != CompositeMethod::ClipPath) return true;
+
         //Half translucent requires intermediate composition.
-        if (opacity == 255 || opacity == 0) return false;
+        if (opacity == 255) return false;
 
         //If scene has several children or only scene, it may require composition.
         if (paints.count > 1) return true;
--- a/src/lib/tvgScene.cpp
+++ b/src/lib/tvgScene.cpp
@@ -25,7 +25,7 @@
 /* External Class Implementation                                        */
 /************************************************************************/
 
-Scene::Scene() : pImpl(new Impl())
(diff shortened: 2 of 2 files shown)


mgrudzinska:
@hermet I know this issue was assigned to you. I just found this solution while making some tests. If you have a better solution please fell free to close this pr

src/lib/tvgSceneImpl.h
```diff
@@ -81,12 +86,21 @@ struct Scene::Impl
...
+
         //Half translucent requires intermediate composition.
-        if (opacity == 255 || opacity == 0) return false;
+        if (opacity == 255 || opacity == 0)
+            return false;
 
         //If scene has several children or only scene, it may require composition.
-        if (paints.count > 1) return true;
-        if (paints.count == 1 && (*paints.data)->identifier() == TVG_CLASS_ID_SCENE) return true;
+        if (paints.count > 1)
```
JSUYA, line 99: Shouldn't it be modified with `if (paints.count > 0)`? (no method check)
Are there any exception cases?

mgrudzinska: I'm not sure what do you mean. I didn;t modify this part, just the new lines are added (sorry for that).
Sure we can check whether there are some paints on the very beginning of this function, but I guess it's not what you meant...

JSUYA: Sorry for the confusion.
What I meant was to perform composition unconditionally if there is paint in the scene.
(However, this method doesn't seem appropriate because it will return true even in unnecessary composition cases.)

The needComposition() function is getting a bit complicated. How about clean up priorities a bit?
I think at least opacity==0 should be on top in all cases.
like this?
```cpp
            if (opacity == 0) return false;

        auto compMethod = scene->composite(nullptr);
        if (paints.count > 0 && compMethod != CompositeMethod::None && compMethod != CompositeMethod::ClipPath) return true;

        //Half translucent requires intermediate composition.
        if (opacity == 255) return false;

        //If scene has several children or only scene, it may require composition.
        if (paints.count > 1) return true;
        if (paints.count == 1 && (*paints.data)->identifier() == TVG_CLASS_ID_SCENE) return true;
        return false;
    }
```

+)
nit : please remove the newline after the if statement.

src/lib/tvgSceneImpl.h
```diff
@@ -81,8 +86,14 @@ struct Scene::Impl
...
+
+        //Masking may require composition (even if opacity == 255)
+        auto compMethod = scene->composite(nullptr);
+        if (compMethod != CompositeMethod::None && compMethod != CompositeMethod::ClipPath) return true;
+
         //Half translucent requires intermediate composition.
-        if (opacity == 255 || opacity == 0) return false;
+        if (opacity == 255) return false;
 
         //If scene has several children or only scene, it may require composition.
```
mgrudzinska, line 98: these can be replaced by:
```
    if (paints.count == 1 && (*paints.data)->identifier() != TVG_CLASS_ID_SCENE) return false;
    return true;
```
but I think it's less readable so I kept this as it was.

JSUYA (approved):
LGTM

hermet (approved):
good job thanks!

PR #1301, opened 2023-01-06
svg_loader: fixing crash for to big buffer

mgrudzinska:
Crash observed on macOS for the image-embeded-*.svg files. Since the alloca function was used the stack allocation failure could not be handled.


mgrudzinska:
I know why we used alloca. So the other solution could be to use some arbitrary chosen MAX_BUF_LENGTH value.

hermet (approved):

hermet:
just curious how big size was requested here?

mgrudzinska:
> just curious how big size was requested here?

mote than 0.5MB (0.54MB and 0.63MB to be precise)

PR #1308, opened 2023-01-17
svg_loader: overwrite the clip's opacity/alpha

mgrudzinska:
According to the svg standard the clips opacity
doesn't affect the final rendering. In order to
avoid any confusion the opacity values are
overwritten by the max possible value.

sample:
```
<svg viewBox="0 0 100 100">
  <g fill-opacity="0">
  <clipPath id="myClip">
    <circle cx="40" cy="35" r="35" opacity="0" fill-opacity="0.9"/>
    <circle cx="70" cy="35" r="15" fill-opacity="0"/>
  </clipPath>
 </g>
 <path d="M10,30 A20,20,0,0,1,50,30 A20,20,0,0,1,90,30 Q90,60,50,90 Q10,60,10,30 Z" clip-path="url(#myClip)" />
</svg>
```

before:
Zrzut ekranu 2023-01-17 o 18 38 33

after:
Zrzut ekranu 2023-01-17 o 18 37 57

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -255,7 +255,6 @@ static void _applyComposition(Paint* paint, const SvgNode* node, const Box& vBox
             node->style->clipPath.applying = true;
 
             auto comp = Shape::gen();
-            comp->fill(255, 255, 255, 255);
             if (node->transform) comp->transform(*node->transform);
 
             auto child = compNode->child.data;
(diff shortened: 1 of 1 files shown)


hermet (approved):

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -264,6 +263,8 @@ static void _applyComposition(Paint* paint, const SvgNode* node, const Box& vBox
             for (uint32_t i = 0; i < compNode->child.count; ++i, ++child) {
                 if (_appendChildShape(*child, comp.get(), vBox, svgPath)) valid = true;
             }
+            comp->fill(255, 255, 255, 255);
```
hermet, line 266: probably, we can call the fill()/opacity() in the valid condition
```
if (valid) {
   comp->fill();
   comp->opacity();
   paint->composite(move(comp), CompositeMethod::ClipPath);
}

hermet (approved):

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: done

PR #1309, opened 2023-01-17
paint: overwriting the clip's alpha/opacity

mgrudzinska:
Depending on the clip's shape, setting its alpha/opacity to zero produced different results.
Now these values are overwritten for clips in order to gain consistent results (there's no need to set the clip's color anymore).

@Issue #1192 

Proposition instead of #1296


mgrudzinska:
It's a bad solution, I'm closing it

PR #1311, opened 2023-01-20
loader: Support ABGR colorspace

JSUYA:
Since the color space is set at the time of specifying the target buffer of the canvas, there is no way to know the color space when the picture is loaded. So, check the color space applied to SwCanvas at the time of reload() and change the color space.

There is an issue of BGR color space support for each loader. The external_jpg loader resets the TJPF color space and calls read() to get a new buffer. In the case of external_png, we need to change the color value directly because it have to start over from begin_read_*.

This solution can affect performance as much as it access again image buffer that have already been `read()` done. However, this only happens once.

issue: https://github.com/Samsung/thorvg/issues/1310

### DIFF
--- a/src/lib/tvgLoadModule.h
+++ b/src/lib/tvgLoadModule.h
@@ -37,6 +37,7 @@ public:
     float vw = 0;
     float vh = 0;
     float w = 0, h = 0;         //default image size
+    uint32_t colorSpace = SwCanvas::ARGB8888;
 
     virtual ~LoadModule() {}
 
@@ -49,7 +50,7 @@ public:
 
     virtual bool read() = 0;
     virtual bool close() = 0;
-    virtual unique_ptr<Surface> bitmap() { return nullptr; }
+    virtual unique_ptr<Surface> bitmap(uint32_t colorSpace) { return nullptr; }
     virtual unique_ptr<Paint> paint() { return nullptr; }
 };
 
--- a/src/lib/tvgPictureImpl.h
+++ b/src/lib/tvgPictureImpl.h
@@ -69,6 +69,7 @@ struct Picture::Impl
     void* rdata = nullptr;            //engine data
     float w = 0, h = 0;
     bool resizing = false;
+    uint32_t rendererColorSpace = 0;
 
     ~Impl()
     {
@@ -104,7 +105,7 @@ struct Picture::Impl
                 }
             }
             free(surface);
-            if ((surface = loader->bitmap().release())) {
+            if ((surface = loader->bitmap(rendererColorSpace).release())) {
                 loader->close();
                 return RenderUpdateFlag::Image;
             }
@@ -128,6 +129,7 @@ struct Picture::Impl
 
     void* update(RenderMethod &renderer, const RenderTransform* pTransform, uint32_t opacity, Array<RenderData>& clips, RenderUpdateFlag pFlag, bool clipper)
     {
+        rendererColorSpace = renderer.colorSpace();
         auto flag = reload();
 
         if (surface) {
--- a/src/lib/tvgRender.h
+++ b/src/lib/tvgRender.h
@@ -203,6 +203,8 @@ public:
     virtual Compositor* target(const RenderRegion& region) = 0;
     virtual bool beginComposite(Compositor* cmp, CompositeMethod method, uint32_t opacity) = 0;
     virtual bool endComposite(Compositor* cmp) = 0;
+
+    virtual uint32_t colorSpace() = 0;
 };
 
 }
--- a/src/loaders/external_jpg/tvgJpgLoader.cpp
+++ b/src/loaders/external_jpg/tvgJpgLoader.cpp
@@ -37,6 +37,24 @@ void JpgLoader::clear()
     freeData = false;
 }
 
+uint32_t convertColorSpaceType(uint32_t colorSpace)
+{
+    uint32_t tjpfColorSpace = TJPF_RGBX;
+    switch (colorSpace)
+    {
+        case SwCanvas::ARGB8888:
+        case SwCanvas::ARGB8888_STRAIGHT:
+        default:
+           tjpfColorSpace = TJPF_BGRX;
+        break;
+        case SwCanvas::ABGR8888:
+        case SwCanvas::ABGR8888_STRAIGHT:
+           tjpfColorSpace = TJPF_RGBX;
+        break;
+    }
+    return tjpfColorSpace;
+}
+
 /************************************************************************/
 /* External Class Implementation                                        */
 /************************************************************************/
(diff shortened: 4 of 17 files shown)


bchrescionko:
I checked it on svg test file and now it works fine, colors are appropriate for different color spaces.

src/lib/tvgPictureImpl.h
```diff
@@ -89,7 +89,7 @@ struct Picture::Impl
         return ret;
     }
 
-    uint32_t reload()
+    uint32_t reload(int32_t colorSpace)
```
hermet, line 92: Suggest setting the default value here.
uint32_t reload(uint32_t colorSpace = -1)

src/loaders/external_jpg/tvgJpgLoader.h
```diff
@@ -48,6 +50,7 @@ class JpgLoader : public LoadModule
     unsigned char *image = nullptr;
     unsigned long size = 0;
     bool freeData = false;
+    uint32_t tjpfColorSpace = TJPF_BGRX;
```
hermet, line 53: I think a new variable is not necessary because Loader already has a colorSpace variable.
you can add a local converter function and immediately call it when it's necessary (see mentioned the above comment)

src/loaders/external_jpg/tvgJpgLoader.cpp
```diff
@@ -149,16 +148,32 @@ bool JpgLoader::close()
 }
 
 
-unique_ptr<Surface> JpgLoader::bitmap()
+unique_ptr<Surface> JpgLoader::bitmap(int32_t colorSpace)
 {
     if (!image) return nullptr;
+    if (colorSpace != -1 && static_cast<uint32_t>(colorSpace) != this->colorSpace) {
```
hermet, line 154: probably you can make a separate converter that returns converted value. In this case, you can call it immediately when it's necessary, no need an additional member variable.
```
uint32_t convert(uint32_t colorSpace)
{
   uint32 tjpfColorSpace;
    ...
   return tjpfColorSpace;
}
```

src/loaders/external_png/tvgPngLoader.cpp
```diff
@@ -110,16 +130,20 @@ bool PngLoader::close()
     return true;
 }
 
-unique_ptr<Surface> PngLoader::bitmap()
+unique_ptr<Surface> PngLoader::bitmap(int32_t colorSpace)
 {
     if (!content) return nullptr;
+    if (colorSpace != -1 && static_cast<uint32_t>(colorSpace) != this->colorSpace) {
```
hermet, line 136: This converting compatibility looks quite fragile. At least you can add a description here(other loaders as well) what it does and how this must be taken over the possible coming other colorspace types.

src/loaders/external_png/tvgPngLoader.h
```diff
@@ -37,11 +37,12 @@ class PngLoader : public LoadModule
...
     bool close() override;
 
-    unique_ptr<Surface> bitmap() override;
+    unique_ptr<Surface> bitmap(int32_t colorSpace) override;
 
 private:
     png_imagep image = nullptr;
-    const uint32_t* content = nullptr;
+    uint32_t* content = nullptr;
+    int pngFormat = PNG_FORMAT_BGRA;
```
hermet, line 45: we can remove this as well??

src/loaders/external_png/tvgPngLoader.cpp
```diff
@@ -97,6 +116,7 @@ bool PngLoader::read()
         free(buffer);
         return false;
     }
+    if (content) free((void*)content);
```
hermet, line 119: this looks an unrelated change... does this correct?

src/loaders/external_jpg/tvgJpgLoader.cpp
JSUYA: thank you for your advice. i added convert function.

src/loaders/external_png/tvgPngLoader.cpp
JSUYA: removed

src/lib/tvgRender.h
```diff
@@ -37,7 +37,7 @@ struct Surface
     uint32_t* buffer;
     uint32_t  stride;
     uint32_t  w, h;
-    uint32_t  cs;
+    uint32_t  colorSpace;
```
hermet, line 40: This is actually out of this goal. Next time please commit an individual refactoring change before the original mission.

hermet:
Unnecessary commits are welcomed. I'd like to suggest you to use "git rebase --interactive" to update the commits in the existing PR. Otherwise all commits will be squashed at the end.

src/lib/tvgLoadModule.h
```diff
@@ -49,7 +50,7 @@ class LoadModule
 
     virtual bool read() = 0;
     virtual bool close() = 0;
-    virtual unique_ptr<Surface> bitmap() { return nullptr; }
+    virtual unique_ptr<Surface> bitmap(bool changeColorSpace) { return nullptr; }
```
hermet, line 53: Thinking about this... 
changeColorSpace is too ambiguous. In the end, the requested colorspace value has to be specified. 

How about this?
```
enum Colorspace { RGBA, ... };
class Coverter { 
  static RGBA2ABGR();
  ...
 };  //in tvgRender.h
...
LoadModule::bitmap(Colorspace cs);
```

src/loaders/external_png/tvgPngLoader.cpp
```diff
@@ -42,6 +42,24 @@ static void _premultiply(uint32_t* data, uint32_t w, uint32_t h)
 }
 
 
+static inline uint32_t CHANGE_COLORSPACE(uint32_t c)
```
hermet, line 45: We don't need this duplicate logic if you consider the general Converter mentioned above.

hermet (changes requested):
More feedbacks. please check it. thank you.

src/lib/tvgRender.h
JSUYA: reverted the code

src/lib/tvgLoadModule.h
JSUYA: I changed the bool changeColorSpace to the colorSpace value and compare it inside the loader.
The colorspace change used now is ARGB <-> ABGR.
ARGBtoABGR or ABGRtoARGB have the same logic, so I think, The current logic doesn't have enough functionality to make it a class.
Some of the common functions can be moved to the header, but I think it is in the same context as _premultiply(), and after this PR, we can work on a separate PR to organize these.

hermet: Ok, actually both are spagetties. Back then I allowed it because I have no better idea at that time. Even now. (Do you?) However colorspace converter is different. changeColorSpace() with out any specification is too ambugiuous because there will be more colorspace formats such as RGB, ARGB, BGR, BGRA, GRAYSCALE, etc. 

It's my opinion. Overall, I agree on your contibution code, anyway it helps ThorVG better than before. But I would like to let you know there is a possibibility to revise the whole base code.

hermet (approved):
thanks.

PR #1315, opened 2023-01-28
clipper refactoring

hermet:
Keep small size & neat apis.

The main focus is to reduce the paint size and remove an api at the out of user demand.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -366,20 +366,6 @@ public:
     CompositeMethod composite(const Paint** target) const noexcept;
 
     /**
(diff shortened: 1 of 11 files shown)


hermet:
@mgrudzinska could you please verify this?

mgrudzinska:
@hermet, sorry for late replay. I'll take a look today

hermet:
@mgrudzinska please don't be pushed yourself. i don't think this is a serious patch.

mgrudzinska:
@hermet kinda late, but I've read the code and checked, works fine

PR #1318, opened 2023-02-02
svg_loader: handling svgs without the viewBox attribs

mgrudzinska:
the first approach was made in #1248 but since many changes have been made since then,  a new pr is opened.

this solves #1239 and hopefully https://github.com/godotengine/godot/issues/72478

### DIFF
--- a/src/lib/tvgPicture.cpp
+++ b/src/lib/tvgPicture.cpp
@@ -98,6 +98,7 @@ Result Picture::size(float w, float h) noexcept
 Result Picture::size(float* w, float* h) const noexcept
 {
     if (!pImpl->loader) return Result::InsufficientCondition;
+    pImpl->reload();
     if (w) *w = pImpl->w;
     if (h) *h = pImpl->h;
     return Result::Success;
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -3186,6 +3186,12 @@ SvgLoader::~SvgLoader()
 
 void SvgLoader::run(unsigned tid)
 {
+    //According to the SVG standard the value of the width/height of the viewbox set to 0 disables rendering
+    if (renderingDisabled) {
+        root = Scene::gen();
+        return;
+    }
+
     if (!simpleXmlParse(content, size, true, _svgLoaderParser, &(loaderData))) return;
 
     if (loaderData.doc) {
@@ -3316,7 +3322,7 @@ bool SvgLoader::read()
     if (((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Viewbox) &&
         (fabsf(vw) <= FLT_EPSILON || fabsf(vh) <= FLT_EPSILON)) {
         TVGLOG("SVG", "The <viewBox> width and/or height set to 0 - rendering disabled.");
-        return false;
+        renderingDisabled = true;
     }
 
     TaskScheduler::request(this);
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -831,20 +831,32 @@ static bool _attrParseSvgNode(void* data, const char* key, const char* value)
 
     if (!strcmp(key, "width")) {
         doc->w = _toFloat(loader->svgParse, value, SvgParserLengthType::Horizontal);
+        doc->viewFlag = (SvgViewFlag)((uint32_t)doc->viewFlag | (uint32_t)SvgViewFlag::Width);
     } else if (!strcmp(key, "height")) {
         doc->h = _toFloat(loader->svgParse, value, SvgParserLengthType::Vertical);
+        doc->viewFlag = (SvgViewFlag)((uint32_t)doc->viewFlag | (uint32_t)SvgViewFlag::Height);
     } else if (!strcmp(key, "viewBox")) {
         if (_parseNumber(&value, &doc->vx)) {
             if (_parseNumber(&value, &doc->vy)) {
                 if (_parseNumber(&value, &doc->vw)) {
-                    _parseNumber(&value, &doc->vh);
-                    loader->svgParse->global.h = doc->vh;
+                    if (_parseNumber(&value, &doc->vh)) {
+                        doc->viewFlag = (SvgViewFlag)((uint32_t)doc->viewFlag | (uint32_t)SvgViewFlag::Viewbox);
+                        loader->svgParse->global.h = doc->vh;
+                    }
+                    loader->svgParse->global.w = doc->vw;
                 }
-                loader->svgParse->global.w = doc->vw;
+                loader->svgParse->global.y = doc->vy;
             }
-            loader->svgParse->global.y = doc->vy;
+            loader->svgParse->global.x = doc->vx;
         }
-        loader->svgParse->global.x = doc->vx;
+        if (((uint32_t)doc->viewFlag & (uint32_t)SvgViewFlag::Viewbox) && (doc->vw < 0.0f || doc->vh < 0.0f)) {
+            doc->viewFlag = (SvgViewFlag)((uint32_t)doc->viewFlag & ~(uint32_t)SvgViewFlag::Viewbox);
+            TVGLOG("SVG", "Negative values of the <viewBox> width and/or height - the attribute invalidated.");
+        }
+        if (!((uint32_t)doc->viewFlag & (uint32_t)SvgViewFlag::Viewbox)) {
+            loader->svgParse->global.x = loader->svgParse->global.y = 0.0f;
+            loader->svgParse->global.w = loader->svgParse->global.h = 1.0f;
+        }
     } else if (!strcmp(key, "preserveAspectRatio")) {
         _parseAspectRatio(&value, &doc->align, &doc->meetOrSlice);
     } else if (!strcmp(key, "style")) {
@@ -1289,22 +1301,22 @@ static SvgNode* _createSvgNode(SvgLoaderData* loader, SvgNode* parent, const cha
     if (!loader->svgParse->node) return nullptr;
     SvgDocNode* doc = &(loader->svgParse->node->node.doc);
 
-    loader->svgParse->global.w = 0;
-    loader->svgParse->global.h = 0;
+    loader->svgParse->global.w = 1.0f;
+    loader->svgParse->global.h = 1.0f;
 
     doc->align = AspectRatioAlign::XMidYMid;
     doc->meetOrSlice = AspectRatioMeetOrSlice::Meet;
+    doc->viewFlag = SvgViewFlag::None;
     func(buf, bufLength, _attrParseSvgNode, loader);
 
-    if (loader->svgParse->global.w == 0) {
-        if (doc->w < FLT_EPSILON) loader->svgParse->global.w = 1;
-        else loader->svgParse->global.w = doc->w;
+    if (!((uint32_t)doc->viewFlag & (uint32_t)SvgViewFlag::Viewbox)) {
+        if ((uint32_t)doc->viewFlag & (uint32_t)SvgViewFlag::Width) {
+            loader->svgParse->global.w = doc->w;
+        }
+        if ((uint32_t)doc->viewFlag & (uint32_t)SvgViewFlag::Height) {
+            loader->svgParse->global.h = doc->h;
+        }
     }
-    if (loader->svgParse->global.h == 0) {
-        if (doc->h < FLT_EPSILON) loader->svgParse->global.h = 1;
-        else loader->svgParse->global.h = doc->h;
-    }
-
     return loader->svgParse->node;
 }
 
@@ -3192,7 +3204,7 @@ void SvgLoader::run(unsigned tid)
 
         _updateStyle(loaderData.doc, nullptr);
     }
-    root = svgSceneBuild(loaderData.doc, vx, vy, vw, vh, w, h, align, meetOrSlice, svgPath);
+    root = svgSceneBuild(loaderData.doc, vx, vy, vw, vh, w, h, align, meetOrSlice, svgPath, viewFlag);
 }
 
 
@@ -3205,28 +3217,37 @@ bool SvgLoader::header()
     if (!loaderData.svgParse) return false;
 
     loaderData.svgParse->flags = SvgStopStyleFlags::StopDefault;
+    viewFlag = SvgViewFlag::None;
 
     simpleXmlParse(content, size, true, _svgLoaderParserForValidCheck, &(loaderData));
 
     if (loaderData.doc && loaderData.doc->type == SvgNodeType::Doc) {
-        //Return the brief resource info such as viewbox:
-        vx = loaderData.doc->node.doc.vx;
-        vy = loaderData.doc->node.doc.vy;
-        w = vw = loaderData.doc->node.doc.vw;
-        h = vh = loaderData.doc->node.doc.vh;
-
-        //Override size
-        if (loaderData.doc->node.doc.w > 0) {
-            w = loaderData.doc->node.doc.w;
-            if (vw < FLT_EPSILON) vw = w;
-        }
-        if (loaderData.doc->node.doc.h > 0) {
-            h = loaderData.doc->node.doc.h;
-            if (vh < FLT_EPSILON) vh = h;
-        }
-
+        viewFlag = loaderData.doc->node.doc.viewFlag;
         align = loaderData.doc->node.doc.align;
         meetOrSlice = loaderData.doc->node.doc.meetOrSlice;
+        w = 1.0f;
+        h = 1.0f;
+
+        //Return the brief resource info such as viewbox:
+        if ((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Width) {
+            w = loaderData.doc->node.doc.w;
+        }
+        if ((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Height) {
+            h = loaderData.doc->node.doc.h;
+        }
+        //Override size
+        if ((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Viewbox) {
+            vx = loaderData.doc->node.doc.vx;
+            vy = loaderData.doc->node.doc.vy;
+            vw = loaderData.doc->node.doc.vw;
+            vh = loaderData.doc->node.doc.vh;
+
+            if (!((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Width)) w = vw;
+            if (!((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Height)) h = vh;
+        } else {
+            vw = w;
+            vh = h;
+        }
     } else {
         TVGLOG("SVG", "No SVG File. There is no <svg/>");
         return false;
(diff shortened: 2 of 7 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3294,6 +3317,8 @@ bool SvgLoader::read()
 
     TaskScheduler::request(this);
 
+    if (!isBBoxGiven) this->done();
+
```
mgrudzinska, line 3321: @hermet @JSUYA 
this is an important change comparing to the #1248 
in case the isBBoxGiven == false, the vBox of the svg is known only after the whole file is handled.

hermet: @mgrudzinska please remove the draft if you are ready to get a review.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -788,3 +788,16 @@ unique_ptr<Scene> svgSceneBuild(SvgNode* node, float vx, float vy, float vw, flo
 
     return root;
 }
+
+void applyBBoxFlag(const Scene* scene, float& vx, float& vy, float& vw, float& vh, float& w, float& h, BBoxFlag bboxFlag)
```
mgrudzinska, line 792: The only reason I added this function is because I didn;t want to change the loader members (vx, vy, vw, vh) from the svgSceneBuild function. calling a dedicated function seems to be more readable, but still, it should be placed in some common space I think, but maybe you have some better idea in general how to deal with no viewBox info than bboxflag

src/loaders/svg/tvgSvgLoader.h
```diff
@@ -51,6 +51,7 @@ class SvgLoader : public LoadModule, public Task
     unique_ptr<Paint> paint() override;
 
 private:
+    BBoxFlag bboxFlag;
```
JSUYA, line 54: How about initializing it to `BBoxFlag::None`?

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -146,6 +146,27 @@ enum class SvgParserLengthType
...
+};
+inline BBoxFlag operator|(BBoxFlag b1, BBoxFlag b2) {
+    return (BBoxFlag)((uint32_t)b1 | (uint32_t)b2);
+}
+inline BBoxFlag operator&(BBoxFlag b1, BBoxFlag b2) {
+    return (BBoxFlag)((uint32_t)b1 & (uint32_t)b2);
+}
+inline BBoxFlag& operator|=(BBoxFlag& b1, BBoxFlag b2) {
+    return (BBoxFlag&)((uint32_t&)b1 |= (uint32_t)b2);
+}
```
JSUYA, line 168: nice try! I don't know if it's appropriate to declare it here. Haha.. I expect that the readability will be much better if the existing masking codes are handled like this.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -788,3 +788,16 @@ unique_ptr<Scene> svgSceneBuild(SvgNode* node, float vx, float vy, float vw, flo
...
+{
+    if ((bboxFlag & BBoxFlag::Viewbox) != BBoxFlag::Viewbox) {
```
JSUYA, line 794: nit: BBoxFlag::Viewbox is full masking, so the code below is same.
`if (bboxFlag != BBoxFlag::Viewbox)`

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -788,3 +788,16 @@ unique_ptr<Scene> svgSceneBuild(SvgNode* node, float vx, float vy, float vw, flo
...
+        scene->bounds(nullptr, nullptr, &vw, &vh, false);
+        vx = 0.0f;
+        vy = 0.0f;
+        if ((bool)(bboxFlag & BBoxFlag::Width)) vw = w;
+        if ((bool)(bboxFlag & BBoxFlag::Height)) vh = h;
+    }
+    if (!(bool)(bboxFlag & BBoxFlag::Width)) w = vw;
+    if (!(bool)(bboxFlag & BBoxFlag::Height)) h = vh;
```
JSUYA, line 802: nit: It is recommended to use c++ style casts. (static_cast)

JSUYA (commented):
I'm not sure BBox is an appropriate name.
`BBoxFlag` enum is for masking the definition state(?) for width/height and viewbox.
And width/height and ViewboxW/ViewboxH are the same value, and there is no separate logic to distinguish them.
What do you think?
(I think there is no problem with the overall logic.)

mgrudzinska:
@JSUYA first of all I'm really very sorry for such a late answer. I have to read my code from the beginning and I guess you'll have to the the same... sorry

> I'm not sure BBox is an appropriate name. `BBoxFlag` enum is for masking the definition state(?) for width/height and viewbox. And width/height and ViewboxW/ViewboxH are the same value, and there is no separate logic to distinguish them. What do you think? (I think there is no problem with the overall logic.)

Thx for this note! since there was a bug - Width and ViewboxW shouldn't be the same. great that you've caught that :)
 
 I'll made changes asap (this week) and push them

mgrudzinska:
1.
> I'm not sure BBox is an appropriate name. `BBoxFlag` enum is for masking the definition state(?) for width/height and viewbox. 
I've changed the name BBox -> SvgViewFlag (since it has flags regarding both, the viewbox and the viewport).

2.
> And width/height and ViewboxW/ViewboxH are the same value, and there is no separate logic to distinguish them. What do you think? (I think there is no problem with the overall logic.)
It was a bug, they are not the same. Although I've changed the enum now - only if all of the viewBox params are correct the flag is set. 

3.
```
inline BBoxFlag operator|(BBoxFlag b1, BBoxFlag b2) {
    return (BBoxFlag)((uint32_t)b1 | (uint32_t)b2);
}
inline BBoxFlag operator&(BBoxFlag b1, BBoxFlag b2) {
    return (BBoxFlag)((uint32_t)b1 & (uint32_t)b2);
}
inline BBoxFlag& operator|=(BBoxFlag& b1, BBoxFlag b2) {
    return (BBoxFlag&)((uint32_t&)b1 |= (uint32_t)b2);
}
```
For now I used the same logic that we have for all other flags. I do agree that the readability has dropped, but the code is consistent now. This change I'll push in a separate pr and for all of the flags (or these that need this). 
Also I've used C-style casting, as it is for the other flags - separate pr, maybe after the above is applied.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -831,20 +834,32 @@ static bool _attrParseSvgNode(void* data, const char* key, const char* value)
...
+                loader->svgParse->global.y = doc->vy;
             }
-            loader->svgParse->global.y = doc->vy;
+            loader->svgParse->global.x = doc->vx;
+        }
+        if (((uint32_t)doc->viewFlag & (uint32_t)SvgViewFlag::Viewbox) && (doc->vw < 0.0f || doc->vh < 0.0f)) {
+            doc->viewFlag = (SvgViewFlag)((uint32_t)doc->viewFlag & ~(uint32_t)SvgViewFlag::Viewbox);
+            TVGLOG("SVG", "Negative values of the <viewBox> width and/or height - the attribute invalidated.");
+        }
+        if (!((uint32_t)doc->viewFlag & (uint32_t)SvgViewFlag::Viewbox)) {
```
mgrudzinska, line 859: cases like ```viewBox="0 0 200"``` - this should be ignored and wasn't before

mgrudzinska:
@hermet @JSUYA I hope that this time I covered in tests all of the cases...

[tests.tar.gz](https://github.com/thorvg/thorvg/files/10946876/tests.tar.gz)

results for master:
Zrzut ekranu 2023-03-11 o 00 46 26

results after applying this pr:
Zrzut ekranu 2023-03-11 o 00 43 40

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: removed for now -> todo in a separate pr

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: it was my bug, doesn;t exist after changes

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3192,7 +3217,8 @@ void SvgLoader::run(unsigned tid)
 
         _updateStyle(loaderData.doc, nullptr);
     }
-    root = svgSceneBuild(loaderData.doc, vx, vy, vw, vh, w, h, align, meetOrSlice, svgPath);
+    root = svgSceneBuild(loaderData.doc, vx, vy, vw, vh, w, h, align, meetOrSlice, svgPath, viewFlag);
+    applySvgViewFlag(root.get(), vx, vy, vw, vh, w, h, viewFlag);
```
hermet, line 3221: How about put this call in svgSceneBuild inside()? otherwise we don't need to expose the method.

hermet (commented):
Thanks. please have a look at my comments.

capnm:
> the first approach was made in #1248 but since many changes have been made since then, a new pr is opened.
> 
> this solves #1239 and hopefully [godotengine/godot#72478](https://github.com/godotengine/godot/issues/72478)

There are still issues in images without the _width/height_ atribute, The change makes it recognized in applications like Godot, but it cannot be used because of the clipping. See https://github.com/godotengine/godot/issues/72478 

texture_01
Gnome Image viewer :
grafik

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -77,6 +77,9 @@
 #define PX_PER_MM 3.779528f //1 in = 25.4 mm -> PX_PER_IN/25.4
 #define PX_PER_CM 37.79528f //1 in = 2.54 cm -> PX_PER_IN/2.54
 
+#define DEFAULT_VIEWPORT_WIDTH  300.0f
+#define DEFAULT_VIEWPORT_HEIGHT 150.0f
```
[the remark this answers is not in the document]
mgrudzinska: it's not so simple - please check my last comment. I.ve changed the approach and removed these values.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -831,20 +834,32 @@ static bool _attrParseSvgNode(void* data, const char* key, const char* value)
...
-            loader->svgParse->global.y = doc->vy;
+            loader->svgParse->global.x = doc->vx;
+        }
+        if (((uint32_t)doc->viewFlag & (uint32_t)SvgViewFlag::Viewbox) && (doc->vw < 0.0f || doc->vh < 0.0f)) {
+            doc->viewFlag = (SvgViewFlag)((uint32_t)doc->viewFlag & ~(uint32_t)SvgViewFlag::Viewbox);
+            TVGLOG("SVG", "Negative values of the <viewBox> width and/or height - the attribute invalidated.");
+        }
+        if (!((uint32_t)doc->viewFlag & (uint32_t)SvgViewFlag::Viewbox)) {
+            loader->svgParse->global.x = loader->svgParse->global.y =
+            loader->svgParse->global.w = loader->svgParse->global.h = 0;
```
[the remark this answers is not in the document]
mgrudzinska: I've changed the def values: ```loader->svgParse->global.w = loader->svgParse->global.h = 1.0;``` in the ```_createSvgNode```. so if there is no reason to change them, they are already correct.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3174,6 +3194,11 @@ SvgLoader::~SvgLoader()
 
 void SvgLoader::run(unsigned tid)
 {
+    if (((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Viewbox) && (fabsf(vw) < FLT_EPSILON || fabsf(vh) < FLT_EPSILON)) {
```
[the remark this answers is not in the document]
mgrudzinska: I moved the log to the header, but this check is still here, so I can return from the func. 
I don;t want to return false in a heder, since then the ```load``` api returns InvalidArguments.
is it what you meant?

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: ```if (!isBBoxGiven) this->done();``` - removed, also after changes it wasn;t necessary any more

applySvgViewFlag -> internal fun of the scene biulder, but now svgSceneBuild changes the vx, vy, vw, vh values

mgrudzinska:
@hermet and @capnm as I wrote in #1248 : 
> 1. from https://www.w3.org/TR/SVG2/geometry.html#Sizing:
>    "The value auto for width and height on the svg element is treated as 100%."
> 2. from (for css) https://svgwg.org/specs/integration/#svg-css-sizing:
>    "When SVG element is the top level element, the [width](https://svgwg.org/svg2-draft/geometry.html#Sizing) and [height](https://svgwg.org/svg2-draft/geometry.html#Sizing) are set to '100%'
>    To resolve 'auto' value on ‘[svg](https://svgwg.org/svg2-draft/struct.html#SVGElement)’ element if the ‘[viewBox](https://svgwg.org/svg2-draft/coords.html#ViewBoxAttribute)’ attribute is not specified:
> 
> * Use author specified [width](https://svgwg.org/svg2-draft/geometry.html#Sizing) and [height](https://svgwg.org/svg2-draft/geometry.html#Sizing) ‘[svg](https://svgwg.org/svg2-draft/struct.html#SVGElement)’ element attributes.
> * If any of the sizing attributes are missing, resolve the missing ‘[svg](https://svgwg.org/svg2-draft/struct.html#SVGElement)’ element [width](https://svgwg.org/svg2-draft/geometry.html#Sizing) to '300px' and missing [height](https://svgwg.org/svg2-draft/geometry.html#Sizing) to '150px' (using [CSS 2.1 replaced elements size calculation](https://www.w3.org/TR/CSS21/visudet.html#inline-replaced-width)).
>   To resolve 'auto' value on ‘[svg](https://svgwg.org/svg2-draft/struct.html#SVGElement)’ element if the ‘[viewBox]"
> 
> 3. from https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/width#svg:
> Default value:	auto (treated as 100%)

so I think that the def values of the viewBox (300x150) aren't the svg standard, it's rather the css std. so I made some changes. it should be useful for godot now.

but:
- vx and vy are still 0 - I remember that there where cases in which this was a problem (shield and field svgs) - I need to read about that in the svg std one more time...
- for the UserSpaceOnUse coord sys some attribs have to be refreshed (like grads)

so after we decide the final version of this pr, I'll open (or not:) ) the issues regarding the above

src/loaders/svg/tvgSvgLoader.cpp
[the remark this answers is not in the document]
hermet: ok

src/loaders/svg/tvgSvgLoader.cpp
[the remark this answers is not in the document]
hermet: No one would not understand these values if these are out of standards. That's why I suggest you to leave any useful comments in the code for the maintanence - what brings you to use these specific values and if any fragiles when adjust the values.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -755,21 +754,35 @@ static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox,
 }
 
 
+void _applySvgViewFlag(const Scene* scene, float& vx, float& vy, float& vw, float& vh, float& w, float& h, SvgViewFlag viewFlag)
```
hermet, line 757: static void

src/loaders/svg/tvgSvgLoader.cpp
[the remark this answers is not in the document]
hermet: If you argue the condition is "Success" then it's ok otherwise InvalidArguments would be better or we can return some else.(you can define and suggest)

Regardless of that, there is no need to run a thread since we could know the condition prior to run() method.
Maybe you can check this at read() instead.

src/loaders/svg/tvgSvgLoader.cpp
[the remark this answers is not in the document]
mgrudzinska: I removed them, since they seems to be a css std, not svg.

src/loaders/svg/tvgSvgLoader.cpp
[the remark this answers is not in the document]
mgrudzinska: I moved this check to read, now Result::Unknown is returned + log

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: done

mgrudzinska:
@hermet I still need to resolve this (I know what's happening, I'll solve this asap)
https://github.com/godotengine/godot/issues/72478#issuecomment-1465735893

hermet:
@mgrudzinska LGTM. please have @JSUYA confirm if possible. Also please squash 3 naugty commits instead rebase & merge. thanks!

mgrudzinska:
@hermet please taka a look on the last commit. I needed this to solve the issue described in (the size 1x1) https://github.com/godotengine/godot/issues/72478#issuecomment-1465735893

@JSUYA @capnm 
Please take a look and comment if necessary

src/lib/tvgLoadModule.h
```diff
@@ -36,7 +36,7 @@ class LoadModule
     float vy = 0;
     float vw = 0;
     float vh = 0;
-    float w = 0, h = 0;         //default image size
+    float w = 1.0f, h = 1.0f;         //default image size
```
hermet, line 39: You should not change the base policy for SVG. If necessary, please arrange the default values in SvgLoader.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3292,6 +3314,9 @@ bool SvgLoader::read()
 {
     if (!content || size == 0) return false;
 
+    if (((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Viewbox) &&
+        (fabsf(vw) <= FLT_EPSILON || fabsf(vh) <= FLT_EPSILON)) return false;
```
hermet, line 3318: IMHO, If this is a wierd case, we could print TVGLOG for hints.

hermet (changes requested):
tvgLoadModule shouldn't be changed

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: done

src/lib/tvgLoadModule.h
mgrudzinska: done

hermet (approved):
LGTM thanks!

mgrudzinska:
@hermet the additional commit contains changes that we discussed, regarding width and height of the viewbox = 0. nothing changed in the previous one, that was already accepted

hermet (approved):
LGTM by me, also welcome @JSUYA's opinion.

src/lib/tvgPicture.cpp
```diff
@@ -98,6 +98,7 @@ Result Picture::size(float w, float h) noexcept
 Result Picture::size(float* w, float* h) const noexcept
 {
     if (!pImpl->loader) return Result::InsufficientCondition;
+    pImpl->reload();
```
JSUYA, line 101: Should reload() called on size() and viewbox() getters?
Is there any overhead from this?

If the issue is the size when the user calls load() and immediately uses the size(), viewbox() getter
It seems that we can call resize() after reading in load() function.

JSUYA (approved):
I am so sorry tooooo late. 
I agree about the changes related to the loader's viewbox.
I left a comment regarding the reload() used for Picture class.
If there is no problem, please merge this. :)

src/lib/tvgPicture.cpp
mgrudzinska: yes, if a user calls size/viewbox after load in a 'normal' case the information are already available, since header() was already called. but for svgs without width/height and viewbox info, this information is available after run() is finished, that's why I need to call reload here, to be sure that run is done.

Issue #1320, opened 2023-02-07
Wrong colorspace in static png loader


JSUYA:
`meson build -Dloaders="all" -Dexamples=true -Dstatic=true`
Running the PicturePng example with a static png loader outputs an image with the wrong colorspace applied.

<img src="https://github.com/JSUYA/jsuya.github.io/blob/master/thorvg_picturepng.png?raw=true/">

JSUYA:
The color_type for lode png is `LCT_RGBA`. so the output looks like the example above.
We can merge https://github.com/thorvg/thorvg/pull/1311 and add a patch that fixes this.

JSUYA:
https://github.com/thorvg/thorvg/pull/1321

PR #1321, opened 2023-02-10
png_loader(static): Fix the colorspace of an image with an alpha channel

JSUYA:
Set colorspace to ABGR when colortype of lodepng is LCT_RGBA.
Since an image without an alpha channel becomes an ARGB colorspace with LCT_RGB,
it is the same as the default colorspace.

related issue : https://github.com/thorvg/thorvg/issues/1320

### DIFF
--- a/src/loaders/png/tvgPngLoader.cpp
+++ b/src/loaders/png/tvgPngLoader.cpp
@@ -213,5 +213,7 @@ void PngLoader::run(unsigned tid)
 
     lodepng_decode(&image, &width, &height, &state, data, size);
 
+    if (state.info_png.color.colortype == LCT_RGBA) colorSpace = SwCanvas::ABGR8888;
+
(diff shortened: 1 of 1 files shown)


src/loaders/png/tvgPngLoader.cpp
```diff
@@ -213,5 +213,7 @@ void PngLoader::run(unsigned tid)
 
     lodepng_decode(&image, &width, &height, &state, data, size);
 
+    if (state.info_png.color.colortype == LCT_RGBA) colorSpace = SwCanvas::ABGR8888;
```
hermet, line 216: If possible the values must be updated earlier. So it looks open() is a better place to put?

JSUYA: I move the code. thanks

hermet (approved):

PR #1322, opened 2023-03-09
svg_loader: fix zero width/height handling

mgrudzinska:
Svgs with width/height attrib set to zero
were rendered although they shouldn't have
been.

sample svg file:
```
<svg viewBox="0 0 10 10" height="100" width="0">
    <path stroke="none" fill="red" d="M0 0 L 100 0 100 100 0 100 0 0"/>
    <path fill="none" stroke="#00ff00" stroke-width="4" d="M0 0 L 100 100"/>
</svg>
```

expected result - empty canvas
master branch result:
Zrzut ekranu 2023-03-9 o 01 32 42


JSUYA (approved):
LGTM

mgrudzinska:
this is handled in a more general way in #1318

PR #1323, opened 2023-03-09
svg_loader: improved safety while style node freeing

patrykka:


mgrudzinska:
@patryk could you describe the problem?

["If ptr is a null pointer, the function does nothing."](https://en.cppreference.com/w/cpp/memory/c/free)

hermet:
Literally this seems doens't helpful.

patrykka:
@mgrudzinska  @hermet 
I observed crash in one of tests. 
#4  __GI___libc_free (mem=0x80) at /usr/src/debug/glibc-2.30/malloc/malloc.c:3117
       ar_ptr = <optimized out>
       p = <optimized out>
       hook = 0x0
#5  0xa8c993d0 in _freeNodeStyle (style=0xb1c358) at /usr/src/debug/thorvg-0.7.9/builddir/../src/loaders/svg/tvgSvgLoader.cpp:2886
No locals.

But Mira was right about std::free descripton, I wonder if that pointers are marked NULL here.

mgrudzinska:
@patrykka since I'm not reproducing could you please provide more details ?
also, we are talking about thorvg (updated master branch) and not about tizenvg, right?

hermet:
@patrykka the memory was broken outside of the pointer. maybe style itself or others. so this change won't be helpful.

hermet (changes requested):
please check the above comment.

patrykka:
Thank you. Closing this PR.

PR #1324, opened 2023-03-13
svg_loader: fix memory leak

patrykka:


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2837,6 +2837,7 @@ static void _svgLoaderParserXmlCssStyle(SvgLoaderData* loader, const char* conte
         length -= next - content;
         content = next;
 
+        free(node->id);
         free(tag);
```
hermet, line 2841: It looks wrong position to free it. I believe @mgrudzinska could help this.

mgrudzinska: @patrykka as far as I see the node (and everything what it contains, like id) is freed in the ```_freeNode```, but I may be missing something. could you please provide an example in which you observed mem leak?

mgrudzinska (changes requested):

mgrudzinska:
@patrykka for such a standard case:
```
<svg viewBox="0 0 10 10" xmlns="http://www.w3.org/2000/svg">
  <style id="idStyle">
    circle.mineStyle {
      fill: gold;
      stroke: maroon;
      stroke-width: 2px;
    }
  </style>

  <circle id="idCircle" cx="5" cy="5" r="4" class="mineStyle"/>
</svg>
```
with your change the same memory is freed twice (AddressSanitizer: heap-use-after-free)

PR #1325, opened 2023-03-13
sw_enigne texmap: imporved safety

patrykka:


src/lib/sw_engine/tvgSwRasterTexmap.h
```diff
@@ -292,6 +292,8 @@ static AASpans* _AASpans(float ymin, float ymax, const SwImage* image, const SwB
 
     aaSpans->lines = static_cast<AALine*>(calloc(height, sizeof(AALine)));
 
+    if (!aaSpans->lines) return nullptr;
```
hermet, line 295: Either wya, this is fault since aaSpans memory could be dangled. Current tvg engines don't catch-up any OOM scenario. That's our policy. So please just leave it.

hermet (changes requested):

src/lib/sw_engine/tvgSwRasterTexmap.h
patrykka: Ok, thank you

PR #1326, opened 2023-03-13
common Accessor: fix uninitialized field

patrykka:


src/lib/tvgAccessor.cpp
```diff
@@ -70,11 +75,11 @@ unique_ptr<Picture> Accessor::set(unique_ptr<Picture> picture, function<bool(con
 
 Accessor::~Accessor()
 {
-
+    delete(pImpl);
 }
 
 
-Accessor::Accessor()
+Accessor::Accessor() : pImpl(new Impl())
```
hermet, line 82: Good catch but there is no point to alloc memory here...
I suggest you to apply _TVG_DISABLE_CTOR() instead _TVG_DECLARE_PRIVATE()
have a look at Accessor class.

hermet (changes requested):

hermet:
@patrykka  https://github.com/thorvg/thorvg/commit/5ebde2c9b682fd7bc69cb3d6f14eacb6b6c30e1b

PR #1330, opened 2023-03-16
svg_loader: handling zero width/height viewbox

mgrudzinska:
contains #1318


mgrudzinska:
@hermet I've opened this as a separate pr since it's more a question. 
For svgs with the width and/or height value set to zero rendering is disabled in #1318 - the load api returns Result:Unknown
and draw - Result::InsufficientCondition.
In this approach an empty scene is added, so that both, load and draw, return Result::Success.
it seems to be more intuitive.

also I'd do exactly the same for width=0 and height=0, despite it's not in the svg standard. but for now there is no way to change the size in such cases, so in my opinion the proposition is reasonable

mgrudzinska:
consistent with #1318

PR #1333, opened 2023-03-21
svg_loader: invalid strokes' width set to zero

mgrudzinska:
In case the "stroke" attrib is set to "none",
the width of the stroke is set to zero. Thanks
to that it isn't taken into account while
establishing the viewbox boundaries for svgs
without a viewbox and width/height information.

```
<svg>
      <linearGradient id="Gradient" x1="0" x2="50%" y1="3" y2="1" gradientUnits="userSpaceOnUse">
      <stop offset="0%" stop-color="red" />
      <stop offset="50%" stop-color="black" stop-opacity="0" />
      <stop offset="100%" stop-color="green" />
    </linearGradient>

  <path fill="blue" d="M 0 0 h 400 v 400 z"/>
  <path stroke="pink" stroke-width="80" fill="green" d="M -100 400  h 400 v -300 z"/>
    <path stroke="none" fill="url(#Gradient)" d="M0 0 L100 0 100 100 0 100 0 0"/>
    <path fill="none" stroke="#00ff00" stroke-width="80" d="M0 0 L 100 100"/>
</svg>
```

before:
Zrzut ekranu 2023-03-22 o 00 19 03

after:
Zrzut ekranu 2023-03-22 o 00 20 10

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -346,7 +346,7 @@ static void _applyProperty(SvgNode* node, Shape* vg, const Box& vBox, const stri
 
     //If stroke property is nullptr then do nothing
     if (style->stroke.paint.none) {
-        //Do nothing
+        vg->stroke(0.0f);
     } else if (style->stroke.paint.gradient) {
         Box bBox = vBox;
         if (!style->stroke.paint.gradient->userSpace) bBox = _boundingBox(vg);


JSUYA (approved):
Thank you for your work.
I've checked that this works fine.
```
    <path stroke="none" fill="url(#Gradient)" stroke-width="1000" d="M0 0 L100 0 100 100 0 100 0 0"/>
```

+)
Gradient is displayed wrong.
It seems to be another problem that occurs when there is no size of svg.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -757,7 +757,9 @@ static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox,
 static void _applySvgViewFlag(const Scene* scene, float& vx, float& vy, float& vw, float& vh, float& w, float& h, SvgViewFlag viewFlag)
 {
     if (!((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Viewbox)) {
-        scene->bounds(nullptr, nullptr, &vw, &vh, false);
+        scene->bounds(&vx, &vy, &vw, &vh, false);
```
hermet, line 760: stroke and viewbox looks having no relavent..
Curious two fixes are in one.

mgrudzinska: right. this change moved to #1334 (with a change)

hermet (approved):

PR #1334, opened 2023-03-24
svg_loader: fixing viewbox width/height

mgrudzinska:
For svg without any viewbox and width/height
information the established width and height
values of the viewbox should be calculated
starting in the 0,0 point.

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -67,7 +67,7 @@ struct Box
 };
 
 static bool _appendShape(SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath);
-static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox, const string& svgPath, bool mask, int depth, bool* isMaskWhite = nullptr);
+static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox, Box& bBox, const string& svgPath, bool mask, int depth, bool* isMaskWhite = nullptr);
 
 
 static inline bool _isGroupType(SvgNodeType type)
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -757,7 +757,9 @@ static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox,
 static void _applySvgViewFlag(const Scene* scene, float& vx, float& vy, float& vw, float& vh, float& w, float& h, SvgViewFlag viewFlag)
 {
     if (!((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Viewbox)) {
-        scene->bounds(nullptr, nullptr, &vw, &vh, false);
+        scene->bounds(&vx, &vy, &vw, &vh, false);
```
hermet, line 760: This looks leading counter dependency... bounds() depends on svg loader result.

hermet: @mgrudzinska Thank you for fixing it. From a design point of view, calling scene->bounds() is not ideal for the reasons mentioned above. I have reviewed it twice, nonetheless, I believe that applying original patch is better since we now understand what we are doing. What are your thoughts? You can leave this patch for the future refactoring. Apologies for my previous hasty comment.

mgrudzinska: @hermet no problem, there were so many things to think of in this 'no viewbox and width/height' issue, that as tthis point I was confused what was applied, what wasn;t... sorry for that, thx for your all revs.
I separated the first commit from this pr, so we can aplly it and I'll close this one - I don;t like hanging prs, if I'll need this one I can find it in closed once.

mgrudzinska:
first commit present in #1344 , second one not needed (after considerations)

PR #1335, opened 2023-03-24
picture: paint and surface reloading separation

mgrudzinska:
I'd rather use https://github.com/thorvg/thorvg/pull/1338

For svg files without any viewbox and widht/height
information it has to be ensured, that the loading
process is finished before any api like size/viewbox/bounds
is called.

### DIFF
--- a/src/lib/tvgPictureImpl.h
+++ b/src/lib/tvgPictureImpl.h
@@ -90,24 +90,30 @@ struct Picture::Impl
         return ret;
     }
 
+    bool reloadPaint()
+    {
+        if (loader && !paint) {
+            if (auto p = loader->paint()) {
+                paint = p.release();
+                loader->close();
+                if (w != loader->w || h != loader->h) {
+                    if (!resizing) {
+                        w = loader->w;
+                        h = loader->h;
+                    }
+                    loader->resize(paint, w, h);
+                    resizing = false;
+                }
+                if (paint) return true;
+            }
+        }
+        return false;
+    }
+
     uint32_t reload()
     {
         if (loader) {
-            if (!paint) {
-                if (auto p = loader->paint()) {
-                    paint = p.release();
-                    loader->close();
-                    if (w != loader->w || h != loader->h) {
-                        if (!resizing) {
-                            w = loader->w;
-                            h = loader->h;
-                        }
-                        loader->resize(paint, w, h);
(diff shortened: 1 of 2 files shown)


src/lib/tvgPictureImpl.h
```diff
@@ -205,6 +205,7 @@ struct Picture::Impl
             if (w) *w = max.x - min.x;
             if (h) *h = max.y - min.y;
         } else {
+            reload();
```
hermet, line 208: Thinking about a better approach...

Two considerations by me.
1. Now we have no more async beneifts by recent reload() calls.
2. Unecessary reload for bitmap images...

mgrudzinska: right, this has to be changed, asap

mgrudzinska: @hermet I made some changes, but still, I don;t like this solution too much, it's not cleat in the code what is happening and why. but I need it, because if an svg file doesn;t have any viewbox and width/height information, we're taking these values from the bounding box of all its paints. so to get this information up to date, we need to ensure, that the loading process is completed. 

I wanted to handle this very specific case in the svg loader, but you didn;t agree back then - maybe we didn't understand each other - the [discussion](https://github.com/thorvg/thorvg/pull/1318#discussion_r1133008968).

please take a look at https://github.com/thorvg/thorvg/pull/1338 - this was +- the original proposition how to solve this problem.

src/lib/tvgPictureImpl.h
```diff
@@ -90,24 +90,30 @@ struct Picture::Impl
...
-                        if (!resizing) {
-                            w = loader->w;
-                            h = loader->h;
-                        }
-                        loader->resize(paint, w, h);
-                        resizing = false;
+        if (loader && !paint) {
+            if (auto p = loader->paint()) {
+                paint = p.release();
+                loader->close();
```
JSUYA, line 98: #1338 forces reload, so reload doesn't seem necessary. I think loader->w, h are already updated. Can't we just update this->w, h?

src/lib/tvgPictureImpl.h
mgrudzinska: I htink it's much easier when this is handled internaly in the svg loader

mgrudzinska:
solved in #1338

Issue #1327, opened 2023-03-14
It looks like you need a null check for image->rle  in tvgSwRaster.cpp


JoogabYun:
Looking at the code below, 

**if image->rle is null** in _rasterRGBAImage, _transformedRGBAImage is called.

_rasterTexmapPolygon is called on _transformedRGBAImage.

In _rasterTexmapPolygon, **imge->rle is null**. If you check image->rle->size == 0 here, it seems to be a problem. isn't it?

```c++
static bool _rasterRGBAImage(SwSurface* surface, SwImage* image, const Matrix* transform, const SwBBox& region, uint32_t opacity)
{
    //RLE Image
    if (image->rle) {
        ...
    //Whole Image
    } else {
        ...
        else return _transformedRGBAImage(surface, image, transform, region, opacity); // this image->rle is must be null
    }
}

static bool _transformedRGBAImage(SwSurface* surface, const SwImage* image, const Matrix* transform, const SwBBox& region, uint32_t opacity)
{
    ...
     return _rasterTexmapPolygon(surface, image, transform, &region, opacity, _alpha); // this image->rle is must be null
    ...
}

static bool _rasterTexmapPolygon(SwSurface* surface, const SwImage* image, const Matrix* transform, const SwBBox* region, uint32_t opacity, uint32_t (*blendMethod)(uint32_t))
{
    //Exceptions: No dedicated drawing area?
    if (!region && image->rle->size == 0) return false; // Maybe the problem is because image->rle is null here, isn't it?
   ...
}
```

hermet:
@JoogabYun Thanks for pointing it out, it sounds like an exceptional case. probably you can submit a null check there or wait for fixing it in a week.

PR #1337, opened 2023-03-25
sw_engine: fix safety check

mgrudzinska:
An incorrect check introduced in bf07eb11f25909d2c73aa7e5baeb01a80af384d0 as a solution of the #1327 issue. As a consequence rotated images were not drawn. Fixed.


hermet (approved):
oh thanks!

PR #1338, opened 2023-03-26
svg_loader: forcing file loading completion

mgrudzinska:
Svgs without any viewbox and width/height information have to be loaded before any other action is taken. This is necessary to get the valid sive/viewbox info.

I'd use this instead of #1335 and extra reload() calls added in the picture code


JSUYA (approved):
I think this way is more efficient than reloading every time in the viewbox() api.

hermet (approved):
make sense. thanks.

PR #1342, opened 2023-04-01
sw_engine: Fix one pixel blank

### LINKED ISSUE #1329: Incorrect fill result when two sub path sharing one edge but has different direction or winding number
```
auto path = tvg::Shape::gen();

  path->moveTo(10, 10);
  path->lineTo(100, 10);
  path->lineTo(10, 100);
  path->close();

  path->moveTo(100, 10);
  path->lineTo(10, 100);
  path->lineTo(100, 100);
  path->close();

  path->fill(0, 255, 0, 255);
```
The result in thorvg is shown below:
image

RuiwenTang:
This PR fix #1329 

If pixel covered by two edge with same slope with different direction. The `area` value is zero, but the winding number is not.


hermet:
@RuiwenTang thanks for your patch. 

I think enhancing this feature may be a bit more complex because the current TVG renderer handles closed shapes with separate drawing units, which is different from the triangle-list mesh composition in OpenGL and D3D. As a result, there may be visible anti-aliasing gaps between the shapes, especially in a rotated version (see the following figure). I've considered a proper and efficient model to handle this, but it's not clear yet. Perhaps the TVG canvas could merge the closed shapes into one paint before moving on to rendering, although this may not be efficient for general cases. Instead, I would suggest that users request a single rectangle directly, as it would be much cheaper! Any opinions?

shot-2023-04-02_00-42-42

RuiwenTang:
Yes,  this problem is more complicated than I thought, I prefoer merge these closed pathes. I'm going to dig a little deeper into this issue

hermet:
@RuiwenTang I mentioned earlier that we need the shapes-merge step prior to the rasterization stage. If we include this fix, it could become redundant. I don't want to leave any dead code that will be difficult to maintain in the future. The problem is that if we change the height of the box (triangles), the patch won't work, as shown in the attached figure. This could confuse users and make them think it's a bug, which would hurt the consistency of the behavior. Currently, we don't support merged shapes. However, if you argue that this fix is necessary for your project, we can temporarily compromise.

thorvg

RuiwenTang:
Ok,and currently I am working on merge these closed shapes. I have implemented a similar algorithm before in my local project.

hermet:
we have the merging shapes function in https://github.com/thorvg/thorvg/commit/b26672a1f6c20718d8ba329c849bce465cc59365

RuiwenTang:
@hermet I pushed a sweep line algorithm. I used this algorithm in my own engine to do polygon triangulation and complex polygon clip. I hope this code can help to fix this case. 
The result is showing below, the outline is rendered on second column:
# the one pixel blank case I wish to fix:
image
# Path with intersection:
image
# Path with intersection and different filltype
image
image

RuiwenTang:
Here is a simple sample code:
```c++
std::array<std::unique_ptr<tvg::Shape>, 2> gen_shape() {
  std::array<std::unique_ptr<tvg::Shape>, 2> ret{};

  // stroke
  ret[0] = tvg::Shape::gen();
  ret[0]->moveTo(100, 10);
  ret[0]->lineTo(40, 180);
  ret[0]->lineTo(190, 60);
  ret[0]->lineTo(10, 60);
  ret[0]->lineTo(160, 180);
  ret[0]->fill(tvg::FillRule::EvenOdd);
  ret[0]->close();
  ret[0]->stroke(0, 255, 0, 255);
  ret[0]->stroke(3.f);
  // fill
  ret[1] = tvg::Shape::gen();
  ret[1]->moveTo(100, 10);
  ret[1]->lineTo(40, 180);
  ret[1]->lineTo(190, 60);
  ret[1]->lineTo(10, 60);
  ret[1]->lineTo(160, 180);
  ret[1]->close();
  ret[1]->fill(0, 255, 0, 255);
  ret[1]->fill(tvg::FillRule::EvenOdd);

  return ret;
};

auto paths = gen_shape();


  auto outline = paths[0]->extractOutline();

  outline->translate(200, 100);
  outline->stroke(0, 255, 0, 255);
  outline->stroke(3.f);

  canvas->push(std::move(paths[0]));

  paths[1]->translate(0, 200);

  canvas->push(std::move(paths[1]));

  canvas->push(std::move(outline));
```

hermet:
Thank you, @RuiwenTang, for your contribution.

1. ThorVG can apply this feature after the v0.9 release.
2. Could you provide more information about the algorithm you applied here? Do you have any reference links or sources for the code? We would like to double-check for any licensing issues.
3. I believe this change will require a significant amount of polishing to be smoothly integrated into the ThorVG source code. If you are interested, we could go through additional review stages. Otherwise, you can leave it to us.

Againk thanks, you did really good job.

RuiwenTang:
The basic scanline idea comes from this video: https://www.youtube.com/watch?v=qkhUNzCGDt0&t=293s .

Some data struct and geometric calculation in this code is inspired by SKIA: https://github.com/google/skia/blob/main/src/gpu/ganesh/geometry/GrTriangulator.cpp 
Under BSD-3 license. 

I made a lot of changes to make the code simple and can do more things like extract the boundary or do advance polygon boolean operation. I can share these changes if thorvg needs these feature.

RuiwenTang:
image
take some time to rebase my branch and  test the curve case :)

PR #1343, opened 2023-04-01
picture: remove unnecessary reloads

mgrudzinska:
After #1338 is applied there is no need to reload
the pictures while aksing about its viewbox or its size.


mgrudzinska:
sorry guys, this change was made in one of my prs, but since we've changed the approach it wasn't merged and I forgot to add this change to the applied one...

hermet (approved):
thanks!

Issue #1332, opened 2023-03-21
svg loader: viewbox of svgs without any viewbox information


mgrudzinska:
it's partially regarding the https://github.com/godotengine/godot/issues/72478. 
It has to be confirmed whether svgs without any viewbox and widht/height information should be clipped starting in 0,0 point or the whole image should be visible and no clipping should be applied.

mgrudzinska:
I'm not 100% sure, but I think this is the answer : https://www.w3.org/TR/SVG2/geometry.html#XProperty
the default value of x and y attributes is zero. I don't see any reason why this should be changed in case no viewbox and width/height information is given in the svg file.

capnm:
I think, at least from the user's point of view, it should be handled like _auto_, that is, treated as a 100% image size.
https://www.w3.org/TR/SVG2/geometry.html#Sizing

mgrudzinska:
result for now (master + #1344)

```
<svg>
    <path stroke-width="100" stroke="red" fill="green" d="M-50 100 v 550 h 600 Z M 600 -50 h -600 v 500 Z"/>
</svg>
```
<img width="324" alt="Zrzut ekranu 2023-04-2 o 14 24 51" src="https://user-images.githubusercontent.com/67589014/229352708-dc8be839-1b0c-4d30-ab13-ea1b8d09c5e7.png">


as you can see the image is cut, so that it starts in 0,0. That's the whole image:

<img width="375" alt="Zrzut ekranu 2023-04-2 o 14 31 14" src="https://user-images.githubusercontent.com/67589014/229352981-3d2dd253-9548-4966-b718-cea45d15e0e0.png">

@capnm as I asked in one of the Godot issues - do you think that the whole image should be visible? these are very specific cases, I'm not sure how often do they appear and what is the expected result from the user perspective...

capnm:
Ah sorry, now I understand what you meant... The cut at 0,0 unfortunately makes most of these images unusable. The main problem is the thickness of a path touching the 0,0 axes, you can only see the half of the thickness. You can take a look at the svg's in https://kenney.nl/assets/pattern-pack-2 for example.

![grafik](https://user-images.githubusercontent.com/4047289/229371853-3e8628b5-e498-4b80-bb87-76fda4fa9d1a.png)

mgrudzinska:
@hermet @junsu could you please comment?
[...]

PR #1344, opened 2023-04-01
svg_loader: fixing viewbox width/height

mgrudzinska:
For svgs without any viewbox and width/height
information the established width and height
values of the viewbox should be increased
by its initial x and y values - the reason is
that the viewbox starting point is forced to
be (0,0).

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -757,9 +757,7 @@ static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox,
 static void _applySvgViewFlag(const Scene* scene, float& vx, float& vy, float& vw, float& vh, float& w, float& h, SvgViewFlag viewFlag)
 {
     if (!((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Viewbox)) {
-        scene->bounds(nullptr, nullptr, &vw, &vh, false);
-        vx = 0.0f;
-        vy = 0.0f;
+        scene->bounds(&vx, &vy, &vw, &vh, false);
         if ((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Width) vw = w;
         if ((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Height) vh = h;
     }


mgrudzinska:
changes present in one of the commits in #1334

hermet (approved):

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -757,7 +757,9 @@ static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox,
 static void _applySvgViewFlag(const Scene* scene, float& vx, float& vy, float& vw, float& vh, float& w, float& h, SvgViewFlag viewFlag)
 {
     if (!((uint32_t)viewFlag & (uint32_t)SvgViewFlag::Viewbox)) {
-        scene->bounds(nullptr, nullptr, &vw, &vh, false);
+        scene->bounds(&vx, &vy, &vw, &vh, false);
+        vw += vx;
```
mgrudzinska, line 761: removing these 4 lines:
```
        vw += vx;
        vh += vy;
        vx = 0.0f;
        vy = 0.0f;
```
ensures that whole images without viewbox and width/height info are drawn

JSUYA (approved):

src/loaders/svg/tvgSvgSceneBuilder.cpp
JSUYA: I already pressed approve, but please check again
I tested this pr using godot [test image.](https://www.kenney.nl/assets/board-game-icons)
As tested on this page, the images didn't look aligned. (https://github.com/thorvg/thorvg/issues/1332)
image
I don't know what i missed.

Anyway, if my understanding is correct
After applying this patch, all svg images with no viewbox defined will be aligned to 0,0.
In ubuntu-imageviewer, chrome, or firefox, the results don't sorted by 0,0.
Is there an svg viewer example that shows this sorted by 0,0? (Not an image editing tool..)

(I don't think there is a standard(?) definition for this.(I cant find it) so.,.  maybe we can support this.)

mgrudzinska: @JSUYA as I mentioned in #1332, I forced x=y=0. but I'm not sure what should be drawn in such cases. and I agree that since we are not sure, and godot need whole images, lets solve it. If someone will point out that this is an error, we'll take care of that

Issue #1340, opened 2023-03-31
Unsupported attribute: paint-order


mgrudzinska:
```
<svg xmlns:svg="http://www.w3.org/2000/svg" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 400 400" >
<path style="fill:#4140ff;stroke:#ff00ff;stroke-width:19;paint-order:markers stroke fill" d="m 100,100 h 200 v 200 z "/>
</svg>
```

found [here](https://github.com/godotengine/godot/pull/75508#pullrequestreview-1366082740)

JSUYA:
I think, we can change the paint order by extracting the stroke and fill raster code of SwRenderer::renderShape() as each function and changing the order in Shape::Impl::render(). it seems that Shape's API needs to be added for this. (for use in the loader)
This is not likely to be a commonly used API. (we will probably have to use BETA_API.)
Any good idea?

mgrudzinska:
@JSUYA fyi #1346 (still draft - I need to refactor a bit)
this api will be more useful after we implement markers :)

hermet:
fyi, Marker is out of current thorvg feature. it's neither out of SVG tiny spec.

mgrudzinska:
solved in https://github.com/thorvg/thorvg/pull/1371

PR #1346, opened 2023-04-05
common: order(bool strokeFirst) api introduced

mgrudzinska:
The new api is introduced to handle the rendering
order of a stroke and a fill. By default fill is
rendered as the first one, stroke as the second
one. This order can be reversed by calling
order(true).

@issue: https://github.com/thorvg/thorvg/issues/1340

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -977,6 +977,18 @@ public:
      */
     Result fill(FillRule r) noexcept;
 
+
+    /**
+     * @brief Sets the rendering order of the stroke and the fill.
+     *
+     * @param[in] strokeFirst If @c true the stroke is rendered before the fill, otherwise the stroke is rendered as the second one (the default option).
+     *
+     * @return Result::Success when succeed, Result::FailedAllocation otherwise.
+     * @BETA_API
+     */
+    Result order(bool strokeFirst) noexcept;
+
+
     /**
      * @brief Gets the commands data of the path.
      *
--- a/src/lib/tvgRender.h
+++ b/src/lib/tvgRender.h
@@ -94,6 +94,7 @@ struct RenderStroke
     uint32_t dashCnt = 0;
     StrokeCap cap = StrokeCap::Square;
     StrokeJoin join = StrokeJoin::Bevel;
+    bool strokeFirst = false;
 
     ~RenderStroke()
     {
--- a/src/lib/tvgShape.cpp
+++ b/src/lib/tvgShape.cpp
@@ -288,6 +288,14 @@ const Fill* Shape::fill() const noexcept
 }
 
 
+Result Shape::order(bool strokeFirst) noexcept
+{
+    if (strokeFirst && !pImpl->strokeFirst()) return Result::FailedAllocation;
+
+    return Result::Success;
+}
+
+
 Result Shape::stroke(float width) noexcept
 {
     if (!pImpl->strokeWidth(width)) return Result::FailedAllocation;
--- a/src/lib/sw_engine/tvgSwRenderer.cpp
+++ b/src/lib/sw_engine/tvgSwRenderer.cpp
@@ -77,7 +77,7 @@ struct SwShapeTask : SwTask
     bool clipper = false;
 
     void run(unsigned tid) override
-    {    
+    {
         if (opacity == 0 && !clipper) return;  //Invisible
 
         uint8_t strokeAlpha = 0;
@@ -117,7 +117,7 @@ struct SwShapeTask : SwTask
                    shape outline below stroke could be full covered by stroke drawing.
                    Thus it turns off antialising in that condition.
                    Also, it shouldn't be dash style. */
-                auto antiAlias = (strokeAlpha == 255 && rshape->strokeWidth() > 2 && rshape->strokeDash(nullptr) == 0) ? false : true;
+                auto antiAlias = strokeAlpha < 255 || rshape->strokeWidth() <= 2 || rshape->strokeDash(nullptr) > 0 || (rshape->stroke && rshape->stroke->strokeFirst);
 
                 if (!shapeGenRle(&shape, rshape, antiAlias)) goto err;
             }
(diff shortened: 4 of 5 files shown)


inc/thorvg.h
```diff
@@ -1021,6 +1034,17 @@ class TVG_EXPORT Shape final : public Paint
      */
     FillRule fillRule() const noexcept;
 
+    /**
```
hermet, line 1037: I think the requirement is close to 0. Let's drop getter. I think it's not necessary.

inc/thorvg.h
```diff
@@ -977,6 +977,19 @@ class TVG_EXPORT Shape final : public Paint
...
+     *
+     * @param[in] fill Number specifying the order in which a fill will be rendered [0 ~ 1]. Rendered first by default (0).
+     * @param[in] stroke number specifying the order in which a stroke will be rendered [0 ~ 1]. Rendered second by default (1).
+     *
+     * @retval Result::Success When succeed.
+     * @retval Result::InvalidArguments If any of the arguments has the same value as another or is outside the range 0-1.
+     *
+     * @BETA_API
+     */
+    Result order(uint8_t fill, uint8_t stroke) noexcept;
```
hermet, line 991: I think multiple parameters are ambugious to use. even the value's range.

Let's simplify this. How about this?

Result stroke(bool above == true);    //stroke order. if true, draw stroke above the shape fill. otherwise, draw stroke under the shape fill.

hermet (commented):
+ quick interface feedbacks.

mgrudzinska:
In the renderer the 'above' is a property of a stroke, since we decided to implement it this way.
the the svg loader the 'paint-order' is a paint property since that how it is in the svg std.
I think it's ok, just highlighting it to you

inc/thorvg.h
hermet: First, I think when ideas are conflicted, I'd like to recommend you to consensus first then code.
Second, float / bool are obviously different types. any issues of it?

hermet: So I asked you a question an actual issue of it.
Were there any complie error? warning? what problems would occur when you call it?

mgrudzinska: sorry, I didn't fully understand what you wrote.
yes, calling stroke(4) will cause an error: ```error: call to member function 'stroke' is ambiguous```
if we can live with that, I'll remove the enum and write as you suggested, stroke(bool).

hermet: ok, thanks for info. When i tested it with msvc it didn't have any issues but gcc has different result.

//fill above stroke. 
So... What do you think?
Result Shape::fill(bool above) noexcept;

mgrudzinska: @hermet please check now. I applied changes

src/lib/tvgShape.cpp
```diff
@@ -288,6 +288,14 @@ const Fill* Shape::fill() const noexcept
 }
 
 
+Result Shape::fill(bool above) noexcept
+{
+    pImpl->rs.fillUnder = !above;
```
mgrudzinska, line 293: fillUnder not a fillAbove, because I didn;t want to negate here:
```if (task->rshape->fillUnder) {```

File: src/loaders/svg/tvgSvgLoaderCommon.h [hunk 4]
```diff
@@ -394,6 +395,7 @@ struct SvgStyleProperty
     SvgColor color;
     bool curColorSet;
     char* cssClass;
+    bool paintOrder; //true if default (fill, stroke), false otherwise
```
mgrudzinska, line 398: ```isPaintOrderDefault``` ? a bit long, but more descriptive

hermet:
Here, you could create two atomic patches:

1. A feature addition at common/engine.
2. An SVG enhancement through the new API.

Good programmers always create patches to get reviewed easier.

src/loaders/svg/tvgSvgLoaderCommon.h
hermet: Remember this, "When in Rome, do as the Romans do"

But if you think it's necessary then please do it.

src/lib/tvgRender.h
```diff
@@ -119,6 +119,7 @@ struct RenderShape
     RenderStroke *stroke = nullptr;
     uint8_t color[4] = {0, 0, 0, 0};    //r, g, b, a
     FillRule rule = FillRule::Winding;
+    bool fillUnder = true;
```
hermet, line 122: I'd like put this in the stroke structure since it's stroking dependency feature so far. it could save the memory.

hermet (changes requested):
Please search for the word 'antiAlias' in tvgSwRenderer.cpp. TVG intentionally disables anti-aliasing when stroking hides the shape edges, so it should not be turned on in the above scenario.

src/lib/tvgRender.h
mgrudzinska: but then:
```
    shape->appendRect(0, 0, 200, 200, 0, 0);
    shape->fill(255, 255, 0, 255);
    shape->fill(true);       //useless, but user can call it
```
user called only the fill apis, no stroke will be drawn, fill(true) is useless.
but because of fill(true) is called I have to create a stroke just to put there the information, that won't be used.

if the api is shape->stroke(true/false) then it's ok to create a stroke - a user should be aware, that calling some 'stroke' api will create a stroke. but calling a fill api that creates a stroke is strange...

can we introduce a different name, like shape->order(bool strokeUnder) ? then it's not so confusing.

hermet: @mgrudzinska 

shape->order(true) //useless, but user can call it  
shape->fill(true) //userless but user can call it.
shape->stroke(true) //useless but user can call it.

Nothing different.

Only a few thoughtless programmers would call fill(true) without using stroke. It's best to ignore such cases as they are not critical and cannot be covered.

hermet: @mgrudzinska let's go with this:

Result order(bool strokeFirst) noexcept;        //drawing order

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: moved to #1371

src/lib/tvgShape.cpp
```diff
@@ -288,6 +288,14 @@ const Fill* Shape::fill() const noexcept
 }
 
 
+Result Shape::order(bool strokeFirst) noexcept
+{
+    if (strokeFirst && !pImpl->strokeFirst()) return Result::FailedAllocation;
```
hermet, line 293: it doesn't take cover the false scenario?

hermet (changes requested):
please check a comment.

hermet:
duplicated: https://github.com/thorvg/thorvg/pull/1371

Issue #1336, opened 2023-03-24
Bug in SVG path stroke drawing causes horns to appear


AlfishSoftware:
This SVG renders incorrectly:
```svg
<svg id="svg" width="64" height="64" xmlns="http://www.w3.org/2000/svg" viewBox="320 192 64 64"><path d="m 365.51212,202.55632 C 362.29764,202.26797 360.28835,203.80976 360,206 l -2,16 v -18 c 0,-2.20914 -1.79086,-4 -4,-4 -2.20914,0 -4,1.79086 -4,4 v 18 l -2,-16 c -0.28835,-2.19024 -2.29765,-3.73203 -4.48788,-3.44368 l -10 30 h 40 z" fill="red" stroke="blue" stroke-linejoin="round" stroke-width="4"/></svg>
```
It should render like this (works fine in browser, Ubuntu Image Viewer and in other tools like NanoSVG):
![thor-breaker](https://user-images.githubusercontent.com/101953341/227435896-b04d50c5-32c9-4028-840f-8e81c5df24bf.svg)
But in ThorVG it's buggy. This is from ThorVG Viewer:
![image](https://user-images.githubusercontent.com/101953341/227434703-1ada0813-c40a-424e-bf65-fc28f9e2ba28.png)
Notice the creepy evil horns of despair.
It's consistently appearing in all my shapes with this particular exact curve. Changing `stroke-linejoin` or `stroke-width` has no effect on the bug, so it must be caused by something in the path calculations. Perhaps some division by zero or similar? Or maybe it's drawing strokes using circle slices for some reason, with a distant center and incorrect fill? No idea.

JSUYA:
Hi @hermet 
```cpp
    auto shape1 = tvg::Shape::gen();

    //Appends Paths
    shape1->moveTo(200,200 );
    shape1->lineTo(200,400);
    shape1->lineTo(200+30,200);

    shape1->close();
    shape1->fill(0, 0, 255, 255);
    shape1->stroke(40);
    shape1->stroke(255, 0, 0, 255);;
    //shape1->stroke(tvg::StrokeJoin::Round);
    if (canvas->push(move(shape1)) != tvg::Result::Success) return;

    auto shape2 = tvg::Shape::gen();

    //Appends Paths
    shape2->moveTo(40,20 );
    shape2->lineTo(40,40);
    shape2->lineTo(40+3,20);

    shape2->close();
    shape2->fill(0, 0, 255, 255);
    shape2->stroke(4);
[...]

PR #1347, opened 2023-04-05
sw_engine: fix the inproper stroke raster.

hermet:
We have an incorrect value comparison when the stroke corner center is adjusted. All coordinates in the stroke raster have been scaled by the given values (sx, sy), while the stroke width remains at its original size due to it not being clear how it should be scaled.

This brings wierd rendering results at times. Now fix it.

@Issue: https://github.com/thorvg/thorvg/issues/1336

### DIFF
--- a/src/lib/sw_engine/tvgSwStroke.cpp
+++ b/src/lib/sw_engine/tvgSwStroke.cpp
@@ -302,7 +302,7 @@ static void _inside(SwStroke& stroke, int32_t side, SwFixed lineLength)
     bool intersect = false;
 
     /* Only intersect borders if between two line_to's and both
-       lines are long enough (line length is zero fur curves). */
+       lines are long enough (line length is zero for curves). */
     if (border->movable && lineLength > 0) {
         //compute minimum required length of lines
         SwFixed minLength = abs(mathMultiply(stroke.width, mathTan(theta)));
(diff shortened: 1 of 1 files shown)


JSUYA (approved):

src/lib/sw_engine/tvgSwStroke.cpp
```diff
@@ -382,9 +382,16 @@ static void _lineTo(SwStroke& stroke, const SwPoint& to)
     if (delta.zero()) return;
 
     //compute length of line
-    auto lineLength =  mathLength(delta);
     auto angle = mathAtan(delta);
 
+    /* The lineLength is used to determine the intersection of strokes outlines.
+       The scale needs to be reverted since the stroke width has not been scaled.
+       An alternative option is to scale the width of the stroke properly by
+       calculating the mixture of the sx/sy rating on the stroke direction. */
```
mgrudzinska, line 390: I may misunderstand the comment - isn;t the alternative option that you have mentioned done in 395-397?

hermet: No, I mean here we can scale the stroke width instead reverting lineLength scale. The problem is, stroke width is a scalar value but we have two axis scale factors from transform matrix which means we could scale it alternatively by computing the total scale value from sx, xy by measuring the direction of the stroke.

Issue #524, opened 2021-06-28
clippath doesn't work with scene


JSUYA:
Segmentation fault occurs when scene is input as composite( with clippath) argument.

```
btThread 1 "Scene" received signal SIGSEGV, Segmentation fault.
tvg::Task::done (this=0x0) at ../src/lib/tvgTaskScheduler.h:55
55	        if (!pending) return;
(gdb) bt
#0  tvg::Task::done (this=0x0) at ../src/lib/tvgTaskScheduler.h:55
#1  0x00007ffff7bc264d in tvg::SwRenderer::prepareCommon (flags=tvg::All, clips=..., opacity=255, transform=0x0, task=0x555555dcb8a0, this=0x555555d85410)
    at ../src/lib/sw_engine/tvgSwRenderer.cpp:553
#2  tvg::SwRenderer::prepareCommon (this=0x555555d85410, task=0x555555dcb8a0, transform=0x0, opacity=255, clips=..., flags=tvg::All) at ../src/lib/sw_engine/tvgSwRenderer.cpp:542
#3  0x00007ffff7bbcad0 in tvg::Shape::Impl::update (pFlag=<optimized out>, clips=..., opacity=<optimized out>, transform=<optimized out>, renderer=..., this=0x555555d854f0)
    at ../src/lib/tvgShapeImpl.h:234
#4  tvg::PaintMethod<tvg::Shape::Impl>::update (this=<optimized out>, renderer=..., transform=<optimized out>, opacity=<optimized out>, clips=..., flag=<optimized out>)
    at ../src/lib/tvgPaint.h:142
#5  0x00007ffff7bba1ff in tvg::Paint::Impl::update (this=0x555555d854c0, renderer=..., pTransform=pTransform@entry=0x0, opacity=<optimized out>, opacity@entry=255, clips=..., 
    pFlag=pFlag@entry=127) at ../src/lib/tvgPaint.cpp:231
#6  0x00007ffff7bbb87c in tvg::Scene::Impl::update (flag=<optimized out>, clips=..., opacity=<optimized out>, transform=0x0, renderer=..., this=0x555555d841e0)
    at ../src/lib/tvgSceneImpl.h:75
#7  tvg::PaintMethod<tvg::Scene::Impl>::update (this=<optimized out>, renderer=..., transform=0x0, opacity=<optimized out>, clips=..., flag=<optimized out>)
    at ../src/lib/tvgPaint.h:142
[...]

PR #1355, opened 2023-04-09
Support Scene ClipPath

hermet:
https://github.com/thorvg/thorvg/issues/524


hermet:
@mgrudzinska  if you have any svg issues which expect scene clipping, please test them with applying this.

Issue #1209, opened 2022-04-03
Incorrect results when converting SVG to PNG


qarmin:
![Screenshot from 2022-04-03 10-44-56](https://user-images.githubusercontent.com/41945903/161419682-224a2216-908b-4b82-a93b-9ee980e79c20.png)

[Example.zip](https://github.com/Samsung/thorvg/files/8404102/Example.zip)

mgrudzinska:
2224 - text is not supported 
2229 - after #1214 it's better, but not perfect, since tvg doesn't support filters
2684 - filters are not supported

hermet:
1550, 1552, 2371, 2909 - confirm fixed
2192, 3323 - gradient filling of body is a bit different
2224, 2229, 2684 - no plan to support it yet

hermet:
@qarmin fixed 1550, 1552, 2371, 2909 in https://github.com/Samsung/thorvg/releases/tag/v0.8.1

JSUYA:
> 2224, 2229, 2684 - no plan to support it yet

details
2224 - text
2229 - feColorMatrix filter
2684 - feGaussianBlur filter

+) Confirm fixed with #1213

hermet:
Still need to check:
2192, 3323 - gradient filling of body is a bit different

capnm:
2961 fixed

hermet:
@capnm thanks for confirm it!

capnm:
@hermet 
2192 - has still this gradient-inverted issue

hermet:
Yes, I just checked. it. Thanks.

hermet:
2192 test result

![image](https://user-images.githubusercontent.com/3711518/232265732-8253dd84-58b7-4f62-a7d7-58bfde3e0734.png)

@capnm tvg & major browsers have the same result.

hermet:
Created a separate issue for the filter enhancement : https://github.com/thorvg/thorvg/issues/1367

capnm:
> Ah confused. 2192 is also fixed in the master branch.

The _gradientTransform="matrix(0.5,0,0,-0.48276,-69.03301,54.17094)"_ seems to be here not supported?

```svg
<linearGradient
       id="a-3"
       y1="68.419998"
       y2="-3.454"
       x1="3.1159999"
       x2="61.688999"
       gradientUnits="userSpaceOnUse"
       gradientTransform="matrix(0.5,0,0,-0.48276,-69.03301,54.17094)">
      <stop
[...]

PR #1356, opened 2023-04-09
svg_loader: gradient inheritance implemented 

mgrudzinska:
Till now only the grad stops were inherited. Now all
grad attribs are.
Inheritance from different grad type is not supported.

@issue: https://github.com/thorvg/thorvg/issues/1209

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1891,6 +1891,7 @@ static constexpr struct
     {"height", SvgParserLengthType::Vertical, sizeof("height"), offsetof(SvgUseNode, h)}
 };
 
+
 static void _cloneNode(SvgNode* from, SvgNode* parent, int depth);
 static bool _attrParseUseNode(void* data, const char* key, const char* value)
 {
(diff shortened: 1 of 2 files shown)


mgrudzinska:
split into two separate commits to make it easier to read

hermet (approved):
looks good to me. added some minor comments. @JSUYA's approval is welcomed.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2698,6 +2569,367 @@ static GradientFactoryMethod _findGradientFactory(const char* name)
...
+            }
+            //GradUnits set, coord not set directly
+            if (to->userSpace == from->userSpace) continue;
+            if (gradUnitSet && !coordSet) {
+                radialTags[i].tagInheritedRecalc(loader, to->radial, to->userSpace);
+            }
+        }
+    }
+
+    if (to->stops.count == 0) _cloneGradStops(to->stops, from->stops);
```
JSUYA, line 2637: In the same context as now, can we add `stops` flag?
The reason for adding the flag is to check whether it is set to 0 or not set.
I think `to->stops.count == 0` is ambiguous.
Of course, it seem to be a no problem now. (Maybe?)

JSUYA (approved):
Declaration order changes caused a lot of diffs. It was a bit difficult to find real changes. hahaha...
LGTM :)

mgrudzinska:
> Declaration order changes caused a lot of diffs. It was a bit difficult to find real changes. hahaha... LGTM :)

That's why I made two commits - one with changed order (without any logical changes) and one with the changes important to rev

PR #1358, opened 2023-04-10
common: log msg modified

mgrudzinska:
Introducing colors for different log types,
file name and line number added (for warnings
and errors only).

### DIFF
--- a/src/lib/tvgCommon.h
+++ b/src/lib/tvgCommon.h
@@ -65,8 +65,11 @@ using namespace tvg;
 enum class FileType { Tvg = 0, Svg, Raw, Png, Jpg, Unknown };
 
 #ifdef THORVG_LOG_ENABLED
-    #define TVGLOG(tag, fmt, ...) fprintf(stdout, tag ": " fmt "\n", ##__VA_ARGS__)  //Log Message for notifying user some useful info
-    #define TVGERR(tag, fmt, ...) fprintf(stderr, tag ": " fmt "\n", ##__VA_ARGS__)  //Error Message for us to fix it
+    constexpr auto ErrorColor = "\033[31m";  //red
+    constexpr auto LogColor = "\033[32m";    //green
+    constexpr auto ResetColor = "\033[0m";   //default
+    #define TVGERR(tag, fmt, ...) fprintf(stderr, "%s" tag " (%s l.%d): " fmt "%s\n", ErrorColor, __FILE__, __LINE__, ##__VA_ARGS__, ResetColor)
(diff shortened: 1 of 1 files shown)


hermet (changes requested):
0. "log msg modified." hmm... I think it should be something like this "enhance the logging feature."
1. I think this patch shouldn't introduce the warning type.
2. please add the Issue link in the commit msg. (I believe you already know our rule.)
https://github.com/thorvg/thorvg/blob/master/CONTRIBUTING.md

src/lib/tvgCommon.h
```diff
@@ -65,8 +65,11 @@ using namespace tvg;
 enum class FileType { Tvg = 0, Svg, Raw, Png, Jpg, Unknown };
 
 #ifdef THORVG_LOG_ENABLED
-    #define TVGLOG(tag, fmt, ...) fprintf(stdout, tag ": " fmt "\n", ##__VA_ARGS__)  //Log Message for notifying user some useful info
-    #define TVGERR(tag, fmt, ...) fprintf(stderr, tag ": " fmt "\n", ##__VA_ARGS__)  //Error Message for us to fix it
+    constexpr auto ErrorColor = "\033[31m";  //red
+    constexpr auto LogColor = "\033[32m";    //green
+    constexpr auto ResetColor = "\033[0m";   //default
+    #define TVGERR(tag, fmt, ...) fprintf(stderr, "%s" tag "(%s l.%d): " fmt "%s\n", ErrorColor, __FILE__, __LINE__, ##__VA_ARGS__, ResetColor)
+    #define TVGLOG(tag, fmt, ...) fprintf(stderr, "%s" tag ": " fmt "%s\n", LogColor, ##__VA_ARGS__, ResetColor)
```
hermet, line 72: I think TVGLOG also needs FILE, LINE. any reason TVGLOG doen't contain them?

hermet (approved):
+ one more comment.

src/lib/tvgCommon.h
mgrudzinska: because I thought that log is more for the user then for us, and I'm not sure how important for a user is to see the file&line. the msg is less readable in my opinion:
```SVG (../src/loaders/svg/tvgXmlParser.cpp l.523): Unsupported attributes used [Elements type: Path][Id : path4657][Attribute: enable-background][Value: accumulate]```
vs
```SVG: Unsupported attributes used [Elements type: Path][Id : path4657][Attribute: enable-background][Value: accumulate]```

but I consulted this and apparently people prefer to see file&line :D so I changed it

hermet (approved):

Issue #1351, opened 2023-04-07
Enhance logging feature


hermet:
1. If TVG could define colors for the log message types, it would be easier to read and distinguish them.
2. Printing the log source[file/line] will provide an intuitive way to identify the issue being pointed to.

mgrudzinska:
@hermet - like this #1358?

hermet:
@mgrudzinska yeah. guess so. good job.

hermet:
@mgrudzinska right now, i'm closing this. if you think warning type is necessary, later you can suggest it.

JSUYA:
log has improved a lot!!
A small wish is to put a small symbol in front of the log.
log -> [L] . Error -> [E], (warning(?) -> [W] , info (?)-> [I])
This is a log message from another project I'm involved in.
<img src="https://github.com/JSUYA/jsuya.github.io/blob/master/log_example.png?raw=true"/>
how about this?

hermet:
We've experienced this with EFL logs as well.

As the result, I'd like to make the prefix simplify and let users focus on the REAL context because too long log message bothers to focus on the context. Even I'd like to remove the filepath(../src/lib) since we already have the domain name in the head. (or vice versa) Long messages bothers readers to read them specially when the terminal width is not enough large. 

So... Maybe. but in my opinion, people could notice the message type by color. it's just 2 types, quite simple.

hermet:
Come to think of it, the tag might be useful when user filters the log data.

mgrudzinska:
> Even I'd like to remove the filepath(../src/lib) since we already have the domain name in the head. (or vice versa)
I don;t like these long paths as well, but I spoke with a few other programmers and they said, that a good log contains a path, so... :)

hermet:
@mgrudzinska We need logical thinking and sufficient experience in our discussions. Unfortunately, we have no idea about the other programmers except you.

PR #1363, opened 2023-04-12
common: enhancement++ of the logging method

mgrudzinska:
Logging tags are marked with the corresponding
log type color, the file and the line information
in gray, and the message content in the default
color.

@Issue: https://github.com/thorvg/thorvg/issues/1351

Zrzut ekranu 2023-04-12 o 16 50 06


hermet:
I think grey, red, green might be enough to use?

mgrudzinska:
@hermet I can not print rgb on my terminal, maybe it's not supported.. no idea. So if red, green grey are ok, that's great :)

hermet (approved):
great thanks!

PR #1369, opened 2023-04-16
svg_loader: fixing parsing empty tags

mgrudzinska:
Self-closing tags (empty elements) were parsed until
the appearance of the '>' character, causing the '/'
char to be appended to the tag name. The final '/'
char should be omitted.

sample:
```
<svg
  viewBox="0 0 10 10"
  xmlns="http://www.w3.org/2000/svg"
<defs/>
  <defs>
    <linearGradient id="myGradient" gradientTransform="rotate(90)">
      <stop />
      <stop offset="95%" stop-color="red" />
    </linearGradient>
  </defs>

  <!-- using my linear gradient -->
  <circle cx="5" cy="5" r="4" fill="url('#myGradient')" />
</svg>
```

before:
Zrzut ekranu 2023-04-16 o 13 42 59
Zrzut ekranu 2023-04-16 o 13 43 20

after:
Zrzut ekranu 2023-04-16 o 13 46 01

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2976,6 +2976,7 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content, 
         //Parse the empty tag
         attrs = content;
         while ((attrs != nullptr) && *attrs != '>') attrs++;
+        if (empty) attrs--;
     }
 
     if (attrs) {


hermet (approved):

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2975,7 +2975,7 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content,
     if (!attrs) {
         //Parse the empty tag
         attrs = content;
-        while ((attrs != nullptr) && *attrs != '>') attrs++;
+        while ((attrs != nullptr) && strncmp(attrs, "/>", 2) != 0) attrs++;
```
hermet, line 2978: if attrs ends at "/", it would occur crash.

hermet (changes requested):

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: @hermet please check now - I tried to crash it, but no success :)
by definition, if ```empty == true```  the two last signs are "\>". This is ensured by the xml parser - for so called empty elements type ```SimpleXMLType::OpenEmpty``` is set, and for this type empty is set to be true. that's why it is safe to do ```attrs--```.

hermet: @mgrudzinska ok thanks.

hermet (approved):

PR #1371, opened 2023-04-16
svg_loader: paint-order attrib handled 

mgrudzinska:
@issue: https://github.com/thorvg/thorvg/issues/1340

### DIFF
--- a/src/lib/tvgShape.cpp
+++ b/src/lib/tvgShape.cpp
@@ -288,6 +288,14 @@ const Fill* Shape::fill() const noexcept
 }
 
 
+Result Shape::order(bool strokeFirst) noexcept
+{
+    if (!pImpl->strokeFirst(strokeFirst)) return Result::FailedAllocation;
+
+    return Result::Success;
+}
+
+
 Result Shape::stroke(float width) noexcept
 {
     if (!pImpl->strokeWidth(width)) return Result::FailedAllocation;
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -252,6 +252,36 @@ static SvgMaskType _toMaskType(const char* str)
 }
 
 
+//The default rendering order: fill, stroke, markers
+//If any is omitted, will be rendered in its default order after the specified ones.
+static bool _toPaintOrder(const char* str)
+{
+    uint8_t position = 1;
+    uint8_t strokePosition = 0;
+    uint8_t fillPosition = 0;
+
+    while (*str != '\0') {
+        str = _skipSpace(str, nullptr);
+        if (!strncmp(str, "fill", 4)) {
+            fillPosition = position++;
+            str += 4;
+        } else if (!strncmp(str, "stroke", 6)) {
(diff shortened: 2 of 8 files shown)


hermet:
@mgrudzinska  You can keep this PR and close this one: https://github.com/thorvg/thorvg/pull/1346 it's duplicated.

src/lib/tvgShape.cpp
```diff
@@ -288,6 +288,14 @@ const Fill* Shape::fill() const noexcept
 }
 
 
+Result Shape::order(bool strokeFirst) noexcept
+{
+    if (strokeFirst && !pImpl->strokeFirst()) return Result::FailedAllocation;
```
hermet, line 293: Please take cover the strokeFirst == false scenario.

hermet (changes requested):

src/lib/tvgShape.cpp
mgrudzinska: done

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -252,6 +252,36 @@ static SvgMaskType _toMaskType(const char* str)
...
+            str += 6;
+        } else if (!strncmp(str, "markers", 7)) {
+            str += 7;
+        } else {
+            return _toPaintOrder("fill stroke");
+        }
+    }
+
+    if (strokePosition == 0) fillPosition = position++;
+    if (strokePosition == 0) strokePosition = position++;
```
hermet, line 279: twice strokePosition. looks error?

hermet (changes requested):
could you please check a comment?

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: thx!

mgrudzinska: this step is needed, since not all paints have to be listed and in case they weren't we have to apply to them the default order

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1263,6 +1301,8 @@ static SvgNode* _createNode(SvgNode* parent, SvgNodeType type)
     node->style->stroke.join = StrokeJoin::Miter;
     node->style->stroke.scale = 1.0;
 
+    node->style->paintOrder = _toPaintOrder("fill stroke");
```
JSUYA, line 1304: I think it is better to write true than `_toPaintOrder("fill stroke")` because "fill storke" always returns `true`.

JSUYA (approved):

hermet:
good job.

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: my motivation was to make the setting the def values to be as independent as possible from the paintOrder variable. to be honest it was for me very hard to think about paint-order without markers :D I kept them in my mind all the time, that's why maybe this was more complicated than it should be

PR #1376, opened 2023-04-18
svg_loader: fixing clipper transformation

mgrudzinska:
In a case where both the clipper and the clippee
are transformed, the final clipper transformation
matrix should be calculated as the multiplication
of both matrices.

@Issue: https://github.com/thorvg/thorvg/issues/1255

KOSZ_893:
before:
Zrzut ekranu 2023-04-18 o 01 45 56

after:
Zrzut ekranu 2023-04-18 o 01 48 26

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -254,13 +254,18 @@ static void _applyComposition(Paint* paint, const SvgNode* node, const Box& vBox
             node->style->clipPath.applying = true;
 
             auto comp = Shape::gen();
-            if (node->transform) comp->transform(*node->transform);
 
             auto child = compNode->child.data;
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -263,6 +262,12 @@ static void _applyComposition(Paint* paint, const SvgNode* node, const Box& vBox
                 if (_appendChildShape(*child, comp.get(), vBox, svgPath)) valid = true;
             }
 
+            auto m = comp->transform();
```
hermet, line 265: you can move in if() scope?

mgrudzinska: of course, sorry

hermet (approved):

Issue #1381, opened 2023-04-19
compilation with Visual Studio


vtorri:
Just fyi, I compiled the lib and tools. There are some warnings though (if you are interested)

hermet:
@vtorri I have removed the warnings on my end. Hopefully, you won't encounter them either. 
https://github.com/thorvg/thorvg/pull/1384

PR #1384, opened 2023-04-20
Make it silent MSVC's trivial compilation warnings.

hermet:
@Issue: https://github.com/thorvg/thorvg/issues/1381


vtorri:
there is also a malloc which returned value is not checked

hermet:
> there is also a malloc which returned value is not checked

Currently, thorvg doesn't take care of OOM.

Issue #1382, opened 2023-04-19
Windows (mingw-w64): compilation error with tests and capi enabled


vtorri:
error:

```
[151/164] Compiling C++ object test/capi/capiUnitTests.exe.p/capiLinearGradient.cpp.obj
FAILED: test/capi/capiUnitTests.exe.p/capiLinearGradient.cpp.obj
"c++" "-Itest/capi/capiUnitTests.exe.p" "-Itest/capi" "-I../test/capi" "-Iinc" "-I../inc" "-I." "-I.." "-Isrc/bindings/capi" "-I../src/bindings/capi" "-fdiagnos
tics-color=always" "-D_FILE_OFFSET_BITS=64" "-Wall" "-Winvalid-pch" "-std=c++14" "-Os" "-g" "-DEXAMPLE_DIR=\"C:/Documents/msys2/home/vincent.torri/gitroot_64/th
orvg_vtorri/src/examples/images\"" "-DTEST_DIR=\"C:/Documents/msys2/home/vincent.torri/gitroot_64/thorvg_vtorri/test/images\"" -MD -MQ test/capi/capiUnitTests.e
xe.p/capiLinearGradient.cpp.obj -MF "test/capi/capiUnitTests.exe.p/capiLinearGradient.cpp.obj.d" -o test/capi/capiUnitTests.exe.p/capiLinearGradient.cpp.obj "-c
" ../test/capi/capiLinearGradient.cpp
In file included from ../test/capi/capiLinearGradient.cpp:25:
../test/capi/../catch.hpp: In instantiation of 'bool Catch::compareEqual(const LhsT&, const RhsT&) [with LhsT = const Tvg_Color_Stop*; RhsT = long long int]':
../test/capi/../catch.hpp:2345:34:   required from 'const Catch::BinaryExpr<LhsT, const RhsT&> Catch::ExprLhs<LhsT>::operator==(const RhsT&) [with RhsT = long l
ong int; LhsT = const Tvg_Color_Stop* const&]'
../test/capi/capiLinearGradient.cpp:165:5:   required from here
../test/capi/../catch.hpp:2316:98: error: ISO C++ forbids comparison between pointer and integer [-fpermissive]
 2316 |     auto compareEqual( LhsT const& lhs, RhsT const& rhs ) -> bool { return static_cast<bool>(lhs == rhs); }
      |                                                                                              ~~~~^~~~~~
[...]

PR #1386, opened 2023-04-20
tests: fixing error on windows (mingw-64)

mgrudzinska:
Error while comparing a pointer and an integer.
Solved by using a null pointer const instead of
the NULL macro.

@issue: https://github.com/thorvg/thorvg/issues/1382

### DIFF
--- a/test/capi/capiLinearGradient.cpp
+++ b/test/capi/capiLinearGradient.cpp
@@ -162,7 +162,7 @@ TEST_CASE("Linear Gradient clear data", "[capiLinearGradient]")
 
     REQUIRE(tvg_gradient_set_color_stops(gradient, NULL, 0) == TVG_RESULT_SUCCESS);
     REQUIRE(tvg_gradient_get_color_stops(gradient, &color_stops_ret, &color_stops_count_ret) == TVG_RESULT_SUCCESS);
-    REQUIRE(color_stops_ret == NULL);
+    REQUIRE(color_stops_ret == nullptr);
     REQUIRE(color_stops_count_ret == 0);
 
     REQUIRE(tvg_gradient_del(gradient) == TVG_RESULT_SUCCESS);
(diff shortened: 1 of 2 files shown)


mgrudzinska:
using nullptr solves this as well, but I tried to be more 'c-style'

test/capi/capiLinearGradient.cpp
```diff
@@ -162,7 +162,7 @@ TEST_CASE("Linear Gradient clear data", "[capiLinearGradient]")
 
     REQUIRE(tvg_gradient_set_color_stops(gradient, NULL, 0) == TVG_RESULT_SUCCESS);
     REQUIRE(tvg_gradient_get_color_stops(gradient, &color_stops_ret, &color_stops_count_ret) == TVG_RESULT_SUCCESS);
-    REQUIRE(color_stops_ret == NULL);
+    REQUIRE(color_stops_ret == (const Tvg_Color_Stop*)NULL);
```
hermet, line 165: I guess we could achieve success by applying compiler flags. Maybe we could try using the -Wnonnull-compare flag?

mgrudzinska: the problem here is comparison between pointer and integer, not a non-null ptr with null, so I don;t see how this flag should help here. beside, I don;t like to suppress warnings, I prefer to solve them...

hermet: 1. It would be better if you addressed the warning msg in this fix so that we can identify the issue easier. You can refer to this commit (https://github.com/thorvg/thorvg/commit/e576e1d8e6a78b97e23bbb443777a97a4ce85cfd) how I share the issue info and where the compiler notify the warning flags us.

2. Comparing NULL(0) is absolutely c-style code. "(const Tvg_Color_Stop*)NULL" this one is more alien-wise. Later some other programmers would come up with NULL compare again, we will face this issue again. 

3. Current test code is based on C++ due to the testframework, So you can just use nullptr or just shutdown warning to prevent later same issues. It's more readable and maintainable.

vtorri: the question is why there is NULL, rather than 0

vtorri: nullptr is probably better. NULL is defined from an int, while nullptr is a pointer (since C++11)

mgrudzinska: @hermet no alien-wise code :D nullptr used

as I wrote in the pr title - it's an error, not a warning.

also, I forgot to add the issue number in the commit msg... sorry. 
I hope the issue number is enough and there is no need to put the error msg in the commit msg. Please let me know if you do not agree, so in the future I'll apply the preferred way of writing commit msg,

hermet (approved):

Issue #1234, opened 2022-08-17
about TVG_EXPORT and Windows


vtorri:
Hello

about TVG_EXPORT:
 * it should be available on Windows for both Microsoft compiler and mingw, so line 21 of thorvg.h : ` #if defined(_WIN32) && !defined(__clang__)`
 * I've not tested yet but I think that clang support __declspec
 * enum class are a type and should not be exported with TVG_EXPORT
 * when TVG_BUILD is not set, and no static lib is created, TVG_EXPORT should be set to __declspec(dllimport)

hermet:
@vtorri thanks. will check it.

p.s) it would be more helpful if you send patches directly.

vtorri:
I can send a PR for enum classes if you agree. I don't know the status of this on linux, but I think it's the same than on Windows: a type, even an enum, is not a symbol.

vtorri:
actually; the difficult part is managing the static build on Windows.

Btw, fo me,static build on Windows is not a good idea.

hermet:
> num classes

Sure, you can go onto this if you think it's necessary.

hermet:
> I can send a PR for enum classes if you agree. I don't know the status of this on linux, but I think it's the same than on Windows: a type, even an enum, is not a symbol.

@vtorri Sure, you can go on this if you think it's necessary.

PR #1387, opened 2023-04-21
infra: specify TVG_API for both static/dynamic linking.

hermet:
Issue: https://github.com/thorvg/thorvg/issues/1234

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -19,17 +19,32 @@
 #include <memory>
 #include <string>
 
-#ifdef TVG_BUILD
-    #if defined(_WIN32) && !defined(__clang__)
-        #define TVG_EXPORT __declspec(dllexport)
-        #define TVG_DEPRECATED __declspec(deprecated)
+#ifdef TVG_API
+    #undef TVG_API
+#endif
+
+#if defined(_WIN32) && !defined(__clang__)
+    #if TVG_BUILD
+        #if TVG_EXPORT
+            #define TVG_API __declspec(dllexport)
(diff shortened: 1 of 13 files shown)


capnm:
@hermet 
> issue: #1234 when TVG_BUILD is not set, and no static lib is created, TVG_EXPORT should be set to __declspec(dllimport)

This change caused now Windows issues by Godot's _external_ static linking, i.e. the __declspec(dllimport). Here the code must be now changed directly :thinking: 
-#define TVG_API __declspec(dllimport)
+#define TVG_API

https://github.com/godotengine/godot/pull/76960#issuecomment-1544491279 ...
https://github.com/godotengine/godot/actions/runs/4951907523

vtorri:
the rule is:
* if thorvg lib is built as a shared lib:
  * if the lib is built: TVG_API must be set to __declspec(dllexport)
  * if the lib is used: TVG_API must be set to __declspec(impexport)
 * otherwise (thorvg built as a static lib), TVG_API must be set to nothing

so at some point, you should have a TVG_STATIC set when static lib is requested (and not set for shared lib)

vtorri:
this patch is wrong for binaries. TVG_BUILD must NOT be set for binaries, ONLY for the lib

hermet:
@capnm @vtorri thanks for noti.
Issue created: https://github.com/thorvg/thorvg/issues/1446
will visit it before the next minor update.

vtorri:
@capnm try to patch git repo with #1448 and report in that patch if it fixes your problem

Issue #1370, opened 2023-04-16
Random crashes due data races


qarmin:
When converting some svg with svg2png tool, I sometimes have crash.
When I check this files manually only once, almost always this works fine, but when I convert them in loop 100-1000 times, then I have very frequently multiple crashes.

Example file:
https://user-images.githubusercontent.com/41945903/232339294-4afce7ae-2af2-459a-b2f5-c30172a19c42.svg

Try to run conversion in loop (and compile app with asan sanitizer)
```
 for i in {0..1000}; do /home/rafal/test/thorvg/build/src/bin/svg2png/svg2png /home/rafal/Desktop/SVGBigPack/KOSZ_20043.svg -r 400x400; done;
```
this may happens very rarely, so I suggest to use https://github.com/qarmin/SVG-regression-finder in which problem is more visible(tool automatically will use all avaialble threads)

Ubsan/Asan Backtraces:
```
==87591==ERROR: AddressSanitizer: heap-use-after-free on address 0x603000001bde at pc 0x7fb0ad6b96b4 bp 0x7fff8c432850 sp 0x7fff8c432840
READ of size 1 at 0x603000001bde thread T0
    #0 0x7fb0ad6b96b3 in _decomposeOutline ../src/lib/sw_engine/tvgSwRle.cpp:733
    #1 0x7fb0ad6b96b3 in _genRle ../src/lib/sw_engine/tvgSwRle.cpp:767
    #2 0x7fb0ad6bbd38 in rleRender(SwRleData*, SwOutline const*, SwBBox const&, bool) ../src/lib/sw_engine/tvgSwRle.cpp:963
    #3 0x7fb0ad6c138a in shapeGenRle(SwShape*, tvg::RenderShape const*, bool) ../src/lib/sw_engine/tvgSwShape.cpp:517
    #4 0x7fb0ad6b1158 in SwShapeTask::run(unsigned int) ../src/lib/sw_engine/tvgSwRenderer.cpp:122
    #5 0x7fb0ad6adfa6 in tvg::Task::done(unsigned int) ../src/lib/tvgTaskScheduler.h:63
    #6 0x7fb0ad6a8bc0 in tvg::SwRenderer::renderShape(void*) ../src/lib/sw_engine/tvgSwRenderer.cpp:386
    #7 0x7fb0ad63cdb0 in tvg::Paint::Impl::render(tvg::RenderMethod&) ../src/lib/tvgPaint.cpp:177
[...]

PR #1388, opened 2023-04-21
Poc Regression finder

qarmin:
This is only proof of concept of possible solution to find regressions

Logic of checking for regressions: 
- Compile thorvg from this PR
- Download and compile thorvg from develop branch
- Download and compile regression finder - https://github.com/qarmin/SVG-regression-finder - allow to find differences between png generated by two apps(or same app with two versions)
- Downloads pack with ~1000 SVG files
- Run tests
- Store files that looks different between versions
- Store files that failed to test(crashes/fails to open them etc.)

Regression finder(some of parameters are available to set via config file): 
- Collect svg files to test(removes from list files with filters, text etc.)
- Creates png files from SVG with same resolution
- Removes alpha channel(used algorithms have problems with finding differences between two images with alpha)
- Compare two pictures, print results and copy them to folders if are broken

Other:
- Comments about number of found differences cannot be added to PR - I got this error message(`regression_check.sh` was based on other sh files, so this is quite strange)
```
  "message": "Resource not accessible by integration",
  "documentation_url": "https://docs.github.com/rest/reference/issues#create-an-issue-comment"
```
- 1000 random files was chosen by me - https://github.com/qarmin/SVG-regression-finder/releases/download/0.2.0/SVG1000.zip(some are properly converted by ThorVG, some are not) - I tried to choose big files, to test as much as possible. 
- Each PR will show random number of broken/problematic files until #1370 is fixed

This CI job should never fail - app checks for differences, but cannot verify if changes are good or bad.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -1,0 +1,67 @@
+name: Regression Testing
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install things
+        run: |
+          sudo apt install -y wget2 librsvg2-dev librsvg2-bin ninja-build git gcc-multilib g++-multilib
+          sudo pip3 install meson
+
+      - name: Compile Thorvg PR
+        run: |
+          meson . build -Dtools=svg2png -Db_sanitize=address,undefined
+          sudo ninja -C build install
+
+      - name: Compile Thorvg Develop
+        run: |
+          git clone https://github.com/thorvg/thorvg.git thorvg_develop
+          cd thorvg_develop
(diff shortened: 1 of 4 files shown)


hermet:
Thank you for your help. It looks useful. 

Before confirm it, I want to address my two thoughts here.

1. To identify actual regression bugs caused by a patch, we should add only successful SVG samples based on the current TVG version. This will make it clearer and easier to figure out problems. We could prepare a set of SVGs on the ThorVG side for this purpose.

2. The Git action should fail immediately when new issues such as program corruption, memory violation, or SVG drawing failures are detected. This will allow developers to deal with the issues promptly.

hermet (approved):

hermet:
@mgrudzinska  @JSUYA  @qarmin  Apply it now, anyone could enhance the script whenever it's necessary.

Issue #1389, opened 2023-04-22
Svg example crash for thread num > 0


mgrudzinska:
```
    #0 0x101647e11 in wrap_setlocale+0x181 (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x30e11)
    #1 0x100b9fdd6 in svgPathToTvgPath(char const*, tvg::Array<tvg::PathCommand>&, tvg::Array<tvg::Point>&) tvgSvgPath.cpp:560
    #2 0x100ba73d0 in _appendShape(SvgNode*, tvg::Shape*, Box const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) tvgSvgSceneBuilder.cpp:398
    #3 0x100ba29a5 in _sceneBuildHelper(SvgNode const*, Box const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, bool, int, bool*) tvgSvgSceneBuilder.cpp:741
    #4 0x100ba2708 in _sceneBuildHelper(SvgNode const*, Box const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, bool, int, bool*) tvgSvgSceneBuilder.cpp:733
    #5 0x100ba0a0b in svgSceneBuild(SvgNode*, float&, float&, float&, float&, float&, float&, AspectRatioAlign, AspectRatioMeetOrSlice, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, SvgViewFlag) tvgSvgSceneBuilder.cpp:796
    #6 0x100b881d4 in SvgLoader::run(unsigned int) tvgSvgLoader.cpp:3491
    #7 0x100b4a72d in tvg::Task::operator()(unsigned int) tvgTaskScheduler.h:81
    #8 0x100b49fa8 in tvg::TaskSchedulerImpl::run(unsigned int) tvgTaskScheduler.cpp:138
    #9 0x100b49c21 in void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, tvg::TaskSchedulerImpl::TaskSchedulerImpl(unsigned int)::'lambda'()> >(void*) thread:298
    #10 0x7ff8171c7258 in _pthread_start+0x7c (libsystem_pthread.dylib:x86_64+0x6258)
[...]

PR #1390, opened 2023-04-22
svg_loader: removing setting the locale

mgrudzinska:
The locale was set while reading the path attr,
which could have caused race conditions between
threads and potentially led to crashes.
Now removed as unnecessary.

@Issue: https://github.com/thorvg/thorvg/issues/1389

### DIFF
--- a/src/loaders/svg/tvgSvgPath.cpp
+++ b/src/loaders/svg/tvgSvgPath.cpp
@@ -52,7 +52,6 @@
 
 #include <cstring>
 #include <math.h>
-#include <clocale>
 #include <ctype.h>
 #include "tvgSvgLoaderCommon.h"
 #include "tvgSvgPath.h"
@@ -545,20 +544,12 @@ bool svgPathToTvgPath(const char* svgPath, Array<PathCommand>& cmds, Array<Point
     char cmd = 0;
     bool isQuadratic = false;
     char* path = (char*)svgPath;
-    char* curLocale;
-
-    curLocale = setlocale(LC_NUMERIC, NULL);
-    if (curLocale) curLocale = strdup(curLocale);
-    setlocale(LC_NUMERIC, "POSIX");
 
     while ((path[0] != '\0')) {
         path = _nextCommand(path, &cmd, numberArray, &numberCount);
         if (!path) break;
         if (!_processCommand(&cmds, &pts, cmd, numberArray, numberCount, &cur, &curCtl, &startPoint, &isQuadratic)) break;
     }
-
-    setlocale(LC_NUMERIC, curLocale);
-    if (curLocale) free(curLocale);
 
     return true;
 }


mgrudzinska:
from [here](https://en.cppreference.com/w/cpp/locale/setlocale):
```Because setlocale modifies global state which affects execution of locale-dependent functions, it is undefined behavior to call it from one thread, while another thread is executing any of the following functions:...long list```

I don't know why setlocale was called only while path was read. doesn;t it make sense to use it globally? 
although I know, that user can change it between init and term, so this pr isn'l a perfect solution. maybe adding a comment in the init and term descriptions is a good idea?

hermet:
> from [here](https://en.cppreference.com/w/cpp/locale/setlocale): `Because setlocale modifies global state which affects execution of locale-dependent functions, it is undefined behavior to call it from one thread, while another thread is executing any of the following functions:...long list`
> 
> I don't know why setlocale was called only while path was read. doesn;t it make sense to use it globally? although I know, that user can change it between init and term, so this pr isn'l a perfect solution. maybe adding a comment in the init and term descriptions is a good idea?



> from [here](https://en.cppreference.com/w/cpp/locale/setlocale): `Because setlocale modifies global state which affects execution of locale-dependent functions, it is undefined behavior to call it from one thread, while another thread is executing any of the following functions:...long list`
> 
> I don't know why setlocale was called only while path was read. doesn;t it make sense to use it globally? although I know, that user can change it between init and term, so this pr isn'l a perfect solution. maybe adding a comment in the init and term descriptions is a good idea?

@mgrudzinska Obviously, app needs to change locale in a certain scenario so this is not a proper solution. Back to the origin, we need to revisit the svg parser it really needs setscale(). Currently, I can't remeber...

hermet (changes requested):
This souldn't be applied.

hermet:
We need to check out the history of setlocale here...

mgrudzinska:
> We need to check out the history of setlocale here...

setlocale was used in efl, but we are parsing numbers differently - we don;t use locale sensitive functions like strtod, strotof, atoi etc (we have svgUtilStrtof)

but I just find out the isalpha, that we are using is locale sensitive. so I need to investigate.

mgrudzinska:
@hermet @JSUYA

in efl setlocale was used while parsing the path. In the SVG specs, the syntax for the path is described using EBNF, it is very well defined and there is no word about any locale specific characters that may appear there, so parsing should be independent on the locale settings.
In thorvg while path is parsed, isspace, isalpha and internally implemented svgUtilStrtof are used. I commented below in more details, but in general - the parsing is independent on locale setting.

svg tags like```text``` and ```id``` should be locale sensitive as they are right now in thorvg. 


In thirvg the following locale sensitive functions:
- ```isalpha()``` - used for parsing the path - only few ascii letters possible ([ref](https://www.w3.org/TR/SVG2/paths.html#TheDProperty)). whatever locale is set these symbols will be correctly recognized
- ```issapce()``` - SVG files are based on XML, which supports the Unicode character set. from [xml specs](https://www.w3.org/TR/2008/REC-xml-20081126/#NT-S) the possible white spaces: (#x20 | #x9 | #xD | #xA) (space/tab/carriage return/newline) and all of them will be universally recognized and not affected by locale settings.
- ```isdigit()``` and ```isxdigit()``` - from [here](https://en.cppreference.com/w/cpp/string/byte/isdigit): 
  ```isdigit and isxdigit are the only standard narrow character classification functions that are not affected by the currently installed C locale. although some implementations (e.g. Microsoft in 1252 codepage) may classify additional single-byte characters as digits.```
both used to parse path data, numbers, and coordinates that use the standard ASCII characters for decimal and hexadecimal digits, so the additional characters recognized as digit shouldn't affect anything.
- ```tolower()``` - used in ```svgUtilStrtof ``` to parse of special floating-point values (inf, infinity, and nan) - locale doesn;t matter
- ```atoi()``` - used to pars a thorvg version in tvg loader, tvgInitializer and svg2png - I can imagine that in the last one someone can call ```-r 4,000x4,000``` but it's so unlikely and counterintuitive...

JSUYA:
In https://github.com/thorvg/thorvg/issues/1389 I believe the reason for the crash is because locale is not thread safe(maybe?).
According to this article https://stackoverflow.com/questions/4057319/is-setlocale-thread-safe-function using `uselocale()` may be another workaround.

As you analyzed, we are using custom strtof instead of strtod, And I don't think removing is a problem if your analysis is correct.
However, I think we should take a more careful if we run into issues with this in the future, as it's very hard to track down.

mgrudzinska:
> https://stackoverflow.com/questions/4057319/is-setlocale-thread-safe-function

yes, if we need to set the locale, we have to provide solution for unix and for windows (uselocale is for unix only). 
sure it can be done, but the question is whether it is needed. 
also if we think that we need it, shouldn;t it be applied for the whole loading process? why path is so special?
I still think that it is not necessary to set the locale

hermet (approved):
@mgrudzinska If you think it's unnecessary, please push it.

JSUYA (approved):

PR #1392, opened 2023-04-22
capi: type identifier apis added

mgrudzinska:
The tvg_gradient_get_identifier and tvg_paint_get_identifier apis have been introduced. They set the newly introduced Tvg_Type_Identifier enum, which indicates the paint and gradient types, respectively.

### DIFF
--- a/src/bindings/capi/thorvg_capi.h
+++ b/src/bindings/capi/thorvg_capi.h
@@ -128,6 +128,23 @@ typedef enum {
 
 
 /**
+ * \brief Enumeration indicating the ThorVG class type.
+ *
+ * \ingroup ThorVGCapi_Paint
+ *
+ * \since 0.9
+ */
+typedef enum {
+    TVG_IDENTIFIER_UNDEF = 0,   ///< Undefined type.
+    TVG_IDENTIFIER_SHAPE,       ///< A shape type paint.
+    TVG_IDENTIFIER_SCENE,       ///< A scene type paint.
+    TVG_IDENTIFIER_PICTURE,     ///< A picture type paint.
+    TVG_IDENTIFIER_LINEAR_GRAD, ///< A linear gradient type.
+    TVG_IDENTIFIER_RADIAL_GRAD  ///< A radial gradient type.
+} Tvg_Identifier;
+
+
+/**
  * \addtogroup ThorVGCapi_Shape
  * \{
  */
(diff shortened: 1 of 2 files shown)


src/bindings/capi/thorvg_capi.h
```diff
@@ -127,6 +127,21 @@ typedef enum {
 } Tvg_Composite_Method;
 
 
+/**
+ * \brief Enumeration indicating the ThorVG class type.
+ *
+ * \ingroup ThorVGCapi_Paint
```
hermet, line 133: since 0.9

src/bindings/capi/thorvg_capi.h
```diff
@@ -899,6 +914,18 @@ TVG_API Tvg_Result tvg_paint_set_composite_method(Tvg_Paint* paint, Tvg_Paint* t
...
+/**
+* \brief Gets the unique id value of the paint instance indicating the instance type.
+*
+* \param[in] paint The Tvg_Paint object of which to get the identifier value.
+* \param[out] identifier The unique identifier of the paint instance type.
+*
+* \return Tvg_Result enumeration.
+* \retval TVG_RESULT_SUCCESS Succeed.
+* \retval TVG_RESULT_INVALID_ARGUMENT In case a @c nullptr is passed as the argument.
+*/
```
hermet, line 927: since 0.9

src/bindings/capi/thorvg_capi.h
```diff
@@ -1737,6 +1764,18 @@ TVG_API Tvg_Result tvg_gradient_set_transform(Tvg_Gradient* grad, const Tvg_Matr
...
+/**
+* \brief Gets the unique id value of the gradient instance indicating the instance type.
+*
+* \param[in] grad The Tvg_Gradient object of which to get the identifier value.
+* \param[out] identifier The unique identifier of the gradient instance type.
+*
+* \return Tvg_Result enumeration.
+* \retval TVG_RESULT_SUCCESS Succeed.
+* \retval TVG_RESULT_INVALID_ARGUMENT In case a @c nullptr is passed as the argument.
+*/
```
hermet, line 1776: since 0.9

hermet (commented):
Could you please add since tag?

hermet (approved):

src/bindings/capi/thorvg_capi.h
```diff
@@ -127,6 +127,23 @@ typedef enum {
...
+ * \since 0.9
+ */
+typedef enum {
+    TVG_TYPE_IDENTIFIER_UNDEF = 0,   ///< Undefined type.
+    TVG_TYPE_IDENTIFIER_SHAPE,       ///< A shape type paint.
+    TVG_TYPE_IDENTIFIER_SCENE,       ///< A scene type paint.
+    TVG_TYPE_IDENTIFIER_PICTURE,     ///< A picture type paint.
+    TVG_TYPE_IDENTIFIER_LINEAR_GRAD, ///< A linear gradient type.
+    TVG_TYPE_IDENTIFIER_RADIAL_GRAD  ///< A radial gradient type.
+} Tvg_Type_Identifier;
```
hermet, line 144: What do you think about removing the word "TYPE" from the enumeration name (resulting in "Tvg_Identifier")?
That would improve consistency.

hermet (commented):
Please check one more comment.

src/bindings/capi/thorvg_capi.h
mgrudzinska: I prefer to keep it, otherwise it's an indicator of the whole thorvg? doesn;t make sense. now it's clear that it's an indicator of a type, but as you wish I can change it.

hermet: Could you please tell us what practical problems you are concerned about? If you feel like it doesn't make sense. We can discuss it further. Personally, I don't see any logical problems either way. But according to your argument, all should have "TYPE" as well. 

`
Tvg_Result
Tvg_Composite_Method
Tvg_Path_Command
Tvg_Stroke_Cap
Tvg_Stroke_Join
Tvg_Stroke_Fill
Tvg_Fill_Rule
`

In my experience, naming is often a matter of personal preference, and many people have different preferences. Therefore, consistency should be the top priority criterion.

hermet: And yes, the identifier covers a whole set of classes in ThorVG's like Tvg_Result.

mgrudzinska: changed :)

PR #1393, opened 2023-04-23
tests: ++coverage of the svg_loader

mgrudzinska:
before:
Zrzut ekranu 2023-04-23 o 22 46 51

after:
Zrzut ekranu 2023-04-24 o 00 29 37


hermet (commented):
tag.svg is necessary here?

hermet:
> tag.svg is necessary here?

ah it's draft. nevermind.

hermet (approved):

PR #1394, opened 2023-04-23
svg_loader: fixing 'infinity' handling in svgUtilStrtof

mgrudzinska:
The word 'infinity' was not read to the end due to the pointer shifting.


hermet (approved):
good catch, thanks.

PR #1395, opened 2023-04-23
svg_loader: removed unused code

mgrudzinska:
While css styling applied one of conditions was
always false. It is now removed and the function
that was called is refactored

### DIFF
--- a/src/loaders/svg/tvgSvgCssStyle.cpp
+++ b/src/loaders/svg/tvgSvgCssStyle.cpp
@@ -155,12 +155,12 @@ SvgNode* cssFindStyleNode(const SvgNode* style, const char* title, SvgNodeType t
 
 SvgNode* cssFindStyleNode(const SvgNode* style, const char* title)
 {
-    if (!style) return nullptr;
+    if (!style || !title) return nullptr;
 
     auto child = style->child.data;
     for (uint32_t i = 0; i < style->child.count; ++i, ++child) {
         if ((*child)->type == SvgNodeType::CssStyle) {
-            if ((title && (*child)->id && !strcmp((*child)->id, title))) return (*child);
+            if ((*child)->id && !strcmp((*child)->id, title)) return (*child);
         }
     }
     return nullptr;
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgCssStyle.cpp
```diff
@@ -175,9 +175,6 @@ void cssUpdateStyle(SvgNode* doc, SvgNode* style)
             if (auto cssNode = cssFindStyleNode(style, nullptr, (*child)->type)) {
                 cssCopyStyleAttr(*child, cssNode);
             }
-            if (auto cssNode = cssFindStyleNode(style, nullptr)) {
```
hermet, line 178: how about the call at 196 line?

src/loaders/svg/tvgSvgCssStyle.cpp
```diff
@@ -175,9 +175,6 @@ void cssUpdateStyle(SvgNode* doc, SvgNode* style)
             if (auto cssNode = cssFindStyleNode(style, nullptr, (*child)->type)) {
                 cssCopyStyleAttr(*child, cssNode);
```
JSUYA, line 176: It looks like this line is also not used.
If the title is null, nothing is done in `cssFindStyleNode()`, so the entire line 175 to 180 looks unused.
I put the test code(printf()) on lines 176 and 179 along with samples from https://github.com/thorvg/thorvg/issues/1255 and tested it.
And if cssNode is not found here, it is expected to be found in cssApplyStyleToPostponeds().

mgrudzinska: @JSUYA from ```cssFindStyleNode```:
```if ((!title && !(*child)->id) || (title && (*child)->id && !strcmp((*child)->id, title))) return (*child);```

and sample:
```
 <defs>
  <style>
    rect {
        fill: #00cc00;
        stroke: #99ff33;
    }
  </style>
 </defs>
 <rect x="5" y="5" width="50" height="50" />
</svg>
```
we have in images ```css-style.svg``` sample file.

File: src/loaders/svg/tvgSvgCssStyle.cpp (hunk 1 above)
Comment (mgrudzinska) [reply src/loaders/svg/tvgSvgCssStyle.cpp:178]: it is needed.
sample that uses l.196:
```
 <circle cx="5" cy="5" r="50" class="mineStyle" />
 <defs>
  <style>
    .mineStyle {
        fill: red;
        stroke: yellow;
        stroke-width: 10;
    }
  </style>
 </defs>
</svg>
```

hermet (approved):

JSUYA (approved):

Issue #1409, opened 2023-04-28
Thorvg breaks down svg files randomly


qarmin:
When testing latest thorvg version(9fb66973f0f56c1a5da1d9640e16fa253a5ad664), I found that sometimes in CI with exactly same list of svg files, I got different number of broken images each time when I run CI job.

I found one file, that looks different almost each time when I run svg2png.

For me look like logic bug/data races.

Without reproducible results, regression finder logic cannot be integrated fully with CI, because there would be a lot of false positives

[AA_5.svg.zip](https://github.com/thorvg/thorvg/files/11356766/AA_5.svg.zip)

https://user-images.githubusercontent.com/41945903/235237094-6f947c09-6158-4455-8047-e7e8a3d8896c.mp4

Even thread sanitizer shows errors
```
WARNING: ThreadSanitizer: data race (pid=577198)
  Write of size 4 at 0x7b4c0000001c by thread T1 (mutexes: write M0):
    #0 _applySvgViewFlag ../src/loaders/svg/tvgSvgSceneBuilder.cpp:784 (libthorvg.so.0+0x4abde) (BuildId: 4908c25bb435795c049df95548e2e26737843685)
    #1 svgSceneBuild(SvgNode*, float&, float&, float&, float&, float&, float&, AspectRatioAlign, AspectRatioMeetOrSlice, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, SvgViewFlag) ../src/loaders/svg/tvgSvgSceneBuilder.cpp:799 (libthorvg.so.0+0x4abde)
    #2 SvgLoader::run(unsigned int) ../src/loaders/svg/tvgSvgLoader.cpp:3495 (libthorvg.so.0+0x45143) (BuildId: 4908c25bb435795c049df95548e2e26737843685)
    #3 tvg::TaskSchedulerImpl::run(unsigned int) <null> (libthorvg.so.0+0x1bbc5) (BuildId: 4908c25bb435795c049df95548e2e26737843685)
    #4 std::thread::_State_impl<std::thread::_Invoker<std::tuple<tvg::TaskSchedulerImpl::TaskSchedulerImpl(unsigned int)::{lambda()#1}> > >::_M_run() <null> (libthorvg.so.0+0x1bc4c) (BuildId: 4908c25bb435795c049df95548e2e26737843685)
[...]

PR #1411, opened 2023-04-29
svg_loader: fix memory access violation

hermet:
@Issue: https://github.com/thorvg/thorvg/issues/1409


qarmin:
Most of the time, thread sanitizer shows same big number warnings(usually 200) as without this patch

qarmin:
Thread sanitizer still shows a lot of warnings - I don't see here any improvement.

Converting is now a lot of better, but still not 100% reproducible - sometimes png is wrong, but sometimes looks right, but it have different size than most of the time

https://user-images.githubusercontent.com/41945903/235309288-10708bab-6bf3-4d00-b7b4-821de620b4f0.mp4

Python script to verify changes
```
import os
import subprocess

svg_path = "/home/rafal/test/thorvg/build/src/bin/svg2png/svg2png"
image_input = "AA_5.svg"
image_output = "AA_5.png"

args = [svg_path, image_input, "-r", "100x100"]
sizes = {}
for i in range(101):
    subprocess.call(args, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT)
    new_size = os.path.getsize(image_output)
    if new_size in sizes:
        sizes[new_size] += 1
    else:
        sizes[new_size] = 1
    
print(sizes)
```
output
```
{11030: 26, 11002: 56, 10697: 13, 10868: 4, 10668: 2}
```

capnm:
_-r 100x100_ should here IMO return an error or a blank image if thorvg cannot calculate the canvas size due to unsupported svg features.

hermet:
> _-r 100x100_ should here IMO return an error or an empty image if thorvg cannot calculate the canvas size due to unsupported svg features. Such images should then be removed from the regression test and otherwise images should be added for new supported features like SVGs without viewbox/viewport.

Yeah, we arehandling this as another issue: https://github.com/thorvg/thorvg/pull/1412

hermet:
@mgrudzinska I would appreciate if you could review this:
https://github.com/thorvg/thorvg/pull/1411/commits/842dd383e2c755541ab7adb2a8942a1dbddf64f2

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3513,35 +3514,33 @@ bool SvgLoader::header()
...
-        if (viewFlag & SvgViewFlag::Height) {
-            h = loaderData.doc->node.doc.h;
-        }
-        //Override size
-        if (viewFlag & SvgViewFlag::Viewbox) {
-            vx = loaderData.doc->node.doc.vx;
-            vy = loaderData.doc->node.doc.vy;
-            vw = loaderData.doc->node.doc.vw;
-            vh = loaderData.doc->node.doc.vh;
+        vx = loaderData.doc->node.doc.vx;
```
JSUYA, line 3518: loaderData.doc->node.doc.vx, vy is probably not initialized.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -816,5 +809,12 @@ unique_ptr<Scene> svgSceneBuild(SvgNode* node, float& vx, float& vy, float& vw,
     auto root = Scene::gen();
     root->push(move(compositeLayer));
 
+    loaderData.doc->node.doc.vx = vBox.x;
+    loaderData.doc->node.doc.vy = vBox.y;
+    loaderData.doc->node.doc.vw = vBox.w;
+    loaderData.doc->node.doc.vh = vBox.h;
+    loaderData.doc->node.doc.w = w;
+    loaderData.doc->node.doc.h = h;
```
JSUYA, line 817: I have question. Why do we need to update the value of loaderData here?

hermet: If there were no any given viewbox & size information, It finds the default size from the scene geometry. To get the final size from it, it updates the info at the last.

src/loaders/svg/tvgSvgLoader.cpp
hermet: For simplication, let it expect the proper values are returned. Otherwise it will override them after the sequence. (3529 ~ 3535)

src/loaders/svg/tvgSvgSceneBuilder.cpp
JSUYA: Are there any cases where loaderData needs to be updated again?
The loader do the parsing again, so there is no need to reuse this.
Even if it is modified to be reused(in the future?), it seems that redundant calculations will occur because w, and h are calculated once again in the above logic.

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: In your code (532~3535), only vw and vh are overried. If `<viewbox>` is not defined, no values are set for loaderData.doc->node.doc.vx and loaderData.doc->node.doc.vy.
(The old code is logically set only when vx and vy are initialized in the class and the viewbox is declared.)
It's common for structs to be set to 0, but I know not all compilers guarantee this.
Therefore, in _createSvgNode(), I think we need to explicitly initialize the values of `doc`


`vx, vy -999.000000 -999.000000 loaderData.doc->node.doc vx vy -999.000000 -999.000000`

```diff
diff --git a/src/loaders/svg/tvgSvgLoader.cpp b/src/loaders/svg/tvgSvgLoader.cpp
index 03a8fee1..36f33ee9 100644
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1347,6 +1347,10 @@ static SvgNode* _createSvgNode(SvgLoaderData* loader, SvgNode* parent, const cha
     doc->align = AspectRatioAlign::XMidYMid;
     doc->meetOrSlice = AspectRatioMeetOrSlice::Meet;
     doc->viewFlag = SvgViewFlag::None;
+
+    doc->vx = -999; // Cases where the compiler does not guarantee initialization
+    doc->vy = -999;
+
     func(buf, bufLength, _attrParseSvgNode, loader);

     if (!(doc->viewFlag & SvgViewFlag::Viewbox)) {
@@ -3535,6 +3539,7 @@ bool SvgLoader::header()
             else vh = h = 1.0f;
             run(0);
         }
+        printf("vx, vy %f %f loaderData.doc->node.doc vx vy %f %f\n", vx, vy, loaderData.doc->node.doc.vx,loaderData.doc->node.doc.vy);

         return true;
     }

```

```diff
--- a/src/examples/images/favorite_on.svg
+++ b/src/examples/images/favorite_on.svg
@@ -2,7 +2,7 @@
 <!-- Generator: Adobe Illustrator 16.0.3, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
 <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
 <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-        width="80px" height="80px" viewBox="-14 -14 80 80" enable-background="new -14 -14 80 80" xml:space="preserve">
+        width="80px" height="80px"  enable-background="new -14 -14 80 80" xml:space="preserve">
```

src/loaders/svg/tvgSvgSceneBuilder.cpp
hermet: @JSUYA this is a fix for the regression of https://github.com/thorvg/thorvg/pull/1248

hermet: @JSUYA and this one. It needs the get the completed size by this https://github.com/thorvg/thorvg/pull/1338

hermet: I missed the logic overriding, I will update it again soon. 👍

hermet: @JSUYA updated thanks!

src/loaders/svg/tvgSvgLoader.cpp
hermet: overriding code has been missed by my fault. now it does.

PR #1412, opened 2023-04-30
svg_loader: handling svg width/height in percentages

mgrudzinska:
The percentages should refer to the size of the viewbox. This was not the case for not knowing the viewbox before reading the width/height.

@Issue: https://github.com/thorvg/thorvg/issues/1409

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -861,10 +861,18 @@ static bool _attrParseSvgNode(void* data, const char* key, const char* value)
 
     if (!strcmp(key, "width")) {
         doc->w = _toFloat(loader->svgParse, value, SvgParserLengthType::Horizontal);
-        doc->viewFlag = (doc->viewFlag | SvgViewFlag::Width);
+        if (strstr(value, "%")) {
+            doc->sw = svgUtilStrtof(value, nullptr) / 100.0f;
+        } else {
+            doc->viewFlag = (doc->viewFlag | SvgViewFlag::Width);
+        }
     } else if (!strcmp(key, "height")) {
         doc->h = _toFloat(loader->svgParse, value, SvgParserLengthType::Vertical);
-        doc->viewFlag = (doc->viewFlag | SvgViewFlag::Height);
+        if (strstr(value, "%")) {
+            doc->sh = svgUtilStrtof(value, nullptr) / 100.0f;
+        } else {
+            doc->viewFlag = (doc->viewFlag | SvgViewFlag::Height);
+        }
     } else if (!strcmp(key, "viewBox")) {
         if (_parseNumber(&value, &doc->vx)) {
             if (_parseNumber(&value, &doc->vy)) {
@@ -1347,6 +1355,8 @@ static SvgNode* _createSvgNode(SvgLoaderData* loader, SvgNode* parent, const cha
     doc->align = AspectRatioAlign::XMidYMid;
     doc->meetOrSlice = AspectRatioMeetOrSlice::Meet;
     doc->viewFlag = SvgViewFlag::None;
+    doc->sw = 1.0f;
+    doc->sh = 1.0f;
     func(buf, bufLength, _attrParseSvgNode, loader);
 
     if (!(doc->viewFlag & SvgViewFlag::Viewbox)) {
@@ -3530,8 +3540,16 @@ bool SvgLoader::header()
             vw = loaderData.doc->node.doc.vw;
             vh = loaderData.doc->node.doc.vh;
 
-            if (!(viewFlag & SvgViewFlag::Width)) w = vw;
-            if (!(viewFlag & SvgViewFlag::Height)) h = vh;
+            if (!(viewFlag & SvgViewFlag::Width)) {
+                w = vw * loaderData.doc->node.doc.sw;
+                loaderData.doc->node.doc.sw = 1.0f;
+                viewFlag = (viewFlag | SvgViewFlag::Width);
+            }
+            if (!(viewFlag & SvgViewFlag::Height)) {
+                h = vh * loaderData.doc->node.doc.sh;
+                loaderData.doc->node.doc.sh = 1.0f;
+                viewFlag = (viewFlag | SvgViewFlag::Height);
+            }
         } else {
             vw = w;
             vh = h;
--- a/src/loaders/svg/tvgSvgLoaderCommon.h
+++ b/src/loaders/svg/tvgSvgLoaderCommon.h
@@ -260,6 +260,8 @@ struct SvgDocNode
     float vy;
     float vw;
     float vh;
+    float sw;
+    float sh;
     SvgViewFlag viewFlag;
     SvgNode* defs;
     SvgNode* style;
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -796,7 +796,10 @@ unique_ptr<Scene> svgSceneBuild(SvgNode* node, float& vx, float& vy, float& vw, 
 
     Box vBox = {vx, vy, vw, vh};
     auto docNode = _sceneBuildHelper(node, vBox, svgPath, false, 0);
+
     _applySvgViewFlag(docNode.get(), vx, vy, vw, vh, w, h, viewFlag);
+    w *= node->node.doc.sw;
+    h *= node->node.doc.sh;
 
     if (!mathEqual(w, vw) || !mathEqual(h, vh)) {
         Matrix m = _calculateAspectRatioMatrix(align, meetOrSlice, w, h, {vx, vy, vw, vh});


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -857,14 +857,24 @@ static bool _attrParseSvgNode(void* data, const char* key, const char* value)
 {
     SvgLoaderData* loader = (SvgLoaderData*)data;
     SvgNode* node = loader->svgParse->node;
-    SvgDocNode* doc = &(node->node.doc);
+    SvgDocNode* doc = &(node->node.doc) ;
 
     if (!strcmp(key, "width")) {
         doc->w = _toFloat(loader->svgParse, value, SvgParserLengthType::Horizontal);
-        doc->viewFlag = (doc->viewFlag | SvgViewFlag::Width);
+        auto s = svgUtilStrtof(value, nullptr);
```
hermet, line 864: s could be inside if() {}

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -260,6 +260,8 @@ struct SvgDocNode
     float vy;
     float vw;
     float vh;
+    float wScalingFactor;
```
hermet, line 263: You are using different names (sw, sh / wScalingFactor, hScalingFactor) for the same purpose. Please maintain consistency by using sw and sh.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -797,6 +797,8 @@ unique_ptr<Scene> svgSceneBuild(SvgNode* node, float& vx, float& vy, float& vw,
     Box vBox = {vx, vy, vw, vh};
     auto docNode = _sceneBuildHelper(node, vBox, svgPath, false, 0);
     _applySvgViewFlag(docNode.get(), vx, vy, vw, vh, w, h, viewFlag);
+    w *= sw;
+    h *= sh;
```
hermet, line 801: Header() is designed to return necessary information to the user. However, since header() does not account for percentages, users might acquire incorrect results. Unless we can understand each other, I may have to give you a negative rating.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3515,6 +3527,8 @@ bool SvgLoader::header()
         meetOrSlice = loaderData.doc->node.doc.meetOrSlice;
         w = 1.0f;
         h = 1.0f;
+        wScalingFactor = loaderData.doc->node.doc.wScalingFactor;
```
hermet, line 3530: What is this used for?

hermet (changes requested):
Thanks for fixing, I added nitpicks and one remarkable opinion.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3492,7 +3504,7 @@ void SvgLoader::run(unsigned tid)
         if (loaderData.gradients.count > 0) _updateGradient(&loaderData, loaderData.doc, &loaderData.gradients);
         if (defs) _updateGradient(&loaderData, loaderData.doc, &defs->node.defs.gradients);
     }
-    root = svgSceneBuild(loaderData.doc, vx, vy, vw, vh, w, h, align, meetOrSlice, svgPath, viewFlag);
+    root = svgSceneBuild(loaderData.doc, vx, vy, vw, vh, w, h, wScalingFactor, hScalingFactor, align, meetOrSlice, svgPath, viewFlag);
```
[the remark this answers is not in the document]
mgrudzinska: sure, removed

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: renamed

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: nowhere, removed from the loader

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: done

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: in a case no viewbox was given and:
- now width/height 
or
- widrth and height  were given in %
header doesn;t have info about the actual width and height. I see that you've made some changes here https://github.com/thorvg/thorvg/commit/842dd383e2c755541ab7adb2a8942a1dbddf64f2 so probably there you are solving this (didn;t check very carefully yet), but at this stage I can not change this. 
do you want me to wait with this pr until your changes are applied?

hermet: @mgrudzinska  Okay, I understand your point. Since you've established the sizing policy for SVG, which includes factors such as pt, %, viewport, and viewbox, I would like to request that you include our current implementation policy documentation within the code as comments. The logic is a bit complex, and the rules are scattered and incomplete, so we need to establish concrete rules to prevent them from being broken by someone else in the future. Any comments and opinions are welcome.

As an example, we could include the following comments within the code (in header function or a new unified sizing method) to outline the SVG sizing policy:

For example:

/*  Here is the SVG sizing policy. Derive the SVG size based on the next priority: 
 1. Use ViewSize if it's valid: 
    - if viewSize is in Percentage, then calc the size after applying it to the Viewbox size.
 3. ViewSize is invalid, Use Viewbox.
    - ...
4. Invalid ViewSize & Viewbox? 
  - GetBound from the actual content and use the size as ...
5. ...  */

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3530,8 +3540,16 @@ bool SvgLoader::header()
             vw = loaderData.doc->node.doc.vw;
             vh = loaderData.doc->node.doc.vh;
 
-            if (!(viewFlag & SvgViewFlag::Width)) w = vw;
-            if (!(viewFlag & SvgViewFlag::Height)) h = vh;
+            if (!(viewFlag & SvgViewFlag::Width)) {
```
hermet, line 3543: one more question, 
What do you think if we apply this scaling in _attrParseSvgNode().

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -796,7 +796,10 @@ unique_ptr<Scene> svgSceneBuild(SvgNode* node, float& vx, float& vy, float& vw,
 
     Box vBox = {vx, vy, vw, vh};
     auto docNode = _sceneBuildHelper(node, vBox, svgPath, false, 0);
+
     _applySvgViewFlag(docNode.get(), vx, vy, vw, vh, w, h, viewFlag);
+    w *= node->node.doc.sw;
```
hermet, line 801: So in this case, you concered no given Viewbox, no given pt ViewSize but only Pencentage View SIze.
Does it the SVG standard or just you idea? Otherwise we could ignore the scaling here?

hermet (commented):
Now, the logic looks good to me but please check more minor comments.

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: I think it's good to add flags(or another way) for `sw` and `sh` and do the final calculation in svgSceneBuild() of tvgSvgSceneBuilder.cpp.

hermet: We are fixing the deferred the calculation messed up.

JSUYA: If width or height is defined as `%`, then w,h and sw,sh will be the same value. Therefore, we don't actually need sw or sh separately. So I guess we can define this as a flag.
And the final svg width and height are determined by _applySvgViewFlag(). I think we can calc the final size according to the flag set here.

`w h (0.020000 0.020000), sw sh (0.020000 0.020000)`

```
@@ -3526,6 +3526,7 @@ bool SvgLoader::header()
         w = 1.0f;
         h = 1.0f;

+printf("w h (%f %f), sw sh (%f %f) \n", loaderData.doc->node.doc.w, loaderData.doc->node.doc.h, loaderData.doc->node.doc.sw, loaderData.doc->node.doc.sw);

```
```
--- a/src/examples/images/favorite_on.svg
+++ b/src/examples/images/favorite_on.svg
@@ -2,7 +2,7 @@
 <!-- Generator: Adobe Illustrator 16.0.3, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
 <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
-        width="80px" height="80px" viewBox="-14 -14 80 80" enable-background="new -14 -14 80 80" xml:space="preserve">
+        width="2%" height="2%" viewBox="-14 -14 80 80" enable-background="new -14 -14 80 80" xml:space="preserve">

```

hermet (approved):
Let's go forward.

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: @JSUYA sure, flag would work as well. sorry, I was offline for a few days

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: I'm not 100% percent... we had some time ago discussion about this with no conclusions actually. but we need proper width and height for gradient in case userSpaceOnUse is set. and this way it works

Issue #1414, opened 2023-05-03
Default generated png size is different than in rsvg


qarmin:
Issue from comments in #1409

```
svg2png file.svg
```
produce different results than
```
rsvg-convert file.svg -o file.png
```

Examples of differences
```
SVG_FILE_50623_rsvg.png 16x17, 		SVG_FILE_50623_thorvg.png 16x16 - 	diff 0x1
SVG_FILE_53546_rsvg.png 509x243, 	SVG_FILE_53546_thorvg.png 508x242 - 	diff 1x1
SVG_FILE_137825_rsvg.png 48x48, 	SVG_FILE_137825_thorvg.png 23x23 - 	diff 25x25
SVG_FILE_142665_rsvg.png 10000x10000, 	SVG_FILE_142665_thorvg.png 4320x4320 - 	diff 5680x5680
SVG_FILE_41413_rsvg.png 144x144, 	SVG_FILE_41413_thorvg.png 1x1 - 	diff 143x143
SVG_FILE_126678_rsvg.png 600x600, 	SVG_FILE_126678_thorvg.png 3600x3600 - 	diff 3000x3000
SVG_FILE_142647_rsvg.png 6500x10000, 	SVG_FILE_142647_thorvg.png 2808x4320 - 	diff 3692x5680
SVG_FILE_51860_rsvg.png 512x512, 	SVG_FILE_51860_thorvg.png 5x5 - 	diff 507x507
```
Most of files differs in max 1x1 pixels

This is caused sometimes by different handling of float values, in `SVG_FILE_50623.svg` contains two properties `height='16.00001'` and `width='16'` and rsvg output have is `16x17` but thorvg `16x16`

SVG with differences - [All differences.zip](https://github.com/thorvg/thorvg/files/11389829/All.differences.zip)

SVG with bigger differences(bigger than 1x1) - [Bigger Differences.zip](https://github.com/thorvg/thorvg/files/11389832/Bigger.Differences.zip)

mgrudzinska:
I think f82c274444f86f3c70a7406cc4a8442f61e34cac is responsible for this bug, at least in some cases I've checked

qarmin:
Looks that now differences > 1x1px are fixed, but what with rest(1x1, 0x1, 1x0) ?
Goal is to do same thing like rsvg or cutting image is good enough?

mgrudzinska:
[...]

PR #1416, opened 2023-05-03
svg_loader: fixing percent. calculations of width/height

mgrudzinska:
The conversion to percentages was applied twice.
Additionally, when a viewbox value was provided before the width/height, incorrect scaling occurred.

@Issue: https://github.com/thorvg/thorvg/issues/1414


hermet (approved):

hermet:
thanks!

Issue #1278, opened 2022-11-18
`PngBuilder::build` segfault on a SVG with embedded PNG image


akien-mga:
This is a follow-up to https://github.com/godotengine/godot/issues/68822, where we use thorvg 0.8.2 in Godot, which fails to load the specific SVG in this zip: 
[testingSVG_fromInkscape_0.svg.zip](https://github.com/Samsung/thorvg/files/10045038/testingSVG_fromInkscape_0.svg.zip)

It fails loading (without crash) at this step in Godot: https://github.com/godotengine/godot/blob/c9c09ad608754a9025ca227358fca5a02553f2cb/modules/svg/image_loader_svg.cpp#L109-L113

Then I tried to see if thorvg `master` (0aec18dc) could process it by using `svg2png`, which triggers a segfault:

```
$ gdb ./build/src/bin/svg2png/svg2png
(gdb) r testingSVG_fromInkscape_0.svg 
Starting program: /home/akien/tmp/thorvg/build/src/bin/svg2png/svg2png testingSVG_fromInkscape_0.svg
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[New Thread 0x7ffff794f6c0 (LWP 615859)]
[New Thread 0x7ffff714e6c0 (LWP 615860)]
[New Thread 0x7ffff694d6c0 (LWP 615861)]
[New Thread 0x7ffff614c6c0 (LWP 615862)]
[New Thread 0x7ffff594b6c0 (LWP 615863)]
[New Thread 0x7ffff514a6c0 (LWP 615864)]
[New Thread 0x7ffff49496c0 (LWP 615865)]
[New Thread 0x7ffff41486c0 (LWP 615866)]
SVG: Unsupported attributes used [Elements type: Svg][Id : NO_ID][Attribute: stroke-miterlimit][Value: 10]
SVG: Unsupported attributes used [Elements type: Svg][Id : NO_ID][Attribute: clip-rule][Value: evenodd]
SVG: Unsupported attributes used [Elements type: Svg][Id : NO_ID][Attribute: sodipodi:docname][Value: testingVectornatorExport_fromInkscape.svg]
SVG: Unsupported attributes used [Elements type: Svg][Id : NO_ID][Attribute: inkscape:version][Value: 1.2 (dc2aeda, 2022-05-15)]
[...]

PR #1417, opened 2023-05-03
svg2png: changing image size

mgrudzinska:
For resolution > 8k image's width and height are changed, but the change was not applied to the image itself. Now fixed.


hermet (approved):

hermet:
btw, just wonder how it should have 8k limit.... @_@?

JSUYA:
> btw, just wonder how it should have 8k limit.... @_@?

Here https://github.com/thorvg/thorvg/pull/1279, Maybe heap overflow occurred (issue : https://github.com/thorvg/thorvg/issues/1278)

PR #1420, opened 2023-05-04
infra: Improved regression finder

qarmin:
CI contains 3 parts:
- Testing for regressions - testing valid files if differs between master branch and PR
- Testing for differences in currently broken files
- Testing if png files are created in reproducible way(in size)

Output is collected and everything is added into comment into PR if anything is changed.

Differences from first and last step are critical, so CI will fail if found any regression.

In second step, there is no method to say if changes are good or bad, so that is why it never fails.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -28,35 +28,82 @@ jobs:
           sudo ninja -C build install
           cd ..
 
-      - name: Download SVG Regression finder
+      - name: Download SVG Regression finder and setup settings
         run: |
-          wget https://github.com/qarmin/SVG-regression-finder/releases/download/0.2.0/svg_tester
+          wget https://github.com/qarmin/SVG-regression-finder/releases/download/0.3.0/svg_tester
           chmod +x ./svg_tester
-
-      - name: Prepare files to test
-        run: |
-          wget https://github.com/qarmin/SVG-regression-finder/releases/download/0.2.0/SVG1000.zip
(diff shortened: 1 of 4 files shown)


hermet:
@qarmin just for a tip for our contribution rule. https://github.com/thorvg/thorvg/blob/master/CONTRIBUTING.md

That would be better if you add a domain info in the commit msg like:

Improved regression finder
-> infra: Improved regression finder
or -> github: Improved regression finder
or -> test: Improved regression finder

Do not override FAIL_CI variable with 0
-> infra: Do not override FAIL_CI variable with 0
or -> github: Improved regression finder
or -> test: Improved regression finder

thanks!

.github/workflows/regression.yml
```diff
@@ -28,40 +28,92 @@ jobs:
           sudo ninja -C build install
           cd ..
 
-      - name: Download SVG Regression finder
+      - name: Download SVG Regression finder and setup settings
         run: |
-          wget https://github.com/qarmin/SVG-regression-finder/releases/download/0.2.0/svg_tester
+          wget https://github.com/qarmin/SVG-regression-finder/releases/download/0.3.0/svg_tester
```
hermet, line 33: What do you think if it has a "latest" folder instead of the specific verison? We don't need to update thorvg everytime.

hermet (approved):
Thanks, please review the comment. If it has a valid reason, let's proceed!

PR #1437, opened 2023-05-09
webp_loader: Introduce Webp image loader

JSUYA:
draft


src/examples/PictureWebp.cpp
```diff
@@ -0,0 +1,186 @@
+/*
+ * Copyright (c) 2020 - 2023 the ThorVG project. All rights reserved.
```
hermet, line 2: copyright -> 2023

src/loaders/webp/tvgWebpLoader.cpp
```diff
@@ -0,0 +1,68 @@
+/*
```
hermet, line 1: This incomplete static loader shouldn't be in this PR.

src/loaders/external_webp/tvgWebpLoader.h
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2021 - 2023 the ThorVG project. All rights reserved.
```
hermet, line 2: -> 2023

src/loaders/external_webp/tvgWebpLoader.cpp
```diff
@@ -0,0 +1,177 @@
+/*
```
hermet, line 1: -> 2023

src/examples/PictureWebp.cpp
```diff
@@ -0,0 +1,186 @@
...
+{
+    if (!canvas) return;
+
+    //Background
+    auto bg = tvg::Shape::gen();
+    bg->appendRect(0, 0, WIDTH, HEIGHT, 0, 0);    //x, y, w, h, rx, ry
+    bg->fill(255, 255, 255, 255);                 //r, g, b, a
+    canvas->push(move(bg));
+
+    //Load png file from path
```
hermet, line 40: png -> webp

src/loaders/external_webp/tvgWebpLoader.cpp
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    int width, height;
+    if (!WebPGetInfo(this->data, size, &width, &height)) return false;
+
+    w = static_cast<float>(width);
+    h = static_cast<float>(height);
+    cs = ColorSpace::ARGB8888;
+    this->size = size;
+
```
hermet, line 122: -- empty line

src/loaders/external_webp/tvgWebpLoader.cpp
```diff
@@ -0,0 +1,177 @@
...
+    if (((size = ftell(webpFile)) < 1)) goto finalize;
+    if (fseek(webpFile, 0, SEEK_SET)) goto finalize;
+
+    data = (unsigned char *) malloc(size);
+    if (!data) goto finalize;
+
+    freeData = true;
+
+    if (fread(data, size, 1, webpFile) < 1) goto failure;
+
```
hermet, line 79: -- empty line

src/loaders/meson.build
```diff
@@ -30,6 +30,17 @@ if all_loaders or get_option('loaders').contains('jpg') == true
     endif
 endif
 
+if all_loaders or get_option('loaders').contains('webp') == true
+    if get_option('static') == true
```
hermet, line 34: No one have a plan to complete it. I think adding static loader in this PR is not a good idea.

src/loaders/webp/tvgWebpLoader.h
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright (c) 2021 - 2023 the ThorVG project. All rights reserved.
```
hermet, line 2: 2023

hermet (changes requested):
Please check comments.

hermet (changes requested):
Please check comments.

src/loaders/meson.build
JSUYA: I know we don't have that plan.
But, we provide loader 'all' and 'static' options in meson options.
The above combination of options can make problems since the dependency on libwebp is not strictly required for thorvg builds.
Therefore, I think it is enough to output an error message that does not support static webp.

Of course I can add a macro to external_webp and wrap the macro inside every function in external_webp.

example)
``` 
//external_webp/meson.build
 if webp_dep.found()
     config_h.set10('THORVG_WEBP_DEPENDENCY', true)
 endif
```

Wouldn't a skeleton class be better than this?

hermet: Rather than fake-build success, User needs to notice that feature is not supported in the build-time.
probably, like this... (src/loaders/meson.build)
image

hermet:
Unnecessary commits have been screwed up during our patch review process,
Maybe we can squash them in one patch at the last otherwise, 
I recommend you to use 
"git rebase --interactive ...." command to update the existing commits then
"git push origin {branch} --force"

src/loaders/meson.build
JSUYA: Making the build fail simplifies the problem.
I've updated the PR. I deleted the webp directory.

JSUYA:
> Unnecessary commits have been screwed up during our patch review process, Maybe we can squash them in one patch at the last otherwise, I recommend you to use "git rebase --interactive ...." command to update the existing commits then "git push origin {branch} --force"

I rebased pr. but why not use github's squash and merge feature?
It works the same as rebase -i and leaves the PR number in the title, so it is easy to track.

hermet:
> > Unnecessary commits have been screwed up during our patch review process, Maybe we can squash them in one patch at the last otherwise, I recommend you to use "git rebase --interactive ...." command to update the existing commits then "git push origin {branch} --force"
> 
> I rebased pr. but why not use github's squash and merge feature? It works the same as rebase -i and leaves the PR number in the title, so it is easy to track.

1. The main purpose of rebasing is to leave only valid commits. Reviewers should not waste their time reviewing unnecessary changes.
2. By squashing patches into one, we would lose the ability to track changes commit by commit easily.
3. Either way, the history of rebasing is tracked by the Github system.
4. Not every contributor has squash/merge permissions.

.github/workflows/build_win.yml
```diff
@@ -19,11 +19,11 @@ jobs:
         pip install meson==0.58.0 ninja
     - name: Build
       run: |
-        meson --backend=ninja build -Dlog=true -Dloaders="all" -Dsavers="all" -Dbindings="capi" -Dtools="all"
+        meson --backend=ninja build -Dlog=true -Dloaders="tvg, svg, png, jpg" -Dsavers="all" -Dbindings="capi" -Dtools="all"
```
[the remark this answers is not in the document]
JSUYA: We need to install libwebp separately in the window container, but I haven't prepared it yet.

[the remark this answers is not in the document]
hermet: We can keep it all but not break the build. It just could try all the valid options at the time.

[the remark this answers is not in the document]
JSUYA: The reason the build don't stop is that if there is no external loader, the static loader is used instead.
Since our webp does not provided a static loader, it is logical that the build breaks without libwebp. (In this case, it is because webp(or all) is explicitly enabled even though static is not specified and libwebp dependency exists.)

Nevertheless, if we want to keep the build success for `all` options without libwebp, I can move libwebp's dependency check to before declaring `THORVG_WEBP_LOADER_SUPPORT`.

Please hope this is the solution.

JSUYA:
It is currently written based on libwebp 0.6.1. I guess I need to upgrade the libwebp version.

.github/workflows/build_win.yml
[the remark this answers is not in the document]
hermet: I see, please keep your change now, we can revisit this later. it's not important at this moment.

meson.build
```diff
@@ -45,6 +45,24 @@ if all_loaders or get_option('loaders').contains('jpg') == true
     config_h.set10('THORVG_JPG_LOADER_SUPPORT', true)
 endif
 
+if all_loaders or get_option('loaders').contains('webp') == true
```
hermet, line 48: Unusual script here. Commens will be helpful.
"#We should check the dependency first because static webp is not supported."

meson.build
```diff
@@ -45,6 +45,24 @@ if all_loaders or get_option('loaders').contains('jpg') == true
...
+    if get_option('static') == true
+        if all_loaders
+            message('static webp is not available, disable webp loader.')
+        elif get_option('loaders').contains('webp') == true
+            error('static webp is not available, disable webp loader.')
+        endif
+    else
+        if webp_dep.found()
+            config_h.set10('THORVG_WEBP_LOADER_SUPPORT', true)
+        else
```
hermet, line 60: else if all_loaders -> message
else -> error

Now I think we can keep the "all" option in CI script.

.github/workflows/build_win.yml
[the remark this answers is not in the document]
hermet: Please review the additional comments and then make a decision on whether you think any further changes are necessary.

JSUYA:
> There is a problem of not finding webp/decode.h in Window ci container. I'll be back when I fix this.

I still haven't been able to find a way.
The problem does not occur even with latest libwebp.
I couldn't find the location (path) of the webp header in the window container(ilammy/msvc-dev-cmd).
If anyone knows a solution to this problem please advise me
https://github.com/thorvg/thorvg/actions/runs/4972439406/jobs/8897651443?pr=1437#step:5:141

JSUYA:
> @JSUYA What do you think if we exclude webp from all_loaders since it's not 100% compatible? We could leave it under the development, this means user should explicitly turn on that option until we make it perfect both dynamic/static.

@hermet 
I think it's a good way. I'll make a new commit soon.
One more thing. I think we can add `experimental_loaders` to options of meson_option. 
what do you think?
```
option('loaders',
   type: 'array',
   choices: ['', 'tvg', 'svg', 'png', 'jpg', 'all'],
   value: ['svg', 'tvg'],
   description: 'Enable File Loaders in thorvg')

option('experimental_loaders',
   type: 'array',
   choices: ['', 'webp','lottie'],
   value: ['webp'],
   description: 'Enable file Loaders in thorvg(experimental)')
```

hermet:
> > @JSUYA What do you think if we exclude webp from all_loaders since it's not 100% compatible? We could leave it under the development, this means user should explicitly turn on that option until we make it perfect both dynamic/static.
> 
> @hermet I think it's a good way. I'll make a new commit soon. One more thing. I think we can add `experimental_loaders` to options of meson_option. what do you think? It will be useful in the loader that we will add in the future.
> 
> ```
> option('loaders',
>    type: 'array',
>    choices: ['', 'tvg', 'svg', 'png', 'jpg', 'all'],
>    value: ['svg', 'tvg'],
>    description: 'Enable File Loaders in thorvg')
> 
> option('experimental_loaders',
>    type: 'array',
>    choices: ['', 'webp','lottie'],
>    value: ['webp'],
>    description: 'Enable file Loaders in thorvg(experimental)')
> ```

In my opinion, that experimet list are volatile. Sometimes items are empty, In this case it's meaningless to keep that option. I think we can just add the experimental items naming like "webp_beta" in the loader, later change it "webp"

JSUYA:
> > > @JSUYA What do you think if we exclude webp from all_loaders since it's not 100% compatible? We could leave it under the development, this means user should explicitly turn on that option until we make it perfect both dynamic/static.
> > 
> > 
> > @hermet I think it's a good way. I'll make a new commit soon. One more thing. I think we can add `experimental_loaders` to options of meson_option. what do you think? It will be useful in the loader that we will add in the future.
> > ```
> > option('loaders',
> >    type: 'array',
> >    choices: ['', 'tvg', 'svg', 'png', 'jpg', 'all'],
> >    value: ['svg', 'tvg'],
> >    description: 'Enable File Loaders in thorvg')
> > 
> > option('experimental_loaders',
> >    type: 'array',
> >    choices: ['', 'webp','lottie'],
> >    value: ['webp'],
> >    description: 'Enable file Loaders in thorvg(experimental)')
> > ```
> 
> In my opinion, that experimet list are volatile. Sometimes items are empty, In this case it's meaningless to keep that option. I think we can just add the experimental items naming like "webp_beta" in the loader, later change it "webp", The point is which one is simpler to maintain the meson script.

Sorry late. I updated pr. Added webp_beta. The meson script has been simplified.

hermet (approved):
thanks!

Issue #1451, opened 2023-05-15
Incorrectly renegered SVG files (from Godot)


JonnyTech:
Not sure if this should be linked to #1255 but here are a couple of broken svg files that I have found - this is linked from https://github.com/godotengine/godot/issues/77094

https://raw.githubusercontent.com/lipis/flag-icons/main/flags/4x3/eu.svg
https://raw.githubusercontent.com/lipis/flag-icons/main/flags/4x3/us.svg

clemens-tolboom:
# IN
![EUR](https://raw.githubusercontent.com/lipis/flag-icons/main/flags/4x3/eu.svg)
# OUT
![d](https://user-images.githubusercontent.com/3584732/238156906-db744227-ab0f-4dac-9af4-b081d007e8dd.png)

JSUYA:
https://raw.githubusercontent.com/lipis/flag-icons/main/flags/4x3/eu.svg
I've simplified the problem a bit more.
`<use>` declared in `<defs>` is not being drawn. @mgrudzinska Could you please check this?
```html
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 100 100">
  <defs>
    <g id="d">
        <path id="a" d="m10 10 h10 v10 h-10 z"/>
        <use fill="#f00" xlink:href="#a" y="-15" />
    </g>
  </defs>
    <use fill="#fc0" xlink:href="#d" width="100%" height="100%" y="+15"/>
</svg>
```

https://raw.githubusercontent.com/lipis/flag-icons/main/flags/4x3/us.svg
@hermet @mgrudzinska `<marker-mid>` is used in this file.
https://www.w3.org/TR/SVG2/painting.html#VertexMarkerProperties
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/marker-mid
We need discussion about supporting features related to markers.

mgrudzinska:
after both, the #1452 and #1453 are applied the eu flag is correct. the us flag looks better, but since markers are used, only one star is visible. 

implementing markers sounds fun for me, but I don't know how much appreciated it will be, it's not commonly used feature....

JonnyTech:
Thank you all - are you happy for me to close this issue?

capnm:
Please have some patience, the developers ask you or close the issue after a commit itself. 
(us.svg: Unsupported elements used [Elements: marker])

[...]

PR #1452, opened 2023-05-15
svg_loader: copy the use node properties

mgrudzinska:
A bug was observed when a 'use' node was
referenced in another 'use' node.

@Issue: https://github.com/thorvg/thorvg/issues/1451

sample:
```
<svg viewBox="0 0 80 20" xmlns="http://www.w3.org/2000/svg">

  <use id="a" href="#myDot" x="5" y="5" style="fill:red" />  
  <use href="#a" x="20" y="5" style="fill:blue; opacity:0.4" />

  <symbol id="myDot" width="10" height="10" viewBox="0 0 2 2">
    <circle cx="1" cy="1" r="1" />
  </symbol>
</svg>
```

before:
Zrzut ekranu 2023-05-16 o 01 01 39

after:
Zrzut ekranu 2023-05-16 o 01 01 59

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2934,6 +2934,16 @@ static void _copyAttr(SvgNode* to, const SvgNode* from)
             }
             break;
         }
+        case SvgNodeType::Use: {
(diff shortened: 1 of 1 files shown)


JSUYA (approved):
LGTM

PR #1453, opened 2023-05-15
svg_loader: fixing finding defs node

mgrudzinska:
In the case when a 'use' node was used
in a 'defs' block  and the element it
referenced was also defined within the same
'defs' block, the reference node could not
be cloned. This was because the 'defs' node
could not be located.

@Issue: https://github.com/thorvg/thorvg/issues/1451

sample: https://github.com/thorvg/thorvg/issues/1451#issuecomment-1548089583


JSUYA (approved):
LGTM

JSUYA (approved):
LGTM

PR #1457, opened 2023-05-16
++ pkg enhancement

hermet:


vtorri:
@hermet you set the variable `stripping` to `true` by default, then if b_sanitize is != none, set it to `true`

in the target:

strip: stripping,

to have something similar to what you had

vtorri:
if you force stripping with this:

```
@@ -53,6 +53,7 @@ thorvg_lib = library(
     dependencies           : thorvg_lib_dep,
     install                : true,
     cpp_args               : compiler_flags,
+    strip                  : true,
     gnu_symbol_visibility  : 'hidden',
     override_options       : override_options
 )
```

if **you install** the lib (strip is done on install rule), are the symbols stripped ?

hermet:
> if **you install** the lib (strip is done on install rule), are the symbols stripped ?

Ah, I see. I checked the result in the build folder, but the stripping process is actually performed during the installation.

capnm:
Clang moved the stripping part from compiler to the linker ...
https://stackoverflow.com/questions/6085491/gcc-vs-clang-symbol-stripping

hermet:
> if you force stripping with this:
> 
> ```
> @@ -53,6 +53,7 @@ thorvg_lib = library(
>      dependencies           : thorvg_lib_dep,
>      install                : true,
>      cpp_args               : compiler_flags,
> +    strip                  : true,
>      gnu_symbol_visibility  : 'hidden',
>      override_options       : override_options
>  )
> ```
> 
> if **you install** the lib (strip is done on install rule), are the symbols stripped ?

So as far as I tested, strip works in default. when I turn it off with meson option, it doesn't invovled. so looks fine.

capnm:
Can confirm, meson does it in the install part:
```
[41/42] meson install --no-rebuild
Installing src/libthorvg.so.0.9.99 to thorvg_64/lib/x86_64-linux-gnu
...
Stripping target 'src/libthorvg.so.0.9.99'.
...
1.592.864 svg2png_build/src/libthorvg.so.0.9.99
  200.672 thorvg_64/lib/x86_64-linux-gnu/libthorvg.so.0.9.99
```

vtorri:
i would do the contrary:
 1. by default, the option is debugoptimized, so you shouln't strip by default. So in the project() function : `strip : false`
 2. in src/meson.build, before the b_sanitize test:

```
stripping = get_option('strip')
```

 3. in src/meson.build, in the b_sanitize test:

```
stripping = true
```

 4. below, in the library() call (around line 54), add:

```
strip: stripping,
```

hermet:
> 5\. in src/meson.build, in the b_sanitize test:
> 
> 
> ```
> stripping = true
> ```

@vtorri Thanks for comments. 2 things.
1. Sanitize test needs to strip? Just wonder why.
2. We'd rather to keep the strip in default for someone who has no idea about it at all. Even though it's enabled, no troubles in debugging most cases. I expect people could disable the strip option when it's necessary.

vtorri (approved):

capnm (approved):

Issue #1372, opened 2023-04-17
ThorVG 1.0 official APIs


hermet:
Finally, the major version of thorvg is going to be changed. This is the opportunity to clear up any mess.

- [ ] Add a Canvas parameter to all paints gen()
- [ ] Replace duplicate() with gen()
- [x] Support Paints list manipulation. 
https://github.com/thorvg/thorvg/issues/2957 
https://github.com/thorvg/thorvg/issues/2598
- [x] Remove all unique_ptr in APIs
- [x] Clear the Experimental APIs
- [x] Support Paint reference Counter
- [x] Remove all deprecated APIs
- [x] Remove all strings in the paramter

**Remove:**
- [x] Result Picture::viewbox(float* x, float* y, float* w, float* h) const;           //experimetal
- [x] Result Picture::bounds(float* x, float* y, float* w, float* h) const;            //deprecated
- [x] Result Picture::load(const char* data, uint32_t size, bool copy = false);  //derpecated 
- [x] Result Canvas::reserve(uint32_t size)    //deprecated
- [x] Result Scene::reserve(uint32_t size)      //deprecated
- [x] const uint32_t* Picture::data(uint32_t* w, uint32_t* h) const;    //experimetal
- [x] Polygon //experimetal
- [x] Vertex //experimetal
- [x] Result Picture::mesh(const Polygon* triangles, uint32_t triangleCnt); //experimetal
- [x] uint32_t Picture::mesh(const Polygon** triangles) const; //experimetal
- [x] uint32_t Paint::identifier() const;   //deprecated
- [x] uint32_t Fill::identifier() const;   //deprecated
- [x] uint32_t LinearGradient::identifier();   //deprecated
- [x] uint32_t RadialGradient::identifier();   //deprecated
- [x] uint32_t Shape::identifier();   //deprecated
- [x] uint32_t Picture::identifier();   //deprecated
- [x] uint32_t Scene::identifier();   //deprecated
- [x] uint32_t Text::identifier();   //deprecated
- [x] TVG binary format support
- [x] Result::Shape::appendArc(float cx, float cy, float radius, float startAngle, float sweep, bool pie)
https://github.com/thorvg/thorvg/issues/2632
- [ ] Result SwCanvas::mempool(MempoolPolicy policy);

//Discard the plan. Keep this for SVG
~~- [ ] Matrix Fill::transform() const https://github.com/thorvg/thorvg/issues/2859~~
[...]

PR #1458, opened 2023-05-17
API clean up

hermet:
BETA -> Official 0.10
- unique_ptr<Picture> Accessor::set(std::unique_ptr<Picture> picture, std::function<bool(const Paint* paint)> func);
- static Accessor::std::unique_ptr<Accessor> gen();
- Result Shape::order(bool strokeFirst);

Modify:
- SwCanvas::Colorspace::ABGR8888_STRAIGHT -> SwCanvas::Colorspace::ABGR8888S
- SwCanvas::Colorspace::ARGB8888_STRAIGHT -> SwCanvas::Colorspace::ARGB8888S

Removal:
- Result Picture::viewbox(float* x, float* y, float* w, float* h) const;

@Issue: https://github.com/thorvg/thorvg/issues/1372

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -990,7 +990,8 @@ public:
      * @param[in] strokeFirst If @c true the stroke is rendered before the fill, otherwise the stroke is rendered as the second one (the default option).
      *
      * @return Result::Success when succeed, Result::FailedAllocation otherwise.
-     * @BETA_API
+     *
+     * @since 0.10
      */
     Result order(bool strokeFirst) noexcept;
 
@@ -1602,7 +1603,7 @@ public:
  *
  * @warning We strongly warn you not to change the paints of a scene unless you really know the design-structure.
  *
- * @BETA_API
+ * @since 0.10
  */
 class TVG_API Accessor final
 {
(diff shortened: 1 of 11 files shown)


hermet:
@mgrudzinska I would appreciate it if you could provide the Shape::order() CAPI binding and test.

mgrudzinska (commented):
since we are considering introducing markers, the 'order' api will be deprecated sooner or later. so maybe we should change it before promoting?
like ```order(int fill, int stroke, int markers)``` ?
or don;t promote it?

hermet:
> since we are considering introducing markers, the 'order' api will be deprecated sooner or later. so maybe we should change it

@mgrudzinska I'm not certain about anything. While the feature could potentially be implemented on the SVGloader side (due to it being part of the SVG spec, not others), I currently have no plans to undertake the task of adding this feature myself. This means that markers will not be introduced unless someone intentionally attempts to do so. Instead we keep the API in beta, we might consider making it official now and then deprecating it in a future 2.0 release.

That's my opinion and I would like to accept your choice since that api was introduced by you. If you like to keep it under beta. then it's ok. but no marker parameters now.

mgrudzinska:
@hermet, let's do it your way. let's promote order api, since as you wrote, we are not sure whether and when markers will be introduced

mgrudzinska (approved):

PR #1463, opened 2023-05-18
fix compiler warnings on MacOS

hermet:
-- compiler warnings


capnm:
General clang issue :+1:
https://reviews.llvm.org/D119670 [clang] Warn on unqualified calls to std::move and std::forward

PR #1466, opened 2023-05-20
Fix clang compiler warnings.

capnm:
[clang] Warn on unqualified calls to std::move and std::forward. See: https://reviews.llvm.org/D119670

I have found some others.


hermet (approved):
thanks!

Issue #404, opened 2021-05-31
Support missing mask options


hermet:
//Mask Value
InvLumaMask = ~(0.2125*R + 0.7154*G + 0.0721*B)

//Masking Operation
AddMask  =  (source * sourceAlpha) + (target * (1 - sourceAlpha))
SubtractMask  = (source * sourceAlpha) - (target * (1 - sourceAlpha))
IntersectMask  = source * min(sourceAlpha, targetAlpha)
DifferenceMask  = abs(source - target * (1 - sourceAlpha))

Examples)
shape->composite(mask, CompositeMethod::AddMask);
shape->composite(mask, CompositeMethod::SubMask);
shape->composite(mask, CompositeMethod::DifferenceMask);
shape->composite(mask, CompositeMethod::IntersectMask);

hermet:
examples:
shape->composite(mask, Composite::AddMask | Composite::AlphaMask);
shape->composite(mask, Composite::AddMask | Composite::InvAlphaMask);
shape->composite(mask, Composite::SubMask | Composite:;AlphaMask);
shape->composite(mask, Composite::DiffMask | Composite:;LumaMask);
shape->composite(mask, Composite::IntMask | Composite:;InvLumaMask);

hermet:
https://helpx.adobe.com/fi/after-effects/using/alpha-channels-masks-mattes.html

hermet:
![shot-2023-05-27_00-14-57](https://github.com/thorvg/thorvg/assets/3711518/f9ccb8d5-7812-42ba-8b30-ededc3432c7c)

hermet:
![image](https://github.com/thorvg/thorvg/assets/3711518/d7692e04-f315-4544-852b-23689d01ba23)

PR #1470, opened 2023-05-22
Enhance the masking operations

hermet:
Issue: https://github.com/thorvg/thorvg/issues/404

image

binary size diff: 201307 -> 22693 = (+21kb)

Current ver. doesn't invole individual rgb channels composition but only alpha at the overlapped region. It needs enhance it?


hermet:
@qarminHello, it would be nicer if you could review the Regression Testing failures. It seems that the designated server is no longer valid.

hermet:
@qarmin Hello, it would be nicer if you could review the Regression Testing failures. It seems that the designated server is no longer valid.

capnm:
> It seems that the designated server is no longer valid.

Looks like some repository download limits have been reached
and the access to the  hxxp://azure.archive.ubuntu.com/ubuntu server gets blocked.

Issue #1472, opened 2023-05-22
Support SVG stroke-miterlimit


capnm:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit
https://www.w3.org/TR/SVGTiny12/painting.html#StrokeProperties

See the PR ... improve some SVG (UI) icons: https://github.com/godotengine/godot/pull/77318#issuecomment-1557575370

+ svg
![test](https://github.com/thorvg/thorvg/assets/4047289/53a45b91-6753-47df-bb05-f3a10995989e)
+ png
![test](https://github.com/thorvg/thorvg/assets/4047289/b6697330-97ff-4fa6-a3ca-8a603832783c)

```
svg2png test.svg -r 512x512
[L] SVG (../thorvg-git/src/loaders/svg/tvgXmlParser.cpp 363): 
Unsupported attributes used [Elements type: Path][Id : NO_ID][Attribute: stroke-miterlimit][Value: 1]
[L] SVG (../thorvg-git/src/loaders/svg/tvgXmlParser.cpp 363): 
Unsupported attributes used [Elements type: Path][Id : NO_ID][Attribute: stroke-miterlimit][Value: 8]
```

PR #1481, opened 2023-06-06
API refinement

hermet:
change the plan to set the default value to alpha properties in the fill/stroke apis.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -928,7 +928,7 @@ public:
      *
      * @return Result::Success when succeed, Result::FailedAllocation otherwise.
      */
-    Result stroke(uint8_t r, uint8_t g, uint8_t b, uint8_t a) noexcept;
+    Result stroke(uint8_t r, uint8_t g, uint8_t b, uint8_t a = 255) noexcept;
 
     /**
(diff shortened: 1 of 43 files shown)


hermet:
For the sake of SVG, the opacity for stroke/fill is useful. Let's discard this plan.

capnm:
@hermet I'm thinking about  #1472
Would this be similar to the alpha OK? But I don't know what to do in capi.
```
      * The join style is used for joining the two line segment while stroking the path.
      *
      * @param[in] join The join style value. The default value is @c StrokeJoin::Bevel.
+     * @param[in] miterlimit Optional Miterlimit of the stroke.
      *
      * @return Result::Success when succeed, Result::FailedAllocation otherwise.
      */
-    Result stroke(StrokeJoin join) noexcept;
+    Result stroke(StrokeJoin join, float miterlimit = 4.0f) noexcept;

+   /**
+     * @brief Gets the stroke miterlimit.
+     *
+     * @return The stroke miterlimit value when succeed, -1 if no stroke was set.
+     */
+    float strokeMiterlimit() const noexcept;
```

hermet:
@capnm hello, that change will lead the abi break. So we should introduce with a separate api.
like:
Result strokeMiterlimit(float v);
float strokeMiterlimit();

PR #1482, opened 2023-06-07
Fix clang compiler warnings in unit tests.

capnm:
[clang] Warn on unqualified calls to std::move and std::forward See: https://reviews.llvm.org/D119670
```
[141/166] Compiling C++ object test/tvgUnitTests.p/testAccessor.cpp.o ../thorvg-git/test/testAccessor.cpp:58:29: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
    picture = accessor->set(move(picture), nullptr);
                            ^
                            std::
...
```
fgrep -rlZ '(move(' test/ |  xargs -0 sed -i 's/(move(/(std::move(/g' :)

@jsuya perhaps remove the optional alpha values (except two for test) here as well.

### DIFF
--- a/test/testAccessor.cpp
+++ b/test/testAccessor.cpp
@@ -55,7 +55,7 @@ TEST_CASE("Set", "[tvgAccessor]")
     REQUIRE(accessor);
 
     //Case 1
-    picture = accessor->set(move(picture), nullptr);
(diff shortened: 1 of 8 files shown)


hermet (approved):
thanks!

Issue #1203, opened 2022-03-07
Feature request: Remove Paint from Canvas


projectitis:
Feature request:

It is possible to update a specific Paint object on the canvas using `canvas->update(paint);`.
I would also like to be able to remove a Paint object from the canvas. For example `canvas->remove(paint);`.

It would be great if this was possible!

projectitis:
If you think this issue should be included, I don't mind working on it?

hermet:
@projectitis You mean you'd like to contribute it?

projectitis:
Yes. 

I have just completed it. Am testing that it works correctly, and will then create PR and you can determine if you'd like it included or not (or other changes).

hermet:
@projectitis thanks for your contribution :) currently I'm working for tvg 0.8 release, new features can be considered after it. I will review it after the release.

PR #1483, opened 2023-06-07
introduce paints() apis that returns the list of paints

hermet:
These new apis would enable users to dynamically modify the motion scene, The data structure of the paints has been changed from an array to a list.

@APIs:
std::list<Paint*>& Canvas::paints() noexcept;
std::list<Paint*>& Scene::paints() noexcept;

@Deprecated:
Result Canvas::reserve(uint32_t size) noexcept;
Result Scene::reserve(uint32_t size) noexcept;

@Issue: https://github.com/thorvg/thorvg/issues/1203

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -18,6 +18,7 @@
 #include <functional>
 #include <memory>
 #include <string>
+#include <list>
 
 #ifdef TVG_API
     #undef TVG_API
@@ -548,14 +549,25 @@ public:
      *
      * @return Result::Success when succeed.
      */
-    Result reserve(uint32_t n) noexcept;
+    TVG_DEPRECATED Result reserve(uint32_t n) noexcept;
+
+    /**
+     * @brief Returns the list of the paints that currently held by the Canvas.
+     *
+     * This function provides the list of paint nodes, allowing users a direct opportunity to modify the scene tree.
+     *
+     * @warning  Please avoid accessing the paints during Canvas update/draw. You can access them after calling sync().
+     * @see Canvas::sync()
+     *
+     * @BETA_API
+     */
+    std::list<Paint*>& paints() noexcept;
 
     /**
      * @brief Passes drawing elements to the Canvas using Paint objects.
      *
      * Only pushed paints in the canvas will be drawing targets.
      * They are retained by the canvas until you call Canvas::clear().
-     * If you know the number of the pushed objects in advance, please call Canvas::reserve().
      *
      * @param[in] paint A Paint object to be drawn.
      *
@@ -564,7 +576,7 @@ public:
      * @retval Result::InsufficientCondition An internal error.
      *
      * @note The rendering order of the paints is the same as the order as they were pushed into the canvas. Consider sorting the paints before pushing them if you intend to use layering.
-     * @see Canvas::reserve()
+     * @see Canvas::paints()
      * @see Canvas::clear()
      */
     virtual Result push(std::unique_ptr<Paint> paint) noexcept;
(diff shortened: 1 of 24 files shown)


JSUYA (commented):
I have a question.
There was a reason(maybe lib size?) in the past that we used tvgArray.h instead of std::list(or other std functions) . If you simply need a paints list, how about add an Array API that returns it?

hermet:
> I have a question. There was a reason(maybe lib size?) in the past that we used tvgArray.h instead of std::list(or other std functions) . If you simply need a paints list, how about add an Array API that returns it?

@JSUYA Yes, that's correct. However, using an array would not be suitable in this case because it assumes a scenario where the user needs to insert or delete a specific element in the list, which is not possible with an array. Additionally, the current usage of the list is limited in terms of size. As you can see, there are only two cases on the thorvg side, so it doesn't significantly impact the size (216226 -> 216350, +130 bytes). On the other hand, thorvg internally uses an array with many more cases, so size becomes a relevant factor unlike the situation with the list. Also once we open the data-structure interfaces with our own, it would need a lot of burdensome to make it possible for covering the whole use-cases. I'm not sure it's good to way go.

JSUYA:
> > I have a question. There was a reason(maybe lib size?) in the past that we used tvgArray.h instead of std::list(or other std functions) . If you simply need a paints list, how about add an Array API that returns it?
> 
> @JSUYA Yes, that's correct. However, using an array would not be suitable in this case because it assumes a scenario where the user needs to insert or delete a specific element in the list, which is not possible with an array. Additionally, the current usage of the list is limited in terms of size.

tvg builds a paint tree every frame and draws it.  I don't know if insert and delete scenarios are suitable our goals(?) in the old days. hahaha... I'm not against it. 
Just because I don't know why this feature is needed. It is expected for lottie support, but I don't know which scenario of lottie needs insert and delete. (This is just a curiosity.)

Doesn't Picture provide it? If it is provided in tvg::Picture, since Accessor and concept actually overlap, Accessor can be deprecated.

JSUYA (approved):

hermet:
> > > I have a question. There was a reason(maybe lib size?) in the past that we used tvgArray.h instead of std::list(or other std functions) . If you simply need a paints list, how about add an Array API that returns it?
> > 
> > 
> > @JSUYA Yes, that's correct. However, using an array would not be suitable in this case because it assumes a scenario where the user needs to insert or delete a specific element in the list, which is not possible with an array. Additionally, the current usage of the list is limited in terms of size.
> 
> tvg builds a paint tree every frame and draws it. I don't know if insert and delete scenarios are suitable our goals(?) in the old days. hahaha... I'm not against it. Just because I don't know why this feature is needed. It is expected for lottie support, but I don't know which scenario of lottie needs insert and delete. (This is just a curiosity.)
> 
> Doesn't Picture provide it? If it is provided in tvg::Picture, since Accessor and concept actually overlap, Accessor can be deprecated.

I believe our goal hasn't been concrete so far. The project is still in development and open to user feedback and requests. If necessary, we can make compromises. For example, as mentioned in the request at https://github.com/thorvg/thorvg/issues/1203, Also there has been a demand for supporting Lottie animation on the Tizen EFL side. The motion scene tree can be modified, and the choice is left to the users whether they want to keep the scene tree and re-use it for the optimal performance(this is the approach currently tizen efl is doing) or flush and reconstruct it for their convenience. Regardless of this feature, they can do it anyway. Lastly, the internal scene tree would'nt be exposed from the tvg::Picture because user don't make it. So Accessor is still valid for tvg::Picture use. tvg::Scene? user knows exactly the scene tree because they manually compose it. At least they are a big different point.

Issue #1471, opened 2023-05-22
rgb composition optimization


hermet:
Currently, the composition assumes that the bitmap images have RGBA channels. If it detects the alpha channel in advance and branches out the composition method between RGBA and RGB, it would enhance performance.

PR #1488, opened 2023-06-12
sw_engine: optimize 64-bit rasterization.

hermet:
Write the pixels per 64-bit data.
This optimization can potentially double the performance in the best case scenario. It's visually noticeable, especially when the resolution is sufficiently large.

https://github.com/thorvg/thorvg/issues/1471


hermet:
Let's lower the priority for considerations regarding 32-bit systems.
64-bit is the primary platform we are going to focus on.
(This doesn't means this patch breaks the 32 bits system)

PR #1504, opened 2023-06-20
common canvas: Alter the concept of the Canvas::clear() API.

hermet:
We've changed the behavior of Canvas::clear(false).

The canvas::clear(false) now retains the paints, allowing the user to update the next frame more easily without having to recompose the paint list.

Previously, clear(false) removed the paint list from the canvas, requiring the user to re-push all of them in the next frame.

Now, we offer the Canvas::paints() API, allowing users to modify the paint list directly instead.

This change will break compatibility with previous versions of Canvas.

Therefore, this update is included in ThorVG 1.0.

Issue: https://github.com/thorvg/thorvg/issues/1372


hermet:
This patch might break the compatibility.
Thus, it should be included after tvg 1.10 release.

src/lib/tvgCanvasImpl.h
```diff
@@ -64,13 +64,14 @@ struct Canvas::Impl
         if (!renderer || !renderer->clear()) return Result::InsufficientCondition;
 
         //Free paints
-        for (auto paint : paints) {
-            paint->pImpl->dispose(*renderer);
-            if (free) delete(paint);
+        if (free) {
+            for (auto paint : paints) {
+                paint->pImpl->dispose(*renderer);
```
JSUYA, line 69: We could use `clear(false)` if the paint object was managed externally.
Um.. maybe, paint's dispose is not called separately in paint's destructor, so I concerned about memory leaks in this part. Are there other alternatives without breaking backwards compatibility?

JoogabYun:
@JSUYA 
thank you very much!!

hermet:
> Hi @JoogabYun :) If you are planning to update the code your repository, i recommend paying attention to this PR. DALi uses `clear(false)`.

@JoogabYun  @JSUYA  You can easily replace the usage:

//canvas->clear(false);
auto& paints = canvas->paiints().reset();
paints.clear();

src/lib/tvgCanvasImpl.h
hermet: dispose() is changed again.

Upgrading to a major version provides an good opportunity to eliminate poor APIs or usages. 
The clear(false) method has demonstrated this need, as indicated by a @warning.
When we have valid reasons, we make changes. This decision is consistent with our policy.

Let's see.

PR #1508, opened 2023-06-26
common array: ++enhance the usability.

hermet:
+ end() which indicates the end of the data pointer.


vtorri:
why don't you use a std::array ?

hermet:
to squash up the binary size.

PR #1510, opened 2023-06-28
introduce Paint::scale(float sx, float sy);

hermet:


hermet:
not quite necessary.

hermet:
keep this if it's necessary later.

PR #1511, opened 2023-06-29
common array: revise the interface.

hermet:
Implement a pair of access methods: last() and first().

### DIFF
--- a/src/lib/tvgArray.h
+++ b/src/lib/tvgArray.h
@@ -73,9 +73,14 @@ struct Array
         return data + count;
     }
 
-    T* last() const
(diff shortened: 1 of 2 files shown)


src/lib/tvgArray.h
```diff
@@ -73,9 +73,14 @@ struct Array
         return data + count;
     }
 
-    T* last() const
+    T& last() const
```
vtorri, line 76: if the method is const, return a const ref.

hermet: @vtorri thanks, but please patches! :)

PR #1515, opened 2023-07-03
common: code refactoring

hermet:
Replace individual array implementation with a common feature.


vtorri (commented):
like free, no need to test if the pointer is NULL or not (lines 114, 175, etc...)

any reason you don't implement copy constructor for Array ?

hermet:
@vtorri no reason so far. we can add copy constructor anytime if it's useful.

PR #1523, opened 2023-07-14
Refactoring of GlGeometry::decomposeOutline due to an internal change  in RenderShape data structure.

vpxyz:
Hi, 
due to the recent change to RenderShape the opengl backend no longer compiles. This commit fixes that. 
Thank you.

Vincenzo


hermet (approved):
thanks!

PR #1528, opened 2023-07-18
examples webp: --compiler warning.

hermet:
1 warning generated.
[122/183] Compiling C++ object src/examples/PictureWebp.p/PictureWebp.cpp.o ../thorvg-git/src/examples/PictureWebp.cpp:38:18: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
    canvas->push(move(bg));
                 ^
                 std::
../thorvg-git/src/examples/PictureWebp.cpp:53:26: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
        if (canvas->push(move(picture)) != tvg::Result::Success) return;
                         ^
                         std::
../thorvg-git/src/examples/PictureWebp.cpp:77:18: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
    canvas->push(move(picture));
                 ^
                 std::

### DIFF
--- a/src/examples/PictureWebp.cpp
+++ b/src/examples/PictureWebp.cpp
@@ -35,7 +35,7 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     auto bg = tvg::Shape::gen();
     bg->appendRect(0, 0, WIDTH, HEIGHT);    //x, y, w, h
     bg->fill(255, 255, 255);                //r, g, b
-    canvas->push(move(bg));
+    canvas->push(std::move(bg));
 
     //Load webp file from path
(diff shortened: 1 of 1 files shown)


vtorri:
in my c++ coding, i **never** remove the `std` namespace (with namespace directive or declaration) because of this kind of problems

capnm:
> in my c++ coding, i **never** remove the `std` namespace (with namespace directive or declaration) because of this kind of problems

IMO it is fine to do this locally to avoid very long, unreadable names, but it isn't recommended to do this in include files.

Issue #1526, opened 2023-07-17
warnings 'comparison of integer expressions of different signedness'


vtorri:
on Windows (msys2 mingw-w64):

```
[30/173] Compiling C++ object src/libthorvg.a.p/lib_sw_engine_tvgSwRaster.cpp.obj
../src/lib/sw_engine/tvgSwRaster.cpp: In function 'void _rasterMaskedRectInt(SwSurface*, const SwBBox&, uint8_t, uint8_t, uint8_t, uint8_t)':
../src/lib/sw_engine/tvgSwRaster.cpp:312:58: warning: comparison of integer expressions of different signedness: 'uint32_t' {aka 'unsigned int'} and 'SwCoord' {
aka 'long int'} [-Wsign-compare]
  312 |     for (uint32_t y = surface->compositor->bbox.min.y; y < surface->compositor->bbox.max.y; ++y) {
      |                                                        ~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/lib/sw_engine/tvgSwRaster.cpp:314:15: warning: comparison of integer expressions of different signedness: 'uint32_t' {aka 'unsigned int'} and 'const SwCo
ord' {aka 'const long int'} [-Wsign-compare]
  314 |         if (y == region.min.y) {
      |             ~~^~~~~~~~~~~~~~~
../src/lib/sw_engine/tvgSwRaster.cpp:315:38: warning: comparison of integer expressions of different signedness: 'uint32_t' {aka 'unsigned int'} and 'const SwCo
ord' {aka 'const long int'} [-Wsign-compare]
  315 |             for (uint32_t y2 = y; y2 < region.max.y; ++y2) {
      |                                   ~~~^~~~~~~~~~~~~~
../src/lib/sw_engine/tvgSwRaster.cpp: In function 'void _rasterMaskedRleInt(SwSurface*, SwRleData*, uint8_t, uint8_t, uint8_t, uint8_t)':
../src/lib/sw_engine/tvgSwRaster.cpp:502:58: warning: comparison of integer expressions of different signedness: 'uint32_t' {aka 'unsigned int'} and 'SwCoord' {
aka 'long int'} [-Wsign-compare]
  502 |     for (uint32_t y = surface->compositor->bbox.min.y; y < surface->compositor->bbox.max.y; ++y) {
      |                                                        ~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #1530, opened 2023-07-22
sw_engine: fix compiler warning

mgrudzinska:
Warning: comparison of integer expressions of different signedness

@Issue: https://github.com/thorvg/thorvg/issues/1526


src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -309,10 +309,10 @@ static void _rasterMaskedRectInt(SwSurface* surface, const SwBBox& region, uint8
     auto h = static_cast<uint32_t>(region.max.y - region.min.y);
     auto cstride = surface->compositor->image.stride;
 
-    for (uint32_t y = surface->compositor->bbox.min.y; y < surface->compositor->bbox.max.y; ++y) {
+    for (signed long y = surface->compositor->bbox.min.y; y < surface->compositor->bbox.max.y; ++y) {
```
hermet, line 312: "auto" is more neat than "signed long".

PR #1532, opened 2023-07-25
svg_loader: clip composed of multiple transformed shapes

mgrudzinska:
Up until now, loading a clip composed of multiple
shapes worked correctly provided that they did not
undergo separate transformations. Now, such a case
has been handled.
Also the 'use' node inside the clip is supported.

this pr is a response to the [comment](https://github.com/thorvg/thorvg/pull/985#discussion_r746232286)

sample:
```

<clipPath id="clip">
    <circle cx="30" cy="50" r="30" transform="matrix(1 0.44 1 1 0 10)"/>
    <circle cx="40" cy="60" r="10" transform="matrix(1 0.4 0 1 30 -50)"/>
</clipPath>

<rect x="20" y="20" width="60" height="60" fill="blue" fill-opacity="0.2"/>
<rect x="0" y="0" width="100" height="100"  fill="red" clip-path="url(#clip)"
      fill-opacity="0.5"/>

<circle cx="30" cy="50" r="30" transform="matrix(1 0.44 1 1 0 10)" opacity="0.5"/>
<circle cx="40" cy="60" r="10" transform="matrix(1 0.4 0 1 30 -50)" opacity="0.5"/>

</svg>
```

before:
Zrzut ekranu 2023-07-25 o 02 24 39

after (expected):
Zrzut ekranu 2023-07-25 o 02 23 01

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -62,6 +62,7 @@
 /************************************************************************/
 
 static bool _appendShape(SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath);
+static bool _appendClipShape(SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath, const Matrix* transform);
 static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox, const string& svgPath, bool mask, int depth, bool* isMaskWhite = nullptr);
 
 
@@ -219,20 +220,30 @@ static unique_ptr<RadialGradient> _applyRadialGradientProperty(SvgStyleGradient*
 }
 
 
-static bool _appendChildShape(SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath)
+//The SVG standard allows only for 'use' nodes that point directly to a basic shape.
+static bool _appendClipUseNode(SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath)
 {
-    auto valid = false;
+    if (node->child.count != 1) return false;
+    auto child = *(node->child.data);
 
-    if (_appendShape(node, shape, vBox, svgPath)) valid = true;
-
-    if (node->child.count > 0) {
-        auto child = node->child.data;
-        for (uint32_t i = 0; i < node->child.count; ++i, ++child) {
-            if (_appendChildShape(*child, shape, vBox, svgPath)) valid = true;
-        }
+    Matrix finalTransform = {1, 0, 0, 0, 1, 0, 0, 0, 1};
+    if (node->transform) finalTransform = *node->transform;
+    if (node->node.use.x != 0.0f || node->node.use.y != 0.0f) {
+        Matrix m = {1, 0, node->node.use.x, 0, 1, node->node.use.y, 0, 0, 1};
+        finalTransform = mathMultiply(&finalTransform, &m);
     }
+    if (child->transform) finalTransform = mathMultiply(child->transform, &finalTransform);
 
-    return valid;
+    return _appendClipShape(child, shape, vBox, svgPath, mathIdentity((const Matrix*)(&finalTransform)) ? nullptr : &finalTransform);
+}
+
+
+static bool _appendClipChild(SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath, bool clip)
+{
+    if (node->type == SvgNodeType::Use) {
+        return _appendClipUseNode(node, shape, vBox, svgPath);
+    }
+    return _appendClipShape(node, shape, vBox, svgPath, nullptr);
 }
 
 
@@ -251,10 +262,10 @@ static void _applyComposition(Paint* paint, const SvgNode* node, const Box& vBox
             auto comp = Shape::gen();
 
             auto child = compNode->child.data;
-            auto valid = false; //Composite only when valid shapes are existed
+            auto valid = false; //Composite only when valid shapes exist
 
             for (uint32_t i = 0; i < compNode->child.count; ++i, ++child) {
-                if (_appendChildShape(*child, comp.get(), vBox, svgPath)) valid = true;
+                if (_appendClipChild(*child, comp.get(), vBox, svgPath, compNode->child.count > 1)) valid = true;
             }
 
             if (node->transform) {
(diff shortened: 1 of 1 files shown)


hermet:
ThorVG supports the scene clipper. You might wanna compose a scene using multiple clippers, rather than adopting this approach?

https://github.com/thorvg/thorvg/issues/524

mgrudzinska:
> ThorVG supports the scene clipper. You might wanna compose a scene with multiple clippers, rather than adopting this approach?
> 
> #524

I know :) but the svg std doesn't really allow for clip being too complex: 

[A ‘clipPath’ element can contain ‘path’ elements, ‘text’ elements, basic shapes (such as ‘circle’) or a ‘use’ element. If a ‘use’ element is a child of a ‘clipPath’ element, it must directly reference ‘path’, ‘text’ or basic shape elements. Indirect references are an error (see Error processing).](https://www.w3.org/TR/2012/WD-SVG2-20120828/masking.html#EstablishingANewClippingPath)

so while the clip is build we have to check, whether there is not too much nesting, ```<g>``` tags etc.
or we can assume that the clip is more or less correct...

@hermet do you prefer to use scene for a clip rather than this approach?

hermet:
> > ThorVG supports the scene clipper. You might wanna compose a scene with multiple clippers, rather than adopting this approach?
> > #524
> 
> I know :) but the svg std doesn't really allow for clip being too complex:
> 
> [A ‘clipPath’ element can contain ‘path’ elements, ‘text’ elements, basic shapes (such as ‘circle’) or a ‘use’ element. If a ‘use’ element is a child of a ‘clipPath’ element, it must directly reference ‘path’, ‘text’ or basic shape elements. Indirect references are an error (see Error processing).](https://www.w3.org/TR/2012/WD-SVG2-20120828/masking.html#EstablishingANewClippingPath)
> 
> so while the clip is build we have to check, whether there is not too much nesting, `<g>` tags etc. or we can assume that the clip is more or less correct...
> 
> @hermet do you prefer to use scene for a clip rather than this approach?

@mgrudzinska I don't prefer the case-by-case approach for the special logic; a more intuitive and simpler method is preferred for better maintenance. Please consider this; the decision is ultimately up to you. I respect your choice.

mgrudzinska:
As @hermet suggested: @1546 solves this as well and also more

hermet:
@mgrudzinska What is better to go? this one? or 1546?

mgrudzinska:
> @mgrudzinska What is better to go? this one? or 1546?

I'd stay with 1546. same logic for mask and clip, easier to maintain, as you said. but I waited for you decision before closing this one

hermet:
@mgrudzinska In my opinion, the performance of this case is not a significant factor in real-world scenarios. Therefore, maintenance is more important.

mgrudzinska:
@hermet since there are still some issues with a scene used as a clip I think that we could reconsider this patch. 
in this approach I need to add a use node support, but before spending time on it I'd prefer to know your opinion - do you prefer to apply the #1546 instead of this one?

hermet:
@mgrudzinska, as I mentioned earlier, the decision is ultimately up to you, and I respect your choice. Please push it if you think it's better. If you believe it's beneficial to leave any history for later maintenance or refactoring, you can retain the information in the code. That way, someone could review it later and switch to the scene clipper appropriately.

mgrudzinska:
@JSUYA  I'd really appreciate any comment/question from you. I checked all the samples that we considered in #1546 and they should be drawn properly

JSUYA:
> @JSUYA I'd really appreciate any comment/question from you. I checked all the samples that we considered in https://github.com/thorvg/thorvg/pull/1546 and they should be drawn properly

I checked the SVG example mentioned in #1546. And I was check that clipping is done as a group type tag.
(I don't know why `<circle>` in `defs` is being drawn to the screen. Was this originally like this? Or is it a bug?.. anyway..)

#524 mentioned by @hermet above seems to have been fixed, but still seems to need to fix.
This is sample code
```cpp
    if (!canvas) return;
    shape->appendRect(0, 0, WIDTH, HEIGHT);          //x, y, w, h
    shape->fill(255, 255, 255);                      //r, g, b

    if (canvas->push(std::move(shape)) != tvg::Result::Success) return;
    scene->opacity(125);

    shape1->appendCircle(200, 200, 400, 400);    //cx, cy, radiusW, radiusH
    shape1->fill(0, 0, 0, 255);              //r, g, b, a
    scene->push(move(shape1));

    //Create another Scene
    auto scene2 = tvg::Scene::gen();

    shape4->appendCircle(150, 150, 200, 200);    //cx, cy, radiusW, radiusH
    shape4->fill(0, 0, 255, 125);
    scene2->push(move(shape4));

    auto shape5 = tvg::Shape::gen();
    shape5->appendCircle(250, 250, 200, 200);    //cx, cy, radiusW, radiusH
    shape5->fill(255, 0, 0, 125);
    scene2->push(move(shape5));

    scene->composite(move(scene2), tvg::CompositeMethod::ClipPath);

    //Draw the Scene onto the Canvas
    canvas->push(move(scene));
```
It looks like the scenes clipped to each shape in scene2 are composited twice.

But because we have a lot of things(issues.. job..)
I think it's a good idea to focus on scene clipping again after applying this patch first.
Sorry for not being able to actively help contribute code
:)

mgrudzinska:
> I don't know why `<circle>` in `defs` is being drawn to the screen. Was this originally like this? Or is it a bug?.. anyway..

it's because there's a typo :) ```def``` instead of ```defs``` :)

> But because we have a lot of things(issues.. job..) I think it's a good idea to focus on scene clipping again after applying this patch first. Sorry for not being able to actively help contribute code :)

the sample you mentioned - this is a problem with clipping, but the problem exists inside the renderer, not in the svg_loader, so this has nothing to do with this pr. 

> Sorry for not being able to actively help contribute code

I really appreciate you fast answer, thank you very much :)


@JSUYA My main question is: What do you think about using **_this approach_** in the SVG loader and not making any changes even after the scene clipping issues are resolved in the renderer? By **_this approach_**, I mean not using a scene as clipping in the svg loader since it's not necessary. We can meet all the SVG standards using just a single shape.

JSUYA (approved):
> @JSUYA My main question is: What do you think about using this approach in the SVG loader and not making any changes even after the scene clipping issues are resolved in the renderer? By this approach, I mean not using a scene as clipping in the svg loader since it's not necessary. We can meet all the SVG standards using just a single shape.

I expected there to be a more complicated structure of SVG file, but I haven't been able to find it yet. This is enough to satisfy the needs (spec), and if a problem arises, I think we should think about it then.

PR #1546, opened 2023-07-29
svg_loader: clipPath read as a scene

mgrudzinska:
After introducing scene-based clipping support
in #1533, it is possible to load the clipPath
element as a scene instead of a shape. This enables correct transformation of each clipPath element,
which was not previously possible.

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -62,12 +62,12 @@
 /************************************************************************/
 
 static bool _appendShape(SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath);
-static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox, const string& svgPath, bool mask, int depth, bool* isMaskWhite = nullptr);
+static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox, const string& svgPath, SvgNodeType compositionType, int depth, bool* isMaskWhite = nullptr);
 
 
 static inline bool _isGroupType(SvgNodeType type)
 {
-    if (type == SvgNodeType::Doc || type == SvgNodeType::G || type == SvgNodeType::Use || type == SvgNodeType::ClipPath || type == SvgNodeType::Symbol) return true;
+    if (type == SvgNodeType::Doc || type == SvgNodeType::G || type == SvgNodeType::Use || type == SvgNodeType::Symbol) return true;
     return false;
 }
 
@@ -248,22 +248,16 @@ static void _applyComposition(Paint* paint, const SvgNode* node, const Box& vBox
         if (compNode && compNode->child.count > 0) {
             node->style->clipPath.applying = true;
 
-            auto comp = Shape::gen();
+            auto comp = _sceneBuildHelper(compNode, vBox, svgPath, SvgNodeType::ClipPath, 0);
+            if (comp) {
+                if (node->transform) {
+                    auto m = comp->transform();
+                    m = mathMultiply(node->transform, &m);
+                    comp->transform(m);
+                }
 
-            auto child = compNode->child.data;
-            auto valid = false; //Composite only when valid shapes are existed
-
-            for (uint32_t i = 0; i < compNode->child.count; ++i, ++child) {
-                if (_appendChildShape(*child, comp.get(), vBox, svgPath)) valid = true;
+                paint->composite(std::move(comp), CompositeMethod::ClipPath);
             }
-
-            if (node->transform) {
-                auto m = comp->transform();
-                m = mathMultiply(node->transform, &m);
-                comp->transform(m);
-            }
-
-            if (valid) paint->composite(std::move(comp), CompositeMethod::ClipPath);
 
             node->style->clipPath.applying = false;
         }
@@ -280,7 +274,7 @@ static void _applyComposition(Paint* paint, const SvgNode* node, const Box& vBox
             node->style->mask.applying = true;
 
             bool isMaskWhite = true;
-            auto comp = _sceneBuildHelper(compNode, vBox, svgPath, true, 0, &isMaskWhite);
+            auto comp = _sceneBuildHelper(compNode, vBox, svgPath, SvgNodeType::Mask, 0, &isMaskWhite);
             if (comp) {
                 if (node->transform) comp->transform(*node->transform);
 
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -718,26 +712,27 @@ static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox,
...
         // For a Symbol node, the viewBox transformation has to be applied first - see _useBuildHelper()
-        if (!mask && node->transform && node->type != SvgNodeType::Symbol) scene->transform(*node->transform);
+        if (node->transform && node->type != SvgNodeType::Symbol) scene->transform(*node->transform);
 
-        if (node->display && node->style->opacity != 0) {
+        if ((node->display || compositionType == SvgNodeType::ClipPath) && node->style->opacity != 0) {
             auto child = node->child.data;
             for (uint32_t i = 0; i < node->child.count; ++i, ++child) {
                 if (_isGroupType((*child)->type)) {
                     if ((*child)->type == SvgNodeType::Use)
```
JSUYA, line 724: Symbol types also know that clip-path can be used. could you please check it?

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -718,26 +712,27 @@ static unique_ptr<Scene> _sceneBuildHelper(const SvgNode* node, const Box& vBox,
...
                         scene->push(_useBuildHelper(*child, vBox, svgPath, depth + 1, isMaskWhite));
-                    else
-                        scene->push(_sceneBuildHelper(*child, vBox, svgPath, false, depth + 1, isMaskWhite));
+                    // According to the SVG standard ClipPath doesn't support <g> tag
+                    else if (compositionType != SvgNodeType::ClipPath)
+                        scene->push(_sceneBuildHelper(*child, vBox, svgPath, SvgNodeType::Unknown, depth + 1, isMaskWhite));
```
JSUYA, line 728: ClipPath can also have children. I haven't tested it, but it seems that the clip path node won't build the scene even if it has children. Is that right?

mgrudzinska: @JSUYA can you elaborate on the children? 

1) from https://www.w3.org/TR/2012/WD-SVG2-20120828/masking.html#ClippingPaths:
A ‘[clipPath](https://www.w3.org/TR/2012/WD-SVG2-20120828/masking.html#ClipPathElement)’ element can contain ‘[path](https://www.w3.org/TR/2012/WD-SVG2-20120828/paths.html#PathElement)’ elements, ‘[text](https://www.w3.org/TR/2012/WD-SVG2-20120828/text.html#TextElement)’ elements, [basic shapes](https://www.w3.org/TR/2012/WD-SVG2-20120828/shapes.html) (such as ‘[circle](https://www.w3.org/TR/2012/WD-SVG2-20120828/shapes.html#CircleElement)’) or a ‘[use](https://www.w3.org/TR/2012/WD-SVG2-20120828/struct.html#UseElement)’ element. If a ‘[use](https://www.w3.org/TR/2012/WD-SVG2-20120828/struct.html#UseElement)’ element is a child of a ‘[clipPath](https://www.w3.org/TR/2012/WD-SVG2-20120828/masking.html#ClipPathElement)’ element, it must directly reference ‘[path](https://www.w3.org/TR/2012/WD-SVG2-20120828/paths.html#PathElement)’, ‘[text](https://www.w3.org/TR/2012/WD-SVG2-20120828/text.html#TextElement)’ or [basic shape](https://www.w3.org/TR/2012/WD-SVG2-20120828/shapes.html) elements. Indirect references are an error (see [Error processing](https://www.w3.org/TR/2012/WD-SVG2-20120828/implnote.html#ErrorProcessing)).

2) from https://www.w3.org/TR/SVG2/render.html#ClippingAndMasking:
SVG supports the following clipping/masking features:
clipping paths, which either uses any combination of ‘[path](https://www.w3.org/TR/SVG2/paths.html#PathElement)’, ‘[text](https://www.w3.org/TR/SVG2/text.html#TextElement)’ and [basic shapes](https://www.w3.org/TR/SVG2/shapes.html#TermBasicShapeElement) or basic shapes to serve as the outline of a (in the absence of anti-aliasing) 1-bit mask, where everything on the "inside" of the outline is allowed to show through but everything on the outside is masked out
masks, which are [container elements](https://www.w3.org/TR/SVG2/struct.html#TermContainerElement) which can contain [graphics elements](https://www.w3.org/TR/SVG2/struct.html#TermGraphicsElement) or other container elements which define a set of graphics that is to be used as a semi-transparent mask for compositing foreground objects into the current background.
Both, clipping and masking, are specified in the module CSS Masking [[css-masking-1](https://www.w3.org/TR/SVG2/refs.html#ref-css-masking-1)].

am I missing something?

JSUYA: Sorry my explanation wasn't good enough.

`clipPath` can have `path`, `text`, or basic shapes. (according to the spec.)
These elements need to be combined into one scene, not individually clipped.

This is the test svg.
[this pr result]


[chrome]


```html
<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
  <clipPath id="myClip">
    <circle cx="40" cy="20" r="40" opacity="0.5" />
    <circle cx="40" cy="60" r="20" fill="#FF0" opacity = "0.5"/>
  </clipPath>

  <circle id="test" cx="50" cy="50" r="80" fill="#000" opacity = "0.5" clip-path="url(#myClip)"/>
</svg>

```

JSUYA: Ah ha! I understand it. 
If so, it seems that this PR could be merged after the scene clipping issue is resolved.
it can be affects the current SVG output result.

I found out through this review...

> or a ‘use’ element. If a ‘use’ element is a child of a ‘clipPath’ element, it must directly reference ‘path’, ‘text’ or basic shape elements. Indirect references are an error (see Error processing).
(If the indirect references are correct in the example below)

This case didn't work properly. (it wasn't supported before.)

```html
<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
  <def>
    <circle id="myCircle" r="30"  />
  </def>

  <clipPath id="myClip">
    <circle cx="50" cy="30" r="30" opacity="0.5" />
    <circle cx="30" cy="70" r="30" fill="#FF0" opacity = "0.5"/>
    <use href="#myCircle" x="70" y="70" fill="blue" />
  </clipPath>

  <circle id="test" cx="50" cy="50" r="40" fill="#000" opacity = "0.5" clip-path="url(#myClip)"/>
</svg>

```

since it adds clipPath elements to the scene(This PR), I expected at least an overlapping translucent circle when `<use>` was used.
Of course, additional PR may be needed, but I think it can be solved within this PR.

What do you think?

mgrudzinska: @JSUYA thx! but this is also a problem with a scene clipping, not the svg loader: #1573

mgrudzinska:
either this approach or the #1532 should be applied

mgrudzinska:
#1532 fulfills all of the SVG standard regarding clipPath, so this pr is no more needed

PR #1547, opened 2023-07-29
svg_loader: support other units for mask and clip

mgrudzinska:
Added support for the objectBoundingBox value of
the clipPathUnits and for the userSpaceOnUse value
of the maskContentUnits.

@issue: https://github.com/thorvg/thorvg/issues/429

CONTAINS #1546


mgrudzinska:
needs to be rebased after we decide between #1532 and #1546

mgrudzinska:
replaced by #1576

PR #1564, opened 2023-08-01
Clean up APIs

hermet:


inc/thorvg.h
```diff
@@ -372,26 +372,9 @@ class TVG_API Paint
      *
      * @return Result::Success when the blending method is successfully set.
      *
-     * @BETA_API
```
mgrudzinska, line 375: @hermet don't we want to add "@since..." ?

hermet: @mgrudzinska updated, thanks

inc/thorvg.h
```diff
@@ -1490,12 +1421,12 @@ class TVG_API SwCanvas final : public Canvas
...
      * @brief Enumeration specifying the methods of combining the 8-bit color channels into 32-bit color.
      */
-    enum Colorspace
+    enum Colorspace : uint8_t
     {
         ABGR8888 = 0,      ///< The channels are joined in the order: alpha, blue, green, red. Colors are alpha-premultiplied. (a << 24 | b << 16 | g << 8 | r)
         ARGB8888,          ///< The channels are joined in the order: alpha, red, green, blue. Colors are alpha-premultiplied. (a << 24 | r << 16 | g << 8 | b)
-        ABGR8888S,         ///< @BETA_API The channels are joined in the order: alpha, blue, green, red. Colors are un-alpha-premultiplied.
-        ARGB8888S,         ///< @BETA_API The channels are joined in the order: alpha, red, green, blue. Colors are un-alpha-premultiplied.
+        ABGR8888S,         ///< The channels are joined in the order: alpha, blue, green, red. Colors are un-alpha-premultiplied. (a << 24 | b << 16 | g << 8 | a) @since 1.0
```
mgrudzinska, line 1428: ```a << 24 | b << 16 | g << 8 | a``` -> ```a << 24 | b << 16 | g << 8 | r``` (```r``` on the end)

mgrudzinska (commented):
@hermet any reason why enums: Result/PathCommand/CanvasEngine/MempoolPolicy were not forced to be uint8_t like the others?

inc/thorvg.h
hermet: fixed thanks.

hermet:
> @hermet any reason why enums: Result/PathCommand/CanvasEngine/MempoolPolicy were not forced to be uint8_t like the others?

Probably, we can additionally include the PathCommand here, but there is no point in changing them because they are usually used as volatile values in most cases. 32 bits/64 bits are better to use in decent architectures. So it's more like the perf vs consistency problem.

hermet:
Need to consider PathCommand to change or not. It breaks the C API efficiency.

PR #1568, opened 2023-08-01
tests: bounds test ++

mgrudzinska:
The 'transformed' arg of the 'bounds' api
wasn't tested. Added.


hermet:
> @hermet for now the reset() api doesn't reset the transformation matrix. would you be open for changing it or is this an unnecessary change?

Considering the scene graph, retaining the transform is better. Let's keep the transform; it was the original intention.

hermet (approved):

PR #1569, opened 2023-08-01
utils introduced + strtof in lottie loader

mgrudzinska:


hermet:
here is my suggstion:

//1. speicify the util functions more concretely like tvgMath.h
tvgUtil.h -> tvgStr.h:

//functions names? (or you can suggest better names)
namespace tvg {
    strDup() or strDuplicate()
    strToFloat()
}

//2. Promote the functions to common layer.

# option1 (simple way. Other utilities are in src/lib now)
 put it in src/lib 

#option 2:
introduce src/utility(or you can suggest a better name) and reorganize the structure like this.
-src/utility
    - tvgMath
    - tvgBezier
    - tvgArray
    - tvgLzw
    - tvgStr

mgrudzinska:
@hermet please check now. I'd prefer names that keep the origin fun name, like tvgStrtof and tvgStrndup, so it is clear what they are doing, but if you wish I'll rename them

hermet:
> @hermet please check now. I'd prefer names that keep the origin fun name, like tvgStrtof and tvgStrndup, so it is clear what they are doing, but if you wish I'll rename them

@mgrudzinska Internal functions doesn't need tvg Prefix. we can use tvg namespace instead. please keep the consistency.

hermet (changes requested):
please check a comment.

src/utils/tvgStr.h
```diff
@@ -0,0 +1,31 @@
...
+ * SOFTWARE.
+ */
+
+#ifndef _TVG_STR_H_
+#define _TVG_STR_H_
+
+namespace tvg
+{
+    float strToFloat(const char *nPtr, char **endPtr);
+    char* strNDuplicate(const char *str, size_t n);
```
hermet, line 29: The last consideration by me is this.

I'm wondering if 'N' is really important in the naming. First of all, we would not have both strDuplicate() and strNDuplicate(). Secondly, strDuplicate(...) is a cleaner name. Third, strDuplicate(str, n) itself is enough to understand what it does.

src/utils/tvgStr.cpp
```diff
@@ -0,0 +1,220 @@
...
+ * src should be one of the following form :
+ *
+ * [whitespace] [sign] {digits [radix digits] | radix digits} [{e | E} [sign] digits]
+ * [whitespace] [sign] {INF | INFINITY}
+ * [whitespace] [sign] NAN [sequence]
+ *
+ * No hexadecimal form supported
+ * no sequence supported after NAN
+ */
+    float strToFloat(const char *nPtr, char **endPtr) {
```
hermet, line 58: The indentation of the documentation and the function header don't align. I'd rather remove the spacing of the function.

mgrudzinska: done

src/utils/tvgStr.h
mgrudzinska: done

hermet (approved):

PR #1570, opened 2023-08-01
webp_loader: libwebp set to be a required depend

mgrudzinska:
While building on windows (MinGW) no 'libwebp' was found, which resulted in build errors. The building process should have been stopped earlier for clarity.

### DIFF
--- a/src/loaders/external_webp/meson.build
+++ b/src/loaders/external_webp/meson.build
@@ -3,12 +3,10 @@ source_file = [
    'tvgWebpLoader.cpp',
 ]
 
-webp_dep = dependency('libwebp', required: false)
+webp_dep = dependency('libwebp', required: true)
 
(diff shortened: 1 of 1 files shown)


mgrudzinska:
@JSUYA I'd like to make a similar change for other loaders. am I missing some logic behind not requiring this dependency?

JSUYA (approved):
The reason why require is set to false in the external loader is to automatically use the static loader when the external library is not found.
Since webp doesn't have a static loader yet, I agree this change. (We can revert this patch again when the static loader is ready.)
But other loaders don't need requiring.

+ Actually, there was a problem of not finding libwebp in our CI's window environment. So user should not use `webp_beta` in windows build environment.
Maybe.. this PR may not be necessary if the window(MinGW) build you are talking about is local, not CI.

PR #1576, opened 2023-08-04
svg_loader: support other units for mask and clip 

mgrudzinska:
Added support for the objectBoundingBox value of
the clipPathUnits and for the userSpaceOnUse value
of the maskContentUnits.

@issue: https://github.com/thorvg/thorvg/issues/429

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -24,6 +24,7 @@
 #include <cstring>
 #include <string>
 #include "tvgCompressor.h"
+#include "tvgPaint.h"
 #include "tvgSvgLoaderCommon.h"
 #include "tvgSvgSceneBuilder.h"
 #include "tvgSvgPath.h"
@@ -220,6 +221,26 @@ static bool _appendClipChild(SvgNode* node, Shape* shape, const Box& vBox, const
 }
 
 
+static Matrix _compositionTransform(Paint* paint, const SvgNode* node, const SvgNode* compNode, SvgNodeType type)
+{
+    Matrix m = {1, 0, 0, 0, 1, 0, 0, 0, 1};
+    //The initial mask transformation ignored according to the SVG standard.
+    if (node->transform && type != SvgNodeType::Mask) {
+        m = *node->transform;
+    }
+    if (compNode->transform) {
+        m = mathMultiply(&m, compNode->transform);
+    }
+    if (!compNode->node.clip.userSpace) {
+        float x, y, w, h;
+        P(paint)->bounds(&x, &y, &w, &h, false, false);
+        Matrix mBBox = {w, 0, x, 0, h, y, 0, 0, 1};
+        m = mathMultiply(&m, &mBBox);
+    }
+    return m;
+}
+
+
 static void _applyComposition(Paint* paint, const SvgNode* node, const Box& vBox, const string& svgPath)
 {
     /* ClipPath */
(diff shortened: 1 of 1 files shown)


mgrudzinska:
replacing #1547

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -92,6 +110,51 @@ static Box _boundingBox(const Shape* shape)
...
+    for (auto child : children) {
+        childBBox = _boundingBox(child);
+        _mergeBoxes(box, childBBox);
+    }
+
+    return box;
+}
+
+
+static Box _boundingBox(const Paint* paint)
```
hermet, line 141: paint->bounds() will return the merged region of its internal elements. You are recomputing the bounds here. Is this necessary?

Also, It looks  "static Box _boundingBox(const Shape* shape)" can be removed?

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -292,9 +373,9 @@ static void _applyComposition(Paint* paint, const SvgNode* node, const Box& vBox
             node->style->mask.applying = true;
 
             bool isMaskWhite = true;
-            auto comp = _sceneBuildHelper(compNode, vBox, svgPath, true, 0, &isMaskWhite);
-            if (comp) {
-                if (node->transform) comp->transform(*node->transform);
+            if (auto comp = _sceneBuildHelper(compNode, vBox, svgPath, true, 0, &isMaskWhite)) {
+                Matrix finalTransform = _compositionTransform(paint, node, compNode, SvgNodeType::Mask);
+                if (!mathIdentity((const Matrix*)(&finalTransform))) comp->transform(finalTransform);
```
hermet, line 378: --mathIdentity
https://github.com/thorvg/thorvg/pull/1603

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -269,13 +351,12 @@ static void _applyComposition(Paint* paint, const SvgNode* node, const Box& vBox
...
             }
 
-            if (node->transform) {
-                auto m = comp->transform();
-                m = mathMultiply(node->transform, &m);
-                comp->transform(m);
-            }
+            if (valid) {
+                Matrix finalTransform = _compositionTransform(paint, node, compNode, SvgNodeType::ClipPath);
+                if (!mathIdentity((const Matrix*)(&finalTransform))) comp->transform(finalTransform);
```
hermet, line 356: --mathIdentity
https://github.com/thorvg/thorvg/pull/1603

hermet (commented):
Thanks, Please check comments.

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: I need to know the bounds without stroke. for a single shape I can remove it, but for a scene it's not possible, I need to traverse all elements. long time ago I wanted to change the bounds api accordingly, so I could use it here, but as I remember it wasn't approved.

hermet: hmm but your logic looks including the stroking width as well.

hermet: ah ok I mis-read the code.

hermet:
This might be helpful:  https://github.com/thorvg/thorvg/pull/1608
```
#include "tvgPaint.h"
...
paint->pImpl->bounds(x, y,w, h, transform, stroking = false);
```

mgrudzinska:
> This might be helpful: #1608
> 
> ```
> #include "tvgPaint.h"
> ...
> P(paint)->bounds(x, y,w, h, transform, stroking = false);
> ```


yes, I'll change this pr after #1608 is on the main

hermet (approved):
Thanks!

PR #1531, opened 2023-07-23
Try to get the GL backend working

RuiwenTang:
This is an other draft which I am trying to rewrite exists GL backend code and make it work as expected.

Changes under examples directory can be ignored. I change to use SDL2 because I can not create a OpenGL context with EFL on my macbook. The clang-format and some code format changes can also be ignored.

Because I want to use some common feature in modern Open GL/GLES, like vertex array object, uniform buffer object, so I changed all existing buffer and shader creation code. Also I thought no matter what GPU backend we choose in the future, they all need to convert polygons to triangles. So I put the tessellation algorithm outside the gl_engine directory.

Current state:

- [x]  Fill Path
- [x] Stroke Path
- [x] Dash Effect In Stroke
- [x]  Clip Path
- [x] Linear Gradient
- [x]  Radial Gradient
- [x]  Image
- [ ]  Blending
- [x]  Masking
- [ ]  Contour AA

PR #1584, opened 2023-08-07
utils: add sort method in tvgArray

RuiwenTang:
This PR does the following things:
* Add a new method in tvg::Array to support sort with custom comparer

The tessellation code in #1531 use tvg::Array as much as possible, and the algorithm need to sort Point , Edge.
So I add the sort method in tvg::Array. This is a new method so there will be no impact on existing code.

### DIFF
--- a/src/utils/tvgArray.h
+++ b/src/utils/tvgArray.h
@@ -24,6 +24,7 @@
 #define _TVG_ARRAY_H_
 
 #include <memory.h>
+#include <cstdint>
 
 namespace tvg
 {
@@ -110,6 +111,12 @@ struct Array
         return count == 0;
     }
 
+    template<class COMPARE>
+    void sort()
+    {
+        qsort<COMPARE>(data, 0, static_cast<int32_t>(count) - 1);
+    }
+
     void operator=(const Array& rhs)
     {
         reserve(rhs.count);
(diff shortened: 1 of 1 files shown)


src/utils/tvgArray.h
```diff
@@ -121,6 +128,34 @@ struct Array
...
+            T       tmp = arr[low];
+            while (i < j) {
+                while (i < j && !COMPARE{}(arr[j], tmp)) --j;
+                // while (i < j && arr[j] >= tmp) --j;
+                if (i < j) {
+                    arr[i] = arr[j];
+                    ++i;
+                }
+                while (i < j && COMPARE{}(arr[i], tmp)) ++i;
+                // while (i < j && arr[i] <= tmp) ++i;
```
hermet, line 148: does this comment necessary?

hermet (changes requested):
Please remove the test-case. We don't add it for the internal functions. Only extern APIs would have it.  Otherwise, Looks good to me. 

Thanks. :-)

RuiwenTang:
@hermet Please take a look again :)

hermet (approved):
Great, looks good to me. thanks!

hermet:
Ah one more, could you please revise the commit message?

RuiwenTang:
> Ah one more, could you please revise the commit message? Remove -> Add a associate unite test case to check the sort logical

Sorry about the commit meesage, I have removed them

Issue #1558, opened 2023-07-30
Support Focal property in Radial Gradient


hermet:
The Radial Gradient lacks the "focal" property, which is commonly utilized in vector formats. It's necessary to provide support for this property.

https://github.com/thorvg/thorvg/issues/1555

mgrudzinska:
@reptofrog
regarding [the example:](https://github.com/thorvg/thorvg/issues/1555#issue-1827560930) it would be great if you provide some samples svgs that you have in mind.
I tried to reconstruct your sample:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 200">
  <defs>
    <radialGradient id="radialGradient3f" r="50" cx="100" cy="100" 
      fr="0" fx="65" fy="65"
      gradientUnits="userSpaceOnUse">
      <stop style="stop-color:#ff00ff;stop-opacity:1;" offset="0"/>
      <stop style="stop-color:#00ff00;stop-opacity:1;" offset="1"/>
    </radialGradient>
  </defs>

  <circle cx="100" cy="100" r="50" fill="url(#radialGradient3f)"/>
</svg>
```
result:
<img width="398" alt="Zrzut ekranu 2023-08-7 o 21 22 54" src="https://github.com/thorvg/thorvg/assets/67589014/2232b214-08e1-46d1-a8a4-4cb31b74126d">

but on your screen the focal point is outside the circle, right? or does inkskape prevent such cases?

the focal support is reachable, but I'm no so sure about all of the corner cases...

mgrudzinska:
I found the answer to my previous comment - the behavior may be as I guessed if inkskape supports svg1.1, this was changed in svg2: https://www.w3.org/TR/SVG2/pservers.html#RadialGradientNotes

reptofrog:
@mgrudzinska 

Here is the example file with focal point set to be **outside of circle** and **outside of canvas**: https://github.com/thorvg/thorvg/assets/36744029/1c8cf513-2ff1-46c6-b6da-c4ea88b0d7c2

Here is one with focal point set to be **outside of circle** but **inside the canvas**: https://github.com/thorvg/thorvg/assets/36744029/7aa9172c-b2ef-40fd-b717-2cb61ec37d41

And finally, an example with focal point both **inside the circle** and **inside the canvas**: https://github.com/thorvg/thorvg/assets/36744029/86af735a-979f-4b0f-bba6-9e1cccc95524

---

[...]

PR #1585, opened 2023-08-07
sw_engine: support for radial grad focal attribs

mgrudzinska:
Till now only the end circle values of the radial grad were supported. Now also the start circle values are. These are the focal point and the radius of the start circle.
In the svg_loader support for the 'fr' attrib of the radial grad is added.

@Issue: https://github.com/thorvg/thorvg/issues/1558

sample:
```
  <defs>
    <radialGradient id="grad" r="50" cx="100" cy="100" 
      fr="21" fx="80" fy="80"
      gradientUnits="userSpaceOnUse">
      <stop style="stop-color:#ff00ff;stop-opacity:1;" offset="0"/>
      <stop style="stop-color:#0000ff;stop-opacity:1;" offset="0.5"/>
      <stop style="stop-color:#00ff00;stop-opacity:1;" offset="1"/>
    </radialGradient>
  </defs>

  <circle cx="100" cy="100" r="50" fill="url(#grad)"/>
</svg>
```

result:
Zrzut ekranu 2023-08-7 o 23 09 06

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -765,7 +765,7 @@ public:
      *
      * @return Result::Success when succeed, Result::InvalidArguments in case the @p r value is zero or less.
      */
-    Result radial(float cx, float cy, float r) noexcept;
+    TVG_DEPRECATED Result radial(float cx, float cy, float r) noexcept;
 
     /**
      * @brief Sets the radial gradient attributes.
@@ -784,6 +784,9 @@ public:
      *
      * @return Result::Success when succeed, Result::InvalidArguments in case the @p r or @p fr value is zero or less.
      *
+     * @note Should you choose not to specify the focal point, ensure that the values for @p fx and @p fy match
+     * those of @p cx and @p cy, and set @p fr to 0.
+     *
      * @BETA_API
      */
     Result radial(float cx, float cy, float r, float fx, float fy, float fr) noexcept;
@@ -799,7 +802,7 @@ public:
     *
     * @return Result::Success when succeed.
     */
-    Result radial(float* cx, float* cy, float* r) const noexcept;
+    TVG_DEPRECATED Result radial(float* cx, float* cy, float* r) const noexcept;
 
     /**
      * @brief Gets the radial gradient attributes.
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -757,28 +757,67 @@ public:
     /**
      * @brief Sets the radial gradient bounds.
      *
-     * The radial gradient bounds are defined as a circle centered in a given point (@p cx, @p cy) of a given radius.
+     * The radial gradient bounds are defined as a circle centered in a given point (@p cx, @p cy) of a given radius @p r.
      *
      * @param[in] cx The horizontal coordinate of the center of the bounding circle.
      * @param[in] cy The vertical coordinate of the center of the bounding circle.
-     * @param[in] radius The radius of the bounding circle.
+     * @param[in] r The radius of the bounding circle.
      *
-     * @return Result::Success when succeed, Result::InvalidArguments in case the @p radius value is zero or less.
+     * @return Result::Success when succeed, Result::InvalidArguments in case the @p r value is zero or less.
      */
-    Result radial(float cx, float cy, float radius) noexcept;
+    Result radial(float cx, float cy, float r) noexcept;
 
     /**
-     * @brief Gets the radial gradient bounds.
+     * @brief Sets the radial gradient attributes.
      *
-     * The radial gradient bounds are defined as a circle centered in a given point (@p cx, @p cy) of a given radius.
+     * The radial gradient is defined by the end circle with a center (@p cx, @p cy) and a radius @p r and
+     * the start circle with a center - focal point (@p fx, @p fy) and a radius @p fr.
+     * The gradient will be rendered in a way that aligns the 100% gradient stop with the edge of the end circle
+     * and 0% gradient stop with the edge of the start circle.
      *
-     * @param[out] cx The horizontal coordinate of the center of the bounding circle.
-     * @param[out] cy The vertical coordinate of the center of the bounding circle.
-     * @param[out] radius The radius of the bounding circle.
+     * @param[in] cx The horizontal coordinate of the center of the end circle.
+     * @param[in] cy The vertical coordinate of the center of the end circle.
+     * @param[in] r The radius of the end circle.
+     * @param[in] fx The horizontal coordinate of the center of the start circle.
+     * @param[in] fy The vertical coordinate of the center of the start circle.
+     * @param[in] fr The radius of the start circle.
+     *
+     * @return Result::Success when succeed, Result::InvalidArguments in case the @p r or @p fr value is zero or less.
+     *
+     * @BETA_API
+     */
+    Result radial(float cx, float cy, float r, float fx, float fy, float fr) noexcept;
+
+    /**
+    * @brief Gets the radial gradient bounds.
+    *
+    * The radial gradient bounds are defined as a circle centered in a given point (@p cx, @p cy) of a given radius.
+    *
+    * @param[out] cx The horizontal coordinate of the center of the bounding circle.
+    * @param[out] cy The vertical coordinate of the center of the bounding circle.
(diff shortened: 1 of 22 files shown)


inc/thorvg.h
```diff
@@ -757,28 +757,52 @@ class TVG_API RadialGradient final : public Fill
...
+     * @param[out] fx The horizontal coordinate of the center of the start circle.
+     * @param[out] fy The vertical coordinate of the center of the start circle.
+     * @param[out] fr The radius of the start circle.
      *
      * @return Result::Success when succeed.
+     *
+     * @see Result radial(float cx, float cy, float r, float fx, float fy, float fr) noexcept
      */
-    Result radial(float* cx, float* cy, float* radius) const noexcept;
+    Result radial(float* cx, float* cy, float* r, float* fx = nullptr, float* fy = nullptr, float* fr = nullptr) const noexcept;
```
mgrudzinska, line 805: @hermet I removed the api that we had till now and added this one with the default values - is it ok? it is not possible have both at the same time, unless I remove the default values.

mgrudzinska:
There are still some corner cases that need to be solved, that's why this is a draft

inc/thorvg.h
hermet: We can't guarantee our milestone 100% yet so I think DEPRECATION is better than removing now.

src/lib/sw_engine/tvgSwRaster.cpp
```diff
@@ -74,6 +74,24 @@ struct FillRadial
     }
 };
 
+struct FillRadialEdgeCase
```
mgrudzinska, line 77: @hermet I added this, because I wanted to check asap, which 'fillRadial' function to call. I didn't want to add any checks inside the fillRadial and I needed to distinguish between cases a == 0 and a > 0.

mgrudzinska:
sample for the 'edgeCases':
    <defs>
        <radialGradient id="grad" r="50" cx="100" cy="100"
                        fr="25" fx="100" fy="75"
                        spreadMethod="reflect" 
                        gradientUnits="userSpaceOnUse">
            <stop style="stop-color:#ff00ff;stop-opacity:1;" offset="0"/>
            <stop style="stop-color:#0000ff;stop-opacity:1;" offset="0.5"/>
            <stop style="stop-color:#00ff00;stop-opacity:1;" offset="1"/>
        </radialGradient>
    </defs>

    <circle cx="100" cy="100" r="50" fill="url(#grad)"/>
</svg>

src/lib/sw_engine/tvgSwFill.cpp
```diff
@@ -228,82 +240,162 @@ static inline uint32_t _pixel(const SwFill* fill, float pos)
...
+ * quadratic equation with the following coefficients (rx and ry defined in the _calculateCoefficients()):
+ * A = a  // fill->radial.a
+ * B = 2 * (dr * fr + rx * dx + ry * dy)
+ * C = fr^2 - rx^2 - ry^2
+ * Derivatives are computed with respect to dx.
+ * This procedure aims to optimize and eliminate the need to calculate all values from the beginning
+ * for consecutive x values with a constant y. The Taylor series expansions are computed as long as
+ * its terms are non-zero.
+ */
+static void _calculateCoefficients(const SwFill* fill, uint32_t x, uint32_t y, float& b, float& delta_b, float& det, float& delta_det, float& delta_delta_det, float* ddd = nullptr, float* delta_ddd = nullptr, float* delta_delta_ddd = nullptr)
```
hermet, line 253: please check naming convention:

aa_bb (x)
aaBB (o)
aabb (o)

src/lib/sw_engine/tvgSwRaster.cpp
hermet: I didn't expect the methods to be separated (RadialFill  vs RadialFillEdgeCase). I suppose the focal version could encompass the previous radial gradient behavior mathematically.

hermet (commented):
Please check comments.

inc/thorvg.h
```diff
@@ -757,28 +757,67 @@ class TVG_API RadialGradient final : public Fill
...
      * @param[in] cx The horizontal coordinate of the center of the bounding circle.
      * @param[in] cy The vertical coordinate of the center of the bounding circle.
-     * @param[in] radius The radius of the bounding circle.
+     * @param[in] r The radius of the bounding circle.
      *
-     * @return Result::Success when succeed, Result::InvalidArguments in case the @p radius value is zero or less.
+     * @return Result::Success when succeed, Result::InvalidArguments in case the @p r value is zero or less.
      */
-    Result radial(float cx, float cy, float radius) noexcept;
+    Result radial(float cx, float cy, float r) noexcept;
```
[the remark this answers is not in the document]
mgrudzinska: I thought that we want to have both setters, don't we?
otherwise the user have to always specify the values of fx, fy, fr. we can not use def values, since for fx and fy these are cx and cy.

inc/thorvg.h
```diff
@@ -757,28 +757,67 @@ class TVG_API RadialGradient final : public Fill
...
+    *
+    * The radial gradient bounds are defined as a circle centered in a given point (@p cx, @p cy) of a given radius.
+    *
+    * @param[out] cx The horizontal coordinate of the center of the bounding circle.
+    * @param[out] cy The vertical coordinate of the center of the bounding circle.
+    * @param[out] r The radius of the bounding circle.
+    *
+    * @return Result::Success when succeed.
+    */
+    Result radial(float* cx, float* cy, float* r) const noexcept;
```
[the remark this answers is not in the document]
mgrudzinska: I didn;t add this tag, because I thought that at some point we could have only one api:
```Result radial(float* cx, float* cy, float* r, float* fx = nullptr, float* fy = nullptr, float* fr = nullptr) noexcept;```
I thought that we can introduce it right now and remove the current one, but as I understand, you don;t think it's a good idea. 
so the plan is to deprecate this one, keep only the new one: ```Result radial(float* cx, float* cy, float* r, float* fx, float* fy, float* fr) const noexcept;``` and then at some point add the default values?

File: src/lib/sw_engine/tvgSwRaster.cpp (hunk 2 above)
Comment (mgrudzinska) [reply src/lib/sw_engine/tvgSwRaster.cpp]: @hermet please check the second commit - now the check ```if (fill->radial.a < FLT_EPSILON)``` is done every time fillRadial is called. my plan was to avoid it. 
do you prefer the solution from the second commit? (I'll squash them)

File: src/lib/sw_engine/tvgSwFill.cpp [hunk 6]
```diff
@@ -250,53 +251,27 @@ static inline uint32_t _pixel(const SwFill* fill, float pos)
  * for consecutive x values with a constant y. The Taylor series expansions are computed as long as
  * its terms are non-zero.
  */
-static void _calculateCoefficients(const SwFill* fill, uint32_t x, uint32_t y, float& b, float& delta_b, float& det, float& delta_det, float& delta_delta_det, float* ddd = nullptr, float* delta_ddd = nullptr, float* delta_delta_ddd = nullptr)
```
mgrudzinska: all *ddd args were here by mistake...

inc/thorvg.h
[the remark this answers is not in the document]
hermet: Thorvg maintains optimal number of APIs. I believe that the new API won't cause any troubles for users. we can address the guide in doc.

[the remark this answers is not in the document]
mgrudzinska: @hermet please check now. I've changed the apis as you suggested. I kept the beta_api tag, added a comment what values to use for the focal point.

mgrudzinska:
I'd like to squash the commits once we have the final version of this pr.

hermet (changes requested):
We don't allow api break now.
please make a separate patch if it does. (thorvg_capi.h ? ) we can apply only it later.

hermet:
@mgrudzinska if you need any patches that breaks apis, please make individual changes. I will cherry-pick to here. https://github.com/thorvg/thorvg/pull/1564

hermet:
Replace the major changes with https://github.com/thorvg/thorvg/pull/1631

PR #1527, opened 2023-07-17
Infra: allow access to pImpl when in TVG dev mode.

hermet:
some infra changes for upcoming lottie animation feature.

PR #1586, opened 2023-08-08
Clarify build/export definition.

hermet:
The meaning of these two flags is somewhat confusing.

* The TVG_EXPORT flag is used to build external APIs.
* The TVG_BUILD flag is used to build the internal thorvg codebase.

See: https://github.com/thorvg/thorvg/pull/1527

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -26,7 +26,7 @@
 
 #ifndef TVG_STATIC
     #ifdef _WIN32
-        #if TVG_BUILD
+        #if TVG_EXPORT
             #define TVG_API __declspec(dllexport)
         #else
             #define TVG_API __declspec(dllimport)
--- a/src/bindings/capi/thorvg_capi.h
+++ b/src/bindings/capi/thorvg_capi.h
@@ -27,7 +27,7 @@
 
 #ifndef TVG_STATIC
     #ifdef _WIN32
-        #if TVG_BUILD
+        #if TVG_EXPORT
             #define TVG_API __declspec(dllexport)
         #else
             #define TVG_API __declspec(dllimport)
--- a/src/meson.build
+++ b/src/meson.build
@@ -1,10 +1,10 @@
-compiler_flags = []
+compiler_flags = ['-DTVG_BUILD']
 override_options = []
 
 lib_type = get_option('default_library')
 
 if (lib_type == 'shared')
-    compiler_flags += ['-DTVG_EXPORT', '-DTVG_BUILD']
+    compiler_flags += ['-DTVG_EXPORT']
 else
     compiler_flags += ['-DTVG_STATIC']
 endif


capnm:
To be 100% sure that I've got it right:

For the external `*.obj` linking I have to make sure that the TVG sources are built with the compiler flags `-DTVG_BUILD` and `-DTVG_STATIC`.

hermet:
> Also to be 100% sure I have understood it correctly: for the external `*.obj` linking I have to make now sure that the TVG sources are built with the compiler flags `-DTVG_BUILD` and `-DTVG_STATIC` ??

@capnm Hmm... Godot is supposed to use the -DTVG_STATIC flag to enable static linking. Otherwise, let's forget about this patch for now. we don't need to worry about this for now. please just let me know then I will close this.

capnm:
@hermet OK. But should keep  (for future improvements) the `-DTVG_BUILD` also there or remove it?

hermet:
> @hermet OK. But should I keep (for future improvements) the `-DTVG_BUILD` also there or remove it?

@capnm  Yes, Godot currently builds tvg itself, so it would be better to keep it. I believe you don't need to change Godot's build infrastructure, as it works as it did before.

vtorri:
TVG_EXPORT is useless. See for example:

https://github.com/davea42/libdwarf-code/blob/main/src/lib/libdwarf/libdwarf.h#L48

capnm:
> > @hermet OK. But should I keep (for future improvements) the `-DTVG_BUILD` also there or remove it?
> 
> @capnm Yes, Godot currently builds tvg itself, so it would be better to keep it. I believe you don't need to change Godot's build infrastructure, as it works as it did before.

There is a policy to avoid changes to the source code of external libraries as much as possible. The own build system (uses static object linking to eliminate user troubles with dynamic libraries) can be further improved to follow the library changes :-)

Also for `v0.10.0` I currently removed  the `-DTVG_BUILD` flag, disabled and removed loaders/saves that are in a application-context for SVG useless (e.g. there is no way to support [embedded raster images](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/image)):
[update-thorvg.sh](https://github.com/godotengine/godot/pull/80095/files#diff-978bba2de6e2b21e72dacd96162d9bdb53bf424feb16c65e1d995c48fd821666)

If you have any ideas for improvement, please let me know! 

I'm investigating if it's feasible to create external plugins for the recent more curve-oriented implementation of [blender.org](https://code.blender.org/2023/05/the-next-big-step-grease-pencil-3-0/) Grease Pencil animations -> Lottie -> Godot. WDYT?

vtorri:
also TVG_BUILD and TVG_EXPORT are useless in non Windows builds

hermet:
@capnm Confused. I'm not sure what's going on with Godot. Godot should not be troubled in 0.10 since nothing has changed about the build flags after 0.9 and this wouldn't be applied. Also, at least Godot should use the SVG loader (or Lottie as well, removing the saver is fine.) The enabled loaders will be statically complied with thorvg binary.

Regarding grease pencil animation, I'm not sure in detail at the moment, but it seems like Lottie animation features are quite aligned with many grease pencil animation features in terms of vector scopes. Most advanced features, such as 3D transformations, filter effects, and skeletal animation, would depend on how the plugin (exporter) converts the original source data from Blender to Lottie. It seems like it would be a big and interesting task. However, some advanced features, such as full-3D complex-meshed models, might not be appropriate for Lottie. Lottie animations are designed for simpler cut scenes and animations.

capnm:
> @capnm Confused. I'm not sure what's going on with Godot. Godot should not be troubled in 0.10 since nothing has changed about the build flags after 0.9 and this wouldn't be applied. Also, at least Godot should use the SVG loader (or Lottie as well, removing the saver is fine.) The enabled loaders will be statically complied with thorvg binary.

In 0.9 we had the issue with static windows linking, which I can now solve simply by adding the `-DTVG_STATIC` flag in SCons.

If you take a look at the [Godot PR](https://github.com/godotengine/godot/pull/80095) comments, it is important for maintainers to reduce the in the repository copied code and Godot binary size. So I need to copy and activate only parts that are currently implemented for SVG images:
```
#define THORVG_SW_RASTER_SUPPORT
#define THORVG_SVG_LOADER_SUPPORT
```
All pixel-based images are processed and rasterized internally, also the other ThorVG loaders are not in use.

> Regarding grease pencil animation, I'm not sure in detail at the moment, but it seems like Lottie animation features are quite aligned with many grease pencil animation features in terms of vector scopes. Most advanced features, such as 3D transformations, filter effects, and skeletal animation, would depend on how the plugin (exporter) converts the original source data from Blender to Lottie. It seems like it would be a big and interesting task. However, some advanced features, such as full-3D complex-meshed models, might not be appropriate for Lottie. Lottie animations are designed for simpler cut scenes and animations.

Thanks a lot for your comments! Yes, the glTF exporter plugin and  format that my students currently use for texture-based 2D/3D animations in Godot must also use only part of Blender's internal capabilities, and as you say, I think Lottie could be a good open format to transport simple vector-based 2D animations from Blender to Godot. I should definitely take the time and read Lottie specs :)

PR #1587, opened 2023-08-08
common: add new method to get mat4x4 directly from mat3x3

RuiwenTang:
This PR add a new method to get 4x4 matrix directly from exsited 3x3 matrix in column major order.

This is also a basic math calculations required by #1531

### DIFF
--- a/src/lib/gl_engine/tvgGlGeometry.h
+++ b/src/lib/gl_engine/tvgGlGeometry.h
@@ -60,6 +60,37 @@
         } \
     }
 
+/**
+ *  mat3x3               mat4x4
+ *
+ * [ e11 e12 e13 ]     [ e11 e12 0 e13 ]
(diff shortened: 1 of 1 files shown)


hermet (commented):
Thanks!

I think this helper is not commonly used for now. 
If it's necessary, you can add this helper function in gl_raster side.

hermet:
> f we only plan to support OpenGL/GLES or WebGL for now, this method can be placed inside gl raster.

@RuiwenTang GPU is just a compuation power unit, it doesn't fix the matrix size, but it's up to the implemenation approach. When it's really necessary, we can move the function to the common. :-)

RuiwenTang:
Ok, I move the code into tvgGlGeometry as macro

hermet (approved):

PR #1589, opened 2023-08-09
gl_engine: upgrade min gl version

RuiwenTang:
This PR changes the min gl version used inside GL Engine

Since most device supported morden gl (GLES 3.0 for mobile GPU, GL 3.3 for desktop GPU)
 
And I plan to use some advance and convenient API like [VAO](https://www.khronos.org/opengl/wiki/Vertex_Specification#Vertex_Array_Object) and [UBO](https://www.khronos.org/opengl/wiki/Uniform_Buffer_Object)

So I changed the example code to create a GLES3.x context with EFL. And the header file included inside GL Engine.
And the rest code in #1531 rely on this.

### DIFF
--- a/src/lib/gl_engine/meson.build
+++ b/src/lib/gl_engine/meson.build
@@ -19,9 +19,9 @@ source_file = [
    'tvgGlShaderSrc.cpp',
 ]
 
-egl_dep = meson.get_compiler('cpp').find_library('EGL')
 gles_dep = meson.get_compiler('cpp').find_library('GLESv2')
-external_dep = [egl_dep, gles_dep]
+
+external_dep = [gles_dep]
 
 engine_dep += [declare_dependency(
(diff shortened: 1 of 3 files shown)


hermet (approved):
good to go. thanks.

src/lib/gl_engine/meson.build
```diff
@@ -19,9 +19,9 @@ source_file = [
    'tvgGlShaderSrc.cpp',
 ]
 
-egl_dep = meson.get_compiler('cpp').find_library('EGL')
 gles_dep = meson.get_compiler('cpp').find_library('GLESv2')
```
dkaraush, line 22: Should `gles_dep` also be updated to `GLESv3`? Tried building on Android, got error `undefined reference to `glGenVertexArrays'`, upgrading library here fixed a problem.

PR #1592, opened 2023-08-10
gl_engine: use vertex array and index buffer when render primitive

RuiwenTang:
This PR use [Vertex Array Object](https://www.khronos.org/opengl/wiki/Vertex_Specification#Vertex_Array_Object) and `Element Array Buffer`  in GlGeometry 

VAO is new object in modern GL API, which can split and manage vertex buffer and attribute state. This is optional in GLES and is mandatory in GL.

Also index buffer is faster then pass cpu data in GL draw call, and must be used with a valid `Vertex Array Object`


hermet (approved):
thanks!

PR #1594, opened 2023-08-11
loader: avoid attempting forceful loading when the given MIME type is…

hermet:
… unknown.

Loading images when the MIME type is unknown is unnecessary and often results in failure, adding unnecessary processing overhead.

Pictures should have a clear MIME type assigned by users.


hermet:
close due to the compatibility break.

hermet:
replaced with https://github.com/thorvg/thorvg/pull/1595

PR #1601, opened 2023-08-15
common: continue searching for a loader

mgrudzinska:
Previously, if an incorrect mime type was provided, no other loaders were sought. Now fixed.


hermet (changes requested):
The incorrect mimetype is the user's fault, which leads to inefficient results. The user must take notice of this and fix it.

mgrudzinska:
ok, now I see that you've changed this lately :) 
the log msg has to be updated, it was misleading

hermet (approved):

PR #1604, opened 2023-08-17
gl_engine: Fix mistake in glbuffer id check

RuiwenTang:
`GL_INVALID_VALUE` is generate during `glGenBuffers` and is checked by calling `glGetError()` inside `GL_CHECK` macro.

While the buffer id is valid if the value is not zero.

### DIFF
--- a/src/lib/gl_engine/tvgGlGpuBuffer.cpp
+++ b/src/lib/gl_engine/tvgGlGpuBuffer.cpp
@@ -29,7 +29,7 @@
 GlGpuBuffer::GlGpuBuffer()
 {
     GL_CHECK(glGenBuffers(1, &mGlBufferId));
-    assert(mGlBufferId != GL_INVALID_VALUE);
+    assert(mGlBufferId != 0);
 }
 
 


src/lib/gl_engine/tvgGlGpuBuffer.cpp
```diff
@@ -29,7 +29,7 @@
 GlGpuBuffer::GlGpuBuffer()
 {
     GL_CHECK(glGenBuffers(1, &mGlBufferId));
-    assert(mGlBufferId != GL_INVALID_VALUE);
```
hermet, line 32: doesn't GL_INVALID_VALUE zero? it is more obvious to use.

RuiwenTang: `GL_INVALID_VALUE ` is a [OpenGL error code](https://www.khronos.org/opengl/wiki/OpenGL_Error),  and the value is 0x501.
The buffer id may reach this value if glGenBuffers is called multiple times, but it is a valid buffer id

hermet: gotcha! thanks.

hermet (approved):

PR #1608, opened 2023-08-18
renderer/paint: added an option for stroking when acquiring bounding …

hermet:
…box size.

The 'Paint' function now includes an internal stroking option, allowing the bounding box to be calculated with or without the stroke size taken into account.

This feature is particularly useful for SVG loaders, and we might consider making the interface available for user demands in the future.


mgrudzinska (approved):
yes, this is what I needed in the svg loader

PR #1616, opened 2023-08-23
Hermet/stroke

hermet:


hermet:
@mgrudzinska  the offset logic applied, this only reverts the api breaks. I expected the deprecate the previous ones and beta apis for the new ones.

You can submit additional changes for the stroke-offset apis like this pending change:
https://github.com/thorvg/thorvg/pull/1564

We will apply them when 1.0 is get ready.

p.s offset feature is buggy im looking into that.

PR #1621, opened 2023-08-28
common/array: enhance the interface

hermet:
easy access to a specific data with a operator.

### DIFF
--- a/src/common/tvgArray.h
+++ b/src/common/tvgArray.h
@@ -74,9 +74,34 @@ struct Array
         return reserve(count + size);
     }
 
-    T* end() const
+    const T& operator[](size_t idx) const
+    {
+        return data[idx];
+    }
+
+    T& operator[](size_t idx)
+    {
+        return data[idx];
+    }
+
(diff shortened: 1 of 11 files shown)


src/common/tvgArray.h
```diff
@@ -69,6 +69,11 @@ struct Array
         return true;
     }
 
+    T& operator()(size_t idx) const
```
RuiwenTang, line 72: should we  use square brackets operator ` operator[](size_t) `?

And one more thing.
I prefer use const function to return const reference, and use non const function to return mutable reference. Just like the code in STL container.
for example:
```
const T& operator()[size_t idx] const { return data[idx]; }
T& operator()[size_t idx] { return data[idx]; }
```
Because this is meanfull and can make sure if pass the const reference array into some function, the array contents  can not be changed.

hermet: code has been updated. hopefully you are satisfiled with that. :)

RuiwenTang (approved):

Issue #1555, opened 2023-07-29
Radial gradient's focal point is broken / not used


reptofrog:
It appears that ThorVG is not using the radial gradient's focal point.

| Inkscape  | ThorVG Viewer |
| ------------- | ------------- |
| ![Screenshot from 2023-07-29 20-54-42](https://github.com/thorvg/thorvg/assets/36744029/d8a81446-319b-46de-856c-3cd19f42aaee)|![image](https://github.com/thorvg/thorvg/assets/36744029/c6b6d6aa-f8d1-47a6-a353-273d106da139)|
|| Sometimes the gradient **is not rendered at all** when the focal point is shifted beyond the canvas, but I could not determine under what conditions that happens. |
||**UPD: I think ThorVG always renders the radial gradient no matter what, just in the way that ignores the focal point if it's outside the boundaries.**|

hermet:
Need to enhance the thorvg canvas radial gradient feature to support the focal property.

reptofrog:
@hermet 

It appears that the rendering is broken at the moment.

I couldn't figure out how to test it locally, so I tested it through the online viewer.

[Screencast from 2023-09-15 14-13-50.webm](https://github.com/thorvg/thorvg/assets/36744029/7f698b4a-c839-4800-95ce-6c2cceb4032d)

hermet:
@reptofrog Could you please attach the resource here? I will check it. Thanks.

reptofrog:
@hermet 

https://github.com/thorvg/thorvg/assets/36744029/8031e961-0f79-428d-8159-1fdf8b62dfd8

hermet:
@reptofrog hopefully you can checked the fixed result in v0.10.7

![image](https://github.com/thorvg/thorvg/assets/3711518/73008a9a-436d-4087-a76f-413b192affaf)

hermet:
a regression by:
```
commit 3426531dd217fdb679574e1ae4559fc4f8b6b17e (HEAD)
Author: Mira Grudzinska <mira@lottiefiles.com>
Date:   Wed Jun 18 12:13:39 2025 +0200

    sw_engine: fix focal point clamping

    This approach should be introduced in d8ebd8b4f517a013479bfbebbcdd3523ca3173ee.
    Using FLOAT_EPSILON is not sufficient due to numerical precision,
    since it's later inverted, and its very small value leads to
    significant errors during inversion.
```

PR #1659, opened 2023-09-18
sw_engine fill: fixed radial fill focal issue.

hermet:
This might be a workaround to correct the issue.
The threshold value comes from the mathematical inaccuracy.

@Issue:https://github.com/thorvg/thorvg/issues/1555


hermet:
@mgrudzinska hello, please have a look and leave any comments if you have a better solution. thanks!

Issue #121, opened 2020-10-27
Join lines attribute misbehaviour


mgrudzinska:
Drawing dashed strokes for various join lines attributes erratically (on corners).

![dash_cap_join](https://user-images.githubusercontent.com/67589014/97369175-a95fc700-18ac-11eb-954b-40bd297f9f62.PNG)

hermet:
![스크린샷 2023-10-05 010343](https://github.com/thorvg/thorvg/assets/3711518/4fd40191-fdaa-449a-bccb-3fa8c180f971)

PR #1685, opened 2023-10-04
sw_engine/stroke: Enhance the quality of the dash line corners.

hermet:
Previously, the engine didn't properly cover the dash line corner styles because it considered a new line to start at the corner.

This update modifies the logic to recognize curved lines as a single line, including the corners.

There may still be some quality issues,
but it's an improvement over the previous version.

@Issue: https://github.com/thorvg/thorvg/issues/121

스크린샷 2023-10-05 010343

### DIFF
--- a/src/renderer/sw_engine/tvgSwCommon.h
+++ b/src/renderer/sw_engine/tvgSwCommon.h
@@ -205,6 +205,7 @@ struct SwDashStroke
     float* pattern = nullptr;
     uint32_t cnt = 0;
     bool curOpGap = false;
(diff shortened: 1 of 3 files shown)


capnm:
I did a quick test of the PR while working on other Godot issues:

image

I have noticed that the **stroke-dashoffset** isn't calculated correctly:

thi_121-join-lines-attribute-misbehaviour_opt

hermet:
@capnm thanks for info. let's revisit the dash-offset problem with a separate issue ticket.

PR #1695, opened 2023-10-11
gl_engine: use glScissor to support fast clip

RuiwenTang:
This PR add scissor box info in GLRenderTask, and enable `GL_SCISSOR_TEST` during rendering to support the fast path clip (Rect Clip)

The reason we choose glScissor instead of glViewport is that if drawCall is outside the viewport it is still executed in the gpu but can be discared by GL_SCISSOR_TEST

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -63,6 +63,11 @@ bool GlRenderer::target(TVG_UNUSED uint32_t* buffer, uint32_t stride, uint32_t w
     surface.w = w;
     surface.h = h;
 
+    mViewport.x = 0;
+    mViewport.y = 0;
+    mViewport.w = surface.w;
+    mViewport.h = surface.h;
+
     return true;
 }
 
@@ -74,6 +79,7 @@ bool GlRenderer::sync()
     // Blend function for straight alpha
     GL_CHECK(glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_ONE, GL_ONE_MINUS_SRC_ALPHA));
     GL_CHECK(glEnable(GL_BLEND));
+    GL_CHECK(glEnable(GL_SCISSOR_TEST));
 
     mGpuBuffer->bind();
 
@@ -84,6 +90,8 @@ bool GlRenderer::sync()
     mGpuBuffer->unbind();
 
     mRenderTasks.clear();
+
+    GL_CHECK(glDisable(GL_SCISSOR_TEST));
 
     return true;
 }
@@ -204,8 +212,6 @@ bool GlRenderer::renderShape(RenderData data)
     uint8_t r = 0, g = 0, b = 0, a = 0;
     size_t flags = static_cast<size_t>(sdata->updateFlag);
 
-    GL_CHECK(glViewport(0, 0, (GLsizei)sdata->viewWd, (GLsizei)sdata->viewHt));
-
     if (flags & (RenderUpdateFlag::Gradient | RenderUpdateFlag::Transform))
     {
         auto gradient = sdata->rshape->fill;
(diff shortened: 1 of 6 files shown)


SergeyLebedkin (approved):

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -63,6 +64,18 @@ bool GlRenderer::target(TVG_UNUSED uint32_t* buffer, uint32_t stride, uint32_t w
     surface.w = w;
     surface.h = h;
 
+    // most platform the stride is just equal with width
+    // but when running on MacOS or iOS with retina display
+    // stride is not equal with width and usually is larger than logical width
+    // and viewport and scissor in OpenGL is based on physical pixel size
+    // so we need to calculate the density
+    mDensity = static_cast<float>(surface.stride) / (surface.w * BPP);
```
hermet, line 72: This looks weird to me. "density" concept shouldn't be involved in this calculation. You should only use the stride when determining the y offset from the surface.

In essence, the surface can be larger than the actually required surface, as existing surfaces can be reused by the graphics system. Experimentally, graphics system do that. It may allocate a new surface buffer exactly same user requested or reuse an existing one (in this case it can be larger than requuested one) for the sake of caching effects. That's why tvg engine(or some other systems) expects both stride/width value to distinguish it.

hermet (changes requested):
Could you please check my comment?

src/renderer/gl_engine/tvgGlRenderer.cpp
RuiwenTang: My statement may not be accurat.

The problem I want to solve should only exist on iOS and MacOS.
If using glfw on MacOS like this:
```
GLFWwindow* window = glfwCreateWindow(800,800, title, nullptr, nullptr);

int32_t pp_width, pp_height;
glfwGetFramebufferSize(window_, &pp_width, &pp_height);
```
pp_width and pp_height may be 1600 or 2400 depending on the actual screen resolution
 the x, y, width, height pass to glScissor need to be calculated based on pp_width/pp_height value

hermet: Thanks, let's say mDensity like you addressed.
If so, do we adjust the drawing coordinates for triangles with the density as well?

hermet: This figure is what I am assuming the scenario:
image

RuiwenTang: No, we can use the logical coordinates on the CPU side to calculate the position of the triangle, which will then be normalized in the vertex shader by multiplying by the projection matrix (computed in `GlGeometry::updateTransform`). OpenGL rasterizer can handle physical coordinate mapping

hermet: I think the correct approach is the given width/height should be 1600x1600 from the begining.

RuiwenTang: This may be a little complicated.
If pass pp_width/pp_height to canvas like
```
glCanvas->target(nullptr, 1600 * BPP, 1600, 1600);
```
this means the logical size of this canvas is 1600 x 1600
```
shape->appendRect(0, 0, 100, 100);

canvas->push(shape)
```
the shape rendered in this canvas will look smaller than 
```
glCanvas->target(nullptr,800 * BPP, 800, 800);
```
**The physical size is only effect the glSicssor calculation, But the logical size can effect the rendering result**

I'm not sure if I described the problem clearly

hermet: this is the scenario why we applied the stride:
image

RuiwenTang: > this is the scenario why we applied the stride: image

I know using stride looks weird, maybe we can add another parameters or api to pass the information to gpu renderer ?

hermet: @RuiwenTang As far as I know, mac uses super-sampling for the retina display, so the framebuffer is double-sized. This is bad and device-specific ompatibility problem. Logically, the provided width and height values are incorrect from the beginning. They should match the framebuffer size. 

I wonder this call resolves this problem.

```
glfwWindowHint(GLFW_COCOA_RETINA_FRAMEBUFFER, GL_FALSE)
```

RuiwenTang: Ok, I will delete the density calculation.
If we have plans to support iOS in the future. I prefer to use Metal directly to avoid this problem.

hermet: thanks. let's revisit this when the thing comes more clear.

hermet (approved):
thanks. let's revisit this when the thing comes more clear.

PR #1690, opened 2023-10-06
[draft]extension/window: Introduce tvg::Window based on GLFW

JSUYA:
Add an extension that uses GLFW-based tvg::Window.
This window has a canvas built into it.
The update() function receives a function that has tvg::Canvas as a parameter.
Users can build paint into the scene.
And call tvg::Window::loop().

Please refer to example/Window.cpp.

[APIs]
tvg::Window::gen(int width, int height, std::string name, tvg::CanvasEngine engine)
void close();
void resize(int width, int height);
void init(std::function<bool(tvg::Canvas*)> on_update);
void update(std::function<bool(tvg::Canvas*)> on_update);
static bool loop();

This idea was inspired by these:
https://github.com/DeriveSDK/experiments/tree/main/glfw_imgui_tvg_window
https://github.com/thorvg/thorvg/issues/1244#issuecomment-1666604901

PR #1702, opened 2023-10-13
renderer/initializer: Support for initializing SW and GL engines together

JSUYA:
Initialize both renderers to allow `init(SW | GL);`

refer to https://github.com/thorvg/thorvg/pull/1690#issuecomment-1759634391

### DIFF
--- a/src/renderer/tvgInitializer.cpp
+++ b/src/renderer/tvgInitializer.cpp
@@ -95,24 +95,27 @@ static bool _buildVersionInfo()
 Result Initializer::init(CanvasEngine engine, uint32_t threads) noexcept
 {
     auto nonSupport = true;
+    if (static_cast<int>(engine) == 0) return Result::InvalidArguments;
 
     if (engine & CanvasEngine::Sw) {
         #ifdef THORVG_SW_RASTER_SUPPORT
             if (!SwRenderer::init(threads)) return Result::FailedAllocation;
             nonSupport = false;
         #endif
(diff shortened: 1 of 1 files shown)


src/renderer/tvgInitializer.cpp
```diff
@@ -91,19 +91,22 @@ static bool _buildVersionInfo()
 Result Initializer::init(CanvasEngine engine, uint32_t threads) noexcept
 {
     auto nonSupport = true;
+    if (!(engine & CanvasEngine::Sw) && !(engine & CanvasEngine::Gl)) {
```
hermet, line 94: if ((int) engine == 0) return InvalidArguments;

src/renderer/tvgInitializer.cpp
```diff
@@ -125,19 +128,22 @@ Result Initializer::term(CanvasEngine engine) noexcept
     if (_initCnt == 0) return Result::InsufficientCondition;
 
     auto nonSupport = true;
+    if (!(engine & CanvasEngine::Sw) && !(engine & CanvasEngine::Gl)) {
```
hermet, line 131: if ((int) engine == 0) return InvalidArguments;

hermet (commented):

src/renderer/tvgInitializer.cpp
JSUYA: I updated it.

JSUYA:
Rebased patch

hermet (approved):

PR #1708, opened 2023-10-17
Readme: Fix lottie sample code

JSUYA:
tvg::cast(picture) -> tvg::cast<tvg::Picture>(picture)

This change should also be applied on our tutorial site.(https://www.thorvg.org/tutorial)


hermet:
Thanks for notifying this. We can replace the usage with https://github.com/thorvg/thorvg/pull/1711

Issue #1710, opened 2023-10-17
Missing blend functions  in capi bindings


onecoolx:
There is no functions to set  BlendMethod  in capi bindings

PR #1713, opened 2023-10-17
capi: introduce missing blend functions in capi bindings

SergeyLebedkin:
[issues 1710: CAPI](https://github.com/thorvg/thorvg/issues/1710) Missing blend functions in capi bindings

Introduced setter and getter for composite methods of the paint object:
`TVG_API Tvg_Result tvg_paint_set_composite_method(Tvg_Paint* paint, Tvg_Paint* target, Tvg_Composite_Method method);`
 `TVG_API Tvg_Result tvg_paint_get_composite_method(const Tvg_Paint* paint, const Tvg_Paint** target, Tvg_Composite_Method* method);`


hermet (approved):
thanks!

Issue #1712, opened 2023-10-17
Support gif saver


hermet:
- src/savers/gif: Add a saver module to export the scenes by converting and exporting them to a GIF file.
- Additionally, we can include another tool: lottie2gif.

hermet:
@JSUYA  I'm working on my side, based on your progress.

JSUYA:
> @JSUYA I'm working on my side, based on your progress.

I will update PR as soon as possible. I am sorry for being late.

PR #1715, opened 2023-10-18
saver GifSaver: Introduce GifSaver for animation

JSUYA:
Add save() API that takes tvg::Animation as a parameter.
This API uses gif.h to create each animation frame as a gif frame.
Gif creation do not support threads because they must be added sequentially.
Please see example/GifSaver.cpp

ex)
```cpp
auto animation = tvg::Animation::gen();
picture->load(EXAMPLE_DIR"/walker.json");
auto saver = tvg::Saver::gen();
saver->save(std::move(animation), EXAMPLE_DIR"/test.gif");
saver->sync();
```

result : test.gif
test

issue: https://github.com/thorvg/thorvg/issues/1712


inc/thorvg.h
```diff
@@ -1821,6 +1821,23 @@ class TVG_API Saver final
...
+     *
+     * @retval Result::Success When succeed.
+     * @retval Result::InsufficientCondition If currently saving other resources.
+     * @retval Result::NonSupport When trying to save a file with an unknown extension or in an unsupported format.
+     * @retval Result::MemoryCorruption An internal error.
+     * @retval Result::Unknown In case an empty paint is to be saved.
+     *
+     * @BETA_API
+     */
+    Result save(std::unique_ptr<Animation> animation, const std::string& path, int frameDelay = 2) noexcept;
```
hermet, line 1839: The inclusion of 'frameDelay' as a parameter seems unnecessary since 'Animation' already contains frame-related information, such as the number of frames and their duration. 'frameDelay' becomes relevant only when dealing with multiple raw scene frames that lack animation timing information. Not here.

src/renderer/tvgSaver.cpp
```diff
@@ -124,6 +139,32 @@ Result Saver::save(std::unique_ptr<Paint> paint, const string& path, bool compre
 }
 
 
+Result Saver::save(std::unique_ptr<Animation> animation, const string& path, int frameDelay) noexcept
```
hermet, line 142: It appears to be duplicated code with the 'ResultSaver::save(std::unique_ptr<Paint> paint, const string& path, bool compress)' method. We can consolidate this method and reuse it.

src/savers/tvg/tvgTvgSaver.cpp
```diff
@@ -820,3 +820,11 @@ bool TvgSaver::save(Paint* paint, const string& path, bool compress)
 
     return true;
 }
+
+
+bool TvgSaver::save(Animation* animation, const string& path, int frameDelay)
+{
+    TVGLOG("TVG_SAVER", "tvg:Animation to tvg saver is not supported.");
```
hermet, line 827: "tvg:Animation to tvg saver is not supported." -> "Animation is not supported"

src/savers/gif/tvgGifSaver.cpp
```diff
@@ -0,0 +1,122 @@
...
+        free(buffer);
+        buffer = nullptr;
+    }
+    return true;
+}
+
+
+bool GifSaver::save(Paint* paint, const string& path, bool compress)
+{
+    TVGLOG("GIF_SAVER", "tvg:Paint to Gif saver is not supported yet.");
```
hermet, line 93: tvg:Paint to Gif saver is not supported yet. -> Paint is not supported.

src/savers/gif/tvgGifSaver.cpp
```diff
@@ -0,0 +1,122 @@
...
+    size_t frameCount = animation->totalFrame();
+    for (size_t frame = 0; frame < frameCount; frame++) {
+        animation->frame(frame);
+        canvas->update(animation->picture());
+
+        if (canvas->draw() == tvg::Result::Success) { 
+            canvas->sync();
+        }
+
+        GifWriteFrame(&handle, reinterpret_cast<uint8_t*>(buffer), width, height, frameDelay);
```
hermet, line 63: I think here it needs to compute a proper frameDelay.

hermet (changes requested):
thanks, please check comments.

inc/thorvg.h
JSUYA: You're right. The delay of gif.h is not ms, it is 1/100th of a second. I will modify it.

src/renderer/tvgSaver.cpp
JSUYA: > We can consolidate this method and reuse it.
I have a question.
1) Is there a way to access the tvg::Animation object from the tvg::Picture(or Impl ?) object?
2) Do tvg::Animation and tvg::Paint need to inherit a new class? (ex tvg::Object ?)
3) compress parameter does not seem to be a valid option in tvg::Animation. Should I just leave it as an unsupported API?

hermet: 1. Not currently.
2. Probably, we can consider it when it proves genuinely effective for thorvg.
3. Yes, a compression parameter can be included to apply it optionally.

Never mind my previous feedback. I misunderstood the logic you implemented; the animation and picture couldn't be shared among the methods.

hermet:
@JSUYA Please check the API that I'd like to suggest.
`   
 /**
     * @brief Export the provided animation data to the specified file path.
     *
     * This function exports the given animation data to the provided file path. You can also specify the desired frame rate in frames per second (FPS) by providing the fps parameter.
     *
     * @param[in] animation The animation to be saved, including all associated properties.
     * @param[in] path The path to the file where the animation will be saved.
     * @param[in] quality The encoded quality level. @c 0 is the minimum, @c 100 is the maximum value(recommended).
     * @param[in] fps The desired frames per second (FPS). For example, to encode data at 60 FPS, pass 60. Pass 0 to keep the original frame data.
     *
     * @return Result::Success if the export succeeds.
     * @return Result::InsufficientCondition if there are ongoing resource-saving operations.
     * @return Result::NonSupport if an attempt is made to save the file with an unknown extension or in an unsupported format.
     * @return Result::MemoryCorruption in case of an internal error.
     * @return Result::Unknown if attempting to save an empty paint.
     *
     * @note A higher frames per second (FPS) would result in a larger file size. It is recommended to use the default value.
     * @note Saving can be asynchronous if the assigned thread number is greater than zero. To guarantee the saving is done, call sync() afterwards.
     *
     * @see Saver::sync()
     *
     * @BETA_API
     */
    Result save(std::unique_ptr<Animation> animation, const std::string& path, uint32_t quality = 100, uint32_t fps = 0) noexcept;
`

hermet:
@JSUYA quality @p is a reserved parameter for encoding compression metholody like jpg/png etc. In gif it wouldn't be used.

src/savers/gif/tvgGifSaver.cpp
```diff
@@ -0,0 +1,129 @@
...
+bool GifSaver::makeFrame(Animation* animation, const string& path, uint32_t width, uint32_t height, uint32_t fps)
+{
+    std::unique_ptr<tvg::SwCanvas> canvas = nullptr;
+    canvas = tvg::SwCanvas::gen();
+    if (canvas == nullptr) { 
+        return false;
+    }
+
+    buffer = (uint32_t*)realloc(buffer, sizeof(uint32_t) * width * height);
+    canvas->push(tvg::cast<tvg::Picture>(animation->picture()));
```
hermet, line 47: Simpler code:
```
cast(animation->picture());
```

src/savers/gif/tvgGifSaver.cpp
```diff
@@ -0,0 +1,129 @@
...
+    auto sw_canvas = reinterpret_cast<tvg::SwCanvas*>(canvas.get());
+    sw_canvas->target(buffer, width, width, height, tvg::SwCanvas::ABGR8888); 
+
+    GifWriter handle;
+    GifBegin(&handle, path.c_str(), width, height, 2);
```
hermet, line 52: "2" is it right?

src/savers/gif/tvgGifSaver.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+    float totalFrame = animation->totalFrame();
+    uint32_t frameDelay = 0;
+    if (fps != 0) {
+        frameDelay = static_cast<uint32_t>((animation->duration() / fps) * 100.0f);
```
hermet, line 57: Does this work correctly?

It suppose to have something like this code...

```
void GifSaver::run(unsigned tid)
{
    auto canvas = tvg::SwCanvas::gen();
    if (!canvas) return;

    //Do not share the memory pool since this canvas could be running on a thread.
    canvas->mempool(SwCanvas::Individual);
    
    auto w = static_cast<uint32_t>(vsize[0]);
    auto h = static_cast<uint32_t>(vsize[1]);

    buffer = (uint32_t*)realloc(buffer, sizeof(uint32_t) * w * h);
    canvas->target(buffer, w, w, h, tvg::SwCanvas::ABGR8888);
    canvas->push(tvg::cast<tvg::Picture>(animation->picture()));

    //use the default fps
    if (mathZero(fps)) {
        fps = (animation->totalFrame() / animation->duration());
    }

    auto delay = (fps / 60.0f);

    GifWriter writer;
    GifBegin(&writer, path, w, h, uint32_t(delay));

    auto duration = animation->duration() * 1000.0f;

    for (auto p = 0.0f; p < duration; p += delay) {
        auto frameNo = animation->totalFrame() * (p / duration);
        animation->frame(frameNo);
        canvas->update();
        if (canvas->draw() == tvg::Result::Success) { 
            canvas->sync();
        }
        GifWriteFrame(&writer, reinterpret_cast<uint8_t*>(buffer), w, h, uint32_t(delay));
    }

    GifEnd(&writer);
}
```

hermet (changes requested):
Please check comments.
- I think  GifSaver will be able to derive the Task class.
- Lzw Compression is already in thorvg/common. Probably we can consider a better optimal code.

hermet:
@JSUYA  Also, please check this revised version: https://github.com/thorvg/thorvg/pull/1749/commits/784419cc4eb818d11f4b5fb2c0e599e5bee1f972

src/renderer/tvgSaver.cpp
```diff
@@ -124,6 +139,32 @@ Result Saver::save(std::unique_ptr<Paint> paint, const string& path, bool compre
 }
 
 
+Result Saver::save(std::unique_ptr<Animation> animation, const string& path, uint32_t fps) noexcept
+{
```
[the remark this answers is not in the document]
JSUYA: I added if condition

src/savers/gif/tvgGifSaver.cpp
JSUYA: > ```
>     auto duration = animation->duration() * 1000.0f;
> ```

Since GifWriteFrame is 1/100 second, I modified it to *100.0.f.

hermet (approved):

hermet:
https://github.com/thorvg/thorvg/pull/1753

PR #1732, opened 2023-10-23
wg_engine: Added ability to draw multiple radial gradient filled shapes

SergeyLebedkin:
[issues 1479: RadialGradient](thorvg#1479)

In order to build you need third party libraries. Before you start please read this: [LearnWebGPU](https://eliemichel.github.io/LearnWebGPU/getting-started/hello-webgpu.html)

Usage example:

    // init glfw
    glfwInit();

    // create a windowed mode window and its opengl context
    glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API);
    GLFWwindow* window = glfwCreateWindow(800, 800, "WebGPU base app", nullptr, nullptr);

    // get window size
    int width{}, height{};
    glfwGetWindowSize(window, &width, &height);

    // init engine webgpu
    tvg::Initializer::init(tvg::CanvasEngine::Wg, 0);

    // create wg canvas
    auto canvasWg = tvg::WgCanvas::gen();
    canvas_wg->target(glfwGetWin32Window(window), width, height);

    // gradient color stops
    colorStops[0] = { 0, 255, 255, 255, 255 };
    colorStops[1] = { 1, 0, 0, 0, 255 };
    // radial gradient
    auto fill1 = tvg::RadialGradient::gen();
    fill1->radial(200, 200, 200);
    fill1->colorStops(colorStops, 2);
    // prepare rectangle
    shape1->appendRect(0, 0, 400, 400); //x, y, w, h
    shape1->fill(std::move(fill1));
    canvas_wg->push(std::move(shape1));

    // gradient color stops
    tvg::Fill::ColorStop colorStops2[3];
    colorStops2[0] = {0, 255, 0, 0, 255};
    colorStops2[1] = {0.5, 255, 255, 0, 255};
    colorStops2[2] = {1, 255, 255, 255, 255};
    //radial gradient
    auto fill2 = tvg::RadialGradient::gen();
    fill2->radial(400, 400, 200);
    fill2->colorStops(colorStops2, 3);
    // prepare circle
    shape2->appendCircle(400, 400, 200, 200); //cx, cy, radiusW, radiusH
    shape2->fill(std::move(fill2));
    canvas_wg->push(std::move(shape2));

    // gradient color stops
    tvg::Fill::ColorStop colorStops3[4];
    colorStops3[0] = {0, 0, 127, 0, 127};
    colorStops3[1] = {0.25, 0, 170, 170, 170};
    colorStops3[2] = {0.5, 200, 0, 200, 200};
    colorStops3[3] = {1, 255, 255, 255, 255};
    // radial gradient
    auto fill3 = tvg::RadialGradient::gen();
    fill3->radial(600, 600, 150);
    fill3->colorStops(colorStops3, 4);
    // prepare ellipse
    auto shape3 = tvg::Shape::gen();
    shape3->appendCircle(600, 600, 150, 100); //cx, cy, radiusW, radiusH
    shape3->fill(std::move(fill3));
    canvas_wg->push(std::move(shape3));

    while (!glfwWindowShouldClose(window)) {
        // webgpu
        canvas_wg->draw();
        canvas_wg->sync();

        // pull events
        glfwPollEvents();
    }

    // terminate engine and window
    tvg::Initializer::term(tvg::CanvasEngine::Wg);
    glfwDestroyWindow(window);
    glfwTerminate();

### DIFF
--- a/src/renderer/wg_engine/tvgWgPipelineRadial.cpp
+++ b/src/renderer/wg_engine/tvgWgPipelineRadial.cpp
@@ -1,0 +1,303 @@
+/*
+ * Copyright (c) 2023 the ThorVG project. All rights reserved.
+
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "tvgWgPipelineRadial.h"
+#include "tvgWgShaderSrc.h"
+
+//************************************************************************
+// WgPipelineBindGroupRadial
+//************************************************************************
+
+void WgPipelineDataRadial::updateGradient(RadialGradient* radialGradient) {
+    const Fill::ColorStop* stops = nullptr;
+    auto stopCnt = radialGradient->colorStops(&stops);
+
+    uGradientInfo.nStops[0] = stopCnt;
+    uGradientInfo.nStops[1] = 0.5f;
+
+    for (uint32_t i = 0; i < stopCnt; ++i) {
+        uGradientInfo.stopPoints[i] = stops[i].offset;
+        uGradientInfo.stopColors[i * 4 + 0] = stops[i].r / 255.f;
+        uGradientInfo.stopColors[i * 4 + 1] = stops[i].g / 255.f;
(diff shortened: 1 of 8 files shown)


src/renderer/wg_engine/tvgWgPipelineRadial.cpp
```diff
@@ -0,0 +1,303 @@
...
+ */
+
+#include "tvgWgPipelineRadial.h"
+#include "tvgWgShaderSrc.h"
+
+//************************************************************************
+// WgPipelineBindGroupRadial
+//************************************************************************
+
+void WgPipelineDataRadial::updateGradient(RadialGradient* radialGradient) {
```
hermet, line 30: So, the class logic in "WgPipelineDataRadial" is very similar to "WgPipelineDataLinear." This could lead to more maintenance costs. Can't we use a base class to share the main code and then derive from it for both Radial and Linear? It appears possible with the appropriate design for the class.

hermet (commented):
Please check a comment

src/renderer/wg_engine/tvgWgPipelineRadial.cpp
SergeyLebedkin: First of all, there are not only 2 classes, but at least 3 classes with the similar logic: Solid, Radial and Linear. But that’s only from the first look they are the similar. At least they have a different uniform layout, that needs to be described separately (they have a different buffer set, sizes, indexes).
Also, we have additional pipelines with absolutely different logic such as Empty pipeline (just for stencil writing), Image pipeline. And I'm not sure, that we don’t need additional pipeline for stroking. Only what we can disaggregate, its pipeline settings and shaders description, but it will over-complicate logic. In my logic pipeline and shader is an integral, solid object.
My main idea is: **single source file is single pipeline/shader** (no meter about logic). That’s really helpful for pipeline debugging and modification. 
From other point of view, we can add additional layers of inheritance. 

In current state it looks like this:
```
 PipelineBase -> PipelineEmpty
 PipelineBase -> PipelineImage
 PipelineBase -> (PipelineSolid, PipelineLinear, PipelineRadial)
```
But it will looks like this:
```
 PipelineBase -> PipelineEmpty
 PipelineBase -> PipelineImage
 PipelineBase -> PipelineFill -> PipelineSolid
 PipelineBase -> PipelineFill -> PipelineGradient -> (PipelineLinear, PipelineRadial)
```
But it this case we will add 2 addininal layers (PipelineFill and PipelineGradient) to 3 existing classes. Is that not over complicated?

hermet: I think the complexity wouldn't result from the number of classes/structures if it helps reduce lines of code (LOC), binary size, or improve performance, etc.

Understood. Either way, it's fine so far. Thanks.

hermet (approved):

PR #1736, opened 2023-10-26
renamed stroke apis family.

hermet:
@Issue: https://github.com/thorvg/thorvg/issues/1372

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -946,7 +946,7 @@ public:
      *
      * @return Result::Success when succeed, Result::FailedAllocation otherwise.
      */
-    Result stroke(float width) noexcept;
+    Result strokeWidth(float width) noexcept;
 
     /**
      * @brief Sets the color of the stroke for all of the figures from the path.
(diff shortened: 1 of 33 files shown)


hermet:
@JSUYA @SergeyLebedkin @RuiwenTang @mgrudzinska Hellow, please let me know if you have any opinion on this. Otherwise I will push this soon.

inc/thorvg.h
```diff
@@ -958,7 +958,7 @@ class TVG_API Shape final : public Paint
      *
      * @return Result::Success when succeed, Result::FailedAllocation otherwise.
      */
-    Result stroke(uint8_t r, uint8_t g, uint8_t b, uint8_t a = 255) noexcept;
+    Result strokeFill(uint8_t r, uint8_t g, uint8_t b, uint8_t a = 255) noexcept;
```
JSUYA, line 961: I think it is good if the parameters and names are intuitive. What about `strokeColor?'

hermet: I intentionally aligned with Shape::fill(uint8_t r, uint8_t g, uint8_t b, uint8_t a = 255);

hermet: fill is also acceptable to me, Otherwise, we could change `Shape::fill()` to `Shape::color()` as well. It would result in significant changes, and I would prefer to minimize these changes.

PR #1751, opened 2023-11-01
examples: Support mac os to execute all

tinyjin:
We can now call `examples/all.sh` on the mac os.

Added condition to detect OS and it replaces find command so the mac doesn't cause error.


hermet (approved):
thanks!

PR #1753, opened 2023-11-02
saver GifSaver: Introduce GifSaver for animation

hermet:
Add save() API that takes tvg::Animation as a parameter. This API uses gif.h to create each animation frame as a gif frame. Gif creation do not support threads because they must be added sequentially. Please see example/GifSaver.cpp

ex)
auto animation = tvg::Animation::gen();
picture->load(EXAMPLE_DIR"/walker.json");
auto saver = tvg::Saver::gen();
saver->save(std::move(animation), EXAMPLE_DIR"/test.gif"); saver->sync();

New API:
Result Saver::save(std::unique_ptr<Animation> animation, const std::string& path, uint32_t quality = 100, uint32_t fps = 0);

Issue: https://github.com/thorvg/thorvg/issues/1712

### DIFF
--- a/src/renderer/tvgSaver.cpp
+++ b/src/renderer/tvgSaver.cpp
@@ -26,6 +26,9 @@
 #ifdef THORVG_TVG_SAVER_SUPPORT
     #include "tvgTvgSaver.h"
 #endif
+#ifdef THORVG_GIF_SAVER_SUPPORT
+    #include "tvgGifSaver.h"
+#endif
 
 /************************************************************************/
 /* Internal Class Implementation                                        */
@@ -50,6 +53,12 @@ static SaveModule* _find(FileType type)
 #endif
             break;
         }
+        case FileType::Gif: {
+#ifdef THORVG_GIF_SAVER_SUPPORT
+            return new GifSaver;
+#endif
+            break;
+        }
         default: {
             break;
         }
@@ -60,6 +69,10 @@ static SaveModule* _find(FileType type)
     switch(type) {
         case FileType::Tvg: {
             format = "TVG";
+            break;
+        }
+        case FileType::Gif: {
+            format = "GIF";
             break;
         }
         default: {
@@ -78,6 +91,8 @@ static SaveModule* _find(const string& path)
     auto ext = path.substr(path.find_last_of(".") + 1);
     if (!ext.compare("tvg")) {
         return _find(FileType::Tvg);
+    } else if (!ext.compare("gif")) {
+        return _find(FileType::Gif);
     }
     return nullptr;
 }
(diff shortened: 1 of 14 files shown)


src/renderer/tvgSaver.cpp
```diff
@@ -124,6 +139,34 @@ Result Saver::save(std::unique_ptr<Paint> paint, const string& path, uint32_t qu
 }
 
 
+Result Saver::save(std::unique_ptr<Animation> animation, const string& path, uint32_t quality, uint32_t fps) noexcept
+{
+    if (mathZero(animation->totalFrame())) return Result::InsufficientCondition;
+
+    auto a = animation.release();
+    if (!a) return Result::MemoryCorruption;
```
JSUYA, line 147: Since animation may have been released, checking should given priority.

src/savers/gif/tvgGifSaver.cpp
```diff
@@ -0,0 +1,145 @@
...
+        if (canvas->draw() == tvg::Result::Success) {
+            canvas->sync();
+        }
+        GifWriteFrame(&writer, reinterpret_cast<uint8_t*>(buffer), w, h, uint32_t(delay * 100.0f));
+    }
+
+    GifEnd(&writer);
+}
+
+
```
JSUYA, line 81: empty line

src/savers/gif/tvgGifSaver.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <stdlib.h>
+#include "gif.h"
+#include "tvgGifSaver.h"
+
+
+/************************************************************************/
+/* Internal Class Implementation                                        */
+/************************************************************************/
+
+void GifSaver::run(unsigned tid)
```
JSUYA, line 32: Actually, since run() is a overrided public method, shouldn't it be declared in an external class?

JSUYA (commented):
Why did you create a new PR? Is it a rebase issue?

hermet:
> Why did you create a new PR? Is it a rebase issue?

Yes, your patch has been already verified. and this is a fork with your authoring.

src/savers/gif/tvgGifSaver.cpp
hermet: Other classes should not invoke run() directly.

JSUYA (approved):

hermet:
Somewhat this gif code is buggy:

generated frame image no 37 by gif:
image

frame capture from the canvas at no.37:
image

JSUYA:
> Somewhat this gif code is buggy:

This seems to be a problem with gif.h. I found the same problem with rlotte. I will track it down and resolve it.

Issue #1754, opened 2023-11-02
doxygen warnings


vtorri:
with doxygen 1.9.8

```
warning: Tag 'OUTPUT_TEXT_DIRECTION' at line 102 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'CLANG_ASSISTED_PARSING' at line 1160 of file 'Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature ena
bled.
warning: Tag 'CLANG_ADD_INC_PATHS' at line 1166 of file 'Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature ena
bled.
warning: Tag 'CLANG_OPTIONS' at line 1174 of file 'Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature ena
bled.
warning: Tag 'CLANG_DATABASE_PATH' at line 1187 of file 'Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature ena
bled.
warning: Tag 'HTML_TIMESTAMP' at line 1334 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'FORMULA_TRANSPARENT' at line 1639 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'LATEX_SOURCE_CODE' at line 1952 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'LATEX_TIMESTAMP' at line 1968 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'RTF_SOURCE_CODE' at line 2042 of file 'Doxyfile' has become obsolete.
[...]

PR #1755, opened 2023-11-02
docs: Remove all warnings from doxygen

tinyjin:
Fixed all warnings on building doxygen, wrong annotations are replaced from both `thorvg.h` and `thorvg_capi.h`.
Also I found unnecessary configs from `Doxyfile` and I commented it out.

@issue: #1754 

now it's clean without any warnings!


hermet:
> I'm checking on windows

I think this is an irrelevant issue, so you can ignore it.

hermet (approved):
LGTM. Thanks!

Issue #1764, opened 2023-11-06
Is it possible for picture raw loader to support `un-alpha-premultiplied`?


onecoolx:
I use thorvg's capi to develop vector drawing functions in the [LVGL](https://github.com/lvgl/lvgl) project. I found that when I wanted to use the raw loader to load a decoded ARGB8888 image, thorVG would draw it as `alpha-premultiplied`, which caused an error. It is caused by the following code：

```
unique_ptr<Surface> RawLoader::bitmap()
{
    if (!content) return nullptr;

    //TODO: It's better to keep this surface instance in the loader side
    auto surface = new Surface;
    surface->buf32 = content;
    surface->stride = static_cast<uint32_t>(w);
    surface->w = static_cast<uint32_t>(w);
    surface->h = static_cast<uint32_t>(h);
    surface->cs = cs;
    surface->channelSize = sizeof(uint32_t);
    surface->premultiplied = true;   ///  <----------------
    surface->owner = true;

    return unique_ptr<Surface>(surface);
}
```

Is it better to add a setting interface to set this value?

hermet:
@onecoolx Let me double-check; I think that option can be introduced upon demand. If you can quickly acquire the raw image with the pre-alpha version, it's better to use it directly since we can skip the conversion.

onecoolx:
> @onecoolx Let me double-check; I think that option can be introduced upon demand. If you can quickly acquire the raw image with the pre-alpha version, it's better to use it directly since we can skip the conversion.

Yes, is true, but currently LVGL does not support creating premultiplied image buffers directly. This is my  problem currently.

hermet:
@onecoolx Let's consider premultiplied image buffers in LVGL as well. Eventually, modern systems should output premultiplied pixels. It would be best to skip the conversion through the pipeline if possible. 

By the way, ThorVG is moving towards the 1.0 release, and we are considering breaking some APIs at this point. However, I don't think introducing a Surface interface for ThorVG is a good idea since WebGPU/GL APIs are not concrete and can be fragile. Instead, we can probably modify the load() API like this in this time:

Result Picture::load(uint32_t* data, uint32_t w, uint32_t h, bool copy);
-> Result Picture::load(uint32_t* data, uint32_t w, uint32_t h, bool premultiplied, bool copy);

Could you please confirm this is satisified with your demand?

onecoolx:
[...]

PR #1768, opened 2023-11-07
picture: added ability to support premultiplied for picture raw loader

SergeyLebedkin:
[issues 1479: picture raw loader to support premultiplied](https://github.com/thorvg/thorvg/issues/1764)

api changes:
    Result Picture::load(uint32_t* data, uint32_t w, uint32_t h, bool copy);
    Result Picture::load(uint32_t* data, uint32_t w, uint32_t h, bool premultiplied, bool copy);

capi changes
    TVG_API Tvg_Result tvg_picture_load_raw(Tvg_Paint* paint, uint32_t *data, uint32_t w, uint32_t h, bool copy);
    TVG_API Tvg_Result tvg_picture_load_raw(Tvg_Paint* paint, uint32_t *data, uint32_t w, uint32_t h, bool premultiplied, bool copy);

### DIFF
--- a/src/examples/Blending.cpp
+++ b/src/examples/Blending.cpp
@@ -132,7 +132,7 @@ void tvgDrawCmds(tvg::Canvas* canvas)
 
     //Lighten
     auto picture = tvg::Picture::gen();
-    if (picture->load(data, 200, 300, true) != tvg::Result::Success) return;
+    if (picture->load(data, 200, 300, true, true) != tvg::Result::Success) return;
     picture->translate(800, 700);
     picture->rotate(40);
     picture->blend(tvg::BlendMethod::Lighten);
@@ -147,7 +147,7 @@ void tvgDrawCmds(tvg::Canvas* canvas)
 
     //ColorBurn
     auto picture2 = tvg::Picture::gen();
-    if (picture2->load(data, 200, 300, true) != tvg::Result::Success) return;
+    if (picture2->load(data, 200, 300, true, true) != tvg::Result::Success) return;
     picture2->translate(600, 250);
     picture2->blend(tvg::BlendMethod::ColorBurn);
     picture2->opacity(150);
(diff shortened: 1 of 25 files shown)


src/examples/Blending.cpp
```diff
@@ -132,7 +132,7 @@ void tvgDrawCmds(tvg::Canvas* canvas)
 
     //Lighten
     auto picture = tvg::Picture::gen();
-    if (picture->load(data, 200, 300, true) != tvg::Result::Success) return;
+    if (picture->load(data, 200, 300, false, true) != tvg::Result::Success) return;
```
hermet, line 135: Exisiting default value was true.

hermet (changes requested):

src/examples/Blending.cpp
SergeyLebedkin: it keep true (the lastone parameter)

hermet: The last one indicates "copy", and we used the premultiplied version before. You don't need to break it.
See: tvgRawLoader.cpp:92

SergeyLebedkin: Git it. Fixed

src/examples/Duplicate.cpp
```diff
@@ -123,7 +123,7 @@ void tvgDrawCmds(tvg::Canvas* canvas)
         file.close();
 
         auto picture1 = tvg::Picture::gen();
-        if (picture1->load(data, 200, 300, true) != tvg::Result::Success) return;
+        if (picture1->load(data, 200, 300, false, true) != tvg::Result::Success) return;
```
hermet, line 126: Come on, could you please fix all other usages as well?

hermet (changes requested):

src/examples/Duplicate.cpp
SergeyLebedkin: Yes.. I know... On the way

hermet (approved):
LGTM

PR #1778, opened 2023-11-10
sw_engine: Clear buffer at the proper time.

hermet:
Clear the buffer when canvas->clear() is called.

TODO: We need to add a color value parameter to clear it.


hermet:
@capnm Hello, It will be great if you could help us confirm any Godot breakage by this.
@SergeyLebedkin Could you please double-check if this patch breaks the current wgpu backend?

hermet:
@SergeyLebedkin ping?

SergeyLebedkin:
> @SergeyLebedkin ping?
Looks good on webgpu side

Issue #1769, opened 2023-11-07
Support transparent GIF?


hermet:
-The original GIF doesn't have an alpha channel. 
-It needs to be confirmed whether the GIF allows for 32-bit (RGBA) channel data. 
-The current GIFSaver has been updated to fill the background with opaque white color. 
-Are there any portability issues?

hermet:
- How to decode/play the transparent gif?
- Custom gif won't be compatible with ohters unless we have a concrete specification,
- WebP might be better go.

hermet:
![abstract_circle](https://github.com/thorvg/thorvg/assets/3711518/d8770365-af86-4be9-9e28-3d684bb541a7)

PR #1782, opened 2023-11-13
Support Saver a background setting.

hermet:
https://github.com/thorvg/thorvg/issues/1769


hermet:
abstract_circle
abstract_circle1
abstract_circle2
abstract_circle3

Issue #1479, opened 2023-06-03
Support WebGPU raster backend


hermet:
Example working condition:

- [x] Shape
- [x] MultiShapes
- [x] Path
- [x] PathCopy
- [x] Scene
- [x] Transforms
- [x] Accessor
- [x] CustomTransform
- [x] Animation
- [x] Update
- [x] PictureSvg
- [x] DataLoad
- [x] LinearGradient
- [x] RadialGradient
- [x] PictureJpg
- [x] PicturePng
- [x] PictureWebp
- [x] SceneTransform
- [x] GradientTransform
- [x] FillRule
- [x] ImageScaleDown
- [x] ImageScaleUp
- [x] Arc
- [x] StrokeLine
- [x] DirectUpdate
- [x] Text
- [x] Retaining
- [x] Performance
- [x] Stroke
- [x] AnimateMasking
- [x] FillSpread
- [x] InvLumaMasking
- [x] Duplicate
- [ ] Viewport
- [ ] StrokeTrim
- [ ] PictureRaw
- [ ] InvMasking
- [ ] SceneBlending
- [ ] Masking
- [ ] LumaMasking
- [ ] GradientMasking
- [ ] ClipPath
- [ ] LottieExtensions
- [ ] Opacity
- [ ] StrokeMiterLimit
- [ ] MaskingMethods
- [ ] Texmap
- [ ] GradientStroke
- [ ] Blending
- [ ] PictureTvg
- [ ] Svg
- [ ] Tvg
- [ ] Lottie

hermet:
https://github.com/thorvg/thorvg/pull/1693

hermet:
https://github.com/thorvg/thorvg/wiki/24'-Development-Roadmap

hermet:
@kunitoki Yes, we do now.

hermet:
Currently 32 resources are incomplete support in Lottie examples:
- [ ] 11555.json
- [ ] 27746-joypixels-partying-face-emoji-animation.json
- [ ] 5317-fireworkds.json
- [ ] 5344-honey-sack-hud.json
- [ ] a_mountain.json
- [ ] batman.json
- [ ] driving.json
- [ ] eid_mubarak.json
- [ ] fiery_skull.json
- [ ] fly_in_beaker.json
- [ ] fun_animation.json
- [ ] game_finished.json
- [ ] gradient_background.json
- [ ] growup.json
- [ ] guitar.json
- [ ] hamburger.json
[...]

PR #1784, opened 2023-11-14
wg_engine: introduced images drawing support for webgpu renderer

SergeyLebedkin:
[issues 1479: pictures](https://github.com/thorvg/thorvg/issues/1479)

Exampe of usage:
```
    picture->load("images/test.png");
    picture->translate(0, 0);
    picture->size(100, 100);
    picture->opacity(255);
    canvas->push(std::move(picture));
```

### DIFF
--- a/src/renderer/wg_engine/tvgWgPipelineImage.cpp
+++ b/src/renderer/wg_engine/tvgWgPipelineImage.cpp
@@ -1,0 +1,414 @@
+/*
+ * Copyright (c) 2023 the ThorVG project. All rights reserved.
+
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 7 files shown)


src/renderer/wg_engine/tvgWgRenderData.h
```diff
@@ -70,15 +74,18 @@ class WgRenderDataShape: public WgRenderData {
 public:
     Array<WgGeometryData*> mGeometryDataShape;
     Array<WgGeometryData*> mGeometryDataStroke;
+    Array<WgGeometryData*> mGeometryDataImage;
```
hermet, line 77: In most cases, shape render data doesn't include images. Ultimately, it requires more precise packed data and handling.
```
WgRenderDataImage : public WgRenderDataShape
```

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -175,7 +179,21 @@ RenderData WgRenderer::prepare(const Array<RenderData>& scene, RenderData data,
...
 RenderData WgRenderer::prepare(Surface* surface, const RenderMesh* mesh, RenderData data, const RenderTransform* transform, Array<RenderData>& clips, uint8_t opacity, RenderUpdateFlag flags) {
-    return nullptr;
+    // get or create render data shape
+    auto renderDataShape = (WgRenderDataShape*)data;
+    if (!renderDataShape) {
+        renderDataShape = new WgRenderDataShape();
+        renderDataShape->initialize(mDevice);
+        renderDataShape->mPipelineBindGroupImage.initialize(mDevice, mPipelineImage, surface);
+    }
+    if (flags & (RenderUpdateFlag::Image | RenderUpdateFlag::Transform)) {
```
hermet, line 189: ++ RenderUpdateFlag::Color  //for opacity change.

hermet (commented):
Overall, look good to me, please check feedbacks.

src/renderer/wg_engine/tvgWgRenderer.cpp
SergeyLebedkin: Good point. Fixed

src/renderer/wg_engine/tvgWgPipelineImage.cpp
```diff
@@ -0,0 +1,421 @@
...
+ */
+
+#include "tvgWgPipelineImage.h"
+#include "tvgWgShaderSrc.h"
+
+//************************************************************************
+// WgPipelineDataImage
+//************************************************************************
+
+void WgPipelineDataImage::updateColor(const uint8_t* color) {
```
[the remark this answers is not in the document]
SergeyLebedkin: Yes... It should be removed

src/renderer/wg_engine/tvgWgRenderData.h
SergeyLebedkin: Good point. I have that idea too. But this case must be solved in more complex way: we need to made some refuctoring for pipelines, context and render targets. I believe that we will made it but not in this PR, bacouse it will impact to all rendering enteties, but not olny for images

hermet (approved):
LGTM!

Issue #1779, opened 2023-11-10
Revise the Canvas::clear() API.


hermet:
Add a chance to clear buffer with a specific color.

Expected API:
virtual Result Canvas::clear(uint8_t r, uint8_t g, uint8_t b, uint8_t a)

hermet:
@SergeyLebedkin @RuiwenTang Hello, Please let me know if you have any opinion.

RuiwenTang:
There is already a `clear` function on the canvas, which means clear all paints pushed into this canvas. Will this cause ambiguity?

BTW，does the `Canvas::clear(r,g,b,a)` method designed to be called repeatedly during a frame of drawing like this:
```
canvas->push(shape)
canvas->clear(0, 0, 0, 0)
canvas->push(shape)
```
Or is it just to clear the remaining drawing content at the beginning of a frame like this:
```
canvas->push(shape)

canvas->clear(0, 0, 0, 0)
canvas->draw()
canvas->sync()

```

hermet:
> here is already a `clear` function on the canvas, which means clear all paints pushed into this canvas. Will this cause ambiguity?

Let's see how we can remove the clear() for paints and merge it with new one, even though both resides within the codebase,  they are quite simple and explicitly identified with different parameters.

> BTW，does the `Canvas::clear(r,g,b,a)` method designed to be called repeatedly during a frame of drawing like this:

The second case. Clear(r, g,b , a) is not mandatory btw, if user fill the explicit background with a shape.

RuiwenTang:
> The second case. Clear(r, g,b , a) is not mandatory btw, if user fill the explicit background with a shape.

For this case. Can we put the `color parameters` in `Canvas::draw()` like 
```Canvas::draw(uint8_t r = 0, uint8_t g = 0, uint8_t b = 0, uint8_t a = 0)```
Since when begin a RenderPass in GPU backend, can specify a clean color if the color attachment load_op is `Clear` which is the most common use case.

hermet:
clear buffer is not

> > The second case. Clear(r, g,b , a) is not mandatory btw, if user fill the explicit background with a shape.
> 
> For this case. Can we put the `color parameters` in `Canvas::draw()` like `Canvas::draw(uint8_t r = 0, uint8_t g = 0, uint8_t b = 0, uint8_t a = 0)` Since when begin a RenderPass in GPU backend, can specify a clean color if the color attachment load_op is `Clear` (which is the most common use case.)

[...]

PR #1790, opened 2023-11-15
ThorVG API: Revise the Canvas::clear() API

SergeyLebedkin:
[issues 1779: Revise the Canvas::clear() API](https://github.com/thorvg/thorvg/issues/1779)

new api function was added

    /**
     * @brief Clear the target buffer that used for the drawing.
     *
     * @param[in] r The red color channel value in the range [0 ~ 255]. The default value is 0.
     * @param[in] g The green color channel value in the range [0 ~ 255]. The default value is 0.
     * @param[in] b The blue color channel value in the range [0 ~ 255]. The default value is 0.
     * @param[in] a The alpha channel value in the range [0 ~ 255], where 0 is completely transparent and 255 is opaque. The default value is 0.
     *
     * @return Result::Success when succeed, Result::InsufficientCondition otherwise.
     */
    virtual Result clear(uint8_t r, uint8_t g, uint8_t b, uint8_t a) noexcept;

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderTask.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderTask.cpp
@@ -106,16 +106,25 @@ GlComposeTask::~GlComposeTask()
     mTasks.clear();
 }
 
+void GlComposeTask::clear(GLfloat clearColor[4]) {
+    mClearColor[0] = clearColor[0];
+    mClearColor[1] = clearColor[1];
+    mClearColor[2] = clearColor[2];
+    mClearColor[3] = clearColor[3];
+    mClear = true;
+}
+
 void GlComposeTask::run()
 {
     GL_CHECK(glBindFramebuffer(GL_FRAMEBUFFER, getSelfFbo()));
 
     // clear this fbo
     GLenum color_buffer = GL_COLOR_ATTACHMENT0;
-    const float transparent[] = {0.f, 0.f, 0.f, 0.f};
-
     GL_CHECK(glDrawBuffers(1, &color_buffer));
-    GL_CHECK(glClearBufferfv(GL_COLOR, 0, transparent));
+    if (mClear) {
+        GL_CHECK(glClearBufferfv(GL_COLOR, 0, mClearColor));
+        mClear = false;
+    }
 
     for(uint32_t i = 0; i < mTasks.count; i++) {
         mTasks[i]->run();
--- a/src/renderer/sw_engine/tvgSwRenderer.cpp
+++ b/src/renderer/sw_engine/tvgSwRenderer.cpp
@@ -374,13 +374,11 @@ SwRenderer::~SwRenderer()
     if (rendererCnt == 0 && initEngineCnt == 0) _termEngine();
 }
 
-
-bool SwRenderer::clear()
+bool SwRenderer::clear(uint8_t r, uint8_t g, uint8_t b, uint8_t a)
 {
-    if (surface) return rasterClear(surface, 0, 0, surface->w, surface->h);
-    return false;
+    if (surface) return rasterClear(surface, surface->join(r, g, b, a), 0, 0, surface->w, surface->h);
+    return true;
 }
-
 
 bool SwRenderer::sync()
 {
@@ -668,7 +666,7 @@ Compositor* SwRenderer::target(const RenderRegion& region, ColorSpace cs)
     cmp->w = cmp->compositor->image.w;
     cmp->h = cmp->compositor->image.h;
 
-    rasterClear(cmp, x, y, w, h);
+    rasterClear(cmp, 0x00, x, y, w, h);
 
     //Switch render target
     surface = cmp;
--- a/src/renderer/sw_engine/tvgSwRenderer.h
+++ b/src/renderer/sw_engine/tvgSwRenderer.h
@@ -50,7 +50,7 @@ public:
     bool blend(BlendMethod method) override;
     ColorSpace colorSpace() override;
 
-    bool clear() override;
+    bool clear(uint8_t r, uint8_t g, uint8_t b, uint8_t a) override;
     bool sync() override;
     bool target(pixel_t* data, uint32_t stride, uint32_t w, uint32_t h, ColorSpace cs);
     bool mempool(bool shared);
--- a/src/renderer/tvgCanvas.h
+++ b/src/renderer/tvgCanvas.h
@@ -76,11 +76,16 @@ struct Canvas::Impl
         return update(p, true);
     }
 
-    Result clear(bool free)
+    Result clear(uint8_t r, uint8_t g, uint8_t b, uint8_t a)
     {
         //Clear render target
-        if (!renderer || !renderer->clear()) return Result::InsufficientCondition;
+        if (!renderer || !renderer->clear(r, g, b, a)) return Result::InsufficientCondition;
 
+        return Result::Success;
+    }
+
+    Result clear(bool free)
+    {
         clearPaints(free);
 
         return Result::Success;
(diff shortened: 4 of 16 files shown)


src/renderer/sw_engine/tvgSwRenderer.cpp
```diff
@@ -374,10 +374,15 @@ SwRenderer::~SwRenderer()
     if (rendererCnt == 0 && initEngineCnt == 0) _termEngine();
 }
 
+bool SwRenderer::clear(uint8_t r, uint8_t g, uint8_t b, uint8_t a)
+{
+    this->clearColor = JOIN(a, b, g, r);
+    if (surface) return rasterClear(surface, this->clearColor, 0, 0, surface->w, surface->h);
```
hermet, line 380: this->clearColor -> surface->join(r, g, b, a)

src/renderer/sw_engine/tvgSwRenderer.cpp
```diff
@@ -668,7 +673,7 @@ Compositor* SwRenderer::target(const RenderRegion& region, ColorSpace cs)
     cmp->w = cmp->compositor->image.w;
     cmp->h = cmp->compositor->image.h;
 
-    rasterClear(cmp, x, y, w, h);
+    rasterClear(cmp, clearColor, x, y, w, h);
```
hermet, line 676: this is an intermediate target clear, so it should come up with "0" value. (not clearColor)

src/renderer/sw_engine/tvgSwRenderer.h
```diff
@@ -72,6 +73,7 @@ class SwRenderer : public RenderMethod
     SwMpool*             mpool;                       //private memory pool
     RenderRegion         vport;                       //viewport
     bool                 sharedMpool = true;          //memory-pool behavior policy
+    uint32_t             clearColor = 0;
```
hermet, line 76: --

src/renderer/tvgRender.h
```diff
@@ -260,6 +260,7 @@ class RenderMethod
     virtual bool blend(BlendMethod method) = 0;
     virtual ColorSpace colorSpace() = 0;
 
+    virtual bool clear(uint8_t r, uint8_t g, uint8_t b, uint8_t a) = 0;
     virtual bool clear() = 0;
```
hermet, line 264: please remove clear()

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -221,6 +221,15 @@ ColorSpace WgRenderer::colorSpace() {
     return ColorSpace::Unsupported;
 }
 
+bool WgRenderer::clear(uint8_t r, uint8_t g, uint8_t b, uint8_t a) {
+    mClearColor.r = r / 255.0;
```
hermet, line 225: 255.0f;

hermet (changes requested):
+ some quick feedback

src/renderer/gl_engine/tvgGlRenderTask.cpp
```diff
@@ -106,16 +106,24 @@ GlComposeTask::~GlComposeTask()
...
     GL_CHECK(glBindFramebuffer(GL_FRAMEBUFFER, getSelfFbo()));
 
     // clear this fbo
     GLenum color_buffer = GL_COLOR_ATTACHMENT0;
-    const float transparent[] = {0.f, 0.f, 0.f, 0.f};
-
     GL_CHECK(glDrawBuffers(1, &color_buffer));
-    GL_CHECK(glClearBufferfv(GL_COLOR, 0, transparent));
+    if (mClear) {
+        GL_CHECK(glClearBufferfv(GL_COLOR, 0, mClearColor));
```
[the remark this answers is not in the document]
SergeyLebedkin: not clearly undestand

src/renderer/wg_engine/tvgWgRenderer.cpp
SergeyLebedkin: image

src/renderer/sw_engine/tvgSwRenderer.cpp
```diff
@@ -374,14 +374,12 @@ SwRenderer::~SwRenderer()
     if (rendererCnt == 0 && initEngineCnt == 0) _termEngine();
 }
 
-
-bool SwRenderer::clear()
+bool SwRenderer::clear(uint8_t r, uint8_t g, uint8_t b, uint8_t a)
 {
-    if (surface) return rasterClear(surface, 0, 0, surface->w, surface->h);
-    return false;
+    if (surface) return rasterClear(surface, JOIN(a, b, g, r), 0, 0, surface->w, surface->h);
```
hermet, line 379: surface->join(r, g, b, a)

src/renderer/tvgCanvas.h
```diff
@@ -76,11 +76,16 @@ struct Canvas::Impl
         return update(p, true);
     }
 
-    Result clear(bool free)
+    Result clear(uint8_t r, uint8_t g, uint8_t b, uint8_t a)
     {
         //Clear render target
-        if (!renderer || !renderer->clear()) return Result::InsufficientCondition;
+        if (!renderer || !renderer->clear(r, g, b,a)) return Result::InsufficientCondition;
```
hermet, line 82: b, a

hermet (changes requested):

src/renderer/gl_engine/tvgGlRenderTask.cpp
```diff
@@ -106,16 +106,25 @@ GlComposeTask::~GlComposeTask()
...
 void GlComposeTask::run()
 {
     GL_CHECK(glBindFramebuffer(GL_FRAMEBUFFER, getSelfFbo()));
 
     // clear this fbo
     GLenum color_buffer = GL_COLOR_ATTACHMENT0;
-    const float transparent[] = {0.f, 0.f, 0.f, 0.f};
-
     GL_CHECK(glDrawBuffers(1, &color_buffer));
-    GL_CHECK(glClearBufferfv(GL_COLOR, 0, transparent));
```
RuiwenTang, line 118: GlComposeTask always need to clear the color-buffer since it may used by a off-screen rendering.
So I think, just change keep the default value for the mClearColor as transparent, and change it if outside user need a different background color

hermet: Since the graphics library (GL) is under development, either way is fine with me. However, so far, raster engines don't need to determine this behavior like OpenGL does. We open the `clear()` API, and users can choose it according to their demand—simple. The `clear()` function is not designed to fill the background color; it is designed for composing with other existing(previous) images. If we clear it, they have no choice but to do so.

RuiwenTang: Now, in the GL backend, the canvas's content is rendered to a off-screen framebuffer first, and then blit to the outside framebuffer. If need to composite with outside framebuffer ,the only change need to do is **render the texture** to outside framebuffer (like compose the SubRenderPass to the main RenderPass) not **blit the texture**.

hermet: Ultimately, we will consider the target to be either the main or intermediate target. Secondly, even though the GL target serves as the framebuffer object (FBO), users might want to add more elements on top of the previous image. IMHO, There are no clear benefits to clearing the buffer by the engine itself, but just add more restrcited usage.

hermet: An even more important point is that a consistent behavior through the raster backends for the clear() API is required for ThorVG.

RuiwenTang: Maybe we can define canvas loading operations (clear, load) like WebGPU and let the backend implement the behavior

hermet:
Moved to https://github.com/thorvg/thorvg/pull/1798

PR #1792, opened 2023-11-16
gif: fixed a regresion bug by clear() change

hermet:
by 66305f3e6d0001ffca70e4337275900cdda8e5bb


src/savers/gif/gif.h
```diff
@@ -1,3 +1,26 @@
+/*
```
JSUYA, line 1: It looks that gif.h has been replaced with tvgGifEncoder.h.
Is there a reason to keep this?

hermet: 1. `gif.h` has been changed since its introduction; it's no longer just pure "gif.h" now. 
2. By this reason, it follows thorvg file name convention.
3. It's buggy, It needs to revise it, otherwise gif.h is not useful for production level.
4. lzw encoding will be replaced with us addtionally.

These are my consideration.

JSUYA: oh I mean git.h can be removed.

hermet: Oh my misunderstanding, it's just my fault.

JSUYA (approved):

PR #1797, opened 2023-11-19
webgpu: add support for textures color space formats

SergeyLebedkin:
[Issues 1479: pictures](https://github.com/thorvg/thorvg/issues/1479)

Formats supported:
    ABGR8888
    ARGB8888
    ABGR8888S
    ARGB8888S

### DIFF
--- a/src/renderer/wg_engine/tvgWgShaderSrc.cpp
+++ b/src/renderer/wg_engine/tvgWgShaderSrc.cpp
@@ -281,7 +281,10 @@ struct Matrix {
 
 // ColorInfo
 struct ColorInfo {
-    color: vec4f
+    format: u32,
+    dummy0: f32,
+    dummy1: f32,
+    opacity: f32
 };
 
 // vertex output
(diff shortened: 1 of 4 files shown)


src/renderer/wg_engine/tvgWgShaderSrc.cpp
```diff
@@ -281,7 +281,10 @@ struct Matrix {
 
 // ColorInfo
 struct ColorInfo {
-    color: vec4f
+    format: u32,
+    dummy0: f32,
```
hermet, line 285: Question: what are these dummies for?

SergeyLebedkin: > Question: what are these dummies for?

There are 2 reasons:
1. Its reserved values for future functionality, f.e. UV flipping
2. Its alligmant for uniform data structures. The most effective data structure should have size alligned to 32x4 bits. Thats why we have 4 values by 4 bytes and two of them are dummy for now

hermet (approved):

PR #1798, opened 2023-11-20
apis/engines: Revise the clear() buffer behavior.

hermet:
ThorVG has offered an option to clear the buffer since version 1.0. This is essential when users utilize the canvas target buffer with the main render target. They share the buffer and need to draw contents onto the existing contents.

API:
Result Canvas::clear(bool free = true)
-> Result Canvas::clear(bool free = true, bool buffer = true)

Issue: https://github.com/thorvg/thorvg/issues/1779


src/renderer/gl_engine/tvgGlRenderTask.cpp
```diff
@@ -111,11 +111,10 @@ void GlComposeTask::run()
     GL_CHECK(glBindFramebuffer(GL_FRAMEBUFFER, getSelfFbo()));
 
     // clear this fbo
-    GLenum color_buffer = GL_COLOR_ATTACHMENT0;
-    const float transparent[] = {0.f, 0.f, 0.f, 0.f};
-
-    GL_CHECK(glDrawBuffers(1, &color_buffer));
```
hermet, line 117: @RuiwenTang  removed this color buffer.

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -273,9 +274,13 @@ bool WgRenderer::sync() {
             WGPURenderPassColorAttachment colorAttachment{};
             colorAttachment.view = backBufferView;
             colorAttachment.resolveTarget = nullptr;
-            colorAttachment.loadOp = WGPULoadOp_Clear;
+            if (mClearBuffer) {
```
hermet, line 277: @SergeyLebedkin Please confirm

src/renderer/gl_engine/tvgGlRenderTask.cpp
RuiwenTang: If examples/Animation can be rendered correctly, I think it can be removed.

hermet: Sure, it works.

RuiwenTang (approved):

hermet:
@gaelyvan We gonna apply this change for skipping buffer clearing. Please let us know if this doesn't work for you.

src/renderer/wg_engine/tvgWgRenderer.cpp
SergeyLebedkin: colorAttachment.clearValue = { 0.0, 0.0, 0.0, 1.0 };
move floats to doubles
alpha channel should be 1.0, i guess

hermet: I intentionally changed it. the cleared buffer should become total transparent, should not have any alpha values. So the previous is inccorect.

SergeyLebedkin: OK, And by the spec color channels values have a double type

Issue #1793, opened 2023-11-17
Need a set of file path when loading Picture data from the memory


hermet:
The data loaded from memory can internally reference different resources. For example, SVG and Lottie can internally have embedded bitmap images, and a method is needed to specify their location path. We can ad a parameter to the load function to specify the resource path. This will break API in 1.0 release.

```
Result Picture::load(const char* data, uint32_t size, const std::string& mimeType, bool copy = false)
-> Result Picture::load(const char* data, uint32_t size, const std::string& mimeType, bool copy = false, const std::string& resourcePath = "")  

TVG_API Tvg_Result tvg_picture_load_data(Tvg_Paint* paint, const char *data, uint32_t size, const char *mimetype, bool copy);
-> TVG_API Tvg_Result tvg_picture_load_data(Tvg_Paint* paint, const char *data, uint32_t size, const char *mimetype, bool copy, const char* resource_path);
```

See also: tvgLottieParser.cpp(834):
LottieImage* LottieParser::parseImage(const char* key)

hermet:
@samuelOsborne

hermet:
[sample.zip](https://github.com/thorvg/thorvg/files/13388713/sample.zip)

PR #1802, opened 2023-11-21
lottie/loader: Support external image with memory animation

tinyjin:
tvgLottie wasn't able to load external image when loading through memory (not file), added new parameter `resourcePath`(optional, default = `""`).

API:
Result Picture::load(const char* data, uint32_t size, const string& mimeType, bool copy)
-> Result Picture::load(const char* data, uint32_t size, const string& mimeType, bool copy, const string& resourcePath)

New parameter:
- `resourcePath`: absolute directory path where resource directory is included.

Example:
Animation and asset directory are located in `/home/anims` and lottie has assets like
```json
"assets": [
    {
      "id": "image_0",
      "w": 200,
      "h": 300,
      "u": "images/",
      "p": "img_0.png",
      "e": 0
    }
  ]
```
The parameter `resourcePath` should be `/home/anims` then tvgLottieLoader will load external image `/home/anims/images/img_0.png`

Usage:
```cpp
picture->load(data, size, "lottie", true, "/absolute/path/to/dir")
```

Issue: #1793

### DIFF
--- a/src/bindings/capi/thorvg_capi.h
+++ b/src/bindings/capi/thorvg_capi.h
@@ -1993,6 +1993,7 @@ TVG_API Tvg_Result tvg_picture_load_raw(Tvg_Paint* paint, uint32_t *data, uint32
 * \param[in] size The size in bytes of the memory occupied by the @p data.
 * \param[in] mimetype Mimetype or extension of data such as "jpg", "jpeg", "svg", "svg+xml", "lottie", "png", etc. In case an empty string or an unknown type is provided, the loaders will be tried one by one.
 * \param[in] copy If @c true the data are copied into the engine local buffer, otherwise they are not.
+* \param[in] resourcePath Directory path to load external images.
 *
 * \return Tvg_Result enumeration.
 * \retval TVG_RESULT_SUCCESS Succeed.
@@ -2002,7 +2003,7 @@ TVG_API Tvg_Result tvg_picture_load_raw(Tvg_Paint* paint, uint32_t *data, uint32
 *
 * \warning: It's the user responsibility to release the @p data memory if the @p copy is @c true.
 */
-TVG_API Tvg_Result tvg_picture_load_data(Tvg_Paint* paint, const char *data, uint32_t size, const char *mimetype, bool copy);
+TVG_API Tvg_Result tvg_picture_load_data(Tvg_Paint* paint, const char *data, uint32_t size, const char *mimetype, bool copy, const char* resourcePath = "");
 
 
 /*!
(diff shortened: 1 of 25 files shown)


hermet:
@tinyjin please double-check build fail at windows.
image

hermet (commented):
Please check comments.

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -227,9 +227,8 @@ bool LottieLoader::open(const char* data, uint32_t size, bool copy, const std::s
     this->copy = copy;
 
     if (!resourcePath.empty()) {
-        auto path = resourcePath.c_str();
-        auto len = strlen(path);
-        this->dirName = strDuplicate(path, len);
+        if (resourcePath.back() == '/') this->dirName = strdup(resourcePath.c_str());
```
hermet, line 230: this patch shouldn't have this change.

hermet (changes requested):
Please check a comment.

src/bindings/capi/thorvg_capi.h
```diff
@@ -20,6 +20,7 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <string>
```
hermet, line 23: Ah, please remove string, it's not allowed in C language.

src/bindings/capi/thorvg_capi.h
```diff
@@ -2002,7 +2004,7 @@ TVG_API Tvg_Result tvg_picture_load_raw(Tvg_Paint* paint, uint32_t *data, uint32
 *
 * \warning: It's the user responsibility to release the @p data memory if the @p copy is @c true.
 */
-TVG_API Tvg_Result tvg_picture_load_data(Tvg_Paint* paint, const char *data, uint32_t size, const char *mimetype, bool copy);
+TVG_API Tvg_Result tvg_picture_load_data(Tvg_Paint* paint, const char *data, uint32_t size, const char *mimetype, bool copy, const std::string& resourcePath = "");
```
hermet, line 2007: string is out of C language.
const std::string& resourcePath = ""  ->   const char* resourcePath

hermet (changes requested):
Please check one more missing point.

src/loaders/lottie/tvgLottieLoader.cpp
hermet: Ok, This is the last part.

Since the other platforms might not same with '/' but '\\'.  it's getting become complex. 
So, rather then appending "/" with a more complex condition here,  I'd like to suggest add it all time in tvgLottieParser.cpp (872)
```
 snprintf(image->path, len, "%s%s%s", dirName, subPath, data); ->
        snprintf(image->path, len, "%s/%s%s", dirName, subPath, data);
```

hermet (commented):
The last comment. Please check it.

hermet (approved):

Issue #1687, opened 2023-10-05
Support GL raster backend


hermet:
The current gl engine is under development; the features are unstable, and its coverage is quite low.

Confirmed example working

- [x] Shape
- [x] MultiShapes
- [x] Path
- [x] PathCopy
- [x] Scene
- [x] Transform
- [x] Accessor
- [x] CustomTransform
- [x] Animation
- [x] Update
- [x] PictureSvg
- [x] DataLoad
- [x] LinearGradient
- [x] RadialGradient
- [x] PictureJpg
- [x] PicturePng
- [x] PictureWebp
- [x] SceneTransform
- [x] GradientTransform
- [x] FillRule
- [x] ImageScaleDown
- [x] ImageScaleUp
- [x] Arc
- [x] StrokeLine
- [x] DirectUpdate
- [x] Text
- [x] Retaining
- [x] Stroke
- [x] AnimateMasking
- [x] InvLumaMasking
- [x] FillSpread
- [x] PictureRaw
- [x] InvMasking
- [x] SceneBlending
- [x] Masking
- [x] LumaMasking
- [x] GradientMasking
- [x] ClipPath
- [x] LottieExtensions
- [x] Opacity
- [x] MaskingMethods
- [x] Texmap
- [x] GradientStroke
- [x] Svg
- [x] Tvg
- [ ] StrokeMiterLimit
- [ ] Performance
- [ ] Viewport
- [ ] StrokeTrim
- [ ] Duplicate
- [ ] Blending
- [ ] PictureTvg
- [ ] Lottie

hermet:
https://github.com/thorvg/thorvg/pull/1531

hermet:
https://github.com/thorvg/thorvg/pull/1704

hermet:
Radial Gradient fill is somewhat different.
![shot-2023-12-27_16-44-51](https://github.com/thorvg/thorvg/assets/3711518/8ecadd48-a432-4b9b-b53d-866b3df6a7bd)

hermet:
@RuiwenTang Radial Gradient fill is a bit different. I remember it worked properly before?
![shot-2023-12-27_16-44-51](https://github.com/thorvg/thorvg/assets/3711518/8ecadd48-a432-4b9b-b53d-866b3df6a7bd)

hermet:
Retaining doesn't updated.

RuiwenTang:
Maybe my recent commit broke it, I'll check

hermet:
[...]

PR #1806, opened 2023-11-22
gl_engine: support advance compose method

RuiwenTang:
This PR implement the ComposeMethods in GL backends.
> In some cases, the rendering may be incorrect, and off-screen rendering needs to be optimized. 

## example/MaskingMethods
截屏2023-11-22 下午11 32 15

https://github.com/thorvg/thorvg/issues/1687

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -117,7 +117,7 @@ bool GlRenderer::sync()
 
 RenderRegion GlRenderer::region(TVG_UNUSED RenderData data)
 {
-    return {0, 0, 0, 0};
+    return {0, 0, static_cast<int32_t>(surface.w), static_cast<int32_t>(surface.h)};
 }
 
 
@@ -151,18 +151,14 @@ bool GlRenderer::beginComposite(Compositor* cmp, CompositeMethod method, uint8_t
 {
     if (!cmp) return false;
 
-    // TODO handle other composite method with recursive begin composite
-
     cmp->method = method;
     cmp->opacity = opacity;
 
-    if (cmp->method == CompositeMethod::None) {
-        if (mPoolIndex >= mComposePool.size()) {
-            mComposePool.emplace_back(make_unique<GlRenderTarget>(surface.w, surface.h));
-            mComposePool.back()->init(mTargetFboId);
-        }
-        mRenderPassStack.emplace_back(GlRenderPass(mComposePool[mPoolIndex++].get()));
+    if (mPoolIndex >= mComposePool.size()) {
+        mComposePool.emplace_back(make_unique<GlRenderTarget>(surface.w, surface.h));
+        mComposePool.back()->init(mTargetFboId);
     }
+    mRenderPassStack.emplace_back(GlRenderPass(mComposePool[mPoolIndex++].get()));
 
     return true;
 }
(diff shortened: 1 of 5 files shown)


src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -627,25 +633,25 @@ void GlRenderer::prepareCmpTask(GlRenderTask* task)
...
+    vertices.push(opacity);
     vertices.push(1.f);
     vertices.push(1.f);
     // right bottom point
     vertices.push(right);
     vertices.push(bottom);
-    vertices.push(1.f);
+    vertices.push(opacity);
     vertices.push(1.f);
     vertices.push(0.f);
```
SergeyLebedkin, line 657: Should we pass additional opacity value via vertex data? As I can see there is general value for the mesh, and we are just doubling this information

src/renderer/gl_engine/tvgGlShaderSrc.h
```diff
@@ -30,5 +30,14 @@ extern const char* LINEAR_GRADIENT_FRAG_SHADER;
...
+extern const char* MASK_ALPHA_FRAG_SHADER;
+extern const char* MASK_INV_ALPHA_FRAG_SHADER;
+extern const char* MASK_LUMA_FRAG_SHADER;
+extern const char* MASK_INV_LUMA_FRAG_SHADER;
+extern const char* MASK_ADD_FRAG_SHADER;
+extern const char* MASK_SUB_FRAG_SHADER;
+extern const char* MASK_INTERSECT_FRAG_SHADER;
+extern const char* MASK_DIFF_FRAG_SHADER;
 
 #endif /* _TVG_GL_SHADERSRC_H_ */
```
SergeyLebedkin, line 43: Should we use 8 different shaders, or maybe use a single shader with switch inside of it?

SergeyLebedkin (commented):
Add some comments

src/renderer/gl_engine/tvgGlShaderSrc.h
RuiwenTang: Complex if-else branch in fragment shader may negatively impact performance. 
And also need additional uniform variable to control these branch.

Consider the performance and for clean shader code, I choose split these branches into sperate shaders

src/renderer/gl_engine/tvgGlRenderer.cpp
RuiwenTang: This is just for convenience as it's just a single float value and I don't want to create a separate props and uniform buffer to store and pass it around. :)

hermet:
@RuiwenTang Hello, do you have any consideration for an AA mechanism for later apply?

RuiwenTang:
> @RuiwenTang Hello, do you have any consideration for an AA mechanism for later apply?

Yes, but I plan to do the following first:
* Implement all missing features in GL engine
* Optimize the triangulation algorithm for now it has some bad case can not handled
* Optimize the off-screen rendering, the current rendering strategy is somewhat wasteful

src/renderer/gl_engine/tvgGlShaderSrc.h
hermet: Maybe we can use them in the shader code when shaders are quite simlar.
 #if  ~   #elif  ~  #endif

hermet (commented):
LGTM.

SergeyLebedkin:
LGTM

SergeyLebedkin (approved):
LGTM

hermet (approved):

PR #1807, opened 2023-11-22
api: updated the recent changed api again.

hermet:
Reordered by the data packing size.
Also removed a wrong capi default parameter value.

Result Picture::load(const char* data, uint32_t size, const std::string& mimeType, bool copy = false, const std::string& rpath = "") -> Result load(const char* data, uint32_t size, const std::string& mimeType, const std::string& rpath = "", bool copy = false)

Tvg_Result tvg_picture_load_data(Tvg_Paint* paint, const char *data, uint32_t size, const char *mimetype, bool copy, const char* resourcePath) -> Tvg_Result tvg_picture_load_data(Tvg_Paint* paint, const char *data, uint32_t size, const char *mimetype, const char* resourcePath, bool copy);

### DIFF
--- a/src/examples/PictureJpg.cpp
+++ b/src/examples/PictureJpg.cpp
@@ -60,7 +60,7 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     file.close();
 
     auto picture = tvg::Picture::gen();
-    if (picture->load(data, size, "jpg", true) != tvg::Result::Success) {
+    if (picture->load(data, size, "jpg", "", true) != tvg::Result::Success) {
         cout << "Couldn't load JPG file from data." << endl;
         return;
     }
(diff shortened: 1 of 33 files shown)


src/examples/PictureJpg.cpp
```diff
@@ -60,7 +60,7 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     file.close();
 
     auto picture = tvg::Picture::gen();
-    if (picture->load(data, size, "jpg", true) != tvg::Result::Success) {
+    if (picture->load(data, size, "jpg", "", true) != tvg::Result::Success) {
```
JSUYA, line 63: How about declaring a macro for `""`?
```
#define TVG_NO_EXTERNAL_RESOURCE ""
```

JSUYA (approved):

src/examples/PictureJpg.cpp
hermet: Well... I have no idea about it but I'd like to change string -> char*

PR #1822, opened 2023-12-07
[build][task-scheduler]: Added THORVG_THREAD_SUPPORT macro in order to be able to not use threads

### LINKED ISSUE #1821: Not able to compile thorvg without thread support
I'm trying to create a version of Godot 4.x that would compile without thread support. 

This is to make the editor and to export games on a non-COOP/COEP web environment. Due to Meltdown and Spectre, the web platform requires COOP and COEP in order to have access to `SharedArrayBuffer`, which is required by Emscripten to compile and use threads.

Unfortunately, I can compile thorvg without `USE_PTHREADS=1`, but it will inevitably crash when trying to instantiate `tvg::TaskScheduler`, even if we could set `tvg::TaskScheduler::async(false)` afterwards.

Capture d’écran du 2023-12-06 14-09-53

Ideally, it should be possible to compile thorvg in a way to forgo the use of threads.

adamscott:
This PR adds the `THORVG_THREAD_SUPPORT` macro and the `-Dthreads` build option (by default `true`).

This makes it possible to disable threads compile time.

Fixes #1821

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -87,6 +87,11 @@ if get_option('vector') == true
   endif
 endif
 
+#Threads
+if get_option('threads') == true
+    config_h.set10('THORVG_THREAD_SUPPORT', true)
+endif
+
 #Bindings
 if get_option('bindings').contains('capi') == true
     config_h.set10('THORVG_CAPI_BINDING_SUPPORT', true)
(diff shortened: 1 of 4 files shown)


adamscott:
image

hermet:
@adamscott thanks for your patch. I mentioned the proper solution in the https://github.com/thorvg/thorvg/issues/1821 Could you please double checkit it? We've adopted the thorvg wasm for the web viewers already with that approach.

adamscott:
Thanks for your tip and your work, @hermet !  Closing this PR as this isn't necessary.

Issue #1824, opened 2023-12-08
Rounded rectangles are slightly incorrect


MewPurPur:
![image](https://github.com/thorvg/thorvg/assets/85438892/ce829f8b-9bb8-4280-b550-60caea2bac3d)

Here, the rectangle and the ellipse both have rx=4, ry=2

```<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><rect x="1" y="9" width="10" height="6" rx="4" ry="2" fill="#fca"/><ellipse cx="5" cy="11" rx="4" ry="2" fill="#caf" fill-opacity="0.4"/></svg>```

This is counterpoint to the SVG specification:

![image](https://github.com/thorvg/thorvg/assets/85438892/3bb9c1c7-1846-4d38-88aa-fb48219d1c70)

The path description stated here does overlap perfectly with an ellipse:

![image](https://github.com/thorvg/thorvg/assets/85438892/db559801-921b-4d36-b277-2690c866befa)

hermet:
@MewPurPur Hello, thanks for report. Hopefully, that issue has been fixed by this: https://github.com/thorvg/thorvg/pull/1825

![image](https://github.com/thorvg/thorvg/assets/3711518/e58013ee-780d-4573-bf4b-0134e55daf06)

hermet:
@MewPurPur @capnm thorvg 0.11.6 with this fix, will be released at the end of this week. Thanks.

PR #1825, opened 2023-12-09
renderer/shape: Apply the magic number kappa to achieve rounded corners.

hermet:
The magic number kappa (0.552284), which is associated with the bezier curve, has been introduced. This formula is supposed to be applied to the rounded corners of the rectangle to ensure consistent drawing results.

Issue: https://github.com/thorvg/thorvg/issues/1824


capnm:
The  magic did it :magic_wand:. 
LGTM.

editor_screenshot_2023-12-09T135644

PR #1832, opened 2023-12-13
gl_engine: optimize framebuffer creation and save some runtime memory

RuiwenTang:
This PR changed the GL fbo creation logical to save much more runtime memory while keep the rendering result correct.

For the same example: **MaskingMethods**

### Before, need create framebuffer 80 times:
before

After this changes, only need create 3 times:
after


hermet (approved):

hermet:
@SergeyLebedkin ping?

Issue #1836, opened 2023-12-14
Expression of data smaller than FLT_EPSILON


JSUYA:
This is a simplified example of issue #1834.
```cpp
    auto shape1 = tvg::Shape::gen();
    shape1->appendCircle(0, 0, 16, 16);    //x, y, w, h

    auto fill = tvg::LinearGradient::gen();
    fill->linear(0, 0, 289.64, 0);

    tvg::Fill::ColorStop colorStops[2];
    colorStops[0] = {0, 255, 0, 0, 255};

    fill->colorStops(colorStops, 1);

    shape1->fill(std::move(fill));
    shape1->scale(10);
    if (canvas->push(std::move(shape1)) != tvg::Result::Success) return;
```

In this case, the value of linear.len with the transformation applied in _prepareLinear() (tvgSwFill.cpp) becomes smaller than FLT_EPSILON.

```cpp
if (isTransformation) {
    Matrix invTransform;
    if (!mathInverse(&gradTransform, &invTransform)) return false;

    fill->linear.offset += fill->linear.dx * invTransform.e13 + fill->linear.dy * invTransform.e23;

    auto dx = fill->linear.dx;
    fill->linear.dx = dx * invTransform.e11 + fill->linear.dy * invTransform.e21;
    fill->linear.dy = dx * invTransform.e12 + fill->linear.dy * invTransform.e22;

    fill->linear.len = fill->linear.dx * fill->linear.dx + fill->linear.dy * fill->linear.dy;
+   printf("fill->linear.len : %0.20f\n", FLT_EPSILON);
+   printf("fill->linear.len : %0.20f\n", fill->linear.len);

```

result
```
fill->linear.len : 0.00000011920928955078
fill->linear.len : 0.00000011920180753577
```

So it will not be drawn in _rasterLinearGradientRle().

hermet:
It doesn't draw anything by:
![image](https://github.com/thorvg/thorvg/assets/3711518/b36da8fe-a2c9-429f-b06c-60ace2cc5a98)

hermet:
Please reopen it if I miss. Thanks.

PR #1837, opened 2023-12-14
sw_engine: Fix wrong return

JSUYA:
Checking for values smaller than FLT_EPSILON should return false.

### DIFF
--- a/src/renderer/sw_engine/tvgSwFill.cpp
+++ b/src/renderer/sw_engine/tvgSwFill.cpp
@@ -140,7 +140,7 @@ bool _prepareLinear(SwFill* fill, const LinearGradient* linear, const Matrix* tr
     fill->linear.dy = y2 - y1;
     fill->linear.len = fill->linear.dx * fill->linear.dx + fill->linear.dy * fill->linear.dy;
 
-    if (fill->linear.len < FLT_EPSILON) return true;
+    if (fill->linear.len < FLT_EPSILON) return false;
 
     fill->linear.dx /= fill->linear.len;
     fill->linear.dy /= fill->linear.len;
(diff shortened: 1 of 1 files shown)


hermet:
@JSUYA hello, any issue of this?

JSUYA:
Sorry for not enough explanation.
This PR is related to https://github.com/thorvg/thorvg/issues/1836. But it doesn't solve it.

`fill->linear.len` value of the gradient is determined and there are several if checks to check this.
If false was being returned normally here, we would have found sooner than the rasterizer step that `len` value is less than 
And as a result, because the renderer does not draw related shapes, I expected that this would be `more strict check` as shapes in unsupported ranges would not be drawn.

I checked SVG example and didn't find any issues.
(In fact, it will be very difficult to find....)


+
As I was writing this comment..  I think it could also seem like another bug.
If you think this is not appropriate, I'll close it. Thank you for your review.

SergeyLebedkin (approved):

hermet:
Let's keep this for now. We need an obvisus a bug fix by this.

JSUYA:
> Let's keep this for now. We need an obvisus a bug fix by this.

There are some advantages, but the problems associated with that bug will likely persist unless we change the float type to a double type. (I'm not sure if there's a better way for fix.)

hermet:
@JSUYA I want to make it sure this would not occur any side effects. So unless any obivous advantage, I'm hesitage to correct the code. Please let me know If those advantages are obvious.

JSUYA:
> @JSUYA I want to make it sure this would not occur any side effects. So unless any obivous advantage, I'm hesitage to correct the code. Please let me know If those advantages you have.

@hermet Oh I meant float type's advantage. (memory & performance).

hermet:
> > @JSUYA I want to make it sure this would not occur any side effects. So unless any obivous advantage, I'm hesitage to correct the code. Please let me know If those advantages you have.
> 
> @hermet Oh I meant float type's advantage. Could converting to SwCoord be a solution?

No, it wouldn't. let me see it...

hermet:
This appears to be intentional code for distinguishing between drawing and filling operations. Unfilled shapes might be used for composition purposes.

PR #1845, opened 2023-12-18
wg_engine: refactor render targets handling

SergeyLebedkin:
For further development of features, we need to create off-screen buffers that will allow us to implement functionality related to composition and blending, as well as for loading data to system memory from the framebuffer. Separating the framebuffer into a separate entity allows you to create several instances of them, switch between them, and blend them according to given rules.

For current time we have only a single render target instance, that have a handle to drawing into surface surface, like a native window. 

New approach allows:
- offscreen rendering
- render pass handling
- switching between render targets
- ability to render images, strokes and shapes into independent render targets

### DIFF
--- a/src/renderer/wg_engine/meson.build
+++ b/src/renderer/wg_engine/meson.build
@@ -5,6 +5,7 @@ source_file = [
     'tvgWgPipelines.h',
     'tvgWgRenderData.h',
     'tvgWgRenderer.h',
+    'tvgWgRenderTarget.h',
     'tvgWgShaderSrc.h',
     'tvgWgShaderTypes.h'
     'tvgWgBindGroups.cpp',
@@ -13,6 +14,7 @@ source_file = [
     'tvgWgPipelines.cpp',
     'tvgWgRenderData.cpp',
     'tvgWgRenderer.cpp',
+    'tvgWgRenderTarget.cpp',
     'tvgWgShaderSrc.cpp',
     'tvgWgShaderTypes.cpp'
 ]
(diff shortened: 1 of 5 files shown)


hermet (changes requested):
@SergeyLebedkin Hello, I'd like to confirm the purpose of this patch. Is it intended for refactoring or feature enhancement? "Introducing a new functionality sounds like a feature enhancement, not refactoring. Could you also provide more details in the commit message? It would be helpful if you mentioned the specific purpose of the patch. For example, "why did you make changes."

hermet (approved):
Super clear.

src/renderer/wg_engine/meson.build
```diff
@@ -5,6 +5,7 @@ source_file = [
     'tvgWgPipelines.h',
     'tvgWgRenderData.h',
     'tvgWgRenderer.h',
+    'tvgWgRenderTarget.h',
```
hermet, line 8: alphabetical order?

src/renderer/wg_engine/meson.build
```diff
@@ -13,6 +14,7 @@ source_file = [
     'tvgWgPipelines.cpp',
     'tvgWgRenderData.cpp',
     'tvgWgRenderer.cpp',
+    'tvgWgRenderTarget.cpp',
```
hermet, line 17: alphabetical order?

src/renderer/wg_engine/meson.build
SergeyLebedkin: Hm... Yes

src/renderer/wg_engine/meson.build
SergeyLebedkin: > alphabetical order?
Yes

SergeyLebedkin: image

Issue #1848, opened 2023-12-19
Path data not invalid even though it doesn't start with MoveTo.


MewPurPur:
https://www.w3.org/TR/SVG2/paths.html#PathDataMovetoCommands

![image](https://github.com/thorvg/thorvg/assets/85438892/af7fa5b4-e1f7-4e6c-83ea-2ded9e93fa76)

So if we don't start with a MoveTo command, everything following should be treated as invalid. But ThorVG tries to do something with it nonetheless (see the white shape):

![image](https://github.com/thorvg/thorvg/assets/85438892/afe33ec2-9220-417c-8e05-f69aed2d3966)

`<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path d="c2 7 2 14 10 5" fill="#fff"/></svg>`

JSUYA:
Hi @hermet 

When using PathCommand (also Path APIs), there is a difference between drawing with {CubicTo, Close} and drawing with {MoveTo, CubicTo, Close}.
If CubicTo does not have an initial point (the previous last point(current point, ctrl point)), it may be decided invalid or an arbitrary initial point of 0,0 can be added internally. However, adding internally may differ from the user's intention.

In order to fix this issue, SVG can add logic that does not draw paths if the path does not start with MoveTo.
However, since the PathCommand rule must come first, please let me know your opinion.

Note: The rules(spec) for PathCommand do not strictly follow the SVG specification.

{CubicTo, Close} case
```cpp
    tvg::PathCommand cmds2[2];
    cmds2[0] = tvg::PathCommand::CubicTo;
    cmds2[1] = tvg::PathCommand::Close;

    //CubicTo 1
    tvg::Point pts2[3];
    pts2[0] = {1, 1.5};      //Ctrl1
    pts2[1] = {1.5, 4.5};      //Ctrl2
    pts2[2] = {9.5, 1};                   //To
    auto shape2 = tvg::Shape::gen();
    shape2->appendPath(cmds2, 2, pts2, 3);     //copy path data
    shape2->fill(255, 255, 0);
    shape2->scale(50);
    if (canvas->push(std::move(shape2)) != tvg::Result::Success) return;
```

{MoveTo, CubicTo, Close} case
```cpp
    tvg::PathCommand cmds2[3];
    cmds2[0] = tvg::PathCommand::MoveTo;
    cmds2[1] = tvg::PathCommand::CubicTo;
[...]

PR #1853, opened 2023-12-20
loader/svg: optimize the path conversion.

hermet:
Push the path data into a given shape directly.
This will skip unnecessary memory copies.

https://github.com/thorvg/thorvg/issues/1848

### DIFF
--- a/src/loaders/svg/tvgSvgPath.cpp
+++ b/src/loaders/svg/tvgSvgPath.cpp
@@ -53,6 +53,7 @@
 #include <cstring>
 #include <math.h>
 #include <ctype.h>
+#include "tvgShape.h"
 #include "tvgSvgLoaderCommon.h"
 #include "tvgSvgPath.h"
 #include "tvgStr.h"
@@ -534,7 +535,7 @@ static char* _nextCommand(char* path, char* cmd, float* arr, int* count)
 /************************************************************************/
 
 
-bool svgPathToTvgPath(const char* svgPath, Array<PathCommand>& cmds, Array<Point>& pts)
+bool svgPathToShape(const char* svgPath, Shape* shape)
 {
     float numberArray[7];
     int numberCount = 0;
@@ -545,11 +546,16 @@ bool svgPathToTvgPath(const char* svgPath, Array<PathCommand>& cmds, Array<Point
     bool isQuadratic = false;
     char* path = (char*)svgPath;
 
+    auto& pts = P(shape)->rs.path.pts;
+    auto& cmds = P(shape)->rs.path.cmds;
+    auto lastCmds = cmds.count;
+
     while ((path[0] != '\0')) {
         path = _nextCommand(path, &cmd, numberArray, &numberCount);
         if (!path) break;
         if (!_processCommand(&cmds, &pts, cmd, numberArray, numberCount, &cur, &curCtl, &startPoint, &isQuadratic)) break;
     }
 
+    if (cmds.count > lastCmds && cmds[lastCmds] != PathCommand::MoveTo) return false;
     return true;
 }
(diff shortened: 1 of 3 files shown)


hermet:
@JSUYA After considering a better approach, this would likely improve performance and issue handling.

SergeyLebedkin (approved):

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -396,13 +396,9 @@ static bool _recognizeShape(SvgNode* node, Shape* shape)
...
         case SvgNodeType::Path: {
             if (node->node.path.path) {
-                Array<PathCommand> cmds;
-                Array<Point> pts;
-                if (svgPathToTvgPath(node->node.path.path, cmds, pts)) {
-                    if (shape->appendPath(cmds.data, cmds.count, pts.data, pts.count) != Result::Success) {
-                        TVGERR("SVG", "Invalid path information.");
-                        return false;
-                    }
+                if (!svgPathToShape(node->node.path.path, shape)) {
```
JSUYA, line 399: tvgSvgPath only accessed data related to Path. This way causes unnecessary includes.
I think below is better than modifying the function parameters. what do you think?

```suggestion
auto& pts = P(shape)->rs.path.pts;
auto& cmds = P(shape)->rs.path.cmds;
   if (!svgPathToTvgPath(node->node.path.path, cmds, pts)) {
...

**Review by JSUYA — COMMENTED** [review]
I had no idea that copy could be reduced. That's a good idea.
But I left few comments. Please check it.

File: src/loaders/svg/tvgSvgPath.cpp [hunk 2]
```diff
@@ -545,11 +546,16 @@ bool svgPathToTvgPath(const char* svgPath, Array<PathCommand>& cmds, Array<Point
...
+    auto& cmds = P(shape)->rs.path.cmds;
+    auto lastCmds = cmds.count;
+
     while ((path[0] != '\0')) {
         path = _nextCommand(path, &cmd, numberArray, &numberCount);
         if (!path) break;
         if (!_processCommand(&cmds, &pts, cmd, numberArray, numberCount, &cur, &curCtl, &startPoint, &isQuadratic)) break;
     }
 
+    if (cmds.count > lastCmds && cmds[lastCmds] != PathCommand::MoveTo) return false;
```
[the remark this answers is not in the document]
hermet: Yes, we talked that. But it would not work in this optmization.

src/loaders/svg/tvgSvgSceneBuilder.cpp
hermet: Could you please suggest the beneficial point? Either way it will be a thorvg specific internal function. For now, less parameter, simpler function prototypes are better I think. or probably we can remove that function since nowhere uses it.

hermet: Also, at least now, the helper function guides you to pass the Shape so that it will optimally append the data in the under the hood. Otherwise someone might call it again with the previous way with an unnecessary memory copy.

JSUYA: > Could you please suggest the beneficial point?

1. No unnecessary header inclusion.(`#include "tvgShape.h"`) And appendPath() API is not used, memory copy does not occur.
2. `tvgSvgPath.cpp` code is basically responsible for translating data related to SVG Path. So we need is a Path String and the PathComand we passed it to. (I think this is why we separated this function, which was only used here, into a separate file.)
3. I think that having access to Shape objects is beyond the scope of tvgSvgPath's role.
 
> Otherwise someone might call it again with the previous way - an unnecessary memory copy.

Even if memory copy occurs when someone uses this API, it is because they use appendPath and not because of this API.

Anyway, I am not against to applying this patch because it does not have any functional problems.
I tested several expected paths and they work without problem.

JSUYA (approved):

src/loaders/svg/tvgSvgSceneBuilder.cpp
hermet: Ok, if you find out any real problems then please revise the code yourself. Otherwise let's keep it simple and easy code.

Issue #1519, opened 2023-07-06
About coding style, do we need a clang-format configuration


RuiwenTang:
As mentioned in the contribution guidelines, the coding style check must passed. 
So I think a clang-format file with specified style config is helpful.

By the way, I tried the code_format.sh script, but the result seems not consistent with the existing code

capnm:
++ EOL issues. I see that by converting and commiting the ThorVG code to the Godot github repository.
See https://github.com/godotengine/godot/blob/master/.gitattributes

hermet:
I agree though I don't like to change the tvg coding style at this point since it will hurts the usability of the code history tracking like git blame info...  

code_format.sh has been experimentally added in the past, we need to improve it to work properly or renewal the official tvg codiing style at the 1.0 release with a more popular one? I just ignore this task since it's annoying to me so far. A volunteer is welcomed. :-)

capnm:
> I agree though I don't like to change the tvg coding style at this point since it will hurts the usability of the code history tracking like git blame info...

Yes, reformatting code also hides usable line-blame annotations in code editors. Maybe just keep some visual consistency that is human readable;)

lmdsp:
I noticed this style pattern is used mostly everywhere for conditions:

Result draw()
{
    if (drawing || paints.empty() || !renderer || !renderer->preRender()) return Result::InsufficientCondition;

    bool rendered = false;
    for (auto paint : paints) {
        if (paint->pImpl->render(renderer)) rendered = true;
    }

    if (!rendered || !renderer->postRender()) return Result::InsufficientCondition;

    drawing = true;

    return Result::Success;
}

This makes it impossible to set breakpoints for debugging.
May I suggest doing something like this instead

Result draw()
{
    if (drawing || paints.empty() || !renderer || !renderer->preRender())
    {
       return Result::InsufficientCondition; // breakpoint can be set here
    }

    bool rendered = false;
    for (auto paint : paints) {
        if (paint->pImpl->render(renderer))
        {
          rendered = true;
        }
    }

    if (!rendered || !renderer->postRender())
    {
      return Result::InsufficientCondition; // breakpoint can be set here
    }

    drawing = true;

    return Result::Success;
}

PR #1857, opened 2023-12-22
wg_engine: shape bbox based rendering (optimization)

SergeyLebedkin:
Before the current changes, all surfaces were painted using a full-screen overlay, no matter how large the object was rendered. This approach is redundant and required reorganization. At the moment, all objects are rendered using an overlay equal to the box of the object itself, which reduces the cost of filling the surface. 
Also surfaces and images were divided into different entities, which reduces the pressure on memory. 
Also geometry data for rendering and geometry data for calculations in system memory were logically separated.

### DIFF
--- a/src/renderer/wg_engine/tvgWgCommon.cpp
+++ b/src/renderer/wg_engine/tvgWgCommon.cpp
@@ -90,6 +90,7 @@ void WgContext::initialize()
     assert(queue);
 }
 
+
 void WgContext::release()
 {
     if (device) {
@@ -99,6 +100,7 @@ void WgContext::release()
     if (adapter) wgpuAdapterRelease(adapter);
     if (instance) wgpuInstanceRelease(instance);
 }
+
 
 void WgContext::executeCommandEncoder(WGPUCommandEncoder commandEncoder)
 {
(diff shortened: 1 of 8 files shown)


src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -55,144 +52,126 @@ void WgRenderer::release()
...
     return true;
 }
 
-
 bool WgRenderer::blend(TVG_UNUSED BlendMethod method)
 {
     return false;
 }
 
-
```
hermet, line 183: Please try to keep the coding consistency as we discussed it.

hermet (changes requested):

hermet:
see also: https://github.com/thorvg/thorvg/issues/1519

src/renderer/wg_engine/tvgWgCommon.cpp
```diff
@@ -90,6 +90,7 @@ void WgContext::initialize()
     assert(queue);
 }
 
+
```
hermet, line 93: --

hermet (changes requested):

hermet (approved):

PR #1865, opened 2023-12-27
api: promote to official APIs.

hermet:
New API:
- SwCanvas::Colorspace::ABGR8888S
- SwCanvas::ColorSpace::ARGB8888S

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -212,7 +212,7 @@ enum class BlendMethod : uint8_t
  */
 enum class CanvasEngine : uint8_t
 {
-    All = 0,       ///< All feasible rasterizers.
+    All = 0,       ///< All feasible rasterizers. @since 1.0
     Sw = (1 << 1), ///< CPU rasterizer.
     Gl = (1 << 2), ///< OpenGL rasterizer.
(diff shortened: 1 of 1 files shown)


hermet:
Note: CanvasEngine::All will not be in v0.12.x

inc/thorvg.h
```diff
@@ -212,7 +212,7 @@ enum class BlendMethod : uint8_t
  */
 enum class CanvasEngine : uint8_t
 {
-    All = 0,       ///< All feasible rasterizers.
+    All = 0,       ///< All feasible rasterizers. @since 1.0
     Sw = (1 << 1), ///< CPU rasterizer.
     Gl = (1 << 2), ///< OpenGL rasterizer.
```
RuiwenTang, line 217: does GL backend needs to be marked as experimetal API ?

hermet: @RuiwenTang It's been already there from the beginning. Instead, we should not change the enum value.  GlCanvas has that experimental tag so It think it wouldn't be a problem.

hermet: @RuiwenTang fyi, "All" was newly added since 1.11.

SergeyLebedkin (approved):

tinyjin (approved):

Issue #1882, opened 2024-01-02
web/lottie-player: code consistency improvement


tinyjin:
1. missing '=' in a condition
```ts
if (this.currentState == PlayerState.Playing) {
// + if (this.currentState === PlayerState.Playing) {
```
issued by @Aidosmf from [comment](https://github.com/thorvg/thorvg/pull/1817#discussion_r1432248897)

2. missing `@param`
```ts
/**
   * Configure and load
   * @since 1.0
 */
public async load(src: string | object): Promise<void> {
  ..
}
```

3. align string quotes to `'`
```ts
// good
export enum ExportableType {
  GIF = 'gif',
  TVG = 'tvg',
}

// bad
export enum PlayMode {
  Bounce = "bounce",
  Normal = "normal",
}
```

4. wrong explanation: `LottieFiles` > `Lottie animation`
```ts
 /**
  * LottieFiles JSON data or URL to JSON.
  * @since 1.0
  */
```

PR #1890, opened 2024-01-04
web: extra features & minor bug fixes

tinyjin:
Updated features & props as ThorVG Viewer requirements

Added:
- `size` - `TVG.size()` wrapper
- `resize()` - method for dynamic resizing
- Support `save('png')`

Fixed:
- animation loading bug when <lottie-player> doesn't have `src` prop
- animation stop behavior -> cursor 0 frame when stopped

### DIFF
--- a/web/src/lottie-player.ts
+++ b/web/src/lottie-player.ts
@@ -252,6 +252,15 @@ export class LottiePlayer extends LitElement {
   @property({ type: Number })
   public currentState: PlayerState = PlayerState.Loading;
 
+  /**
+   * original size of the animation (readonly)
+   * @since 1.0
+   */
+  @property({ type: Float32Array })
+  public get size(): Float32Array {
+    return Float32Array.from(this._TVG?.size() || [0, 0]);
+  }
+
   private _TVG?: TvgModule;
   private _canvas?: HTMLCanvasElement;
   private _imageData?: ImageData;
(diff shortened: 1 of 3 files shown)


web/src/lottie-player.ts
```diff
@@ -252,6 +252,15 @@ export class LottiePlayer extends LitElement {
   @property({ type: Number })
   public currentState: PlayerState = PlayerState.Loading;
 
+  /**
+   * current resolution of animation (readonly)
+   * @since 1.0
+   */
+  @property({ type: Float32Array })
+  public get resolution(): Float32Array {
```
hermet, line 270: 1. Is it really necessary? If so,
2. I think it requires a consistent naming convention, such as 'size()' since the input function is 'resize()'.

hermet (commented):
Reviewed "web: Add extra feature & property"

web/src/lottie-player.ts
tinyjin: > 1. Is it really necessary? If so,
think so, Viewer needs to show original size (not canvas size)

Screenshot 2024-01-04 at 4 39 55 PM

tinyjin: > 2. I think it requires a consistent naming convention, such as 'size()' since the input function is 'resize()'

gotcha, will change to `size()`

web/src/lottie-player.ts
```diff
@@ -252,6 +252,15 @@ export class LottiePlayer extends LitElement {
   @property({ type: Number })
   public currentState: PlayerState = PlayerState.Loading;
 
+  /**
+   * original size of animation (readonly)
```
hermet, line 266: -> original size of the animation (readonly)

web/src/lottie-player.ts
```diff
@@ -491,6 +500,21 @@ export class LottiePlayer extends LitElement {
     this._render();
   }
 
+  /**
+   * Resize canvas size
```
hermet, line 515: -> Adjust the canvas size.

hermet (commented):
polishing, doc.

hermet:
For your information, some descriptions start with an uppercase letter, while others do not. @_@

tinyjin:
> For your information, some descriptions start with an uppercase letter, while others do not. @_@

Ghaha,,, that's my fault, I'll add to #1882

hermet (approved):

PR #1895, opened 2024-01-08
ttf: Fixed an invalid unicode encoding.

hermet:
Ensured the data count is correctly multiplied by the data size.

### DIFF
--- a/src/loaders/ttf/tvgTtfLoader.cpp
+++ b/src/loaders/ttf/tvgTtfLoader.cpp
@@ -125,7 +125,7 @@ static uint32_t* _codepoints(const char* text, size_t n)
 
     auto utf8 = text;
     //preserve approximate enough space.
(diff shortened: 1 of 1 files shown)


src/loaders/ttf/tvgTtfLoader.cpp
```diff
@@ -148,7 +148,7 @@ static uint32_t* _codepoints(const char* text, size_t n)
             return nullptr;
         }
     }
-    utf32[i - 1] = 0;   //end of the unicdoe
+    utf32[i] = 0;   //end of the unicdoe
```
capnm, line 151: // End of Unicode.

hermet: Thanks, your contribution will be very helpful for ThorVG!

PR #1896, opened 2024-01-09
improvement for the lottie2gif functionality

tinyjin:
Usage example:
```cpp
uint32_t size = 0;
auto data = P(animation->picture())->data(size);
```

### DIFF
--- a/src/renderer/tvgPicture.h
+++ b/src/renderer/tvgPicture.h
@@ -81,7 +81,7 @@ struct Picture::Impl
 
     ~Impl()
     {
-        LoaderMgr::retrieve(loader);
+        if (LoaderMgr::retrieve(loader))
         delete(paint);
     }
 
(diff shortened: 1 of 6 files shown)


src/renderer/tvgPicture.h
```diff
@@ -241,6 +241,11 @@ struct Picture::Impl
         else return nullptr;
     }
 
+    uint32_t* data(uint32_t& size)
+    {
```
hermet, line 245: please add nullcheck for loader

src/renderer/tvgPicture.h
```diff
@@ -241,6 +241,11 @@ struct Picture::Impl
         else return nullptr;
     }
 
+    uint32_t* data(uint32_t& size)
```
hermet, line 244: since the input is const char*, let's change the return type  to  "const char*"

hermet (commented):
please check comments

hermet:
https://github.com/thorvg/thorvg/pull/2017

PR #1898, opened 2024-01-09
fix the prototype of mpoolInit

kisvegabor:
Hi,

While I was trying to build ThorVG for an STM32 MCU I found that in `tvgSwCommon.h` `mpoolInit` has `uint32_t` parameter while in `tvgSwCommon.cpp` it's `unsigned`. The difference caused link error on a 32 bit MCU.


kisvegabor:
I found an other mismatch: [9566244](https://github.com/thorvg/thorvg/pull/1898/commits/9566244f461560db6b384a63240b8c7610fdb1c5)

hermet (approved):

SergeyLebedkin (approved):

Issue #1900, opened 2024-01-10
Support pure sync mode without mutexes and tasks


kisvegabor:
When I built ThorVG on STM32 it turned out the `mutex` and some other task related C++ modules are not implemented there. I haven't tried, but I suspect that it's true for other vendors/platforms too.

As we are using ThorVG only in sync mode I removed all the mutex and task related parts from `tvgTaskScheduler.cpp` and `tvgTaskScheduler.h`. See the changes here https://github.com/lvgl/lvgl/pull/5249

Now it works, but I wonder if it would make sense to allow enabling a simple "sync only" mode with a define.

What do you think?

hermet:
@kisvegabor Hello, I agree with your opinion. We can remove the threading feature from the compilation on the end side. Please let me consider making it compilable for both scenarios. I will come up with the solution.

kisvegabor:
Thank you! :blush: 

We copy pasted the source code ThorVG into LVGL and modified it a little bit. Therefore for us the best would be to have a define in a config file we can alter this behavior.

PR #1911, opened 2024-01-15
common: STM32 portability enhancement

hermet:
Some systems such as micro-processor might not support the thread feature on the system.

Enhance the portability by compiling the thorvg with toggling the threading depepdency through the build option.

For this, thorvg newly introduced the internal Key/ScopedLock abstraction for transparent thread-locking dependnecy.

To turn off the thread feature, please use the next build option:

$meson setup build -Dthreads=false ...

Note that the thread feature is enabled in default.
Turning off the thread feature could reduce the binary size by 7kb.

issue: https://github.com/thorvg/thorvg/issues/1900

### DIFF
--- a/src/renderer/tvgTaskScheduler.h
+++ b/src/renderer/tvgTaskScheduler.h
@@ -25,22 +25,13 @@
 
 #include <mutex>
 #include <condition_variable>
+
 #include "tvgCommon.h"
 #include "tvgInlist.h"
 
-namespace tvg
-{
+namespace tvg {
 
-struct Task;
-
-struct TaskScheduler
-{
-    static unsigned threads();
-    static void init(unsigned threads);
-    static void term();
-    static void request(Task* task);
-    static void async(bool on);
-};
+#ifdef THORVG_THREAD_SUPPORT
 
 struct Task
 {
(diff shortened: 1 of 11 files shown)


kisvegabor:
Hi,

Thank you for the prompt action! :100: 

If we are using a managed build can we manually add `#define THORVG_THREAD_SUPPORT 0` to `config.h`?

hermet:
Turning off the thread feature could reduce the binary size by 7kb.

> #define THORVG_THREAD_SUPPORT 0

@kisvegabor Hi, I think you can do that. Alternatively, you can just ignore the option declaration.

src/renderer/tvgTaskScheduler.h
```diff
@@ -86,7 +77,36 @@ struct Task
...
+private:
+    friend struct TaskSchedulerImpl;
+};
+
+#endif  //THORVG_THREAD_SUPPORT
+
+
+struct TaskScheduler
+{
+    static unsigned threads();
```
kisvegabor, line 102:
```suggestion
    static uint32_t threads();
```

kisvegabor (commented):
I've tested it on STM32 and it works, except the minor return type mismatch.

kisvegabor (approved):

kisvegabor:
By the way, how could you request review from me? At LVGL I can do it only for people with write access.

hermet:
@kisvegabor Well... If I'm correct, all registered project members(at least above the triage) on github have permissions to review/approve PRs though only the members with write access could allow merging them.

kisvegabor:
Ah I see. I've tried to check my role, but couldn't find it anywhere. I assume it's "Triage" but seemingly it's not high enough to check my own role :smile: 
Thank you anyway!

PR #1922, opened 2024-01-18
wg_engine: introduced compute pipeline entities

SergeyLebedkin:
introduces posibility to create compute pipelines
does not affect functionality


src/renderer/wg_engine/tvgWgBindGroups.cpp
```diff
@@ -334,49 +409,43 @@ void WgBindGroupBlit::initialize(WGPUDevice device, WGPUQueue queue, WGPUSampler
...
-        bindGroup->update(context.device, context.queue, opacity);
-        mUsed.push(bindGroup);
-        mPool.pop();
-    }
-    return bindGroup;
+void WgBindGroupOpacityPool::initialize(WgContext& context)
+{
+    release(context);
+    mPool.reserve(256);
+    for (uint32_t i = 0; i < mPool.reserved; i++)
```
hermet, line 425: This loop looks unnecessary or probably this is better?
```
memset(mPool.data, 0x00, sizeof(WgBindGroupOpacity*) * 256);
```

src/renderer/wg_engine/tvgWgBindGroups.cpp
```diff
@@ -334,49 +409,40 @@ void WgBindGroupBlit::initialize(WGPUDevice device, WGPUQueue queue, WGPUSampler
...
-    } else {
-        bindGroup = mPool.last();
-        bindGroup->update(context.device, context.queue, opacity);
-        mUsed.push(bindGroup);
-        mPool.pop();
-    }
-    return bindGroup;
+void WgBindGroupOpacityPool::initialize(WgContext& context)
+{
+    memset(mPool, 0x00, sizeof(mPool));
```
hermet, line 423: hmm this looks wrong... you might wanna this:
```
mPool.reserve(256);
memset(mPool.data, 0x00, sizeof(WgBindGroupOpacity*) * 256);
```

hermet (changes requested):

src/renderer/wg_engine/tvgWgBindGroups.cpp
SergeyLebedkin: No, now we are using simple array, without any wrapper.
 `WgBindGroupOpacity* mPool[256];`

hermet (commented):
I'm really not certain if caching every single alpha value is allowed. Instead of preparing the binding group for every single opacity value, can't the opacity value be a parameter? This looks over the top. Or it would be nice If you can suggest how the memory usage in the max can be differentiated.

SergeyLebedkin:
> I'm really not certain if caching every single alpha value is allowed. Instead of preparing the binding group for every single opacity value, can't the opacity value be a parameter? This looks over the top.

No, it can`t be... any parameter must be a binding group, and must be alive and unic all the time before command buffer will be executed (WGPU have "postponed" architecture). But for each opacity value bind group will be created only once in the moment of first reaching (RAII) and will be reusable for the next access with the same opacity value.

hermet:
the b

> > I'm really not certain if caching every single alpha value is allowed. Instead of preparing the binding group for every single opacity value, can't the opacity value be a parameter? This looks over the top.
> 
> No, it can`t be... any parameter must be a binding group, and must be alive and unic all the time before command buffer will be executed (WGPU have "postponed" architecture). But for each opacity value bind group will be created only once in the moment of first reaching (RAII) and will be reusable for the next access with the same opacity value.



> > I'm really not certain if caching every single alpha value is allowed. Instead of preparing the binding group for every single opacity value, can't the opacity value be a parameter? This looks over the top.
> 
> No, it can`t be... any parameter must be a binding group, and must be alive and unic all the time before command buffer will be executed (WGPU have "postponed" architecture). But for each opacity value bind group will be created only once in the moment of first reaching (RAII) and will be reusable for the next access with the same opacity value.

It looks caching the bind group for every opacity value is overly aggressive and inefficient. How about check the used memory size briefly when the maximum binding group is allocated?

SergeyLebedkin:
> It looks caching the bind group for every opacity value is overly aggressive and inefficient. How about check the used memory size briefly when the maximum binding group is allocated?

Initialy we allocate only 2KB of memory (in 64-bit arcs) for a pointer array. Bind groups will be allocated only by request. And I believ that in animation process they all be used f.e. in case of fading by tranparensy. And its expencive not by memoty usage, but by each BindGroup allocations time on GPU side. So thats why we need to create instances as earlier as we can

hermet:
> > It looks caching the bind group for every opacity value is overly aggressive and inefficient. How about check the used memory size briefly when the maximum binding group is allocated?
> 
> Initialy we allocate only 2KB of memory (in 64-bit archs) for a pointers array. Bind groups will be allocated only by request. And I believ that in animation process they all be used f.e. in case of fading by tranparensy. And its expencive **not by memory usage**, but **by allocations time** for each BindGroup on the GPU side. So thats why we need to create instances as earlier as we can and keep them alive as long, as we can

1. Consider a scenario involving a single animation shot that transitions in opacity from 255 to 0. This process could be burdensome due to the need for initializing the opacity binding group for every frame.
2. You have already implemented the opacity binding group with an opacity parameter, which eliminates the need for this pooling overhead. I'm curious about how this refactoring improves upon the previous approach.

hermet (approved):

Issue #1937, opened 2024-01-26
SVG: rx/ry equal to 0 for elliptical arc parameters not handled correctly.


MewPurPur:
https://www.w3.org/TR/SVG/paths.html#ArcOutOfRangeParameters

One of the SVG specifications:

> If either rx or ry is 0, then this arc is treated as a straight line segment (a "lineto") joining the endpoints.

This is not handled correctly, instead, nothing is drawn. Example of a SVG with the issue:

`<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path d="M11 8A0 2 0 005 8" fill="none" stroke="red" stroke-width="1.5" stroke-linecap="round"/></svg>`

PR #1938, opened 2024-01-26
loader/svg: Apply specification of out-of-range elliptical arc parameters

JSUYA:
The three specifications below apply.

Specification: https://www.w3.org/TR/SVG/paths.html#ArcOutOfRangeParameters

- If the endpoint (x, y) of the segment is identical to the current point (e.g., the endpoint of the previous segment),
then this is equivalent to omitting the elliptical arc segment entirely.
- If either rx or ry is 0, then this arc is treated
as a straight line segment(a "lineto") joining the endpoints.
- If either rx or ry have negative signs, these are dropped; the absolute value is used instead.

related issue : https://github.com/thorvg/thorvg/issues/1937

### DIFF
--- a/src/loaders/svg/tvgSvgPath.cpp
+++ b/src/loaders/svg/tvgSvgPath.cpp
@@ -51,7 +51,6 @@
 #define _USE_MATH_DEFINES       //Math Constants are not defined in Standard C/C++.
 
 #include <cstring>
-#include <math.h>
 #include <ctype.h>
 #include "tvgMath.h"
 #include "tvgShape.h"
@@ -472,13 +471,12 @@ static bool _processCommand(Array<PathCommand>* cmds, Array<Point>* pts, char cm
         }
         case 'a':
         case 'A': {
-            if (arr[0] < FLT_EPSILON || arr[1] < FLT_EPSILON) {
+            if (mathZero(arr[0]) || mathZero(arr[1])) {
                 Point p = {arr[5], arr[6]};
                 cmds->push(PathCommand::LineTo);
                 pts->push(p);
                 *cur = {arr[5], arr[6]};
-            }
-            else if (!mathEqual(cur->x, arr[5]) || !mathEqual(cur->y, arr[6])) {
+            } else if (!mathEqual(cur->x, arr[5]) || !mathEqual(cur->y, arr[6])) {
                 _pathAppendArcTo(cmds, pts, cur, curCtl, arr[5], arr[6], fabsf(arr[0]), fabsf(arr[1]), arr[2], arr[3], arr[4]);
                 *cur = *curCtl = {arr[5], arr[6]};
                 *isQuadratic = false;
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgPath.cpp
```diff
@@ -471,9 +472,17 @@ static bool _processCommand(Array<PathCommand>* cmds, Array<Point>* pts, char cm
...
         case 'a':
         case 'A': {
-            _pathAppendArcTo(cmds, pts, cur, curCtl, arr[5], arr[6], arr[0], arr[1], arr[2], arr[3], arr[4]);
-            *cur = *curCtl = {arr[5], arr[6]};
-            *isQuadratic = false;
+            if (arr[0] < FLT_EPSILON || arr[1] < FLT_EPSILON) {
+                Point p = {arr[5], arr[6]};
+                cmds->push(PathCommand::LineTo);
+                pts->push(p);
+                *cur = {arr[5], arr[6]};
```
hermet, line 479: Neat code.
```
   *cur = p;
} else if ( ...) 
```

src/loaders/svg/tvgSvgPath.cpp
```diff
@@ -53,6 +53,7 @@
 #include <cstring>
 #include <math.h>
 #include <ctype.h>
+#include "tvgMath.h"
```
hermet, line 56: -- #include math.h
++ #include tvgMath.h

capnm:
It looks like there is still an issue with the first spec. Could you take a look at this example?

```xml
svg viewBox="0 0 120 90" xmlns="http://www.w3.org/2000/svg">

  <path d="
    M110 80
    A1 -2.5  0 0 0 50 80
    A0 2     0 0 1 100 55
    A1 2     0 0 0 100 55
    L 10 10" 

  fill="none" stroke="red" stroke-width="5" stroke-linecap="round"/>
</svg>
```

editor_screenshot_2024-01-26T214044

0_example_x10pr

src/loaders/svg/tvgSvgPath.cpp
```diff
@@ -471,9 +472,17 @@ static bool _processCommand(Array<PathCommand>* cmds, Array<Point>* pts, char cm
         }
         case 'a':
         case 'A': {
-            _pathAppendArcTo(cmds, pts, cur, curCtl, arr[5], arr[6], arr[0], arr[1], arr[2], arr[3], arr[4]);
-            *cur = *curCtl = {arr[5], arr[6]};
-            *isQuadratic = false;
+            if (arr[0] < FLT_EPSILON || arr[1] < FLT_EPSILON) {
```
capnm, line 475: Check the absolute values.

fabsf ...

JSUYA: Thank you for your review. I modified code.

hermet (approved):

PR #1948, opened 2024-02-01
examples/svg: Add target(-t) option.

JSUYA:
Add -t option to the argument of svg example.
-t can specify a svg file located within the example's resource.
You can use this when you want to check the output of just one SVG file.

ex)
$)./build/src/examples/Svg -t cartman.svg

### DIFF
--- a/src/examples/Svg.cpp
+++ b/src/examples/Svg.cpp
@@ -35,6 +35,8 @@ static int counter = 0;
 
 static std::vector<unique_ptr<tvg::Picture>> pictures;
 
+const char* targetSvgFileName = nullptr;
+
 void svgDirCallback(const char* name, const char* path, void* data)
 {
     if (counter >= NUM_PER_ROW * NUM_PER_COL) return;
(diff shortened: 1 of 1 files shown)


JSUYA:
No longer need to clean svg files in resource dir and modify NUM_PER_ROW/COL to check for issues.

hermet:
What do you think if we introduce PicturrSVG like PictureTVG. I don't really like to make examples functional.

JSUYA:
> What do you think if we introduce PicturrSVG like PictureTVG. I don't really like to make examples functional.

https://github.com/JSUYA/thorvg/commit/6338bf686765d0a9a036449e4a451d4952659e55
I made an example that draws one simple SVG(logo.svg).
I think that if we add various functions (transformation test, opacity etc.) like PicturePng to this example, inconveniences will still remain.
I still think this PR is a simple feature for contributors.

Is it okay if I add a PictureSvg that outputs 1 SVG?

hermet:
The functionality I mentioned is not actually a separate feature you mentioned(transform, opacity etc); those functions are included under the ThorVG showcases. What I was referring to as 'functional' are actually the command-line functions. These are not about thorvg and user don't need to be digressed by it. I think we can add PictureSVG as like other PictureXXX families.

JSUYA:
I made new PR. https://github.com/thorvg/thorvg/pull/1962

PR #1959, opened 2024-02-04
renderer/canvas: Optimize logic by removing paint verification.

hermet:
Allow direct updates to the paint object without prior validation. The verification process is deemed inefficient;
users are expected to ensure the paint is updated using a canvas that contains it.


hermet:
--compatiblity break.

Issue #1427, opened 2023-05-08
Support static WebP loader


hermet:
WebP is the compact and optimal image format in the industry. 
It is widely used among platforms and apps these days. Let's prioritize supporting external WebP.

https://chromium.googlesource.com/webm/libwebp/

vtorri:
jpeg-xl is far superior to webp, but less used

hermet:
Some history here:
https://www.roboleary.net/webdev/2023/03/06/next-web-image-format-not-jpegxl.html

hermet:
for static decoder: https://github.com/jacklicn/libwebp

PR #1966, opened 2024-02-06
loader/webp: Support static webp loader option

hermet:
Built to libwebp code in to support the static option of webp loader. (code from :https://github.com/jacklicn/libwebp)

This forking version only contains the only webp decoding feature.

Binary Size: +68kb

issue: https://github.com/thorvg/thorvg/issues/1427


hermet:
+69kb

JSUYA (approved):
It's nice! Thank you for your work.

PR #1967, opened 2024-02-06
gl_engine: add stencil and cover render task

RuiwenTang:
This PR contains the following changes:

* add new render task to do stencil and cover rendering which is a fallback rendering method to handle cases that trianglation tessellation failed
* add a new tessellator to generate stencil and cover vertex mesh

**After using this fallback method, more cases can be rendered successfully:**

## Before this patch
image

----

## After this patch
截屏2024-02-06 18 44 12

### DIFF
--- a/src/renderer/gl_engine/tvgGlTessellator.h
+++ b/src/renderer/gl_engine/tvgGlTessellator.h
@@ -177,6 +177,23 @@ private:
     GlPoint mPtCur;
 };
 
+class BWTessellator
+{
+public:
+    BWTessellator(Array<float>* points, Array<uint32_t>* indices);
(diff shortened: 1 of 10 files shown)


SergeyLebedkin (approved):

src/renderer/gl_engine/tvgGlTessellator.h
```diff
@@ -177,6 +177,23 @@ class DashStroke
     GlPoint mPtCur;
 };
 
+class BWTessellator
```
hermet, line 180: What is "BW"?

hermet (approved):

Issue #1808, opened 2023-11-23
Support Property Overriding Feature in Lottie


hermet:
scheme: https://github.com/LottieFiles/lottie-docs/blob/fc7d3de485b1da75a447aceeb9993eb0b525b3d1/docs/schema/animation/animation.json#L116

example:
```
{
  "v": "5.5.2",
  "fr": 60,
  "ip": 0,
  "op": 120,
  "w": 300,
  "h": 100,
  "assets": [],
  "slots": {
    "text_document_data": {
      "p": {
        "k": [
          {
            "s": {
              "f": "Ubuntu Light Italic",
              "fc": [1, 0, 1],
              "s": 50,
              "t": "Hello World",
              "j": 0
            },
            "t": 0
          }
        ]
      }
    }
  },
  "fonts": {
    "list": [
      {
        "fPath": "https://fonts.gstatic.com/s/ubuntu/v15/4iCp6KVjbNBYlgoKejZftWyI.ttf",
        "fFamily": "Ubuntu",
        "fStyle": "Light Italic",
        "fName": "Ubuntu Light Italic",
        "origin": 3
      }
    ]
  },
  "layers": [
    {
      "nm": "T",
      "ty": 5,
      "sr": 1,
      "ks": { "p": { "k": [5, 80], "a": 0 } },
      "ip": 0,
      "op": 120,
      "st": 0,
      "t": {
        "a": [],
        "d": {
          "k": [
            {
              "s": {
                "f": "Ubuntu Light Italic",
                "fc": [0, 0, 0],
                "s": 100,
                "t": "Hello",
                "j": 0
              },
              "t": 0
            }
          ],
          "sid": "text_document_data"
        },
        "m": { "a": { "k": [0, 0], "a": 0 }, "sid": "text_alignment" },
        "p": {}
      }
    }
  ]
}
```

hermet:
Here is a brief idea to introduce the slot feature.

1. Introduce the SlotParsing in LottieParser?

2. Introduce Override the properties with the given slot.

3. Introduce LottieAnimation with slot method

```
//candidate 1
class LottieAnimation extends Lottie
 - const char* addSlot(const char* slot json);  //return slot name
 - Result removeSlot(const char* slot name);
 - Result apply(const char* slot name);

//candidate 2
class Lottie Animation extends Lottie
- Result override(const char* slot json);
```


[...]

PR #1974, opened 2024-02-07
Lottie Slot Feature Support

tinyjin:
Internal model & parser changes to parse "sid" and obtain their data into the `LottieComposition`.
It will support dynamic-change following lottie objects:
- LottieSolidStroke
- LottieSolidFill
- LottieGradientStroke
- LottieGradientFill
- LottieTextDoc

Related: #1808


src/loaders/lottie/tvgLottieModel.h
```diff
@@ -456,7 +457,10 @@ struct LottieSolidFill : LottieObject
 };
 
 
-struct LottieGradientFill : LottieObject, LottieGradient
+struct LottieGradientObject : LottieObject, LottieGradient {};
```
hermet, line 460: Can't see any necessity of LottieGradientObject in this commit context. Could you please explain this?

tinyjin: In this change, `parseProperty` will have parameter `LottieObject` to parse `sid` into itself.

When system calls `parseProperty` from `parseGradient`:
```cpp
void LottieParser::parseGradient(LottieGradient* gradient, const char* key)
{
    // ...
    // Can't pass original lottie object because LottieGradient isn't child of LottieObject for this scope.
    else if (!strcmp(key, "o")) parseProperty(gradient->opacity, gradient); // [!] Type error
    // ...
}
```

Even though, `parseGradient` is called with LottieObject such as
- LottieGradientFill
- LottieGradientStroke

the parameter is being type-changed because of the explicit type of `LottieGradient`, it wasn't able to pass as `LottieObject`.


But if we declare `LottieGradientObject`, the `LottieObject` will be correctly passed.
```cpp
// the `gradient` is matched the `LottieObject`
void LottieParser::parseGradient(LottieGradientObject* gradient, const char* key)

// template<typename T> void parseProperty(T& prop, LottieObject* parent)
```

hermet (commented):
Please check my comments.

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -618,6 +639,7 @@ struct LottieComposition
     Array<LottieObject*> assets;
     Array<LottieInterpolator*> interpolators;
     Array<LottieFont*> fonts;
+    Array<LottieSlot*> slots;
```
hermet, line 642: the slots in the array should be freed at ~LottieComposition();

hermet (commented):
Thanks, one more comment; everything else looks fine so far. I'm looking for the following implementation patches.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -468,12 +468,20 @@ void LottieParser::parsePropertyInternal(T& prop)
 
 
 template<typename T>
-void LottieParser::parseProperty(T& prop)
+void LottieParser::parseProperty(T& prop, LottieSlot::Type type, LottieObject* parent)
```
hermet, line 472: Considering again, the term 'parent' here, does it function conceptually as a parent in a hierarchical structure, or is it simply the 'object' associated with a property? I thin it's more likely to be => LottieObject* obj

hermet (commented):
Reviewed: "lottie/parser: Added sid parsing logic"

inc/thorvg.h
```diff
@@ -1769,6 +1769,18 @@ class TVG_API Animation
...
+     * @brief Rewrite lottie properties through the slot data
+     *
+     * @param[in] slotJson The lottie slot data
+     *
+     * @retval Result::Success When succeed.
+     * @retval Result::InsufficientCondition When the given parameter is invalid.
+     *
+     * @note Experimental API
+     */
+    Result override(const char* slotJson) noexcept;
```
hermet, line 1782: This concept is not suitable for general Animation. Please remove the API implementation from this patch. We can consider introducing LottieAnimation instead.

src/loaders/lottie/tvgLottieOverrider.h
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright (c) 2023 - 2024 the ThorVG project. All rights reserved.
```
hermet, line 2: -> 2024

src/loaders/lottie/tvgLottieOverrider.cpp
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright (c) 2023 - 2024 the ThorVG project. All rights reserved.
```
hermet, line 2: -> 2024

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -613,6 +613,28 @@ struct LottieComposition
 {
     ~LottieComposition();
 
+    LottieObject* findObject(const char* sid) {
```
hermet, line 616: Suggestion: 
Essentially, ThorVG prefers simpler method names (if no any conflict) that consist of just a verb, like.
```
LottieObject* find(const char* sid);
```
As you can see, we can easily figure out the purpose of this method through method name and the return type value + the given parameter.

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -613,6 +613,28 @@ struct LottieComposition
...
+        auto targets = this->root->children;
+        return getObject(sid, root->children);
+    }
+
+    LottieObject* getObject(const char* sid, const Array<LottieObject*> children) {
```
hermet, line 621: Should this method be introduced as a member of LottieComposition()? I think this is not relevant to it. Probably you can add this as a just static helper function (see 29 line in tvgLottieModel.cpp )

Also, Just for your information, Please double-check the difference in binary size between inline (.h) and (.cpp) implementations when you adding a inline function. We will use the one with the smaller size.

src/renderer/tvgAnimation.cpp
```diff
@@ -24,6 +24,8 @@
 #include "tvgFrameModule.h"
```
hermet, line 24: Let's make a Animation change with a separate commit.

src/loaders/lottie/tvgLottieParser.h
```diff
@@ -35,6 +35,7 @@ struct LottieParser : LookaheadParserHandler
     }
 
     bool parse();
+    bool parseSlot(LottieComposition* comp, LottieSlot* slot);
```
hermet, line 38: We can remove the 'comp' parameter since this parser already has a designated 'comp' (see line 40).
``` parseSlot() -> parse(LottieSlot* slot);  ```
However, I expect the ```parse(const char* slot)```,  parsing json logic sholud be in the parser. We can eliminate the dependency on RapidJSON from the overrider if the parsing tasks can be transferred to LottieParser.

File: src/loaders/lottie/tvgLottieParser.cpp [hunk 11]
```diff
@@ -1243,6 +1244,67 @@ void LottieParser::postProcess(Array<LottieGlyph*>& glyphes)
 /* External Class Implementation                                        */
 /************************************************************************/
 
+bool LottieParser::parseSlot(LottieComposition* comp, LottieSlot* slot)
```
hermet, line 1247: please double check indentaion.

src/loaders/lottie/tvgLottieOverrider.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include "tvgLottieParser.h"
+#include "tvgLottieLoader.h"
+
+using namespace std;
+using namespace rapidjson;
+
+/************************************************************************/
+/* Internal Class Implementation                                        */
+/************************************************************************/
+
```
hermet, line 39: please double check indentation.

src/renderer/tvgAnimation.cpp
```diff
@@ -62,6 +65,21 @@ Animation::Animation() : pImpl(new Impl)
 }
 
 
+Result Animation::override(const char* slotJson) noexcept
+{
+    if (!pImpl->overrider) {
+        auto loader = static_cast<LottieLoader*>(pImpl->picture->pImpl->loader);
```
hermet, line 71: This should be something like this:
```
static_cast<LottieLoader*>(pImpl->picture->pImpl->loader)->override(slot);
```
Then, the LottieOverrider can reside within LottieLoader.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1243,6 +1244,67 @@ void LottieParser::postProcess(Array<LottieGlyph*>& glyphes)
...
+    enterObject();
+
+    // Generate context
+    LottieParser::Context context;
+    this->context = &context;
+
+    // Find target parent object
+    auto target = comp->findObject(slot->sid);
+
+    switch (slot->type)
```
hermet, line 1261: please check a coding convention about switch() at 374 line tvgSwShape.cpp

hermet (changes requested):
reviewed "lottie: Support the slot overriding feature"

src/examples/LottieSlot.cpp
```diff
@@ -0,0 +1,227 @@
+/*
+ * Copyright (c) 2023 - 2024 the ThorVG project. All rights reserved.
```
hermet, line 2: 2024

hermet (commented):
reviewd "examples: Added lottie slot example"

src/loaders/lottie/tvgLottieParser.cpp
tinyjin: Yes, it's just associated object, renamed it

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -233,4 +249,9 @@ LottieComposition::~LottieComposition()
     for (auto s = slots.data; s < slots.end(); ++s) {
         delete(*s);
     }
+}
+
+LottieObject* LottieComposition::find(const char* sid) {
+    auto targets = this->root->children;
```
hermet, line 255: one liner:
```
    return _getObject(sid, this->root->children);
```

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -29,6 +29,22 @@
 /* Internal Class Implementation                                        */
 /************************************************************************/
 
+static LottieObject* _getObject(const char* sid, const Array<LottieObject*> children) {
```
hermet, line 32: Avoid memory copying:
```
const Array<LottieObject*> children -> , const Array<LottieObject*>& children
```

src/loaders/lottie/tvgLottieOverrider.cpp
```diff
@@ -0,0 +1,115 @@
...
+    parser->comp = this->comp;
+
+    auto sid = parser->getSid();
+    if (!sid) return false;
+
+    auto slots = comp->slots;
+    for (auto s = slots.data; s < slots.end(); ++s) {
+        auto slot = (*s);
+
+        if (!strcmp(slot->sid, sid)) {
```
hermet, line 51:
```
if (!strcmp(slot->sid, sid)) return parser->parse(slot);
```

src/loaders/lottie/tvgLottieOverrider.cpp
```diff
@@ -0,0 +1,115 @@
...
+            parser->parse(slot);
+            break;
+        }
+    }
+
+    return true;
```
hermet, line 57: probably false?

src/loaders/lottie/tvgLottieOverrider.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+
+void LottieOverrider::overrideGradient(LottieObject* target, LottieGradient* gradient)
```
hermet, line 61: Let's use a simple name (override) since we can easily figure out the objective with the parameters.

src/loaders/lottie/tvgLottieOverrider.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+    switch (target->type)
+    {
```
hermet, line 64: Please double check the convention:
image

src/loaders/lottie/tvgLottieOverrider.cpp
```diff
@@ -0,0 +1,115 @@
...
+        TVGERR("LOTTIE", "Invalid target type for gradient override!");
+        break;
+    }
+}
+
+
+void LottieOverrider::overrideColor(LottieObject* target, LottieObject* color)
+{
+    auto solid = static_cast<LottieSolidFill*>(color);
+    switch (target->type)
```
hermet, line 88: fix convention:

src/loaders/lottie/tvgLottieOverrider.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+
+void LottieOverrider::overrideGradient(LottieObject* target, LottieGradient* gradient)
+{
+    switch (target->type)
+    {
+        case LottieObject::GradientFill: {
+            auto fill = static_cast<LottieGradientFill*>(target);
+            fill->colorStops = gradient->colorStops;
```
hermet, line 67: 1. Maybe it needs to free the ex-memory... ```fill->colorStops->frames``` overriding  operator=() ?
2. It seems here polymorphism helps to make code neat & clean. Please consider the interface of the next:
```
virtual bool LottieObject::override(LottieObject* prop) {}

bool LottieGradientFill::override(LottieObject* prop) override
{
    auto fill = static_cast<LottieGradientFill*>(prop);
    this->colorStops = gradient->ColorStops;
}

```

hermet (changes requested):
Reviewed: "lottie: Support the slot overriding feature"

src/loaders/lottie/tvgLottieAnimation.cpp
```diff
@@ -0,0 +1,55 @@
...
+/************************************************************************/
+
+LottieAnimation::~LottieAnimation()
+{
+    delete(pImpl);
+}
+
+Result LottieAnimation::override(const char* slotJson) noexcept
+{
+    if (static_cast<LottieLoader*>(pImpl->picture->pImpl->loader)->override(slotJson)) {
```
hermet, line 44: if (!pImpl->picture->pImpl->loader) return InsufficientCondition;

src/renderer/tvgAnimation.cpp
```diff
@@ -20,33 +20,15 @@
  * SOFTWARE.
  */
 
-#include "tvgCommon.h"
 #include "tvgFrameModule.h"
-#include "tvgPaint.h"
 #include "tvgPicture.h"
+#include "tvgLottieLoader.h"
```
hermet, line 25: this file doesn't refer: 
-- #include "tvgLottieLoader.h"

-- #include "tvgPicture.h"   //can be included in tvgAnimation.h

src/renderer/tvgAnimation.h
```diff
@@ -0,0 +1,47 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef _TVG_ANIMATION_H_
+#define _TVG_ANIMATION_H_
+
+#include "tvgCommon.h"
+#include "tvgPaint.h"
```
hermet, line 27: ++ #include "tvgPicture.h"

inc/thorvg.h
```diff
@@ -1849,6 +1849,45 @@ class TVG_API Animation
...
```
hermet, line 1878: slotJson -> slot

inc/thorvg.h
```diff
@@ -1849,6 +1849,45 @@ class TVG_API Animation
...
+    ~LottieAnimation();
+
+    /**
+     * @brief Rewrite lottie properties through the slot data
+     *
+     * @param[in] slotJson The lottie slot data
+     *
+     * @retval Result::Success When succeed.
+     * @retval Result::InsufficientCondition When the given parameter is invalid.
+     *
```
hermet, line 1875: ++InsufficientCondition

hermet (changes requested):
reviewd "lottie: Introduce the LottieAnimation class"

src/examples/LottieSlot.cpp
```diff
@@ -0,0 +1,227 @@
...
+    auto shape = tvg::Shape::gen();
+    shape->appendRect(0, 0, WIDTH, HEIGHT);
+    shape->fill(50, 50, 50);
+
+    if (canvas->push(std::move(shape)) != tvg::Result::Success) return;
+
+    const char* slotData = R"(
+    {
+        "gradient_fill": {
+            "p": {
```
hermet, line 57: this json is ok?

hermet:
@tinyjin thanks for your update, please check my comments.

hermet:
Imagine one another scenario, if the user intends to cancel the override of the slot by using:
```
animation->override(nullptr);
```

src/renderer/tvgAnimation.cpp
tinyjin: Without `tvgLottieLoader.h`, the `LottieLoader` cannot be found

> identifier "LottieLoader" is undefinedC/C++(20)
```cpp
if (static_cast<LottieLoader*>(pImpl->picture->pImpl->loader)->override(slot)) {
```

inc/thorvg.h
tinyjin: line: 1874
`InsufficientCondition` is already there?

src/renderer/tvgAnimation.cpp
hermet: Please double-check it, I couldn't see any LottieLoader usage in tvgAnimation.cpp.
Even the call is belonged to tvgLottieAnimation.cpp, not tvgAnimation.cpp

src/loaders/lottie/tvgLottieOverrider.cpp
```diff
@@ -0,0 +1,115 @@
...
+ * SOFTWARE.
+ */
+
+#include "tvgLottieOverrider.h"
+#include "tvgCommon.h"
+#include "tvgLottieOverrider.h"
+#include "tvgLottieParser.h"
+#include "tvgLottieLoader.h"
+
+using namespace std;
```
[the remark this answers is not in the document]
hermet: I meant the 29 line.

src/renderer/tvgAnimation.cpp
tinyjin: Oops, I checked wrong file. that's correct.

src/examples/LottieSlot.cpp
tinyjin: Screenshot 2024-02-21 at 1 00 42 AM
Yes, In some white-space cases, RapidJSON cannot parse, but this example JSON can be parsed.

We might need to find the crashed pattern with the white-space.

hermet:
> > Imagine one another scenario, if the user intends to cancel the override of the slot by using:
> > ```
> > animation->override(nullptr);
> > ```
> 
> @hermet Sounds good.
> 
> And wondering:
> 
> * We discussed this overriding features shouldn't take a memory to record what's overrided.
> * Do we have efficient way to clear overridden status without saving changes?

Please check this: https://github.com/thorvg/thorvg/pull/1998

```
LottieSlot
{
  LottieProperty::Type type;
  char* sid;
  LottieProperty* property  //we can backup the origin here:
};
```
Probably you can make the slot rewind the sequence with the original property in LottiSlot.

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -590,6 +592,25 @@ struct LottieLayer : LottieGroup
 };
 
 
+struct LottieSlot
+{
+    char* sid = nullptr;
+
+    enum Type : uint8_t
```
hermet, line 631: Could you please cherry pick this first, and make this based on it?
https://github.com/thorvg/thorvg/pull/1998

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -228,4 +228,9 @@ LottieComposition::~LottieComposition()
     for (auto f = fonts.begin(); f < fonts.end(); ++f) {
         delete(*f);
     }
+
+    //delete slots
+    for (auto s = slots.begin(); s < slots.end(); ++s) {
+        delete(*s);
```
hermet, line 248: Considering again, freeing the resource by Slot is better here instead of 124 line.
```
free(s->sid)
```

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -620,6 +641,7 @@ struct LottieComposition
     Array<LottieObject*> assets;
     Array<LottieInterpolator*> interpolators;
     Array<LottieFont*> fonts;
+    Array<LottieSlot*> slots;
```
hermet, line 678: Reviewing again, there is no point to use dynamic memory allocation for slots. 
```
Array<LottieSlot*> slots -> Array<LottieSlot> slots
```

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -457,12 +457,20 @@ void LottieParser::parsePropertyInternal(T& prop)
 
 
 template<typename T>
```
hermet, line 460: Probably, we can remove a parameter, by adding: 
```
template<typename T, LottieProperty type>
void LottieParser::parseProperty(T& prop, LottieObject* obj)
```

See https://github.com/thorvg/thorvg/pull/1998

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1228,6 +1229,77 @@ void LottieParser::postProcess(Array<LottieGlyph*>& glyphes)
...
+            auto gradient = new LottieGradient;
+            this->context->gradient = gradient;
+
+            while (auto key = nextObjectKey()) {
+                if (!strcmp(key, "p")) parseProperty(gradient->colorStops);
+                else skip(key);
+            }
+
+            LottieOverrider::override(targetObj, gradient);
+            delete(gradient);
```
hermet, line 1264: assuming this memory deletion also delete the property that has been belonged to the targetObj. So memory corruption... Please double check,

src/loaders/lottie/tvgLottieOverrider.h
```diff
@@ -0,0 +1,43 @@
...
+#define _TVG_LOTTIE_OVERRIDER_H_
+
+#include "tvgCommon.h"
+#include "tvgLottieModel.h"
+
+struct LottieOverrider
+{
+    bool override(const char* data);
+
+    static void override(LottieObject* target, LottieGradient* gradient);
```
hermet, line 33: these 3 override() looks unnecessary...

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1228,6 +1229,77 @@ void LottieParser::postProcess(Array<LottieGlyph*>& glyphes)
...
+                if (!strcmp(key, "p")) parseProperty(text->doc);
+                else skip(key);
+            }
+
+            LottieOverrider::override(targetObj, text);
+            delete(text);
+            break;
+        }
+        default: {
+            TVGERR("LOTTIE", "Unsupported slot type");
```
hermet, line 1292: You can put this log message in LottieObject::override()...

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -468,6 +488,12 @@ struct LottieGradientFill : LottieGradient
         if (LottieGradient::prepare()) statical = false;
     }
 
+    void override(LottieObject* prop) override
+    {
+        this->colorStops = static_cast<LottieGradient*>(prop)->colorStops;
```
hermet, line 493: definitely memory leak due to the previous property value.

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -350,6 +364,12 @@ struct LottieSolidFill : LottieSolid
         if (color.frames || opacity.frames) statical = false;
     }
 
+    void override(LottieObject* prop) override
+    {
+        this->color = static_cast<LottieSolid*>(prop)->color;
```
hermet, line 369: definitely memory leak due to the previous property value.

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -339,6 +347,12 @@ struct LottieSolidStroke : LottieSolid, LottieStroke
         LottieObject::type = LottieObject::SolidStroke;
         if (color.frames || opacity.frames || LottieStroke::dynamic()) statical = false;
     }
+
+    void override(LottieObject* prop) override
+    {
+        this->color = static_cast<LottieSolid*>(prop)->color;
```
hermet, line 353: definitely memory leak due to the previous property value.

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -183,6 +185,12 @@ struct LottieText : LottieObject
         LottieObject::type = LottieObject::Text;
     }
 
+    void override(LottieObject* prop) override
+    {
+        this->doc = static_cast<LottieText*>(prop)->doc;
```
hermet, line 190: definitely memory leak due to the previous property value.

hermet (changes requested):
reviewed "lottie: Support the slot overriding feature"

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -121,9 +121,11 @@ struct LottieObject
     virtual ~LottieObject()
     {
         free(name);
+        free(sid);
```
hermet, line 124: doesn't this sid double free? => 234 line

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -456,13 +456,21 @@ void LottieParser::parsePropertyInternal(T& prop)
...
         if (!strcmp(key, "k")) parsePropertyInternal(prop);
-        else skip(key);
+        else if (obj && !strcmp(key, "sid")) {
+            auto slot = new LottieSlot;
+            if (!slot) return;
+
+            slot->sid = getStringCopy();
+            obj->sid = slot->sid;
+            slot->type = type;
+            comp->slots.push(*slot);
```
hermet, line 473: Please check the next, comp.slots doesn't have elements with the memory pointer anymore.
```
comp->slots.push({getStringCopy(), type});
obj->sid = comp->slots.last().sid;
```

hermet (changes requested):
reviewed "lottie/parser: Added sid parsing logic"

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -29,6 +29,20 @@
...
+static LottieObject* _getObject(const char* sid, const Array<LottieObject*>& children) {
+    for (auto c = children.begin(); c < children.end(); ++c) {
+        auto child = *c;
+        if (child->type == LottieObject::Type::Group || child->type == LottieObject::Type::Layer) {
+            auto group = static_cast<LottieGroup*>(child);
+            auto foundObj = _getObject(sid, group->children);
+            if (foundObj) return foundObj;
+        }
+
+        if (child->sid && !strcmp(sid, child->sid)) return child;
```
hermet, line 41: This condition should be at 34 ~ 35 line, if the searching is at group itself?

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -124,6 +124,10 @@ struct LottieObject
         free(sid);
     }
 
+    virtual void override(LottieObject* prop) {
```
hermet, line 127: convention fix:
```
virtual void override(LottieObject* prop)
{
 ...
}
```

src/loaders/lottie/tvgLottieOverrider.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+/************************************************************************/
+/* Internal Class Implementation                                        */
+/************************************************************************/
+
+/************************************************************************/
+/* External Class Implementation                                        */
+/************************************************************************/
+
+bool LottieOverrider::override(const char* slot)
```
hermet, line 35: LottieOverride doesn't need to hold dirName / comp as member variables?
```
bool LottieOverrider::override(LottieComposition* comp, const char* slot, const char* dirName);
```

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -249,6 +249,19 @@ struct LottieGenericProperty : LottieProperty
         return frame->interpolate(frame + 1, frameNo);
     }
 
+    T& operator=(const T& other)
+    {
+        if (this != &other) {
```
hermet, line 254: Let's remove 254 condition. (we don't expect this scenario)

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -249,6 +249,19 @@ struct LottieGenericProperty : LottieProperty
...
+            value = other.value;
+            if (other.frames) {
+                memcpy(frames, other.frames, sizeof(LottieScalarFrame<T>) * other.frames->count);
+                delete(other.frames);
```
hermet, line 258: i think it needs to delete frames of "this" before memory copy, not "others".

hermet (changes requested):
reviewed "lottie: Support the slot overriding feature"

hermet:
@tinyjin Please check more comments. thanks!

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -124,6 +124,7 @@ struct LottieObject
     }
 
     char* name = nullptr;
+    char* sid = nullptr;
```
hermet, line 132: Could you please double check this?
```
Slot {
   char* sid;
   type;
   Array<LottieObject*> objs;   //assuming multiple properties uses the same sid.
};
```
I think LottieObject doesn't need to have sid member.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1223,6 +1224,76 @@ void LottieParser::postProcess(Array<LottieGlyph*>& glyphes)
...
+bool LottieParser::parse(LottieSlot* slot)
+{
+    enterObject();
+
+    // Generate context
+    LottieParser::Context context;
+    this->context = &context;
+
+    // Find target object
+    auto targetObj = comp->find(slot->sid);
```
hermet, line 1246: you can quickly find the targetObjs from the slots objs.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1223,6 +1224,76 @@ void LottieParser::postProcess(Array<LottieGlyph*>& glyphes)
...
+        case LottieProperty::Type::ColorStop: {
+            auto gradient = new LottieGradient;
+            this->context->gradient = gradient;
+
+            while (auto key = nextObjectKey()) {
+                if (!strcmp(key, "p")) parseProperty(gradient->colorStops);
+                else skip(key);
+            }
+
+            targetObj->override(gradient);
```
hermet, line 1258: And like this:
```
for (obj = slot.objs.begin(); obj < slot.objs.end(); ++obj) {
   (*obj)->override(gradient);
}
```

hermet (commented):
Almost there, please check once more for the optimal way.

hermet:
+please rebase the branch as well.

src/loaders/lottie/tvgLottieOverrider.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef _TVG_LOTTIE_OVERRIDER_H_
+#define _TVG_LOTTIE_OVERRIDER_H_
+
+#include "tvgCommon.h"
+#include "tvgLottieModel.h"
+
+struct LottieOverrider
+{
+    bool override(LottieComposition* comp, const char* slot, const char* dirName);
```
hermet, line 31: Now, LottieOverrider is too simple, 
Let's migrate the LottieOverrider function with 
```LottieLoader::override(LottieComposition* comp, const char* slot, const char* dirName);```
and then remove LottieOverrider.

hermet:
https://github.com/thorvg/thorvg/pull/2003

Issue #1686, opened 2023-10-05
incorrect dash offset position


hermet:
Dash-offset result is different with inkscape version. 

![image](https://github.com/thorvg/thorvg/assets/3711518/27f53589-755a-404a-8122-3f62fcb517f4)

```
<svg width="512" height="512" version="1.1" viewBox="0 0 512 512" xmlns="http://www.w3.org/2000/svg">
	<g fill="none">
		<g stroke-width="20">
			<path d="m20 40h120l-20 90-80-10v-100" stroke="#f00" stroke-linecap="round" stroke-linejoin="bevel"/>
			<path d="m170 40h120l-20 90-80-10v-100" stroke="#ff0" stroke-linecap="square" stroke-linejoin="bevel"/>
			<path d="m320 40h120l-20 90-80-10v-100" stroke="#0f0" stroke-linecap="square" stroke-miterlimit="5"/>
		</g>
		<g stroke-dasharray="40, 40, 60" stroke-width="20">
			<path d="m20 175.21h120l-20 90-80-10v-100" stroke="#f00" stroke-linecap="round" stroke-linejoin="round"/>
			<path d="m170 175.21h120l-20 90-80-10v-100" stroke="#ff0" stroke-linecap="square" stroke-linejoin="bevel"/>
			<path d="m320 175.21h120l-20 90-80-10v-100" stroke="#0f0" stroke-linecap="square" stroke-miterlimit="5"/>
		</g>
		<g stroke-width="15">
			<path d="m137.52 295.83a100 100 0 0 1-58.193 77.9l-40.912-91.248z" stroke="#f00" stroke-linecap="round" stroke-linejoin="round"/>
			<path d="m287.52 295.83a100 100 0 0 1-58.193 77.9l-40.912-91.248z" stroke="#ff0" stroke-linecap="square" stroke-linejoin="bevel"/>
			<path d="m437.52 295.83a100 100 0 0 1-58.193 77.9l-40.912-91.248z" stroke="#0f0" stroke-linecap="square" stroke-miterlimit="5"/>
		</g>
		<g stroke-dashoffset="10" stroke-width="10">
			<path d="m137.52 405.83a100 100 0 0 1-58.193 77.9l-40.912-91.248z" stroke="#f00" stroke-dasharray="30,20" stroke-linecap="round" stroke-linejoin="round"/>
[...]

PR #1985, opened 2024-02-16
sw_engine: Increasing accuracy for dashed curves

mgrudzinska:
Dashed curves require greater precision in calculating their lengths. Otherwise, it results in visual
discrepancies compared to the expected outcomes.

issue: https://github.com/thorvg/thorvg/issues/1686

before:

after:

### DIFF
--- a/src/common/tvgBezier.cpp
+++ b/src/common/tvgBezier.cpp
@@ -29,7 +29,7 @@
 /* Internal Class Implementation                                        */
 /************************************************************************/
 
-static float _lineLength(const Point& pt1, const Point& pt2)
+static float _lineLengthApprox(const Point& pt1, const Point& pt2)
 {
     /* approximate sqrt(x*x + y*y) using alpha max plus beta min algorithm.
        With alpha = 1, beta = 3/8, giving results with the largest error less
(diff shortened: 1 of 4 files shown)


hermet (approved):

hermet:
Hello, probably, we can make the old version of the logic redundant.

src/renderer/tvgShape.cpp
```diff
@@ -130,11 +130,11 @@ Result Shape::appendCircle(float cx, float cy, float rx, float ry) noexcept
     auto ryKappa = ry * PATH_KAPPA;
 
     pImpl->grow(6, 13);
-    pImpl->moveTo(cx, cy - ry);
-    pImpl->cubicTo(cx + rxKappa, cy - ry, cx + rx, cy - ryKappa, cx + rx, cy);
+    pImpl->moveTo(cx + rx, cy);
```
hermet, line 133: Changing it might introduce a regression bug, which we can address with another patch.

hermet (changes requested):

src/renderer/tvgShape.cpp
mgrudzinska: I just wanted to comment this :)
@JSUYA - just so you know, it may cause some problems in tests in efl, as I remember correctly

mgrudzinska: this change is moved to https://github.com/thorvg/thorvg/pull/1987

src/common/tvgBezier.cpp
```diff
@@ -41,14 +41,75 @@ static float _lineLength(const Point& pt1, const Point& pt2)
 }
 
 
+static float _lineExactLength(const Point& pt1, const Point& pt2)
+{
+    Point diff = {pt2.x - pt1.x, pt2.y - pt1.y};
+    return sqrtf(diff.x * diff.x + diff.y * diff.y);
```
hermet, line 47: We only need a single line for this fix. I think you are going overboard with the changes?

mgrudzinska:
tmr I'll check test on windows

hermet:
So It increases the complexity. I think replacing the old trick with sqrt() is the better option.

mgrudzinska:
> Hello, probably, we can make the old version of the logic(wrt sqrt trick) redundant.

@hermet sure I can remove the old logic. but please keep in mind that the ''bezAt'' is used inside the 'interpolate' in the LottieVectorFrame struct. it is called many times. 
can you please confirm that you agree with removing the approx calculations?

hermet:
> > Hello, probably, we can make the old version of the logic(wrt sqrt trick) redundant.
> 
> @hermet sure I can remove the old logic. but please keep in mind that the ''bezAt'' is used inside the 'interpolate' in the LottieVectorFrame struct. it is called many times. can you please confirm that you agree with removing the approx calculations?

Unfortunately I'm unsure if the phrase 'many times' significantly impacts the performance. It's best to measure the performance difference using an advanced profiler, such as Intel VTune, with a Lottie example. Otherwise I will prevent the unwilling code that just make code maintainance hard. I will verify this when I have time or you can conduct this.

hermet (approved):

mgrudzinska:
oh, merged, I'll do these tests regardless, so that we don't keep any unnecessary code if it turns out to be redundant.

hermet:
@mgrudzinska Binary diff is just around 50 (ignorable). performance was certainly dropped (~5%) with Lottie example.

PR #1987, opened 2024-02-19
common: Changing the circle/ellipse starting point

mgrudzinska:
According to the SVG standard, drawing a circle/ellipse should start at the '3 o'clock' position and proceed in a clock-wise direction. Until now, this point was set at '12 o'clock'.
The differences in the outcome were visible for dashed strokes.

issue: https://github.com/thorvg/thorvg/issues/1686

before:
el_before

after:
el_after

### DIFF
--- a/src/renderer/tvgShape.cpp
+++ b/src/renderer/tvgShape.cpp
@@ -130,11 +130,11 @@ Result Shape::appendCircle(float cx, float cy, float rx, float ry) noexcept
     auto ryKappa = ry * PATH_KAPPA;
 
     pImpl->grow(6, 13);
-    pImpl->moveTo(cx, cy - ry);
-    pImpl->cubicTo(cx + rxKappa, cy - ry, cx + rx, cy - ryKappa, cx + rx, cy);
+    pImpl->moveTo(cx + rx, cy);
     pImpl->cubicTo(cx + rx, cy + ryKappa, cx + rxKappa, cy + ry, cx, cy + ry);
     pImpl->cubicTo(cx - rxKappa, cy + ry, cx - rx, cy + ryKappa, cx - rx, cy);
     pImpl->cubicTo(cx - rx, cy - ryKappa, cx - rxKappa, cy - ry, cx, cy - ry);
+    pImpl->cubicTo(cx + rxKappa, cy - ry, cx + rx, cy - ryKappa, cx + rx, cy);
     pImpl->close();
 
     return Result::Success;


src/renderer/tvgShape.cpp
```diff
@@ -130,11 +130,11 @@ Result Shape::appendCircle(float cx, float cy, float rx, float ry) noexcept
     auto ryKappa = ry * PATH_KAPPA;
 
     pImpl->grow(6, 13);
-    pImpl->moveTo(cx, cy - ry);
-    pImpl->cubicTo(cx + rxKappa, cy - ry, cx + rx, cy - ryKappa, cx + rx, cy);
+    pImpl->moveTo(cx + rx, cy);
     pImpl->cubicTo(cx + rx, cy + ryKappa, cx + rxKappa, cy + ry, cx, cy + ry);
     pImpl->cubicTo(cx - rxKappa, cy + ry, cx - rx, cy + ryKappa, cx - rx, cy);
     pImpl->cubicTo(cx - rx, cy - ryKappa, cx - rxKappa, cy - ry, cx, cy - ry);
+    pImpl->cubicTo(cx + rxKappa, cy - ry, cx + rx, cy - ryKappa, cx + rx, cy);
```
hermet, line 137: Please see line 218 as well; we can reuse this instead of creating an individual implementation.

mgrudzinska: thx for tip with line 218! fixed.
I'm not sure what do you mean by 'reuse this'. here we have a circle. there is a rounded rectangle. I can add there an extra case for circles, but is it worth it?

mgrudzinska:
results from both commits (after #1985 is applied):
mira_circ_popr

before:
mira_circ

src/renderer/tvgShape.cpp
mgrudzinska: @hermet do you want me to make a special case for circles in appendRect and in the appendCircle call appendRect?

hermet: @mgrudzinska Apologies for the confusion. I believe the appendRect() method should not be modified; the previous version seems correct. Please revert the changes made to appendRect(). Also, I had previously modified appendCircle() to start at 90 degrees, but I had to revert the change due to an issue caused by Lottie. I will double-check it, basically I agree on this change.

mgrudzinska: @hermet I double checked in the svg standard. the rounded rect should start in a different point than it is in the current version of the code.  so both this changes are correct. 
but if they cause some other consequences lets investigate it and keep this pr open till then.

hermet: @mgrudzinska Got it, Thanks. soon, I will double-check any breaks by this change.

hermet:
Condition: Need to tweak this circle trimpath regression bug, before changing the circle commands:
image
[happy.json](https://github.com/thorvg/thorvg/files/14338743/happy.json)

@mgrudzinska probably, you can separate changes from circle and rect so that we can apply one of them first.

mgrudzinska:
@hermet rect moved to #1996

mgrudzinska:
> Condition: Need to tweak this circle trimpath regression bug, before changing the circle commands: image [happy.json](https://github.com/thorvg/thorvg/files/14338743/happy.json)
> 
> @mgrudzinska probably, you can separate changes from circle and rect so that we can apply one of them first.

I see the problem. will be present for both, circle and rect. I just need to test the solution before sending it, so I opened https://github.com/thorvg/thorvg/issues/1997 so we both don't spent time on the same task

hermet (approved):

PR #1996, opened 2024-02-20
common: Changing the rounded rect starting point

mgrudzinska:
According to the SVG standard, drawing a rectangle starts at (x+rx, y) point. Till now it was (x+w, y+ry). The difference was visible only for dashed strokes.

### DIFF
--- a/src/renderer/tvgShape.cpp
+++ b/src/renderer/tvgShape.cpp
@@ -215,20 +215,20 @@ Result Shape::appendRect(float x, float y, float w, float h, float rx, float ry)
         pImpl->lineTo(x + w, y + h);
         pImpl->lineTo(x, y + h);
         pImpl->close();
-    //circle
+    //rounded rectangle or circle
     } else {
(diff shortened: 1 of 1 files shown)


mgrudzinska:
changes visualisation:
https://github.com/thorvg/thorvg/pull/1987#issuecomment-1952500264

src/renderer/tvgShape.cpp
```diff
@@ -215,20 +215,20 @@ Result Shape::appendRect(float x, float y, float w, float h, float rx, float ry)
         pImpl->lineTo(x + w, y + h);
         pImpl->lineTo(x, y + h);
         pImpl->close();
-    //circle
-    } else {
+        //rounded rectangle or circle
```
hermet, line 218: wrong indentation?

mgrudzinska: fixed

hermet (approved):

Issue #1997, opened 2024-02-20
Divergence of starting points between SVG and Lottie standards


mgrudzinska:
https://github.com/thorvg/thorvg/pull/1987#issuecomment-1953394872

#1987 and #1996 are fixes to meet the svg standard. Since the lottie starting points are defined differently, there needs to be added some 'shift'.

hermet:
`appendShapes` is a utility function for app developers, whereas SVG/Lottie can directly utilize `appendPath` to compose any shapes in accordance with their specifications.

PR #1999, opened 2024-02-20
lottie: use internal append rect/ellipse func

mgrudzinska:
The lottie format defines the starting point for
drawing rectangles and ellipses differently than
the API available in the TVG. In the case of using dashes or trim paths, this leads to discrepancies. Therefore, when using the lottie builder, internal functions for appending rectangles and circles are utilized.

@issue: https://github.com/thorvg/thorvg/issues/1997

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -34,6 +34,8 @@
 /************************************************************************/
 /* Internal Class Implementation                                        */
 /************************************************************************/
+constexpr auto PATH_KAPPA = 0.552284f;
+
 
 struct RenderRepeater
 {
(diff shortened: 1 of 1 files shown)


hermet:
1. Is this a better approach that compromises renderer compatibility and then modifies the Lottie? Or can we simply tweak the SVG alone?
2. Have you reviewed the specifications for rectangles here? If not, we may need to update only the problematic circle.

mgrudzinska:
> 1. Is this a better approach that compromises renderer compatibility and then modifies the Lottie? Or can we simply tweak the SVG alone?
> 2. Have you reviewed the specifications for rectangles here? If not, you may need to update only the problematic circle.

I'm not sure if I understand the question, but I'll try to answer. 
The renderer is set to a certain starting point for the circle and rectangle. In TVG, it is consistent with the SVG standard, though it could well be different. But as long as it's different from the Lottie standard, we need to do something about it so that Lottie files are displayed correctly.

Here, I proposed such a solution because it's very simple and we can easily change it in the future. In this way, we can also merge prs #1987 and #1996

As for the ultimate solution - I've read more about the Lottie standard and I think that the implementation of 'direction' will result in the solution proposed here being able to be modified as well.

And ofc I checked the lottie specs for rects. That's why I included some changes regarding rects into this pr as well.

@hermet if you think this isn't a good approach, please close this pr. I'm new in the lottie format, I'm probably missing many aspects of it right now.

hermet:
> The renderer is set to a certain starting point for the circle and rectangle. In TVG, it is consistent with the SVG standard, though it could well be different.

@mgrudzinska The renderer itself is a open method, it could not follow any specification, I've never designed it from the get go, it is a just a pure graphics engine.

hermet:
I'm considering the less change as possible for the best of both worlds.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -371,9 +373,68 @@ static void _repeat(LottieGroup* parent, unique_ptr<Shape> path, RenderContext*
...
+            {x + w - rx + hrx, y}, {x + w, y + ry - hry}, {x + w, y + ry} //cubicTo
+    };
+
+    shape->appendPath(commands, commandsCnt, points, pointsCnt);
+}
+
+
+static void _appendRect(Shape* shape, float x, float y, float w, float h, float r)
+{
+    if (mathZero(r)) {
```
hermet, line 428: suggestion: 2 lines are enough:
```
if (mathZero(r)) _appendSharpRect(shape, x, y, w, h, r);
else _appendRoundedRect(shape, x, y, w, h, r);
```

hermet (commented):
thanks, just a few nit-picks.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -371,9 +373,68 @@ static void _repeat(LottieGroup* parent, unique_ptr<Shape> path, RenderContext*
 }
 
 
+static void _appendSharpRect(Shape* shape, float x, float y, float w, float h, float r)
+{
+    constexpr int commandsCnt = 5;
+    PathCommand commands[commandsCnt] = {
+            PathCommand::MoveTo, PathCommand::LineTo, PathCommand::LineTo,
```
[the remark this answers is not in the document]
mgrudzinska: sorry for that, changed everywhere

hermet (approved):

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -371,9 +373,65 @@ static void _repeat(LottieGroup* parent, unique_ptr<Shape> path, RenderContext*
...
+    constexpr int pointsCnt = 4;
+    Point points[pointsCnt] = {{x + w, y}, {x + w, y + h}, {x, y + h}, {x, y}};
+
+    shape->appendPath(commands, commandsCnt, points, pointsCnt);
+}
+
+static void _appendRoundedRect(Shape* shape, float x, float y, float w, float h, float r)
+{
+    constexpr int commandsCnt = 10;
+    PathCommand commands[commandsCnt] = {
```
hermet, line 393: this command list style would be good to apply in tvgShape.cpp as well.

mgrudzinska: @hermet  I'm not sure if I understand... in 'appendRect' we use functions from pImpl like 'moveTo', 'cubicTo' etc. would you like to replace them with 'append' from pImpl?

PR #2004, opened 2024-02-22
Lottie loader: fix compilation on Windows

vtorri:
strlen() was used in tvgLottieModel.h without the inclusion of cstring. This patch fixes this error.


vtorri:
btw, test fail alsos here : 

```
../test/testInitializer.cpp:54: FAILED:
  REQUIRE( Initializer::init(0, CanvasEngine(100)) == Result::NonSupport )
with expansion:
  0 == 5

stderr:
[E] TVG (../src/loaders/tvg/tvgTvgBinInterpreter.cpp 494): Corrupted tvg file.
Catch will terminate because it needed to throw an exception.
The message was: Test failure requires aborting test!
terminate called without an active exception
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――


```

hermet (approved):

hermet:
you shouldn't add the tag.tvg in this PR. i replaced with this: https://github.com/thorvg/thorvg/pull/2005 
anyway thanks!

PR #2017, opened 2024-03-01
wasm: rectify the gif/tvg saving function

hermet:


hermet:
https://github.com/thorvg/thorvg/pull/2143

PR #2025, opened 2024-03-04
gl_engine: use stencil and cover tessellator by default

RuiwenTang:
Since we choose multisample antialiasing, use stencil then cover to render polygon can get better performance.
Also the code is much easier to understand.


SergeyLebedkin (approved):

hermet (approved):
@RuiwenTang Can we expect any improvement in the rendering results from this, or is it just intermediate progress for the next updates?

RuiwenTang:
> @RuiwenTang Can we expect any improvement in the rendering results from this, or is it just intermediate progress for the next updates?

Yes, A few more patches are needed to enable MSAA

Issue #2014, opened 2024-02-28
Corrupt radial gradient on Apple Silicon


lmdsp:
I've encountered a strange bug when rendering an svg with certain focal point values.
Strangely enough, the rendering is correct on Intel x64 (mac/PC) and the online ThorVG viewer,
but Apple Silicon / ARM show weird horizontal lines instead of the expected gradient.

```svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg width="800" height="590" version="1.1" id="svg13" xmlns="http://www.w3.org/2000/svg"
   xmlns:svg="http://www.w3.org/2000/svg">
   <defs>
      <radialGradient cx="429" cy="3" fx="429" fy="-600" gradientUnits="userSpaceOnUse" id="b" r="600">
         <stop offset="0" stop-color="red" id="stop10" />
         <stop offset="1" stop-color="white" id="stop11" />
      </radialGradient>
   </defs>

   <path d="M 800,200 H 0 V 590 H 800 V 0 Z" fill="url(#b)" id="svg_4" />
</svg>
```

Note that if `fx` is changed to 600, then there are no artefacts anymore.
This is using latest ThorVG 0.12.6, XCode 15.2 and macos Sonoma 14.3.1.

ThorVG viewer:

![image](https://github.com/thorvg/thorvg/assets/1681470/30731652-e095-4d41-948e-b3012a1d05de)

Test application | x64:

![image](https://github.com/thorvg/thorvg/assets/1681470/64663b72-799e-42a7-89cd-000e9b071b92)

Test application | Apple Silicon:

![image](https://github.com/thorvg/thorvg/assets/1681470/251bf7c7-dac1-41c7-a816-e81b4b233e45)

mgrudzinska:
@lmdsp could you please try after this slight change?
https://github.com/thorvg/thorvg/pull/2018
this is a random guess since for now I'm able to test on apple silicon. 

Also could you please test what will happen if you change fy="-600" to fy="-597" ?

lmdsp:
@mgrudzinska Thanks for looking into this.

Changing from fy="-600" to fy="-597" doesn't change anything.
Changing line 207 from `if (fill->radial.a > 0);` to `if (fill->radial.a > RADIAL_A_THRESHOLD);` doesn't fix the problem either I'm afraid.

Perhaps a hint; those are the contents before and after executing the instructions at line 198
```cpp
 //This condition fulfills the SVG 1.1 std:
[...]

PR #2027, opened 2024-03-05
sw_engine: fix radial gradient when focal point is outside circle on Apple Silicon

lmdsp:
sw_engine: radial gradient

https://github.com/thorvg/thorvg/issues/2014

Radial gradient results in a corrupted image when the focal point is outside the circle on Apple Silicon. This happens because some compilers use FMA to optimize the a = dr² - dx² - dy² calculation, which cause loss of precision.

We use compiler specific float contraction control pragmas to avoid this when they exist, and revert to manual de-optimization in other cases.

### DIFF
--- a/src/renderer/sw_engine/tvgSwFill.cpp
+++ b/src/renderer/sw_engine/tvgSwFill.cpp
@@ -201,7 +201,13 @@ bool _prepareRadial(SwFill* fill, const RadialGradient* radial, const Matrix* tr
         fill->radial.fy = cy + r * (fy - cy) / dist;
         fill->radial.dx = cx - fill->radial.fx;
         fill->radial.dy = cy - fill->radial.fy;
-        fill->radial.a = fill->radial.dr * fill->radial.dr - fill->radial.dx * fill->radial.dx - fill->radial.dy * fill->radial.dy;
+        // Prevent loss of precision on Apple Silicon when dr=dy and dx=0 due to FMA
+        // https://github.com/thorvg/thorvg/issues/2014
+        auto dr2 = fill->radial.dr * fill->radial.dr;
+        auto dx2 = fill->radial.dx * fill->radial.dx;
+        auto dy2 = fill->radial.dy * fill->radial.dy;
+
+        fill->radial.a = dr2 - dx2 - dy2;
     }
 
     if (fill->radial.a > 0) fill->radial.invA = 1.0f / fill->radial.a;
--- a/AUTHORS
+++ b/AUTHORS
@@ -26,3 +26,4 @@ SergeyLebedkin <sergii@lottiefiles.com>
 Jinny You <jinny@lottiefiles.com>
 Nattu Adnan <nattu@reallynattu.com>
 Gabor Kiss-Vamosi <kisvegabor@gmail.com>
+Lorcán Mc Donagh <lorcan@lmdsp.com>
--- a/src/examples/resources/radial_gradient_focal_outside_circle.svg
+++ b/src/examples/resources/radial_gradient_focal_outside_circle.svg
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="800" height="590" version="1.1" id="https://github.com/thorvg/thorvg/issues/2014" xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+   <defs>
+      <radialGradient cx="429" cy="3" fx="429" fy="-600" gradientUnits="userSpaceOnUse" id="b" r="600">
+         <stop offset="0" stop-color="red" id="stop10" />
+         <stop offset="1" stop-color="white" id="stop11" />
+      </radialGradient>
+   </defs>
+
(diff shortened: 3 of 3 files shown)


lmdsp:
@hermet @mgrudzinska Ideally I would add a unit test for this, but from what I see the existing tests only check the tvg::Result return values of operations.

Wouldn't it be better to also do some kind of check of the output image against a known/expected reference ? This would allow catching all sorts of rendering artifacts.

Comparing exact pixels migh be too complicated to implement, but may I suggest at least checking the image mean/min/max values match as a first step ?
At least if the test fails one can look at the image manually and determine what's wrong.

hermet:
@lmdsp Yes, we are already conducting a specific comparison using the regression test and SVG examples. So far, the best approach is to push an SVG samples(src/examples/resources) to address those scenarios, after which the cases will be covered by our regular testing and infrastructure.

src/renderer/sw_engine/tvgSwFill.cpp
```diff
@@ -201,7 +202,17 @@ bool _prepareRadial(SwFill* fill, const RadialGradient* radial, const Matrix* tr
         fill->radial.fy = cy + r * (fy - cy) / dist;
         fill->radial.dx = cx - fill->radial.fx;
         fill->radial.dy = cy - fill->radial.fy;
+#if TVG_FLOAT_CONTRACT_AVAILABLE
         fill->radial.a = fill->radial.dr * fill->radial.dr - fill->radial.dx * fill->radial.dx - fill->radial.dy * fill->radial.dy;
+#else
+        // Prevent loss of precision on Apple Silicon when dr=dy and dx=0 due to FMA
```
hermet, line 208: I think we can replace the new logic with the old one. Maintaining both logics just increases the maintenance burden.

hermet:
+ it will be nicer if you add you in https://github.com/thorvg/thorvg/blob/main/AUTHORS

src/renderer/sw_engine/tvgSwFill.cpp
lmdsp: @hermet Yes I agree this would be simpler, however I'm not absolutely certain that this trick will guarantee that FMA isn't used on here future compilers or other platforms.
But if we have a test in place to catch this I suppose yes it's the simplest solution.

On another note I suspect this kind of incorrect optimization could affects other parts of the code, we'll need to keep an eye on this.

lmdsp:
> @lmdsp Yes, we are already conducting a specific comparison using the regression test and SVG examples. So far, the best approach is to push an SVG samples(src/examples/resources) to address those scenarios, after which the cases will be covered by our regular testing and infrastructure.

Great, thanks, I will add a simple sample for this then. Is there anything else I need to do besides adding the .svg image ?

hermet:
.svg is enough!

src/renderer/sw_engine/tvgSwFill.cpp
```diff
@@ -201,7 +202,17 @@ bool _prepareRadial(SwFill* fill, const RadialGradient* radial, const Matrix* tr
...
+        // https://github.com/thorvg/thorvg/issues/2014
```
hermet, line 204: one more, could you please remove this issue history? we can track the history by git blame with commit msgs.

src/renderer/sw_engine/tvgSwFill.cpp
```diff
@@ -201,7 +200,13 @@ bool _prepareRadial(SwFill* fill, const RadialGradient* radial, const Matrix* tr
         fill->radial.fy = cy + r * (fy - cy) / dist;
         fill->radial.dx = cx - fill->radial.fx;
         fill->radial.dy = cy - fill->radial.fy;
-        fill->radial.a = fill->radial.dr * fill->radial.dr - fill->radial.dx * fill->radial.dx - fill->radial.dy * fill->radial.dy;
+        // Prevent loss of precision on Apple Silicon when dr=dy and dx=0 due to FMA
+        // https://github.com/thorvg/thorvg/issues/2014
+        const float dr2 = fill->radial.dr * fill->radial.dr;
```
hermet, line 205: Neat better?
```
const float -> auto
```

src/renderer/sw_engine/tvgSwFill.cpp
```diff
@@ -234,7 +239,6 @@ bool _prepareRadial(SwFill* fill, const RadialGradient* radial, const Matrix* tr
     return true;
 }
 
-
```
hermet, line 237: please do not modify here.

src/renderer/sw_engine/tvgSwFill.cpp
```diff
@@ -172,7 +172,6 @@ bool _prepareLinear(SwFill* fill, const LinearGradient* linear, const Matrix* tr
...
```
hermet, line 175: please do not modify here

hermet (commented):
Thanks, overall LGTM, there a few nit-picks I'd like to suggest more.

src/renderer/sw_engine/tvgSwFill.cpp
lmdsp: Thanks for the insights, I've integrated you requests and squashed the history.

hermet (approved):

Issue #2029, opened 2024-03-06
Add c/c++ test suites for lottie slot 


hermet:
Cover the recently introduced APIs in unit tests.
```
thorvg/test/tvgLottie.cpp
thorvg/test/capi/capiLottie.cpp
```

PR #2030, opened 2024-03-06
test: Updated test code for the lottie slot

tinyjin:
This patch adds the test case for the recently introduced lottie slot API

Issue: #2029


test/capi/capiLottie.cpp
```diff
@@ -0,0 +1,56 @@
...
+    REQUIRE(animation);
+
+    Tvg_Paint* picture = tvg_animation_get_picture(animation);
+    REQUIRE(picture);
+
+    Tvg_Identifier id = TVG_IDENTIFIER_UNDEF;
+    REQUIRE(tvg_paint_get_identifier(picture, &id) == TVG_RESULT_SUCCESS);
+    REQUIRE(id == TVG_IDENTIFIER_PICTURE);
+
+    REQUIRE(tvg_picture_load(picture, TEST_DIR"/invalid.json") == TVG_RESULT_INVALID_ARGUMENT);
```
JSUYA, line 44: Invalid argument test for picture is already tested in capiAnimation.cpp, so this test line is unnecessary.

test/testLottie.cpp
```diff
@@ -0,0 +1,53 @@
...
+TEST_CASE("Lottie Slot", "[tvgLottie]")
+{
+    REQUIRE(Initializer::init(0) == Result::Success);
+
+    auto animation = LottieAnimation::gen();
+    REQUIRE(animation);
+
+    auto picture = animation->picture();
+    REQUIRE(picture->identifier == Picture::identifier);
+
```
hermet, line 44: Probably, you might wanna add a negative test at this point,
override() before load()

test/capi/capiLottie.cpp
```diff
@@ -0,0 +1,55 @@
...
+    Tvg_Animation* animation = tvg_lottie_animation_new();
+    REQUIRE(animation);
+
+    Tvg_Paint* picture = tvg_animation_get_picture(animation);
+    REQUIRE(picture);
+
+    Tvg_Identifier id = TVG_IDENTIFIER_UNDEF;
+    REQUIRE(tvg_paint_get_identifier(picture, &id) == TVG_RESULT_SUCCESS);
+    REQUIRE(id == TVG_IDENTIFIER_PICTURE);
+
```
hermet, line 43: Probably, you might wanna add a negative test at this point,
override() before load()

test/testLottie.cpp
```diff
@@ -0,0 +1,60 @@
...
+    //Slot override before loaded
+    REQUIRE(animation->override(slotJson) == Result::InsufficientCondition);
+
+    //Slot override
+    REQUIRE(picture->load(TEST_DIR"/lottieslot.json") == Result::Success);
+    REQUIRE(animation->override(slotJson) == Result::Success);
+
+    //Slot override with invalid JSON
+    REQUIRE(animation->override("") == Result::InsufficientCondition);
```
hermet, line 55: Shouldn't it be InvalidArgument?

tinyjin: Screenshot 2024-03-06 at 3 57 43 PM

For now, returns InsufficientCondition in invalid JSON, should we change the result?

hermet: Yes, we can change it if you think it makes sense.

tinyjin: agree, #2033

PR #2033, opened 2024-03-06
lottie/api: specify clear result for override API

tinyjin:
In case the given JSON is invalid, it's more closed to `Result::InvalidArguments`.


hermet:
@tinyjin could you please double-check the capi as well?

hermet (approved):

Issue #2006, opened 2024-02-23
static webp memory sanitizer detection


hermet:
memory misaligned here, ignorable or a potential issue?

![image](https://github.com/thorvg/thorvg/assets/3711518/7ee1134e-2e91-4b04-84f3-143e4b310fe9)

PR #2041, opened 2024-03-08
loader/webp: Remove WEBP_FORCE_ALIGNED and use memcpy() instead

JSUYA:
The google's libwebp source uses aligned memory access.
This patch is that applies the two commits below to our static lib code.

refer to:
[Remove WEBP_FORCE_ALIGNED and use memcpy() instead]
https://github.com/webmproject/libwebp/commit/3884972e3f9ae887e0c16ca342851eff3ae67485
[bit_reader.c: s/VP8L_USE_UNALIGNED_LOAD/VP8L_USE_FAST_LOAD/]
https://github.com/webmproject/libwebp/commit/ac49e4e4dc05d6dcd4083aaa41965713e42c1c35
source :
https://chromium.googlesource.com/webm/libwebp/+/refs/heads/main/src/utils/bit_reader_inl_utils.h#80

related issue: https://github.com/thorvg/thorvg/issues/2006

### DIFF
--- a/src/loaders/webp/utils/bit_reader.cpp
+++ b/src/loaders/webp/utils/bit_reader.cpp
@@ -16,6 +16,7 @@
 #endif
 
 #include "./bit_reader_inl.h"
+#include "./utils.h"
 
 //------------------------------------------------------------------------------
 // VP8BitReader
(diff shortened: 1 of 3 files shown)


src/loaders/webp/meson.build
```diff
@@ -1,5 +1,14 @@
 webp_deb = []
 
+config_h = configuration_data()
+
+config_h.set('WEBP_FORCE_ALIGNED', true)
```
hermet, line 5: Instead of making the script complex, how about we enable it directly in the header (bit reader). The point is this script is not a configurable at all.

hermet (commented):
Need to confirm the webp history when unaligned memory usage is valid.

src/loaders/webp/meson.build
JSUYA: Hi I updated this PR.
The google lib is no longer use unaligned memory access. so i found 2 commit related this. and i applied it.

hermet (approved):
cool, thanks!

Issue #2051, opened 2024-03-13
Storing the address of local variable


kisvegabor:
Hi,

While compiling ThorVG for ESP32 I get this warning (converted to error):

```
error: storing the address of local variable 'context' in '*this.LottieParser::context' [-Werror=dangling-pointer=]
 1255 |     this->context = &context;
```

hermet:
@kisvegabor hello, you might wanna confirm fix by this https://github.com/thorvg/thorvg/pull/2052

PR #2052, opened 2024-03-14
lottie: fixed a compiler warning:

hermet:
error: storing the address of local variable 'context' in '*this.LottieParser::context' [-Werror=dangling-pointer=]
 1255 |     this->context = &context;

issue: https://github.com/thorvg/thorvg/issues/2051

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -252,7 +252,7 @@ void LottieParser::getValue(ColorStop& color)
 {
     if (peekType() == kArrayType) enterArray();
 
-    color.input = new Array<float>(context->gradient->colorStops.count);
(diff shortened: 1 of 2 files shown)


hermet:
@qarmin Hello. The svg-regression test server recently experienced a breakdown. Could you please take a moment to review it? Thank you.

qarmin:
I was supposed to check what has changed in Thorvg a long time ago, but I couldn't find the time for it - I'll try to check the fuzzer and add appropriate images for the tests this week.

This issue seems to be more of a GitHub problem because the resource is still available and can be manually downloaded, so probably restarting CI will fix problem.

hermet:
@qarmin thanks, since on this topic, if we have additional SVG resources that we want to add to the tests, how can we contribute them in the official package?

kisvegabor:
Sorry, I'm late, but I confirm that this PR solves the issue.

PR #2057, opened 2024-03-14
Better regression testing

qarmin:
Few changes
- Update outdated actions(also needed to change artifacts name, because CI started to fail because were non unique)
- Fix path in regression finder - looks that path changed few months ago and tester silently ignored errors so CI was always green since that moment.
- Fix step with finding differences between current and develop branch - it should now find a lot of more broken files - previously most of them were ignored
- Rewrite regression commenting code into python
- Removed undefined/address flags from regression testing - I have with this flags strange freezes

I not changed here images that will be used in regression finding.

Currently such files are inside https://github.com/qarmin/SVG-regression-finder/releases/tag/0.3.0 release page, but I think that this files should be moved into this repository.

To add new files, such steps needs to be done:
- Downloading files from - https://github.com/qarmin/SVG-regression-finder/releases/download/0.3.0/ThorvgNotValidFiles.zip or https://github.com/qarmin/SVG-regression-finder/releases/download/0.3.0/ThorvgValidFiles.zip
- Unpacking them, putting new files inside/removing old and packing it  again
- Adding files to tag like - https://github.com/thorvg/thorvg/releases/tag/v0.12.7 or adding it to issues(like in video)

https://github.com/thorvg/thorvg/assets/41945903/69df3603-1fd0-4a4d-a858-c817d6c6f65e

- Copying link and replacing with it, link from `.github/workflows/regression.yml` file


.github/workflows/regression_check.py
```diff
@@ -0,0 +1,140 @@
...
+
+    print()
+    create_comment_v, fail_ci_v, github_comment_v = check_changes_in_valid_files()
+    print()
+    create_comment_i, fail_ci_i, github_comment_i = check_changes_in_invalid_files()
+    print()
+    create_comment_s, fail_ci_s, github_comment_s = check_changes_in_image_size()
+    print()
+
+    create_comment = create_comment_v or create_comment_i or create_comment_s
```
tinyjin, line 130: Is this checking whether `github_comment` created?
I guess `create_comment` is more like function name. The name `has_comment` would be cleaner :)

.github/workflows/regression_check.py
```diff
@@ -0,0 +1,140 @@
...
+    fail_ci = fail_ci_v or fail_ci_i or fail_ci_s
+    github_comment = github_comment_v + github_comment_i + github_comment_s
+
+    if fail_ci == 1:
+        # Just create file, we
```
tinyjin, line 135: minor thing:
comment is ambigious, after `we` you might wanna explain something else?

.github/workflows/regression_check.py
```diff
@@ -0,0 +1,140 @@
...
+    panics = []
+    possible_problems = []
+    with open(file_name, "r") as file:
+        print(f"Checking file {file_name}")
+        for line in file:
+            line = line.strip()
+            if RUST_PANIC in line:
+                panics.append(line)
+                continue
+            if POSSIBLE_PROBLEM in line:
```
tinyjin, line 29: we can save the line and make it readable with
```python
if RUST_PANIC in line:
    panics.append(line)
elif POSSIBLE_PROBLEM in line:
```

.github/workflows/regression_check.py
```diff
@@ -0,0 +1,140 @@
...
+    print()
+
+    create_comment = create_comment_v or create_comment_i or create_comment_s
+    fail_ci = fail_ci_v or fail_ci_i or fail_ci_s
+    github_comment = github_comment_v + github_comment_i + github_comment_s
+
+    if fail_ci == 1:
+        # Just create file, we
+        with open("EXIT_REQUESTED", "w") as file:
+            pass
```
tinyjin, line 137: To just create file, we don't need to use `with` and `pass` expressions.
```python
file = open("EXIT_REQUESTED", "w")
```

tinyjin (changes requested):
Hello, I love new regression testing code in python. And I've reviewed.
Please take them into consideration!

.github/workflows/regression_check.py
qarmin: Comment removed, because file is no longer created

hermet (approved):

tinyjin (approved):

Issue #2062, opened 2024-03-16
Embedded compilers doesn't like `_A`, `_B`, and `_C` function names.


kisvegabor:
Hi,

I'm not sure why, but I get an error for these 3 function names. Maybe these are reserved or used internally somewhere. I got this with error with multiple compilers.

```
/home/kisvegabor/projects/lvgl/esp-workspace/esp-bsp/examples/display_lvgl_demos/components/lvgl/src/libs/thorvg/tvgLottieInterpolator.cpp:44:21: error: expected unqualified-id before numeric constant
   44 | static inline float _B(float aA1, float aA2) { return 3.0f * aA2 - 6.0f * aA1; }
      |                     ^~
```
Renaming the function to `tvg_A`, `tvg_B`, and `tvg_C` solves the issue. Basically adding any letter before the `_` is fine.

PR #2063, opened 2024-03-17
lottie: Renamed function names with prefix

tinyjin:
Fixed naming conflict in short names.

issue: #2062

### DIFF
--- a/src/loaders/lottie/tvgLottieInterpolator.cpp
+++ b/src/loaders/lottie/tvgLottieInterpolator.cpp
@@ -40,20 +40,20 @@
 #define SUBDIVISION_MAX_ITERATIONS 10
 
 
-static inline float _A(float aA1, float aA2) { return 1.0f - 3.0f * aA2 + 3.0f * aA1; }
-static inline float _B(float aA1, float aA2) { return 3.0f * aA2 - 6.0f * aA1; }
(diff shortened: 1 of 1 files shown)


src/loaders/lottie/tvgLottieInterpolator.cpp
```diff
@@ -40,20 +40,20 @@
 #define SUBDIVISION_MAX_ITERATIONS 10
 
 
-static inline float _A(float aA1, float aA2) { return 1.0f - 3.0f * aA2 + 3.0f * aA1; }
-static inline float _B(float aA1, float aA2) { return 3.0f * aA2 - 6.0f * aA1; }
-static inline float _C(float aA1) { return 3.0f * aA1; }
+static inline float tvg_A(float aA1, float aA2) { return 1.0f - 3.0f * aA2 + 3.0f * aA1; }
```
hermet, line 43: The issue likely arises from the pre-processor converting A, B, and C into numeric values. I don't know about the compiler much, assuming the current name is too short for it.

Here is the static function name style guideline:
```
static float _funcName(..);

//so like:
_A(...) -> _constA(...)     //the macro is to determine the constant value of the equation.
_B(...) -> _constB(...)
_C(...) -> _constC(...)
```

kisvegabor (approved):

hermet (approved):

PR #2065, opened 2024-03-18
loader/svg: Remove duplicate if-check in pathAppendArcTo

JSUYA:
The if-check to skip-rule when drawing an arc path is already checked in line 476.
In addition, since the float type equal check is performed in the range of 1/256,
unintentional skiped may occur. Therefore, remove duplicate code. https://www.w3.org/TR/SVG2/paths.html#ArcOutOfRangeParameters

test file:
SVG_FILE_147893
SVG_FILE_147939


JSUYA:
Rebased branch

hermet:
@JSUYA  Could you please add SVG_FILE_147939.svg to our thorvg/src/examples/resources?

hermet (approved):

Issue #2070, opened 2024-03-18
Memory leak lottie2gif fails to converte file


qarmin:
Command
```
~/test/thorvg/build/src/tools/lottie2gif/lottie2gif ../Untitled\ Folder\ 3/about.json -r 300x300
```
json (real json, but not lottie file)

[about.json.zip](https://github.com/thorvg/thorvg/files/14642588/about.json.zip)

```
Failed Converting Gif file : ../Untitled Folder 3/manifest.json

=================================================================
==400865==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 120 byte(s) in 1 object(s) allocated from:
    #0 0x7f3aa04dfba8 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:95
    #1 0x7f3a9ff1e443 in LottieParser::parse() ../src/loaders/lottie/tvgLottieParser.cpp:1299
    #2 0x7f3a9fef6046 in LottieLoader::run(unsigned int) ../src/loaders/lottie/tvgLottieLoader.cpp:58
    #3 0x7f3a9fef2f33 in LottieLoader::header() ../src/loaders/lottie/tvgLottieLoader.cpp:92
    #4 0x7f3a9fdbd745 in LoaderMgr::loader(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool*) ../src/renderer/tvgLoader.cpp:315
    #5 0x7f3a9fddda8d in tvg::Picture::Impl::load(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ../src/renderer/tvgPicture.h:152
    #6 0x7f3a9fddda8d in tvg::Picture::load(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ../src/renderer/tvgPicture.cpp:175
    #7 0x55f051fd1aeb in App::convert(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) ../src/tools/lottie2gif/lottie2gif.cpp:74
    #8 0x55f051fd408c in App::convert(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) ../src/tools/lottie2gif/lottie2gif.cpp:102
[...]

PR #2071, opened 2024-03-19
lottie: Prevent memory leak when file is invalid

JSUYA:
When json file is invalid in the parser,
the LottieComposition object is not released and the parse() returns false.
To prevent memory leaks, free the memory before returning false.

related issue : https://github.com/thorvg/thorvg/issues/2070

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -207,7 +207,7 @@ float LottieLayer::remap(float frameNo)
 
 LottieComposition::~LottieComposition()
 {
-    if (!initiated) delete(root->scene);
+    if (!initiated && root) delete(root->scene);
 
     delete(root);
     free(version);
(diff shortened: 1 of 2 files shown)


src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -207,7 +207,7 @@ float LottieLayer::remap(float frameNo)
 
 LottieComposition::~LottieComposition()
 {
-    if (!initiated) delete(root->scene);
+    if (!initiated && root) delete(root->scene);
```
hermet, line 210: initialized shouldn't be true if the root is null. So it doesn't need to change this.

JSUYA: The case where parse() returns false is when json is invalid or there is no root layer. When initialized is false, there may be cases where the root layer is null. (1313 line in tvgLottieParser.cpp)
Without this change, root->scene will be called with root being null.

hermet: ok, got it. thanks

hermet (approved):

PR #2073, opened 2024-03-20
lottie/slot: Support reverting feature

tinyjin:
Support the slot reverting feature corresponding to the overriding.
`override(nullptr)` reverts all slots once called.

Usage:
```cpp
//Revert slot
if (animation->override(nullptr) == tvg::Result::Success) {
    canvas->update();
}
```

### DIFF
--- a/src/loaders/lottie/tvgLottieLoader.cpp
+++ b/src/loaders/lottie/tvgLottieLoader.cpp
@@ -302,7 +302,44 @@ Paint* LottieLoader::paint()
 
 bool LottieLoader::override(const char* slot)
 {
-    if (!slot || !comp || comp->slots.count == 0) return false;
+    if (!comp || comp->slots.count == 0) return false;
+
+    //reverting whole slots
+    if (!slot) {
+        if (!this->overriden) return true;
+
+        for (auto s = comp->slots.begin(); s < comp->slots.end(); ++s) {
+            for (auto pair = (*s)->pairs.begin(); pair < (*s)->pairs.end(); ++pair) {
+                if ((*pair).prop) {
+                    switch ((*s)->type) {
+                        case LottieProperty::Type::ColorStop: {
+                            auto target = static_cast<LottieGradient*>((*pair).obj);
+                            target->colorStops.release();
+                            target->colorStops = *static_cast<LottieColorStop*>((*pair).prop);
+                            break;
+                        }
+                        case LottieProperty::Type::Color: {
+                            auto target = static_cast<LottieSolid*>((*pair).obj);
+                            delete(target->color.frames);
+                            target->color = *static_cast<LottieColor*>((*pair).prop);
+                            break;
+                        }
+                        case LottieProperty::Type::TextDoc: {
+                            auto target = static_cast<LottieText*>((*pair).obj);
+                            target->doc.release();
+                            target->doc = *static_cast<LottieTextDoc*>((*pair).prop);
+                            break;
+                        }
+                        default: break;
+                    }
+                }
+            }
+        }
+
+        this->overriden = false;
+        return true;
+    }
+
 
     //TODO: Crashed, does this necessary?
     auto temp = strdup(slot);
@@ -324,6 +361,7 @@ bool LottieLoader::override(const char* slot)
     if (idx < 1) success = false;
 
     free(temp);
+    this->overriden = success;
     return success;
 }
 
--- a/src/loaders/lottie/tvgLottieModel.h
+++ b/src/loaders/lottie/tvgLottieModel.h
@@ -632,18 +632,47 @@ struct LottieLayer : LottieGroup
 
 struct LottieSlot
 {
+    struct Pair {
+        LottieObject* obj;
+        LottieProperty* prop;
+
+        void override(LottieObject* target, LottieProperty::Type type)
+        {
+            delete(prop);
+
+            switch (type) {
+                case LottieProperty::Type::ColorStop: {
+                    prop = static_cast<LottieGradient*>(obj)->colorStops.copy();
+                    break;
+                }
+                case LottieProperty::Type::Color: {
+                    prop = static_cast<LottieSolid*>(obj)->color.copy();
+                    break;
+                }
+                case LottieProperty::Type::TextDoc: {
+                    prop = static_cast<LottieText*>(obj)->doc.copy();
+                    break;
+                }
+                default: break;
+            }
+
+            obj->override(target);
+        }
+    };
+
     char* sid;
-    Array<LottieObject*> objs;
+    Array<Pair> pairs;
     LottieProperty::Type type;
 
     LottieSlot(char* sid, LottieObject* obj, LottieProperty::Type type) : sid(sid), type(type)
     {
-        objs.push(obj);
+        pairs.push({obj});
     }
 
     ~LottieSlot()
     {
         free(sid);
+        for (auto p = pairs.begin(); p < pairs.end(); ++p) delete((*p).prop);
     }
 };
 
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -478,7 +478,7 @@ void LottieParser::parseProperty(T& prop, LottieObject* obj)
             //append object if the slot already exists.
             for (auto slot = comp->slots.begin(); slot < comp->slots.end(); ++slot) {
                 if (strcmp((*slot)->sid, sid)) continue;
-                (*slot)->objs.push(obj);
+                (*slot)->pairs.push({obj});
                 return;
             }
             comp->slots.push(new LottieSlot(sid, obj, type));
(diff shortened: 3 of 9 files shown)


src/examples/LottieExtension.cpp
```diff
@@ -85,6 +85,11 @@ void tvgDrawCmds(tvg::Canvas* canvas)
         cout << "Failed to override the slot" << endl;
     }
 
+    //Revert slot
```
hermet, line 88: Rather than example, I think it's better to include this minor showcase in the unit tests.

hermet (commented):
Reviewed "examples: Added lottie slot reverting example"

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -302,7 +302,18 @@ Paint* LottieLoader::paint()
 
 bool LottieLoader::override(const char* slot)
 {
-    if (!slot || !comp || comp->slots.count == 0) return false;
+    if (!comp || comp->slots.count == 0) return false;
+
+    //reverting whole slots
```
hermet, line 307: Let's say, if there hasn't been any overriding before, there's no need to reset at all.
```
bool LottieLoader::override(const char* slot) {
   if (!comp || comp->slots.count == 0) return false;
   if (!this->overridden) return true;
```

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -97,6 +97,32 @@ struct LottieMask
 };
 
 
+struct LottieObject;
```
hermet, line 100: please put this on 34 line.

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -130,6 +156,16 @@ struct LottieObject
         TVGERR("LOTTIE", "Unsupported slot type");
     }
 
+    virtual void copy(LottieSlot::Pair* pair)
```
hermet, line 159: It's not a good idea because LottieObject (a pure base class) depends on the LottieSlot (an extension) feature. Even copying/reverting doesn't necessarily require a virtual and overriding approach, as the implementation can be common.

See: 
```
template<typename T>
assign(LottieSlot:Pair *pair, T& prop)
{
    delete(pair->prop);
    pair->prop = prop.copy();
}

File: src/loaders/lottie/tvgLottieModel.h [hunk 5]
```diff
@@ -194,6 +230,21 @@ struct LottieText : LottieObject
         this->prepare();
     }
 
+    void copy(LottieSlot::Pair* pair) override
+    {
+        if (pair->prop) delete(pair->prop);
```
hermet, line 235: no need to check:
```
if (pair->prop) 
```

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -194,6 +230,21 @@ struct LottieText : LottieObject
...
+        pair->prop = this->doc.copy();
+    }
+
+    void revert(LottieProperty* prop) override
+    {
+        this->doc.release();
+
+        auto target = static_cast<LottieTextDoc*>(prop);
```
hermet, line 243: Probably this works:
```
this->doc = target;
```

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1278,8 +1278,9 @@ bool LottieParser::parse(LottieSlot* slot)
     if (!obj || Invalid()) return false;
 
     //apply slot object to all targets
-    for (auto target = slot->objs.begin(); target < slot->objs.end(); ++target) {
-        (*target)->override(obj);
+    for (auto target = slot->pairs.begin(); target < slot->pairs.end(); ++target) {
+        (*target).obj->copy(target);
```
hermet, line 1282: ```(*target).obj -> target->obj```

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -451,10 +459,26 @@ struct LottieColorStop : LottieProperty
         fill->colorStops(result.data, count);
     }
 
+    LottieColorStop* copy()
+    {
+        //copy, used for slot reverting
+        auto ret = new LottieColorStop();
+        if (frames) ret->frames = frames;
+        else {
+            ret->value = {};
```
hermet, line 468: ret->value{ value.data, };

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -249,15 +249,14 @@ struct LottieGenericProperty : LottieProperty
         return frame->interpolate(frame + 1, frameNo);
     }
 
-    T& operator=(const T& other)
```
hermet, line 252: I couldn't see any differences between before and after. 
Could you please explain why this [operator -> shallowCopy] is necessary?

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -249,15 +249,14 @@ struct LottieGenericProperty : LottieProperty
...
+    void shallowCopy (const LottieGenericProperty<T>& other)
     {
         //shallow copy, used for slot overriding
         delete(frames);
```
hermet: You should not delete the actual property data for backup(reverting) ??

hermet (commented):
reviewed "lottie/slot: Support reverting feature"

hermet:
Please check comments:
1. please update the api doc(header) for reverting case.
2. please add unit-test for reverting scenario.
3. please double-check any memory leaks detection with: 
```
$meson -Db_sanitize="address,undefined" ...          //enable the memory violation detector
```

test/capi/capiLottie.cpp
```diff
@@ -50,6 +50,12 @@ TEST_CASE("Lottie Slot", "[capiLottie]")
     REQUIRE(tvg_picture_load(picture, TEST_DIR"/lottieslot.json") == TVG_RESULT_SUCCESS);
     REQUIRE(tvg_lottie_animation_override(animation, slotJson) == TVG_RESULT_SUCCESS);
 
+    //Slot revert
+    REQUIRE(tvg_lottie_animation_override(animation, nullptr) == TVG_RESULT_SUCCESS);
```
hermet, line 54: probably, you can add this as well before the first slot overriding call (45 line?)

src/loaders/lottie/tvgLottieProperty.h
tinyjin: Yes. It's the revising operator commit.
The deletion in shallow copy case has been removed in next commit.

src/loaders/lottie/tvgLottieProperty.h
tinyjin: I also see it's not necessary in current code.
Will drop the commit.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -40,6 +40,28 @@ static char* _int2str(int num)
 }
 
 
+static void _assign(LottieSlot::Pair *pair, LottieProperty::Type type)
+{
+    delete(pair->prop);
+
+    switch (type) {
+        case LottieProperty::Type::ColorStop: {
+            pair->prop = static_cast<LottieGradient*>(pair->obj)->colorStops.copy();
```
hermet, line 49: it looks copy() doesn't necessary:
```
pair->prop = &static_cast<LottieGradient*>(pair->obj)->colorStops;
```

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -249,10 +249,18 @@ struct LottieGenericProperty : LottieProperty
...
+        auto ret = new LottieGenericProperty<T>(value);
+        if (frames) ret->frames = frames;
+        else ret->value = value;
+        return ret;
+    }
+
     T& operator=(const T& other)
     {
         //shallow copy, used for slot overriding
-        delete(frames);
```
hermet, line 255: Simply removing the memory deallocation will result in memory leaks. I think you should not remove them once the review changes have been made based on the feedback.

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -302,7 +302,22 @@ Paint* LottieLoader::paint()
...
+    if (!slot) {
+        if (!this->overriden) return true;
+
+        for (auto s = comp->slots.begin(); s < comp->slots.end(); ++s) {
+            for (auto pair = (*s)->pairs.begin(); pair < (*s)->pairs.end(); ++pair) {
+                if ((*pair).prop) (*pair).obj->revert((*pair).prop);
```
hermet, line 313: like the assign() with my comments, you can just revert the value. (it would not require the new revert functions)

hermet (changes requested):
Please check the comments. 
Note that, the memory handles are not being managed properly.

src/loaders/lottie/tvgLottieParser.cpp
tinyjin: Without copy(), I think the shallow copy operator will change the value, and it will effect slot reverting in this code.
Or would changing shallow copy operator to deep copy approach be better?

test/capi/capiLottie.cpp
```diff
@@ -42,12 +42,23 @@ TEST_CASE("Lottie Slot", "[capiLottie]")
     REQUIRE(id == TVG_IDENTIFIER_PICTURE);
 
     const char* slotJson = R"({"gradient_fill":{"p":{"a":0,"k":[0,0.1,0.1,0.2,1,1,0.1,0.2,0.1,1]}}})";
-
+
```
hermet, line 45: -- white space

hermet (changes requested):
No sense to add this as a separate patch. 
Please squash this patch with 'lottie: Support the slot reverting feature'.

hermet:
https://github.com/thorvg/thorvg/pull/2077

Issue #2078, opened 2024-03-21
svg2png: stack overflow and crash


qarmin:
```
svg2png file.svg -r 300x300
```
files - [thorvg.zip](https://github.com/thorvg/thorvg/files/14693550/thorvg.zip)

```
../src/renderer/sw_engine/tvgSwRle.cpp:716:31: runtime error: member access within null pointer of type 'struct SwOutline'
AddressSanitizer:DEADLYSIGNAL
=================================================================
==71421==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000010 (pc 0x7fdd12e62613 bp 0x7fdd0cbf54e0 sp 0x7fdd0cbf52a0 T5)
==71421==The signal is caused by a READ memory access.
==71421==Hint: address points to the zero page.
    #0 0x7fdd12e62613 in tvg::Array<unsigned int>::begin() ../src/common/tvgArray.h:100
    #1 0x7fdd12e62613 in _decomposeOutline ../src/renderer/sw_engine/tvgSwRle.cpp:716
    #2 0x7fdd12e62613 in _genRle ../src/renderer/sw_engine/tvgSwRle.cpp:755
    #3 0x7fdd12e64572 in rleRender(SwRleData*, SwOutline const*, SwBBox const&, bool) ../src/renderer/sw_engine/tvgSwRle.cpp:989
    #4 0x7fdd12e67a79 in shapeGenRle(SwShape*, tvg::RenderShape const*, bool) ../src/renderer/sw_engine/tvgSwShape.cpp:477
    #5 0x7fdd12e588e1 in SwShapeTask::run(unsigned int) ../src/renderer/sw_engine/tvgSwRenderer.cpp:153
    #6 0x7fdd12e07a83 in tvg::Task::operator()(unsigned int) ../src/renderer/tvgTaskScheduler.h:64
    #7 0x7fdd12e09fbb in tvg::TaskSchedulerImpl::run(unsigned int) (/home/rafal/test/thorvg/build/src/tools/svg2png/../../libthorvg.so.0+0x409fbb) (BuildId: c671918882fb3c4597fa585d8556c27a6c92e052)
    #8 0x7fdd126e6332  (/lib/x86_64-linux-gnu/libstdc++.so.6+0xe6332) (BuildId: 102661775a59c09123f226bc3021556fd42bc563)
    #9 0x7fdd11a97b59 in start_thread nptl/pthread_create.c:444
[...]

Issue #494, opened 2021-06-23
Stack overflow when trying to convert image


qarmin:
[e-clipPath-029.svg.zip](https://github.com/Samsung/thorvg/files/6699341/e-clipPath-029.svg.zip)
```
==21929==ERROR: AddressSanitizer: stack-overflow on address 0x7f58395ffe68 (pc 0x7f583e858728 bp 0x7f58396006b0 sp 0x7f58395ffe50 T1)
    #0 0x7f583e858727 in __interceptor_free (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x10d727)
    #1 0x7f583d5dacaa in setlocale (/lib/x86_64-linux-gnu/libc.so.6+0x33caa)
    #2 0x7f583e7cad73  (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x7fd73)
    #3 0x7f583e5496d4 in svgPathToTvgPath(char const*, tvg::Array<tvg::PathCommand>&, tvg::Array<tvg::Point>&) ../src/loaders/svg/tvgSvgPath.cpp:521
    #4 0x7f583e550ffa in _appendShape ../src/loaders/svg/tvgSvgSceneBuilder.cpp:281
    #5 0x7f583e55472b in _appendChildShape ../src/loaders/svg/tvgSvgSceneBuilder.cpp:159
    #6 0x7f583e55090e in _applyComposition ../src/loaders/svg/tvgSvgSceneBuilder.cpp:187
    #7 0x7f583e5545d9 in _applyProperty ../src/loaders/svg/tvgSvgSceneBuilder.cpp:261
    #8 0x7f583e5545d9 in _appendShape ../src/loaders/svg/tvgSvgSceneBuilder.cpp:326
    #9 0x7f583e55472b in _appendChildShape ../src/loaders/svg/tvgSvgSceneBuilder.cpp:159
    #10 0x7f583e55090e in _applyComposition ../src/loaders/svg/tvgSvgSceneBuilder.cpp:187
    #11 0x7f583e5545d9 in _applyProperty ../src/loaders/svg/tvgSvgSceneBuilder.cpp:261
    #12 0x7f583e5545d9 in _appendShape ../src/loaders/svg/tvgSvgSceneBuilder.cpp:326
    #13 0x7f583e55472b in _appendChildShape ../src/loaders/svg/tvgSvgSceneBuilder.cpp:159
[...]

PR #2079, opened 2024-03-21
loader/svg: Check whether href id of use is parent or not

JSUYA:
When finding the 'nodeFrom' referenced by `<use>`,
if it is the parent, it is not referenced.

https://www.w3.org/TR/SVG2/struct.html#UseElement
Specification:
If the referenced element is a (shadow-including) ancestor of the ‘use’ element,
then this is an invalid circular reference and the ‘use’ element is in error.

related issue: https://github.com/thorvg/thorvg/issues/2078 SVG_FILE_65171.svg
SVG_FILE_65172.svg

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1919,6 +1919,19 @@ static SvgNode* _findNodeById(SvgNode *node, const char* id)
 }
 
 
+static SvgNode* _findParentById(SvgNode* node, char* id, SvgNode* doc)
+{
+    SvgNode *parent = node->parent;
+    while (parent != nullptr && parent != doc) {
+        if (parent->id && !strcmp(parent->id, id)) {
+            return parent;
+        }
+        parent = parent->parent;
+    }
+    return nullptr;
+}
+
+
 static constexpr struct
 {
     const char* tag;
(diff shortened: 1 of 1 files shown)


hermet:
_findNodeById() is used elsewhere (_updateComposite()). You might wanna check for any introduced circular dependency inside _findNodeById() to ensure it addresses any potential issues in one place. Also, we could probably log a message in the error scenario to notify the user that the SVG is invalid data.

hermet:
Perhaps, using a bit-flag or a boolean could be a better option to prevent unnecessary traversing just for error-checking validation. See: https://github.com/thorvg/thorvg/issues/494

JSUYA:
> _findNodeById() is used elsewhere (_updateComposite()). You might wanna check for any introduced circular dependency inside _findNodeById() to ensure it addresses any potential issues in one place. Also, we could probably log a message in the error scenario to notify the user that the SVG is invalid data.

I think this case is different from finding circular references. The purpose of the `<use>` spec is to determine whether something is a parent or not before making a circular reference. Also, `Infinite recursive call` is checked through `depth` check of _cloneNode(). findNodeById() takes a long time as it searches top down. It is not suitable for finding and checking parent. If I modify findNodeById() for that, it will have a big impact on other functions.

hermet:
Ok, rather than "circular dependency", I actually meat to _findParentById()" this one. basically, idea is not different.

hermet (approved):

Issue #2044, opened 2024-03-11
Lottie: Animation Segment(Marker) Support


hermet:
The concept of using markers in Lottie animations is known, where markers are used to control animation playback by specifying start and end points without needing to know the exact frame numbers.

Sample:
[marker.json](https://github.com/thorvg/thorvg/files/14552673/marker.json)

API usage:
```
lottie->load("sample.json");

lottie->segment("sectionA") ;                    //specify the SectionA marked by Lottie spec.
auto frame = lottie->frameCnt();   //returns the "SectionA" frame count
auto duration = lottie->duration();  //returns the SectionA duration in seconds
...

lottie->segment(begin: 0.3, end: 0.7);        //specify the segment 0.3 ~ 0.7 ?
auto frame = lottie->frameCnt();       //returns the 0.3 ~ 0.7 frame count
auto duration = lottie->duration();    //returns the 0.3 ~ 0.7 duration in seconds
...

**Comment by hermet** [comment]
Acquire Marker List
```
uint32_t LottieAnimation::markerCnt();  //return the marker count
const char* LottieAnimation::markers(uint32_t idx); //return the marker name by a given index.
```

**Comment by fall2019** [comment]
Okay. I will add it.

**Comment by hermet** [comment]
@theashraf

> * Is there a way to determine the beginning and ending frame number of a given marker ? cause we have a use case where the user want to know the segment value for UI purposes

Makes sense. Probably, thorvg could provide an extra api:
```
//Return the begin, end values of the given marker.
Result LottieAnimation::segment(const char* marker, float* begin, float* end)
```

> * Would applying a segment or a marker update the animation total_frames and duration information we get from `tvg_get_total_frames` and `tvg_get_duration` apis ?

Yes.


I anticipate that dotLottie will replace the segment with the marker name. Could you please clarify in what scenarios dotLottie requires the use of this API?
```
Result LottieAnimation::segment(float begin, float end);  //Set the segment with a specific range.
```

**Comment by theashraf** [comment]
@hermet

In dotLottie, we offer support for both named segments (Markers) and standard segments. So, the following API would be beneficial:

```cpp
Result LottieAnimation::segment(float begin, float end);
```

This is useful in scenarios where users want to specify a custom segment for playback that does not have an associated named marker within the Lottie JSON

PR #2080, opened 2024-03-21
Lottie: Animation Segment(Marker) Support

fall2019:
issue: https://github.com/thorvg/thorvg/issues/2044


hermet:
@fall2019, could you please squash your unnecessary commits to the one?

fall2019:
Okay. I usually use squash on merge option. Will try to squash later.

fall2019:
Squashed all commits into one.

SergeyLebedkin (approved):

inc/thorvg.h
```diff
@@ -5,6 +5,7 @@
 #include <memory>
 #include <string>
 #include <list>
+#include <string_view>
```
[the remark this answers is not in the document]
fall2019: Oh maybe come changes were lost. I just forced pushed some commits.Thanks for reminding me.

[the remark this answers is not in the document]
fall2019: Just checked files. Maybe it's in an outdated version of this branch. Right now there is no string_view in files.

tinyjin (changes requested):
Hello, Thank you so much for the efforts.
We'd love to have your patches.

Please check these comments:

**1. Conflict Resolution**
- `src/loaders/lottie/tvgLottieLoader.h` currently has a merge conflict.
- Please ensure that the patch is conflict-free before merging.

**2. Enhance Commit Message Clarity**
- The current message simply states "POC"
- It would be beneficial to provide a more descriptive message that outlines the purpose of these changes.
- We strongly recommend referring to this [guidance](https://github.com/thorvg/thorvg/blob/main/CONTRIBUTING.md#commit-message) for composing meaningful commit messages.

**3. Improve Commit Description Formatting:**
- The commit description appears unclear.
- It seems to be just squashed without tidying the explanations, there are a lot of typos or unnecessary texts.
- Please check again the description is clear and concise.
Screenshot 2024-03-22 at 7 32 49 PM

fall2019:
Okay. Thanks for reviewing and your article is really helpful. Will resolve the comments later.

inc/thorvg.h
```diff
@@ -1341,6 +1341,53 @@ class TVG_API Picture final : public Paint
...
+    /**
+     * @brief Return segment duration
+     *
+     * @param[out] dur segment duration
+     *
+     * @note Will call sync()
+     *
+     * @note Experimental API
+     */
+    Result duration(float&dur);
```
hermet, line 1379: Using `duration()` and `frameCount()` wouldn't be necessary because the Animation class already provides these APIs. Users could utilize the existing ones instead of introducing new ones. Even these are not proper to Picture class.

inc/thorvg.h
```diff
@@ -1341,6 +1341,53 @@ class TVG_API Picture final : public Paint
...
+    /**
+     * @brief Specify segment by marker
+     *
+     * @param[out] marker segment name
+     *
+     * @note Set empty to reset specified segment
+     *
+     * @note Experimental API
+     */
+    Result segment(const std::string& marker);
```
hermet, line 1355: 1. This segment API should be moved to the class LottieAnimation in thorvg_lottie.h
2. thorvg would not use string, please replace the string with const char*

inc/thorvg.h
```diff
@@ -1341,6 +1341,53 @@ class TVG_API Picture final : public Paint
...
+     *
+     * @param[in] begin  begin position
+     *
+     * @param[in] end  end position
+     *
+     * @note range from 0.0 to 1.0
+     *
+     * @note Experimental API
+     */
+    Result segment(float begin, float end);
```
hermet, line 1368: This segment API should be moved to the class Animation in thorvg.h

hermet (changes requested):
Thanks for your contribution, quickly examined the API design.

fall2019:
You're welcome. Will try to resolve soon.

fall2019:
Resolved all comments.

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -345,7 +345,23 @@ bool LottieLoader::frame(float no)
 
 float LottieLoader::totalFrame()
 {
-    return frameCnt;
+    if(this->selectedSection==NONE)
+        return frameCnt;
+    if(!this->comp)
+        sync();
+    if(this->selectedSection==MARKER)
```
tinyjin, line 352: `this->` is not mandatory in this context. Would be better having `this` as less as possible.

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -354,12 +370,43 @@ float LottieLoader::curFrame()
...
+        float frames = frameCnt * distance;
+        return frames/this->comp->frameRate;
+    }
+    return -1.0f;
+}
+
+void LottieLoader::segment(const char* marker)
+{
+    this->marker = marker;
+    if(marker==nullptr)
```
tinyjin, line 399: We can simply check `nullptr` with modern way:
```cpp
// It'd be readable
if (!marker)
```

src/loaders/lottie/tvgLottieLoader.h
```diff
@@ -29,7 +29,7 @@
 
 struct LottieComposition;
 struct LottieBuilder;
-
+enum SelectedSegment{NONE,MARKER,RANGE};
```
tinyjin, line 32: The enumeration is only needed with marker. Maybe we can define them in LottieMarker?
You can refer to the other model defination in `tvgLottieModel.h`
```cpp
struct LottieMarker
{
    enum Segment : uint8_t { None = 0, Marker, Range };

    char* name = nullptr;
    int time = 0;
    int duration = 0;

    ~LottieMarker()
    {
        free(name);
    }
};
```

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -676,6 +691,7 @@ struct LottieComposition
     Array<LottieInterpolator*> interpolators;
     Array<LottieFont*> fonts;
     Array<LottieSlot*> slots;
+    std::map<const char*,LottieMarker*,StrCompare> markers;
```
tinyjin, line 694: Should we really have `std::map` for this?
Using 

`StrCompare` only provides a comparison function, I think we don't have to use in struct.
We can express like?:
```cpp
// Matching parameters and helper functions into `LottieMarker`
Array<LottieMarker*> markers;
```

tinyjin: Please double-check binary size as well if you use new library.

```sh
cd ./build/src
strip libthorvg.so
size libthorvg.so
# compare output size with latest version of thorvg binary
```

inc/thorvg.h
```diff
@@ -1859,6 +1859,30 @@ class TVG_API Animation
      * @% Experimental API
      */
     float duration() const noexcept;
+    
+    /**
+     * @brief Specify segment by marker
+     *
+     * @param[out] marker segment name
+     *
+     * @note Set nullptr to reset specified segment
```
tinyjin, line 1868:
```cpp
@note Set @c nullptr to reset specified segment.
```

tinyjin (changes requested):
Simply reviewed in code consistency.
Do apply same convention to the other lines even not mentioned.

src/loaders/lottie/tvgLottieModel.h
fall2019: Resolved. But I think It's common for modern computers to have gigabytes or even terabytes of storage. Why we care about binary size? Just curious.😂

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -345,7 +345,23 @@ bool LottieLoader::frame(float no)
 
 float LottieLoader::totalFrame()
 {
-    return frameCnt;
+    if(this->selectedSection==NONE)
```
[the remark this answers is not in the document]
fall2019: Resolved. Maybe we could use smt like [clang-format](https://releases.llvm.org/14.0.0/tools/clang/docs/ClangFormat.html) in the future?

src/loaders/lottie/tvgLottieModel.h
hermet: @fall2019 ThorVG has been also considered for the web module, even for small-sized embedded IoT devices.
@tinyjin maintainers can do this, contributors don't need to be bothered by this.

tinyjin: Ok, @fall2019 Never mind the binary size check!
But just fyi, the size is really important in the reason Hermet said ;)

fall2019: Yeah. That makes sense.

inc/thorvg.h
```diff
@@ -1859,6 +1859,30 @@ class TVG_API Animation
...
+    /**
+     * @brief Specify segment by marker
+     *
+     * @param[out] marker segment name
+     *
+     * @note Set @c nullptr to reset specified segment
+     *
+     * @note Experimental API
+     */
+    void segment(const char* marker);
```
hermet, line 1872: This api should be in src/loaders/lottie/thorvg_lottie.h

hermet (commented):
api description is too weak. probably you can describe about marker / segment methods... at least definition of the marker should be described there.

i.e)
"Markers are used to control animation playback by specifying start and end points, eliminating the need to know the exact frame numbers. Generally, markers are designated at the design level, meaning the callers must know the marker name in advance to use it."

.gitignore
```diff
@@ -7,3 +7,4 @@ doc
 *.tvg
 *.tmp
 *.code-workspace
+.DS_Store
```
tinyjin, line 10: It shouldn't be included in this patch? (unrelated with marker)

src/renderer/tvgAnimation.cpp
```diff
@@ -92,6 +92,25 @@ float Animation::duration() const noexcept
     return static_cast<FrameModule*>(loader)->duration();
 }
 
+void Animation::segment(float begin, float end)
+{
+    auto loader = pImpl->picture->pImpl->loader;
+
+    if (!loader) return;
+    if (!loader->animatable()) return;
```
tinyjin, line 100: `if (!loader || !loader->animatable()) return;`

src/renderer/tvgAnimation.cpp
```diff
@@ -92,6 +92,25 @@ float Animation::duration() const noexcept
...
+
+    static_cast<FrameModule*>(loader)->segment(begin,end);
+}
+
+void Animation::segment(const char* marker)
+{
+    auto loader = pImpl->picture->pImpl->loader;
+
+    if (!loader) return;
+    if (!loader->animatable()) return;
```
tinyjin, line 110: `if (!loader || !loader->animatable()) return;`

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -180,6 +180,19 @@ struct LottieFont
     Origin origin = Embedded;
 };
 
+struct LottieMarker
+{
+    enum SelectedSegment:std::uint8_t { NONE = 0, MARKER, RANGE };
```
tinyjin, line 185: `enum SelectedSegment : uint8_t { NONE = 0, MARKER, RANGE };`

(Plus) I think it doesn't have state (whether it's selected or not) it self in this context. So the prefix `Selected` could be ambigious.

src/renderer/tvgPicture.cpp
```diff
@@ -20,6 +20,7 @@
  * SOFTWARE.
  */
 
+#include "tvgFrameModule.h"
```
tinyjin, line 23: Probably, it's not necessary?

src/loaders/lottie/tvgLottieLoader.h
```diff
@@ -39,6 +39,11 @@ class LottieLoader : public FrameModule, public Task
     float frameCnt = 0.0f;
     float frameDuration = 0.0f;
 
+    float selectedSegmentBegin=0.0f;
+    float selectedSegmentEnd=0.0f;
```
tinyjin, line 43: operand spacing

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -943,6 +942,31 @@ void LottieParser::parseAssets()
...
+    }
+    
+    return marker;
+}
+
+void LottieParser::parseMarkers()
+{
+    enterArray();
+    while (nextArrayValue())
+    {
```
tinyjin, line 965: You can check original code style for while-loop.
`while (nextArrayValue()) {` (no newline in brace)

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -943,6 +942,31 @@ void LottieParser::parseAssets()
...
+        auto marker = parseMarker();
+        comp->markers.push(marker);
```
tinyjin, line 967: Make a line?
`comp->markers.push(parseMarker());`

src/examples/Animation.cpp
```diff
@@ -58,6 +62,13 @@ void tvgDrawCmds(tvg::Canvas* canvas)
         cout << "Lottie is not supported. Did you enable Lottie Loader?" << endl;
         return;
     }
+    
+    frameRate=animation->totalFrame() / animation->duration();
```
tinyjin, line 66: spacing.
`frameRate = animation->totalFrame() / animation->duration();`

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -345,7 +345,19 @@ bool LottieLoader::frame(float no)
...
+    {
+        for(auto it : comp->markers)
+            if(std::strcmp(selectedMarker.c_str(), it->name))
+                return it->time;
+    }
+    else if(selectedSection == LottieMarker::SelectedSegment::RANGE)
+    {
+        return frameCnt * selectedSegmentBegin;
+    }
+    return -1.0f;
```
tinyjin, line 360: For enumeration condition, rather than `if-else`, `switch-case` would be readable.

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -354,12 +366,38 @@ float LottieLoader::curFrame()
...
+        float frames = frameCnt * distance;
+        return frames / comp->frameRate;
+    }
+    return -1.0f;
+}
+
+void LottieLoader::segment(const char* marker)
+{
+    selectedMarker = marker;
+    if(marker == nullptr) selectedSection = LottieMarker::SelectedSegment::NONE;
```
tinyjin, line 391: `if (!marker)`

tinyjin (changes requested):
Thanks for previous changes. Reviewed again.

fall2019:
Resolved comments.

fall2019:
Okay. Thanks for comments. Will resolve soon.  🤔It has been a mystery for me of how others come up with same format style. I didn't find formatter config in the repo so I just manually format it one after another.

tinyjin:
@fall2019 Well, currently we don't have specific convention(formatter), but we're trying to keep as consistent style as possible.

inc/thorvg.h
```diff
@@ -1848,6 +1848,15 @@ class TVG_API Animation
...
     float totalFrame() const noexcept;
+    
+    /**
+     * @brief Retrieves the start frame number specified by segment.
+     *
+     * @return Start frame number specified by segment. If there is no segment returns first frame;
+     *
+     * @note Experimental API
+     */
+    float startFrame() const noexcept;
```
hermet, line 1859: I'd like to say, the startFrame() API is not necessary. Once a segment is established, the frame number could be internally adjusted by the engine (lottie loader), eliminating the need for users to handle this directly.

Here's an overview of related APIs:
```
float Animation::totalFrame()  //Returns the total number of frames in the segment.
float Animation::duration()     //Returns the duration of the segment.
float Animation::curFrame()   //Returns the current frame number, remapped to the segment (i.e., the current frame number - the segment's starting frame number).
float Animation::frame(float no)  //Remaps the frame number (no) to the segment(i.e., @p no + the segment's starting frame number).
```

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -180,6 +180,16 @@ struct LottieFont
     Origin origin = Embedded;
 };
 
+struct LottieMarker
+{
+    char* name = nullptr;
+    int time = 0;
```
hermet, line 186: I guess time/duration should be float type.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -19,7 +19,6 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-
```
hermet, line 22: Please do not touch unnecessary lines.

src/loaders/lottie/tvgLottieAnimation.cpp
```diff
@@ -54,3 +54,20 @@ unique_ptr<LottieAnimation> LottieAnimation::gen() noexcept
...
     return unique_ptr<LottieAnimation>(new LottieAnimation);
 }
+
+bool LottieAnimation::segment(const char* marker)
+{
+    auto loader = pImpl->picture->pImpl->loader;
+    if (!loader || !loader->animatable()) return false;
+    
+    if (!marker) {
+        static_cast<FrameModule*>(loader)->segment(0.0, 1.0);
```
hermet, line 64: It might be necessary to cast directly to LottieLoader rather than to FrameModule.

src/renderer/tvgFrameModule.h
```diff
@@ -38,7 +38,9 @@ class FrameModule: public ImageLoader
     virtual float totalFrame() = 0;         //return the total frame count
     virtual float curFrame() = 0;           //return the current frame number
     virtual float duration() = 0;           //return the animation duration in seconds
-
+    virtual float startFrame() { return 0.0f; }; //return start frame
+    virtual bool getSegment(float& beign, float& end, const char* marker = nullptr) { return false; }
```
hermet, line 42: The marker is a lottie specific concept. It should not be in this FrameModule interface. You might wanna add add those necessary interfaces in the LottieLoader

hermet (changes requested):
Please check more feedbacks. Thanks for your contribution!

fall2019:
Thanks for comments. All resolved rn.

src/loaders/lottie/tvgLottieLoader.h
```diff
@@ -26,10 +26,10 @@
 #include "tvgCommon.h"
 #include "tvgFrameModule.h"
 #include "tvgTaskScheduler.h"
+#include "tvgLottieModel.h"
```
tinyjin, line 29: This import statement seems useless in new code lines.
Could you check?

src/loaders/lottie/tvgLottieLoader.h
```diff
@@ -26,10 +26,10 @@
...
 
 struct LottieComposition;
 struct LottieBuilder;
-
```
tinyjin, line 32: You should avoid pointless code line touch

examples/LottieExtension.cpp
```diff
@@ -24,71 +24,105 @@
...
 /************************************************************************/
 
-static unique_ptr<tvg::LottieAnimation> animation;
-static Elm_Transit *transit;
-static bool updated = false;
+static unique_ptr<tvg::LottieAnimation> slotAnimation;
+static unique_ptr<tvg::LottieAnimation> markerAnimation;
+static Elm_Transit *slotTransit;
+static Elm_Transit *markerTransit;
+static int updated = 0;
```
tinyjin, line 31: `LottieExtension` will have been keeping Lottie specified features here.
These definitions seem heavy, it will definitely need duplicated code.

Can we rewrite the multi animation examples under the struct or array?
Something like this:
```cpp
struct LottieExample
{
    enum Type : uint8_t { Slot = 0, Marker }
    unique_ptr<tvg::LottieAnimation> animation;
    Elm_Transit transition;
}

static std::vector<LottieExample> examples;
```

examples/LottieExtension.cpp
```diff
@@ -24,71 +24,105 @@
...
 
 void tvgUpdateCmds(tvg::Canvas* canvas, tvg::LottieAnimation* animation, float progress)
 {
-    if (updated || !canvas) return;
+    if (updated >= 2 || !canvas) return;
```
tinyjin, line 35: We may wanna avoid hard coded condition.
You can define constant value in top of file and reuse them.
```cpp
#define NUM_TOTAL_UPDATE 2
if (updated >= NUM_TOTAL_UPDATE || !canvas) return;
```

just FYI. we don't need to use `updated` condition, please find out `examples/Lottie.cpp`

examples/LottieExtension.cpp
```diff
@@ -24,71 +24,105 @@
...
-    picture->size(&w, &h);
-
-    if (w > h) {
-        scale = WIDTH / w;
-        shiftY = (HEIGHT - h * scale) * 0.5f;
-    } else {
-        scale = HEIGHT / h;
-        shiftX = (WIDTH - w * scale) * 0.5f;
+    //Slot
+    {
```
tinyjin, line 54: Pointless brackets. I think just comment is enough to divide code contexts.

tinyjin (changes requested):
Thank you @fall2019 :)
I've reviewed for updated codes.

fall2019:
Thanks. Resolved.

tinyjin (commented):
The API looks good, a few comments on example side.

tinyjin:
@hermet I guess the example part(LottieExtension) that I've just reviewed is not main part of this Marker feature. So let me know if I've overreviewed.

I can also improve LottieExtension part in this issue #2122

hermet:
@tinyjin, I think you can focus on the feature enhancement first. Actually, examples should be added as a separate commit. @fall2019, Basically, we prefer atomic patches, the best form of the PR is to separate patches for feature enhancement and examples because feature enhancement would not depend on examples. So if you separate them, that will be better for us.

fall2019:
Alright. Resolved.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -987,7 +1010,6 @@ void LottieParser::parseFonts()
     }
 }
 
-
```
hermet, line 990: X

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -736,7 +746,6 @@ struct LottieSlot
     bool overriden = false;
 };
 
-
```
hermet, line 739: X

hermet (commented):
Reviewed "Animation Segment(Marker) Support" commit. It looks almost there. Thanks,

hermet:
@fall2019 One more, please add yourself "https://github.com/thorvg/thorvg/blob/main/AUTHORS"  please.

fall2019:
Thanks. Resolved all comments.

tinyjin (commented):
Please check comments.

tinyjin:
@hermet 
I've tested `segment` API in various suites, seems nicely working. (some minor consideration -> commented)

Here, I wanna make sure two things.

**1. Unit test**
Should we include unit test for marker & segment in this patch? Otherwise, I can also add all cases I've checked into `testLottie.cpp` and `capiLottie.cpp` right after merging.

**2. Support marker comment in JSON objects.**
```json
"markers": [
    { "tm": 0, "cm": "{\\"name\\":\\"bird\\"}", "dr": 22 },
    { "tm": 22, "cm": "{\\"name\\":\\"explosion\\"}", "dr": 11 },
    { "tm": 33, "cm": "{\\"name\\":\\"feather\\"}", "dr": 77 }
]
```
This is also from RFC, the `markers[].cm` could be JSON string. Do you think it must be in this patch? I think we can gradually support this case in next patch.

hermet:
> **1. Unit test**
> Should we include unit test for marker & segment in this patch? Otherwise, I can also add all cases I've checked into `testLottie.cpp` and `capiLottie.cpp` right after merging.

@tinyjin Yes, It would be better to perform unit testing, but it's not mandatory in this PR. 

@fall2019 Since on this topic, Returning the list of marker names will be essential, so users can inquire about the available markers in the animation. Probably, like this:

```
uint32_t LottieAnimation::markerCnt();  //return the marker count
const char* LottieAnimation::markers(uint32_t idx); //return the marker name by a given index.
```
The items(unit test and apis) mentioned above are not the main focus at this point but will become essential eventually. Therefore, it's up to you. If you'd like to include them, please update the PR accordingly. Thanks.

</br>
</br>

> **2. Support marker comment in JSON objects.**
> 
> ```json
> "markers": [
>     { "tm": 0, "cm": "{\\"name\\":\\"bird\\"}", "dr": 22 },
>     { "tm": 22, "cm": "{\\"name\\":\\"explosion\\"}", "dr": 11 },
>     { "tm": 33, "cm": "{\\"name\\":\\"feather\\"}", "dr": 77 }
> ]
> ```
> 
> This is also from RFC, the `markers[].cm` could be JSON string. Do you think it must be in this patch? I think we can gradually support this case in next patch.

@tinyjin No, it shouldn't be.

examples/LottieExtension.cpp
```diff
@@ -72,23 +58,63 @@ void tvgDrawCmds(tvg::Canvas* canvas)
...
+/************************************************************************/
+void tvgDrawCmds(tvg::Canvas* canvas)
+{
+    //Background
+    auto shape = tvg::Shape::gen();
+    shape->appendRect(0, 0, WIDTH, HEIGHT);
+    shape->fill(50, 50, 50);
+    if (canvas->push(std::move(shape)) != tvg::Result::Success) return;
+    
+    for (auto& it : examples)
```
hermet, line 92: convention
```
for () {
   ...
}
```

examples/LottieExtension.cpp
```diff
@@ -189,19 +219,25 @@ int main(int argc, char **argv)
 
         elm_init(argc, argv);
 
-        transit = elm_transit_add();
```
hermet, line 192: Using multiple transitions here is inappropriate. Please refer to the examples/Lottie.cpp to see how it handles multiple animations.

hermet (changes requested):
Reviewed "Add example for Animation Segment(Marker)"

fall2019:
Add api
`uint32_t LottieAnimation::markerCnt();  //return the marker count
const char* LottieAnimation::markers(uint32_t idx); //return the marker name by a given index.`

Resolved all comments.

tinyjin:
@fall2019. So far pretty good.
It's around the corner. I'm going to soon wrap up this patch by my side.

Thanks.

fall2019:
Okay. On your side.

tinyjin:
#2164

PR #2082, opened 2024-03-21
wg_engine: stroke trim

SergeyLebedkin:
[issues 1479: strokes](#1479)

Introduced stroke trim setting

left: (0.0, 0.95)
middle: (0.15, 1.0)
right: (0.15, 0.95)

image

### DIFF
--- a/src/renderer/wg_engine/tvgWgGeometry.cpp
+++ b/src/renderer/wg_engine/tvgWgGeometry.cpp
@@ -204,6 +204,40 @@ void WgGeometryData::appendMesh(const RenderMesh* rmesh)
 };
 
 
+WgPoint WgGeometryData::interpolate(float t, uint32_t& index)
+{
+    assert(t >= 0.0f);
+    assert(t <= 1.0f);
+    // get total length
+    float totalLength = 0.0f;
+    for (uint32_t i = 0; i < positions.count - 1; i++)
+        totalLength += positions[i].dist(positions[i+1]);
+    float targetLength = totalLength * t;
+    // find current segment index and interpolation factor
+    float currentFactor = 0.0f;
+    float currentLength = 0.0f;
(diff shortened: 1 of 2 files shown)


src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -328,24 +354,30 @@ void WgGeometryDataGroup::stroke(const RenderShape& rshape)
...
     assert(rshape.stroke);
     if (rshape.stroke->dashPattern) {
-        // first step: decode path data
+        // decode path data
         WgGeometryDataGroup segments{};
         decodePath(rshape, &segments);
-        // second step: split path to segments using dash patterns
+        // trim segments
+        WgGeometryDataGroup trimmed{};
+        trimOutlines(&segments, &trimmed, rshape.stroke->trim.begin, rshape.stroke->trim.end);
```
hermet, line 362: I think you can skip the trimOutlines when the trim region is default (begin 0.0, end = 1.0) ?

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -328,24 +354,30 @@ void WgGeometryDataGroup::stroke(const RenderShape& rshape)
...
-        // first step: decode path data
-        WgGeometryDataGroup outlines{};
-        decodePath(rshape, &outlines);
-        // second step: create geometry for strokes
+        // decode path data
+        WgGeometryDataGroup segments{};
+        decodePath(rshape, &segments);
+        // trim segments
+        WgGeometryDataGroup trimmed{};
+        trimOutlines(&segments, &trimmed, rshape.stroke->trim.begin, rshape.stroke->trim.end);
```
hermet, line 377: here as well.

hermet (commented):
+one comment.

hermet (approved):

PR #2087, opened 2024-03-22
lottie/slot: Fixed potential overriding bug

tinyjin:
Slot overriding tries to set property via the `LottieObject`. It potentially causes unexpected changes.

Revised the `override` function. The purpose of this patch is to aim atomic property to be overriden.

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.h
+++ b/src/loaders/lottie/tvgLottieModel.h
@@ -125,7 +125,7 @@ struct LottieObject
         free(name);
     }
 
-    virtual void override(LottieObject* prop)
+    virtual void override(LottieProperty* prop)
     {
         TVGERR("LOTTIE", "Unsupported slot type");
     }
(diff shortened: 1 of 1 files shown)


src/loaders/lottie/tvgLottieModel.h
```diff
@@ -642,28 +642,36 @@ struct LottieSlot
         //apply slot object to all targets
         for (auto pair = pairs.begin(); pair < pairs.end(); ++pair) {
             //backup the original properties before overwriting
```
hermet, line 644: Please review this, probably we can make code simpler like:
```
for (auto pair = pairs.begin(); pair < pairs.end(); ++pair) {
   pair->prop = pair->obj->override(&static_cast<LottieGradient*>(target)->colorStops, overriden);
}
```

hermet (approved):
LGTM, please check a comment, probably you can make a better code.

PR #2090, opened 2024-03-25
loader/svg: Do recalc after gradient inheritance

JSUYA:
Inherit gradientUnits and recalculate each element
radial(cx, cy, r,...), linear(x1, y1, x2, y2).

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2669,7 +2669,7 @@ static void _inheritGradient(SvgLoaderData* loader, SvgStyleGradient* to, SvgSty
         if (to->transform) memcpy(to->transform, from->transform, sizeof(Matrix));
     }
 
-    if (to->type == SvgGradientType::Linear && from->type == SvgGradientType::Linear) {
+    if (to->type == SvgGradientType::Linear) {
         for (unsigned int i = 0; i < sizeof(linear_tags) / sizeof(linear_tags[0]); i++) {
             bool coordSet = to->flags & linear_tags[i].flag;
             if (!(to->flags & linear_tags[i].flag) && (from->flags & linear_tags[i].flag)) {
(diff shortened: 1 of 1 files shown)


JSUYA:
The SVG_FILE_130758.svg problem in the regression test looks to be an inherit problem of the focal feature.
@mgrudzinska  Can you check this?

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2669,11 +2669,13 @@ static void _inheritGradient(SvgLoaderData* loader, SvgStyleGradient* to, SvgSty
         if (to->transform) memcpy(to->transform, from->transform, sizeof(Matrix));
     }
 
-    if (to->type == SvgGradientType::Linear && from->type == SvgGradientType::Linear) {
```
hermet, line 2672: This condition looks like somewhat unnecessary changes. It doesn't need to loop for().

JSUYA: from and to can be different gradients(linear, radial).
Regardless of what the from is, recalc must be performed according to changes in common properties.(unit, transform, spread)

hermet: @JSUYA please see 2675 line.

JSUYA: ok I removed 2675line if condition

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2696,10 +2696,16 @@ static void _inheritGradient(SvgLoaderData* loader, SvgStyleGradient* to, SvgSty
             //GradUnits not set directly, coord set
             if (!gradUnitSet && coordSet) {
                 radialTags[i].tagRecalc(loader, to->radial, to->userSpace);
+                //If fx and fy are not set, set cx and cy.
+                if (!strcmp(radialTags[i].tag, "cx") && (to->flags & SvgGradientFlags::Cx) && !(to->flags & SvgGradientFlags::Fx)) to->radial->fx = to->radial->cx;
```
hermet, line 2700: the center position is mandatory in radial gradient. In the exceptional case, it would overwrite with the default value anyway. Thus this condition looks unnecessary. 
``` !strcmp(radialTags[i].tag, "cx") && (to->flags & SvgGradientFlags::Cx) ```

hermet (commented):
thanks. one more comment.

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: um.. depending on the inheritance status, there may be a case where cx or cy is missing, but it will probably be overwritten with 0.
I removed `(to->flags & SvgGradientFlags::Cx)` condition.
`strcmp()` should be necessary. If not, it is overwritten at every tag, so recalculated fx or ,fy may be overwritten in `fr`.

hermet (approved):

PR #2101, opened 2024-03-27
lottie: Fix crash from invalid masking method

tinyjin:
This change is better for stability. Returns `None` if the `mode` attribute cannot be parsed.

Screenshot 2024-03-27 at 6 49 00 PM

related issue: #2072
> Resolving crashes by segmentation fault for 3 files
- test9_IDX_135_RAND_12722319039069386985.json
- test9_IDX_207_RAND_6442145524277313594.json
- test9_IDX_743_RAND_6211786878291024908.json

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -42,7 +42,10 @@ static char* _int2str(int num)
 
 CompositeMethod LottieParser::getMaskMethod(bool inversed)
 {
-    switch (getString()[0]) {
+    auto mode = getString();
+    if (!mode) return CompositeMethod::None;
+
+    switch (mode[0]) {
         case 'a': {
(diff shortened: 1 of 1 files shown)


src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -42,7 +42,10 @@ static char* _int2str(int num)
 
 CompositeMethod LottieParser::getMaskMethod(bool inversed)
 {
-    switch (getString()[0]) {
+    auto mode = getString();
+    if (!mode || strlen(mode) < 1) return CompositeMethod::None;
```
hermet, line 46: please remove the following condition if it doesn't helpful for the fix. that looks unnecessary.
```
 strlen(mode) < 1
```

hermet (approved):

Issue #2096, opened 2024-03-26
svg: support for set declaration of style name


JSUYA:
```html
<?xml version="1.0" encoding="utf-8"?>
<svg version="1.1" id="Content" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 35 35" xml:space="preserve">
    <defs>
        <style>.cls-2{fill:none;}
               .cls-3{fill:#1a73e8;}
               .cls-4,
               .cls-8{fill:#f00;}
               .cls-4{opacity:0.2;isolation:isolate;}
               .cls-5{fill:#1a237e;fill-opacity:0.2;}.cls-6{fill:#185abc;}.cls-7{fill:#f1f3f4;}.cls-9{fill:#bdc1c6;}</style>
    </defs>
	<path class="cls-8" d="M21.28,10.94H12.17a2.73,2.73,0,0,0-2.73,2.73V20.6a2.72,2.72,0,0,0,2.73,2.73h9.11A2.73,2.73,0,0,0,24,20.6V13.67A2.72,2.72,0,0,0,21.28,10.94Z" />
</svg>
```

In this case both .cls-4 and .cls-8 should be declared with `fill:#f00;`. Both cls-4 and cls-8 can be used as classes.

```
<style>
.a,.b,.c{fill:none;stroke:#000;}
.a,.b{stroke-miterlimit:10;}
.b,.c{stroke-linecap:round;}
.c{stroke-linejoin:round;}
</style>
```
In this case, each class integrates each attribute.
.a -> `fill:none;stroke:#000;stroke-miterlimit:10;`
.b -> `fill:none;stroke:#000;stroke-miterlimit:10;stroke-linecap:round;`
.c -> `stroke-linecap:round;stroke-linecap:round;`

PR #2106, opened 2024-03-28
loader/svg: Fix finding css class name

JSUYA:
Css class name starts with '.'.
Regardless of the previous parsing result,
if name contains '.', it is set to name from here.

related issue: https://github.com/thorvg/thorvg/issues/2096


hermet (commented):
what about line 557?

JSUYA:
> what about line 557?

Oh my approach is wrong. thank you for check.

qarmin:
Regression tool found some differences in files that previously failed to properly convert - https://github.com/thorvg/thorvg/actions/runs/8463217912 

Strange that comment about them was not added to comment section

JSUYA:
The way style_name are parsed needs to be improved.

PR #2111, opened 2024-03-29
lottie: optimized data copying in slot overriding.

hermet:
Generally, the data is already in runtime memory,
ThorVG doesn't need to copy it again.


hermet:
Can't allow this... if the input data is re-used.

PR #2113, opened 2024-03-29
Finding also crashes in regression finder

qarmin:
- previously svg2png crashes, timeouts and leaks have been ignored, but with new step now are now being caught - this depends partially on #2072 
- ~using github actions to create github comments instead curl(which did not work, I do not know why) - thanks to this, a comment will be created for each PR after the first passing of the CI, indicating whether or not changes in test suite have been detected in the PR, and on subsequent passes the same comment will be updated with newer information~
- re-enabling build with address/undefined sanitizer - looks that all strange freeze have been fixed

Not sure why, but comments not working(maybe will work after merging?) due error
```
 /home/runner/work/_actions/peter-evans/create-or-update-comment/v4/dist/index.js:4695
      const error = new requestError.RequestError(toErrorMessage(data), status, {
                    ^

RequestError [HttpError]: Resource not accessible by integration
    at /home/runner/work/_actions/peter-evans/create-or-update-comment/v4/dist/index.js:4695:21
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  status: 403,
```
https://github.com/peter-evans/create-or-update-comment/issues/76

this worked fine on different project - https://github.com/qarmin/SVG-regression-finder/actions/runs/8472470087/job/23214648273
and added/updated comment: https://github.com/qarmin/SVG-regression-finder/pull/5#issuecomment-2025893685

Also ruff uses similar workflow - https://github.com/astral-sh/ruff/blob/main/.github/workflows/pr-comment.yaml

https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/resource-not-accessible

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -14,12 +14,12 @@ jobs:
       - name: Install things
         run: |
           sudo apt-get update
-          sudo apt-get install -y wget2 librsvg2-dev librsvg2-bin ninja-build git gcc-multilib g++-multilib
+          sudo apt-get install -y wget2 librsvg2-dev librsvg2-bin ninja-build git gcc-multilib g++-multilib earlyoom
           sudo pip3 install meson
 
       - name: Compile Thorvg PR
         run: |
-          meson . build -Dtools=svg2png
+          meson . build -Dtools=svg2png,lottie2gif -Dsavers=gif -Db_sanitize=address,undefined
           sudo ninja -C build install
 
       - name: Compile Thorvg Develop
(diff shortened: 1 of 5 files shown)


hermet:
@qarmin Thanks for your contribution. We might need to clarify the regression failure here.

qarmin:
I have no idea how to make comments automatically add to PR, so I just commented them out so the CI is now green and anyone can check what's not working as I have no idea how to fix it(looks like a permissions problem)

Adding comments to a thread is useful when files that didn't convert properly before are changing. In that case, apart from the information in the CI logs, there is no information that the PR has changed anything. In the case where a regression is found, CI fails and this is easily visible.

tinyjin:
Hello, @qarmin.
Would this link help for the commenting problem?
https://github.com/peter-evans/create-or-update-comment/issues/188

```
In public repositories this action does not work in pull_request workflows when triggered by forks.
```

I think it ocurrs because this patch from forked repo. If so, we can just merge and see it works.

tinyjin:
Gotcha, I was misunderstanding. I thought the workflow in PR doesn't work only when Github Action(.yml) changed.
But I see what you're saying is right.

Anyway, this problem only happens in PR which from forked repo due to the token issue.
I've just found this as well, it might solve them.

> Alternatively, use the [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) event to comment on pull requests.

The event `pull_request_target` may unlimit few things on PR instead of `pull_request`.
I am not sure it would simply work by replacing `pull_request` to `pull_request_target`, thinking worth to try.

I see some people do same thing in forked PR like:
```yml
name: Thank contributors

on:
  pull_request_target:
    types:
      - closed

jobs:
  thank:
    if: github.event.pull_request.merged == true
    runs-on: ubuntu-latest
    steps:
      - name: Comment on the pull request
        uses: peter-evans/create-or-update-comment@v3
        with:
          issue-number: ${{ github.event.pull_request.number }}
          body: |
            Thank you for contributing! 🎉
```

Sharing related article :)
https://dev.to/kalkwst/workflow-triggering-events-and-event-actions-5cec#understanding-event-activity-types

hermet:
> But as far as I can see, no one is doing that, and additionally it would not change anything for outsiders

@qarmin You got a write access. please do it if it's necessary, it's recommended for thorvg maintenance. Thanks.

hermet (approved):

PR #2125, opened 2024-04-01
loader/svg: Support hsl color format

JSUYA:
Support parsing of hsl(hue, saturation, brightness) color type and conversion to rgb color.

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -103,15 +103,19 @@ static const char* _skipComma(const char* content)
 }
 
 
-static bool _parseNumber(const char** content, float* number)
+static bool _parseNumber(const char** content, const char** end, float* number)
 {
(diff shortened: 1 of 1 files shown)


JSUYA:
related file https://github.com/thorvg/thorvg/issues/1255#issuecomment-2021859300
SVG_FILE_141599.svg

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -117,6 +117,19 @@ static bool _parseNumber(const char** content, float* number)
 }
 
 
+static bool _parseNumber(const char* content, char** end, float* number)
```
hermet, line 120: Hmm... this looks can be merged with the above... Actually it should be.

JSUYA: I updated it

hermet (approved):

PR #2130, opened 2024-04-02
loader/svg: Check invalid color

hermet:
Checks whether the string that can be specified in Color is valid.

example)
```
     style="stroke:asdasd"
```

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -667,7 +667,7 @@ static bool _hslToRgb(float hue, float satuation, float brightness, uint8_t* red
 }
 
 
-static void _toColor(const char* str, uint8_t* r, uint8_t* g, uint8_t* b, char** ref)
+static bool _toColor(const char* str, uint8_t* r, uint8_t* g, uint8_t* b, char** ref)
 {
     unsigned int len = strlen(str);
     char *red, *green, *blue;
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -735,22 +738,24 @@ static void _toColor(const char* str, uint8_t* r, uint8_t* g, uint8_t* b, char**
                     tb /= 100.0f;
                     brightness = _skipSpace(brightness + 1, nullptr);
                     if (brightness && brightness[0] == ')' && brightness[1] == '\0') {
-                       if (_hslToRgb(th, ts, tb, r, g, b)) return;
+                       return _hslToRgb(th, ts, tb, r, g, b);
```
hermet, line 741: @JSUYA please verify this

hermet:
image

@JSUYA This svg has been broken by this patch. I think you should take care of this...

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -713,6 +715,7 @@ static void _toColor(const char* str, uint8_t* r, uint8_t* g, uint8_t* b, char**
                 }
             }
         }
+        return true;
     } else if (ref && len >= 3 && !strncmp(str, "url", 3)) {
         if (*ref) free(*ref);
         *ref = _idFromUrl((const char*)(str + 3));
```
JSUYA, line 721: missing `return true;`

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -735,22 +738,24 @@ static void _toColor(const char* str, uint8_t* r, uint8_t* g, uint8_t* b, char**
                     tb /= 100.0f;
                     brightness = _skipSpace(brightness + 1, nullptr);
                     if (brightness && brightness[0] == ')' && brightness[1] == '\0') {
-                       if (_hslToRgb(th, ts, tb, r, g, b)) return;
+                       return _hslToRgb(th, ts, tb, r, g, b);
                     }
                 }
             }
         }
+        return true;
```
JSUYA, line 746: please remove this.
If hsl color parsing is successful, it should return at line 741. If line 741 is not touched, _toColor() should be fail.

PR #2134, opened 2024-04-03
gl_engine: hotfix for the main surface drawing issue.

hermet:
GL might need to generate a default target FBO
when the given target ID indicates the main surface.

However, users may want to draw visuals directly onto the main surface.

This policy must be reviewed thoroughly.

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -56,7 +56,7 @@ bool GlRenderer::clear()
 
 bool GlRenderer::target(int32_t id, uint32_t w, uint32_t h)
 {
-    assert(w > 0 && h > 0);
+    if (id == GL_INVALID_VALUE || w == 0 || h == 0) return false;
 
     surface.stride = w;
     surface.w = w;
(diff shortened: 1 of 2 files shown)


hermet:
@RuiwenTang @SergeyLebedkin This PR should be applied in v0.13 release. It would be appreciated if you could review at your earliest convenience.

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -69,6 +69,11 @@ bool GlRenderer::target(int32_t id, uint32_t w, uint32_t h)
 
     mTargetFboId = static_cast<GLint>(id);
 
+    //TODO: It's not allow to draw onto the main surface. Need to confirm the policy.
+    if (mTargetFboId == 0) {
+        GL_CHECK(glGetIntegerv(GL_FRAMEBUFFER_BINDING, &mTargetFboId));
+    }
```
RuiwenTang, line 75: If with EFL, the `mTargetFboId` is not zero, does it mean that at https://github.com/thorvg/thorvg/blob/main/examples/Path.cpp#L113 this point in time, the framebuffer corresponding to GLView has been initialized?

hermet: @RuiwenTang Yes. Actually glview surface fbo in EFL has been initialized at the point, However the gl binding api does not return it. (that's the problem)

RuiwenTang: But the comment says that drawing to the main surface (fbo 0) is not allowed. I think this description is not accurate.

hermet: Because currently it's aborted, not allowed to draw on the main surface. It's a todo fix. please fix it. :-)

PR #2136, opened 2024-04-03
Change C-style code to C++ code style

rinechran:
Change C-style code to C++ code style

### DIFF
--- a/src/common/tvgInlist.h
+++ b/src/common/tvgInlist.h
@@ -27,8 +27,8 @@ namespace tvg {
 
 //NOTE: declare this in your list item
 #define INLIST_ITEM(T) \
-    T* prev; \
-    T* next
+    T* prev = nullptr; \
+    T* next = nullptr
 
 template<typename T>
 struct Inlist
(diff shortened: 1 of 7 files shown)


hermet:
Unless any practical benefits, I think there is no point to change the coding style.

probably, except this:
image

rinechran:
yes The code style remains the same, but the reason for changing a few cases came out as an error in the mscv static integration stone, so I changed it.

hermet:
> yes The code style remains the same, but the reason for changing a few cases came out as an error in the mscv static integration stone, so I changed it.

Could you please share the error message?

rinechran:
Here's a log message.

image
image

ref : https://learn.microsoft.com/en-us/cpp/code-quality/c26440?view=msvc-170&f1url=%3FappId%3DDev17IDEF1%26l%3DEN-US%26k%3Dk(C26440)%26rd%3Dtrue
ref : https://msdn.microsoft.com/query/dev17.query?appId=Dev17IDEF1&l=EN-US&k=k(C26493)&rd=true

rinechran:
@hermet  I fully empathize with your opinion and believe it's appropriate to close the issue at hand. Thank you as always for sharing your thoughts

Issue #2153, opened 2024-04-05
Lottie animation not rendering - potential lack of layer effects support


theashraf:
Hey team

We've encountered an issue with the dotLottie players where a specific animation is not rendering. The main issue is documented here: https://github.com/LottieFiles/dotlottie-web/issues/164.

I ran a Lottie features report using [relottie](https://codepen.io/lottiefiles/pen/eYXyypR) on the mentioned animation and compiled a list of features it utilizes. It appears that ThorVG does not yet support the "layer effect" feature, which is used by this animation. Any suggestions on how to proceed?

Here is the list of features identified:

```text
name
threedimensional
height
width
metadata
metadata-children
generator
composition
composition-children
layer-precomposition
layer-type
time-stretch
start-time
out-point
in-point
hidden
layer-threedimensional
blend-mode-normal
has-mask
auto-orient
layer-transform
layer-transform-children
anchor-point
animated-position
animated
position-keyframe-list
position-keyframe-list-children
position-keyframe
position-keyframe-out-tangent
keyframe-bezier-handle
keyframe-bezier-handle-x-axis
keyframe-bezier-handle-y-axis
position-keyframe-in-tangent
keyframe-value
keyframe-value-children
position-keyframe-time
layer-transform-scale
animated-multidimensional
keyframe-list
keyframe-list-children
keyframe
keyframe-out-tangent
keyframe-in-tangent
keyframe-time
layer-transform-skew
animated-value-static
static-value
translation
rotation-clockwise
animated-value
layer-transform-skew-axis
transform-opacity
reference-id
composition-index
layer-shape
effect-list
effect-list-children
effect-drop-shadow
effect-type
enabled
effect-param-list
effect-param-list-children
effect-param-color
effect-value-color
animated-color
effect-param-slider
effect-value-slider
effect-param-angle
effect-value-angle
shape-list
shape-list-children
shape-path
shape-type
shape-direction-clockwise
animated-shape-prop
animated-shape
shape-keyframe-list
shape-keyframe-list-children
shape-keyframe
shape-keyframe-out-tangent
shape-keyframe-in-tangent
shape-keyframe-start
shape-keyframe-start-children
bezier
bezier-closed
bezier-in-tangents
bezier-in-tangents-children
bezier-out-tangents
bezier-out-tangents-children
bezier-vertices
bezier-vertices-children
shape-keyframe-time
shape-fill
shape-fill-color
fill-rule-non-zero
shape-fill-opacity
version
framerate
[...]

PR #2160, opened 2024-04-09
lottie: corrected the comp/layer size data type.

hermet:
Actually, the these size is expected to be a floating-point value.
This correction addresses a parsing error.

issue: https://github.com/thorvg/thorvg/issues/2153

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -1089,14 +1089,14 @@ void LottieParser::parseText(Array<LottieObject*>& parent)
 }
 
 
-void LottieParser::getLayerSize(uint32_t& val)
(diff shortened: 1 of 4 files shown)


src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1089,14 +1089,14 @@ void LottieParser::parseText(Array<LottieObject*>& parent)
 }
 
 
-void LottieParser::getLayerSize(uint32_t& val)
+void LottieParser::getLayerSize(float& val)
 {
     if (val == 0) {
```
tinyjin, line 1094: `if (val == 0.0f)` might be better? since val becomes `float`!

tinyjin (approved):

Issue #1244, opened 2022-08-20
Replace examples ui toolkits


hermet:
Currently thorvg examples use EFL lib to run a window & animation loop. But EFL is difficult to use in windows platform. Probably we can replace it with an other toolkit which is more portable and maintainable among OSs (i.e. glfw?) Any suggestions?

vtorri:
It is not difficult to use on Windows ! :-D If you need help, I can provide it. You know I'm the Windows maintainer of the EFL. Please tell me what the problems are and what you need

vtorri:
@hermet i have written [ewpi](https://github.com/vtorri/ewpi) for the dependencies of the EFL on Windows. I also have written an NSIS installer for the  dependencies and the EFL. It's on my local computer at home (i'm on holydays). If you want I'll release them.

Also, I have enabled all the options for the EFL. If you want a minimal set of options, please tell me  what you need

hermet:
@vtorri I know. :) I raised this issue because @projectitis looks struggle with EFL on windows. Probably you can help him.

vtorri:
@projectitis if you use IRC, you can join #e on libera.chat.

projectitis:
Sorry, it's bedtime here :) 

I have GLFW3 working with ThorVG that we could provide as an option, but I would gladly accept advice on how to get EFL working. Then I will update the readme with steps for windows users and we won't have to update the examples 😁

It might be good to have a GLFW implementation as an option as it's a reasonably popular framework.

vtorri:
@projectitis the simplest way for you is to use msys2. You install with pacman the dependencies, and you can compile the EFL with meson.

or you wait 1 or 2 weeks for me to create the needed installers.

vpxyz:
Another option is SDL2. It is simple to use and widely used.

RuiwenTang:
Do we still consider another option? 
I am trying to rewrite the OpenGL backend for thorvg. And the biggest problem is the OpenGL context creation when using EFL.

My usual working environment is macbook, but EFL can not create the OpenGL context as I mentioned in #1531 

In Ubuntu, EFL creates Open GLES contexts up to version 3.1. Even if the  GPU driver provides OpenGL 4.6 capability.

BTW. If we plan to choose WebGPU in the future #1479 , I can not find the API to get the raw window handler which is needed to create a WebGPU surface

hermet:
[...]

PR #2162, opened 2024-04-10
gl_engine: using normal texture rendering in final color blit

RuiwenTang:
This PR replace the blit logical with normal texture rendering in the final color blit pass.

Since blit msaa framebuffer to another msaa framebuffer may generate GLError in some platforms. Use normal texture rendering to blit the final color buffer onto target framebuffer.

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderTask.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderTask.cpp
@@ -202,25 +202,26 @@ void GlComposeTask::onResolve() {
 }
 
 GlBlitTask::GlBlitTask(GlProgram* program, GLuint target, GlRenderTarget* fbo, Array<GlRenderTask*>&& tasks)
- : GlComposeTask(program, target, fbo, std::move(tasks))
+ : GlComposeTask(program, target, fbo, std::move(tasks)), mColorTex(fbo->getColorTexture())
 {
-}
-
-void GlBlitTask::setSize(uint32_t width, uint32_t height)
-{
-    mWidth = width;
-    mHeight = height;
 }
 
 void GlBlitTask::run()
 {
     GlComposeTask::run();
 
-    GL_CHECK(glScissor(0, 0, mWidth, mHeight));
     GL_CHECK(glBindFramebuffer(GL_FRAMEBUFFER, getTargetFbo()));
-    GL_CHECK(glBindFramebuffer(GL_READ_FRAMEBUFFER, getSelfFbo()));
 
-    GL_CHECK(glBlitFramebuffer(0, 0, mWidth, mHeight, 0, 0, mWidth, mHeight, GL_COLOR_BUFFER_BIT, GL_NEAREST));
+    if (mClearBuffer) {
+        GL_CHECK(glClearColor(0, 0, 0, 0));
+        GL_CHECK(glClear(GL_COLOR_BUFFER_BIT));
+    }
+
+    // make sure the blending is correct
+    GL_CHECK(glEnable(GL_BLEND));
+    GL_CHECK(glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_ALPHA));
+
+    GlRenderTask::run();
 }
 
 GlDrawBlitTask::GlDrawBlitTask(GlProgram* program, GLuint target, GlRenderTarget* fbo, Array<GlRenderTask*>&& tasks)
--- a/src/renderer/gl_engine/tvgGlRenderTask.h
+++ b/src/renderer/gl_engine/tvgGlRenderTask.h
@@ -130,7 +130,7 @@ protected:
 
     GLuint getResolveFboId();
 
-    virtual void onResolve();
+    void onResolve();
 private:
     GLuint mTargetFbo;
     GlRenderTarget* mFbo;
@@ -143,16 +143,11 @@ public:
     GlBlitTask(GlProgram*, GLuint target, GlRenderTarget* fbo, Array<GlRenderTask*>&& tasks);
     ~GlBlitTask() override = default;
 
-    void setSize(uint32_t width, uint32_t height);
-
     void run() override;
 
-protected:
-    void onResolve() override {}
-
+    GLuint getColorTextore() const { return mColorTex; }
 private:
-    uint32_t mWidth = 0;
-    uint32_t mHeight = 0;
+    GLuint mColorTex = 0;
 };
 
 class GlDrawBlitTask : public GlComposeTask
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -86,8 +86,6 @@ bool GlRenderer::sync()
     //nothing to be done.
     if (mRenderPassStack.size() == 0) return true;
 
-    mGpuBuffer->flushToGPU();
-
     // Blend function for straight alpha
     GL_CHECK(glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_ALPHA));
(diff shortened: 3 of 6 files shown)


RuiwenTang:
@hermet Can you help verify if this PR is worked? I don't have EFL development environment.

hermet:
@RuiwenTang nothing visible after applying this patch.

RuiwenTang:
> @RuiwenTang nothing visible after applying this patch.

I think, this is because, the target framebuffer ID is wrong. As a said, https://github.com/thorvg/thorvg/pull/2134#discussion_r1554848495

hermet:
> > @RuiwenTang nothing visible after applying this patch.
> 
> I think, this is because, the target framebuffer ID is wrong. As I said before [#2134 (comment)](https://github.com/thorvg/thorvg/pull/2134#discussion_r1554848495)

Hmm.. if the framebuffer id is wrong, how it works now then? (fbo is not zero but 1)

RuiwenTang:
> > > @RuiwenTang nothing visible after applying this patch.
> > 
> > 
> > I think, this is because, the target framebuffer ID is wrong. As I said before [#2134 (comment)](https://github.com/thorvg/thorvg/pull/2134#discussion_r1554848495)
> 
> Hmm.. if the framebuffer id is wrong, how it works now then? fbo is not zero but 1 by gl renderer:
> 
> ```
> glGetIntegerv(GL_FRAMEBUFFER_BINDING, &mTargetFboId));
> ```

Can you add break point at https://github.com/thorvg/thorvg/blob/main/examples/Path.cpp#L113 and https://github.com/thorvg/thorvg/blob/main/src/renderer/gl_engine/tvgGlRenderer.cpp#L74 . Then check the framebuffer IDs obtained the secode time are equal ?

hermet:
> > > > @RuiwenTang nothing visible after applying this patch.
> > > 
> > > 
> > > I think, this is because, the target framebuffer ID is wrong. As I said before [#2134 (comment)](https://github.com/thorvg/thorvg/pull/2134#discussion_r1554848495)
> > 
> > 
> > Hmm.. if the framebuffer id is wrong, how it works now then? fbo is not zero but 1 by gl renderer:
> > ```
> > glGetIntegerv(GL_FRAMEBUFFER_BINDING, &mTargetFboId));
> > ```
> 
> Can you add break point at https://github.com/thorvg/thorvg/blob/main/examples/Path.cpp#L113 and https://github.com/thorvg/thorvg/blob/main/src/renderer/gl_engine/tvgGlRenderer.cpp#L74 . Then check the framebuffer IDs obtained the secode time are equal ?

@RuiwenTang That efl api (https://github.com/thorvg/thorvg/blob/main/examples/Path.cpp#L113) doesn't work as it's supposed to do. It returns 0 always. So we put the workaround to acquire the actual framebuffer in gl renderer here: https://github.com/thorvg/thorvg/blob/main/src/renderer/gl_engine/tvgGlRenderer.cpp#L74

RuiwenTang:
> > > > > @RuiwenTang nothing visible after applying this patch.
> > > > 
> > > > 
> > > > I think, this is because, the target framebuffer ID is wrong. As I said before [#2134 (comment)](https://github.com/thorvg/thorvg/pull/2134#discussion_r1554848495)
> > > 
> > > 
> > > Hmm.. if the framebuffer id is wrong, how it works now then? fbo is not zero but 1 by gl renderer:
> > > ```
> > > glGetIntegerv(GL_FRAMEBUFFER_BINDING, &mTargetFboId));
> > > ```
> > 
> > 
> > Can you add break point at https://github.com/thorvg/thorvg/blob/main/examples/Path.cpp#L113 and https://github.com/thorvg/thorvg/blob/main/src/renderer/gl_engine/tvgGlRenderer.cpp#L74 . Then check the framebuffer IDs obtained the secode time are equal ?
> 
> @RuiwenTang That efl api (https://github.com/thorvg/thorvg/blob/main/examples/Path.cpp#L113) doesn't work as it's supposed to do now. It returns 0 always. It's an efl interface issue we can ignore it. Instead we put the workaround to acquire the actual framebuffer(1) in gl renderer here: https://github.com/thorvg/thorvg/blob/main/src/renderer/gl_engine/tvgGlRenderer.cpp#L74 though it's also a workaround of the issue that the drawing to the main surface(0) occurs the abort.

Can we change to use another window toolkit as mensioned in https://github.com/thorvg/thorvg/issues/1244 which can running on MacOS ? I really don't know what happens under the GLContext created by EFL.

hermet:
> > > > > > @RuiwenTang nothing visible after applying this patch.
> > > > > 
> > > > > 
> > > > > I think, this is because, the target framebuffer ID is wrong. As I said before [#2134 (comment)](https://github.com/thorvg/thorvg/pull/2134#discussion_r1554848495)
> > > > 
> > > > 
> > > > Hmm.. if the framebuffer id is wrong, how it works now then? fbo is not zero but 1 by gl renderer:
> > > > ```
> > > > glGetIntegerv(GL_FRAMEBUFFER_BINDING, &mTargetFboId));
> > > > ```
> > > 
> > > 
> > > Can you add break point at https://github.com/thorvg/thorvg/blob/main/examples/Path.cpp#L113 and https://github.com/thorvg/thorvg/blob/main/src/renderer/gl_engine/tvgGlRenderer.cpp#L74 . Then check the framebuffer IDs obtained the secode time are equal ?
> > 
> > 
> > @RuiwenTang That efl api (https://github.com/thorvg/thorvg/blob/main/examples/Path.cpp#L113) doesn't work as it's supposed to do now. It returns 0 always. It's an efl interface issue we can ignore it. Instead we put the workaround to acquire the actual framebuffer(1) in gl renderer here: https://github.com/thorvg/thorvg/blob/main/src/renderer/gl_engine/tvgGlRenderer.cpp#L74 though it's also a workaround of the issue that the drawing to the main surface(0) occurs the abort.
> 
> @hermet Can we change to use another window toolkit as mensioned in #1244 which can running on MacOS ? I really don't know what happens under the GLContext created by EFL.

I wish to. just lack of time for it. :-)

hermet:
@RuiwenTang See TVG Window port in the roadmap: https://github.com/thorvg/thorvg/wiki/24'-Development-Roadmap

RuiwenTang:
@hermet Or can we add a set of OpenGL/GLES examples using GLFW or SDL so we can unify the OpenGL runtime environment and let me continue development ?

SergeyLebedkin:
> @hermet Or can we add a set of OpenGL/GLES examples using GLFW or SDL so we can unify the OpenGL runtime environment and let me continue development ?

Please, do not forget about webgpu (maybe other APIs). We need to get handles to native display/window from different platforms for APIs initializations

RuiwenTang:
> > @hermet Or can we add a set of OpenGL/GLES examples using GLFW or SDL so we can unify the OpenGL runtime environment and let me continue development ?
> 
> Please, do not forget about webgpu (maybe other APIs). We need to get handles to native display/window from different platforms for APIs initializations

I think it's not complicated using GLFW https://github.com/eliemichel/glfw3webgpu

hermet:
> @hermet Or can we add a set of OpenGL/GLES examples using GLFW or SDL so we can unify the OpenGL runtime environment and let me continue development ? 
@RuiwenTang You mean add one sample that composed with glfw in examples?

RuiwenTang:
@hermet Can you try this patch again, I found the problem, thanks

hermet:
@RuiwenTang Cool, it works. thanks.

Please double-check, as I observed three broken examples with the latest update (regardless of this change).
https://github.com/thorvg/thorvg/issues/1687#issuecomment-2054101770

hermet (approved):

PR #2171, opened 2024-04-12
wg_engine: optimize with the internal triangulation algorithm

SergeyLebedkin:
[issues 1479: lottie](#1479)

Optimizations:
    - "on-the-fly" computing polyline segments lengths, total length and bboxes
    - optimaze stroking algorithm by prevent vector normalizations
    - tabulated sin and cos functions
    - reducing system and graphics memory realocations
    - prealocate triangle fans index buffers (less memory writes)
    - using render meshes heaps to reduce webgpu instances re-creations
    - using single instance of pilylines for standard operations such as trim and split
    - "on-the-fly" strokes generations with dash pattern
    - "on-the-fly" mesh objects generation in a process of path decoding
    - merge strokes into single mesh by each shape


hermet:
@SergeyLebedkin Thanks, if possible, it might be better to separate the patches corresponding to the each enhancements. This could make the review process more effective and allow for incremental updates.

PR #2175, opened 2024-04-13
sw_engine : addition of aarch64 neon

rinechran:
Even when I enable the vector option, host_machine.cpu_family() is set to aarch64, so NEON doesn't get activated


SergeyLebedkin (approved):

hermet (changes requested):
please make it sure about 'adv_neon'.

hermet:
Please check this: https://github.com/thorvg/thorvg/pull/2176

PR #2176, opened 2024-04-14
infra: revise the vector build script

hermet:
Note that this change renames the Meson option from 'vector' to 'simd',
separates the NEON types into 'neon-arm' and 'neon-aarch',
and designates the 'mfpu' option solely for 'neon-arm'.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -85,13 +85,16 @@ endif
 #Vectorization
 simd_type = 'none'
 
-if get_option('vector') == true
-  if host_machine.cpu_family() == 'x86' or host_machine.cpu_family() == 'x86_64'
+if get_option('simd') == true
+  if host_machine.cpu_family().startswith('x86')
     config_h.set10('THORVG_AVX_VECTOR_SUPPORT', true)
     simd_type = 'avx'
-  elif host_machine.cpu_family() == 'arm'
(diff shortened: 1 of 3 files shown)


rinechran:
https://github.com/thorvg/thorvg/blob/b4a0933a10f36ed2bb8004e5d925df8ed50a3ae3/src/meson.build#L30

When compiling on aarch64, the error ['-mfpu=neon'] occurs.

reference : https://stackoverflow.com/questions/29851128/gcc-arm64-aarch64-unrecognized-command-line-option-mfpu-neon

hermet:
> https://github.com/thorvg/thorvg/blob/b4a0933a10f36ed2bb8004e5d925df8ed50a3ae3/src/meson.build#L30
> 
> When compiling on aarch64, the error ['-mfpu=neon'] occurs. so I created a separate type called adv_neon to resolve the issue
> 
> reference : https://stackoverflow.com/questions/29851128/gcc-arm64-aarch64-unrecognized-command-line-option-mfpu-neon

Got it.  I think "adv_neon" is ambiguous for users. neon is just neon. there is no need to separate it.

rinechran:
I also think that there is no need to separate them

Issue #2163, opened 2024-04-11
Godot: Heap buffer overflow when loading some SVG image 


hermet:
Need to verify this is a thorvg issue or not.
https://github.com/godotengine/godot/issues/90371

PR #2179, opened 2024-04-15
loader/svg: Add null to the end of data

JSUYA:
Because memcpy() is not guaranteed to copy null at the end of the data array,
it increase the size by 1 and add null.
This prevents invalid access of string functions in parser.

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -3813,10 +3813,11 @@ bool SvgLoader::open(const char* data, uint32_t size, TVG_UNUSED const string& r
     clear();
 
     if (copy) {
-        content = (char*)malloc(size);
+        content = (char*)malloc(size + 1);
         if (!content) return false;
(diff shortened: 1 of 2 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3813,9 +3813,10 @@ bool SvgLoader::open(const char* data, uint32_t size, TVG_UNUSED const string& r
     clear();
 
     if (copy) {
-        content = (char*)malloc(size);
+        content = (char*)malloc(size + 1);
         if (!content) return false;
         memcpy((char*)content, data, size);
+        content[size] = nullptr;
```
hermet, line 3819: maybe ```'\0'``` ?

JSUYA: thanks i updated it

JSUYA:
https://github.com/thorvg/thorvg/issues/2163

hermet (approved):

mgrudzinska (approved):

PR #2180, opened 2024-04-15
common: code refactoring

hermet:
introduced mathDeg2Rad() and mathRad2Deg() for a common implementation.

### DIFF
--- a/src/common/tvgMath.h
+++ b/src/common/tvgMath.h
@@ -44,6 +44,18 @@ bool mathIdentity(const Matrix* m);
 void mathMultiply(Point* pt, const Matrix* transform);
 
 
+static inline float mathDeg2Rad(float degree)
+{
+     return degree * (MATH_PI / 180.0f);
+}
+
+
+static inline float mathRad2Deg(float radian)
+{
+    return radian * (180.0f / MATH_PI);
+}
+
+
 static inline bool mathZero(float a)
 {
     return (fabsf(a) < FLT_EPSILON) ? true : false;
--- a/src/common/tvgLines.cpp
+++ b/src/common/tvgLines.cpp
@@ -238,7 +238,7 @@ float bezAngleAt(const Bezier& bz, float t)
     pt.x *= 3;
     pt.y *= 3;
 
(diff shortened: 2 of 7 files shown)


src/common/tvgMath.h
```diff
@@ -44,6 +44,18 @@ bool mathIdentity(const Matrix* m);
 void mathMultiply(Point* pt, const Matrix* transform);
 
 
+static inline float mathDeg2Rad(float degree)
+{
+     return degree * MATH_PI / 180.0f;
```
JSUYA, line 49: how about re-define `MATH_PI / 180.0f` ?
```
#define MATH_PI3 0.01745329238474369049f
```

src/common/tvgMath.h
```diff
@@ -44,6 +44,18 @@ bool mathIdentity(const Matrix* m);
...
+}
+
+
+static inline float mathRad2Deg(float radian)
+{
+    return radian * 180.0f / MATH_PI;
```
JSUYA, line 55: how about re-define `180.0f / MATH_PI` ?
```
#define MATH_PI4  57.2957763671875f
```

src/common/tvgMath.h
hermet: I don't see any beneficial points. Please think about these points:
1. better performance?
2. less & neat code?
3. enhance the readibility?

JSUYA: 1. A division operation is performed every time. The amount of computation can be reduced.
2.3.4. You have already defined `MATH_PI *0.5f` as `MATH_PI2`. If you use an appropriate name, I think it is better.

hermet: - Compiler will do pre-computation. No benefits practically.
- MATH_PI2 has followed the M_PI2, as in the math definition. This prevents the repetitive code from the usage.
- Current logic is already commonized and clear to read. Please suggest me the proper name if you'd like to argue this. 
- In the end, those definition introduction will just increase the LOC.

SergeyLebedkin (approved):

tinyjin (approved):

Issue #2183, opened 2024-04-15
Lottie/Marker: Runtime crash when parsing error


tinyjin:
Marker APIs have potential crash, which use following LottieLoader functions:
- `uint32_t LottieLoader::markersCnt()`
- `const char* LottieLoader::markers(uint32_t index)`
- `bool LottieLoader::segment(const char* marker, float& begin, float& end)`

[marker_crash.json](https://github.com/thorvg/thorvg/files/14976694/marker_crash.json)

<img width="1287" alt="Screenshot 2024-04-15 at 6 14 16 PM" src="https://github.com/thorvg/thorvg/assets/11167117/b134ed7d-5360-4003-8062-7b5afd53bf35">

So basically, after parsing the Lottie with invalid marker object. It's going to be runtime crash when those Marker APIs called.

The parser isn't affected by this error, but recently introduced APIs seem not covered.

ThorVG must not make crash even if it has parsing error.

### Valid marker object
```json
{
    "tm": 33,
    "cm": "sectionC",
    "dr": 30
}
```

### Invalid marker object (-> crash)
```json
{
    "tm": 33,
    "cm": "sectionC",
    "dr": "error because it's mismatched type"
}
```

PR #2184, opened 2024-04-15
lottie/loader: Fix crash when parsing error

tinyjin:
In any cases from parsing error, Marker APIs come runtime crash.

Added checking nullptr condition to avoid them.

issue: #2183


hermet (commented):
It might need to return the fail at LottieAnimation: 76, 84

image

hermet:
> @hermet You mean we need to revise the api to be this?
> 
> ```c++
> Result LottieAnimation::markersCnt(uint32_t& cnt) noexcept { }
> 
> Result LottieAnimation::marker(uint32_t idx, const char** name) noexcept  { }
> ```
Never mind, your fix looks correct. However, I think this issue isn't limited to just the marker. Please double-check the other loader methods and fix them as necessary.
ex)
image

hermet (approved):

PR #2187, opened 2024-04-15
lottie: ++stability

hermet:
ensure that the loading has been completed before overriding.

### DIFF
--- a/src/loaders/lottie/tvgLottieLoader.cpp
+++ b/src/loaders/lottie/tvgLottieLoader.cpp
@@ -262,7 +262,8 @@ bool LottieLoader::read()
 
 Paint* LottieLoader::paint()
 {
-    this->done();
+    done();
+
     if (!comp) return nullptr;
     comp->initiated = true;
     return comp->root->scene;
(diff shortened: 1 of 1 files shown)


tinyjin (approved):

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -271,6 +272,8 @@ Paint* LottieLoader::paint()
 
 bool LottieLoader::override(const char* slot)
 {
+    if (!comp) done();
+
     if (!comp || comp->slots.count == 0) return false;
```
mgrudzinska, line 277: is  ```!comp``` still needed in this if statement?

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -262,7 +262,8 @@ bool LottieLoader::read()
 
 Paint* LottieLoader::paint()
 {
-    this->done();
+    done();
```
mgrudzinska, line 265: this can be changed in many other places in this file, but out of the scope of this pr

mgrudzinska (approved):

Issue #2156, opened 2024-04-08
sw_engine : Often the test fails.


rinechran:
Occasionally failds while running a test on x86_64

below are the build options and test error logs

```
meson setup build -Dlog=true -Dexamples=true  -Dsavers="all" -Dtests=true
➜  a git:(master) ✗ meson setup build -Dlog=true -Dexamples=true  -Dsavers="all" -Dtests=true
The Meson build system
Version: 0.61.2
Source dir: /mnt/ssd/project/a
Build dir: /mnt/ssd/project/a/build
Build type: native build
Project name: thorvg
Project version: 0.13.99
C++ compiler for the host machine: c++ (gcc 11.4.0 "c++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0")
C++ linker for the host machine: c++ ld.bfd 2.38
Host machine cpu family: x86_64
Host machine cpu: x86_64
Configuring config.h using configuration
Library pthread found: YES
Found pkg-config: /usr/bin/pkg-config (0.29.2)
Run-time dependency elementary found: YES 1.26.2
Message:

Summary:
    ThorVG version:          0.13.99
    Build Type:              debugoptimized
    Prefix:                  /usr/local
    Multi-Tasking:           true
    SIMD Instruction:        none
    Raster Engine (SW):      true
    Raster Engine (GL_BETA): false
    Raster Engine (WG_BETA): false
    Loader (TVG):            true
    Loader (SVG):            true
    Loader (TTF):            false
    Loader (LOTTIE):         true
    Loader (PNG):            false
    Loader (JPG):            false
    Loader (WEBP):           false
    Saver (TVG):             true
    Saver (GIF):             true
    Binding (CAPI):          false
    Binding (WASM_BETA):     false
    Log Message:             true
    Tests:                   true
    Examples:                true
    Tool (Svg2Tvg):          false
    Tool (Svg2Png):          false
    Tool (Lottie2Gif):       false


Build targets in project: 56

thorvg 0.13.99

  User defined options
    examples: true
    log     : true
    savers  : all
    tests   : true

Found ninja-1.10.1 at /usr/bin/ninja

```

```

➜  a git:(master) ✗ ./build/test/tvgUnitTests
[L] RENDERER (../src/renderer/tvgLoader.cpp 169): PNG format is not supported
[L] RENDERER (../src/renderer/tvgLoader.cpp 169): JPG format is not supported
[...]

PR #2189, opened 2024-04-16
loader/svg: Return actual decoded size

JSUYA:
related issue: https://github.com/thorvg/thorvg/issues/2156

### DIFF
--- a/src/loaders/svg/tvgSvgUtil.cpp
+++ b/src/loaders/svg/tvgSvgUtil.cpp
@@ -47,7 +47,6 @@ size_t svgUtilURLDecode(const char *src, char** dst)
     if (length == 0) return 0;
 
     char* decoded = (char*)malloc(sizeof(char) * length + 1);
-    decoded[length] = '\0';
 
     char a, b;
     int idx =0;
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgUtil.cpp
```diff
@@ -46,7 +46,7 @@ size_t svgUtilURLDecode(const char *src, char** dst)
     auto length = strlen(src);
     if (length == 0) return 0;
 
-    char* decoded = (char*)malloc(sizeof(char) * length + 1);
+    char* decoded = (char*)calloc(length + 1, sizeof(char));
     decoded[length] = '\0';
```
hermet, line 50: It could count the actual length of the decoded string (=> idx?). 
For example, this approach might be more efficient than unnecessarily clearing memory.
```
 decoded[idx] = '\0';
```

JSUYA: you are right. I hadn't thought about return size. I updated PR .

hermet (approved):

PR #2192, opened 2024-04-16
wg_engine: optimize with the internal triangulation algorithm and mesh rendering

SergeyLebedkin:
Optimizations:
- "on-the-fly" computing polyline segments lengths, total length and bboxes
- optimaze stroking algorithm by prevent vector normalizations
- tabulated sin and cos functions
- reducing system and graphics memory realocations
- prealocate triangle fans index buffers (less memory writes)
- using render meshes heaps to reduce webgpu instances re-creations
- using single instance of pilylines for standard operations such as trim and split
- "on-the-fly" strokes generations with dash pattern
- "on-the-fly" mesh objects generation in a process of path decoding
- merge strokes into single mesh by each shape

### DIFF
--- a/src/renderer/wg_engine/tvgWgGeometry.cpp
+++ b/src/renderer/wg_engine/tvgWgGeometry.cpp
@@ -23,6 +23,159 @@
 #include "tvgWgGeometry.h"
 
 //***********************************************************************
+// WgPolyline
+//***********************************************************************
+
+WgMath* WgGeometryData::gMath = nullptr;
+
+void WgMath::initialize()
+{
+    if (initialized) return;
+    initialized = true;
+    const uint32_t nPoints = 360;
+    sinus.reserve(360);
+    cosin.reserve(360);
+    for (uint32_t i = 0; i < nPoints; i++) {
+        float angle = i * (2 * M_PI) / nPoints;
+        sinus.push(sin(angle));
+        cosin.push(cos(angle));
+    }
+};
+
+
+void WgMath::release() {
+    sinus.clear();
+    cosin.clear();
+};
+
+//***********************************************************************
+// WgPolyline
+//***********************************************************************
+
+WgPolyline::WgPolyline()
+{
+    pts.reserve(1024);
+    dist.reserve(1014);
+}
+
+
+void WgPolyline::appendPoint(WgPoint pt)
+{
+    if (pts.count > 0) {
+        float distance = pts.last().dist(pt);
+        if (distance > 0) {
+            // update min and max indexes
+            iminx = pts[iminx].x >= pt.x ? pts.count : iminx;
+            imaxx = pts[imaxx].x <= pt.x ? pts.count : imaxx;
+            iminy = pts[iminy].y >= pt.y ? pts.count : iminy;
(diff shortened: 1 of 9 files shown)


src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -23,126 +23,233 @@
...
-    if (positions.count <= 2) return;
-    indexes.reserve((positions.count - 2) * 3);
-    for (size_t i = 0; i < positions.count - 2; i++) {
-        indexes.push(0);
-        indexes.push(i + 1);
-        indexes.push(i + 2);
+    if (initialized) return;
+    initialized = true;
+    const uint32_t nPoints = 360;
+    sinus.reserve(360);
```
mgrudzinska, line 36: nPoints

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -23,126 +23,233 @@
...
+};
+
+//***********************************************************************
+// WgPolyline
+//***********************************************************************
+
+WgPolyline::WgPolyline()
+{
+    pts.reserve(1024);
+    dist.reserve(1014);
```
mgrudzinska, line 58: any reason why 1014 not 1024?

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -23,126 +23,233 @@
...
+    pmax.x = pts[imaxx].x;
+    pmax.y = pts[imaxy].y;
+}
+
+//***********************************************************************
+// WgGeometryData
+//***********************************************************************
+
+WgGeometryData::WgGeometryData() {
+    positions.pts.reserve(10240);
```
mgrudzinska, line 183: 1024? since this number appears in few places I'd use constexpr

mgrudzinska (commented):
in some places there are magic numbers. would be easier to debug in the future, if we keep them in some constexpr ?
8? 3? 2? 
(sure not all of them :) )
also M_PI -> MATH_PI ?

SergeyLebedkin:
> in some places there are magic numbers. would be easier to debug in the future, if we keep them in some constexpr ? 8? 3? 2? (sure not all of them :) ) also M_PI -> MATH_PI ?

M_PI -> MATH_PI I believe that we should use our own PI constant, you are right

hermet:
@SergeyLebedkin Hello, could you please improve your commit messages? Even it currently does not follow our commit format.

src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -49,34 +49,62 @@ class WgPoint
...
+    bool initialized{};
+
+    void initialize();
+    void release();
+};
+
+
+struct WgPolyline
+{
+    // polyline points
```
hermet, line 75: Suggestions:
1. These comments in the WgPolyline structure merely duplicate the variable names and seem unnecessary. Comments should have any hidden purpose or intention more than it.
2. Please append comments on one line if necessary.

hermet (approved):

PR #2193, opened 2024-04-16
loader/svg: Add check to the limits of result of StrToFloat

JSUYA:
The string passed to the svg parser is not guaranteed to always be a valid string.
Certain strings may have numbers that cannot be converted.
Therefore, make sure to check whether the converted value is in the appropriate range before returning.

related issue: https://github.com/thorvg/thorvg/issues/2078#issuecomment-2037495121

### DIFF
--- a/src/common/tvgStr.cpp
+++ b/src/common/tvgStr.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "config.h"
+#include <cmath>
 #include <cstring>
 #include <memory.h>
 #include "tvgMath.h"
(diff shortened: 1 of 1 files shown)


src/common/tvgStr.cpp
```diff
@@ -197,6 +197,8 @@ float strToFloat(const char *nPtr, char **endPtr)
 
 success:
     if (endPtr) *endPtr = (char *)(a);
+    if (val < FLT_EPSILON || val > FLT_MAX) return 0.0f;
```
mgrudzinska, line 200: ```val < FLT_EPSILON``` is a problem for us?
also ```val > FLT_MAX``` won't catch ```NaN```, right? 
so maybe ```std::isfinite``` ?

JSUYA: Thanks! I updated pr

hermet (approved):

PR #2194, opened 2024-04-16
lottie: code clean up

hermet:


tinyjin (approved):
LGTM!

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -356,8 +356,7 @@ float LottieLoader::duration()
     if (!comp) done();
     if (!comp) return 0.0f;
 
-    auto frameNo = frameCnt * (segmentEnd - segmentBegin);
-    return frameNo / comp->frameRate;
+    return frameCnt * (segmentEnd - segmentBegin) / comp->frameRate;
```
mgrudzinska, line 359: I'd put all into one line, but up to you
```return !comp ? 0.0f : frameCnt * (segmentEnd - segmentBegin) / comp->frameRate;```

also in l.371 I'd write:
```return !comp ? 0 :  comp->markers.count;```

hermet: cleaning code is good but not mandatory. 
This kind of requirement would require endless code changes.
Let's review the diff only. Thanks.

mgrudzinska (approved):

PR #2199, opened 2024-04-18
draft: experimental code formatting

RuiwenTang:
As mentioned in #1519
We need an automated code formatting tool and consistent code style
This PR shows how to use clang-format to define a coding style and use `.git-blame-ignore-revs` to ignore the format commit from normal gitblame.
github can apply this ignore list automatically. For local repository, need run the following command:
```
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

image

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,3 @@
+# Defines the Chromium style for automatic reformatting.
+# http://clang.llvm.org/docs/ClangFormatStyleOptions.html
+BasedOnStyle: Google
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,11 @@
+# This file contains a list of Git commit hashes that should be hidden from the
+# regular Git history. Typically, this includes commits involving mass auto-formatting
+# or other normalizations. Commit hashes *must* use the full 40-character notation.
+# To apply the ignore list in your local Git client, you must run:
+#
+#   git config blame.ignoreRevsFile .git-blame-ignore-revs
+#
+# This file is automatically used by GitHub.com's blame view.
+
+# previous path which format thorvg.h
+22fc2d521b3d5fc4af53f1db2b2903852f17dfb0
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -2,63 +2,62 @@
 #define _THORVG_H_
 
 #include <functional>
+#include <list>
 #include <memory>
 #include <string>
-#include <list>
 
 #ifdef TVG_API
-    #undef TVG_API
+#undef TVG_API
 #endif
 
 #ifndef TVG_STATIC
-    #ifdef _WIN32
-        #if TVG_BUILD
-            #define TVG_API __declspec(dllexport)
-        #else
-            #define TVG_API __declspec(dllimport)
-        #endif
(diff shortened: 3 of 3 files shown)


RuiwenTang:
@hermet Do you think this method is acceptable?

hermet:
@hermet Firstly, I'm welcome for this. Here are two major points we need to make a consensus:
- When to apply? This change will bring huge code modifications that may be difficult for maintainers of other systems to integrate with the newer versions of ThorVG. That's why I've considered the change point is at 1.0 release.
- What coding convention standard do we prefer? There are a bunch of different coding standard by the well-known organizations/companies and ThorVG has no connection to Google.

hermet:
> @hermet Do you think this method is acceptable?

@RuiwenTang Firstly, I'm welcome for this. Here are two major points we need to make a consensus:
- When to apply? This change will bring huge code modifications that may be difficult for maintainers of other systems to integrate with the newer versions of ThorVG. That's why I've considered the change point is at 1.0 release.
- What coding convention standard do we prefer? There are a bunch of different coding standard by the well-known organizations/companies and ThorVG has no connection to Google.

RuiwenTang:
> > @hermet Do you think this method is acceptable?
> 
> @RuiwenTang Firstly, I'm welcome to this. Here are two major points we need to make a consensus:
> 
> * When to apply? This change will bring huge code modifications that may be difficult for maintainers of other systems to integrate with the newer versions of ThorVG. That's why I've considered the change point is at 1.0 release.
> * What coding convention standard do we prefer? There are a bunch of different coding standard by the well-known organizations/companies and ThorVG has no connection to Google so far.

I think it all depends on you.
As the number of code and contributors increases, global formatting becomes increasingly expensive. Also, I think the current coding style might cause confusion when others are first contributing to ThorVG

Details about coding style are also up to you, clang-format in this PR is for demonstration only.


> BTW. Every time I develop ThorVG I have to disable the editor's automatic code formatting behavior :(

hermet:
> > > @hermet Do you think this method is acceptable?
> > 
> > 
> > @RuiwenTang Firstly, I'm welcome to this. Here are two major points we need to make a consensus:
> > 
> > * When to apply? This change will bring huge code modifications that may be difficult for maintainers of other systems to integrate with the newer versions of ThorVG. That's why I've considered the change point is at 1.0 release.
> > * What coding convention standard do we prefer? There are a bunch of different coding standard by the well-known organizations/companies and ThorVG has no connection to Google so far.
> 
> I think it all depends on you. As the number of code and contributors increases, global formatting becomes increasingly expensive. Also, I think the current coding style might cause confusion when others are first contributing to ThorVG
> 
> We can discuss the details of coding style together, clang-format in this PR is for demonstration only.
> 
> > BTW. Every time I develop ThorVG I have to disable the editor's automatic code formatting behavior :(

Firstly, most contributors would have the same problem when they contribute to a project which is not familiar with them. They make same mistakes. This is not a thorvg only issue. Rather than it, I would like to say, coding conventions for developers are a minor issue, but integration issues with ThorVG will matter more. Thus I bet this on 1.0 release. Otherwise, Integrating it with large systems like Godot, LVGL, and Tizen could have a big trouble. Maybe we can consider an auto-formatting script that aligns with the current ThorVG style. People could simply write code first and run the script. It's easy, and everyone will be happy.

Issue #2013, opened 2024-02-27
Memory problems that occur when the wasm class is destroyed


MrZillaGold:
When the emscripten class over wasm is no longer needed, you need to call the `delete` method as recommended in the documentation to free up occupied memory and prevent memory leaks.

https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html#memory-management

Memory errors often occur when calling the `delete` method, here are a few error stacks:

```
RuntimeError: memory access out of bounds
    at thorvg-wasm.wasm.tvg::Scene::Impl::clear(bool) (wasm://wasm/thorvg-wasm.wasm-013439fa:wasm-function[619]:0x199d5)
    at thorvg-wasm.wasm.tvg::Scene::clear(bool) (wasm://wasm/thorvg-wasm.wasm-013439fa:wasm-function[618]:0x1992e)
    at thorvg-wasm.wasm.LottieBuilder::update(LottieComposition*, float) (wasm://wasm/thorvg-wasm.wasm-013439fa:wasm-function[1785]:0x65dfb)
    at thorvg-wasm.wasm.LottieLoader::run(unsigned int) (wasm://wasm/thorvg-wasm.wasm-013439fa:wasm-function[1933]:0x6ec0f)
    at thorvg-wasm.wasm.non-virtual thunk to LottieLoader::run(unsigned int) (wasm://wasm/thorvg-wasm.wasm-013439fa:wasm-function[1952]:0x6f56e)
    at thorvg-wasm.wasm.tvg::TaskSchedulerImpl::request(tvg::Task*) (wasm://wasm/thorvg-wasm.wasm-013439fa:wasm-function[728]:0x1d348)
    at thorvg-wasm.wasm.tvg::TaskScheduler::request(tvg::Task*) (wasm://wasm/thorvg-wasm.wasm-013439fa:wasm-function[727]:0x1d2e4)
    at thorvg-wasm.wasm.LottieLoader::frame(float) (wasm://wasm/thorvg-wasm.wasm-013439fa:wasm-function[1970]:0x704f6)
    at thorvg-wasm.wasm.tvg::Animation::frame(float) (wasm://wasm/thorvg-wasm.wasm-013439fa:wasm-function[135]:0x8d50)
    at thorvg-wasm.wasm.TvgLottieAnimation::frame(float) (wasm://wasm/thorvg-wasm.wasm-013439fa:wasm-function[3197]:0xcce0e)
```

```
[...]

Issue #1858, opened 2023-12-22
web/lottie-player: please check this codepen example I created:


hermet:
please check this codepen example I created: https://codepen.io/aidosmf/pen/PoLYzVX

if I would like to change styles programmatically, there is no way to access it from the instance. The public methods are not workings as well

_Originally posted by @Aidosmf in https://github.com/thorvg/thorvg/pull/1817#discussion_r1434875039_

tinyjin:
Thank you! @Aidosmf 
Let me take a look deeper into this example.

Since it's standard web component, I guess the creation should be like this
```ts
// wrong
import { LottiePlayer } from "https://esm.sh/@thorvg/lottie-player";
...
const player = await createPlayer()

// correct
import 'https://unpkg.com/@thorvg/lottie-player@latest/dist/lottie-player.js';
...
const player = document.createElement('lottie-player');
```

unsure yet, I'll prepare valid example for js-creation.

It's really good for improving test suite by the way = )

PR #2201, opened 2024-04-18
web: stability++

tinyjin:
Screenshot 2024-04-18 at 11 41 21 PM

`delete()` function by emscripten would guarantee that the WASM module is cleaned up from the memory when component is destroyed.

see also: https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html#memory-management

This patch includes some refactorings on web component and example to support calling this function.

### DIFF
--- a/web/example/index.html
+++ b/web/example/index.html
@@ -9,13 +9,14 @@
   <body>
     <div>
       <div>
+        <button onclick="playAnimation()">Play</button>
         <button onclick="pauseAnimation()">Pause</button>
         <button onclick="stopAnimation()">Stop</button>
+        <button onclick="createAnimation()">Create</button>
         <button onclick="destroyAnimation()">Destroy</button>
-        <button onclick="playAnimation()">Play</button>
-        <button onclick="save2gif()">save2gif</button>
-        <button onclick="save2tvg()">save2tvg</button>
-        <button onclick="save2png()">save2png</button>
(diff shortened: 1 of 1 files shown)


tinyjin:
This patch indicates official way to render in code-base creation. (find the `createAnimation()` in `web/example/index.html`)
Will close the #1858.

### js-creation
```js
import 'https://unpkg.com/@thorvg/lottie-player@latest/dist/lottie-player.js';

const animation = document.createElement('lottie-player');

animation.autoPlay = true;
animation.loop = true;
animation.src = "https://lottie.host/6d7dd6e2-ab92-4e98-826a-2f8430768886/NGnHQ6brWA.json";
animation.style.width = "500px";
animation.style.height = "500px";

appendChild(animation);
```

tinyjin:
The missing of `delete()` has been found in #2013.

web/src/lottie-player.ts
```diff
@@ -279,29 +277,21 @@ export class LottiePlayer extends LitElement {
...
+    if (!this._timer) {
+      //NOTE: ThorVG Module has loaded, but called this function again
       return;
     }
 
     clearInterval(this._timer);
     this._timer = undefined;
 
-    this._TVG = _tvg;
+    this._TVG = new _module.TvgLottieAnimation();;
```
hermet, line 294: ```;;```

PR #2205, opened 2024-04-19
infra++

hermet:


src/tools/meson.build
```diff
@@ -2,14 +2,14 @@ if (lib_type == 'static')
     compiler_flags += ['-DTVG_STATIC']
 endif
 
-if all_tools or get_option('tools').contains('svg2png') == true
+if svg2png == true
```
mgrudzinska, line 5: ```if svg2png``` is enough

mgrudzinska (approved):
I'd remove ```== true``` from the src/tools/meson.build but otherwise ok

Issue #2206, opened 2024-04-19
svg: a regression bug


hermet:
This breaks the rendering: https://github.com/thorvg/thorvg/pull/2148

```
commit 938946d3ba514f9950ca6e2da4411cb5d79e17a3 (HEAD)
Author: JunsuChoi <jsuya.choi@samsung.com>
Date:   Fri Apr 5 13:42:02 2024 +0900

    loader/svg: Skip to invalid polygon

    If a Polygon's points array is odd, it is not a valid shape.
```

![image](https://github.com/thorvg/thorvg/assets/3711518/ea811a40-853c-4e0a-94e5-81fdbd65c718)

JSUYA:
um... In my environment, svg works fine.
https://github.com/thorvg/thorvg/pull/2148 patch is no longer drawn if polygon type elements's points array do not match in pairs. Therefore, it has nothing to do with color change.
And `gallardo.svg`, which is displayed incorrectly, does not use the polygon element.
`152932619-bd3d6921-72df-4f09-856b-f9743ae32a14.svg` displayed fine using polygon elements in my environment.

It's strange that `1630446379Bahamas-patriotic-flag-symbol.svg` is reversed.

hermet:
@JSUYA Please try it with `-Dlog=true`

JSUYA:
oh! ok i found it. i will make pr soon. thanks

mgrudzinska:
https://github.com/thorvg/thorvg/pull/2212

PR #2207, opened 2024-04-19
loader/svg: Revert part of 75e587a

JSUYA:
This patch reverts part of 75e587a for fix regression bug.
Even if unsupported elements are encountered, further parsing continues.

related issue: https://github.com/thorvg/thorvg/issues/2206

### DIFF
--- a/src/loaders/svg/tvgXmlParser.cpp
+++ b/src/loaders/svg/tvgXmlParser.cpp
@@ -358,7 +358,6 @@ bool simpleXmlParseAttributes(const char* buf, unsigned bufLength, simpleXMLAttr
         if (!func((void*)data, tmpBuf, tval)) {
             if (!_isIgnoreUnsupportedLogAttributes(tmpBuf, tval)) {
                 TVGLOG("SVG", "Unsupported attributes used [Elements type: %s][Id : %s][Attribute: %s][Value: %s]", simpleXmlNodeTypeToString(((SvgLoaderData*)data)->svgParse->node->type), ((SvgLoaderData*)data)->svgParse->node->id ? ((SvgLoaderData*)data)->svgParse->node->id : "NO_ID", tmpBuf, tval ? tval : "NONE");
-                goto error;
             }
         }
     }


hermet (approved):

hermet:
You might wanna write a better commit msg title:
```
//clear to understand the purpose of this commit.
loader/svg: Fix a regression bug. 
loader/svg: Fix a part of patch due to a regression bug
```

mgrudzinska (changes requested):
I think that the whole https://github.com/thorvg/thorvg/commit/75e587a9a9b8becbedc6f1ccacd88da95be1ccb4 should be reverted.

1. from the svg std:
[If an odd number of coordinates is provided, then the element is in error, with the same user agent behavior as occurs with an incorrectly specified path element.](https://www.w3.org/TR/SVG2/shapes.html#PolygonElement)
the path error handling is here: https://www.w3.org/TR/SVG2/paths.html#PathDataErrorHandling

In summary: for path data error handling, data should be read up until the first error occurs, and only the data up to that point should be rendered. 

As I understand, the same principle should apply to the points of a polygon. If there is an error in the data or an odd number of points is provided, then only the data loaded up to that point should be rendered.
@JSUYA do you agree or am I wrong?

2. ```if (!func(buf, bufLength, _attrParsePolygonNode, loader)) return nullptr;```
if we want to add this check, we should handle not only polygons, but all the node types we create. otherwise, the code will not be consistent

@JSUYA please check https://github.com/thorvg/thorvg/pull/2212 as the solution I'd apply

JSUYA:
> I think that the whole [75e587a](https://github.com/thorvg/thorvg/commit/75e587a9a9b8becbedc6f1ccacd88da95be1ccb4) should be reverted.
> 
> 1. from the svg std:
>    [If an odd number of coordinates is provided, then the element is in error, with the same user agent behavior as occurs with an incorrectly specified path element.](https://www.w3.org/TR/SVG2/shapes.html#PolygonElement)
>    the path error handling is here: https://www.w3.org/TR/SVG2/paths.html#PathDataErrorHandling
> 
> In summary: for path data error handling, data should be read up until the first error occurs, and only the data up to that point should be rendered.
> 
> As I understand, the same principle should apply to the points of a polygon. If there is an error in the data or an odd number of points is provided, then only the data loaded up to that point should be rendered. @JSUYA do you agree or am I wrong?
> 

Thank you for your detail information. you are right. I reviewed #2212. 

> 2. `if (!func(buf, bufLength, _attrParsePolygonNode, loader)) return nullptr;`
>    if we want to add this check, we should handle not only polygons, but all the node types we create. otherwise, the code will not be consistent

I think, if there is an object that should not be drawn when parsing is not processed properly, this syntax is resonable.
So far, we have only worked on valid images, so we created nodes regardless of the result of `func()`, but this may be needed more as we need to support it more widely through recent broken file testing. 
However, since it is currently difficult to check the parsing results of all specs, I think we can add more of these parts.
Of course, since this does not apply in this case, it is correct to revert my commit. :)
Thanks!

PR #2209, opened 2024-04-20
lottie: improve markers spec coverage.

hermet:
Allow the both usages:

1. "cm": "sectionB"
2. "cm": { "name": "sectionB", "description": "test decsription" }


src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -955,10 +955,19 @@ LottieMarker* LottieParser::parseMarker()
     auto marker = new LottieMarker;
     
     while (auto key = nextObjectKey()) {
-        if (KEY_AS("cm")) marker->name = getStringCopy();
-        else if (KEY_AS("tm")) marker->time = getFloat();
-        else if (KEY_AS("dr")) marker->duration = getFloat();
-        else skip(key);
+        if (KEY_AS("cm")) {
+            if (peekType() == kObjectType) {
```
tinyjin, line 959: If I remember correctly, the "cm" attribute is written in JSON string not object. We should parse them for string type.

tinyjin:
```js
"markers": [
    { "tm": 0, "cm": "{\\"name\\":\\"bird\\"}", "dr": 22 },
    { "tm": 22, "cm": "{\\"name\\":\\"explosion\\"}", "dr": 11 },
    { "tm": 33, "cm": "{\\"name\\":\\"feather\\"}", "dr": 77 }
]
```

tinyjin (commented):
Please check comment :)

src/loaders/lottie/tvgLottieParser.cpp
hermet: Hmm... It's an object. Any side issues?

tinyjin: I mean, since the "cm" is string type even if it's actually JSON format, we may need to parse stringified JSON to object then parse again in object type.

Screenshot 2024-04-22 at 10 19 50 AM

I tried an example with JSON "cm". It fails.

Screenshot 2024-04-22 at 10 22 18 AM
[marker_sample.json](https://github.com/thorvg/thorvg/files/15056043/marker_sample.json)

hermet: @tinyjin Sorry, I don't see any issue here. It handles the CM name as a string. Could you please suggest the correct logic you think? I also can't see any errors with your sample file.

tinyjin: OK, thanks for check. I will soon test it again.

hermet:
Let's revisit this with a practical usage.

PR #2212, opened 2024-04-21
svg_loader: correct polygon's points loading 

mgrudzinska:
[Revert "loader/svg: Skip to invalid polygon"](https://github.com/thorvg/thorvg/commit/c89dd9e2366a5f363507583131a2c042d209de55) 

This reverts commit https://github.com/thorvg/thorvg/commit/75e587a9a9b8becbedc6f1ccacd88da95be1ccb4.
If incorrect data for the points in a polygon is provided,
the element should still be rendered, but only up to the point
where the error occurs—using an even number of the points that
have been successfully loaded.

[svg_loader: correct polygon's points loading](https://github.com/thorvg/thorvg/commit/54f85d9a04ee1397d8b40b8ad63f76a4a62e200c) 

Only an even number of correctly read points defining
a polygon should be loaded and passed on for rendering.
Any remaining points should be ignored.

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1724,8 +1724,11 @@ static SvgNode* _createEllipseNode(SvgLoaderData* loader, SvgNode* parent, const
 
 static bool _attrParsePolygonPoints(const char* str, SvgPolygonNode* polygon)
 {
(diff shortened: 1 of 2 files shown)


JSUYA (approved):
LGTM

hermet (approved):

Issue #2210, opened 2024-04-21
[SVG] Nested elements rendering when they shouldn't


MewPurPur:
`<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><rect x="2" y="2" width="6" height="4" fill="red"><circle cx="10" cy="10" r="3"/></rect></svg>`

Every renderer I found ignores the circle inside the rect tag, but not ThorVG, which renders both shapes.

I didn't find this behavior stated explicitly anywhere in the SVG specification. This seems like it might be it, though: https://www.w3.org/TR/SVG2/struct.html#InterfaceSVGSVGElement

> To find the non-container graphics elements within a given element _element_, the following steps are run:

> Let result be an initially empty list.
> If element is an ‘[svg](https://www.w3.org/TR/SVG2/struct.html#SVGElement)’ or ‘[g](https://www.w3.org/TR/SVG2/struct.html#GElement)’ element, then for each child element _child_ of _element_, in document order:
> - Let descendants be the result of [finding the non-container graphics elements](https://www.w3.org/TR/SVG2/struct.html#TermFindNonContainerGraphicsElements) within child.
> - Append to result all the elements of descendants.
> 
> Otherwise, if element is a [graphics element](https://www.w3.org/TR/SVG2/struct.html#TermGraphicsElement) then append element to result.
> Return result.

But while at it, I found other related ones that we're doing wrong: https://www.w3.org/TR/SVG2/struct.html#UnknownElement

Here's an example SVG from the specification, trimmed down to only consider the elements we currently support:

```
<svg viewBox="0 0 700 200" xmlns="http://www.w3.org/2000/svg">
  <mysteryElement>
    <path d="M0 0h600v200h-600z" fill="#bdb76b"/>
  </mysteryElement>
  <hideElement xmlns="http://www.example.com/invisibleML">
    <path d="M0 0h300v100h-300z" fill="#f5f5f5"/>
  </hideElement>
  <linearGradient>
    <gradientExtension>
      <path d="M300 0h100v200h-100z" fill="#f0f"/>
    </gradientExtension>
  </linearGradient>
</svg>
```

The rectangle inside the mysteryElement renders, which is correct.

The rectangle inside the hideElement shouldn't render because namespace was overridden with a non-SVG one, but it does render.

The rectangle inside gradientExtension shouldn't render because linearGradient isn't rendered, but it does. It actually does even if we put it directly under linearGradient.

[...]

PR #2213, opened 2024-04-22
loader/svg: Check current graphics node that not closed

JSUYA:
If any of the graphics nodes are not closed,
the graphics nodes between them are ignored.

related issue: https://github.com/thorvg/thorvg/issues/2210

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -3198,6 +3198,14 @@ static void _svgLoaderParserXmlClose(SvgLoaderData* loader, const char* content)
         }
     }
 
+    for (unsigned int i = 0; i < sizeof(graphicsTags) / sizeof(graphicsTags[0]); i++) {
+        if (!strncmp(content, graphicsTags[i].tag, graphicsTags[i].sz - 1)) {
+            loader->currentGraphicsNode = nullptr;
+            loader->stack.pop();
+            break;
+        }
+    }
+
     loader->level--;
 }
 
@@ -3264,6 +3272,11 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content, 
         if (loader->stack.count > 0) parent = loader->stack.last();
         else parent = loader->doc;
         node = method(loader, parent, attrs, attrsLength, simpleXmlParseAttributes);
+        if (node && !empty) {
+            auto defs = _createDefsNode(loader, nullptr, nullptr, 0, nullptr);
+            loader->stack.push(defs);
+            loader->currentGraphicsNode = node;
+        }
     } else if ((gradientMethod = _findGradientFactory(tagName))) {
         SvgStyleGradient* gradient;
         gradient = gradientMethod(loader, attrs, attrsLength);
--- a/src/loaders/svg/tvgSvgLoaderCommon.h
+++ b/src/loaders/svg/tvgSvgLoaderCommon.h
@@ -560,6 +560,7 @@ struct SvgLoaderData
     int level = 0;
     bool result = false;
     bool style = false;
+    SvgNode* currentGraphicsNode = nullptr;
 };
 
 struct Box


mgrudzinska (changes requested):
please check the Svg example. I'm getting pretty strange results. 

but also it's not so simple as 'the shapes can not be nested' - they can not, but the nested shape can be reused in the 'use' tag. check this:
```
<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
  <rect x="2" y="2" width="6" height="4" fill="red">
    <circle id="circ" cx="10" cy="10" r="3"/>
  </rect>
  <use href="#circ" transform="translate(0,4)"/>
</svg>
```

JSUYA:
> please check the Svg example. I'm getting pretty strange results.
> 
> but also it's not so simple as 'the shapes can not be nested' - they can not, but the nested shape can be reused in the 'use' tag. check this:
> 
> ```
> <svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
>   <rect x="2" y="2" width="6" height="4" fill="red">
>     <circle id="circ" cx="10" cy="10" r="3"/>
>   </rect>
>   <use href="#circ" transform="translate(0,4)"/>
> </svg>
> ```

oh thank you for check.
um.. I think handling nested nodes with `display = false` will solve the problem.

mgrudzinska:
I thought rather about using ```level``` or use some separate ```SvgNode*``` to store such elements. but maybe ```display``` will be better...

mgrudzinska:
> oh thank you for check. ~um.. I think handling nested nodes with `display = false` will solve the problem.~ I think we need additional member value other than `display`, such as `nested`.

this seams to be an expensive solution - each node will keep an additional field. so maybe ```SvgNode* unknownNodes``` ? this is a bit connected with the unknown nodes from the svg specs, but any other name will be fine as well

JSUYA:
> I thought rather about using level or use some separate SvgNode* to store such elements. but maybe display will be better...

Because level is the relationship between parent and child, it only applies to group nodes. If nested nodes and groups are repeated, there may eventually be a case where we need to find the graphics name.

> this seams to be an expensive solution - each node will keep an additional field. so maybe `SvgNode* unknownNodes` ? this is a bit connected with the unknown nodes from the svg specs, but any other name will be fine as well

I think unknown nodes and this problem are separate issues. The spec even says it behaves like `<g>`.
https://www.w3.org/TR/SVG2/struct.html#UnknownElement

Nested nodes are generally not drawn, but should be drawn when referenced. Therefore, I think display should not be changed because it is a unique(?) property.
this case.
```
  <rect x="2" y="2" width="6" height="4" fill="red">
    <circle id="circ" cx="10" cy="10" r="3" display="none"/>
  </rect>
  <use href="#circ" />
```

But we also have a problem in this case :)

```
  <circle id="circ" cx="10" cy="10" r="3" display="none"/>
  <use href="#circ" />
```

mgrudzinska:
> Because level is the relationship between parent and child, it only applies to group nodes. If nested nodes and groups are repeated, there may eventually be a case where we need to find the graphics name.

I thought that while reading the data we can check whether the level is equal the stack size. but we have to store them somewhere, so this was just an idea to caught such nodes. anyway, it's just an idea, didn't test it. 

> Nested nodes are generally not drawn, but should be drawn when referenced. Therefore, I think display should not be changed because it is a unique(?) property. this case.

agreed

> But we also have a problem in this case :)
> 
> ```
>   <circle id="circ" cx="10" cy="10" r="3" display="none"/>
>   <use href="#circ" />
> ```

do we? I've added ```fill="red" ```
```
<svg xmlns="http://www.w3.org/2000/svg" width="400" height="400" viewBox="0 0 124 124" fill="none">
  <circle id="circ" cx="80" cy="80" r="30" fill="red" display="non"/>
  <use href="#circ" />
</svg>
```

JSUYA:
> ```
>  <circle id="circ" cx="80" cy="80" r="30" fill="red" display="non"/>
> ```
There is typo. `non` -> `none`


I updated PR. I didn't want it either, but `_sceneBuildHelper` more complicated. If you have any better ideas, please let me know
I did some tests by changing the order of the samples below, but I didn't find any problems.

```
<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
<def>
</def>
  <rect x="2" y="2" width="6" height="4" fill="red">
    <circle id="mk" cx="12" cy="12" fill="red" r="3" />
    <circle id="circ" cx="10" cy="10" fill="red" r="3" />
     <g id="gg" >
    <use href="#circ"  clip-path="url(#mk)" />
     </g>
  </rect>
    <use href="#gg" />
</svg>

```

mgrudzinska:
@JSUYA as I've mentioned, I'm not a big fan of this solution, as it adds a new member to each node and it won't be used very often. can you please comment on why don't you want to go in the solution similar to ```def```? so to separate these nodes, store them in a different object and search for them there if needed?

JSUYA:
> @JSUYA as I've mentioned, I'm not a big fan of this solution, as it adds a new member to each node and it won't be used very often. can you please comment on why don't you want to go in the solution similar to `def`? so to separate these nodes, store them in a different object and search for them there if needed?

@mgrudzinska 
There's no reason not to use it. I think your solution is correct.
But during testing, I thought that 'defs' might not be usable in our current implementation as some examples didn't work as expected.  
For example
```
<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
<defs>
    <circle id="mk" cx="12" cy="14" fill="blue" r="3" />
     <g id="gg" >
        <use id="groupuse" href="#circ" />
     </g>
</defs>

  <rect x="2" y="2" width="6" height="4" fill="red">

     <g id="gg2" >
       <circle id="circ" cx="10" cy="10" fill="red" r="3"/>
     </g>
  </rect>
    <use href="#gg"/>
    <use href="#mk"/>
</svg>
```
However, this case is output incorrectly even if it is not a nested element.
```
<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">

<defs>
    <circle id="mk" cx="12" cy="14" fill="blue" r="3" />

     <g id="gg" >
        <use id="groupuse" href="#circ" />
     </g>

     <g id="gg2" >
       <circle id="circ" cx="10" cy="10" fill="red" r="3"/>
     </g>
</defs>

  <!--<rect x="2" y="2" width="6" height="4" fill="red"> -->
   <!-- Move to defs -->

   <!--  </rect> -->
    <use href="#gg"/>
    <use href="#mk"/>
</svg>
```
Anyway, I modified the code so that if a nested state is found, the child element becomes a child of `defs`.

mgrudzinska:
> @mgrudzinska There's no reason not to use it. I think your solution is correct. But during testing, I thought that 'defs' might not be usable in our current implementation as some examples didn't work as expected. For example
> 
> ...
> 
> Anyway, I modified the code so that if a nested state is found, the child element becomes a child of `defs`.

I don't understand what you mean. I wanted to use something similar to 'defs', but not 'defs' exactly, since these nested nodes have nothing to do with 'defs'. 
However, do I understand correctly, that your examples show that it wasn't possible to use another object and it has to be 'defs' ?

JSUYA:
> > @mgrudzinska There's no reason not to use it. I think your solution is correct. But during testing, I thought that 'defs' might not be usable in our current implementation as some examples didn't work as expected. For example
> > ...
> > Anyway, I modified the code so that if a nested state is found, the child element becomes a child of `defs`.
> 
> I don't understand what you mean. I wanted to use something similar to 'defs', but not 'defs' exactly, since these nested nodes have nothing to do with 'defs'.

Yes, you definitely said ‘similar’.
However, if a separate array is created for this case, where should the array be located?
If we place a separate array in the loader, we need to search this array more time in postpone, updategradient, etc.
And if the nested element is a group, parent management and node management for these elements are also necessary.
If nested elements are to become parents, additional work is needed to check them in sceneBuilder and prevent them from actually being drawn.
So I thought that task was more complicated and decided to take the easiest route. This is because it is not a property declared in `defs`, but it behaves as if it were declared in `defs`.
So I decided to create these as children of `defs`.
Since the loader node must reflect the contents of the SVG file as is, I do not think this approach is 100% good.
Is there a better way? I would like to solve this problem while minimizing side effects.

umm..my another idea is...
Since there is no clear specification for nested elements yet, I think these nested elements can be unsupported. without parsing them.

>  However, do I understand correctly, that your examples show that it wasn't possible to use another object and it has to be 'defs' ?

I just wanted to say that we currently have a bug with this example. The results are different from Chrome.
```
<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">

<defs>
    <circle id="mk" cx="12" cy="14" fill="blue" r="3" />

     <g id="gg" >
        <use id="groupuse" href="#circ" />
     </g>

     <g id="gg2" >
       <circle id="circ" cx="10" cy="10" fill="red" r="3"/>
     </g>
</defs>

    <use href="#gg"/>
    <use href="#mk"/>
</svg>
```

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3260,6 +3268,16 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content,
...
         else parent = loader->doc;
         node = method(loader, parent, attrs, attrsLength, simpleXmlParseAttributes);
+        if (node && !empty) {
+            auto defs = _getDefsNode(node);
+            if (!defs) {
+                defs = _createNode(nullptr, SvgNodeType::Defs);
+                loader->def = defs;
+                loader->doc->node.doc.defs = defs;
+            }
+            loader->stack.push(defs);
```
mgrudzinska, line 3278: I'd replace these lines:
```
            auto defs = _getDefsNode(node);
            if (!defs) {
                defs = _createNode(nullptr, SvgNodeType::Defs);
                loader->def = defs;
                loader->doc->node.doc.defs = defs;
            }
```
by 
```
            auto defs = _createDefsNode(loader, nullptr, nullptr, 0, nullptr);
```
but I leave the decisions to you

mgrudzinska (approved):
>Since the loader node must reflect the contents of the SVG file as is, I do not think this approach is 100% good.

exactly this is what I was thinking :)
I agree that this is very specific case, so maybe lets keep it your way, but maybe can we add a comment in the tvhSvgLoaderCommon.h file:
```
SvgNode* def = nullptr; //occasionally contains elements that are defined within graphic elements
```
or something like that
I also left another comment, but it's up to you whether you want to apply :)

hermet (approved):

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: Ooh.. i didn't add comment in `defs`. I will write this in the next.

PR #2216, opened 2024-04-23
lottie: data optimization

hermet:
lottie exclusively uses lines with the bezier curve. This approach consumes more points memory and requires additional computational processing. The Lottie loader converts 'CubicTo' to 'LineTo' only if possible,
specifically when the shape layer has only a single frame.

We've observed that thousands of commands can be converted with our Lottie example.

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -33,6 +33,17 @@
 #define KEY_AS(name) !strcmp(key, name)
 
 
+static bool _curve2line(Point& a, Point& b, Point& c, Point& d)
+{
+    auto m = (d.y - a.y) / (d.x - a.x);
+    auto by = m * (b.x - a.x) + a.y;
+    if (!mathEqual(by, b.y)) return false;
+    auto cy = m * (c.x - a.x) + a.y;
+    if (!mathEqual(cy, c.y)) return false;    
+    return true;
+}
+
+
 static char* _int2str(int num)
 {
     char str[20];
@@ -179,7 +190,7 @@ void LottieParser::getValue(TextDocument& doc)
 }
 
 
-void LottieParser::getValue(PathSet& path)
+void LottieParser::getValue(PathSet& path, bool optimize)
 {
     Array<Point> outs, ins, pts;
     bool closed = false;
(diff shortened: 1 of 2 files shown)


src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -33,6 +33,17 @@
...
 
+static bool _curve2line(Point& a, Point& b, Point& c, Point& d)
+{
+    auto m = (d.y - a.y) / (d.x - a.x);
+    auto by = m * (b.x - a.x) + a.y;
+    if (!mathEqual(by, b.y)) return false;
+    auto cy = m * (c.x - a.x) + a.y;
+    if (!mathEqual(cy, c.y)) return false;    
+    return true;
+}
```
mgrudzinska, line 44: my propositioin:
```
static bool _isBetween(const Point& check, const Point& a, const Point& b)
{
    return check.x >= mathMin(a.x, b.x) && check.x <= mathMax(a.x, b.x) &&
           check.y >= mathMin(a.y, b.y) && check.y <= mathMax(a.y, b.y);
}

static bool _curve2line(const Point& a, const Point& b, const Point& c, const Point& d)
{
    //vertical line
    if (mathEqual(a.x, d.x) && mathEqual(a.x, c.x) && mathEqual(a.x, b.x)) {
        return _isBetween(b, a, d) && _isBetween(c, a, d);
    }

    auto m = (d.y - a.y) / (d.x - a.x);
    auto by = m * (b.x - a.x) + a.y;
    if (!mathEqual(by, b.y)) return false;
    auto cy = m * (c.x - a.x) + a.y;
    if (!mathEqual(cy, c.y)) return false;
    
    return _isBetween(b, a, d) && _isBetween(c, a, d);
}
```

tvgMath.h has to be included

ex. in the 'world_locations.json' happens that `_isBetween` returns false

mgrudzinska (changes requested):
for a vertical line: (1,1), (1,2), (1,3), (1,4) functions returns ```false ```(also we get nan and inf inside it)

for points: (1,1), (2,2), (10,10), (4,4) - function returns ```true```, although the end point should be changed to be (10,10) instead of (4,4)

PR #2218, opened 2024-04-24
lottie: correct the multiline text alignment

tinyjin:
When the Lottie Text Justify sets to center, text alignment should also be center.
[74_Text_Fonts.json](https://github.com/thorvg/thorvg/files/15090114/74_Text_Fonts.json)

**[Before]**

image

**[After]**

image

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -546,9 +546,36 @@ static void _updateText(LottieGroup* parent, LottieObject** child, float frameNo
     float spacing = text->spacing(frameNo) / scale;
     Point cursor = {0.0f, 0.0f};
     auto scene = Scene::gen();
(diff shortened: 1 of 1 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -546,9 +546,35 @@ static void _updateText(LottieGroup* parent, LottieObject** child, float frameNo
     float spacing = text->spacing(frameNo) / scale;
     Point cursor = {0.0f, 0.0f};
     auto scene = Scene::gen();
+    int line = 0;
 
     //text string
-    while (*p != '\0') {
+    while (true) {
```
hermet, line 552: Could you please leave a TODO comment so that we can try it by any chance later?
```
TODO: remove nested scenes.
```

hermet (approved):

PR #2219, opened 2024-04-24
loader/svg: If link string is empty, not create picture object

JSUYA:
An empty link causes the wrong loader to be found and causes various problems.

related issue : https://github.com/thorvg/thorvg/issues/2078#issuecomment-2067726046

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -557,7 +557,7 @@ static bool _isValidImageMimeTypeAndEncoding(const char** href, const char** mim
 
 static unique_ptr<Picture> _imageBuildHelper(SvgLoaderData& loaderData, SvgNode* node, const Box& vBox, const string& svgPath)
 {
-    if (!node->node.image.href) return nullptr;
+    if (!node->node.image.href || !strlen(node->node.image.href)) return nullptr;
     auto picture = Picture::gen();
 
     TaskScheduler::async(false);    //force to load a picture on the same thread


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -558,6 +558,7 @@ static bool _isValidImageMimeTypeAndEncoding(const char** href, const char** mim
 static unique_ptr<Picture> _imageBuildHelper(SvgLoaderData& loaderData, SvgNode* node, const Box& vBox, const string& svgPath)
 {
     if (!node->node.image.href) return nullptr;
+    if (strlen(node->node.image.href) == 0) return nullptr;
```
mgrudzinska, line 561: I'd merge these two together:
```
if (!node->node.image.href || !strlen(node->node.image.href)) return nullptr;
```
but up to you

mgrudzinska (approved):

src/loaders/svg/tvgSvgSceneBuilder.cpp
JSUYA: Thanks! I updated pr

hermet:
What if href has just "single character" ? Shouldn't it fix in the loader level?

JSUYA:
> What if href has just "single character" ? Shouldn't it fix in the loader level?

If href is empty, I can set it to nullptr in the loader. But it's not really nullptr, so I wanted the builder to handle it.
And even if it is a single character, it is processed in the if statement below and processed as load fail.

hermet:
> > What if href has just "single character" ? Shouldn't it fix in the loader level?
> 
> If href is empty, I can set it to nullptr in the loader. But it's not really nullptr, so I wanted the builder to handle it. And even if it is a single character, it is processed in the if statement below and processed as load fail.

the picture::load() doesn't return fail in the emptry string?

> > What if href has just "single character" ? Shouldn't it fix in the loader level?
> 
> If href is empty, I can set it to nullptr in the loader. But it's not really nullptr, so I wanted the builder to handle it. And even if it is a single character, it is processed in the if statement below and processed as load fail.

Then single character and return fail is ok? 
It looks empty string is un-willing case. If the picture::load() with empty string is the matter here, we might need to consider to fix it in loader side rather than case-by-case exceptions outside.

JSUYA:
> > > What if href has just "single character" ? Shouldn't it fix in the loader level?
> > 
> > 
> > If href is empty, I can set it to nullptr in the loader. But it's not really nullptr, so I wanted the builder to handle it. And even if it is a single character, it is processed in the if statement below and processed as load fail.
> 
> the picture::load() doesn't return fail in the emptry string?

In this case, the svg path information is included in line 601 of tvgSvgSceneBuilder.cpp.
For picture::load() this is not an empty string

> 
> > > What if href has just "single character" ? Shouldn't it fix in the loader level?
> > 
> > 
> > If href is empty, I can set it to nullptr in the loader. But it's not really nullptr, so I wanted the builder to handle it. And even if it is a single character, it is processed in the if statement below and processed as load fail.
> 
> Then single character and return fail is ok? It looks empty string is un-willing case. If the picture::load() with empty string is the matter here, we might need to consider to fix it in loader side rather than case-by-case exceptions outside.

To be more clear, this is a processing issue with picture::load() when the file name is empty.
When the file type is not clear to the loader, we open the file and try to guess what it is.
In the process, if there is only a directory path, the same problem will occur in the process of finding the loader.

this case
```
    snprintf(buf, sizeof(buf), EXAMPLE_DIR"/svg/");

    tvg::Result ret = picture->load(buf);
```
I think this can be handled additionally in picture later (separately from this patch).
We can add more code like this to LoadModule* LoaderMgr::loader() of tvgLoader.

```
#ifdef _WIN32
    auto fileName = path.substr(path.find_last_of("\\") + 1, path.find_last_of(".") - 1);
#else
    auto fileName = path.substr(path.find_last_of("/") + 1, path.find_last_of(".") - 1);
#endif
    if (fileName.empty()) {
        *invalid = true;
        return nullptr;
    }
```

This PR may be done preemptively in svgLoader to avoid additional unnecessary checks.
Is there a direction you want?

hermet:
@JSUYA As you mentioned the PR title:  ==> "An empty link causes the wrong loader to be found" 

I want to make it sure If the picture::load() return "SUCCESS" even in the empty string. If that's the issue point, the picture loader must to be fixed.

JSUYA:
> @JSUYA As you mentioned the PR title: ==> "An empty link causes the wrong loader to be found"
> 
> I want to make it sure If the picture::load() return "SUCCESS" even in the empty string. If that's the issue point, the picture loader must to be fixed.

`Empty link`  mean is for the svg loader side.
If the picture was causing problems with the empty path, I would have written picture: "~~".
picture returns Invalid path in empty string.

I already mentioned it above
From the picture's perspective, it is an invalid file name problem.

> this case
> 
> ```
>     snprintf(buf, sizeof(buf), EXAMPLE_DIR"/svg/");
> 
>     auto picture = tvg::Picture::gen();
>     tvg::Result ret = picture->load(buf);
> ```
> 
> I think this can be handled additionally in picture later (separately from this patch). We can add more code like this to LoadModule* LoaderMgr::loader() of tvgLoader.
> 
> ```
> #ifdef _WIN32
>     auto fileName = path.substr(path.find_last_of("\\") + 1, path.find_last_of(".") - 1);
> #else
>     auto fileName = path.substr(path.find_last_of("/") + 1, path.find_last_of(".") - 1);
> #endif
>     if (fileName.empty()) {
>         *invalid = true;
>         return nullptr;
>     }
> ```

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -557,7 +557,7 @@ static bool _isValidImageMimeTypeAndEncoding(const char** href, const char** mim
 
 static unique_ptr<Picture> _imageBuildHelper(SvgLoaderData& loaderData, SvgNode* node, const Box& vBox, const string& svgPath)
 {
-    if (!node->node.image.href) return nullptr;
+    if (!node->node.image.href || !strlen(node->node.image.href)) return nullptr;
```
hermet, line 560: just one tip: this length is used at 573 line. probably, we can resue the length with a variable.

hermet (approved):

Issue #2225, opened 2024-04-28
Conan package: build failure using c++17 (or later) with MSVC.


vsaulue:
# Description

While trying to use thorvg via `conan` in a c++20 project on Windows (Visual Studio Community 17.9.3), the thorvg build fails with the following first error:

```
C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\shared\rpcndr.h(203,9): error C2872: 'byte': ambiguous symbol [C:\Users\Vincent\.conan2\p\b\thorvd6d3916f25a7e\b\build-debug\src\25a6634@@thorvg@
sta.vcxproj]
  (compiler le fichier source '../../src/src/loaders/ttf/tvgTtfLoader.cpp')
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\shared\rpcndr.h(202,23):
  could be 'unsigned char byte'
  D:\Programmes\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\include\cstddef(34,24):
  or       'std::byte'
```

My conan host profile:
```
[settings]
arch=x86_64
build_type=Debug
compiler=msvc
compiler.cppstd=20
compiler.runtime=dynamic
compiler.runtime_type=Debug
compiler.version=193
os=Windows
[conf]
tools.cmake.cmaketoolchain:generator=Ninja Multi-Config
```

# Reproduction steps

1. Open a Visual Studio command prompt
2. `cd` to an empty folder you don't mind polluting
3. `conan remove thorvg`
4. `conan install --requires="thorvg/0.13.0" --build=missing -s:h build_type=Debug -s:h compiler.cppstd=20`

### [Full build output](https://github.com/thorvg/thorvg/files/15143281/output.txt)

# Root cause ?

The C++ standard library provides [std::byte](https://en.cppreference.com/w/cpp/types/byte) since C++17. The Windows header `rpcndr.h` defines `typedef unsigned char byte;`. A [using namespace std; in thorvg](https://github.com/thorvg/thorvg/blob/5237d0868eb991064b64890e78b43a5f76664082/src/renderer/tvgCommon.h#L29) causes these types to be ambiguous.

# Workaround for users

Force conan to build the thorvg dependency in C++14 for your project, to "undefine" `std::byte`. This can be done by adding `-s thorvg/*:compiler.cppstd=14` in the `conan install` command of your project:
```bash
# Build the project's dependencies with C++20 by default, but use C++14 for thorvg.
[...]

PR #2226, opened 2024-04-29
ttf_loader : windows C++ 17 portability

rinechran:
ttf_loader : windows C++ 17 portability
https://github.com/thorvg/thorvg/issues/2225


rinechran:
The issue was specifically with the ttf_loader, so I applied the fix only in the tvgTtfLoader.cpp. This macro is only activated in MSCV, and I believe it does not pose a major problem for other compilers, hence its application is confined to this file.
image
confirmed that there are no errors after switching to C++ 20

vtorri:
i proposed a better solution in the issue

PR #2227, opened 2024-04-29
Windows: define WIN32_LEAN_AND_MEAN before including windows.h.

vtorri:
Fix issue #2225. It will also speed up compilation on Windows.


hermet:
Perhaps we could use 'config.h' to place the definition using the Meson build script.

vtorri:
in that case, i will have to remove the undef after the inclusion of `windows.h`

hermet (approved):

hermet:
tag.tvg shouldn't be modified in this PR.

vtorri:
closing , see PR #2232

Issue #2172, opened 2024-04-13
Support viewport feature for the scene.


hermet:
Since setting up a viewport region is a common method, we can officially support it as a replacement concept:

Current:
```
//use a shape for clipping
auto shape = Shape::gen();
shape->appendRect(x, y, w, h);
...
//setup its viewport
scene->composite(tvg::CompositeMethod::Clip, move(shape));
```
To be:
```
scene->viewport(x, y, w, h);
```

hermet:
It is necessary to figure out the trade-off between increased maintenance and performance improvement is acceptable.

PR #2229, opened 2024-04-29
scene viewport

mgrudzinska:
[common: scene viewport api introduced](https://github.com/thorvg/thorvg/commit/b4cd8e5207b7f83a6206001f06beba2bbd8229a2) 

New api can be used to replace clipping with a rectangle.
Scenes with viewport set can be still clipped/masked.

@issue: https://github.com/thorvg/thorvg/issues/2172

[loaders: use viewport in svg/lottie loader](https://github.com/thorvg/thorvg/commit/7d821888511029378cbc7a668c248db0c6149171) 

The viewport api is used to replace the rectangular cliping.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1394,6 +1394,36 @@ public:
     std::list<Paint*>& paints() noexcept;
 
     /**
+     * @brief Sets the scene's viewport.
+     *
+     * Allows specifying a rectangular viewport for the scene by providing the coordinates of its top-left corner and its dimensions.
+     *
+     * @param[in] x The horizontal coordinate of the upper left corner of the viewport.
+     * @param[in] y The vertical coordinate of the upper left corner of the viewport.
+     * @param[in] w The width of the viewport.
+     * @param[in] h The height of the viewport.
+     *
+     * @retval Result::InvalidArguments if the provided width @p w or height @p h are less than 0, Result::Success otherwise.
+     *
+     * @note Experimental API
+     */
+    Result viewport(int32_t x, int32_t y, int32_t w, int32_t h) noexcept;
+
+    /**
+     * @brief Gets the scene's viewport.
+     *
+     * @param[out] x The horizontal coordinate of the upper left corner of the viewport.
+     * @param[out] y The vertical coordinate of the upper left corner of the viewport.
+     * @param[out] w The width of the viewport.
+     * @param[out] h The height of the viewport.
+     *
+     * @retval Result::InsufficientCondition if the viewport was not set, Result::Success otherwise.
+     *
+     * @note Experimental API
+     */
+    Result viewport(int32_t* x, int32_t* y, int32_t* w, int32_t* h) noexcept;
+
+    /**
      * @brief Sets the total number of the paints pushed into the scene to be zero.
      * Depending on the value of the @p free argument, the paints are freed or not.
      *
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -997,15 +997,32 @@ static void _updatePrecomp(LottieLayer* precomp, float frameNo, LottieExpression
 
     //clip the layer viewport
     if (precomp->w > 0 && precomp->h > 0) {
-        auto clipper = Shape::gen().release();
-        clipper->appendRect(0, 0, static_cast<float>(precomp->w), static_cast<float>(precomp->h));
-        clipper->transform(precomp->cache.matrix);
+        auto& m = precomp->cache.matrix;
 
-        //TODO: remove the intermediate scene....
-        auto cscene = Scene::gen();
-        cscene->composite(cast(clipper), CompositeMethod::ClipPath);
-        cscene->push(cast(precomp->scene));
-        precomp->scene = cscene.release();
+        if (!mathIdentity((const Matrix*)&m)) {
+            //rotated/skewed rectangle - composition required
+            if (!mathRightAngle(&m) || mathSkewed(&m)) {
+                auto clipper = Shape::gen().release();
+                clipper->appendRect(0, 0, precomp->w, precomp->h);
+                clipper->transform(m);
+
+                //intermediate scene needed
+                if (precomp->scene->composite(nullptr) != CompositeMethod::None) {
+                    auto cscene = Scene::gen();
+                    cscene->composite(cast(clipper), CompositeMethod::ClipPath);
+                    cscene->push(cast(precomp->scene));
+                    precomp->scene = cscene.release();
+                } else {
+                    precomp->scene->composite(cast(clipper), CompositeMethod::ClipPath);
+                }
+            //axis aligned rectangle - in order to optimize replaced by viewport
+            } else {
+                precomp->scene->viewport(static_cast<int32_t>(m.e13), static_cast<int32_t>(m.e23),
+                                         static_cast<int32_t>(precomp->w * m.e11), static_cast<int32_t>(precomp->h * m.e22));
+            }
+        } else {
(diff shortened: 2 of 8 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1000,11 +1000,14 @@ static void _updatePrecomp(LottieLayer* precomp, float frameNo)
         clipper->appendRect(0, 0, static_cast<float>(precomp->w), static_cast<float>(precomp->h));
         clipper->transform(precomp->cache.matrix);
 
-        //TODO: remove the intermediate scene....
-        auto cscene = Scene::gen();
-        cscene->composite(cast(clipper), CompositeMethod::ClipPath);
-        cscene->push(cast(precomp->scene));
-        precomp->scene = cscene.release();
+        if (precomp->scene->composite(nullptr) != CompositeMethod::None) {
```
hermet, line 1003: We assume the use of the new viewport API here.

inc/thorvg.h
```diff
@@ -1393,6 +1393,38 @@ class TVG_API Scene final : public Paint
...
+     *
+     * @retval Result::Success When succeed.
+     * @retval Result::InvalidArguments In case the provided width @p w or height @p h are less than 0.
+     * @retval Result::FailedAllocation An internal error with a memory allocation.
+     *
+     * @note Experimental API
+     */
+    Result viewport(int32_t x, int32_t y, int32_t w, int32_t h) noexcept;
+
+    /**
```
hermet, line 1414: Let's remove the getter API.

src/renderer/tvgPaint.cpp
```diff
@@ -40,6 +40,47 @@
...
+namespace {
+    enum ViewportType: uint8_t
+    {
+        General = 0,
+        AxisAligned,
+        None
+    };
+}
+
+static ViewportType _updateViewport(Paint* const paint, RenderMethod* const renderer, const RenderTransform* const transform, Array<RenderData>& clips, RenderUpdateFlag pFlag, RenderData& rd)
```
hermet, line 52: Viewport is a scene function. It probably could intersect the current viewport with the scene viewport in scene layer?

src/renderer/tvgScene.h
```diff
@@ -77,6 +84,11 @@ struct Scene::Impl
         if (auto renderer = PP(scene)->renderer) {
             renderer->dispose(rd);
         }
+
+        if (vport) {
```
hermet, line 88: --null check

src/renderer/tvgScene.h
```diff
@@ -209,9 +221,32 @@ struct Scene::Impl
...
 
+        if (vport) dup->viewport(vport->vregion.x, vport->vregion.y, vport->vregion.w, vport->vregion.h);
+
         return ret;
     }
 
+    bool viewport(int32_t x, int32_t y, int32_t w, int32_t h)
+    {
+        if (!vport) {
+            vport = static_cast<Viewport*>(calloc(1, sizeof(Viewport)));
```
hermet, line 232: malloc

src/renderer/tvgScene.h
```diff
@@ -209,9 +221,32 @@ struct Scene::Impl
...
+            if (!vport) return false;
```
hermet, line 233: --

src/renderer/tvgScene.cpp
```diff
@@ -72,4 +72,19 @@ Result Scene::clear(bool free) noexcept
 list<Paint*>& Scene::paints() noexcept
 {
     return pImpl->paints;
+}
+
+
+Result Scene::viewport(int32_t x, int32_t y, int32_t w, int32_t h) noexcept
```
hermet, line 78: how about `uint32_t` ? values must be larger than 0

hermet (commented):
quick feedbacks, thanks.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -861,15 +861,9 @@ Scene* svgSceneBuild(SvgLoaderData& loaderData, Box vBox, float w, float h, Aspe
...
-    viewBoxClip->appendRect(0, 0, w, h);
-
-    auto compositeLayer = Scene::gen();
-    compositeLayer->composite(std::move(viewBoxClip), CompositeMethod::ClipPath);
-    compositeLayer->push(std::move(docNode));
-
     auto root = Scene::gen();
-    root->push(std::move(compositeLayer));
+    root->push(std::move(docNode));
+    root->viewport(0, 0, static_cast<int32_t>(w), static_cast<int32_t>(h));
```
JSUYA, line 866: Just curious...
Doesn't this affect the result of `bounds()`? I haven't tested it yet...
The result of `bounds()` after `CompositeMethod::ClipPath` and the result of `bounds()` after viewport are expected to be different. (Until at least one `update()` occurs...)

src/renderer/tvgScene.h
mgrudzinska: I need to check it somewhere, but now I've compressed it:
```
        if (vport) delete(vport->vclip);
        free(vport);
```

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: but the clipper is transformed, it may not be a rectangle

src/loaders/svg/tvgSvgSceneBuilder.cpp
hermet: @mgrudzinska  bounds() wouldn't occupy the compositors regions including clipper. Please double-check it.

src/loaders/lottie/tvgLottieBuilder.cpp
hermet: The primary purpose of this viewport optimization is for this specific application, not others. If it's not applicable here, then it might be unnecessary. Perhaps we should internally consider whether the final transformed viewport is axis-aligned. Otherwise, this idea is wrong approach from the very-beginning.

mgrudzinska:
I'll check the regression bugs

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -861,15 +865,12 @@ Scene* svgSceneBuild(SvgLoaderData& loaderData, Box vBox, float w, float h, Aspe
...
-    auto compositeLayer = Scene::gen();
-    compositeLayer->composite(std::move(viewBoxClip), CompositeMethod::ClipPath);
-    compositeLayer->push(std::move(docNode));
-
     auto root = Scene::gen();
-    root->push(std::move(compositeLayer));
+    root->push(std::move(docNode));
+
+    float a, b, c, d;
+    root->bounds(&a, &b, &c, &d);
```
JSUYA, line 872: Is this code we need?

mgrudzinska: nope! these are leftovers after me checking the bounds, thx!

hermet:
Performance is enhanced by ~3% with Lottie example

src/renderer/tvgScene.h
```diff
@@ -209,9 +256,28 @@ struct Scene::Impl
...
 
+    void viewport(int32_t x, int32_t y, int32_t w, int32_t h)
+    {
+        if (!vport) vport = new Viewport();
+        vport->vregion = {x, y, w, h};
+    }
+
+    bool viewport(int32_t* x, int32_t* y, int32_t* w, int32_t* h)
+    {
+        if (!vport) return false;
```
tinyjin, line 272: Perhaps, It'd be better if we check the case all parameters are `nullptr`
```cpp
if (!vport || !(x || y || w || h)) return false;
```

hermet:
close this since this is unlikely to be applied.

PR #2244, opened 2024-05-05
ttf_loader: fixing font scaling factor

mgrudzinska:
Currently, both svg and ttf loaders assume
a PPI value of 96. Considering a dynamic
determination of PPI for specific devices
should be a global and consistent approach
across all loaders.

### DIFF
--- a/examples/Text.cpp
+++ b/examples/Text.cpp
@@ -62,7 +62,7 @@ void tvgDrawCmds(tvg::Canvas* canvas)
     canvas->push(std::move(text));
 
     auto text2 = tvg::Text::gen();
-    text2->font("Arial", 30, "italic");
+    text2->font("Arial", 40, "italic");
     text2->text("Font = \"Arial\", Size = 40, Style=Italic");
     text2->translate(0, 150);
     text2->fill(255, 255, 255);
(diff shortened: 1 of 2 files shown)


hermet (changes requested):
Could you please suggest a potential bug fix or an improvement here? I agree the DPI should be a variable value in ThorVG, but I haven't observed any improvements with your fix.  This change would just break the behavior, similar to how you had to modify the examples.

mgrudzinska:
> Could you please suggest a potential bug fix or an improvement here? I agree the DPI should be a variable value in ThorVG, but I haven't observed any improvements with your fix. This change would just break the behavior, similar to how you had to modify the examples.

the problem can be observed only while comparing with something else. like in an svg:
```
<svg viewBox="0 0 140 50" xmlns="http://www.w3.org/2000/svg">
  <text x="5" y="28" fill="blue" font-family="Monaco" font-size="30"> I love</text>
  <rect x="5" y="0" width="10" height="28" fill="green" opacity="0.5"/>
</svg>
```
before this change:
Zrzut ekranu 2024-05-6 o 10 02 37

after (as expected):
Zrzut ekranu 2024-05-6 o 10 01 52

we can think about detecting dynamically the ppi, but it's a different task

hermet:
@mgrudzinska I wonder if you have ensured that your system's text size does not apply any additional text scalability. 
Something like this:
image

ThorVG shouldn't take care of these kind of extra scalability variables for text, engine inside.

mgrudzinska:
@hermet can you explain the idea behind the 96/72 factor? where does the 72 come from? I thought we were currently assuming that the display's PPI is 96.

mgrudzinska:
```The size of the font in points. This determines how large the text will appear.```
I was sure it's in pixels. sorry

hermet:
For now, i think you should not touch this if you are considering dpi 96. it should be working correctly.

PR #2251, opened 2024-05-06
common: fix a precision issue in appendArc api

mgrudzinska:
For sweep angles very close to a multiple of pi/2, an extra cubic curve segment was added, which was
populated with NaN values due to finite precision limits.

### DIFF
--- a/src/renderer/tvgShape.cpp
+++ b/src/renderer/tvgShape.cpp
@@ -145,13 +145,15 @@ Result Shape::appendArc(float cx, float cy, float radius, float startAngle, floa
     //just circle
     if (sweep >= 360.0f || sweep <= -360.0f) return appendCircle(cx, cy, radius, radius);
 
+    const float arcPrecision = 1e-5f;
     startAngle = mathDeg2Rad(startAngle);
     sweep = mathDeg2Rad(sweep);
 
-    auto nCurves = ceil(fabsf(sweep / MATH_PI2));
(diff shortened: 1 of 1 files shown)


mgrudzinska:
sample that didn't work on my machine:
```
    shape1->appendArc(150, 250, 180, 350, 90.000012, true);
    shape1->strokeFill(255, 0, 0);
    shape1->strokeWidth(13);
    if (canvas->push(std::move(shape1)) != tvg::Result::Success) return;

    shape2->appendArc(150, 350, 180, 360, 90.000012, true);
    shape2->strokeFill(255, 255, 255);
    shape2->strokeWidth(2);
    if (canvas->push(std::move(shape2)) != tvg::Result::Success) return;
```

before:
the first arc (he second one contained NaN values and caused the program to hang):
Zrzut ekranu 2024-05-6 o 22 41 03


after:
Zrzut ekranu 2024-05-6 o 22 40 06

mgrudzinska:
another example:
```
    auto shape3 = tvg::Shape::gen();
    shape3->appendArc(600, 450, 180, 10, 0.000002, true);
    shape3->strokeFill(255, 255, 255);
    shape3->strokeWidth(2);
    if (canvas->push(std::move(shape3)) != tvg::Result::Success) return;
```
   
before:
Zrzut ekranu 2024-05-6 o 23 34 21

after:
Zrzut ekranu 2024-05-6 o 23 33 30

hermet (approved):

PR #2252, opened 2024-05-07
lottie: enable the expressions feature in default

hermet:
this commit introduces an additional build options:

- lottie expressions: this advanced feature in Lottie can significantly increase binary size. Users now have the option to enable or disable it based on their requirements.

Note that, this change introduces one config definitions:

- THORVG_LOTTIE_EXPRESSIONS_SUPPORT

### DIFF
--- a/test/testAnimation.cpp
+++ b/test/testAnimation.cpp
@@ -36,7 +36,7 @@ TEST_CASE("Animation Basic", "[tvgAnimation]")
     REQUIRE(animation);
 
     auto picture = animation->picture();
-    REQUIRE(picture->identifier == Picture::identifier);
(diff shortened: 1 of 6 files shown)


test/testAnimation.cpp
```diff
@@ -36,7 +36,7 @@ TEST_CASE("Animation Basic", "[tvgAnimation]")
     REQUIRE(animation);
 
     auto picture = animation->picture();
-    REQUIRE(picture->identifier == Picture::identifier);
```
mgrudzinska, line 39: the same should be changed in testLottie l. 45 and 81

mgrudzinska (approved):

Issue #2223, opened 2024-04-26
lottie: inccorect glyph positions


hermet:
An  inccorrect glyph positions caused by a bug or missing property support. Need to investigate.
(left: thorvg, right: lottie-web)

[calculator.json](https://github.com/thorvg/thorvg/files/15124519/calculator.json)

![image](https://github.com/thorvg/thorvg/assets/3711518/a7a57dae-1dfe-4020-bafd-1908a495cdf4)

PR #2254, opened 2024-05-08
lottie/text: Support text tracking

tinyjin:
Text Tracking value("tr") is parsed and never used. Calculate text spacing size via the tracking offset.

I checked Lottie spec docs [here](https://lottiefiles.github.io/lottie-docs/text/#text-document) for "tr" attribute. However it doesn't seem clear to explain them.

I referred original logic on [lottie-web](https://github.com/airbnb/lottie-web/blob/master/build/player/lottie_canvas.js#L10710) to calculate `trackingOffset` value.

Issue: #2223 

[Before]
Screenshot 2024-05-08 at 6 11 33 PM

[After]
Screenshot 2024-05-08 at 6 10 59 PM

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -606,7 +606,7 @@ static void _updateText(LottieGroup* parent, LottieObject** child, float frameNo
                 p += glyph->len; 
 
                 //advance the cursor position horizontally
-                cursor.x += glyph->width + spacing;
+                cursor.x += glyph->width + spacing + doc.tracking;
                 break;
             }
         }
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -183,7 +183,7 @@ void LottieParser::getValue(TextDocument& doc)
         else if (KEY_AS("f")) doc.name = getStringCopy();
         else if (KEY_AS("t")) doc.text = getStringCopy();
         else if (KEY_AS("j")) doc.justify = getInt();
-        else if (KEY_AS("tr")) doc.tracking = getInt();
+        else if (KEY_AS("tr")) doc.tracking = getFloat() * 0.1f;
         else if (KEY_AS("lh")) doc.height = getFloat();
         else if (KEY_AS("ls")) doc.shift = getFloat();
         else if (KEY_AS("fc")) getValue(doc.color);
--- a/src/loaders/lottie/tvgLottieProperty.h
+++ b/src/loaders/lottie/tvgLottieProperty.h
@@ -75,8 +75,8 @@ struct TextDocument
     } stroke;
     char* name = nullptr;
     float size;
+    float tracking;
     uint8_t justify;
-    uint8_t tracking;
 };
 
 


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -545,6 +545,7 @@ static void _updateText(LottieGroup* parent, LottieObject** child, float frameNo
 
     auto scale = doc.size * 0.01f;
     float spacing = text->spacing(frameNo) / scale;
+    float trackingOffset = doc.tracking / 1000.0f * (doc.size / scale);
```
hermet, line 548: same logic?
```
auto tracking = doc.tracking * 0.1f
```

tinyjin: If the `doc.size` is zero, Yes. My test sets also don't have that value, it works.
But i'm not sure if it could cause different result in case it has `doc.size` value.

I've never had lottie file has `doc.size` value, Have you?

hermet: if the doc.size(font) size zero, then it could skip the text at all.
tracking doesn't need to take care of that scenario.

hermet: if the doc.size is zero, then scale value will be 0. it would not draw anything about text. So tracking doesn't need to care about it.

tinyjin: You're right, sorry never mind the comment!

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -543,8 +543,8 @@ static void _updateText(LottieGroup* parent, LottieObject** child, float frameNo
 
     if (!p || !text->font) return;
 
-    auto scale = doc.size * 0.01f;
     float spacing = text->spacing(frameNo) / scale;
+    float trackingOffset = doc.tracking * 0.1f;
```
hermet, line 547: I think we can pre-multiply this in the parsing time?

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -183,7 +183,7 @@ void LottieParser::getValue(TextDocument& doc)
         else if (KEY_AS("f")) doc.name = getStringCopy();
         else if (KEY_AS("t")) doc.text = getStringCopy();
         else if (KEY_AS("j")) doc.justify = getInt();
-        else if (KEY_AS("tr")) doc.tracking = getInt();
+        else if (KEY_AS("tr")) doc.tracking = getInt() * 0.1f;
```
hermet, line 186:
```
getInt() -> getFloat()
```

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -76,7 +76,7 @@ struct TextDocument
     char* name = nullptr;
     float size;
     uint8_t justify;
-    uint8_t tracking;
+    float tracking;
```
hermet, line 79: let's change the line location with `justify` for compact memory packing.

hermet (approved):

Issue #2021, opened 2024-03-03
Lottie regression bug


hermet:
This patch broke the lottie trimpath animations:
https://github.com/thorvg/thorvg/pull/1985

[threads.json](https://github.com/thorvg/thorvg/files/14471740/threads.json)
[sample2.json](https://github.com/thorvg/thorvg/files/14471745/sample2.json)

hermet:
@mgrudzinska

PR #2259, opened 2024-05-09
common: tweaking bezier computation for perf.

hermet:
It will sacrifice precision that is not noticeable and instead try to improve performance.

Approximately 5% has beeen improved at Lottie example

### DIFF
--- a/src/common/tvgLines.cpp
+++ b/src/common/tvgLines.cpp
@@ -23,7 +23,7 @@
 #include "tvgMath.h"
 #include "tvgLines.h"
 
-#define BEZIER_EPSILON 1e-4f
+#define BEZIER_EPSILON 1e-2f
 
 /************************************************************************/
 /* Internal Class Implementation                                        */


mgrudzinska (approved):
I checked the resources from issues that I remember were sensitive to precision in bezier calculations:
https://github.com/thorvg/thorvg/issues/2021
https://github.com/thorvg/thorvg/issues/1686
I don't see any differences. so ok.

generally I think we could ifdef all the places where we reduce precision and set a flag in meson. By default, we would use approximations, but there would be an option to perform precise calculations

hermet:
> I checked the resources from issues that I remember were sensitive to precision in bezier calculations: #2021 #1686 I don't see any differences. so ok.
> 
> generally I think we could ifdef all the places where we reduce precision and set a flag in meson. By default, we would use approximations, but there would be an option to perform precise calculations

Let's see if it's necessary or not.

PR #2263, opened 2024-05-09
lottie: replace cubics with lines where possible

mgrudzinska:
When the loaded control points are zero (in and out), there is no need to apply bezier curves - they can be replaced with straight lines.

### DIFF
--- a/src/common/tvgMath.h
+++ b/src/common/tvgMath.h
@@ -62,6 +62,12 @@ static inline bool mathZero(float a)
 }
 
 
+static inline bool mathZero(const Point& p)
+{
+    return mathZero(p.x) && mathZero(p.y);
+}
+
+
 static inline bool mathEqual(float a, float b)
 {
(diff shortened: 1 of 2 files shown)


hermet:
@mgrudzinska Does this necessary for rounded corner? I've assumed it would but the performance enhancement for Curve 2 Line is not clear yet. Let's carefully review it before confirming.

mgrudzinska:
> @mgrudzinska Does this necessary for rounded corner? I've assumed it would but the performance enhancement for Curve 2 Line is not clear yet. Let's carefully review it before confirming.

@hermet  no it's not necessary for rounded corner. 
also it will require some extra checks in the rounded corners implementation if we apply this. 

I just wanted to point it out that this is an easier solution then using ```_curve2line``` from https://github.com/thorvg/thorvg/pull/2216

mgrudzinska:
1) mathZero -> ==
2) will affect the rounded corners

Issue #2230, opened 2024-04-30
lottie: support Shape Modifier feature


hermet:
- [x] Repeater
- [x] Trim Path
- [x] Rounded Corners
- [x] Offset Path
- [ ] Merge Path

See: https://lottiefiles.github.io/lottie-docs/shapes/

hermet:
Rounded Corners is partially work. Lottie Path(Bezier Curve) is not supported currently.

hermet:
https://github.com/LottieFiles/dotlottie-web/issues/190

hermet:
See also MergeShape: https://github.com/thorvg/thorvg/issues/26

hermet:
https://github.com/thorvg/thorvg/issues/2769

PR #2267, opened 2024-05-12
lottie: support shape modifiers - rounded corner

mgrudzinska:
Implementation of the roundness property
for a polystar. Only the corners that were
not subject to rounding by the inner or
outer roundness parameters are affected.

Implemented rounding of corners between
bezier curves that are straight lines.

@Issue: https://github.com/thorvg/thorvg/issues/2230

before:
before_r

after (as in AE):
after_r

before:
before_p
beforep2

after:
after_p
afterp2

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -614,9 +614,10 @@ static void _updateText(LottieGroup* parent, LottieObject** child, float frameNo
 }
 
 
-static void _updateStar(LottieGroup* parent, LottiePolyStar* star, Matrix* transform, float frameNo, Shape* merging, LottieExpressions* exps)
+static void _updateStar(LottieGroup* parent, LottiePolyStar* star, Matrix* transform, float roundness, float frameNo, Shape* merging, LottieExpressions* exps)
 {
     static constexpr auto POLYSTAR_MAGIC_NUMBER = 0.47829f / 0.28f;
+    static constexpr auto ROUNDED_POLYSTAR_MAGIC_NUMBER = 0.47829f;
 
     auto ptsCnt = star->ptsCnt(frameNo, exps);
     auto innerRadius = star->innerRadius(frameNo, exps);
@@ -633,6 +634,8 @@ static void _updateStar(LottieGroup* parent, LottiePolyStar* star, Matrix* trans
     auto numPoints = size_t(ceilf(ptsCnt) * 2);
     auto direction = (star->direction == 0) ? 1.0f : -1.0f;
     auto hasRoundness = false;
+    bool applyExtRoundness = roundness > ROUNDNESS_EPSILON && (mathZero(innerRoundness) || mathZero(outerRoundness));
+    auto shape = applyExtRoundness ? Shape::gen().release() : merging;
 
     float x, y;
 
@@ -652,17 +655,17 @@ static void _updateStar(LottieGroup* parent, LottiePolyStar* star, Matrix* trans
     }
 
     if (mathZero(innerRoundness) && mathZero(outerRoundness)) {
-        P(merging)->rs.path.pts.reserve(numPoints + 2);
-        P(merging)->rs.path.cmds.reserve(numPoints + 3);
+        P(shape)->rs.path.pts.reserve(numPoints + 2);
+        P(shape)->rs.path.cmds.reserve(numPoints + 3);
     } else {
-        P(merging)->rs.path.pts.reserve(numPoints * 3 + 2);
-        P(merging)->rs.path.cmds.reserve(numPoints + 3);
+        P(shape)->rs.path.pts.reserve(numPoints * 3 + 2);
+        P(shape)->rs.path.cmds.reserve(numPoints + 3);
         hasRoundness = true;
     }
 
     Point in = {x, y};
     if (transform) mathTransform(transform, &in);
-    merging->moveTo(in.x, in.y);
+    shape->moveTo(in.x, in.y);
 
     for (size_t i = 0; i < numPoints; i++) {
         auto radius = longSegment ? outerRadius : innerRadius;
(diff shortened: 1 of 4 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -37,6 +37,8 @@
 /* Internal Class Implementation                                        */
 /************************************************************************/
 
+#define ROUNDNESS_EPSILON 1.0f
```
hermet, line 40: We must standardize a single definition in the roundcorner PR; otherwise, I suggest removing the constant because 1.0f itself already looks clear and deterministic, not like other magic numbers.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -710,16 +715,76 @@ static void _updateStar(LottieGroup* parent, LottiePolyStar* star, Matrix* trans
...
+
+    if (applyExtRoundness) {
+        auto cmdCnt = shape->pathCommands(nullptr);
+        const Point *pts = nullptr;
+        auto ptsCnt = shape->pathCoords(&pts);
+
+        auto len = mathLength(pts[1] - pts[2]);
+        auto r = len > 0.0f ? ROUNDED_POLYSTAR_MAGIC_NUMBER * mathMin(len * 0.5f, roundness) / len : 0.0f;
+
+        if (hasRoundness) {
```
hermet, line 737: You might wanna double-check, we can skip round corner handling in this case—rounding a star—because we do not apply rounded corners to curved paths, only to straight lines.

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: this one is needed in the tvgLottieBuilder.cpp and in the tvgLottieProperty.h so I'd like to keep it as a const in tvgLottieProperty.h

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: if in a bezier ```start == crtl1``` or```ctrl2 == end``` then roundness should be applied

|this branch | AE|
|------------|----|
|Zrzut ekranu 2024-05-13 o 13 11 40|Zrzut ekranu 2024-05-13 o 13 11 45|
|Zrzut ekranu 2024-05-13 o 13 12 34|Zrzut ekranu 2024-05-13 o 13 12 40|
|Zrzut ekranu 2024-05-13 o 13 12 02|Zrzut ekranu 2024-05-13 o 13 12 11|

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -30,6 +30,7 @@
 #include "tvgLottieInterpolator.h"
 #include "tvgLottieExpressions.h"
 
+#define ROUNDNESS_EPSILON 1.0f
```
hermet, line 33: I'm leaving a comment just in case, since the name and the number are a bit suspicious to use as the macro. 

```
//clean to understand, a obvious logic.
if (value == 0)  prepare();
if (value < 1) progress();
if (value == 1) finish();
````

```
//the numbers has been changed to text but practically, readability doesn't improved
//What is Epsilon? Readers might need to look into the EPSILON to understand the logic.
if (value == ZERO) prepare();
if (value < EPSILON) progress();
if (value == EPSILON) finish();
```
.
.
.
.
.
.
.
.
.
.

```
#define ZERO 0
#define EPSILON 1   //Ahhh. it's just 1.
```

IMO, from time to time, CONSTANT MACROS just increases the CODE longer, even it could make the maintenance worse. These guidance might help.

A. The value is **experimental number**, it can be **modifiable and it is used in several places**, allowing us to update them quickly whenever necessary. => Definitely good for maintenance.

B.  Numbers are often unexpectable magic numbers, such as PI (3.1416321321), or PATH_KAPPA (0.552284.......),  POLYSTAR_MAGIC_NUMBER(0.47829f / 0.28f) => helpful for readability.

C. People really have no idea where those complicated numbers come from, so we can apply them with explainable names =>  helpful for readability.

File: src/loaders/lottie/tvgLottieProperty.h [hunk 4]
```diff
@@ -238,6 +239,88 @@ static void _copy(PathSet& pathset, Array<PathCommand>& outCmds)
 }
 
 
+static void _roundCorner(Array<PathCommand>& cmds, Array<Point>& pts, const Point& prev, const Point& curr, const Point& next, float roundness, bool continuity, Point& distPrev)
+{
+    auto lenPrev = mathLength(prev - curr);
```
hermet, line 244: It looks, we can make it optimal by avoiding sqrt() call.

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -455,21 +538,33 @@ struct LottiePathSet : LottieProperty
         return (*frames)[frames->count];
     }
 
-    bool operator()(float frameNo, Array<PathCommand>& cmds, Array<Point>& pts, Matrix* transform)
+    bool operator()(float frameNo, Array<PathCommand>& cmds, Array<Point>& pts, Matrix* transform, float roundness)
     {
         if (!frames) {
+            if (roundness > ROUNDNESS_EPSILON) {
+                _handleCorners(value, cmds, pts, transform, roundness);
```
hermet, line 545: These handleCorners could be inlined, could you please confirm the binary size? probably, implementing functions in cpp file might be better for it. Here is a way to check the bin size.
```
$strip libthorvg.so.xxx
$size libthorvg.so.xxx
```

src/loaders/lottie/tvgLottieProperty.h
mgrudzinska: that was the intention while introducing continuity, but I've mixed up something. now fixed

src/loaders/lottie/tvgLottieProperty.h
mgrudzinska: so the conclusion is that you want me to remove ROUNDNESS_EPSILON?

src/loaders/lottie/tvgLottieProperty.h
mgrudzinska: after introducing cpp the bin size is bigger (if I inline the fun and if I don't)

src/loaders/lottie/tvgLottieProperty.h
hermet: @mgrudzinska If you think this makes sense, feel free to change it; otherwise, you can keep it as it is.

hermet:
abandoned the [lottie: optimize roundCorner()](https://github.com/thorvg/thorvg/pull/2267/commits/69a515d699256111858890bb6bc19ee16e372d91)

Issue #2266, opened 2024-05-10
`tvg_set_frame` API returns imprecise frame number in `tvg_get_current_frame` after upgrade to v0.13.3


theashraf:
When upgrading thorvg from version v0.13.0 to v0.13.3 in [dotLottie-rs](https://github.com/LottieFiles/dotlottie-rs), we encountered a failing test related to the tvg_set_frame API. When tvg_set_frame is called with a specific frame number and tvg_get_current_frame is subsequently called, the returned frame number occasionally differs slightly from what was originally set, causing the player to render a different frame than expected.

During the dotLottie player's event tests, we observed discrepancies between the frame number set by the user and the frame actually rendered. The mismatch is evident in the recorded events from the test case, shown below:

Recorded Events: 
```json
[
    "on_load",
    "on_play",
    "on_frame: 1",
    "on_render: 1",
    "on_frame: 2",
    "on_render: 2",
    "on_frame: 3", // This is when the player calls tvg_set_frame
    "on_render: 3.0000002", // This is when the player triggers a render event with the frame number returned from tvg_get_current_frame()
    "on_frame: 4",
    "on_render: 4",
    "on_frame: 5",
    "on_render: 5",
    "on_frame: 6",
    "on_render: 6.0000005",
    "on_frame: 7",
    "on_render: 7.0000005",
    "on_frame: 8",
    "on_render: 8",
    "on_frame: 9",
    "on_render: 9",
    "on_frame: 10",
    "on_render: 10",
    "on_frame: 11",
    "on_render: 11.000001",
    "on_frame: 12",
    "on_render: 12.000001",
    "on_frame: 13",
    "on_render: 13.000001",
    "on_frame: 14",
    "on_render: 14.000001",
    "on_frame: 15",
    "on_render: 15.000001",
    "on_frame: 16",
    "on_render: 16",
    "on_frame: 17",
    "on_render: 17",
    "on_frame: 18",
    "on_render: 18",
    "on_frame: 19",
    "on_render: 19",
    "on_frame: 20",
    "on_render: 20",
    "on_frame: 21",
    "on_render: 21.000002",
    "on_frame: 22",
    "on_render: 22.000002",
    "on_frame: 23",
    "on_render: 23.000002",
    "on_frame: 24",
    "on_render: 24.000002",
    "on_frame: 25",
    "on_render: 25.000002",
    "on_frame: 26",
    "on_render: 26.000002",
    "on_frame: 27",
    "on_render: 27.000002",
[...]

PR #2270, opened 2024-05-13
lottie: regression fix & improve the frame calculation.

hermet:
issue: https://github.com/thorvg/thorvg/issues/2266

### DIFF
--- a/test/testAnimation.cpp
+++ b/test/testAnimation.cpp
@@ -47,6 +47,44 @@ TEST_CASE("Animation Basic", "[tvgAnimation]")
 
 #ifdef THORVG_LOTTIE_LOADER_SUPPORT
 
+TEST_CASE("Animation Frames Counting", "[tvgAnimation]")
+{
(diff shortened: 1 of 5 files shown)


test/testAnimation.cpp
```diff
@@ -60,10 +60,32 @@ TEST_CASE("Animation Lottie", "[tvgAnimation]")
     REQUIRE(picture->load(TEST_DIR"/invalid.json") == Result::InvalidArguments);
     REQUIRE(picture->load(TEST_DIR"/test.json") == Result::Success);
 
+    REQUIRE(animation->totalFrame() == Approx(120).margin(004004));
```
mgrudzinska, line 63: 004004 ? we have this value in few places in this test

hermet: Just one sample point. Its not the main point in this PR. Please let us not get side tracked.

mgrudzinska (approved):

tinyjin (approved):

Issue #2269, opened 2024-05-13
GLES version upgrade (v2 -> v3)


hermet:
Should `gles_dep` also be updated to `GLESv3`? Tried building on Android, got error `undefined reference to `glGenVertexArrays'`, upgrading library here fixed a problem.

_Originally posted by @dkaraush in https://github.com/thorvg/thorvg/pull/1589#discussion_r1597668757_

SergeyLebedkin:
GLES 3.0 supports from Android 4.3 (2012 year)
So I think it make a sence

hermet:
@SergeyLebedkin could you please make a patch?

SergeyLebedkin:
> @SergeyLebedkin could you please make a patch?

As I can see related patch is already merged. So what we need to do?

PR #2271, opened 2024-05-13
gl_engine: updated GLES minimum requirement from v2 to v3

hermet:
GLES v3 was introduced many years ago;
we are now dropping support for GLES v2.

issue: https://github.com/thorvg/thorvg/issues/2269

### DIFF
--- a/src/renderer/gl_engine/meson.build
+++ b/src/renderer/gl_engine/meson.build
@@ -21,7 +21,7 @@ source_file = [
    'tvgGlTessellator.h',
 ]
 
-gles_dep = meson.get_compiler('cpp').find_library('GLESv2')
+gles_dep = meson.get_compiler('cpp').find_library('GLESv3')
 
 external_dep = [gles_dep]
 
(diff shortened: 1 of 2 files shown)


SergeyLebedkin (approved):

src/renderer/gl_engine/meson.build
```diff
@@ -21,7 +21,7 @@ source_file = [
    'tvgGlTessellator.h',
 ]
 
-gles_dep = meson.get_compiler('cpp').find_library('GLESv2')
+gles_dep = meson.get_compiler('cpp').find_library('GLESv3')
```
RuiwenTang, line 24: Should we add some check on the GLESv3 finding ?
I'm not sure, but the GLES3 implementation of some drivers is also included in libGLESv2.so. I remember mesa seems to be like this.

Is it better to do as below?
```
gl_deps = [ meson.get_compiler('cpp').find_library('GLESv2') ]

glesv3 = meson.get_compiler('cpp').find_library('GLESv3', required: false)

if glesv3.found()
    gl_deps += [glesv3]
endif
```

hermet: @RuiwenTang apis will be duplicated. no api conflicts?

RuiwenTang: I'm not sure, I just remember encountering this situation. Maybe we can add the check in the future if encounter compilation problems.

hermet: @RuiwenTang Ok, you approve?

RuiwenTang (approved):

hermet:
@RuiwenTang come up with https://github.com/thorvg/thorvg/pull/2279

Issue #2255, opened 2024-05-08
Segmentation Fault When Resetting Animation Overrides


theashraf:
I am encountering a segmentation fault error in [dotlottie-rs](https://github.com/LottieFiles/dotlottie-rs) when trying to reset the applied slots overrides object for this Lottie animation: [favorites_active.json](https://github.com/thorvg/thorvg/files/15251790/favorites_active.json).

The applied slots overrides are as follows:

```json
{
    "lottie-icon-outline": {
        "p": {
            "a": 0,
            "k": [1, 1, 0]
        }
    },
    "lottie-icon-solid": {
        "p": {
            "a": 0,
            "k": [0, 0, 1]
        }
    }
}
```

Related Issue [issue #188](https://github.com/LottieFiles/dotlottie-web/issues/188) in the `dotlottie-web`

Please note that the `lottie-icon-solid` property is animated color in the provided animation with keyframes. Changing it to non-animated in the overrides could be causing the issue 👀 ?

tinyjin:
Thanks, @theashraf. We will have a look on this as soon as possible.

tinyjin:
Hello @theashraf.
I tested Slot override & reset. In ThorVG on my machine, I did not encounter segmentation fault error.

I don't find any relevance to segmentation fault from dotlottie-rs.

We also noticed a bug in Slot resetting in this file. After resetting the slot, the styles do not render to their original color(blue), it's rendering a black star, which is not the original look.

Perhaps, some sort of bugs affecting the color could cause segmentation fault in dotlottie-rs, not sure yet. I'll keep track on the resetting bug.

**[Original]**
![CleanShot 2024-05-09 at 15 27 04](https://github.com/thorvg/thorvg/assets/11167117/a301f143-a633-4ec1-8ce6-5f8dbf028445)

**[Slot Overrien]**
![CleanShot 2024-05-09 at 15 34 05](https://github.com/thorvg/thorvg/assets/11167117/0bc92986-a038-4b68-bc70-a281e29c5d96)

**[After resetting]**
![CleanShot 2024-05-09 at 15 30 14](https://github.com/thorvg/thorvg/assets/11167117/817baf59-bf64-48f4-b710-0fd7d90d0541)

hermet:
@tinyjin did you check with sanitizer?

tinyjin:
@hermet Ooops.
I've got a heap UAF with sanitizer. I'll take care of it. This might be potential crash in Rust system.

[...]

PR #2272, opened 2024-05-13
lottie/slot: Fix slot resetting bug

tinyjin:
When resetting back to animated property, system causes an UAF because frames have been freed.

Mark frames in nullptr at the case, so it doesn't use frame data after freed.

Issue: #2255 

CleanShot 2024-05-13 at 17 24 14@2x

**Current, when back to animated property.**

CleanShot 2024-05-13 at 17 29 35@2x

**Seems resetting well now.**

CleanShot 2024-05-13 at 17 25 48


hermet (approved):

hermet:
@tinyjin It would be nicer if you could add one more unit-test for the slot-resetting case in tests/lottie ...

mgrudzinska (approved):
I'd add issue number into commit msg

PR #2275, opened 2024-05-13
test: fix margin value

mgrudzinska:
The precision margin for the comparison
has been set to 0.01f instead of 004004.


test/capi/capiAnimation.cpp
```diff
@@ -81,15 +81,15 @@ TEST_CASE("Animation Lottie", "[capiAnimation]")
 
     float frame;
     REQUIRE(tvg_animation_get_total_frame(animation, &frame) == TVG_RESULT_SUCCESS);
-    REQUIRE(frame == Approx(120).margin(004004));
+    REQUIRE(frame == Approx(120).margin(0.01f));
```
hermet, line 84: By the api guide, `0.001f` is correct range?

hermet (approved):

tinyjin (approved):

Issue #2274, opened 2024-05-13
renderer: support an viewport interface of the canvas


hermet:
Support a feature to set the partial region of the drawing. This would help apps minimize the ThorVG drawing region when only a portion of the canvas is visible on the screen, such as during scrolling.

```
canvas->viewport(x, y, w, h);
```

PR #2276, opened 2024-05-14
renderer/engines: support the canvas viewport function.

hermet:
The viewport function defines the rectangular area of the canvas that will be used for drawing operations.

It is used to clip the rendering output to the boundaries of the rectangle. Apps can use this function to set the drawing region within the canvas.

When the ThorVG canvas is partially inside the screen area such as during scrolling it could help enhance rendering performance.

issue: https://github.com/thorvg/thorvg/issues/2274


src/renderer/sw_engine/tvgSwRenderer.cpp
```diff
@@ -409,6 +409,7 @@ RenderRegion SwRenderer::viewport()
 bool SwRenderer::viewport(const RenderRegion& vp)
 {
     vport = vp;
+    if (surface) vport.intersect({0, 0, (int32_t)surface->w, (int32_t)surface->h});
```
JSUYA, line 412: If it is the same as before, can we skip it? Called every time the canvas is updated or drawn.
Of course, if the comparison check is an unnecessary operation rather than performing intersect() every time, you can ignore this comment.

test/testSwCanvas.cpp
```diff
@@ -94,3 +94,251 @@ TEST_CASE("Memory Pool", "[tvgSwCanvas]")
...
+
+    auto canvas = SwCanvas::gen();
+    REQUIRE(canvas);
+
+    REQUIRE(canvas->viewport(25, 25, 100, 100) == Result::Success);
+
+    uint32_t buffer[100*100];
+    REQUIRE(canvas->target(buffer, 100, 100, 100, SwCanvas::Colorspace::ARGB8888) == Result::Success);
+
+    REQUIRE(canvas->viewport(25, 25, 50, 50) == Result::Success);
```
JSUYA, line 327: Does this code affect the actual rendering results?
For the results below, the final vport result values are 0, 0, 100, 00, and I wanted the actual output to be displayed in the upper left corner.
The output result is at the bottom right.
If the viewport has changed, shouldn't the update be performed internally? 
Because that's how push() works.

```diff
diff --git a/examples/Svg.cpp b/examples/Svg.cpp
index 259fd01f..b9b8ed79 100644
--- a/examples/Svg.cpp
+++ b/examples/Svg.cpp
@@ -120,7 +120,11 @@ void tvgSwTest(uint32_t* buffer)

 void drawSwView(void* data, Eo* obj)
 {
+    swCanvas->viewport(WIDTH - 100, HEIGHT - 100, 100, 100);
+    swCanvas->update();
+    swCanvas->viewport(0, 0, 100, 100);
     if (swCanvas->draw() == tvg::Result::Success) {
+    swCanvas->viewport(WIDTH - 100, 0, 100, 100);
         swCanvas->sync();
     }
 }

```

JSUYA: oh... I mean. It was a call between update() and draw(). 
My explain about `REQUIRE(canvas->viewport(25, 25, 50, 50) == Result::Success);` is wrong.

To summarize again..., `viewport()` called after calling `update()`(or `push()`) is not actually applied and returns Success.
Since this means that our API is affected by the order, my opinion was that, like the push() API, the update() function should be called again internally when the viewport is updated.
I think, if someone who does not know this, sets the viewport between update() and draw()(or after update())and performs a transition, An unintended area may be displayed in the first frame.

This is my test code
```diff
diff --git a/examples/Svg.cpp b/examples/Svg.cpp
index 259fd01f..f792030a 100644
--- a/examples/Svg.cpp
+++ b/examples/Svg.cpp
@@ -120,7 +120,11 @@ void tvgSwTest(uint32_t* buffer)

 void drawSwView(void* data, Eo* obj)
 {
+    printf("1) viewport(924 924 100 100) => %d\n", static_cast<int>(swCanvas->viewport(WIDTH - 100, HEIGHT - 100, 100, 100)));
+    swCanvas->update();
+    printf("2) viewport(0, 0, 100, 100) => %d\n", static_cast<int>(swCanvas->viewport(0, 0, 100, 100)));
     if (swCanvas->draw() == tvg::Result::Success) {
+    printf("3) viewport(WIDTH - 100, 0, 100, 100) => %d\n", static_cast<int>(swCanvas->viewport(WIDTH - 100, 0, 100, 100)));
         swCanvas->sync();
     }
 }
diff --git a/src/renderer/tvgCanvas.h b/src/renderer/tvgCanvas.h
index 7e2b7c3f..1a81431d 100644
--- a/src/renderer/tvgCanvas.h
+++ b/src/renderer/tvgCanvas.h
@@ -98,6 +98,7 @@ struct Canvas::Impl
         auto flag = RenderUpdateFlag::None;
         if (refresh || force) flag = RenderUpdateFlag::All;

+printf("in update function call viewport(vport) : %d %d %d %d\n", vport.x, vport.y, vport.w, vport.h);
         renderer->viewport(vport);

         if (paint) {
@@ -113,6 +114,7 @@ struct Canvas::Impl

     Result draw()
     {
+printf("in draw function call viewport(vport) : %d %d %d %d\n", vport.x, vport.y, vport.w, vport.h);
         renderer->viewport(vport);

         if (drawing || paints.empty() || !renderer->preRender()) return Result::InsufficientCondition;
```

result
```
$ ./build/examples/Svg
...
in update function call viewport(vport) : 0 0 2147483647 2147483647
1) viewport(924 924 100 100) => 0 // < - Success
in update function call viewport(vport) : 924 924 100 100
2) viewport(0, 0, 100, 100) => 0  // <- Success
in draw function call viewport(vport) : 0 0 100 100
3) viewport(WIDTH - 100, 0, 100, 100) => 2 // <- InsufficientCondition
```
The actual results are rendered in 924 924 100 100.

src/renderer/tvgCanvas.h
```diff
@@ -104,12 +108,13 @@ struct Canvas::Impl
             }
             refresh = false;
         }
+        status = Status::Updating;
```
JSUYA, line 111: update() is also called from push(). Therefore, viewport() only operates when the canvas is synchronized(or clean?). after creation and sync(). If this is the intended behavior, I think it would be a good to describe it to the docs.

JSUYA (approved):

PR #2277, opened 2024-05-14
test/lottie: added the slot resetting in animated property case

tinyjin:
Enhance Lottie test coverage in case animated property used


test/testLottie.cpp
```diff
@@ -70,6 +70,33 @@ TEST_CASE("Lottie Slot", "[tvgLottie]")
     REQUIRE(Initializer::term() == Result::Success);
 }
 
+TEST_CASE("Lottie Slot(Keyframe)", "[tvgLottie]")
```
hermet, line 73: I think it's hard to identify the 'Keyframe' in the test suite itself. Let's add it as "Lottie Slot2".

hermet (approved):

PR #2279, opened 2024-05-14
infra/gl_enigne: fallback use gles v2 if the system doesn't have v3.

hermet:
let's keep this for a while until our minimum requirement is clear.

### DIFF
--- a/src/renderer/gl_engine/meson.build
+++ b/src/renderer/gl_engine/meson.build
@@ -21,12 +21,14 @@ source_file = [
    'tvgGlTessellator.h',
 ]
 
-gles_dep = meson.get_compiler('cpp').find_library('GLESv3')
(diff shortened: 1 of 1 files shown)


RuiwenTang (approved):

RuiwenTang:
BTW,should we consider a dynamic loading mechanism similar to [GLAD](https://glad.dav1d.de/)? This can avoids compile time lookup and linking library.

hermet:
> BTW,should we consider a dynamic loading mechanism similar to [GLAD](https://glad.dav1d.de/)? This can avoids compile time lookup and linking library.

It would be ideal as a portable library; however, dynamic linking would be costly for us, and it's still experimental. Let's see. By the way, I think `GlCanvas::gen()` should return `nullptr` during creation when GL is not available on the current system.

PR #2280, opened 2024-05-14
lottie: fix repeater transformation

mgrudzinska:
Matrix multiplication order changed.

before:
Zrzut ekranu 2024-05-15 o 00 45 13

after:
Zrzut ekranu 2024-05-15 o 00 43 30

AE:
Zrzut ekranu 2024-05-15 o 00 44 49

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -40,6 +40,7 @@
 struct RenderRepeater
 {
     int cnt;
+    uint32_t level;
     float offset;
     Point position;
     Point anchor;
@@ -59,7 +60,7 @@ struct RenderContext
     Shape* propagator = nullptr;
     Shape* merging = nullptr;  //merging shapes if possible (if shapes have same properties)
     LottieObject** begin = nullptr; //iteration entry point
-    RenderRepeater* repeater = nullptr;
+    Array<RenderRepeater> repeaters;
     Matrix* transform = nullptr;
     float roundness = 0.0f;
     bool fragmenting = false;  //render context has been fragmented by filling
@@ -74,7 +75,6 @@ struct RenderContext
     ~RenderContext()
     {
         if (ownPropagator) delete(propagator);
-        delete(repeater);
         free(transform);
     }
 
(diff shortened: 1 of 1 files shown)


hermet (changes requested):
@mgrudzinska Hello, it would be nice if you share the test resource here. 
Unfortunately, this breaks the existing resources: in examples/lottie:
`5317-fireworkds.json`, `abstract_circle.json`
Need to confirm all works correctly.

mgrudzinska:
It is a working solution, but I'm not entirely satisfied with the use of array repeaters. 
From what I can see, repeaters, unlike other elements, do not overwrite the previous ones - instead, both are applied. This would justify using an array, but I would prefer to confirm this and discuss it further.

mgrudzinska:
[repeater.json](https://github.com/thorvg/thorvg/files/15375452/repeater.json)

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -59,7 +60,7 @@ struct RenderContext
     Shape* propagator = nullptr;
     Shape* merging = nullptr;  //merging shapes if possible (if shapes have same properties)
     LottieObject** begin = nullptr; //iteration entry point
-    RenderRepeater* repeater = nullptr;
+    Array<RenderRepeater*> repeaters;
```
hermet, line 63:
```
Array<RenderRepeater> repeaters;
```

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -49,6 +49,7 @@ struct RenderRepeater
     uint8_t endOpacity;
     bool interpOpacity;
     bool inorder;
+    bool external;
```
hermet, line 52: We strongly recommend adding a comment for this kind of ambiguous data.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1082,7 +1096,7 @@ static void _updateChildren(LottieGroup* parent, float frameNo, Inlist<RenderCon
                     break;
                 }
                 case LottieObject::Repeater: {
-                    _updateRepeater(parent, child, frameNo, contexts, ctx, exps);
+                    _updateRepeater(parent, child, frameNo, contexts, ctx, exps, groupLevel == level);
```
hermet, line 1099: Here, the level looks over-programmed. You might wanna just confirm whether the repeater is requested after the group.

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: I'll change it

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: I changed the approach, since I've found nested examples. I need to know whether the level of repeater was greater or it was equal to the level of repeated shape

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -88,16 +88,17 @@ struct RenderContext
             propagator = static_cast<Shape*>(rhs.propagator->duplicate());
         }
 
-        if (rhs.repeater) {
-            repeater = new RenderRepeater();
-            *repeater = *rhs.repeater;
+        if (!rhs.repeaters.empty()) {
```
hermet, line 91:
```
if (!rhs.repeaters.empty()) {
  ...
} 
```
-> 

```
repeaters.push(rhs.repeaters);
```

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -40,6 +40,7 @@
 struct RenderRepeater
 {
     int cnt;
+    uint32_t level;
```
hermet, line 43: What do you think if we put this `uint16_t level` in the `RenderContext`.  I guess the code could be more compact & clean.

mgrudzinska:
combined with https://github.com/thorvg/thorvg/pull/2311

PR #2281, opened 2024-05-14
lottie: implement skew transformation

mgrudzinska:
Added support for skew transformation by
a specified angle around a specified axis.

after:
Zrzut ekranu 2024-05-15 o 01 31 25

expected:
Zrzut ekranu 2024-05-15 o 01 36 56

[skewed_stars.json](https://github.com/thorvg/thorvg/files/15314637/skewed_stars.json)

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -129,6 +129,42 @@ static void _rotationZ(Matrix* m, float degree)
 }
 
 
+static void _skew(Matrix* m, float angleDeg, float axisDeg)
+{
+    auto angle = -mathDeg2Rad(angleDeg);
+    float tanVal = tanf(angle);
+
+    axisDeg = fmod(axisDeg, 180.0f);
+    if (fabsf(axisDeg) < 0.01f || fabsf(axisDeg - 180.0f) < 0.01f || fabsf(axisDeg + 180.0f) < 0.01f) {
+        float cosVal = cosf(mathDeg2Rad(axisDeg));
+        auto B = cosVal * cosVal * tanVal;
+        m->e12 += B * m->e11;
+        m->e22 += B * m->e21;
+        return;
+    } else if (fabsf(axisDeg - 90.0f) < 0.01f || fabsf(axisDeg + 90.0f) < 0.01f) {
+        float sinVal = -sinf(mathDeg2Rad(axisDeg));
+        auto C = sinVal * sinVal * tanVal;
+        m->e11 -= C * m->e12;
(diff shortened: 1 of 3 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -154,6 +191,16 @@ static bool _updateTransform(LottieTransform* transform, float frameNo, bool aut
         _rotateX(&matrix, transform->rotationEx->x(frameNo, exps));
     }
 
+    auto skewAngle = transform->skewAngle(frameNo, exps);
```
hermet, line 194: In most cases, `skewAngle` and `skewAxis` will have 0 values (default). We can probably skip these defaults quickly.

mgrudzinska: yes, this check (and more) are done inside ```_skew```:
```if (mathZero(fmod(angleDeg, 360.0f))) return;```
I can remove fmod...360 if it's wasting of time

Comment (hermet) [reply src/loaders/lottie/tvgLottieBuilder.cpp:194]: maybe something like this
```
if (skewAngle != 0.0f) {
   //apply the skew transform
    auto skewAxis = transform->skewAxis(frameNo, exps);
    _skew(&matrix, skewAngle, skewAxis);
}
```

File: src/loaders/lottie/tvgLottieBuilder.cpp [hunk 2]
```diff
@@ -154,6 +191,16 @@ static bool _updateTransform(LottieTransform* transform, float frameNo, bool aut
...
+    // For angles where tangent explodes, the shape degenerates into an infinitely thin line.
+    // This is handled by zeroing out the matrix due to finite numerical precision.
+    if (fabsf(fmod(skewAngle + 90.0f, 180.0f)) < 0.01f) {
```
hermet, line 197: if we don't expect any visual result by this, maybe, we can quickly return false and skip forwarding.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -154,6 +190,16 @@ static bool _updateTransform(LottieTransform* transform, float frameNo, bool aut
         _rotateX(&matrix, transform->rotationEx->x(frameNo, exps));
     }
 
+    auto skewAngle = transform->skewAngle(frameNo, exps);
+    if (fabsf(skewAngle) > 0.01f) {
```
hermet, line 194: If the exceptions are handled in 198, Please consider the most general case.
```
if (skewAngle == 0.0f) 
```
is better obviously.

hermet (approved):

PR #2285, opened 2024-05-16
examples: porting examples to gtk+2.0 on linux

onecoolx:
Hi all，
      I'm trying to porting examples to gtk2 on linux. I had tried porting a version to support both linux+gtk2 and tizen systems. Most of the demos looks like normal, but I'm not sure whether this code is right on tizen / ecore system. I can't verify it. Can anyone try it?
      
Issue: https://github.com/thorvg/thorvg/issues/1244


hermet:
@onecoolx Hello, may I ask what prompted you to switch to GTK? I assume there was a specific need for you.

onecoolx:
> @onecoolx Hello, may I ask what prompted you to switch to GTK? I assume there was a specific need for you.

I prefer developing cross-platform graphics applications. I found that the code related to the local window system on each platform can be abstracted into an encapsulation layer, and then the upper code is isolated from the window system. As you see, the patch I submitted actually supports two UI system at the same time. EvasPlatform.cpp retains the original EFL support, and Gtk2Platform.cpp implements the porting of gtk2.0.

onecoolx:
> @onecoolx Hello, may I ask what prompted you to switch to GTK? I assume there was a specific need for you.

Another reason is that EFL opengl doesn't work well on my ubuntu : )

hermet:
> > @onecoolx Hello, may I ask what prompted you to switch to GTK? I assume there was a specific need for you.
> 
> Another reason is that EFL opengl doesn't work well on my ubuntu : )

Firstly, thank you for your contribution. I assure you that EFL works perfectly with OpenGL on Ubuntu. I believe it's just a simple matter of environment setup. Nonetheless, the framework is an additional usage in ThorVG just for a showcase. My concern is that multiple choices and dependencies will significantly increase the maintenance burden for the ThorVG project. Therefore, in my opinion, ThorVG should depend on one framework at most for the showcase. We need to consider which framework is the easiest and most adaptable among user systems. Perhaps we might be able to replace EFL with GTK but more than it, unlike GTK, Qt, EFL, and others, I prefer the framework having minimal dependencies and being lightweight, so it's quite portable. Again, thanks for your contribution and let's see.

onecoolx:
> > > @onecoolx Hello, may I ask what prompted you to switch to GTK? I assume there was a specific need for you.
> > 
> > 
> > Another reason is that EFL opengl doesn't work well on my ubuntu : )
> 
> Firstly, thank you for your contribution. I assure you that EFL works perfectly with OpenGL on Ubuntu. I believe it's just a simple matter of environment setup. Nonetheless, the framework is an additional usage in ThorVG just for a showcase. My concern is that multiple choices and dependencies will significantly increase the maintenance burden for the ThorVG project. Therefore, in my opinion, ThorVG should depend on one framework at most for the showcase. We need to consider which framework is the easiest and most adaptable among user systems. Perhaps we might be able to replace EFL with GTK but more than it, unlike GTK, Qt, EFL, and others, I prefer the framework having minimal dependencies and being lightweight, so it's quite portable. Again, thanks for your contribution and let's see.

Yes, I agree. But in my opinion, Gtk2 is a better choice than EFL even if it only supports one GUI backend. gtk supports most Linux systems you can find, and supports MacOSX and Windows.

vtorri:
EFL works on mac and windows

onecoolx:
> I prefer the framework having minimal dependencies and being lightweight,

Hmm，SDL ? Maybe better?

hermet:
@onecoolx Hello, thank you for your contribution. However, as discussed [here](https://github.com/thorvg/thorvg/pull/2285), I'm considering using GLFW because of the less size and dependency. If you'd like to suggest more benefits of GTK over GLFW, please feel free to share them before confirming it. Thanks.

onecoolx:
> @onecoolx Hello, thank you for your contribution. However, as discussed [here](https://github.com/thorvg/thorvg/pull/2285), I'm considering using GLFW because of the less size and dependency. If you'd like to suggest more benefits of GTK over GLFW, please feel free to share them before confirming it. Thanks.

I agree with others that GLFW is a better choice for ThorVG.

Issue #2273, opened 2024-05-13
Support for embedded images that are not linked for PNG, JPG or WebP in SVG


fire:
Appears to need extra support implemented for embedded images

https://github.com/godotengine/godot/pull/91901#issuecomment-2107298904

mgrudzinska:
@fire could you please provide any sample?

fire:
Try the one from https://github.com/lovell/sharp/issues/3557

fire:
Expected:

![image](https://github.com/thorvg/thorvg/assets/32321/95103e87-49c4-4313-beb9-14896b8378e1)

Result:

![image](https://github.com/thorvg/thorvg/assets/32321/e5984024-8863-4431-b199-9dfaeef5a6ed)

akien-mga:
A couple more examples from testing https://github.com/godotengine/godot/pull/91901:

![drawing](https://github.com/thorvg/thorvg/assets/4701338/38e4e371-e99f-4fd8-ba8f-dac09941ab7c)
![drawing2](https://github.com/thorvg/thorvg/assets/4701338/60f000b8-eb57-4da1-b82b-8599fb60ffc7)

Those SVGs were generated with Inkscape 1.3.2 on Linux by simply using the "Import" (Ctrl+I) option, selecting a WEBP, PNG, or JPG file, and choosing "Embed" as import type:
![image](https://github.com/thorvg/thorvg/assets/4701338/a15e5739-02e0-42de-b654-a8f31a8c3fed)

mgrudzinska:
> A couple more examples from testing [godotengine/godot#91901](https://github.com/godotengine/godot/pull/91901):
> 

these I'll fix tmr.

[this one ](https://github.com/thorvg/thorvg/issues/2273#issuecomment-2115211312) is more complicated. set an image as svg background, scales it, applies "image-rendering:pixelated"

hermet:
background-image:url in style is not supported:
![download](https://github.com/thorvg/thorvg/assets/3711518/a302037d-35a3-4f97-bfa4-ce4ba7c3051a)

hermet:
We can provide an asset resolver.

PR #2290, opened 2024-05-17
common: fix data parsing

mgrudzinska:
The appearance of xml entities was handled
by the svg parser, but they may also be present
in the embedded data. Thus, while decoding base64, these elements must be removed.

@Issue: https://github.com/thorvg/thorvg/issues/2273

before:
Zrzut ekranu 2024-05-17 o 12 54 50

after:
Zrzut ekranu 2024-05-17 o 12 54 18

### DIFF
--- a/src/loaders/svg/tvgXmlParser.h
+++ b/src/loaders/svg/tvgXmlParser.h
@@ -25,9 +25,9 @@
 
 #include "tvgSvgLoaderCommon.h"
 
-#define NUMBER_OF_XML_ENTITIES 8
-const char* const xmlEntity[] = {"&quot;", "&nbsp;", "&apos;", "&amp;", "&lt;", "&gt;", "&#035;", "&#039;"};
-const int xmlEntityLength[] = {6, 6, 6, 5, 4, 4, 6, 6};
+#define NUMBER_OF_XML_ENTITIES 9
+const char* const xmlEntity[] = {"&#10;", "&quot;", "&nbsp;", "&apos;", "&amp;", "&lt;", "&gt;", "&#035;", "&#039;"};
+const int xmlEntityLength[] = {5, 6, 6, 6, 5, 4, 4, 6, 6};
 
 enum class SimpleXMLType
 {


hermet:
@mgrudzinska Are the issued embedded image types 'webp, png, jpg'? I'm a bit confused as to why XML is necessary for them.

mgrudzinska:
one file has an embedded png file, the other one jpeg. they both contain ```&#10;```
Zrzut ekranu 2024-05-17 o 18 27 55


[here](https://github.com/thorvg/thorvg/issues/2273#issuecomment-2115248898) is described how they were produced.

years ago we had an svg file that contained some xml entities in some values (not in the embedded data).

mgrudzinska:
@hermet oo you think we could duplicate the b64Decode() for the SVG loader and the Lottie loader? 
I'm not sure if skipping xml entities makes sense for lottie files, whereas we do have examples of svgs where they apper.

hermet:
@mgrudzinska I think we can use customizer.
```
typedef char* (*CustomDecoder)(const char* start, const char* end);      //implement this in svg loader.

b64Decode(.........,  CustomDecoder* customizer = nullptr)
{
   ...
   if (customizer) encoded = customizer(encoded, encoded + len);
   ...
}
```

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -570,7 +572,7 @@ static unique_ptr<Picture> _imageBuildHelper(SvgLoaderData& loaderData, SvgNode*
         if (!_isValidImageMimeTypeAndEncoding(&href, &mimetype, &encoding)) return nullptr; //not allowed mime type or encoding
         char *decoded = nullptr;
         if (encoding == imageMimeTypeEncoding::base64) {
-            auto size = b64Decode(href, strlen(href), &decoded);
+            auto size = b64Decode(href, strlen(href), &decoded, &b64DecodeSkipXML);
```
hermet, line 575: I think only `_skipXmlEntities()` is necessary for custom decoding. Should it have an entirely duplicated decoding logic in svgLoader?

hermet (commented):
Something went wrong. Please check the comment. I think only the necessary part can be customized.

mgrudzinska:
@hermet sorry for not noticing earlier, but this small change is enough...

hermet:
@mgrudzinska Great. confirmed the fix above but this looks an another story:

download

mgrudzinska:
> @mgrudzinska Great. confirmed the fix above and this looks an another story:
> 

yes, this one is more complicated, as I wrote [here](https://github.com/thorvg/thorvg/issues/2273#issuecomment-2116411315), we have to set an image as the svg background, scales it (both doable) and apply "image-rendering:pixelated". the last one is something new.

but since these two examples were included in the issue, II've mentioned the issue in the commit msg

PR #2300, opened 2024-05-20
web: ++performance 

hermet:


tinyjin:
This patch improves to little more FPS in my machine as well. (around 3~5 FPS++)

But I'm wondering, this patch consumes much more memories.
In screenshot, as you can see, patched version consumes 20MB and original is 10MB. I have no idea what reason causes it.

However, I'm sure this can improve performance.
Because, In JS, `new Array` could be faster in large calculation, there are plenty benchmarks shows that's faster than `Array.from`. (Theoretically, it should consume lesser memory) 

Thus, approved. I believe this much of memory increasement is not a big deal than performance enhancement.
In thousand animations, memory consumption is almost same. I think only affected in a few animations.

https://www.measurethat.net/Benchmarks/ShowResult/524234
CleanShot 2024-05-21 at 15 27 39@2x


**[Current]**
CleanShot 2024-05-21 at 15 22 27@2x

**[Patched]**
CleanShot 2024-05-21 at 15 22 43@2x

tinyjin (approved):

web/src/lottie-player.ts
```diff
@@ -360,7 +360,7 @@ export class LottiePlayer extends LitElement {
     }
 
     const buffer = this._TVG.render();
-    const clampedBuffer = Uint8ClampedArray.from(buffer);
+    const clampedBuffer = new Uint8ClampedArray(buffer.buffer, buffer.byteOffset, buffer.byteLength);
```
hermet, line 363: @theashraf could you please review this in dotlottie side? This might help dotlottie player to improve the fps (~5fps)

theashraf: Awesome @hermet

theashraf: @hermet in dotlottie-web, we're currently not casting the framebuffer, which should bring better performance since the ThorVG framebuffer is guaranteed to be a Uint8 Array anyway.

https://github.com/LottieFiles/dotlottie-web/blob/991635ac4fefedfe98cae9f3dd69d66273323d89/packages/web/src/dotlottie.ts#L457

hermet: @theashraf thanks, we will double check it. 👍

PR #2310, opened 2024-05-23
common: code refactoring

hermet:
Replace the math functions with operator overloading. This should potentially reduce the code size.


src/common/tvgMath.h
```diff
@@ -36,12 +36,12 @@
...
 
-
 bool mathInverse(const Matrix* m, Matrix* out);
-Matrix mathMultiply(const Matrix* lhs, const Matrix* rhs);
-void mathRotate(Matrix* m, float degree);
 bool mathIdentity(const Matrix* m);
-void mathMultiply(Point* pt, const Matrix* transform);
+void operator*=(Point& pt, const Matrix& m);
+Matrix operator*(const Matrix& lhs, const Matrix& rhs);
+void operator*=(Matrix& m, float degree);
```
mgrudzinska, line 43: maybe could we add some comment here? imo this is counterintuitive that matrix * float is a rotation...

mgrudzinska (approved):

src/common/tvgMath.h
hermet: let's revert it.

PR #2311, opened 2024-05-23
lottie: repeater transformation fix

mgrudzinska:
In the case of a transformation (denoted as T1)
appearing after the repeater, but at the same
level of nesting, the objects to which the repeater is applied should not consider T1. However,
the transformation of the repeater itself should
take T1 into account.

example 20874.json from lottie zip files:

before:
Zrzut ekranu 2024-05-23 o 15 41 24

after:
Zrzut ekranu 2024-05-23 o 15 41 05

Zrzut ekranu 2024-05-23 o 14 29 37

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -40,6 +40,7 @@
 struct RenderRepeater
 {
     int cnt;
+    Matrix transform;
     float offset;
     Point position;
     Point anchor;
@@ -59,7 +60,7 @@ struct RenderContext
     Shape* propagator = nullptr;
     Shape* merging = nullptr;  //merging shapes if possible (if shapes have same properties)
     LottieObject** begin = nullptr; //iteration entry point
-    RenderRepeater* repeater = nullptr;
+    Array<RenderRepeater> repeaters;
     Matrix* transform = nullptr;
     float roundness = 0.0f;
     bool fragmenting = false;  //render context has been fragmented by filling
@@ -74,7 +75,6 @@ struct RenderContext
     ~RenderContext()
     {
         if (ownPropagator) delete(propagator);
-        delete(repeater);
         free(transform);
     }
 
@@ -88,9 +88,8 @@ struct RenderContext
             propagator = static_cast<Shape*>(rhs.propagator->duplicate());
         }
 
-        if (rhs.repeater) {
-            repeater = new RenderRepeater();
-            *repeater = *rhs.repeater;
+        for (auto repeater = rhs.repeaters.begin(); repeater < rhs.repeaters.end(); ++repeater) {
+            repeaters.push(*repeater);
         }
         roundness = rhs.roundness;
     }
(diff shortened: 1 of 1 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -413,6 +440,16 @@ static void _repeat(LottieGroup* parent, unique_ptr<Shape> path, RenderContext*
         mathTranslateR(&m, -repeater->anchor.x, -repeater->anchor.y);
 
         auto pm = PP(shape)->transform();
+
+        if (repeater->transform && !mathIdentity(static_cast<const Matrix*>(repeater->transform))) {
+            m = mathMultiply(repeater->transform, &m);
+            if (pm) {
+                Matrix inverse;
+                mathInverse(repeater->transform, &inverse);
+                *pm = mathMultiply(&inverse, pm);
```
mgrudzinska, line 449: I know it's a bit expensive approach, but I don;t expect this scenario to happen very often

mgrudzinska:
other sample checking different configurations of transformation:

[rep1.json](https://github.com/thorvg/thorvg/files/15421030/rep1.json)
[rep2.json](https://github.com/thorvg/thorvg/files/15421031/rep2.json)
[rep3.json](https://github.com/thorvg/thorvg/files/15421032/rep3.json)
[rep4.json](https://github.com/thorvg/thorvg/files/15421033/rep4.json)

before:
before_rep

after:
after_rep

expected:
|Zrzut ekranu 2024-05-22 o 23 28 48|Zrzut ekranu 2024-05-22 o 23 28 47|
|Zrzut ekranu 2024-05-22 o 23 28 44|Zrzut ekranu 2024-05-22 o 23 28 50|

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -49,6 +50,34 @@ struct RenderRepeater
     uint8_t endOpacity;
     bool interpOpacity;
     bool inorder;
+
+    RenderRepeater() : transform(nullptr) {}
```
hermet, line 54: it does already in 43 line.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -413,6 +440,16 @@ static void _repeat(LottieGroup* parent, unique_ptr<Shape> path, RenderContext*
         mathTranslateR(&m, -repeater->anchor.x, -repeater->anchor.y);
 
         auto pm = PP(shape)->transform();
+
+        if (repeater->transform && !mathIdentity(static_cast<const Matrix*>(repeater->transform))) {
```
hermet, line 444: You might wanna check the mathIdentity == true case. I assume its much less than false scenarios. If so, it hurts the bigger case, it shouldn't be.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -983,6 +1020,11 @@ static void _updateRepeater(TVG_UNUSED LottieGroup* parent, LottieObject** child
     auto repeater= static_cast<LottieRepeater*>(*child);
 
     if (!ctx->repeater) ctx->repeater = new RenderRepeater();
+    auto matrix = ctx->propagator->transform();
+    if (!mathIdentity(static_cast<const Matrix*>(&matrix))) {
```
hermet, line 1024: Please confirm if this identity check is really helpful. I'm concerned it just adds unnecessary checks.

hermet (commented):
Thanks please check my comments.

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: I wanted to avoid unnecessary mem allocation, but now should be more optimal:
changed to null check:
```
    if (auto matrix = PP(ctx->propagator)->transform()) {
        if (!ctx->repeater->transform) ctx->repeater->transform = (Matrix*)malloc(sizeof(Matrix));
        *ctx->repeater->transform = *matrix;
    } else ctx->repeater->transform = nullptr;
    ```

File: src/loaders/lottie/tvgLottieBuilder.cpp (hunk 2 above)
Comment (mgrudzinska) [reply src/loaders/lottie/tvgLottieBuilder.cpp:54]: done

File: src/loaders/lottie/tvgLottieBuilder.cpp [hunk 5]
```diff
@@ -49,6 +50,34 @@ struct RenderRepeater
     uint8_t endOpacity;
     bool interpOpacity;
     bool inorder;
+
+    RenderRepeater() : transform(nullptr) {}
+
+    RenderRepeater(const RenderRepeater& r)
```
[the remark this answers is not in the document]
mgrudzinska: done

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -49,6 +50,34 @@ struct RenderRepeater
...
+              interpOpacity(r.interpOpacity),
+              inorder(r.inorder)
+    {
+        if (r.transform) {
+            transform = static_cast<Matrix*>(malloc(sizeof(Matrix)));
+            if (transform) *transform = *r.transform;
+        }
+    }
+
+    RenderRepeater& operator=(RenderRepeater r) = delete;
```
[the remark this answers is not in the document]
mgrudzinska: done

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: The issue is that at a given level, we already have an accumulated transformation matrix, and we're accumulating that before the repeater appears. I'll think about it some more.

hermet: Tip: simply print out each case tested with Lottie and see which case prints more.

src/loaders/lottie/tvgLottieBuilder.cpp
hermet: Maybe `Matrix transform` in Repeater (43 line) might be better than `Matrix* transform` if, in most cases, the repeater needs to allocate the transform."

mgrudzinska:
combined with https://github.com/thorvg/thorvg/pull/2280

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: not sure any more if I understand - yes, often this won't be an identity matrix. so I removed the check. is that what you meant?

hermet: Checking the identity matrix also comes with a comparison cost. If the identity matrix is used more frequently compared to the non-identity matrix, we can keep it; otherwise, I'd suggest removing it. I think you can' briefly estimate this with `Lottie` example.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -392,45 +391,56 @@ static Shape* _draw(LottieGroup* parent, RenderContext* ctx)
...
-        mathRotate(&m, repeater->rotation * multiplier);
-        mathTranslateR(&m, -repeater->anchor.x, -repeater->anchor.y);
-
-        auto pm = PP(shape)->transform();
-        shape->transform(pm ? (m * *pm) : m);
+    Array<Shape*> propagators;
+    propagators.push(ctx->propagator);
+
+    for (auto repeater = ctx->repeaters.end() - 1; repeater >= ctx->repeaters.begin(); --repeater) {
+        Array<Shape*> shapes(repeater->cnt);
```
hermet, line 398: Suggestion: Reuse this array by putting it outside the loop.
```
Array<Shape*> shapes;
for (... ) {
   shapes.reserve(repeater->cnt);       //reserve the size
   
    ...
   
   shapes.reset();       //re-use the memory pool.
}
```

hermet (approved):

PR #2316, opened 2024-05-24
capi: fix compile warning

XuNeo:
Fix warning during build with lvgl https://github.com/lvgl/lvgl/pull/6081/

```
error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
```


hermet:
@XuNeo could you please add yourself here? https://github.com/thorvg/thorvg/blob/main/AUTHORS

hermet (approved):

Issue #2287, opened 2024-05-16
SwRender is stuck with the specific Lottie file


beicause:
The Lottie file:
[test10.json](https://github.com/thorvg/thorvg/files/15331298/test10.json)
Add this test case in ThorVG 0.13.3:
```cpp
TEST_CASE("Animation Lottie10", "[tvgAnimation]")
{
    REQUIRE(Initializer::init(CanvasEngine::Sw, 0) == Result::Success);

    auto canvas = SwCanvas::gen();
    REQUIRE(canvas);
    uint32_t buffer[512*512];
    REQUIRE(canvas->target(buffer, 512, 512, 512, SwCanvas::Colorspace::ABGR8888S) == Result::Success);

    auto animation = Animation::gen();
    REQUIRE(animation);

    auto picture = animation->picture();
    REQUIRE(picture->identifier() == Picture::identifier());

    REQUIRE(canvas->push(cast(picture)) == Result::Success);

    REQUIRE(picture->load(TEST_DIR"/test10.json") == Result::Success);
    REQUIRE(animation->totalFrame() == Approx(180).margin(004004));
    REQUIRE(animation->curFrame() == 0);
    REQUIRE(animation->duration() == Approx(3).margin(004004));

    for(int i=1;i<180;i++){
        REQUIRE(animation->frame(i) == Result::Success);
        REQUIRE(canvas->update(picture) == Result::Success);
        REQUIRE(canvas->draw() == Result::Success);
        REQUIRE(canvas->sync() == Result::Success);
        REQUIRE(canvas->clear(false) == Result::Success);
    }
    REQUIRE(Initializer::term(CanvasEngine::Sw) == Result::Success);
}
```

Build with:
```sh
meson setup builddir -Dtests=true -Dextra=""
ninja -C builddir
```

Gets stuck sometimes when running ( but seem not stuck with `-Dextra="lottie_expressions"`):
```sh 
./builddir/test/tvgUnitTests
```

Gets stuck always when running:
```sh
./builddir/test/tvgUnitTests "Animation Lottie10"
```

I don't known the exactly reason and the Lottie file is downloaded from website, I am not the author. The call stack shows it's stuck in the while loop in tvgSwRle.cpp  `_cubicTo`.

`lottie2gif` and thorvg-viewer works fine, but Export GIF in thorvg-viewer alerts the error `Unable to save the Gif data.`.

I open this separate issue from https://github.com/godotengine/godot/pull/91580#pullrequestreview-2051394914

[...]

PR #2318, opened 2024-05-24
lottie: fix interpolation issue

mgrudzinska:
Interpolation between closed and open curves
is allowed. Since these curves have a different
number of points, this caused an error. The current behavior enforces both curves to be treated as opened.

@Issue: https://github.com/thorvg/thorvg/issues/2287

[sample](https://github.com/thorvg/thorvg/files/15440592/interp.json)

tvg:
tvg

web:
web

### DIFF
--- a/src/loaders/lottie/tvgLottieProperty.h
+++ b/src/loaders/lottie/tvgLottieProperty.h
@@ -548,7 +548,13 @@ struct LottiePathSet : LottieProperty
                 t = (frameNo - frame->no) / ((frame + 1)->no - frame->no);
                 if (frame->interpolator) t = frame->interpolator->progress(t);
                 if (frame->hold) path = &(frame + ((t < 1.0f) ? 0 : 1))->value;
-                else interpolate = true;
+                else {
+                    if (frame->value.ptsCnt != (frame + 1)->value.ptsCnt) {
+                        path = &frame->value;
+                        TVGLOG("LOTTIE", "Different numbers of points in consecutive frames - interpolation omitted.");
+                    }
+                    else interpolate = true;
+                }
             }
         }
 


mgrudzinska:
The [sample](https://github.com/thorvg/thorvg/files/15440592/interp.json) I created by myself, it has some errors so it can not be uploaded to AE... similar issue with the [test10.json](https://github.com/thorvg/thorvg/files/15440622/test10.json) from the [issue](https://github.com/thorvg/thorvg/issues/2287) - it can be uploaded but with many errors... so hard to see how the interpolation between closed/opened is done there

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -560,24 +560,26 @@ struct LottiePathSet : LottieProperty
 
         auto s = frame->value.pts;
         auto e = (frame + 1)->value.pts;
+        //possible interpolation between closed and opened curves
+        auto minFrame = frame->value.ptsCnt <= (frame + 1)->value.ptsCnt ? frame : frame + 1;
```
hermet, line 564: Logically, this sounds like unexpected behavior because it doesn't make sense to interpolate paths that don't have the same number of points. What do you think about just returning the current path without interpolation if the number of points doesn't match?

hermet (commented):
@mgrudzinska thanks, please check a comment.

src/loaders/lottie/tvgLottieProperty.h
mgrudzinska: If we do as you say, the sample json I provided will be static. I would rather keep interpolation by cutting off closing points. This change from open to closed seems suspicious, but not the need for interpolation itself. 

However, if you think it's better not to interpolate, I'll change it. 

It seems to me that overall the visual effect will be minimal and these won't be a common case

hermet: For now, we can fix the issue with a rather simple solution I suggested. The problem is that this not only cuts off the closing but also cuts when the paths size don't match in general. Logically, This scenario couldn't be expected and we don't need to complicate things for that. Let's add a LOG now and revisit this if it really matters later by users.

hermet:
@mgrudzinska great, your fix has been applied.

PR #2323, opened 2024-05-26
fix(capi): add void as paramter instead of ()

kisvegabor:
Hi,

Empty argument list generates warning with some compilers.


kisvegabor:
Duplicate of https://github.com/thorvg/thorvg/pull/2316

PR #2327, opened 2024-05-28
lottie optimization

hermet:


src/common/tvgCompressor.cpp
```diff
@@ -472,4 +472,19 @@ size_t b64Decode(const char* encoded, const size_t len, char** decoded)
...
+unsigned long djb2Encode(const char* str)
+{
+    unsigned long hash = 5381;
+    int c;
+
+    while ((c = *str++)) {
+        hash = ((hash << 5) + hash) + c; // hash * 33 + c
+    }
+    return hash;
+}
```
tinyjin, line 488: We can reduce code line?
```cpp
unsigned long hash = 5381;

while (int c = *str++) {
    hash = ((hash << 5) + hash) + c; // hash * 33 + c
}
return hash;
```

hermet: Probably, However, rather than just reducing the number of lines of code, we can focus on simplifying the code by reducing its logical complexity.

tinyjin: Yes agree, That should be priority!
LGTM so far :)

tinyjin (approved):

src/common/tvgCompressor.cpp
hermet: See: https://jellyfish.co/library/code-complexity/

I've seen many cases where thorvg contributors don't pay much attention to the  'degree of nesting', 

It's not good.

tinyjin: Thanks, hermet. I'll keep in mind!

PR #2330, opened 2024-05-29
lottie: fix & enhance parsing

mgrudzinska:
The object type key does not need to be provided
first. Previously, content was skipped until the
"ty" type key was found. If the key was placed at
the end, the entire object was ignored.
Now, before parsing subsequent elements, finding
the "ty" key is enforced. Since in-situ parsing
does not support backtracking, the key is searched 
with an internal function.

Introducing support for the "ty" key placed anywhere 
highlighted the use of redundant logic for parsing 
the list of shapes — this duplicate logic has been removed.

before:
Zrzut ekranu 2024-05-30 o 01 42 45

after:
Zrzut ekranu 2024-05-30 o 01 27 31

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -370,7 +370,7 @@ template<typename T>
 bool LottieParser::parseTangent(const char *key, LottieVectorFrame<T>& value)
 {
     if (KEY_AS("ti") && getValue(value.inTangent)) ;
-    else if (KEY_AS("to") && getValue(value.outTangent)) ;       
+    else if (KEY_AS("to") && getValue(value.outTangent)) ;
     else return false;
 
     value.hasTangent = true;
@@ -881,10 +881,12 @@ LottieOffsetPath* LottieParser::parseOffsetPath()
 }
 
 
-LottieObject* LottieParser::parseObject()
+LottieObject* LottieParser::parseObject(const char* type)
 {
-    auto type = getString();
-    if (!type) return nullptr;
+    if (!type) {
+        type = getString();
+        if (!type) return nullptr;
+    }
 
     if (!strcmp(type, "gr")) return parseGroup();
     else if (!strcmp(type, "rc")) return parseRect();
@@ -911,6 +913,23 @@ LottieObject* LottieParser::parseObject()
 void LottieParser::parseObject(Array<LottieObject*>& parent)
 {
     enterObject();
+
+    //object type key is not listed as the first one
+    auto value = peekValue();
+    if (value && strcmp(value->GetString(), "ty")) {
+        if (auto type = findObjectType()) {
+            if (auto child = parseObject(type)) {
+                if (child->hidden) delete (child);
+                else parent.push(child);
+            } else {
+                //skip unsupported type
+                while (auto key = nextObjectKey()) skip(key);
+            }
+            free(type);
+            return;
+        }
+    }
+    //object type key either listed as the first one or never - skip the entire object
     while (auto key = nextObjectKey()) {
         if (KEY_AS("ty")) {
             if (auto child = parseObject()) {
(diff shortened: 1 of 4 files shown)


mgrudzinska:
file to test: [matte_bugs.lottie.json](https://github.com/thorvg/thorvg/files/15491840/matte_bugs.lottie.json) - from this issue: https://github.com/LottieFiles/dotlottie-web/issues/214
the other file may be private (?)

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -857,9 +861,9 @@ LottieRepeater* LottieParser::parseRepeater()
 }
 
 
-LottieObject* LottieParser::parseObject()
+LottieObject* LottieParser::parseObject(const char* type)
 {
-    auto type = getString();
+    if (!type) type = getString();
     if (!type) return nullptr;
```
tinyjin, line 867: In case `type` isn't nullptr from the beginning, we might be able to avoid unnecessary condition check by this code

```cpp
if (!type) {
    type = getString();
    if (!type) return nullptr;
}
```

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -887,6 +891,22 @@ LottieObject* LottieParser::parseObject()
 void LottieParser::parseObject(Array<LottieObject*>& parent)
 {
     enterObject();
+
+    //object type key is not listed as the first one
+    if (peekValue() && strcmp(peekValue()->GetString(), "ty")) {
```
tinyjin, line 896: It could be more optimal and readable?
```cpp
auto value = peekValue()
if (value && strcmp(value->GetString(), "ty"))
```

hermet:
@mgrudzinska Could you please separate the `d` and `ty` patches? If the order of the `ty` patch is changed by other tools, it should be reviewed carefully.

mgrudzinska:
> @mgrudzinska Could you please separate the `d` and `ty` patches? If the order of the `ty` is changed by other tools/plugins (against the bodymovin), it should be reviewed carefully.

I separated them https://github.com/thorvg/thorvg/pull/2333
will rebase this one after 2333 is merged

hermet:
Seems this patch should be confirm & pushed first than https://github.com/thorvg/thorvg/pull/2333

mgrudzinska:
> Seems this patch should be confirm & pushed first than #2333

but what with the 'direction' ? In this pr I removed the whole section of parseShapes. I won't handle direction in group because it doesn;t make sense. I moved direction handling to separate shapes parsing, but you wanted this in a separate pr.

hermet:
Please cancel the draft after resolving the conflict. Thanks.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -546,6 +546,7 @@ LottieRect* LottieParser::parseRect()
 LottieEllipse* LottieParser::parseEllipse()
 {
     auto ellipse = new LottieEllipse;
+    if (!ellipse) return nullptr;
```
hermet, line 549: @mgrudzinska just for your information, I intentionally removed the null check from the memory alloc fails.

mgrudzinska: @hemet it makes perfect sense, since new will throw ex, not return null

mgrudzinska: I added this, sorry, rebasing didn;t go well

hermet:
As you might know, the new usage could affect the performance of the Lottie format quite negatively. I think this is really bad structure for Lottie. Let's review this carefully.

mgrudzinska:
[skottie](https://skottie.skia.org/?h=128&w=128) rearranges the order of attributes and puts all "ty" tags at the end

hermet:
@mgrudzinska could you please rebase this?

hermet:
@mgrudzinska, could you please resubmit this PR from the ThorVG main repo?

mgrudzinska:
> @mgrudzinska, could you please resubmit this PR from the ThorVG main repo?

@hermet https://github.com/thorvg/thorvg/tree/mira/format rebased to main and on the main repo. 
if you prefer I can close this pr and open a new one, I don't see a way how to change a source branch of an already opened pr

mgrudzinska:
the same changes, just different branch: https://github.com/thorvg/thorvg/pull/3194

PR #2333, opened 2024-05-30
lottie: support 'direction' in ellipses

mgrudzinska:
lottie: support 'direction' in ellipses

partially solves 16699.json (without https://github.com/thorvg/thorvg/pull/2330 won't work since direction property is given before the type property))

before:
Zrzut ekranu 2024-07-9 o 15 53 57

after:
Zrzut ekranu 2024-07-9 o 15 49 31

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -553,7 +553,7 @@ static void _updateRect(LottieGroup* parent, LottieObject** child, float frameNo
 }
 
 
-static void _appendCircle(Shape* shape, float cx, float cy, float rx, float ry, Matrix* transform)
+static void _appendCircle(Shape* shape, float cx, float cy, float rx, float ry, Matrix* transform, bool clockwise)
 {
     auto rxKappa = rx * PATH_KAPPA;
     auto ryKappa = ry * PATH_KAPPA;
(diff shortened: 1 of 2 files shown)


mgrudzinska:
[org.json](https://github.com/thorvg/thorvg/files/15499040/org.json)
[rev.json](https://github.com/thorvg/thorvg/files/15499041/rev.json)

main:
Zrzut ekranu 2024-05-30 o 13 23 48

this pr:
Zrzut ekranu 2024-05-30 o 13 22 58

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -527,7 +527,7 @@ static void _updateRect(LottieGroup* parent, LottieObject** child, float frameNo
 }
 
 
-static void _appendCircle(Shape* shape, float cx, float cy, float rx, float ry, Matrix* transform)
+static void _appendCircle(Shape* shape, float cx, float cy, float rx, float ry, float direction, Matrix* transform)
```
hermet, line 530: Direction will be useful only when the trimpath is valid. What do you think about the builder setting the trimpath in a reverse way (1.0, 0) ?

mgrudzinska: is this what you mean? now I don't need ifs
  ```
    auto i = direction > 0 ? 0 : ptsCnt - 1;
    points[i] = {cx, cy - ry}; //moveTo
    points[i += direction] = {cx + rxKappa, cy - ry}; points[i += direction] = {cx + rx, cy - ryKappa}; points[i += direction] = {cx + rx, cy}; //cubicTo
    ...
  ```

also I'd change the def:
```
    uint8_t direction = 0;   //0: clockwise, 2: counter-clockwise, 3: xor(?)
```
to be 
```
    int8_t direction = 1;   //1: clockwise, -1: counter-clockwise, 3: xor(?)
```
and now we could replace:
```
 auto direction = (ellipse->direction == 0) ? 1.0f : -1.0f;
 ```
by:
 ```
  auto direction = (float)ellipse->direction;
  ```
 but I suspect that there was something behind 0/2/3 values, so I didn;t touch it

hermet: > auto i = direction > 0 ? 0 : ptsCnt - 1;

No, we might be able to make the direction just using the trimpath feature. 
- (0 ~ 1) //CW.  
- (0 ~ -1) //CCW.

mgrudzinska: But trimming is more expensive, isn't it? For now, it's implemented as dashing, but even if we change this, some checks will have to be made inside the renderer. We could handle 0:1/0:-1 separately, but it may still be tricky to handle cases involving both direction and trimming. Even now, when there are more than one consecutive trim paths with offset in the json, it doesn't work well.

but if you think we should use trimming for direction handling, I can try to implement it, but this seems to be some other task than the one solved in this pr. maybe we could do this as a separate issue?

hermet: > But trimming is more expensive, isn't it?

Considering the direction, trimming is necessary regardless. Direction without trimming is pointless, IMO.

> We could handle 0:1/0:-1 separately, but it may still be tricky to handle cases involving both direction and trimming.

Let's figure it out.

> Even now, when there are more than one consecutive trim paths with offset in the json, it doesn't work well.

I think this is another story we need to handle.

> if you think we should use trimming for direction handling, I can try to implement it,

Let's figure out a better approach with logic first. That's why we are discussing this.

> but this seems to be some other task than the one solved in this pr. maybe we could do this as a separate issue?

direction is unlikely to be used comparatively and is not urgent to push IMO. So.. probably yes but unless a user reports the issue.  In the end, the *LottiePath* needs to cover this as well. Also implementing direction at the shape level would increase the size and its complexity. Let's double-check the better approach first and have a decision. Maybe `strokeTrim()` could additionally allow the direction for common usage.

mgrudzinska: @hermet since here are two commits and the second one is the problematic, I moved the first into https://github.com/thorvg/thorvg/pull/2343, because I'd need this to rebase https://github.com/thorvg/thorvg/pull/2330

mgrudzinska: so I think we can close this one, fix trimming and then apply direction in a proper way. ok

hermet: Nah, Let's keep this for now and lets see.

hermet:
Please rebase the patch and cancel the draft. Thanks.

mgrudzinska:
@hermet can we merge it ?

hermet (approved):

hermet:
@mgrudzinska merged but can't see any fixes mentioned you above:

image

Issue #2190, opened 2024-04-16
Renderer: Add stroke trim API


hermet:
Stroke trimming is useful for artistic visuals including Lottie Animation. 
ThorVG already has this feature built-in; it could provide the APIs for general usage:

```
Shape::strokeTrim(float begin = 0.0f, float end = 1.0f)
```

hermet:
see this: https://github.com/thorvg/thorvg/pull/2333

hermet:
Final version:
```
Result strokeTrim(float begin, float end, bool simultaneous = true) noexcept;
bool strokeTrim(float* begin, float* end) const noexcept;
```

PR #2337, opened 2024-05-30
common: strokeTrim api

mgrudzinska:
New api sets/gets the trim of the stroke
along the defined path segment, allowing
control over which part of the stroke is
visible.

@Issue: https://github.com/thorvg/thorvg/issues/2190

Zrzut ekranu 2024-05-30 o 21 05 22

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1021,7 +1021,6 @@ public:
      */
     Result strokeJoin(StrokeJoin join) noexcept;
 
-
     /**
      * @brief Sets the stroke miterlimit.
      *
@@ -1032,6 +1031,22 @@ public:
      * @since 0.11
      */
     Result strokeMiterlimit(float miterlimit) noexcept;
+
+    /**
+     * @brief Sets the trim of the stroke along the defined path segment, allowing control over which part of the stroke is visible.
+     *
+     * The values of the arguments @p begin, @p end, and @p offset are in the range of 0.0 to 1.0, representing the beginning of the path and the end, respectively.
+     *
+     * @param[in] begin Specifies the start of the segment to display along the path.
+     * @param[in] end Specifies the end of the segment to display along the path.
+     * @param[in] simultaneous Determines how to trim multiple paths within a single shape. If set to @c true (default), trimming is applied simultaneously to all paths;
+     * Otherwise, all paths are treated as a single entity with a combined length equal to the sum of their individual lengths and are trimmed as such.
+     *
+     * @retval Result::Success when succeed.
+     *
+     * @note Experimental API
+     */
+    Result strokeTrim(float begin, float end, bool simultaneous = true) noexcept;
 
     /**
      * @brief Sets the solid color for all of the figures from the path.
@@ -1072,7 +1087,6 @@ public:
      */
     Result fill(FillRule r) noexcept;
 
-
     /**
      * @brief Sets the rendering order of the stroke and the fill.
      *
@@ -1083,7 +1097,6 @@ public:
      * @since 0.10
      */
     Result order(bool strokeFirst) noexcept;
-
 
     /**
      * @brief Gets the commands data of the path.
@@ -1189,6 +1202,18 @@ public:
      * @since 0.11
      */
     float strokeMiterlimit() const noexcept;
+
+    /**
+     * @brief Gets the trim of the stroke along the defined path segment.
+     *
+     * @param[out] begin The starting point of the segment to display along the path.
+     * @param[out] end Specifies the end of the segment to display along the path.
+     *
+     * @retval @c true if trimming is applied simultaneously to all paths of the shape, @c false otherwise.
+     *
+     * @note Experimental API
+     */
+    bool strokeTrim(float* begin, float* end) const noexcept;
 
     /**
      * @brief Creates a new Shape object.
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -916,7 +916,7 @@ static void _updateTrimpath(TVG_UNUSED LottieGroup* parent, LottieObject** child
         end = (length * end) + pbegin;
     }
 
-    P(ctx->propagator)->strokeTrim(begin, end, trimpath->type == LottieTrimpath::Type::Individual ? true : false);
+    P(ctx->propagator)->strokeTrim(begin, end, trimpath->type == LottieTrimpath::Type::Simultaneous);
 }
 
 
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -121,11 +121,11 @@ LottieImage::~LottieImage()
 
 void LottieTrimpath::segment(float frameNo, float& start, float& end, LottieExpressions* exps)
 {
-    auto s = this->start(frameNo, exps) * 0.01f;
-    auto e = this->end(frameNo, exps) * 0.01f;
+    start = this->start(frameNo, exps) * 0.01f;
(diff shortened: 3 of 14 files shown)


src/renderer/tvgRender.h
```diff
@@ -140,7 +140,7 @@ struct RenderStroke
     struct {
         float begin = 0.0f;
         float end = 1.0f;
-        bool individual = false;
+        bool simultaneous = true;
```
mgrudzinska, line 143: @hermet I changed 'individual' to 'simultaneous' because imo the naming is very unfortunate. 'Individual' means that all paths within a given shape are treated as a whole, rather than individually, which is counterintuitive. 
It's easier to explain to the user that 'simultaneous' means that the trimming is done for all paths simultaneously if set to true, if false it is otherwise
but ofc I can restore the previous version if you don;t agree

hermet:
Could you please replace the internal usage with this api in `_updateTrimpath()` in LottieBuilder? if it reduces the binary size. (I guess it does)

mgrudzinska:
> Could you please replace the internal usage with this api in `_updateTrimpath()` in LottieBuilder? if it reduces the binary size. (I guess it does)

I don't see any difference in size

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -916,7 +916,7 @@ static void _updateTrimpath(TVG_UNUSED LottieGroup* parent, LottieObject** child
         end = (length * end) + pbegin;
     }
 
-    P(ctx->propagator)->strokeTrim(begin, end, trimpath->type == LottieTrimpath::Type::Individual ? true : false);
+    ctx->propagator->strokeTrim(begin, end, trimpath->type == LottieTrimpath::Type::Simultaneous);
```
hermet, line 919: Please compare size with [text+data+bss+dec],  changing is not good...

before:
image
after:
image

mgrudzinska: @hermet am I doing something wrong? def settings, no changes
Zrzut ekranu 2024-05-31 o 12 01 29

hermet: Please try $strip the binary before $size

mgrudzinska: I got the whole list of symbols that can not be stripped, no changes in the size. I'll try to find out later why it doesn;t work, for now I restored the org code.

src/renderer/tvgShape.h
```diff
@@ -216,23 +216,35 @@ struct Shape::Impl
...
-        rs.stroke->trim.individual = individual;
+        rs.stroke->trim.simultaneous = simultaneous;
         flag |= RenderUpdateFlag::Stroke;
 
         return true;
     }
 
+    bool strokeTrim(float* begin, float* end, bool* simultaneous)
+    {
+        if (!rs.stroke) return false;
```
hermet, line 239:
```
if (!rs.stroke) {
   if (begin) *begin = 0.0f;
   if (end) *end = 1.0f;
}
```

src/renderer/tvgShape.h
```diff
@@ -216,23 +216,35 @@ struct Shape::Impl
...
+
+        if (begin) *begin = rs.stroke->trim.begin;
+        if (end) *end = rs.stroke->trim.end;
+        if (simultaneous) *simultaneous = rs.stroke->trim.simultaneous;
+
+        return true;
```
hermet, line 245:
```
return  rs.stroke->trim.simultaneous;
```

inc/thorvg.h
```diff
@@ -1190,6 +1203,19 @@ class TVG_API Shape final : public Paint
...
+     *
+     * @param[out] begin The starting point of the segment to display along the path.
+     * @param[out] end Specifies the end of the segment to display along the path.
+     * @param[out] simultaneous Determines how to trim multiple paths within a single shape.
+     *
+     * @retval Result::Success when succeed, Result::InsufficientCondition otherwise.
+     *
+     * @note Experimental API
+     */
+    Result strokeTrim(float* begin, float* end, bool* simultaneous) const noexcept;
```
hermet, line 1217: Return type is useless practically. let's see this suggestion. we can make apis simpler.
```
// return true when simultaneous, false otherwise.
bool strokeTrim(float* begin, float* end) const noexcept;
```

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -137,28 +137,8 @@ void LottieTrimpath::segment(float frameNo, float& start, float& end, LottieExpr
         return;
     }
 
-    s += o;
```
mgrudzinska, line 140: @hermet please note that this logic in moved into tvgShape.h

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -137,28 +137,8 @@ void LottieTrimpath::segment(float frameNo, float& start, float& end, LottieExpr
...
-    e += o;
-
-    auto loop = true;
```
mgrudzinska, line 143: @hermet please note that this logic in moved into tvgShape.h

src/renderer/tvgShape.h
```diff
@@ -216,23 +216,51 @@ struct Shape::Impl
...
         return true;
     }
 
+    bool strokeTrim(float* begin, float* end)
+    {
+        if (rs.stroke) {
+            if (begin) *begin = rs.stroke->trim.begin;
+            if (end) *end = rs.stroke->trim.end;
+            return rs.stroke->trim.simultaneous;
+        }
```
hermet, line 260:
```
else {
   if (begin) *begin = 0.0f;
   if (end) *end = 1.0f;
}
```

inc/thorvg.h
```diff
@@ -1033,6 +1032,22 @@ class TVG_API Shape final : public Paint
...
+     * @brief Sets the trim of the stroke along the defined path segment, allowing control over which part of the stroke is visible.
+     *
+     * The values of the arguments @p begin, @p end, and @p offset are in the range of 0.0 to 1.0, representing the beginning of the path and the end, respectively.
+     *
+     * @param[in] begin Specifies the start of the segment to display along the path.
+     * @param[in] end Specifies the end of the segment to display along the path.
+     * @param[in] simultaneous Determines how to trim multiple paths within a single shape. If set to @c true (default), trimming is applied simultaneously to all paths;
+     * Otherwise, all paths are treated as a single entity with a combined length equal to the sum of their individual lengths and are trimmed as such.
+     *
+     * @retval Result::Success when succeed, Result::FailedAllocation otherwise.
```
hermet, line 1045: No exceptions so far?
```
     * @retval Result::Success when succeed
```

inc/thorvg.h
```diff
@@ -1190,6 +1203,19 @@ class TVG_API Shape final : public Paint
      */
     float strokeMiterlimit() const noexcept;
 
+    /**
+     * @brief Gets the trim of the stroke along the defined path segment.
+     *
+     * @param[out] begin The starting point of the segment to display along the path.
+     * @param[out] end Specifies the end of the segment to display along the path.
+     * @param[out] simultaneous Determines how to trim multiple paths within a single shape.
```
hermet, line 1211: --

src/renderer/tvgShape.h
```diff
@@ -216,23 +216,54 @@ struct Shape::Impl
         return true;
     }
 
-    bool strokeTrim(float begin, float end, bool individual)
+    bool strokeTrim(float begin, float end, bool simultaneous)
```
hermet, line 219: void return?

mgrudzinska:
rebased

hermet (approved):

mgrudzinska:
rebased

PR #2339, opened 2024-05-31
wg_engine: fix color buffer corruption with wgpu-opengl wrapper (linux)

SergeyLebedkin:
In a case of usage stencil buffer only we need to turn off an color target writes. In other case color buffer fill be filled by unxepcted color if fragment shader did not return any value. It happens in a case on OpenGL realization of webgpu, that used in linux

Befire:
image

After:
image

### DIFF
--- a/src/renderer/wg_engine/tvgWgRenderData.cpp
+++ b/src/renderer/wg_engine/tvgWgRenderData.cpp
@@ -353,8 +353,8 @@ void WgRenderDataShape::releaseMeshes(WgContext &context)
     meshGroupStrokes.release(context);
     meshGroupShapesBBox.release(context);
     meshGroupShapes.release(context);
-    pMin = {0.0f, 0.0f };
-    pMax = {0.0f, 0.0f };
+    pMin = {FLT_MAX, FLT_MAX};
+    pMax = {0.0f, 0.0f};
 }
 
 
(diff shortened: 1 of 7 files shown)


src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -321,7 +321,7 @@ bool WgRenderer::target(void* window, uint32_t w, uint32_t h)
     mTargetSurface.w = w > 0 ? w : 1;
     mTargetSurface.h = h > 0 ? h : 1;
 
-    // TODO: replace solution to cross-platform realization
+        // TODO: replace solution to cross-platform realization
```
hermet, line 324: --

src/renderer/wg_engine/tvgWgRenderData.cpp
```diff
@@ -353,8 +353,8 @@ void WgRenderDataShape::releaseMeshes(WgContext &context)
     meshGroupStrokes.release(context);
     meshGroupShapesBBox.release(context);
     meshGroupShapes.release(context);
-    pMin = {0.0f, 0.0f };
-    pMax = {0.0f, 0.0f };
+    pMin = { FLT_MAX, FLT_MAX };
```
hermet, line 356:
```
    pMin = {FLT_MAX, FLT_MAX};
    pMax = {0.0f, 0.0f};
```

hermet (approved):

PR #2246, opened 2024-05-05
text tag

mgrudzinska:
todo:
- default values handling
- font-family multi values
- fix grad?
- meson - better idea?

Zrzut ekranu 2024-05-6 o 00 43 31

```
<svg viewBox="0 0 125 115">
  <defs>
    <linearGradient id="rainbow" x1="0" x2="0" y1="20%" y2="100%" gradientUnits="userSpaceOnUse" >
      <stop stop-color="#ff00ff" offset="0%"/>
      <stop stop-color="#FF5447" offset="30%"/>
      <stop stop-color="#FF7B21" offset="40%"/>
      <stop stop-color="#EAFC37" offset="60%"/>
      <stop stop-color="#4FCB6B" offset="80%"/>
      <stop stop-color="#0000ff" offset="100%"/> 
    </linearGradient>
  </defs>
    <text font-family="Monaco" fill="#ff00ff" font-size="25" x="0" y="30">gradient</text>
    <text font-family="Monaco" fill="url(#rainbow)" font-size="45" x="0" y="70">text</text>
    <text font-family="Monaco" font-size="20" x="30" y="100">in svg</text>
    <rect x="120" y="0" width="5" height="115" fill="url(#rainbow)"/>
</svg>
```

issue: https://github.com/thorvg/thorvg/issues/1897

Issue #2200, opened 2024-04-18
Unable to open file with Chinese path


zengzm1212:
such as:  std::string path = u8"F:/svg_test/中文路径/中文evenodd.svg";

hermet:
no test resource.

Issue #2055, opened 2024-03-14
Lottie: Support feasibility of accessing a specific Lottie node by identifier


hermet:
Every Lottie layer/property is assigned a name (ID), allowing users to access specific nodes to perform hit testing for user interactions.

ThorVG could assign an ID property to every paint node; however, we may need to refine the process of matching each paint to the corresponding Lottie model node and property.

See also: https://github.com/thorvg/thorvg/issues/1276

Example code user usage:
```
auto accessor = tvg::Accessor::gen();

//The callback function from lambda expression.
//This function will be called for every paint nodes of the picture tree.
auto search = [](const tvg::Paint* paint) -> bool
{
    if (paint->name() == "that i'm looking for") {
        //acquire the geometry info of the "that i'm looking for"
        float x, y, w, h;
        paint->bounds(&x, &y, &w, &h, true);
        
        //stop traversing immediately.
        return false
    }

     //You can return false, to stop traversing immediately.
     return true;
};

auto& paints = canvas->paints();
for (auto paint : paints) {
   paint->set(tvg::cast(paint), search );
}
```

PR #2345, opened 2024-06-02
ttf_loader: support loading from memory

mgrudzinska:
Adding support for reading from memory is
necessary to enable embedded ttf support,
e.g. in svg files.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1552,7 +1552,7 @@ public:
     Result fill(std::unique_ptr<Fill> f) noexcept;
 
     /**
-     * @brief Loads a scalable font data(ttf) from a file.
+     * @brief Loads a scalable font data (ttf) from a file.
      *
      * ThorVG efficiently caches the loaded data using the specified @p path as a key.
      * This means that loading the same file again will not result in duplicate operations;
@@ -1570,6 +1570,34 @@ public:
      * @see Text::unload(const std::string& path)
      */
     static Result load(const std::string& path) noexcept;
+
+    /**
+     * @brief Loads a scalable font data (ttf) from a memory block of a given size.
+     *
+     * ThorVG efficiently caches the loaded font data using the specified @p name as a key.
+     * This means that loading the same fonts again will not result in duplicate operations.
+     * Instead, ThorVG will reuse the previously loaded font data.
+     *
+     * @param[in] name The name under which the font will be stored and accessible (e.x. in a @p font() API).
+     * @param[in] data A pointer to a memory location where the content of the font data is stored.
+     * @param[in] size The size in bytes of the memory occupied by the @p data.
+     * @param[in] mimeType Mimetype or extension of font data. In case an empty string is provided the loader will be determined automatically.
+     * @param[in] copy If @c true the data are copied into the engine local buffer, otherwise they are not (default).
+     *
+     * @retval Result::Success When succeed.
+     * @retval Result::InvalidArguments If no name is provided or if @p size is zero while @p data points to a valid memory location.
+     * @retval Result::NonSupport When trying to load a file with an unsupported extension.
+     * @retval Result::Unknown If an error occurs at a later stage.
+     *
+     * @warning: It's the user responsibility to release the @p data memory.
+     *
+     * @note To unload the font data loaded using this API, pass the proper @p name and @c nullptr as @p data.
+     * @note If you are unsure about the MIME type, you can provide an empty value like @c "", and thorvg will attempt to figure it out.
+     * @note Experimental API
+     *
+     * @see Text::font(const char* name, float size, const char* style)
+     */
+    static Result load(const char* name, const char* data, uint32_t size, const std::string& mimeType = "ttf", bool copy = false) noexcept;
 
     /**
      * @brief Unloads the specified scalable font data (TTF) that was previously loaded.
--- a/src/loaders/ttf/tvgTtfLoader.cpp
+++ b/src/loaders/ttf/tvgTtfLoader.cpp
@@ -193,7 +193,13 @@ static uint32_t* _codepoints(const char* text, size_t n)
 
 void TtfLoader::clear()
 {
-    _unmap(this);
+    if (nomap) {
+        if (freeData) free(reader.data);
+        reader.data = nullptr;
+        reader.size = 0;
+        freeData = false;
+        nomap = false;
+    } else _unmap(this);
     shape = nullptr;
 }
 
@@ -232,6 +238,22 @@ bool TtfLoader::open(const string& path)
 {
     clear();
     if (!_map(this, path)) return false;
+    return reader.header();
+}
+
+
+bool TtfLoader::open(const char* data, uint32_t size, TVG_UNUSED const string& rpath, bool copy)
+{
+    reader.size = size;
+    nomap = true;
+
+    if (copy) {
+        reader.data = (uint8_t*)malloc(size);
+        if (!reader.data) return false;
+        memcpy((char*)reader.data, data, reader.size);
+        freeData = true;
+    } else reader.data = (uint8_t*)data;
+
     return reader.header();
 }
 
--- a/src/loaders/ttf/tvgTtfLoader.h
+++ b/src/loaders/ttf/tvgTtfLoader.h
@@ -37,12 +37,15 @@ struct TtfLoader : public FontLoader
     char* text = nullptr;
     Shape* shape = nullptr;
     bool italic = false;
+    bool nomap = false;
+    bool freeData = false;
 
     TtfLoader();
     ~TtfLoader();
 
     using FontLoader::open;
     bool open(const string& path) override;
+    bool open(const char *data, uint32_t size, const string& rpath, bool copy) override;
     bool resize(Paint* paint, float w, float h) override;
     bool request(Shape* shape, char* text, bool italic = false) override;
     bool read() override;
--- a/src/renderer/tvgText.cpp
+++ b/src/renderer/tvgText.cpp
@@ -71,6 +71,21 @@ Result Text::load(const std::string& path) noexcept
 }
 
 
+Result Text::load(const char* name, const char* data, uint32_t size, const string& mimeType, bool copy) noexcept
+{
+    if (!name || (size == 0 && data)) return Result::InvalidArguments;
+
+    //unload font
+    if (!data) {
+        if (LoaderMgr::retrieve(name)) return Result::Success;
+        return Result::InvalidArguments;
+    }
+
+    if (!LoaderMgr::loader(name, data, size, mimeType, copy)) return Result::NonSupport;
+    return Result::Success;
+}
+
+
 Result Text::unload(const std::string& path) noexcept
 {
     if (LoaderMgr::retrieve(path)) return Result::Success;
--- a/examples/Text.cpp
+++ b/examples/Text.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "Common.h"
+#include <fstream>
 
 /************************************************************************/
 /* Drawing Commands                                                     */
@@ -50,10 +51,20 @@ void tvgDrawCmds(tvg::Canvas* canvas)
         return;
     }
 
-    if (tvg::Text::load(EXAMPLE_DIR"/font/SentyCloud.ttf") != tvg::Result::Success) {
-        cout << "TTF is not supported. Did you enable TTF Loader?" << endl;
-        return;
+    //Load from memory
+    ifstream file(EXAMPLE_DIR"/font/SentyCloud.ttf", ios::binary);
+    if (!file.is_open()) return;
+    file.seekg(0, std::ios::end);
+    auto size = file.tellg();
+    file.seekg(0, std::ios::beg);
+    auto data = (char*)malloc(size);
+    if (!data) return;
+    file.read(data, size);
+    file.close();
+    if (tvg::Text::load("SentyCloud", data, size, "ttf", true) != tvg::Result::Success) {
+        cout << "Error while loading TTF from memory. Did you enable TTF Loader?" << endl;
     }
+    free(data);
 
     auto text = tvg::Text::gen();
     text->font("Arial", 80);
--- a/src/renderer/tvgLoader.cpp
+++ b/src/renderer/tvgLoader.cpp
@@ -433,3 +433,26 @@ LoadModule* LoaderMgr::loader(const uint32_t *data, uint32_t w, uint32_t h, bool
     delete(loader);
     return nullptr;
 }
+
+
+//loads fonts from memory - loader is cached (regardless of copy value) in order to access it while setting font
+LoadModule* LoaderMgr::loader(const char* name, const char* data, uint32_t size, const string& mimeType, bool copy)
+{
+    //TODO: add check for mimetype ?
+    if (auto loader = _findFromCache(name)) return loader;
+
+    if (auto loader = _findByType(mimeType)) {
+        if (loader->open(data, size, "", copy)) {
+            loader->hashpath = strdup(name);
+            loader->pathcache = true;
+            ScopedLock lock(key);
+            _activeLoaders.back(loader);
+            return loader;
+        } else {
+            TVGLOG("LOADER", "The font data \"%s\" could not be loaded.", name);
+            delete(loader);
+        }
+    }
+
+    return nullptr;
+}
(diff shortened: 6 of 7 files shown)


mgrudzinska:
@hermet I need this for https://github.com/thorvg/thorvg/issues/1897

hermet:
> @hermet I need this for #1897

@mgrudzinska Hello, next time, please have a look at some practices of how we can add a new feature for ThorVG. By sharing a plan and open discuss before writing code, we can try to discuss with ThorVG maintainers, contributors, feature requesters to reach a correct consensus with aligning, before pushing a new feature. Let's save our coding time and review process.

https://github.com/thorvg/thorvg/issues/692
https://github.com/thorvg/thorvg/issues/2044
https://github.com/thorvg/thorvg/issues/969
https://github.com/thorvg/thorvg/issues/1640
https://github.com/thorvg/thorvg/issues/2055
https://github.com/thorvg/thorvg/issues/2190
https://github.com/thorvg/thorvg/issues/1808
https://github.com/thorvg/thorvg/issues/1372

mgrudzinska:
> @mgrudzinska Hello, next time, please have a look at some practices of how we can add a new feature for ThorVG. By sharing a plan and open discuss before writing code, we can try to discuss with ThorVG maintainers, contributors, feature requesters to reach a correct consensus with aligning, before pushing a new feature. Let's save our coding time and review process.

after your [comment](https://github.com/thorvg/thorvg/pull/2246#issuecomment-2128325590) about embedded fonts support I didn't think there is much to discuss...

hermet:
> > @mgrudzinska Hello, next time, please have a look at some practices of how we can add a new feature for ThorVG. By sharing a plan and open discuss before writing code, we can try to discuss with ThorVG maintainers, contributors, feature requesters to reach a correct consensus with aligning, before pushing a new feature. Let's save our coding time and review process.
> 
> after your [comment](https://github.com/thorvg/thorvg/pull/2246#issuecomment-2128325590) about embedded fonts support I didn't think there is much to discuss...

Yes, discussion is not necessary for you and me, but it might be necessary for others.

hermet:
@mgrudzinska \+APIs always matter. That's why we need to reach a consensus for new API designs.

inc/thorvg.h
```diff
@@ -1544,16 +1544,42 @@ class TVG_API Text final : public Paint
...
+     * @retval Result::Unknown If an error occurs at a later stage.
+     *
+     * @warning: It's the user responsibility to release the @p data memory.
+     *
+     * @note Experimental API
+     *
+     * @see Text::unload(const std::string& fontId)
+     */
+    static Result load(const std::string& name, const char* data, uint32_t size, const std::string& rpath = "", bool copy = false) noexcept;
+
```
mgrudzinska, line 1576: do we consider support for other formats than ttf?

hermet: It's up to the formats. We can consider others later if users demand but not now.

inc/thorvg.h
```diff
@@ -1544,16 +1544,42 @@ class TVG_API Text final : public Paint
...
+     * @retval Result::NonSupport When trying to load a file with an unsupported extension.
+     * @retval Result::Unknown If an error occurs at a later stage.
+     *
+     * @warning: It's the user responsibility to release the @p data memory.
+     *
+     * @note Experimental API
+     *
+     * @see Text::unload(const std::string& fontId)
+     */
+    static Result load(const std::string& name, const char* data, uint32_t size, const std::string& rpath = "", bool copy = false) noexcept;
```
hermet, line 1575: 1. `name` is really necessary? I think we can remove it.
2. `rpath` neither.

hermet (commented):
Quickly review the API prototype. Please check and leave your opinions if you have any. Thanks.

inc/thorvg.h
mgrudzinska: I'll add mimetype to this api, so there won;t be any problem in the future

inc/thorvg.h
hermet: Probably the api doc could address the `name`  should have the `font name` to guide clearly.

mgrudzinska: > > 1. and how will I refer to this specific font after loading it?
> 
> Ah yes, it needs to be. My fault. Let's use `const char* name` instead.

I using string so this will be consistent with unload() api, which expects now path to be a string. if I change name to be const char* probably I should also change path to be  const char*?

hermet: Ok, let's use a string for now. Just... there is a conflict between the name (string) and font name (char). Let's clean it up under the experimental API later. Probably we should confirm this issue(https://github.com/thorvg/thorvg/issues/2200) and come up with a proper solution.

hermet: Considering this further, we have this option as well:
```
load("name", data, ...)       //load the name data
load("name", nullptr, ...)   //unload the name data.
```
```
load(path);           //load the path data
unload(path);       //unload the path data
```

mgrudzinska:
changed to draft since we discussing apis

inc/thorvg.h
hermet: > say we support both, ttf and woff, and we load two Arial fonts from memory, one from ttf and one from woff file. unloading using name will not work properly, so we should add mimetype to the unload api

In the end, I would like to replace all `std::string` usages with `char*`, both external and internal, to save memory and reduce size if there are no technical issues. `std::string` uses significantly more binary size in comparison.

> so maybe instead of that your approach is better. although it is counterintuitive to use load to unload, but I assume we don't want to add secone api to unload. right? 

I believe users can understand it if we properly address the load API doc. Thus, I don't think it's necessary to add unload() additionally.

inc/thorvg.h
```diff
@@ -1548,6 +1548,36 @@ class TVG_API Text final : public Paint
...
+     *
+     * @warning: It's the user responsibility to release the @p data memory.
+     *
+     * @note To unload the font data loaded using this API, pass the proper @p name and @c nullptr as @p data and zero as @p size.
+     * @note If you are unsure about the MIME type, you can provide an empty value like @c "", and thorvg will attempt to figure it out.
+     * @note Experimental API
+     *
+     * @see Text::font(const char* name, float size, const char* style)
+     */
+    static Result load(const char* name, const char* data, uint32_t size, const std::string& mimeType = "ttf", bool copy = true) noexcept;
```
mgrudzinska, line 1579: @hermet please note, that by default copy is true - it's because I need the loader to be cached and want to make it easier for the user - can use the def setting

hermet: @mgrudzinska Please set the default `false` as other apis do. svg loader could pass `true` there.

src/renderer/tvgText.cpp
```diff
@@ -71,6 +71,22 @@ Result Text::load(const std::string& path) noexcept
 }
 
 
+Result Text::load(const char* name, const char* data, uint32_t size, const string& mimeType, bool copy) noexcept
+{
+    if (!name) return Result::InvalidArguments;
+
+    //unload font
+    if (!data && size == 0) {
+        if (LoaderMgr::retrieve(name)) return Result::Success;
```
JSUYA, line 80: It is not appropriate that the load() function returns Success indicating that the unload was successful when there is no data.
Does load(name, nullptr, 0, ...) mean unload?

src/renderer/tvgText.cpp
```diff
@@ -71,6 +71,22 @@ Result Text::load(const std::string& path) noexcept
...
+        return Result::InvalidArguments;
+
+    } else if (!data || size <= 0) return Result::InvalidArguments;
```
JSUYA, line 83: Why not do an unload in this case?

inc/thorvg.h
mgrudzinska: "If @p copy is set to @c false, the caching mechanism still works, so it is the user's responsibility to ensure the @p data's validity for as long as needed."

this is why I forced true. I'll change to false, but caching still has to be active and it has nothing to do with svg loader. is it ok?

src/renderer/tvgText.cpp
mgrudzinska: yes... it is unloading... I'm not very happy with such method, but please check [this](https://github.com/thorvg/thorvg/pull/2345#discussion_r1625910687) and the discussion around. 
I think we could adopt the unload api... so it expects some char* (which for now is string), being either a path or a name. but with name should come a mime type, otherwise if we have more font loaders, there may be a problem.

src/renderer/tvgText.cpp
mgrudzinska: to prevent unloading by a mistake. to distinguish between wrong args and unloading. like with a dash api - if the user really wants to unload, has to pay attention and pass null and 0.

inc/thorvg.h
hermet: Please set false. That is intended. Data ownership will be on user side. Also, it would be better if the svg loader could optimize by avoiding the data copy including the Pictures.

hermet: @mgrudzinska Please see this design; it's the best scenario we are pursuing.
image

mgrudzinska: @hermet, ofc I changed. I'm just trying to highlight, that caching is working no matter what the copy value is (so no matter whether the data is shareable no not), this is opposite to other loading methods

from tvgLoader.cpp:
```
//loads fonts from memory - loader is cached (regardless of copy value) in order to access it while setting font
LoadModule* LoaderMgr::loader(const char* name, const char* data, uint32_t size, const string& mimeType, bool copy)
{
    //TODO: add check for mimetype ?
    if (auto loader = _findFromCache(name)) return loader;
```  

in other places we have:
```
LoadModule* LoaderMgr::loader(const uint32_t *data, uint32_t w, uint32_t h, bool premultiplied, bool copy)
{
    //Note that users could use the same data pointer with the different content.
    //Thus caching is only valid for shareable.
    if (!copy) {
        if (auto loader = _findFromCache((const char*)(data), w * h, "raw")) return loader;
    }
```

hermet: The loader cache is preferred. It's beneficial if we can support it. let's see

src/loaders/ttf/tvgTtfLoader.cpp
```diff
@@ -193,7 +193,13 @@ static uint32_t* _codepoints(const char* text, size_t n)
...
 void TtfLoader::clear()
 {
-    _unmap(this);
+    if (fromMemory) {
+        if (copy) free(reader.data);
+        reader.data = nullptr;
+        reader.size = 0;
+        copy = false;
+    }
+    else _unmap(this);
```
tinyjin, line 202: found minor convention thing.

```cpp
} else _unmap(this);
```

src/loaders/ttf/tvgTtfLoader.cpp
```diff
@@ -236,6 +241,25 @@ bool TtfLoader::open(const string& path)
 }
 
 
+bool TtfLoader::open(const char* data, uint32_t size, TVG_UNUSED const string& rpath, bool copy)
+{
+    clear();
```
hermet, line 246: 246 ~ 247 line doesn't necessary.

src/loaders/ttf/tvgTtfLoader.h
```diff
@@ -37,12 +37,15 @@ struct TtfLoader : public FontLoader
     char* text = nullptr;
     Shape* shape = nullptr;
     bool italic = false;
+    bool fromMemory = false;
```
hermet, line 40: Let's align the variable name.  (fromMemory -> freeData)
image

src/loaders/ttf/tvgTtfLoader.h
```diff
@@ -37,12 +37,15 @@ struct TtfLoader : public FontLoader
...
+    bool copy = false;
```
hermet, line 41: this looks unnecessary.

src/renderer/tvgText.cpp
```diff
@@ -71,6 +71,22 @@ Result Text::load(const std::string& path) noexcept
 }
 
 
+Result Text::load(const char* name, const char* data, uint32_t size, const string& mimeType, bool copy) noexcept
+{
+    if (!name) return Result::InvalidArguments;
+
+    //unload font
+    if (!data && size == 0) {
```
hermet, line 79:
```
if (!data) {
```

hermet (commented):
Please check comments. thanks.

inc/thorvg.h
```diff
@@ -1572,6 +1572,36 @@ class TVG_API Text final : public Paint
...
+     * @param[in] copy If @c true the data are copied into the engine local buffer, otherwise they are not (default).
+     *
+     * @retval Result::Success When succeed.
+     * @retval Result::InvalidArguments In case no name or no data are provided or the @p size is zero or less (except case when @p data = @c nullptr and @p size = zero are passed to unload the font @p name that was previously loaded).
+     * @retval Result::NonSupport When trying to load a file with an unsupported extension.
+     * @retval Result::Unknown If an error occurs at a later stage.
+     *
+     * @warning: It's the user responsibility to release the @p data memory.
+     *
+     * @note To unload the font data loaded using this API, pass the proper @p name and @c nullptr as @p data and zero as @p size.
```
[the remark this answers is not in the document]
mgrudzinska: I've also changed accordingly the line:
```* @retval Result::InvalidArguments In case no name is provided or the @p size is zero or less.```

File: src/loaders/ttf/tvgTtfLoader.h (hunk 9 above)
Comment (mgrudzinska) [reply src/loaders/ttf/tvgTtfLoader.h:41]: in tvgTtfLoader.cpp fun clear():
```
void TtfLoader::clear()
{
    if (fromMemory) {
        if (freeData) free(reader.data);
        reader.data = nullptr;
        reader.size = 0;
        freeData = false;
    } else _unmap(this);
    shape = nullptr;
}
```
@hermet  I wanted to avoid calling _unmap if _map wasn't called first. this is why I introduced fromMemory.

File: src/renderer/tvgText.cpp (hunk 5 above)
Comment (mgrudzinska) [reply src/renderer/tvgText.cpp]: @JSUYA it is changed now

File: src/renderer/tvgText.cpp [hunk 12]
```diff
@@ -71,6 +71,21 @@ Result Text::load(const std::string& path) noexcept
 }
 
 
+Result Text::load(const char* name, const char* data, uint32_t size, const string& mimeType, bool copy) noexcept
+{
+    if (!name || size <= 0) return Result::InvalidArguments;
```
hermet, line 76: `size <= 0` this looks weird.

src/loaders/ttf/tvgTtfLoader.h
hermet: @mgrudzinska 
1. Can you try to merge 2 vars?  `freeData` & `fromMemory`
2. Suggesting names: 
 - `direct`     //direct read from the data
 - `nomap`   //no mapping

hermet: @mgrudzinska 

Could you please merge the 2 vars? `freeData`, `fromMemory` 
Suggested names:
 - `direct`      //direct read from the data
 - `nomap`    //no mapping

src/renderer/tvgText.cpp
mgrudzinska: ```<=```  ->  ```==```

src/loaders/ttf/tvgTtfLoader.h
mgrudzinska: I can introduce some enum, but with bool I don't see how to do this. and an enum doesn't seem like a good idea.
`if nomap` does not mean that I have to free the data

hermet: > I can introduce some enum, but with bool I don't see how to do this.

it's ok. let me try it later.

> `if nomap` does not mean that I have to free the data

Yes, that's what I asked you to use `freeData` aligning with others. However, the condition says `mapping case` as you argued:

> I wanted to avoid calling _unmap if _map wasn't called first.

`nomap` is straightforward to use. `fromMemory` does not mean to either `free the data` or `map/unmap`.
```
if (nomap) { 
   //no-mapping scenario
} else {
   //mapping scenario
}
```

hermet (approved):

src/renderer/tvgText.cpp
```diff
@@ -71,6 +71,21 @@ Result Text::load(const std::string& path) noexcept
 }
 
 
+Result Text::load(const char* name, const char* data, uint32_t size, const string& mimeType, bool copy) noexcept
+{
+    if (!name || size == 0) return Result::InvalidArguments;
```
hermet, line 76: @mgrudzinska this might be precise:
`(size == 0 && data)`

PR #2351, opened 2024-06-03
common: clarification of returned types

mgrudzinska:
The functions setting stroke's features always
returned a true. Returnig a boolen was a remnant
from a previous implementation. Since now they
never return false, they can be void functions.
The APIs description has been corrected.


mgrudzinska:
capi docs updated

hermet (approved):

mgrudzinska:
rebased

PR #2352, opened 2024-06-03
svg_loader: Add embedded webp and ttf imageMimeTypes

capnm:
Fixes loading webp images embedded in SVG.
(+ ttf font for the near future)

Test image:
[svg_embedded_bitmap_png-jpg-webp.zip](https://github.com/user-attachments/files/15538744/svg_embedded_bitmap_png-jpg-webp.zip)

svg_embedded_bitmap_png-jpg-webp

@mgrudzinska I also did a quick ttf mime test with your font related PRs:
- loads the system default font
- does not parse the unicode godot ghost character

[MiniSans.zip](https://github.com/user-attachments/files/15538766/MiniSans.zip)

svg_embedded-ttf-font


mgrudzinska:
@capnm yes, xml entities are not supported now. do you need this? 
also I found such 'header' in the embedded fonts:  ```data:application/font-woff;base64,```
I assume ```font-ttf``` instead of ```font/ttf``` is also possible? in your sample the 'application' word is skipped, can it be something else? 
that's why for now I didn;t use minetypes...

capnm:
> @capnm yes, xml entities are not supported now. do you need this? 

Yes, this was intentionally included in the test font to check the standard HTML text characters.
See the MiniSans.nfo in the zip file, .e.g.  https://www.w3schools.com/charsets/ref_utf_misc_symbols.asp

> also I found such a 'header' in the embedded fonts: `data:application/font-woff;base64,` I assume `font-ttf` instead of `font/ttf` is also possible? in your sample the 'application' word is skipped, can it be something else? that's why for now I didn;t use minetypes...

[WOFF](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_fonts/WOFF) is a completely different (more complicated) binary font format.  @hermet has, if I see it correctly, implemented support for decoding and rasterizing ttf.

mgrudzinska:
> [WOFF](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_fonts/WOFF) is a completely different (more complicated) binary font format. @hermet has, if I see it correctly, implemented support for decoding and rasterizing ttf.

yes, I know that we support ttf, not woff, but I'm asking whether are you sure that this :
```{"font/ttf", sizeof("font/ttf"), imageMimeTypeEncoding::base64},```
is the only possibility. can it be like this```{"font-ttf", sizeof("font-ttf"), imageMimeTypeEncoding::base64},```?

**Comment by capnm** [comment]
> yes, I know that we support ttf, not woff, but I'm asking whether are you sure that this :
> `{"font/ttf", sizeof("font/ttf"), imageMimeTypeEncoding::base64},`
> is the only possibility. can it be like this`{"font-ttf", sizeof("font-ttf"), imageMimeTypeEncoding::base64},`?

I read your comment too quickly - good question. I just pasted it from [iana](https://www.iana.org/assignments/media-types/media-types.xhtml#font) The correct MIME type for a TTF font file is also *font/ttf*. But interestingly, WOFF has its own registered MIME type, *application/font-woff*. I'm not sure either, what do you think?

**Comment by mgrudzinska** [comment]
> > yes, I know that we support ttf, not woff, but I'm asking whether are you sure that this :
> > `{"font/ttf", sizeof("font/ttf"), imageMimeTypeEncoding::base64},`
> > is the only possibility. can it be like this`{"font-ttf", sizeof("font-ttf"), imageMimeTypeEncoding::base64},`?
> 
> I read your comment too quickly - good question. I just pasted it from [iana](https://www.iana.org/assignments/media-types/media-types.xhtml#font) The correct MIME type for a TTF font file is also _font/ttf_. But interestingly, WOFF has its own registered MIME type, _application/font-woff_. I'm not sure either, what do you think?

the answer is in your link, thx :)
"Deprecated Alias:  The existing registration "application/font-woff" is deprecated in favor of "font/woff"."
there is no such a comment for ttf, so lets focus now on ttf as you implemented and I'll do the corresponding changes in my pr.

**Comment by mgrudzinska** [comment]
@capnm and is this guarantied that the embedded ttf will be base64 encoded? 
also the naming is not perfect - 'imageMimeTypeEncoding' -> ttf is not an image.

@JSUYA can you tell why first we pair the image type with imageMimeTypeEncoding and then we still do the check in _isValidImageMimeTypeAndEncoding() ? do we really need the first step? all info we need is in the embedded data and either way we read it

**Comment by JSUYA** [comment]
> @JSUYA can you tell why first we pair the image type with imageMimeTypeEncoding and then we still do the check in _isValidImageMimeTypeAndEncoding() ? do we really need the first step? all info we need is in the embedded data and either way we read it

Well.., I think `imageMimeTypes` are the encoding types allowed by each mime, and `_isValidImageMimeTypeAndEncoding()` is the process of finding the corresponding encoding type from the actual data.
svg+xml could be both base64 or utf8(According to the code...). so we eventually need to find the actual encoding type of the data.
If there is no actual case, _isValidImageMimeTypeAndEncoding() may be removed, but I do not think it is unnecessary.

**Comment by capnm** [comment]
"Deprecated Alias: The existing registration "application/font-woff" is deprecated in favor of "font/woff"."

I see now, in the current revision of 2024-05-28, same font/xy mime can be seen for the new/old WOFF font:
ttf font/ttf [RFC8081]
woff font/woff [RFC8081]
woff2 font/woff2 [RFC8081]
[iana](https://www.iana.org/assignments/media-types/media-types.xhtml#font) 
Hopefully this will be supported by everyone :)

> @capnm and is this guarantied that the embedded ttf will be base64 encoded?
> also the naming is not perfect - 'imageMimeTypeEncoding' -> ttf is not an image.

I've tested Inkscape (you've already fixed their xml[\n] bug) and various web browser apps.
You can drag and drop the SVG test image into the window, I haven't spotted any issues so far:
 
Screenshot from 2024-06-04 04-56-29

 'imageMimeTypeEncoding' → *MimeTypeEncoding* ?

**Comment by capnm** [comment]
@hermet  the webp embedded support is currently without the mime actually bug/broken: 

image

**Comment by hermet** [comment]
@capnm We will check it, thanks! => https://github.com/thorvg/thorvg/issues/2355

**Comment by capnm** [comment]
> @capnm We will check it, thanks! => #2355

Should I rather remove the *ttf* mime now to avoid confusion? I wrote this PR mainly to fix this issue.

**Comment by hermet** [comment]
@capnm TLDR; sorry. The font might be a bit more controversial. If we separate it, we can fix the WebP issue quickly.

File: src/loaders/svg/tvgSvgSceneBuilder.cpp [hunk 1]
```diff
@@ -508,6 +508,8 @@ static constexpr struct
 } imageMimeTypes[] = {
     {"jpeg", sizeof("jpeg"), imageMimeTypeEncoding::base64},
     {"png", sizeof("png"), imageMimeTypeEncoding::base64},
+    {"webp", sizeof("webp"), imageMimeTypeEncoding::base64},
+    {"font/ttf", sizeof("font/ttf"), imageMimeTypeEncoding::base64},
```
JSUYA, line 512: I think imageMimeTypes is a list of image types that can be assigned to an image element. Therefore, the `font` is not applicable because it is not suitable for image elements. I recommend to leave only webp. nice fix :)

JSUYA (approved):

hermet (approved):

Issue #2349, opened 2024-06-03
lottie: not rendering


theashraf:
I'm not able to play this animation on thorvg viewer, any idea ? 

[test.json](https://github.com/user-attachments/files/15531971/test.json)

tinyjin:
Thank you @theashraf for reporting.

In my machine, `EXE_BAD_ACCEES` has caused.
![CleanShot 2024-06-03 at 19 03 19@2x](https://github.com/thorvg/thorvg/assets/11167117/470856f8-042c-4c76-b08a-34d2937a67bf)

Seems like the animation faces infinite loop between `_updateMatte` and `_updateLayer`.
![CleanShot 2024-06-03 at 19 03 00@2x](https://github.com/thorvg/thorvg/assets/11167117/1d53552a-8c6a-43b8-820a-389308180c95)

tinyjin:
This code line occurs the incorrect loop.

```cpp
if (!_updateMatte(root, layer, frameNo, exps)) return;
```

Without this, it goes render, but matte is broken.
![CleanShot 2024-06-03 at 19 08 35](https://github.com/thorvg/thorvg/assets/11167117/54de3b40-784b-4947-ac61-de64ebbe16d1)

mgrudzinska:
@theashraf already fixed on main, will be present in v0.13.7

PR #2353, opened 2024-06-03
lottie: fix matte layer finding

mgrudzinska:
Layer identifiers do not have to be unique within
the entire file - they are unique within a given
group. Searching the entire composition to find
the referenced matte was an incorrect approach.

Error introduced by 1ee79a6c2afafb6b16b14abc969b60c1dbc7a065

@Issue: https://github.com/thorvg/thorvg/issues/2349


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1315,7 +1315,7 @@ static void _attachFont(LottieComposition* comp, LottieLayer* parent)
 }
 
 
-static bool _buildComposition(LottieComposition* comp, LottieGroup* parent)
+static bool _buildComposition(LottieComposition* comp, LottieLayer* parent)
```
mgrudzinska, line 1318: @hermet  is there a reason why this should be a LottieGroup object?

hermet: @mgrudzinska No should be. we can change it if necessary.

tinyjin:
CleanShot 2024-06-04 at 11 25 14

My test result, it's working well!

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1335,7 +1335,7 @@ static bool _buildComposition(LottieComposition* comp, LottieGroup* parent)
                 }
             //matte layer is specified by an index.
             } else {
-                if (auto matte = comp->layerByIdx(child->mid)) {
+                if (auto matte = parent->layerByIdx(child->mid)) {
```
hermet, line 1338: Since on this change, could you please make it less condition, compact code?
```
child->matteTarget = parent->layerByIdx(child->mid);
```

hermet (approved):

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: one static_cast less, if we keep Lottie Group I have to do 
```child->matteTarget = static_cast<LottieLayer*>(parent)->layerByIdx(child->mid);```

SergeyLebedkin (approved):

PR #2358, opened 2024-06-04
wg_engine: cross-platform support (windows, linux and macos)

SergeyLebedkin:
Make an ability to run ThorVG on linux platform including wsl2 and windows platform

Also it provide changes public API for webgpu canvas interface to provide nessesary handles to native window for different platforms:
```
class TVG_API WgCanvas final : public Canvas
	...
	Result target(void* instance, void* surface, uint32_t w, uint32_t h) noexcept;
```

Example of usage for glfw library

```
// create platform defined canvas target
void createCanvasTarget(GLFWwindow* window, tvg::WgCanvas* canvas) {
    // create webgpu instance
    WGPUInstanceDescriptor instanceDesc{};
    WGPUInstance instance = wgpuCreateInstance(&instanceDesc);

    // create webgpu surface
    WGPUSurface surface = glfwGetWGPUSurface(instance, window);

    // get window width and height
    int width{}, height{};
    glfwGetWindowSize(window, &width, &height);

    // create target
    canvas->target(instance, surface, width, height);
}

```

Also see the example [glfwGetWGPUSurface](https://github.com/eliemichel/glfw3webgpu)

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1745,7 +1745,7 @@ public:
      * @note Experimental API
      * @see Canvas::viewport()
      */
-    Result target(void* window, uint32_t w, uint32_t h) noexcept;
+    Result target(void *disp_inst, void *wind_serf, uint32_t w, uint32_t h) noexcept;
 
     /**
      * @brief Creates a new WgCanvas object.
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -80,7 +80,7 @@ enum class Result
     InsufficientCondition, ///< The value returned in case the request cannot be processed - e.g. asking for properties of an object, which does not exist.
     FailedAllocation,      ///< The value returned in case of unsuccessful memory allocation.
     MemoryCorruption,      ///< The value returned in the event of bad memory handling - e.g. failing in pointer releasing or casting
-    NonSupport,            ///< The value returned in case of choosing unsupported options.
+    NonSupport,            ///< The value returned in case of choosing unsupported engine features(options).
     Unknown                ///< The value returned in all other cases.
 };
 
@@ -956,7 +956,7 @@ public:
      *
      * @param[in] width The width of the stroke. The default value is 0.
      *
-     * @retval Result::Success when succeed, Result::FailedAllocation otherwise.
+     * @retval Result::Success when succeed.
      */
     Result strokeWidth(float width) noexcept;
 
@@ -968,7 +968,7 @@ public:
      * @param[in] b The blue color channel value in the range [0 ~ 255]. The default value is 0.
      * @param[in] a The alpha channel value in the range [0 ~ 255], where 0 is completely transparent and 255 is opaque. The default value is 0.
      *
-     * @retval Result::Success when succeed, Result::FailedAllocation otherwise.
+     * @retval Result::Success when succeed.
      */
     Result strokeFill(uint8_t r, uint8_t g, uint8_t b, uint8_t a = 255) noexcept;
 
@@ -978,8 +978,7 @@ public:
      * @param[in] f The gradient fill.
      *
      * @retval Result::Success When succeed.
-     * @retval Result::FailedAllocation An internal error with a memory allocation for an object to be filled.
-     * @retval Result::MemoryCorruption In case a @c nullptr is passed as the argument.
+     * @retval Result::MemoryCorruption In case a @c nullptr is passed as the argument or an error with accessing it.
      */
     Result strokeFill(std::unique_ptr<Fill> f) noexcept;
 
@@ -1006,7 +1005,7 @@ public:
      *
      * @param[in] cap The cap style value. The default value is @c StrokeCap::Square.
      *
-     * @retval Result::Success when succeed, Result::FailedAllocation otherwise.
+     * @retval Result::Success when succeed.
      */
     Result strokeCap(StrokeCap cap) noexcept;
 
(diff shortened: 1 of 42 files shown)


src/renderer/wg_engine/tvgWgCommon.cpp
```diff
@@ -170,7 +193,6 @@ WGPUTexture WgContext::createTexture2d(WGPUTextureUsageFlags usage, WGPUTextureF
 
 WGPUTexture WgContext::createTexture2dMS(WGPUTextureUsageFlags usage, WGPUTextureFormat format, uint32_t width, uint32_t height, uint32_t sc, char const * label)
 {
-
     WGPUTextureDescriptor textureDesc{};
```
tinyjin, line 196: Probably, unnecessary code line?

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -314,38 +305,15 @@ bool WgRenderer::target(uint32_t* buffer, uint32_t stride, uint32_t w, uint32_t
 
 
 // target for native window handle
-bool WgRenderer::target(void* window, uint32_t w, uint32_t h)
+bool WgRenderer::target(void *disp_inst, void *wind_serf, uint32_t w, uint32_t h)
```
tinyjin, line 308: I don't know much about usage/convention of void pointer.
Maybe this type is necessary for wg code, otherwise i'd use explicit type.

The code convention also changed, it should be `void* disp_inst, void* wind_serf`?

SergeyLebedkin: > Maybe this type is necessary for wg code, otherwise i'd use explicit type

Different patforms used different types for native window description. But in most cases there are always 2 pointers, like pointer to display and window
void* is the most universal type to support all platforms

src/renderer/wg_engine/tvgWgCommon.cpp
SergeyLebedkin: > Probably, unnecessary code line?
simply removed empty string

inc/thorvg.h
```diff
@@ -1745,7 +1745,7 @@ class TVG_API WgCanvas final : public Canvas
      * @note Experimental API
      * @see Canvas::viewport()
      */
-    Result target(void* window, uint32_t w, uint32_t h) noexcept;
+    Result target(void *disp_inst, void *wind_serf, uint32_t w, uint32_t h) noexcept;
```
hermet, line 1748: Please check the parameters, update the documentation as well.
```
(void* display, void* window, uint32_t w, uint32_t h)
```

src/renderer/wg_engine/tvgWgCommon.cpp
```diff
@@ -27,18 +27,22 @@
 // context
 //*****************************************************************************
 
-void WgContext::initialize()
+void WgContext::initialize(void* disp_inst, void* wind_serf, uint32_t w, uint32_t h)
```
hermet, line 30: please take care of the coding convention. thorvg doesn't use `a_b` 
disp_inst -> display
wind_serf -> window

src/renderer/wg_engine/tvgWgRenderer.h
```diff
@@ -52,7 +52,7 @@ class WgRenderer : public RenderMethod
     bool sync() override;
 
     bool target(uint32_t* buffer, uint32_t stride, uint32_t w, uint32_t h);
-    bool target(void* window, uint32_t w, uint32_t h); // temporary solution
+    bool target(void* disp_inst, void* wind_serf, uint32_t w, uint32_t h); // cross-olatform solution
```
hermet, line 55: please remove the useless comment.

hermet (commented):
Please check comments.

src/renderer/wg_engine/tvgWgCommon.cpp
```diff
@@ -327,6 +349,27 @@ void WgContext::releaseIndexBuffer(WGPUBuffer& buffer)
 }
 
 
+WGPUSurface WgContext::createSurface(WGPUInstance instance, void* disp_inst, void* wind_serf)
```
hermet, line 352: I think we have two approaches to consider here:
1. Receive the generated instances of `WGPUSurface` / `WGPUInstance` from the user side instead of the window/display handles.
2. Add an additional parameter for the window type in the API so that we can properly identify the current window system.

So far, I think option 1 is better for ThorVG as it should not depend on the window system. However, we can improve this point further after this PR.

src/renderer/wg_engine/tvgWgCommon.cpp
```diff
@@ -86,6 +90,20 @@ void WgContext::initialize()
     // set device error handling
     wgpuDeviceSetUncapturedErrorCallback(device, onDeviceError, nullptr);
 
+    // configure surface
+    WGPUSurfaceConfiguration surfaceConfiguration{};
+    surfaceConfiguration.nextInChain = nullptr;
+    surfaceConfiguration.device = device;
+    surfaceConfiguration.format = WGPUTextureFormat_BGRA8Unorm;
```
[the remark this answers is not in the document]
SergeyLebedkin: This format must be compatible with native window formats, that are only BGRA in all of the existing platforms. In other case context creation will be failed

src/renderer/wg_engine/tvgWgCommon.cpp
SergeyLebedkin: 1. So it will create dependence from webgpu.h header on the user side, where WGPUSurface/WGPUInstance types are described
2. Yes, its can be implemented. So we need to add public enum for a platform description

src/renderer/wg_engine/tvgWgCommon.cpp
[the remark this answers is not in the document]
hermet: If so, how to generate RGBA image format?

As the WebGPU surface requests the color space requirement, ThorVG should also allow users to have an option to satisfy their working environment. ThorVG is a graphics engine, it should not depend on the window system itself also should not limit the rendering image format. Please consider this, later I can assure, webgpu engine needs to rectify this.

src/renderer/wg_engine/tvgWgCommon.cpp
hermet: > 1. So it will create dependence from webgpu.h header on the user side, where WGPUSurface/WGPUInstance types are described

So far, that's correct. The best approach is for the toolkit layer to abstract this point so apps don't depend on WebGPU directly.

hermet:
@SergeyLebedkin please change the status to "Ready for review"

hermet (approved):

PR #2360, opened 2024-06-05
lottie: Add exception for lottie doesn't have exported glyph

tinyjin:
Added handling logic for missing characters to prevent program hang.

> [text-grouping-words.json](https://github.com/user-attachments/files/15584103/text-grouping-words.json)

I referred to the `lottie-web` for preventing logic, in this case `lottie-web` [returns emptyChar](https://github.com/airbnb/lottie-web/blob/master/build/player/lottie_canvas.js#L7527), which has zero width and size.

ThorVG will also skip for this unknown character.


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1128,9 +1129,14 @@ static void _updateText(LottieLayer* layer, float frameNo)
 
                 //advance the cursor position horizontally
                 cursor.x += glyph->width + spacing + doc.tracking;
+
+                found = true;
                 break;
             }
         }
+
+        //glyph not found
```
hermet, line 1138: -- not a very useful comment.

hermet (approved):

mgrudzinska (approved):

Issue #2350, opened 2024-06-03
Text in svg cannot be displayed correctly


zengzm1212:
Below is the svg file I used for testing
![svg_text](https://github.com/thorvg/thorvg/assets/59642136/bbff97c7-0b71-4a12-89b1-10979f36bcc4)

hermet:
System font is required. It is currently out of ThorVG's spec support.

zengzm1212:
@hermet  Could you load the text information into svgloader first?

hermet:
@zengzm1212 figuring out the system font in thorvg side is not simple. Let's see. :-)

hermet:
@zengzm1212 Alternatively, your app? is direcltly able to load the system font using the ThorVG Text API. This will likely enable SVG to support this text soon.

zengzm1212:
![image](https://github.com/thorvg/thorvg/assets/59642136/b0378325-07f1-4fe7-8392-f974359f298c)

I want to use SvgLoader to load the information in svg and then convert it into my own data structure, but I found that the corresponding SvgNode is not generated when loading text.
I'm going to simply add a TextNode for myself.  Anyway, thank you very much for your answers.

hermet:
@zengzm1212 see -> https://github.com/thorvg/thorvg/pull/2246/files

mgrudzinska:
@zengzm1212 can you please confirm whether we can close this issue? 
[the text node support](https://github.com/thorvg/thorvg/pull/2362) is already merged  into main, will be present in v0.13.8

PR #2362, opened 2024-06-06
svg_loader: handle text node

mgrudzinska:
The text node is handled, but default values
of the font-family and font-size as used in
the user's system are not.
For now font has to be loaded by the user.

@Issue: https://github.com/thorvg/thorvg/issues/2350

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2119,7 +2119,72 @@ static SvgNode* _createUseNode(SvgLoaderData* loader, SvgNode* parent, const cha
 }
 
 
-//TODO: Implement 'text' primitive
+static constexpr struct
+{
+    const char* tag;
+    SvgParserLengthType type;
+    int sz;
+    size_t offset;
+} textTags[] = {
+        {"x", SvgParserLengthType::Horizontal, sizeof("x"), offsetof(SvgTextNode, x)},
+        {"y", SvgParserLengthType::Vertical, sizeof("y"), offsetof(SvgTextNode, y)},
+        {"font-size", SvgParserLengthType::Vertical, sizeof("font-size"), offsetof(SvgTextNode, fontSize)}
+};
+
+
+static bool _attrParseTextNode(void* data, const char* key, const char* value)
+{
+    SvgLoaderData* loader = (SvgLoaderData*)data;
+    SvgNode* node = loader->svgParse->node;
+    SvgTextNode* text = &(node->node.text);
+
+    unsigned char* array;
+    int sz = strlen(key);
+
+    array = (unsigned char*)text;
+    for (unsigned int i = 0; i < sizeof(textTags) / sizeof(textTags[0]); i++) {
+        if (textTags[i].sz - 1 == sz && !strncmp(textTags[i].tag, key, sz)) {
+            *((float*)(array + textTags[i].offset)) = _toFloat(loader->svgParse, value, textTags[i].type);
+            return true;
+        }
+    }
+
+    if (!strcmp(key, "font-family")) {
+        if (text->fontFamily && value) free(text->fontFamily);
+        text->fontFamily = strdup(value);
+    } else if (!strcmp(key, "style")) {
+        return simpleXmlParseW3CAttribute(value, strlen(value), _parseStyleAttr, loader);
+    } else if (!strcmp(key, "clip-path")) {
+        _handleClipPathAttr(loader, node, value);
+    } else if (!strcmp(key, "mask")) {
+        _handleMaskAttr(loader, node, value);
+    } else if (!strcmp(key, "id")) {
+        if (node->id && value) free(node->id);
+        node->id = _copyId(value);
+    } else if (!strcmp(key, "class")) {
+        _handleCssClassAttr(loader, node, value);
+    } else {
+        return _parseStyleAttr(loader, key, value, false);
+    }
+    return true;
+}
+
+
+static SvgNode* _createTextNode(SvgLoaderData* loader, SvgNode* parent, const char* buf, unsigned bufLength, parseAttributes func)
+{
+    loader->svgParse->node = _createNode(parent, SvgNodeType::Text);
+    if (!loader->svgParse->node) return nullptr;
+
+    //TODO: support the def font and size as used in a system?
+    loader->svgParse->node->node.text.fontSize = 10.0f;
+    loader->svgParse->node->node.text.fontFamily = nullptr;
+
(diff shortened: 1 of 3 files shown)


src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -374,6 +374,14 @@ struct SvgCssStyleNode
 {
 };
 
+struct SvgTextNode
+{
+    char* text;
+    float x, y;
+    float fontSize;
+    char* fontFamily;
+};
```
tinyjin, line 383: Ordering variables like this could be more optimal and packed in memory perspective

```cpp
char* text;
char* fontFamily;
float x, y;
float fontSize;
```

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -335,6 +338,46 @@ static void _applyProperty(SvgLoaderData& loaderData, SvgNode* node, Shape* vg,
...
+
+        if (style->fill.paint.gradient->type == SvgGradientType::Linear) {
+            auto linear = _applyLinearGradientProperty(style->fill.paint.gradient, bBox, style->fill.opacity);
+            vg->fill(std::move(linear));
+        } else if (style->fill.paint.gradient->type == SvgGradientType::Radial) {
+            auto radial = _applyRadialGradientProperty(style->fill.paint.gradient, bBox, style->fill.opacity);
+            vg->fill(std::move(radial));
+        }
+    } else if (style->fill.paint.url) {
+        //TODO: Apply the color pointed by url
```
tinyjin, line 361: TVGLOG can efficiently show current coverage of spec for unsupported feature?

```cpp
TVGLOG("LOTTIE", "3D transform(ddd) is not totally compatible.");
```

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2119,7 +2119,72 @@ static SvgNode* _createUseNode(SvgLoaderData* loader, SvgNode* parent, const cha
...
+        _handleMaskAttr(loader, node, value);
+    } else if (!strcmp(key, "id")) {
+        if (node->id && value) free(node->id);
+        node->id = _copyId(value);
+    } else if (!strcmp(key, "class")) {
+        _handleCssClassAttr(loader, node, value);
+    } else {
+        return _parseStyleAttr(loader, key, value, false);
+    }
+    return true;
```
tinyjin, line 2169: `return true;` will never be called?

We can remove the line and replace them with  `return _parseStyleAttr(loader, key, value, false);` by removing else condition.

mgrudzinska: not in this pr, all _attrParse fun are like that, all should be changed together for consistency, but thx, will do

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -772,6 +801,34 @@ static unique_ptr<Scene> _useBuildHelper(SvgLoaderData& loaderData, const SvgNod
...
 
+static unique_ptr<Text> _textBuildHelper(SvgLoaderData& loaderData, const SvgNode* node, const Box& vBox, const string& svgPath)
+{
+    auto textNode = &node->node.text;
+    if (!textNode->text) return nullptr;
+    auto text = Text::gen();
+
+    Matrix textTransform = {1, 0, 0, 0, 1, 0, 0, 0, 1};
+    if (node->transform) textTransform = *node->transform;
+    if (!mathZero(node->node.text.x) || !mathZero(node->node.text.y)) {
```
hermet, line 812: Please check this "mathTranslateR()"

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -304,14 +312,21 @@ static void _applyComposition(SvgLoaderData& loaderData, Paint* paint, const Svg
...
-    if (node->transform && !clip) vg->transform(*node->transform);
-    if (node->type == SvgNodeType::Doc || !node->style->display) return;
 
+static inline void _setFill(Text* text, uint8_t r, uint8_t g, uint8_t b, TVG_UNUSED uint8_t a)
+{
+    text->fill(r, g, b);
+}
+
+
+template<typename Vg, typename std::enable_if<std::is_same<Vg, Shape*>::value || std::is_same<Vg, Text*>::value, int>::type = 0>
```
mgrudzinska, line 327: @hermet please comment whether we are ok with such template

hermet:
@mgrudzinska @JSUYA This is unrelated to the current review... I think the structure of the SVG loader is becoming a tangled mess, like spaghetti. I believe it needs to be redesigned with a clean architecture. The first job should be to separate the 'builder' and 'parser'.

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: by mistake, sorry

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -772,6 +815,34 @@ static unique_ptr<Scene> _useBuildHelper(SvgLoaderData& loaderData, const SvgNod
...
+    if (!textNode->text) return nullptr;
+    auto text = Text::gen();
+
+    Matrix textTransform = {1, 0, 0, 0, 1, 0, 0, 0, 1};
+    if (node->transform) textTransform = *node->transform;
+    if (!mathZero(node->node.text.x) || !mathZero(node->node.text.y)) {
+        Matrix translate = {1, 0, node->node.text.x, 0, 1, node->node.text.y - textNode->fontSize, 0, 0, 1};
+        textTransform = textTransform * translate;
+    }
+    mathLog(&textTransform);
```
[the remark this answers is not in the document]
mgrudzinska: by mistake, sorry

src/loaders/svg/tvgSvgSceneBuilder.cpp
hermet: Please reach a consensus with @JSUYA. I will follow you guys consensus. In my opinion, using complicate templates will hinder the ThorVG project, as newcomers will face a significant barrier in understanding the code. I would not prefer it unless it helps to either a) reduce massive duplicate code (without increasing the binary size much) or b) enhance performance.

mgrudzinska: I don't like this solution (I wrote it :D ) because we still have the same code for a stroke.

hermet (approved):

PR #2364, opened 2024-06-07
examples: fix memory leak

hermet:


hermet:
The Windows unit test failures is not related to this change.

tinyjin (approved):

Issue #2332, opened 2024-05-30
lottie: high cpu usage causing a browser freeze 


theashraf:
I noticed very high CPU usage when playing this animation on dotlottie-rs: [notification.json](https://github.com/thorvg/thorvg/files/15495049/notification.json).

When playing in the browser, the browser freezes. I observed the same issue with the thorvg viewer.

Related issue: [https://github.com/LottieFiles/dotlottie-web/issues/221](https://github.com/LottieFiles/dotlottie-web/issues/221)

hermet:
Huge size shapes with range [503005.187500 ~ -605279.125000] significantly affect the RLE segmentation performance.

baseplate-admin:
Hi @hermet, the file seems to play with [`lottie-player`](https://www.npmjs.com/package/@lottiefiles/lottie-player)

PR #2375, opened 2024-06-07
renderer: clipping optimization

hermet:
issue: https://github.com/thorvg/thorvg/issues/2332

### DIFF
--- a/src/renderer/tvgPaint.cpp
+++ b/src/renderer/tvgPaint.cpp
@@ -41,8 +41,39 @@
     }
 
 
+static Result _clipRect(RenderMethod* renderer, const Point* pts, const RenderTransform* pTransform, RenderTransform* rTransform, RenderRegion& before)
+{
+    //sorting
+    Point tmp[4];
+    Point min = {FLT_MAX, FLT_MAX};
+    Point max = {0.0f, 0.0f};
 
-static Result _compFastTrack(Paint* cmpTarget, const RenderTransform* pTransform, RenderTransform* rTransform, RenderRegion& viewport)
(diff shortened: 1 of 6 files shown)


src/renderer/tvgPaint.cpp
```diff
@@ -88,13 +123,15 @@ static Result _compFastTrack(Paint* cmpTarget, const RenderTransform* pTransform
...
-        viewport.y = static_cast<int32_t>(v1.y);
-        viewport.w = static_cast<int32_t>(ceil(v2.x - viewport.x));
-        viewport.h = static_cast<int32_t>(ceil(v2.y - viewport.y));
+        after.x = static_cast<int32_t>(v1.x);
+        after.y = static_cast<int32_t>(v1.y);
+        after.w = static_cast<int32_t>(ceil(v2.x - after.x));
+        after.h = static_cast<int32_t>(ceil(v2.y - after.y));
+
+        if (after.w < 0) after.w = 0;
+        if (after.h < 0) after.h = 0;
```
mgrudzinska, line 132: in case after.w or after.h <= 0 there is no point of intersecting the regions, right? I think we could set w=h=0 and return

hermet: It's ignorable. No point for a super corner case. Let's keep the code clean.

mgrudzinska (approved):

src/renderer/tvgPaint.cpp
mgrudzinska: ok, so no comments otherwise

PR #2392, opened 2024-06-11
loaders: ensure a null terminator at the end of the data

mgrudzinska:
strlen does not count the null terminator - using it to determine the size of loaded data that is copied results in not copying the null terminator and may lead to memory corruption.

### DIFF
--- a/src/loaders/lottie/tvgLottieLoader.cpp
+++ b/src/loaders/lottie/tvgLottieLoader.cpp
@@ -182,10 +182,11 @@ bool LottieLoader::header()
 bool LottieLoader::open(const char* data, uint32_t size, const std::string& rpath, bool copy)
 {
     if (copy) {
-        content = (char*)malloc(size);
+        content = (char*)malloc(size + 1);
         if (!content) return false;
-        memcpy((char*)content, data, size);
-    } else content = data;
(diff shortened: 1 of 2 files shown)


mgrudzinska:
rebased

hermet:
> strlen does not count the null terminator

hmm you changed about binary data, not strings. if someone uses the strlen() then that's the fix point?

hermet (changes requested):
This doesn't make sense.

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -182,10 +182,11 @@ bool LottieLoader::header()
 bool LottieLoader::open(const char* data, uint32_t size, const std::string& rpath, bool copy)
 {
     if (copy) {
-        content = (char*)malloc(size);
+        content = (char*)malloc(size + 1);
         if (!content) return false;
         memcpy((char*)content, data, size);
```
hermet, line 187:
```
 memcpy(content, data, size);
```

hermet (approved):

hermet:
@mgrudzinska considering about this again, we can't prevent the scenario "non copy". Ultimately, this is an user error they should correct this.

Issue #2389, opened 2024-06-11
lottie: TVG_RESULT_NOT_SUPPORTED error when loading specific Lottie JSON string


theashraf:
I am encountering an issue in [dotlottie-rs](https://github.com/LottieFiles/dotlottie-rs) where the `tvg_picture_load_data` C API returns a `TVG_RESULT_NOT_SUPPORTED` error when attempting to load a specific Lottie JSON string. Below is the JSON string that is causing the issue:

[...]

PR #2393, opened 2024-06-11
lottie: parse a number despite an incorrect type

mgrudzinska:
It may happen that a number which should be an int is actually a float. In such cases, a parsing error occurred and the animation was not rendered. Now,
an attempt is made to parse the number as a float
first, and only if this fails, the parsing is concluded with an error.

@Issue: https://github.com/thorvg/thorvg/issues/2389

### DIFF
--- a/src/loaders/lottie/tvgLottieParserHandler.cpp
+++ b/src/loaders/lottie/tvgLottieParserHandler.cpp
@@ -86,10 +86,14 @@ bool LookaheadParserHandler::nextArrayValue()
 
 int LookaheadParserHandler::getInt()
 {
-    if (state != kHasNumber || !val.IsInt()) {
+    if (state != kHasNumber) {
         Error();
         return 0;
     }
+
+    // attempt to parse as float as a backup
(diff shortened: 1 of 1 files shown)


hermet:
This might become a bit of an issue. Hiding issue by an engine side wouldn't be a good solution if there is a spec error, the usage must be corrected first.

hermet (changes requested):
Please make it sure using getFloat() if the data expects the float.

mgrudzinska:
@hermet please check whether this is what you meant. I pushed changes

hermet:
You don't need to modify getInt() behavior. parser could call the getFloat() instead of getInt() if the data requires the float type.

hermet:
> @hermet am I allowed to introduce getNum() and inside decide which one to call: getInt or getFloat? or you mean that I should take care about this specific json and solve the problem only for the 'm' proparty of the repeater?

Basically, this is not a thorvg problem. I would like to confirm the root-cause first why this issue is occurred as I argued => "if the data requires the float type."

hermet:
Let's replace with this: https://github.com/thorvg/thorvg/pull/2507

PR #2405, opened 2024-06-15
test for text loading from memory + issues caught during testing

mgrudzinska:
issues caught during testing:
 - wrong returned value while unloading
 - lack of support for unprovided mimetype
 - missing file.close in Text example

### DIFF
--- a/src/renderer/tvgLoader.cpp
+++ b/src/renderer/tvgLoader.cpp
@@ -436,23 +436,22 @@ LoadModule* LoaderMgr::loader(const uint32_t *data, uint32_t w, uint32_t h, bool
 
 
 //loads fonts from memory - loader is cached (regardless of copy value) in order to access it while setting font
-LoadModule* LoaderMgr::loader(const char* name, const char* data, uint32_t size, const string& mimeType, bool copy)
+LoadModule* LoaderMgr::loader(const char* name, const char* data, uint32_t size, TVG_UNUSED const string& mimeType, bool copy)
 {
     //TODO: add check for mimetype ?
     if (auto loader = _findFromCache(name)) return loader;
 
-    if (auto loader = _findByType(mimeType)) {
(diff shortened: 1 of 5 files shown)


src/renderer/tvgLoader.cpp
```diff
@@ -441,15 +441,31 @@ LoadModule* LoaderMgr::loader(const char* name, const char* data, uint32_t size,
...
+                _activeLoaders.back(loader);
+                return loader;
+            } else {
+                TVGLOG("LOADER", "The font data \"%s\" could not be loaded.", name);
+                delete (loader);
+            }
+        }
+    }
+    //Unkown MimeType. Try with the candidates in the order
+    for (int i = static_cast<int>(FileType::Ttf); i < static_cast<int>(FileType::Raw); i++) {
```
hermet, line 459: What about directly accessing the font (TTF) loader? I don't think it's necessary to involve the other loaders. Above(444) as well.

src/renderer/tvgLoader.cpp
```diff
@@ -441,15 +441,31 @@ LoadModule* LoaderMgr::loader(const char* name, const char* data, uint32_t size,
...
-    if (auto loader = _findByType(mimeType)) {
-        if (loader->open(data, size, "", copy)) {
-            loader->hashpath = strdup(name);
-            loader->pathcache = true;
-            ScopedLock lock(key);
-            _activeLoaders.back(loader);
-            return loader;
-        } else {
-            TVGLOG("LOADER", "The font data \"%s\" could not be loaded.", name);
+    if (!mimeType.empty()) {
```
hermet, line 444: if (mimeType == null) mimeType = "ttf"

mgrudzinska: please check now, I ignored the mimetype and force to use the ttf loader

hermet (approved):

PR #2414, opened 2024-06-18
infra: --egl dependncy

hermet:
there is no more EGL dependency in gl engine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,7 +494,7 @@ meson setup builddir -Dbindings="capi"
 ## Dependencies
 ThorVG offers versatile support for image loading, accommodating both static and external loaders. This flexibility ensures that, even in environments without external libraries, users can still leverage static loaders as a reliable alternative. At its foundation, the ThorVG core library is engineered to function autonomously, free from external dependencies. However, it is important to note that ThorVG also encompasses a range of optional feature extensions, each with its specific set of dependencies. The dependencies associated with these selective features are outlined as follows:
 
-* GL engine: [EGL](https://www.khronos.org/egl), [GLESv2/GLESv3](https://www.khronos.org/opengles/)
+* GL engine: [GLESv2/GLESv3](https://www.khronos.org/opengles/)
 * WG engine: [webgpu-native](https://github.com/gfx-rs/wgpu-native)
 * External PNG support: [libpng](https://github.com/glennrp/libpng)
 * External JPG support: [turbojpeg](https://github.com/libjpeg-turbo/libjpeg-turbo)
--- a/src/renderer/gl_engine/tvgGlCommon.h
+++ b/src/renderer/gl_engine/tvgGlCommon.h
@@ -39,17 +39,6 @@
           } \
         } while(0)
 
-#define EGL_CHECK(x) \
-    x; \
-    do { \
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -494,7 +494,7 @@ meson setup builddir -Dbindings="capi"
 ## Dependencies
 ThorVG offers versatile support for image loading, accommodating both static and external loaders. This flexibility ensures that, even in environments without external libraries, users can still leverage static loaders as a reliable alternative. At its foundation, the ThorVG core library is engineered to function autonomously, free from external dependencies. However, it is important to note that ThorVG also encompasses a range of optional feature extensions, each with its specific set of dependencies. The dependencies associated with these selective features are outlined as follows:
 
-* GL engine: [EGL](https://www.khronos.org/egl), [GLESv2/GLESv3](https://www.khronos.org/opengles/)
+* GL engine: [GL](https://www.khronos.org/gl)
```
RuiwenTang, line 497: This URL point to 404 page ....

hermet: @RuiwenTang fault! => https://www.khronos.org/opengl/

RuiwenTang:
Strictly speaking, OpenGL and OpenGLES are two different implementations, and the current GLEngine Shader is only compatible with OpenGLES.
Does this need to be marked?

hermet:
@RuiwenTang got it, then we'd rather stick to use gles officially. fine?

RuiwenTang:
If the target is OpenGL, only the shaders need some changes.
And to do that we need to append some API or enumeration to specify whether we are targeting OpenGL or OpenGLES.What do you think about this? @hermet

hermet:
> If the target is OpenGL, only the shaders need some changes. And to do that we need to append some API or enumeration to specify whether we are targeting OpenGL or OpenGLES.What do you think about this? @hermet

@RuiwenTang Basically, I prefer to use GLES for wider platform support. Unless our implementation specifically requires the GL spec, we don't need to use GL. I am not aware of any cases where a user should use GL for ThorVG because I expect GL systems can support GLES as well. It's my opinion but I respect your decision. It's your maintaining domain.

RuiwenTang:
> > If the target is OpenGL, only the shaders need some changes. And to do that we need to append some API or enumeration to specify whether we are targeting OpenGL or OpenGLES.What do you think about this? @hermet
> 
> @RuiwenTang I prefer to use GLES for wider platform support. Unless our implementation specifically requires the GL spec, we don't need to use GL. I am not aware of any cases where a user should use GL for ThorVG because I expect user systems, where supporting GL, also able to support GLES like my local systems. It's my opinion but I respect your decision. It's your maintaining domain.

This depends on the target platform that thorvg wants to support. Not all platforms support OpenGLES, which is why Google brought [Angle](https://github.com/google/angle) to simulate OpenGLES.

For me, it is simpler to only support OpenGLES, and I don't need to worry about OpenGL compatibility issues. :)

hermet:
@RuiwenTang I'm trying to make a decision on the direction ThorVG should take, not for us engineers. :) Either way between gles / gl, this PR resolves the incorrect dependency. Please have a review.

RuiwenTang:
For now, I think we can only support GLES to make things simple

PR #2417, opened 2024-06-19
common: add missing ifdef

mgrudzinska:
In the absence of support for the TTF loader
crashes occurred. Fixed.

### DIFF
--- a/src/renderer/tvgLoader.cpp
+++ b/src/renderer/tvgLoader.cpp
@@ -438,6 +438,7 @@ LoadModule* LoaderMgr::loader(const uint32_t *data, uint32_t w, uint32_t h, bool
 //loads fonts from memory - loader is cached (regardless of copy value) in order to access it while setting font
 LoadModule* LoaderMgr::loader(const char* name, const char* data, uint32_t size, TVG_UNUSED const string& mimeType, bool copy)
 {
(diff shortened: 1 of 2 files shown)


test/testSavers.cpp
```diff
@@ -100,7 +100,7 @@ TEST_CASE("Save scene into tvg", "[tvgSavers]")
 #endif
 
 
-#ifdef THORVG_GIF_SAVER_SUPPORT
+#if defined(THORVG_GIF_SAVER_SUPPORT) && defined(THORVG_LOTTIE_LOADER_SUPPORT)
```
JSUYA, line 103: Hi nice fix
Could you also edit ./test/capi?

src/renderer/tvgLoader.cpp
```diff
@@ -442,6 +442,7 @@ LoadModule* LoaderMgr::loader(const char* name, const char* data, uint32_t size,
     if (auto loader = _findFromCache(name)) return loader;
 
     //function is dedicated for ttf loader (the only supported font loader)
+#ifdef THORVG_TTF_LOADER_SUPPORT
```
hermet, line 445: let's move this to 441 line.

hermet (approved):

Issue #1991, opened 2024-02-19
Basic steps to compile for linux (wg_beta)


boulabiar:
Hi, 
1. I'm testing thorvg but I have problems activating wg_beta renderer 

`src/renderer/wg_engine/meson.build:11:4: ERROR: Expecting rbracket got string.
    'tvgWgBindGroups.cpp',
    ^
For a block that started at 1,14
source_file = [
              ^`

hermet:
@boulabiar Hello, and thank you for your interest in the ThorVG project. We apologize for any inconvenience. Please note that WebGPU is currently in intensive development and is not stable for use by users or outside collaborators at the moment. We recommend waiting for the official release. We anticipate it after this summer.

boulabiar:
No problem, but I want to test it anyway and test examples.
At least be able to compile wg_beta and examples.

do you know how to get those 2 working on linux ?

hermet:
@boulabiar Currently, examples in thorvg repo would not compatible with wg on Linux (we are mostly testing on Windows with local test code). We can provide any brief information about thorvg webgpu condition and progress. However, guiding people the setup of the dev environment is burdensome for us at this phase.

hermet:
@boulabiar https://github.com/thorvg/thorvg/wiki/WebGPU-Raster-Engine-Development

PR #2419, opened 2024-06-19
examples: replace the efl toolkit with sdl

hermet:
issue: https://github.com/thorvg/thorvg/issues/1244 https://github.com/thorvg/thorvg/issues/1991


capnm:
Just thinking out loud: Could the source code possibly be split into "pure tvg*" and the part related to toolkits split into `./framework/sdl/` `./framework/glfw/` ... ?

hermet:
> Just thinking out loud: Could the source code possibly be split into "pure tvg*" and the part related to toolkits split into `./framework/sdl/` `./framework/glfw/` ... ?

Maybe, if we apply the adapter or bridge, or something similar and if thorvg library has a windows layer over other frameworks dependency as you mentioned, definitely it should have like it.

However, the examples here are for illustration purposes only. I don't aim for a perfect design, as this is not a product or software intended for use. It is just a reference, serves as a direct guide on how to integrate the canvas into Windows and user system main loop. We don't need to introduce more complicated design and code for "our" maintanence, and thorvg never gonna introduce multiple frameworks for examples here. 

Anyway, in this improvement, we have made many enhancements to make it more intuitive and allow users to focus solely on the ThorVG practices. Please have a look at it and thanks your opinion.

examples/Capi.cpp
```diff
@@ -272,115 +259,88 @@ void testCapi()
...
+    tvg_swcanvas_set_target(canvas, (uint32_t*)surface->pixels, surface->w, surface->pitch / 4, surface->h, TVG_COLORSPACE_ARGB8888);
+    tvg_swcanvas_set_mempool(canvas, TVG_MEMPOOL_POLICY_DEFAULT);
 
+    contents();
+
+    SDL_Event event;
+    auto running = true;
+    auto ptime = SDL_GetTicks();
+    auto elapsed = 0;
+    uint32_t tickCnt = 0;
```
mgrudzinska, line 297: unused variable warning

examples/Example.h
```diff
@@ -0,0 +1,537 @@
...
+    #ifdef _WIN32
+        auto rpath = _fullpath(buf, path, PATH_MAX);
+    #else
+        auto rpath = realpath(path, buf);
+    #endif
+
+        //open directory
+    #ifdef _WIN32
+        WIN32_FIND_DATA fd;
+        HANDLE h = FindFirstFileEx((path + "\\*").c_str(), FindExInfoBasic, &fd, FindExSearchNameMatch, NULL, 0);
```
SergeyLebedkin, line 85:
```
../examples/Example.h:85:42: error: invalid operands of types 'const char*' and 'const char [3]' to binary 'operator+'
   85 |         HANDLE h = FindFirstFileEx((path + "\\*").c_str(), FindExInfoBasic, &fd, FindExSearchNameMatch, NULL, 0);
      |                                     ~~~~ ^ ~~~~~
      |                                     |      |
      |                                     |      const char [3]
      |                                     const char*
```
Should be some thing like?
`HANDLE h = FindFirstFileEx((std::string(path) + "\\*").c_str(), FindExInfoBasic, &fd, FindExSearchNameMatch, NULL, 0);`

examples/CustomTransform.cpp
```diff
@@ -20,203 +20,93 @@
...
+
+        //scale y
+        m.e22 = 1 + (progress * 2.0f);
+
+        //rotation
+        constexpr auto PI = 3.141592f;
+        auto degree = 45.0f;
+        auto radian = degree / 180.0f * PI;
+        auto cosVal = cosf(radian);
+        auto sinVal = sinf(radian);
```
SergeyLebedkin, line 76: Windows: missed `#include <cmath>`

capnm:
> ... Anyway, in this improvement, we have made many enhancements to make it more intuitive and allow users to focus solely on the ThorVG practices. Please have a look at it and thanks your opinion.

Would sdl also work with the wg_engine? I have only tried the glfw successfully. Yes, this has to serve as a quick demonstration of how it works for "someone" new to tvg. We did some AI lessons recently and my experiment with the current AI tools questioned for a simple ThorVG example resulted in completely incorrect and unusable code, thus showing up as an AI failure. Probably the API docs? are unreadable for the robots ;-)

hermet:
> Would sdl also work with the wg_engine?

Yes, please check this guide: https://github.com/thorvg/thorvg/wiki/WebGPU-Raster-Engine-Development

examples/Example.h
```diff
@@ -0,0 +1,538 @@
...
+#include <cmath>
+#include <vector>
+#include <fstream>
+#include <iostream>
+#include <thread>
+#include <thorvg.h>
+#include <SDL.h>
+#include <SDL_syswm.h>
+#ifdef _WIN32
+    #define WIN32_LEAN_AND_MEAN
```
SergeyLebedkin, line 33: WIN32_LEAN_AND_MEAN redefined in config.h

PR #2420, opened 2024-06-19
swift: Config for Swift Package build

andyf-canva:
Recently we have opened a new ThorVG for Swift repo: [thorvg.swift](https://github.com/thorvg/thorvg.swift).

Code will be migrated from a Swift fork [thorvg-swift](https://github.com/andyf-canva/thorvg-swift). You can see more details here at [the GitHub issue](https://github.com/thorvg/thorvg.swift/issues/1) on the repo.

In order to build the Swift Package, with the thorvg core as a submodule inside the thorvg.swift repo, the following changes need to take place for the upstream thorvg repo. These are what are currently allowing the thorvg-swift repo to build.

## Config header file

When building the thorvg source using the `meson` build system, we spit out a [config.h](https://github.com/andyf-canva/thorvg-swift/blob/22bbf9dc385d9f788c4c11085f0a786f089d5701/src/bindings/capi/config.h) inside the output `builddir/` directory. Many of the C++ source files refer to this at compile time. 

Swift Package Manager doesn't follow the same paradigm, and requires all files to be present in the target directory at compile time. Thus, we need to generate a `config.h` file ahead of time so that SPM can see that header file and successfully build all of the C++ source files.

For time being, I've put this underneath the C API bindings folder, but am open to another destination for this file.


andyf-canva:
@hermet @tinyjin I've now marked this PR as ready to review as @tinyjin and I have been collaborating on the Swift API as you can see [here](https://github.com/thorvg/thorvg.swift/pull/2#discussion_r1686249547). And we're getting close to merging that PR! 🎉 

For this PR, these are the minimal requirements to get the Swift Package building. As mentioned, where I've added the symbolic link and the config file may not be the best place to put them so would love your feedback as code owners of this repo 😄

hermet (commented):
@andyf-canva Thanks! Quick comment here. I think ThorVG doesn't need changes for this. The best approach would be to generate the `config.h` during the build & packaging thorvg swift with the latest throvg release. Or, at the very least, this change should be included `thorvg.swift` side.

andyf-canva:
@hermet thanks for the feedback ❤️ That's a fair call. From memory, I wasn't able to perform a sort of "build step" from the SPM side, but let me take a look again. Because I agree, seems like something that should just happen at build time. Will report back soon. 👍

andyf-canva:
Good news - I've removed the symbolic link because we can point the Swift Package path to a layer higher to include the `inc` directory. I'll also push the changes to the `thorvg.swift` PR.

Last thing to figure out: the `config.h` file. Will report back and see if I can hack together a pre-build script or something.

andyf-canva:
An update: I've removed the `config.h` file and have created a pre-build script that consumers have to run in-order to copy the config file into the correct position to compile the Package. I will add more details in the PR [here](https://github.com/thorvg/thorvg.swift/pull/2).

andyf-canva:
That means that this PR is now not making any changes on the `thorvg` repo and can thus be closed. 👍

PR #2422, opened 2024-06-19
wg_engine: fix incorrect geomatry

SergeyLebedkin:
Fix computation of segments count for cubic curves. 
Using screen coordinates of base points, instead of world coordinates.

Before/After
image

### DIFF
--- a/src/renderer/wg_engine/tvgWgGeometry.cpp
+++ b/src/renderer/wg_engine/tvgWgGeometry.cpp
@@ -87,13 +87,14 @@ void WgPolyline::appendPoint(WgPoint pt)
 }
 
 
-void WgPolyline::appendCubic(WgPoint p1, WgPoint p2, WgPoint p3)
+void WgPolyline::appendCubic(WgPoint p1, WgPoint p2, WgPoint p3, size_t nsegs)
 {
     WgPoint p0 = pts.count > 0 ? pts.last() : WgPoint(0.0f, 0.0f);
-    size_t segs = ((uint32_t)(p0.dist(p3) / 8.0f));
-    segs = segs == 0 ? 1 : segs;
-    for (size_t i = 1; i <= segs; i++) {
-        float t = i / (float)segs;
+    nsegs = nsegs == 0 ? 1 : nsegs;
+    float dt = 1.0f / (float)nsegs;
+    float t = 0.0f;
+    for (size_t i = 1; i <= nsegs; i++) {
+        t += dt;
         // get cubic spline interpolation coefficients
         float t0 = 1.0f * (1.0f - t) * (1.0f - t) * (1.0f - t);
         float t1 = 3.0f * (1.0f - t) * (1.0f - t) * t;
--- a/src/renderer/wg_engine/tvgWgGeometry.h
+++ b/src/renderer/wg_engine/tvgWgGeometry.h
@@ -57,6 +57,7 @@ public:
     inline void normalize() { float rlen = 1.0f / length(); x *= rlen; y *= rlen; }
     inline WgPoint normal() const { float rlen = 1.0f / length(); return { x * rlen, y * rlen }; }
     inline WgPoint lerp(const WgPoint& p, float t) const { return { x + (p.x - x) * t, y + (p.y - y) * t }; };
+    inline WgPoint trans(const Matrix& m) const { return { x * m.e11 + y * m.e12, x * m.e21 + y * m.e22 }; };
 };
 
 struct WgMath
@@ -85,7 +86,7 @@ struct WgPolyline
     WgPolyline();
 
     void appendPoint(WgPoint pt);
-    void appendCubic(WgPoint p1, WgPoint p2, WgPoint p3);
+    void appendCubic(WgPoint p1, WgPoint p2, WgPoint p3, size_t nsegs = 16);
 
     void trim(WgPolyline* polyline, float trimBegin, float trimEnd) const;
 
--- a/src/renderer/wg_engine/tvgWgRenderData.cpp
+++ b/src/renderer/wg_engine/tvgWgRenderData.cpp
@@ -275,7 +275,7 @@ void WgRenderDataShape::updateBBox(WgPoint pmin, WgPoint pmax)
 }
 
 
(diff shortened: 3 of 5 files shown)


src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -57,6 +57,7 @@ class WgPoint
     inline void normalize() { float rlen = 1.0f / length(); x *= rlen; y *= rlen; }
     inline WgPoint normal() const { float rlen = 1.0f / length(); return { x * rlen, y * rlen }; }
     inline WgPoint lerp(const WgPoint& p, float t) const { return { x + (p.x - x) * t, y + (p.y - y) * t }; };
+    inline WgPoint trans(const Matrix& m) const { return { x * m.e11 + y * m.e12, x * m.e21 + y * m.e22 }; };
 };
```
mgrudzinska, line 61: m.e13 and m.e23 (translation) is not important?

SergeyLebedkin: in a case of 2d vector - no

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -87,13 +87,12 @@ void WgPolyline::appendPoint(WgPoint pt)
 }
 
 
-void WgPolyline::appendCubic(WgPoint p1, WgPoint p2, WgPoint p3)
+void WgPolyline::appendCubic(WgPoint p1, WgPoint p2, WgPoint p3, size_t nsegs)
 {
     WgPoint p0 = pts.count > 0 ? pts.last() : WgPoint(0.0f, 0.0f);
```
mgrudzinska, line 92: shouldn't this be handled as an error? starting from 0,0 most probably won't a correct, right?

SergeyLebedkin: That`s a good idia. At least we can handle it as assert

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -87,13 +87,12 @@ void WgPolyline::appendPoint(WgPoint pt)
...
-    size_t segs = ((uint32_t)(p0.dist(p3) / 8.0f));
-    segs = segs == 0 ? 1 : segs;
-    for (size_t i = 1; i <= segs; i++) {
-        float t = i / (float)segs;
+    nsegs = nsegs == 0 ? 1 : nsegs;
+    for (size_t i = 1; i <= nsegs; i++) {
+        float t = i / (float)nsegs;
```
mgrudzinska, line 95: to avoid repeated division I'd introduce dt = 1/nsegs outside the `for` loop and inside `t += dt` would be enough

SergeyLebedkin: make a sence

src/renderer/wg_engine/tvgWgRenderData.cpp
```diff
@@ -298,10 +298,14 @@ void WgRenderDataShape::updateMeshes(WgContext &context, const RenderShape &rsha
         } else if (cmd == PathCommand::Close) {
             polyline.close();
         } else if (cmd == PathCommand::CubicTo) {
+            WgPoint p0 = polyline.pts.last();
+            WgPoint p3 = rshape.path.pts[pntIndex + 2];
+            size_t nsegs = (uint32_t)(p0.trans(rt->m).dist(p3.trans(rt->m)));
```
mgrudzinska, line 303: I may not understand something, so please correct me
in a case, when `p0` and `p3` are so close to each other, that `dist < 1`, but `p1` and `p2` are quite far away:
Zrzut ekranu 2024-06-19 o 18 22 09
will the result still be as we expect it to be?

SergeyLebedkin: Did you have some proposal about this? How to estimate total length of spline? As I undestand there is no general solution about it, but we can play around with p0, p1, p2 and p3 (green lines). But as I can see distance between p0 and p3 is more than OK

mgrudzinska: there is no formula, we can only estimate. I'm just asking whether your estimation is good enough and won't give wrong results in such a case, when distance between p0 and p3 is very small. the first approach in estimating the bezier curve length is the sum of distances between all four points p0-p3

SergeyLebedkin: Thats easy to implement. Lets try

mgrudzinska:
didn't check all the examples, but this solves AnimateMasking.cpp

src/renderer/wg_engine/tvgWgRenderData.cpp
```diff
@@ -298,10 +298,19 @@ void WgRenderDataShape::updateMeshes(WgContext &context, const RenderShape &rsha
         } else if (cmd == PathCommand::Close) {
             polyline.close();
         } else if (cmd == PathCommand::CubicTo) {
+            WgPoint pt0 = polyline.pts.last().trans(rt->m);
```
mgrudzinska, line 301: are you sure at this point that calling last() will not crash? you do this check later in appendCubic, but maybe it should be done here?

src/renderer/wg_engine/tvgWgRenderData.cpp
```diff
@@ -298,10 +298,20 @@ void WgRenderDataShape::updateMeshes(WgContext &context, const RenderShape &rsha
         } else if (cmd == PathCommand::Close) {
             polyline.close();
         } else if (cmd == PathCommand::CubicTo) {
+            assert(polyline.pts.count > 0);
+            WgPoint pt0 = polyline.pts.last().trans(rt->m);
+            WgPoint pt3 = WgPoint(rshape.path.pts[pntIndex + 2]).trans(rt->m);
+            size_t nsegs = (uint32_t)(pt0.dist(pt3));
+            if (nsegs <= 1) {
```
mgrudzinska, line 305: I'd use the approach with using the length p0-p4 in all cases. just in cxase

SergeyLebedkin: Fixed

PR #2431, opened 2024-06-20
tests: resolve warnings

mgrudzinska:


test/capi/capiText.cpp
```diff
@@ -69,7 +69,7 @@ TEST_CASE("Load/unload TTF file from a memory", "[capiText]")
     fseek(file, 0, SEEK_SET);
     char* data = (char*)malloc(data_size);
     REQUIRE(data);
-    REQUIRE(fread(data, 1, data_size, file) == data_size);
```
hermet, line 72: why don't data type casting?

hermet (approved):

test/capi/capiText.cpp
mgrudzinska: the only reason is because in all other places we use this kind of check (>0)

hermet (approved):

PR #2433, opened 2024-06-20
wg_engine: fix stroke miterlimit

SergeyLebedkin:
Change the appliense of stroke mitter corresponding to SVG spec.
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit

See **StrokeMiterlimit** and **Stroke** examples

Before/After
image

### DIFF
--- a/src/renderer/wg_engine/tvgWgGeometry.cpp
+++ b/src/renderer/wg_engine/tvgWgGeometry.cpp
@@ -154,7 +154,7 @@ void WgPolyline::close()
 
 bool WgPolyline::isClosed() const
 {
-    return (pts.count >= 2) && (pts[0].dist2(pts.last()) == 0.0f);
+    return (pts.count >= 2) && (mathZero(pts[0].dist2(pts.last())));
 }
 
 
@@ -351,11 +351,41 @@ void WgGeometryData::appendStrokeDashed(const WgPolyline* polyline, const Render
 }
 
 
+void WgGeometryData::appendStrokeJoin(const WgPoint& v0, const WgPoint& v1, const WgPoint& v2, StrokeJoin join, float halfWidth, float miterLimit)
+{
+    WgPoint dir0 = (v1 - v0).normal();
+    WgPoint dir1 = (v2 - v1).normal();
+    WgPoint nrm0 { +dir0.y, -dir0.x };
+    WgPoint nrm1 { +dir1.y, -dir1.x };
+    WgPoint offset0 = nrm0 * halfWidth;
+    WgPoint offset1 = nrm1 * halfWidth;
+    if (join == StrokeJoin::Round) {
+        appendCircle(v1, halfWidth);
+    } else if (join == StrokeJoin::Bevel) {
+        appendRect(v1 - offset0, v1 + offset0, v1 - offset1, v1 + offset1);
+    } else if (join == StrokeJoin::Miter) {
+        WgPoint nrm = (nrm0 + nrm1);
+        if (!mathZero(dir0.x * dir1.y -  dir0.y * dir1.x)) {
+            nrm.normalize();
+            float cosine = nrm.dot(nrm0);
+            float angle = std::acos(dir0.dot(dir1.negative()));
+            float miterRatio = 1.0f / (std::sin(angle) * 0.5);
+            if (miterRatio <= miterLimit) {
+                appendRect(v1 + nrm * (halfWidth / cosine), v1 + offset0, v1 + offset1, v1);
+                appendRect(v1 - nrm * (halfWidth / cosine), v1 - offset0, v1 - offset1, v1);
+            } else {
+                appendRect(v1 - offset0, v1 + offset0, v1 - offset1, v1 + offset1);
+            }
+        }
+    }
+}
+
+
 void WgGeometryData::appendStroke(const WgPolyline* polyline, const RenderStroke *stroke)
 {
     assert(stroke);
     assert(polyline);
-    float wdt = stroke->width / 2;
+    float wdt = stroke->width * 0.5f;
 
     // single line sub-path
     if (polyline->pts.count == 2) {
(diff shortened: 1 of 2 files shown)


src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -399,11 +398,13 @@ void WgGeometryData::appendStroke(const WgPolyline* polyline, const RenderStroke
                 } else if (stroke->join == StrokeJoin::Miter) {
                     WgPoint nrm = (nrm0 + nrm1).normal();
                     float cosine = nrm.dot(nrm0);
-                    if ((cosine != 0.0f) && (abs(cosine) != 1.0f) && (abs(wdt / cosine) <= stroke->miterlimit * 2)) {
-                        appendRect(v1 + nrm * (wdt / cosine), v1 + offset0, v1 + offset1, v1 - nrm * (wdt / cosine));
+                    float mitterRation = 1.0f / std::sin((std::acos(dir0.dot(dir1.negative()))/2.0f));
```
mgrudzinska, line 401: miterlength? there is only one 't' and 'ration' is not a word

mgrudzinska: /2 -> *0.5

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -399,11 +398,13 @@ void WgGeometryData::appendStroke(const WgPolyline* polyline, const RenderStroke
...
+                    if ((cosine != 0.0f) && (abs(cosine) < 1.0f) && (mitterRation <= stroke->miterlimit)) {
```
mgrudzinska, line 402: cosine != 0 -> mathZero

since vectors used to calculate cosine are normalized, can cosine be > 1?

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -399,11 +398,13 @@ void WgGeometryData::appendStroke(const WgPolyline* polyline, const RenderStroke
...
+                        appendRect(v1 + nrm * (wdt / cosine), v1 + offset0, v1 + offset1, v1);
+                        appendRect(v1 - nrm * (wdt / cosine), v1 - offset0, v1 - offset1, v1);
                     } else {
                         appendRect(v1 - offset0, v1 + offset0, v1 - offset1, v1 + offset1);
-                    }
+                    }
```
mgrudzinska, line 407: --spaces

src/renderer/wg_engine/tvgWgGeometry.cpp
SergeyLebedkin: It can be equal to 1.0 and its not a case for us

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -458,7 +459,8 @@ void WgGeometryData::appendStroke(const WgPolyline* polyline, const RenderStroke
                     } else if (stroke->join == StrokeJoin::Miter) {
                         WgPoint nrm = (nrm0 + nrm1).normal();
                         float cosine = nrm.dot(nrm0);
-                        if ((cosine != 0.0f) && (abs(cosine) < 1.0f) && (abs(wdt / cosine) <= stroke->miterlimit * 2)) {
+                        float mitterRation = 1.0f / std::sin((std::acos(dir0.dot(dir1.negative()))/2.0f));
```
mgrudzinska, line 462: this is not optimal that the same logic has to be copied in some other place. but this comment is a  bit wider than this pr
but please apply the same commits as above

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -458,7 +459,8 @@ void WgGeometryData::appendStroke(const WgPolyline* polyline, const RenderStroke
                     } else if (stroke->join == StrokeJoin::Miter) {
                         WgPoint nrm = (nrm0 + nrm1).normal();
                         float cosine = nrm.dot(nrm0);
-                        if ((cosine != 0.0f) && (abs(cosine) < 1.0f) && (abs(wdt / cosine) <= stroke->miterlimit * 2)) {
+                        float miterRatio = 1.0f / std::sin((std::acos(dir0.dot(dir1.negative())) * 0.5));
+                        if ((!mathZero(cosine)) && (abs(cosine) < 1.0f) && (miterRatio <= stroke->miterlimit)) {
```
mgrudzinska, line 463: can you please comment on this: `(abs(cosine) < 1.0f)` ? even if it happens that cosine = 1.0001 I don;t see what problem will it cause.

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -456,13 +462,18 @@ void WgGeometryData::appendStroke(const WgPolyline* polyline, const RenderStroke
...
-                            appendRect(v1 - nrm * (wdt / cosine), v1 - offset0, v1 - offset1, v1);
-                        } else {
-                            appendRect(v1 - offset0, v1 + offset0, v1 - offset1, v1 + offset1);
+                        WgPoint nrm = (nrm0 + nrm1);
+                        if(!mathZero(nrm.length())) {
+                            nrm.normalize();
+                            float cosine = nrm.dot(nrm0);
+                            float angle = std::acos(dir0.dot(dir1.negative()));
+                            float miterRatio = 1.0f / (std::sin(angle) * 0.5);
+                            if ((abs(cosine) < 1.0f) && (miterRatio <= stroke->miterlimit)) {
```
mgrudzinska, line 471: as I said, I don;t think case when cosine is slightly more than zero is a problem. if it is, there is some underlying cause, not the value of cosine itself. and we should take care about the underlaying case, not the effect.
here one problem I see is division by zero, which should not take place, no matter whether we use this value later or not.

mgrudzinska (changes requested):

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -351,11 +351,41 @@ void WgGeometryData::appendStrokeDashed(const WgPolyline* polyline, const Render
...
+    WgPoint nrm1 { +dir1.y, -dir1.x };
+    WgPoint offset0 = nrm0 * wdt;
+    WgPoint offset1 = nrm1 * wdt;
+    if (join == StrokeJoin::Round) {
+        appendCircle(v1, wdt);
+    } else if (join == StrokeJoin::Bevel) {
+        appendRect(v1 - offset0, v1 + offset0, v1 - offset1, v1 + offset1);
+    } else if (join == StrokeJoin::Miter) {
+        WgPoint nrm = (nrm0 + nrm1);
+        if ((!mathZero(nrm.length()) && (!dir0.equal(dir1)))) {
```
mgrudzinska, line 368: slightly cheaper method - "2d cross product". we check the directions, ignore the orientation:
`if (!mathZero(dir0.x * dir1.y -  dir0.y * dir1.x)) {`

mgrudzinska (commented):
please check last small comments

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -351,11 +351,41 @@ void WgGeometryData::appendStrokeDashed(const WgPolyline* polyline, const Render
 }
 
 
+void WgGeometryData::appendStrokeJoint(WgPoint& v0, WgPoint& v1, WgPoint& v2, StrokeJoin join, float wdt, float limit)
```
mgrudzinska, line 354: `appendStrokeJoin`

src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -51,9 +51,11 @@ class WgPoint
     WgPoint negative() const { return {-x, -y}; }
     inline void negate() { x = -x; y = -y; }
     inline float length() const { return sqrt(x*x + y*y); }
+    inline float length2() const { return x*x + y*y; }
```
mgrudzinska, line 54: remove?

src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -51,9 +51,11 @@ class WgPoint
...
     inline float dot(const WgPoint& p) const { return x * p.x + y * p.y; }
     inline float dist(const WgPoint& p) const { return sqrt(dist2(p)); }
     inline float dist2(const WgPoint& p) const { return ((p.x - x)*(p.x - x) + (p.y - y)*(p.y - y)); }
+    inline bool equal(const WgPoint& p) const { return mathEqual(x, p.x) && mathEqual(y, p.y); }
```
mgrudzinska, line 58: remove?

mgrudzinska (approved):

src/renderer/wg_engine/tvgWgGeometry.h
mgrudzinska: we'll unify all points at some point, so we can keep them for now

src/renderer/wg_engine/tvgWgGeometry.h
mgrudzinska: we'll unify all points at some point, so we can keep them for now

hermet:
@SergeyLebedkin @mgrudzinska Hello, StrokeMiterLimit works for you? I got a crash.

temp

SergeyLebedkin:
> @SergeyLebedkin @mgrudzinska Hello, StrokeMiterLimit works for you? I got a crash.
 It works on wsl and windows

mgrudzinska:
@hermet @SergeyLebedkin it crashes, but the problem is `stroke-miterlimit.png` not the miterlimit alg, so not connected with this pr

PR #2440, opened 2024-06-21
examples: Support window resizable

JSUYA:
Add SDL_WINDOW_RESIZABLE to the sdl window property and call canvas->update() in the resize event.

Afterwards, if the canvas needs to be expanded by the window size by passing the size as an argument to the update(canvas,elapsed) interface, it can be processed individually.

### DIFF
--- a/examples/Example.h
+++ b/examples/Example.h
@@ -268,7 +268,7 @@ struct SwWindow : Window
 
     SwWindow(Example* example, uint32_t width, uint32_t height) : Window(tvg::CanvasEngine::Sw, example, width, height)
     {
-        window = SDL_CreateWindow("ThorVG Example (Software)", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, SDL_WINDOW_HIDDEN);
+        window = SDL_CreateWindow("ThorVG Example (Software)", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE);
     }
 
     bool ready() override
@@ -317,7 +317,7 @@ struct GlWindow : Window
 
     GlWindow(Example* example, uint32_t width, uint32_t height) : Window(tvg::CanvasEngine::Gl, example, width, height)
     {
-        window = SDL_CreateWindow("ThorVG Example (OpenGL)", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN);
+        window = SDL_CreateWindow("ThorVG Example (OpenGL)", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE);
         context = SDL_GL_CreateContext(window);
     }
 


examples/Example.h
```diff
@@ -229,10 +234,13 @@ struct Window
...
                 needResize = false;
+                if (example->update(canvas, elapsed)) {
+                    needDraw = true;
+                }
             }
 
-            if (tickCnt > 0) {
-                if (example->update(canvas, elapsed)) {
+            if (example->isAnimation && tickCnt > 0) {
+                if (!needDraw && example->update(canvas, elapsed)) {
```
hermet, line 243: I think needDraw itself works for resizing as well, isAnimation flag doesn't need.

hermet (changes requested):

examples/Example.h
JSUYA: The example that does not need update per frame did not need to call example->update(,) in every loop, and I added isAnimation to distinguish it. I thought this was a way to make the code less complicated.

My intention(?) is... for check detailed small changes, In some examples needs when the window size is resized the result resized to fit the window size.

JSUYA: > Only the flag can be added here.

If canvas->update() is not called in no animation example, a crash will occur when the window size becomes smaller than the canvas.

hermet: > The example that does not need update per frame

`needDraw` doing this.

hermet: > > Only the flag can be added here.
> 
> If canvas->update() is not called in no animation example, a crash will occur when the window size becomes smaller than the canvas.

It might be related to an SW engine problem that hasn't been addressed before. I have already observed it and decided to address it correctly later.

hermet: @JSUYA Sorry, i might need to review this again, probably, i confused with 'tickCnt' by my dead brain.

hermet: > If canvas->update() is not called in no animation example,

@JSUYA just reviewed code again, `canvas->update()` is always called, so what is your problem?

JSUYA: 1. If the log option is turned on, unnecessary logs are generated.
2. Of course, redraw does not occur because RenderUpdateFlag does not change internally, but I don't think there is a need to travel all paints's update(). I thought it was an unnecessary action for examples that did not implement `Example::update( , )`.

hermet: > 1. If the log option is turned on, unnecessary logs are generated.

Sorry i don't understand the context why suddenly "log" come from in this PR.


> 2\. Of course, redraw does not occur because RenderUpdateFlag does not change internally, but I don't think there is a need to travel all paints's update(). I thought it was an unnecessary action for examples that did not implement `Example::update( , )`.

 Example::update() is an interface to be triggered every frame. Each examples could implement it when they want to update the scene by their demands.

JSUYA: I updated PR!!
The real problem is that I tried to call canvas->update() in Example::update(, ). Because of this, I had to distinguish between animated examples.

+) Regarding the mention of log,,,, for the above reason, when running the no animated example, the log of update() was called every loop. But it's okay now.

examples/Example.h
```diff
@@ -317,7 +318,7 @@ struct GlWindow : Window
 
     GlWindow(Example* example, uint32_t width, uint32_t height) : Window(tvg::CanvasEngine::Gl, example, width, height)
     {
-        window = SDL_CreateWindow("ThorVG Example (OpenGL)", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN);
+        window = SDL_CreateWindow("ThorVG Example (OpenGL)", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE);
```
JSUYA, line 321: When I resize the window, the result differently than expected.
I think this is a bug and should be fixed in the future, so I added a flag.

hermet (approved):

PR #2445, opened 2024-06-22
gl_engine: Fix compile error if only has OpenGL library on MacOS

RuiwenTang:
This PR make the GL backend can be compile and running when link with OpenGL on MacOS. Since GLES is only available in iOS.

BTW:
The Windows platform may have the same problem.

### DIFF
--- a/examples/Example.h
+++ b/examples/Example.h
@@ -317,7 +317,19 @@ struct GlWindow : Window
 
     GlWindow(Example* example, uint32_t width, uint32_t height) : Window(tvg::CanvasEngine::Gl, example, width, height)
     {
-        window = SDL_CreateWindow("ThorVG Example (OpenGL)", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE);
+        SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
+        SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
+        SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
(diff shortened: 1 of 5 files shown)


src/renderer/gl_engine/meson.build
```diff
@@ -36,15 +36,26 @@ if not gles_dep.found()
...
+link_opengl = false
+
 if not gles_dep.found()
   gles_dep = dependency('GL')
+  link_opengl = true
 endif
 
 if not gles_dep.found()
   gles_dep = meson.get_compiler('cpp').find_library('GL')
+  link_opengl
```
hermet, line 48: `link_opengl = true?`

src/renderer/gl_engine/meson.build
```diff
@@ -36,15 +36,26 @@ if not gles_dep.found()
...
+endif
+
+gl_target_define = '-DTHORVG_GL_TARGET_GLES=1'
```
hermet, line 51:
```
if link_opengl
    gl_target_define = '-DTHORVG_GL_TARGET_GL=1'
else
   gl_target_define = '-DTHORVG_GL_TARGET_GLES=1'
endif
```

examples/Example.h
```diff
@@ -41,6 +41,10 @@
     #include <sys/stat.h>
 #endif
 
+#ifdef THORVG_GL_RASTER_SUPPORT
+#include <SDL_opengl.h>
```
hermet, line 45: this looks unnecessary here?

src/renderer/gl_engine/tvgGlCommon.h
```diff
@@ -24,7 +24,15 @@
 #define _TVG_GL_COMMON_H_
 
 #include <assert.h>
+#if defined (THORVG_GL_TARGET_GLES)
 #include <GLES3/gl3.h>
+#elif defined (THORVG_GL_TARGET_GL)
```
hermet, line 29: please consider using indentation:

example:
```
//for MSVC Compat
#ifdef _MSC_VER
    #define TVG_UNUSED
    #define strncasecmp _strnicmp
    #define strcasecmp _stricmp
#else
    #define TVG_UNUSED __attribute__ ((__unused__))
#endif

// Portable 'fallthrough' attribute
#if __has_cpp_attribute(fallthrough)
    #ifdef _MSC_VER
        #define TVG_FALLTHROUGH [[fallthrough]];
    #else
        #define TVG_FALLTHROUGH __attribute__ ((fallthrough));
    #endif
#else
    #define TVG_FALLTHROUGH
#endif

#if defined(_MSC_VER) && defined(__clang__)
    #define strncpy strncpy_s
    #define strdup _strdup
#endif
```

hermet (approved):
please check some minor comments. thanks.

PR #2447, opened 2024-06-23
examples: fix crash in window creation

tinyjin:
SDL flag combination(SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE) causes crash.

CleanShot 2024-06-23 at 13 55 52@2x

### DIFF
--- a/examples/Example.h
+++ b/examples/Example.h
@@ -268,7 +268,7 @@ struct SwWindow : Window
 
     SwWindow(Example* example, uint32_t width, uint32_t height) : Window(tvg::CanvasEngine::Sw, example, width, height)
     {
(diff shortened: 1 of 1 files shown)


hermet (changes requested):
This doesn't make sense to me. The root-cause(thorvg) should be fixed, not example.
Could you please check with this https://github.com/thorvg/thorvg/pull/2441  ?

hermet:
@tinyjin hello, could you please confirm the `surface->pixel`, `surface->w`, `surface->h`, `surface->pitch` values?

image

JSUYA (commented):
Regardless of whether it crashed or not,,,, the reason the window was displayed even if it was HIDDEN is because SDL_ShowWindow() is called.
If you want to change HIDDEN to SHOWN, SDL_ShowWindow() is unnecessary.

mgrudzinska:
@tinyjin please check https://github.com/thorvg/thorvg/pull/2466 whether this solves your case

tinyjin:
@mgrudzinska #2466 with macOS doesn't happen runtime crash in example. Thanks.

PR #2450, opened 2024-06-23
doc: removed useless information for easier maintainence

hermet:
That basic information is addressed in the Results section.
We can only add information that pertains to a specific scenario.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -72,6 +72,10 @@ class Animation;
 
 /**
  * @brief Enumeration specifying the result from the APIs.
+ *
+ * All ThorVG APIs could potentially return one of the values in the list.
+ * Please note that some APIs may additionally specify the reasons that trigger their return values.
+ *
  */
 enum class Result
 {
@@ -278,8 +282,6 @@ public:
      * The rotational axis passes through the point on the object with zero coordinates.
      *
      * @param[in] degree The value of the angle in degrees.
-     *
-     * @retval Result::Success when succeed, Result::FailedAllocation otherwise.
      */
     Result rotate(float degree) noexcept;
 
@@ -287,8 +289,6 @@ public:
      * @brief Sets the scale value of the object.
      *
      * @param[in] factor The value of the scaling factor. The default value is 1.
-     *
-     * @retval Result::Success when succeed, Result::FailedAllocation otherwise.
      */
     Result scale(float factor) noexcept;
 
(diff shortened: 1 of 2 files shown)


mgrudzinska:
Zrzut ekranu 2024-06-24 o 11 20 34

@hermet isn't this a bit confusing and looks like the fun always produces some issue?

hermet:
> Zrzut ekranu 2024-06-24 o 11 20 34
> @hermet isn't this a bit confusing and looks like the fun always produces some issue?

@mgrudzinska could you please make an issue if it is not related to this PR?

mgrudzinska:
> @mgrudzinska could you please make an issue if it is not related to this PR?

but it is related. I've suggested that if we remove all '@retval Result::Success ...` it will be confusing.
if you don't agree, please resolve my comment, no need to discuss

hermet:
@mgrudzinska Fixing the return values of per API would easily harm the compatibility between the app and the library(You've already made this by changing the retval list of some apis). The reason we have listed the return values in enum, is to indicate that all APIs can potentially return these values. Therefore, users can compare the returned values to identify the type of error. This flexible approach ultimately ensures compatibility and easy to use.

```
enum class Result
{
    Success = 0,           ///< The value returned in case of a correct request execution.
    InvalidArguments,      ///< The value returned in the event of a problem with the arguments given to the API - e.g. empty paths or null pointers.
    InsufficientCondition, ///< The value returned in case the request cannot be processed - e.g. asking for properties of an object, which does not exist.
    FailedAllocation,      ///< The value returned in case of unsuccessful memory allocation.
    MemoryCorruption,      ///< The value returned in the event of bad memory handling - e.g. failing in pointer releasing or casting
    NonSupport,            ///< The value returned in case of choosing unsupported engine features(options).
    Unknown                ///< The value returned in all other cases.
};
```
This is not only about ThorVG's mechanism; many libraries, since the Unix system, follow this approach. Now, we can address the retval only when we can help users understand the reason for the non-success in more detail than the general explanation.

hermet:
> if we remove all '@RetVal Result::Success ...` it will be confusing.

> > @mgrudzinska could you please make an issue if it is not related to this PR?
> 
> but it is related. I've suggested that if we remove all '@RetVal Result::Success ...` it will be confusing. if you don't agree, please resolve my comment, no need to discuss

@mgrudzinska It needs a logic. If you think my logic does not make sense, it is better to discuss it rather than ignore it. Please come up with a logic for what will happen to users if every API doesn't mention: "@RetVal Result::Success." I don't anticipate any confusion there.

inc/thorvg.h
```diff
@@ -72,6 +72,10 @@ class Animation;
 
 /**
  * @brief Enumeration specifying the result from the APIs.
+ *
+ * All ThorVG APIs could potentially return one of the values in the list.
+ * Please note that some APIs may additionally specify the reasons that trigger their return values.
```
hermet, line 77: @mgrudzinska please note this

Issue #2282, opened 2024-05-15
gl_engine: enhance portability with safety


hermet:
GlCanvas::gen() should return nullptr when GL is not available on the current system.
Alternatively, GlCanvas::target(int32_t id, uint32_t w, uint32_t h) should return NonSupported.

hermet:
@SergeyLebedkin Please assign yourself when you start working on this. Thanks.

PR #2451, opened 2024-06-23
gl_engine: revise the gl portability

hermet:
- guarantee minimum version requirement
- removed glesv2 dependency
- corrected gl ver dependency
- removed sdl practices

issue: https://github.com/thorvg/thorvg/issues/2282

### DIFF
--- a/examples/Example.h
+++ b/examples/Example.h
@@ -317,19 +317,16 @@ struct GlWindow : Window
 
     GlWindow(Example* example, uint32_t width, uint32_t height) : Window(tvg::CanvasEngine::Gl, example, width, height)
     {
+#ifdef THORVG_GL_TARGET_GLES
+        SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
-        SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
-        window = SDL_CreateWindow("ThorVG Example (OpenGLES)", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE);
-
-        if (window == nullptr) {
-            // try create OpenGLES context failed, create OpenGL context instead
-            SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
-            SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 3);
-            SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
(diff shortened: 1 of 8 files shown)


src/renderer/gl_engine/tvgGlCommon.h
```diff
@@ -26,12 +26,17 @@
...
     #include <GLES3/gl3.h>
+    #define TVG_REQUIRE_GL_MAJOR_VER 3
+    #define TVG_REQUIRE_GL_MINOR_VER 0
 #elif defined (THORVG_GL_TARGET_GL)
     #if defined(__APPLE__) || defined(__MACH__)
         #include <OpenGL/gl3.h>
     #else
-        #include <GL/gl3.h>
+        #define GL_GLEXT_PROTOTYPES 1
+        #include <GL/gl.h>
```
hermet, line 36: @RuiwenTang I couldn't find `<GL/gl3.h>` please confirm this, Thanks.

examples/Example.h
```diff
@@ -317,19 +317,7 @@ struct GlWindow : Window
...
-
-        if (window == nullptr) {
-            // try create OpenGLES context failed, create OpenGL context instead
-            SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
-            SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 3);
-            SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
-            window = SDL_CreateWindow("ThorVG Example (OpenGL)", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE);
-        }
-
+        window = SDL_CreateWindow("ThorVG Example (OpenGL/ES)", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE);
```
hermet, line 320: @RuiwenTang Please let me know if this brings up an issue for you. Thanks.

src/renderer/gl_engine/tvgGlCommon.h
RuiwenTang: which platform do you compile to ?

examples/Example.h
RuiwenTang: If not specify GL context version, on MacOS, SDL will create a GL 2.0 context.

src/renderer/gl_engine/tvgGlCommon.h
hermet: linux(ubuntu)

RuiwenTang: I will check this on my Ubuntu environment.

BTW:
Maybe we should consider dynamically loading GL functions, which would avoid most compilation and linking issues

examples/Example.h
hermet: @RuiwenTang Please double-check it again. The problem might be with the profile (OpenGL ES), not the version. In my case, it's 4.1 by default. I believe this should work.

RuiwenTang: `4.1` means target on OpenGL context, highest version in GLES is 3.2. And I think the `SDL_GL_CONTEXT_PROFILE_CORE` only valid when target on OpenGL.

hermet: @RuiwenTang the main point is it. SDL would use the higher version based on the designated profile. And gl engine should work on it.

RuiwenTang: yes 330 core is the minimum version GLEngine required. It works in higher version

PR #2454, opened 2024-06-23
wg_engine: fix line joins

mgrudzinska:
Shapes were incorrectly closed in certain cases -
the decision to close a shape or not should be based on path commands rather than the number of points and their distances from each other.

before:
Zrzut ekranu 2024-06-24 o 00 27 56

after - expected result is that the second triangle is not closed:
Zrzut ekranu 2024-06-24 o 00 34 57

sample:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 500 500" >
    <path d="M 20 20 h 300 l -150 200 l -150 -200 Z" stroke="blue" stroke-width="30" fill="none" stroke-linejoin="round" opacity="0.5"/>
    <path d="M 20 270 h 300 l -150 200 l -150 -200" stroke="red" stroke-width="30" fill="none" stroke-linejoin="round" opacity="0.5"/>
</svg>
```

### DIFF
--- a/src/renderer/wg_engine/tvgWgGeometry.cpp
+++ b/src/renderer/wg_engine/tvgWgGeometry.cpp
@@ -57,6 +57,7 @@ WgPolyline::WgPolyline()
     constexpr uint32_t nPoints = 360;
     pts.reserve(nPoints);
     dist.reserve(nPoints);
+    closed = false;
 }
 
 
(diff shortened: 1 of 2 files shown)


hermet:
@SergeyLebedkin ping?

SergeyLebedkin (approved):

SergeyLebedkin:
Did not break anything related to rendering approaches, including shapes.

capnm:
godot + v0.14.0 
image

mgrudzinska:
> godot + v0.14.0

@capnm are you confirming that it works as expected, right? because this is the expected result

capnm:
> > godot + v0.14.0
> 
> @capnm are you confirming that it works as expected, right? because this is the expected result

Yes! I am looking for test improvement images for the PR Godot ++ v0.14.0.

Issue #2461, opened 2024-06-24
gl_engine: memory leaks


hermet:
**Build Configure**
`meson setup builddir -Db_sanitize="address, undefined" -Dloaders=all -Dengines=all`

**ClipPath, Lottie, PictureRaw, SVG**
```
Direct leak of 1440 byte(s) in 18 object(s) allocated from:
    #0 0x7f0a70c531e7 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:99
    #1 0x7f0a700575b5 in GlRenderer::drawClip(tvg::Array<void*>&) ../src/renderer/gl_engine/tvgGlRenderer.cpp:844
    #2 0x7f0a70059e83 in GlRenderer::renderShape(void*) ../src/renderer/gl_engine/tvgGlRenderer.cpp:294
    #3 0x7f0a6ffa7152 in tvg::Shape::Impl::render(tvg::RenderMethod*) ../src/renderer/tvgShape.h:60
    #4 0x7f0a6ff9e8e4 in tvg::Paint::Impl::render(tvg::RenderMethod*) ../src/renderer/tvgPaint.cpp:250
    #5 0x7f0a6ffac18e in tvg::Scene::Impl::render(tvg::RenderMethod*) ../src/renderer/tvgScene.h:139
    #6 0x7f0a6ff9eaab in tvg::Paint::Impl::render(tvg::RenderMethod*) ../src/renderer/tvgPaint.cpp:250
    #7 0x7f0a6ffac18e in tvg::Scene::Impl::render(tvg::RenderMethod*) ../src/renderer/tvgScene.h:139
    #8 0x7f0a6ff9eaab in tvg::Paint::Impl::render(tvg::RenderMethod*) ../src/renderer/tvgPaint.cpp:250
    #9 0x7f0a6ffac18e in tvg::Scene::Impl::render(tvg::RenderMethod*) ../src/renderer/tvgScene.h:139
    #10 0x7f0a6ff9eaab in tvg::Paint::Impl::render(tvg::RenderMethod*) ../src/renderer/tvgPaint.cpp:250
    #11 0x7f0a6ffb288b in tvg::Picture::Impl::render(tvg::RenderMethod*) ../src/renderer/tvgPicture.cpp:83
    #12 0x7f0a6ff9ec72 in tvg::Paint::Impl::render(tvg::RenderMethod*) ../src/renderer/tvgPaint.cpp:250
[...]

PR #2464, opened 2024-06-24
gl_engine: Fix memory leak caused by no deleted GlRenderTask

RuiwenTang:
This PR Fix memory leak cuased ClipTask and empty render task

#2461


hermet (approved):
perfect, thanks!

PR #2466, opened 2024-06-25
examples: fix crash

mgrudzinska:
If the initial window size does not fit on the screen, a resize event is triggered. This occurs at the very beginning after pushing to the canvas, at a stage when the buffer should not be accessed and changing the target is not possible. To prevent this situation, resizing is paused until rendering to the initial target is completed.

### DIFF
--- a/examples/Example.h
+++ b/examples/Example.h
@@ -227,6 +227,7 @@ struct Window
             }
 
             if (needResize) {
+                if (tickCnt == 0) draw();
                 resize();
                 needResize = false;
             }


hermet (commented):
@mgrudzinska Could you please share how these values change during resize() calls?.
`surface->pixels pointer`, `surface->w`, `surface->h`

hermet:
@mgrudzinska could you please check with this fix? https://github.com/thorvg/thorvg/pull/2469

mgrudzinska:
yes, I'll check, this is one not the correct way of solving this problem, so I'll close it

PR #2469, opened 2024-06-25
example: synchronize the canvas before the window pops up.

hermet:
This ensures that the window pops up with the appropriate content, since windows can be popped up asynchronously with mainloop.


examples/Example.h
```diff
@@ -439,8 +430,12 @@ struct WgWindow : Window
         Window::canvas = canvas.get();
```
mgrudzinska, line 430: l.427 - constructor returns false

hermet:
@mgrudzinska @tinyjin please confirm if this fixes your problems. thanks.

tinyjin:
@hermet This fixes crash in my macOS
CleanShot 2024-06-25 at 21 54 21@2x

mgrudzinska:
> @mgrudzinska @tinyjin please confirm if this fixes your problems. thanks.

yes, it does.

tinyjin (approved):

PR #2470, opened 2024-06-25
wg_engine: force texture data writing on GPU side

SergeyLebedkin:
Texture must be fully uploaded into GPU memory before we can use or destroy it. This change force texture data to update.
See PicturePng, PictureJpg examples for texture usage


mgrudzinska:
this solves the Lottie example crash as well

hermet (approved):

Issue #2468, opened 2024-06-25
Regression in v0.13.8 for SVG with text node and no embedded font


akien-mga:
Seen in Godot `master` branch after moving from ThorVG 0.13.7 to 0.13.8, the following SVG:

![thanks](https://github.com/thorvg/thorvg/assets/4701338/54acde4d-0cf5-4fcf-a45b-75cd7912048a)
[thanks.zip](https://github.com/user-attachments/files/15968427/thanks.zip)

Is now failing to load. It also fails to render anything on https://thorvg.github.io/thorvg.viewer/

The Godot error is:
```
ERROR: ImageLoaderSVG: Couldn't draw ThorVG pictures on canvas.
   at: create_image_from_utf8_buffer (modules/svg/image_loader_svg.cpp:125)
```
i.e. it's failing on calling `tvg::SwCanvas::draw` here:
https://github.com/godotengine/godot/blob/6b281c0c07b07f2142b1fc8a6b3158091a9b124c/modules/svg/image_loader_svg.cpp#L122-L126

With 0.13.7, it draws ok and looks like this in-engine:

![image](https://github.com/thorvg/thorvg/assets/4701338/8db6a9ca-f76d-4f71-bcb3-b5a71072831c)

So the text was missing, but that's a separate issue, possibly a Godot one. That part isn't a regression, it's the same in ThorVG 0.12.10.

Minimal Godot project with the SVG: [thorvg-svg-embedded-png.zip](https://github.com/user-attachments/files/15968444/thorvg-svg-embedded-png.zip)

mgrudzinska:
@akien-mga can you please elaborate on the regression you are observing?
in v0.13.7 text was not supported in the svg loader.

also we don't support for now `tspan` used in the `thanks.svg` and only embedded fonts are supported (or loaded by the user). 

so there this sample couldn't work in thorvg

akien-mga:
Maybe some background: in Godot, our SVG support for now is limited to rasterizing files to lossless compressed WebP, so each SVG file is imported by the Godot editor to be used as a sprite/texture.

With thorvg 0.13.7, this file could be imported fine, as in it produces a usable texture and no import errors. As pointed out, the actual texture was missing text, so technically it's not fully functional as a 1:1 conversion of the source asset to a texture. But at least it's not workflow breaking, the user can see that the texture is missing text, and find maybe in documentation that it's indeed not a supported feature.

[...]

PR #2479, opened 2024-06-26
common: fix rendering failure

mgrudzinska:
In case of a problem with loading font data, there is no paint inside a text that can be rendered,
but this should not interrupt the rendering process of other paints. Now fixed.

@Issue: https://github.com/thorvg/thorvg/issues/2468


capnm:
I can confirm that it fixes the Godot error issue and PR is also in the 0.14.x branch:
0f2fe0c4d41a87439149aabc8d076cca77e51137

PR #2481, opened 2024-06-26
examples: fix crash after resize

mgrudzinska:
Changing target (resizing) should enforce canvas update.

### DIFF
--- a/examples/AnimateMasking.cpp
+++ b/examples/AnimateMasking.cpp
@@ -72,7 +72,7 @@ struct UserExample : tvgexam::Example
         return true;
     }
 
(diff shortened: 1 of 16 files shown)


JSUYA (commented):
I completely agree with this method, but draw() is called without update() after resize(target() call / https://github.com/thorvg/thorvg/issues/2484), if the size of the buffer changes, a crash occurs. I think it's best to solve this from an engine perspective first.
https://github.com/thorvg/thorvg/pull/2440#discussion_r1648525334

hermet:
https://github.com/thorvg/thorvg/pull/2485/commits/70711311c43e82fc0f9ecf589c5050cd954dac94

PR #2482, opened 2024-06-27
examples: prevent crash for unsupported engines

mgrudzinska:


hermet:
https://github.com/thorvg/thorvg/pull/2483

PR #2490, opened 2024-06-27
examples: fix surface refresh approuch for webgpu

SergeyLebedkin:
Move surface refresh approuch from engine side to user size to syncronize behavior with gl and sw renderers
Checked on static and animated examples


examples/Example.h
```diff
@@ -436,6 +436,11 @@ struct WgWindow : Window
         //Set the canvas target and draw on it.
         verify(canvas->target(instance, surface, width, height));
     }
+
+    void refresh() override 
+    {
+        wgpuSurfacePresent(surface);
```
hermet, line 442: @SergeyLebedkin It looks like this is closer to an engine change. `sync()` seems more appropriate to me since the example sw/gl actually refreshes the SDL window, not their context.

hermet:
> Checked on static and animated examples

@SergeyLebedkin Could you please explain what should be different?

hermet (approved):

Issue #1869, opened 2023-12-28
Support `shape-rendering="crispEdges` in shapes to disable anti-aliasing


OverloadedOrama:
Hello! I am currently developing a pixel art drawing application in Godot, and importing svg images and automatically rasterize them is supported. However, there is no apparent way to disable anti-aliasing in individual shapes. SVGs have a [shape-rendering attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering) that, when set to `crispEdges` does exactly this, but it seems to be unsupported in thorvg. I would greatly appreciate if it was supported.

Example svg:
```svg
<svg width="460" height="200">
  <ellipse cx="210" cy="45" rx="170" ry="15" shape-rendering="crispEdges" style="fill:lime" />
  <ellipse cx="240" cy="100" rx="220" ry="30" style="fill:purple" />
<polyline points="20,20 40,25 60,40 80,120 120,140 200,180"
  style="fill:none;stroke:black;stroke-width:3" shape-rendering="crispEdges" />
<polyline points="60,20 80,25 100,40 120,120 160,140 240,180"
  style="fill:none;stroke:black;stroke-width:3" />
</svg>
```

How it looks as a png exported from ThorVGViewer:
![test(1)](https://github.com/thorvg/thorvg/assets/35376950/c7f5a297-a8cf-46ae-8176-91dbbb0da034)

VS how it's supposed to look (I imported the svg file above into the GNU Image Manipulation Program and exported it as a png):
![test](https://github.com/thorvg/thorvg/assets/35376950/af698e74-ec86-4d76-87f3-bc88d5274951)

In the ThorVGViewer png, all shapes are anti-aliased, while in the GNU Image Manipulation Program png, the polyline on the left and and the green ellipse have aliased crisp edges, and only the polyline on the right and the purple ellipse are anti-aliased. If the differences aren't visible on the browser, try downloading the images and looking at them in an image editing software.

If support for text rendering from svg files gets supported, I would also appreciate if disabling anti-aliasing there would also be supported with [text-rendering](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/text-rendering).

PR #2499, opened 2024-06-30
renderer: support `shape-rendering="crispEdges"` for aliased shapes

OverloadedOrama:
Implements #1869 for the software renderer. If the `shape-rendering="crispEdges"` attribute is used in the SVG, the software renderer does not use anti-aliasing for the shape's lines. The rest of the `shape-rendering` values just default to anti-aliasing, I'm not sure how else to handle them. I tested it with svg2png and Godot and it seems to be working as intended.

Tested using the following svg file:

```svg
<svg width="460" height="200">
  <rect width="140" height="30" x="10" y="10" rx="20" ry="20" style="fill:rgb(0,0,255)" />
  <rect width="140" height="30" x="160" y="10" rx="20" ry="20" style="fill:rgb(0,0,255);stroke-width:1;stroke:red" />
  <rect width="140" height="30" x="310" y="10" rx="20" ry="20" shape-rendering="crispEdges" style="fill:rgb(0,0,255);stroke-width:1;stroke:red" />
  <rect width="140" height="30" x="10" y="60" rx="20" ry="20" fill="none" style="stroke-width:1;stroke:red" />
  <rect width="140" height="30" x="160" y="60" rx="20" ry="20" fill="none" shape-rendering="crispEdges" style="stroke-width:1;stroke:red" />
  <ellipse cx="210" cy="110" rx="170" ry="15" shape-rendering="crispEdges" style="fill:lime" />
  <ellipse cx="210" cy="160" rx="200" ry="30" style="fill:purple" />
  <polyline points="20,20 40,25 60,40 80,120 120,140 200,180"
    style="fill:none;stroke:white;stroke-width:3" shape-rendering="crispEdges" />
  <polyline points="60,20 80,25 100,40 120,120 160,140 240,180"
    style="fill:none;stroke:white;stroke-width:3" />
  <path d="M 10 10 C 20 20, 40 20, 50 10" stroke="black" shape-rendering="crispEdges" fill="transparent"/>
</svg>
```

Result png:
Result png

This is the first time I'm contributing to ThorVG, hopefully I did everything right. 😅

### DIFF
--- a/src/renderer/sw_engine/tvgSwRenderer.cpp
+++ b/src/renderer/sw_engine/tvgSwRenderer.cpp
@@ -85,6 +85,7 @@ struct SwShapeTask : SwTask
        Additionally, the stroke style should not be dashed. */
     bool antialiasing(float strokeWidth)
     {
+        if (rshape->stroke && rshape->stroke->antiAlias == ShapeRendering::CrispEdges) return false;
         return strokeWidth < 2.0f || rshape->stroke->dash.count > 0 || rshape->stroke->strokeFirst || rshape->trimpath() || rshape->stroke->color.a < 255;
     }
 
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -160,6 +160,14 @@ enum class FillRule : uint8_t
     EvenOdd      ///< A line from the point to a location outside the shape is drawn and its intersections with the path segments of the shape are counted. If the number of intersections is an odd number, the point is inside the shape.
 };
 
+/**
+ * @brief Enumeration specifying whether the shapes are anti-aliased or not.
+ */
+enum class ShapeRendering : uint8_t
+{
+    AntiAliased = 0, ///< The shape has anti-aliasing applied to it.
+    CrispEdges      ///< The shape has crisp edges, no anti-aliasing applied to it.
+};
 
 /**
  * @brief Enumeration indicating the method used in the mask of two objects - the target and the source.
@@ -1179,6 +1187,17 @@ public:
     Result fill(FillRule r) noexcept;
 
     /**
+     * @brief Sets the shape rendering method style for the lines of the Shape object.
+     *
+     * The shape rendering method determines whether the lines are anti-aliased or not.
+     *
+     * @param[in] sr The shape rendering method value. The default value is @c ShapeRendering::AntiAliased.
+     *
+     * @retval Result::Success when succeed.
+     */
+    Result shapeRendering(ShapeRendering sr) noexcept;
+
+    /**
      * @brief Sets the rendering order of the stroke and the fill.
      *
      * @param[in] strokeFirst If @c true the stroke is rendered before the fill, otherwise the stroke is rendered as the second one (the default option).
@@ -1228,6 +1247,13 @@ public:
      * @return The fill rule value of the shape.
      */
     FillRule fillRule() const noexcept;
+
+    /**
+     * @brief Gets the shape rendering method value.
+     *
+     * @return The shape rendering method value of the shape.
+     */
+    ShapeRendering shapeRendering() const noexcept;
 
     /**
      * @brief Gets the stroke width.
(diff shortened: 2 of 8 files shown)


hermet:
@OverloadedOrama Hello, Thanks for your contribution. I am curious about the practical use of this feature in user applications. Could you provide some tips on use cases?

hermet:
@OverloadedOrama One more question, do you really need to apply this per shape or scene? or is it applied to the full screen in the ThorVG canvas?

OverloadedOrama:
@hermet Hello! I am working on a pixel art editor made with Godot, and I want to implement support for vector layers in order to have non-destructive shapes, and my idea is to construct a SVG behind the scenes that gets rasterized by Godot/ThorVG in order to be drawn in the canvas. This already can work fine, but all shapes are currently anti-aliased, with no option to turn it off, which doesn't work well with pixel art. So I thought to use the `shape-rendering` SVG attribute to control whether a shape is anti-aliased or not.

I think this needs to be applied per-shape, just like the attribute is per-shape in SVG, in order to allow users to choose which shapes are anti-aliased and which ones are not.

hermet:
@RuiwenTang @SergeyLebedkin Hello, could you please confirm with the GL/WG engine side if they can allow the AA option per shape?

OverloadedOrama:
I'm not sure why unit tests are failing. For some reason, if I uncomment either line 90 or line 91 on `tvgSwShape.cpp`, the unit tests pass fine, but I don't understand why adding line 90 makes them fail.

src/renderer/sw_engine/tvgSwRenderer.cpp
```diff
@@ -87,6 +87,7 @@ struct SwShapeTask : SwTask
        Additionally, the stroke style should not be dashed. */
     bool antialiasing(float strokeWidth)
     {
+        if (rshape->stroke->antiAlias == ShapeRendering::CrispEdges) return false;
```
JSUYA, line 90: > I'm not sure why unit tests are failing. For some reason, if I uncomment either line 90 or line 91 on tvgSwShape.cpp, the unit tests pass fine, but I don't understand why adding line 90 makes them fail.

When strokeWidth is 0, rshape->stroke is null. It looks like a null check for rshape->stroke is needed. 
```suggestion
        if (rshape->stroke && rshape->stroke->antiAlias == ShapeRendering::CrispEdges) return false;
```

hermet:
Some considerations:

- ThorVG can probably support anti-aliasing internally through the SVG loader and renderer, without exposing specific APIs because I don't think AA toggle per shape is a general requirement.
- Because of that reason, I'm not sure whether adding a flag for each shape is a good idea in sake of optimal memory usage, because it would harm the general use-cases, a global option is good though.
- It would be nice if you have a practical demo in production.

OverloadedOrama:
Hello @hermet, apologies for the late reply. I don't have a practical demo ready at the moment, but as I mentioned the reason I need this is because I want to add vector layers in my pixel art software, which is made with Godot, and I'd like to give users the option to disable AA so the shapes can look better for pixel art. If you think that adding a per-shape flag for AA is not needed and it can increase memory usage, then I completely understand and I'd be okay with a more global solution.

What I'd ideally want though is for a way to toggle AA with SVG code. The idea for vector layers is to pass an SVG string and use Godot's [Image.load_svg_from_string](https://docs.godotengine.org/en/stable/classes/class_image.html#class-image-method-load-svg-from-string) to turn SVG into a rasterized image. From what I understand, this uses ThorVG's software renderer behind the scenes. Using the `shape-rendering` tag for each shape is the closest thing I could find to this, and when I compiled Godot with this PR's version of ThorVG, it was working as expected. If there's a better solution for this, I'm of course open to it! If a global option is implemented, perhaps it could be passed as an extra parameter for Image.load_svg_from_string and Image.load_svg_from_buffer, like it happens with the `scale` parameter, but I suppose that needs to be done in Godot's side.

hermet:
Close out old PRs

Issue #2494, opened 2024-06-27
Undocumented error conditions for Canvas methods


dtromb:
When I am calling `Canvas::sync()` under some conditions, it is returning `Result::InsufficientCondition` (equal to 2).

In the API reference (https://www.thorvg.org/apis) , this return code is undocumented, and non-specific, so I don't know what is causing the error or why.   I cannot even to find the place in the code where this is happening, because there are a number of virtual 'pImpl' pointers involved.  

Short of adding descriptive strings to the errors, or more descriptive error codes, could the possible error conditions somehow be documented in the API docs with short explainations of when they occur?

dtromb:
To be more specific here, I am using the software renderer only, and it appears to me as though the SwCanvas impl calls `SwRenderer::sync`, only returning that error code if `sync()` does not return true.   But `sync()` always returns true (tvgSwRenderer.cpp:398), so my analysis must be wrong or I am not understanding the code!)

dtromb:
Looking more closely, it appears that this will happen if `Canvas::sync()` is called twice.   This needs to be documented, as we only have "Guarantees that drawing task is finished" and "must be called" to go on.

Why is it an error to `sync()` twice?  And what is the overall lifecycle of the Canvas?   What changes to its state allow us to draw / sync again?  etc....

hermet:
> Short of adding descriptive strings to the errors, or more descriptive error codes, could the possible error conditions somehow be documented in the API docs with short explainations of when they occur?

@dtromb  hello, as the doc says:
```
     * The Canvas rendering can be performed asynchronously. To make sure that rendering is finished,
     * the sync() must be called after the draw() regardless of threading.
````
if the draw() function has not been successfully conducted before, the return value won't be successful either.

For your information, the canvas operates based on states by default. Each state (repeat: update -> draw -> sync)  can operate based on the previous state, and if the previous state was not successfully reached, it may return information such as `InsufficientCondition`. `InsufficientCondition` does not mean an error; it simply indicates that the requested operation was not performed (or skipped) for some reason. In this case, "Nothing to do sync"

PR #2501, opened 2024-07-01
api: ++description

hermet:
issue: https://github.com/thorvg/thorvg/issues/2494

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -625,6 +625,8 @@ public:
      * The Canvas rendering can be performed asynchronously. To make sure that rendering is finished,
      * the sync() must be called after the draw() regardless of threading.
      *
+     * @retval Result::InsufficientCondition: The canvas is either already in sync condition or in a damaged condition (a draw is required before syncing).
+     *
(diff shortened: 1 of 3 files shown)


inc/thorvg.h
```diff
@@ -625,6 +625,8 @@ class TVG_API Canvas
      * The Canvas rendering can be performed asynchronously. To make sure that rendering is finished,
      * the sync() must be called after the draw() regardless of threading.
      *
+     * @retval Result::InsufficientCondition: The canvas is either already in sync condition or in a damaged condition (a draw is required before syncing).
+     *
      * @see Canvas::draw()
```
mgrudzinska, line 630: I'd write something like this without using 'damaged'
```* @retval Result::InsufficientCondition: The canvas is either already synchronized or in a state not ready for syncing (e.g., requiring a draw before syncing).```

mgrudzinska (approved):
@hermet I added a comment, but it's a matter of individual preference so feel free to ignore it.

mgrudzinska:
I was 3 min too late :D

Issue #2518, opened 2024-07-04
svg: Graphic elements declared in <defs> are displayed.


JSUYA:
```
<svg width="192px" height="192px" viewBox="0 0 192 192" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <defs id="def">
        <linearGradient x1="1" y1="2" x2="3" y2="4" id="l"/>
        </linearGradient>
        <rect x="10" y="10" width="10" height="10"  fill="blue" id="r2"/>

        <linearGradient x1="1" y1="2" x2="3" y2="4" id="l2">
        </linearGradient>
        <rect x="100" y="10" width="10" height="10"  fill="yellow" id="r3"/>
    </defs>

  <rect x="50" y="50" width="50" height="50"  fill="green" id="r1"/>
</svg>
```

Graphics elements declared in `<defs>` must not be displayed on the canvas(without `<use>`).
However, if there is an element of the form `<linearGradient> ... </linearGradient> `in front of the graphics element, the graphics element is displayed on the canvas.

related pr : https://github.com/thorvg/thorvg/pull/2213

PR #2519, opened 2024-07-04
svg_loader: Fix incorrect stack.pop() call in loader

JSUYA:
loader->stack.pop() at line 3271 is called to remove the defs node added to the stack due to nested graphics elements.
However, it is called in an inappropriate situation and the loader's node tree is damaged. 
Fixes an error where the close tag is recognized as a `line` of the graphics tag when it is `linearGradient`.

ex)
```
    <defs id="def">
        <linearGradient x1="1" y1="2" x2="3" y2="4" id="l"/>
        </linearGradient>
        <rect x="10" y="10" width="10" height="10"  fill="blue" id="r2"/>
    </defs>
```

related issue: https://github.com/thorvg/thorvg/issues/2518

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -3254,19 +3254,34 @@ static void _clonePostponedNodes(Array<SvgNodeIdPair>* cloneNodes, SvgNode* doc)
 }
 
 
-static void _svgLoaderParserXmlClose(SvgLoaderData* loader, const char* content)
+static void _svgLoaderParserXmlClose(SvgLoaderData* loader, const char* content, unsigned int length)
 {
+    const char* itr = nullptr;
+    int sz = length;
+    char tagName[20] = "";
+
     content = _skipSpace(content, nullptr);
+    itr = content;
+    while ((itr != nullptr) && *itr != '>') itr++;
+
+    if (itr) {
+        sz = itr - content;
+        while ((sz > 0) && (isspace(content[sz - 1]))) sz--;
+        if ((unsigned int)sz >= sizeof(tagName)) sz = sizeof(tagName) - 1;
+        strncpy(tagName, content, sz);
+        tagName[sz] = '\0';
+    }
+    else return;
 
     for (unsigned int i = 0; i < sizeof(groupTags) / sizeof(groupTags[0]); i++) {
-        if (!strncmp(content, groupTags[i].tag, groupTags[i].sz - 1)) {
+        if (!strncmp(tagName, groupTags[i].tag, sz)) {
             loader->stack.pop();
             break;
         }
     }
 
     for (unsigned int i = 0; i < sizeof(graphicsTags) / sizeof(graphicsTags[0]); i++) {
-        if (!strncmp(content, graphicsTags[i].tag, graphicsTags[i].sz - 1)) {
+        if (!strncmp(tagName, graphicsTags[i].tag, sz)) {
             loader->currentGraphicsNode = nullptr;
             loader->stack.pop();
             break;
@@ -3437,7 +3452,7 @@ static bool _svgLoaderParser(void* data, SimpleXMLType type, const char* content
             break;
         }
         case SimpleXMLType::Close: {
-            _svgLoaderParserXmlClose(loader, content);
+            _svgLoaderParserXmlClose(loader, content, length);
             break;
         }
         case SimpleXMLType::Data:


mgrudzinska (changes requested):
The root cause of this issue is that the `linearGradient` tag is incorrectly recognized as a graphics node `line` (`radialGradient` won't cause this issue). Your solution works, but the logic is not entirely correct.

In this case, everything will work, but the stack.pop() will be called due to the closing of `linearGradient` instead of `circle`:
```
<svg width="192px" height="192px" viewBox="0 0 192 192" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <defs id="def">
       <circle id="mk1" cx="12" cy="14" fill="blue" r="30">
        <rect x="10" y="10" width="10" height="10"  fill="blue" id="r2"/>
        <linearGradient x1="1" y1="2" x2="3" y2="4" id="l2">
        </linearGradient>
        <rect x="100" y="10" width="10" height="10"  fill="yellow" id="r3"/>
      </circle>
    </defs>

  <rect x="50" y="50" width="50" height="50"  fill="green" id="r1"/>
  <use href="#mk1"/>
</svg>
```

This does not cause any visual effect, but it introduces a logical problem. I'm affraid that in the future we may forget about this and if we introduce some changes this will come up and be confusing. 

what do you think?

JSUYA:
> The root cause of this issue is that the `linearGradient` tag is incorrectly recognized as a graphics node `line` (`radialGradient` won't cause this issue). Your solution works, but the logic is not entirely correct.

@mgrudzinska You are right!. Because of `line` of graphics tag, `linearGradient` is recognized as a graphics tag. I will fix it with a patch soon.
Thank you for your review!.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3266,7 +3266,7 @@ static void _svgLoaderParserXmlClose(SvgLoaderData* loader, const char* content)
     }
 
     for (unsigned int i = 0; i < sizeof(graphicsTags) / sizeof(graphicsTags[0]); i++) {
-        if (!strncmp(content, graphicsTags[i].tag, graphicsTags[i].sz - 1)) {
+        if (!strncmp(content, graphicsTags[i].tag, graphicsTags[i].sz - 1) && !isalpha(content[graphicsTags[i].sz - 1])) {
```
mgrudzinska, line 3269: isalpha is locale sensitive... and it's not clear for me what should happen if isalpha gets a space (I expected it to return false, but it returns true - I was checking the case `<line ...><\line>`).
anyway, imo the most elegant way would be to extract the tagname as it is done in the '_svgLoaderParserXmlOpen'.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3254,19 +3254,34 @@ static void _clonePostponedNodes(Array<SvgNodeIdPair>* cloneNodes, SvgNode* doc)
...
+    char tagName[20] = "";
+
     content = _skipSpace(content, nullptr);
+    itr = content;
+    while ((itr != nullptr) && *itr != '>') itr++;
+
+    if (itr) {
+        sz = itr - content;
+        while ((sz > 0) && (isspace(content[sz - 1]))) sz--;
+        if ((unsigned int)sz >= sizeof(tagName)) sz = sizeof(tagName);
```
mgrudzinska, line 3270: `sz = sizeof(tagName) - 1;` otherwise l.3272 will cause a problem

mgrudzinska, line 3270: sz = sizeof(tagName) - 1; otherwise l.3272 will cause a problem

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: @JSUYA ofc I'll review as many times as needed.
I now the code grows and grows... since there are identical parts of code in open and close functions, would you consider to introduce a fun handling it? or do you think it's a bad idea? 
if it's a bad idea, I'll approve, since I've already checked and beside one small comment it's ok

JSUYA: @mgrudzinska Thanks! Similar code already exists in _svgLoaderParserXmlOpen and _svgLoaderParserForValidCheckXmlOpen. I think it would be better to handle this as a new PR since it goes outside the purpose of this patch.  :)

mgrudzinska: ok

PR #2526, opened 2024-07-05
API: revise the APIs.

hermet:
deprecate the `identifier()` APIs by replacing them with `type()`.

ThorVG is going to introduce an instance `id()`,
and this could be confused with the `identifier()` methods.

with this new type() method can reduce the memory size by removing unncessary type data.

New Experimental C APIs:
- enum Tvg_Type
- Tvg_Result tvg_paint_get_type(const Tvg_Paint* paint, Tvg_Type* type)
- Tvg_Result tvg_gradient_get_type(const Tvg_Gradient* grad, Tvg_Type* type)

New Experimental C++ APIs:
- Type Paint::type() const
- Type Fill::type() const
- Type LinearGradient::type() const
- Type RadialGradient::type() const
- Type Shape::type() const
- Type Scene::type() const
- Type Picture::type() const
- Type Text::type() const

Deprecated C APIs:
- enum Tvg_Identifier
- Tvg_Result tvg_paint_get_identifier(const Tvg_Paint* paint, Tvg_Identifier* identifier)
- Tvg_Result tvg_gradient_get_identifier(const Tvg_Gradient* grad, Tvg_Identifier* identifier)

Deprecated C++ APIs:
- enum class Type
- uint32_t Paint::identifier() const
- uint32_t Fill::identifier() const
- static uint32_t Picture::identifier()
- static uint32_t Scene::identifier()
- static uint32_t Shape::identifier()
- static uint32_t LinearGradient:identifier()
- static uint32_T RadialGradient::identfier()

Removed Experimental APIs:
- static uint32_t Text::identifier()

### DIFF
--- a/src/bindings/capi/thorvg_capi.h
+++ b/src/bindings/capi/thorvg_capi.h
@@ -171,11 +171,8 @@ typedef enum {
 
 
 /**
- * \brief Enumeration indicating the ThorVG class type.
- *
- * \ingroup ThorVGCapi_Paint
- *
- * \since 0.9
+ * \see Tvg_Type
+ * \deprecated
  */
 typedef enum {
     TVG_IDENTIFIER_UNDEF = 0,   ///< Undefined type.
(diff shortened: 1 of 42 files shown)


mgrudzinska (approved):

src/bindings/capi/thorvg_capi.h
```diff
@@ -950,15 +970,21 @@ TVG_API Tvg_Result tvg_paint_get_composite_method(const Tvg_Paint* paint, const
 /**
 * \brief Gets the unique id value of the paint instance indicating the instance type.
```
mgrudzinska, line 971: maybe not to use 'id'? but up to you
`Gets the unique value indicating the type of the paint instance.`

mgrudzinska: same with gradient

mgrudzinska:
@hermet in tvgLottieModel.h we use 'uint8_t id = 0;    //1: linear, 2: radial` 
maybe forr consistency you want to change to type? but also up to you, doesn't seem to be important

hermet:
> @hermet in tvgLottieModel.h we use 'uint8_t id = 0; //1: linear, 2: radial` maybe forr consistency you want to change to type? but also up to you, doesn't seem to be important

It's a Lottie loader side independent value. I think it is not directly related to this change.

PR #2529, opened 2024-07-05
wg_engine: fix safe disposing of render data in MT environment

SergeyLebedkin:
Store desposed object in MT-safe list and then despose objects in sync stage

### DIFF
--- a/src/renderer/wg_engine/tvgWgRenderer.cpp
+++ b/src/renderer/wg_engine/tvgWgRenderer.cpp
@@ -50,6 +50,7 @@ void WgRenderer::initialize()
 
 void WgRenderer::release()
 {
+    clearDisposes();
     WgGeometryData::gMath->release();
     delete WgGeometryData::gMath;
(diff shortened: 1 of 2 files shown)


src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -221,15 +222,27 @@ bool WgRenderer::postRender()
...
+        mDisposed.renderDatas.push(data);
+    }
+}
+
+
+void WgRenderer::clearDisposes()
+{
+    for (uint32_t i = 0; i < mDisposed.renderDatas.count; i++) {
+        auto renderData = (WgRenderDataPaint*)mDisposed.renderDatas[i];
+        if (renderData) {
```
hermet, line 238: Here, the null check is not helpful. Please remove it and keep the code compact and neat.

SergeyLebedkin: Done

hermet (approved):

PR #2533, opened 2024-07-06
Renderer: rectified a missing image update flag

hermet:
The flag could be missed when the image is copied with a surface.


SergeyLebedkin (approved):
checked on wg engine.... seems to all issue are fixed now

PR #2535, opened 2024-07-07
gl_engine: make sure solid color not overwrite gradient fill

RuiwenTang:
If there are both gradient colors and solid colors the gradient color is used first for rendering.

Now the example `Duplicate` is correct:
截屏2024-07-07 15 23 41

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -1023,51 +1023,41 @@ bool GlRenderer::renderShape(RenderData data)
     if (bbox.w <= 0 || bbox.h <= 0) return true;
 
     uint8_t r = 0, g = 0, b = 0, a = 0;
-    int32_t drawDepth1 = 0, drawDepth2 = 0, drawDepth3 = 0;
+    int32_t drawDepth1 = 0, drawDepth2 = 0;
 
     size_t flags = static_cast<size_t>(sdata->updateFlag);
 
     if (flags == 0) return false;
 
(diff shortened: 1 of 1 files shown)


src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1048,10 +1048,7 @@ bool GlRenderer::renderShape(RenderData data)
     {
         auto gradient = sdata->rshape->fill;
         if (gradient) drawPrimitive(*sdata, gradient, RenderUpdateFlag::Gradient, drawDepth1);
-    }
-
-    if(flags & (RenderUpdateFlag::Color | RenderUpdateFlag::Transform))
-    {
+    } else if(flags & (RenderUpdateFlag::Color | RenderUpdateFlag::Transform)) {
```
hermet, line 1051: wasn't this?
```
if (transform) {
   if (gradient fill)
   else (solid fill)
   
   if (gradient fill stroke)
   else (solid fill stroke)
}
```

RuiwenTang: Can I only check RenderUpdateFlag::Color and RenderUpdateFlag::Gradient here?

hermet: > Can I only check RenderUpdateFlag::Color and RenderUpdateFlag::Gradient here?

yes, come to see again, my comment is wrong. :)

my intention was like: 
```
if (gradient fill | solid fill | transform) 
   if (gradient fill)...
   else (solid fill) ...
```

hermet:
@RuiwenTang is it ok? if only the transform flag is on. just to make it sure.

RuiwenTang:
> @RuiwenTang is it ok? if only the transform flag is on. just to make it sure.

I ran all the examples and it looked good to me. except for the features that GLEngine doesn't support yet

hermet (approved):

PR #2544, opened 2024-07-09
common/math: introduced custom atan2()

hermet:
the custom atan2 algorithm by Remez.
see: https://en.wikipedia.org/wiki/Remez_algorithm

This improved the Lottie example by ~2ms.
Total atan2 performance time was reduced by 43%

### DIFF
--- a/src/common/tvgMath.cpp
+++ b/src/common/tvgMath.cpp
@@ -22,6 +22,20 @@
 
 #include "tvgMath.h"
 
+//see: https://en.wikipedia.org/wiki/Remez_algorithm
+float mathAtan2(float y, float x)
+{
+    if (y == 0.0f) return 0.0f;
+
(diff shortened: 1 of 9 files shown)


src/common/tvgMath.cpp
```diff
@@ -22,6 +22,18 @@
 
 #include "tvgMath.h"
 
+//see: https://en.wikipedia.org/wiki/Remez_algorithm
+float mathAtan2(float y, float x)
+{
+    auto a = std::min(fabsf(x), fabsf(y)) / std::max(fabsf(x), fabsf(y));
```
JSUYA, line 28: Is it safe for zero divide?
I looked it up that atan2(0,0) is not defined, but cmath's atan2(0,0) returns 0, so it seems to be error-safe.
https://www.medcalc.org/manual/atan2-function.php

Actually, I don't think there is problem(?), but if you think it is a problem, I think you can add a comment.
( `// x !=0 and y != 0')

JSUYA (approved):

PR #2559, opened 2024-07-12
lottie: fix crash caught by asan

mgrudzinska:
Address sanitizer detected a bug introduced
in https://github.com/thorvg/thorvg/commit/d95ec723354d532598b9a7192d959a35c2eeecb0, causing a crash during precomposition update. Fixed.

Crash observed on game_finished.json


hermet (changes requested):
You are not applying the viewport. Currently, the scene can be deleted, but composition must be applied every time.

PR #2568, opened 2024-07-13
examples: prevent seg fault if init fails

mgrudzinska:


hermet:
https://github.com/thorvg/thorvg/pull/2569

Issue #2435, opened 2024-06-21
engines: enhance the raster engines(GL/WG) feature capabilities


hermet:
We are currently developing ThorVG's GPU acceleration engine for high-performance rendering. The purpose of this task is to ensure that the GL/WG engines are compatible with the ThorVG examples. The following table shows the list of feature coverage and the current status that the GL and WebGPU engines need to achieve.

| Examples | OpenGL/ES | WebGPU |
| ------------- | ------------- | ------------- | 
| Shapes | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| Path | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| Scene | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| Transform | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| Accessor | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| CustomTransform | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| Animation | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| Update | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| PictureSvg | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| DataLoad | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| LinearGradient | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| RadialGradient | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| PictureJpg | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| PicturePng | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| PictureWebp | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| SceneTransform | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
| GradientTransform | <ul><li>[x] Works </li></ul>  | <ul><li>[x] Works </li></ul> |
[...]

PR #2571, opened 2024-07-16
wg_engine: fix fill spread artifacts

SergeyLebedkin:
[Fill spread](https://github.com/thorvg/thorvg/issues/2435) example

Fixed range selector edged cases
before:
image

after:
image


mgrudzinska (approved):
I've reproduced the bug for bigger values of x2 and y2 in the linear api. now fixed

hermet (approved):

PR #2572, opened 2024-07-16
wg_engine: fix cubic generation artifacts (GradientStroke example)

SergeyLebedkin:
[GradientStroke](https://github.com/thorvg/thorvg/issues/2435) example

More accurate coefitient computation

Before:
image

After:
image


mgrudzinska (changes requested):
The underlying cause of this error is the finite precision of the floating point numbers. The value of `dt * nseg` should equal 1, but for different values of `nseg`, it might be slightly below or slightly above 1. This becomes a problem when `t` exceeds 1. 
I don't think the applied change is universal. 
Instead I would suggest:
```
for (auto t = 0.0f; t <= 1.0f; t += dt)
```

mgrudzinska (approved):

PR #2573, opened 2024-07-17
sw_engine: ++fast alpha blending

hermet:


mgrudzinska (approved):
in ~90% cases the results are different, but the difference is never bigger than 1 when comparing separate channels (new values are smaller), so almost impossible to notice it

hermet:
in ~90% cases the results are different,
@mgrudzinska Could you please suggest the "input (color,  alpha)"  where the output differs?

hermet:
@mgrudzinska nevermind, it's from rounding up/down differences.

Issue #2580, opened 2024-07-21
lottie: rendering issue when resizing frame buffer after upgrading to thorvg 0.14.x


theashraf:
Hey team, 

after upgrading from thorvg 0.13.x to 0.14.x in the dotLottie-rs project, we are encountering a rendering issue. Specifically, when we resize the buffer, nothing is rendered on the canvas anymore.

Here is our resize logic (we might be missing something here?):

```rust
pub fn resize(&mut self, width: u32, height: u32) -> Result<(), LottieRendererError> {
    if (width, height) == (self.width, self.height) {
        return Ok(());
    }

    if width == 0 || height == 0 {
        return Err(LottieRendererError::InvalidArgument(
            "Width and height must be greater than 0".to_string(),
        ));
    }

    self.width = width;
    self.height = height;

    self.buffer
        .resize((self.width * self.height * 4) as usize, 0);

    self.thorvg_canvas
        .set_target(
            &mut self.buffer,
            self.width,
            self.width,
            self.height,
            get_color_space_for_target(),
        )
        .map_err(LottieRendererError::ThorvgError)?;

    let (scaled_picture_width, scaled_picture_height, shift_x, shift_y) =
        self.layout.compute_layout_transform(
            self.width as f32,
            self.height as f32,
            self.picture_width,
            self.picture_height,
        );

    self.thorvg_animation
        .set_size(scaled_picture_width, scaled_picture_height)?;
    self.thorvg_animation.translate(shift_x, shift_y)?;

    self.thorvg_background_shape.append_rect(
        0.0,
        0.0,
        self.width as f32,
        self.height as f32,
        0.0,
        0.0,
    )?;

    Ok(())
}
```

This appears to be a regression, as the same code works correctly in thorvg 0.13.x.

Have there been any breaking changes or updates in thorvg 0.14.x that could cause this issue? any insights or suggestions on how to resolve this would be greatly appreciated.

hermet:
@theashraf thanks, could you please test this? https://github.com/LottieFiles/dotlottie-rs/pull/206

PR #2581, opened 2024-07-21
wasm: thorvg canvas resize problem.

hermet:
The canvas syncs before changing a target buffer.

issue: https://github.com/thorvg/thorvg/issues/2580

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -181,6 +181,8 @@ public:
         if (!canvas || !animation) return;
         if (this->width == width && this->height == height) return;
 
+        canvas->sync();
+
         this->width = width;
         this->height = height;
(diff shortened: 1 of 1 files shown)


hermet:
@tinyjin you might need to revert: https://github.com/thorvg/thorvg.web/commit/8f217f176b07015508946badc83269531e63c2f1

tinyjin:
> @tinyjin you might need to revert: [thorvg/thorvg.web@8f217f1](https://github.com/thorvg/thorvg.web/commit/8f217f176b07015508946badc83269531e63c2f1)

May I ask how this commit is related to the issue?

hermet:
@tinyjin I guess you made a work-around it due to the target() fails with InsufficientCondition.

tinyjin (approved):
@hermet Yes, it now works well in resize case including DPR change. I will send following patch for thorvg.web as well :)

Issue #2582, opened 2024-07-21
linearGradient doesn't abide by certain specifications


MewPurPur:
The first issue is with userSpaceOnUse, where values outside of [0, 1] are treated incorrectly.

```xml
<svg width="64" height="64" xmlns="http://www.w3.org/2000/svg"><linearGradient id="a" x1="0" y1="0" x2="2" y2="2"><stop offset="0" stop-color="#ff4596"/><stop offset=".25" stop-color="#ffe345"/><stop offset="1" stop-color="#45ffa2"/></linearGradient><rect width="64" height="64" fill="url(#a)"/></svg>
```

Expectation:
![image](https://github.com/user-attachments/assets/e74590a7-0d78-4b2f-b66a-5727f25fc586)

Nothing is displayed though, whereas the specification has the following section: https://svgwg.org/svg2-draft/coords.html#ObjectBoundingBox

> Any numeric value can be specified for values expressed as a fraction or percentage of object bounding box units. In particular, fractions less are zero or greater than one and percentages less than 0% or greater than 100% can be specified.

The second issue concerns gradients that start where they end.

```xml
<svg width="64" height="64" xmlns="http://www.w3.org/2000/svg"><linearGradient id="a" x1=".4" y1=".4" x2=".4" y2=".4" gradientUnits="userSpaceOnUse"><stop offset="0" stop-color="#ff4596"/><stop offset=".25" stop-color="#ffe345"/><stop offset=".5" stop-color="#45ffa2"/></linearGradient><rect width="64" height="64" fill="url(#a)"/></svg>
```

Nothing is displayed, which goes against the specification: https://svgwg.org/svg2-draft/pservers.html#LinearGradientNotes

> If x1 = x2 and y1=y2, then the area to be painted will be painted as a single color using the color and opacity of the last gradient stop.

hermet:
@MewPurPur @mgrudzinska  the fix will be in v0.14.5, thanks.

PR #2583, opened 2024-07-22
sw_engine: fix linear grad issue

mgrudzinska:
The restriction of the 'length' of a linear gradient
to values greater than FLT_EPSILON was causing rendering
issues in cases where it was a valid gradient with well
defined dx and dy values.

@issue: https://github.com/thorvg/thorvg/issues/2582

### DIFF
--- a/src/renderer/sw_engine/tvgSwRaster.cpp
+++ b/src/renderer/sw_engine/tvgSwRaster.cpp
@@ -1590,8 +1590,6 @@ static bool _rasterSolidGradientRect(SwSurface* surface, const SwBBox& region, c
 
 static bool _rasterLinearGradientRect(SwSurface* surface, const SwBBox& region, const SwFill* fill)
 {
-    if (fill->linear.len < FLOAT_EPSILON) return false;
-
     if (_compositing(surface)) {
         if (_matting(surface)) return _rasterGradientMattedRect<FillLinear>(surface, region, fill);
         else return _rasterGradientMaskedRect<FillLinear>(surface, region, fill);
--- a/src/renderer/sw_engine/tvgSwCommon.h
+++ b/src/renderer/sw_engine/tvgSwCommon.h
@@ -134,7 +134,6 @@ struct SwFill
 {
     struct SwLinear {
         float dx, dy;
-        float len;
         float offset;
     };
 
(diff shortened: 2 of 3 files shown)


src/renderer/sw_engine/tvgSwRaster.cpp
```diff
@@ -1590,8 +1590,6 @@ static bool _rasterSolidGradientRect(SwSurface* surface, const SwBBox& region, c
 
 static bool _rasterLinearGradientRect(SwSurface* surface, const SwBBox& region, const SwFill* fill)
 {
-    if (fill->linear.len < FLOAT_EPSILON) return false;
```
hermet, line 1593: I think the length equation is incorrect. It should be:
- Linear:  skip rendering, if  ((x1 - x2) * (y1 - y2) == 0)
- Radial: skip rendering if  (radius == 0)

hermet (commented):
The second patch could affect the performance a bit, please separate PR.

src/renderer/sw_engine/tvgSwRaster.cpp
mgrudzinska: do you mean `==` or `tvg::zero` ?

hermet: yes.

mgrudzinska:
> The second patch could affect the performance a bit, please separate PR.

https://github.com/thorvg/thorvg/pull/2588

src/renderer/sw_engine/tvgSwRaster.cpp
hermet: @mgrudzinska I'm not talking about dx/dy, Skip rendering if the given linear/radial gradient bound is 0.

hermet: @mgrudzinska I just noticed that two patches should be merged together. They can't exist separately. This looks good to me if [this pull request](https://github.com/thorvg/thorvg/pull/2588) is properly handled altogether. Otherwise, we can consider adding skipping rendering (currently, the radial gradient does this badly as well).

hermet (approved):

PR #2585, opened 2024-07-22
infra: switch the optimization option to level 3

hermet:
In case of modern systems, size is not a big problem, so we turn the optimization option to default
for better performance.

FPS: ~20% enhanced
Size: ~25% increased

Note that this doesn't affect to the web player.


hermet:
Applying this only to the release branch.

PR #2588, opened 2024-07-22
sw_engine: handle grad edge cases 

mgrudzinska:
For a linear gradient defined by identical start and end
points, and for a radial gradient with a radius of 0,
the rendered shape should have the color of the last
specified color stop.
The documentation has been updated accordingly.

@Issue: https://github.com/thorvg/thorvg/issues/2582

note:
for start=end svg standard defines that the last stop color should be used to fill the shape. for lottie files tested in skottie and AE - both seems to use the same def as svg

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -705,7 +705,8 @@ public:
      * @param[in] x2 The horizontal coordinate of the second point used to determine the gradient bounds.
      * @param[in] y2 The vertical coordinate of the second point used to determine the gradient bounds.
      *
-     * @note In case the first and the second points are equal, an object filled with such a gradient fill is not rendered.
+     * @note In case the first and the second points are equal, an object is filled with a single color using the last color specified in the colorStops().
+     * @see Fill::colorStops()
      */
     Result linear(float x1, float y1, float x2, float y2) noexcept;
 
@@ -771,6 +772,8 @@ public:
      * @param[in] radius The radius of the bounding circle.
      *
      * @retval Result::InvalidArguments in case the @p radius value is zero or less.
+     *
+     * @note In case the @p radius is zero, an object is filled with a single color using the last color specified in the colorStops().
      */
     Result radial(float cx, float cy, float radius) noexcept;
 
--- a/src/renderer/sw_engine/tvgSwCommon.h
+++ b/src/renderer/sw_engine/tvgSwCommon.h
@@ -153,6 +153,7 @@ struct SwFill
     uint32_t* ctable;
     FillSpread spread;
 
+    bool solid = false; //solid color fill with the last color from colorStops
     bool translucent;
 };
 
@@ -521,6 +522,7 @@ void imageReset(SwImage* image);
 void imageFree(SwImage* image);
 
 bool fillGenColorTable(SwFill* fill, const Fill* fdata, const Matrix& transform, SwSurface* surface, uint8_t opacity, bool ctable);
+const Fill::ColorStop* fillFetchSolid(const SwFill* fill, const Fill* fdata);
 void fillReset(SwFill* fill);
 void fillFree(SwFill* fill);
 
@@ -556,11 +558,11 @@ SwOutline* mpoolReqDashOutline(SwMpool* mpool, unsigned idx);
 void mpoolRetDashOutline(SwMpool* mpool, unsigned idx);
 
 bool rasterCompositor(SwSurface* surface);
-bool rasterGradientShape(SwSurface* surface, SwShape* shape, Type type);
+bool rasterGradientShape(SwSurface* surface, SwShape* shape, const Fill* fdata, uint8_t opacity);
 bool rasterShape(SwSurface* surface, SwShape* shape, uint8_t r, uint8_t g, uint8_t b, uint8_t a);
 bool rasterImage(SwSurface* surface, SwImage* image, const RenderMesh* mesh, const Matrix& transform, const SwBBox& bbox, uint8_t opacity);
 bool rasterStroke(SwSurface* surface, SwShape* shape, uint8_t r, uint8_t g, uint8_t b, uint8_t a);
-bool rasterGradientStroke(SwSurface* surface, SwShape* shape, Type type);
+bool rasterGradientStroke(SwSurface* surface, SwShape* shape, const Fill* fdata, uint8_t opacity);
 bool rasterClear(SwSurface* surface, uint32_t x, uint32_t y, uint32_t w, uint32_t h);
 void rasterPixel32(uint32_t *dst, uint32_t val, uint32_t offset, int32_t len);
 void rasterGrayscale8(uint8_t *dst, uint8_t val, uint32_t offset, int32_t len);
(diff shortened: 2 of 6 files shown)


inc/thorvg.h
```diff
@@ -705,7 +705,8 @@ class TVG_API LinearGradient final : public Fill
      * @param[in] x2 The horizontal coordinate of the second point used to determine the gradient bounds.
      * @param[in] y2 The vertical coordinate of the second point used to determine the gradient bounds.
      *
-     * @note In case the first and the second points are equal, an object filled with such a gradient fill is not rendered.
+     * @note In case the first and the second points are equal, an object is filled with a single color using the last color specified in the colorStops().
```
hermet, line 708: This hurts the performance. Maybe we can consider to switch using solid fill that case.

mgrudzinska: I saw the comment, I'll handle this asap, but I don;t think this will affect perf so often, this is a very rare case

mgrudzinska: @hermet please take a look in the tvgSwFill. this is already optimized, for both, radial and linear grad. 
radial was drawn before this change, but with a wrong color, linear was sometimes drawn sometimes not. if it was drawn, then the color was wrong

mgrudzinska:
files to test:
[lin.svg.zip](https://github.com/user-attachments/files/16395974/lin.svg.zip)
[rad.svg.zip](https://github.com/user-attachments/files/16395975/rad.svg.zip)


before:
Zrzut ekranu 2024-07-26 o 21 21 29

after:
Zrzut ekranu 2024-07-26 o 21 20 26

hermet:
if the color is solid, we can redirect it to the solid fill function. It's quite cheap and less size, avoiding those bad code duplication.

src/renderer/sw_engine/tvgSwRaster.cpp
```diff
@@ -1920,6 +1920,12 @@ bool rasterGradientShape(SwSurface* surface, SwShape* shape, Type type)
 {
     if (!shape->fill) return false;
 
+    if (shape->fill->forceSolid) {
```
hermet, line 1923:
```
if (auto color = FillFetchSolid(shape->fill)) {
   return raster...(c->r, c->g, c->b, MULTIPLY(c->a, opacity));
}
```

src/renderer/sw_engine/tvgSwFill.cpp
```diff
@@ -214,7 +223,12 @@ bool _prepareLinear(SwFill* fill, const LinearGradient* linear, const Matrix* tr
     fill->linear.dy = y2 - y1;
     auto len = fill->linear.dx * fill->linear.dx + fill->linear.dy * fill->linear.dy;
 
-    if (len < FLOAT_EPSILON) return true;
+    if (len < FLOAT_EPSILON) {
+        if (tvg::zero(fill->linear.dx) && tvg::zero(fill->linear.dy)) {
+            fill->forceSolid = true;
+        }
+        return true;
```
hermet, line 230:
```
return false;
```

src/renderer/sw_engine/tvgSwFill.cpp
```diff
@@ -820,14 +837,16 @@ bool fillGenColorTable(SwFill* fill, const Fill* fdata, const Matrix* transform,
 
     fill->spread = fdata->spread();
 
-    if (ctable) {
-        if (!_updateColorTable(fill, fdata, surface, opacity)) return false;
-    }
-
     if (fdata->type() == Type::LinearGradient) {
-        return _prepareLinear(fill, static_cast<const LinearGradient*>(fdata), transform);
+        if (_prepareLinear(fill, static_cast<const LinearGradient*>(fdata), transform)) {
```
hermet, line 841:
```
if (!_prepareLinear(fill, static_cast<const LinearGradient*>(fdata), transform)) return false;
```

src/renderer/sw_engine/tvgSwFill.cpp
```diff
@@ -125,15 +125,24 @@ static void _applyAA(const SwFill* fill, uint32_t begin, uint32_t end)
 
 static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface* surface, uint8_t opacity)
 {
+    const Fill::ColorStop* colors;
```
hermet, line 128: We can move this part to: (see 1923)
```
Fill::ColorStop* FillFetchSolid(...);
```

src/renderer/sw_engine/tvgSwCommon.h
```diff
@@ -152,6 +152,8 @@ struct SwFill
 
     uint32_t* ctable;
     FillSpread spread;
+    bool forceSolid = false;
+    uint8_t forcedSolid[4] = {0, 0, 0, 0};    //r, g, b, a
```
hermet, line 156: looks a single var is enough for the unusual case.
```
bool solid = false;   //fill solid color with the last color value.
```

hermet (commented):
Thanks. Please check the comments.

src/renderer/sw_engine/tvgSwFill.cpp
mgrudzinska: there is no reason to return false, this is a valid case. right now if I return false nothing will be rendered

src/renderer/sw_engine/tvgSwFill.cpp
mgrudzinska: I understand we don;t want to repeat the same code. so I applied, but I removed this part:
```
    //LOG: What type of gradient?!
    return false;
 ```

File: src/renderer/sw_engine/tvgSwFill.cpp (hunk 5 above)
Comment (mgrudzinska) [reply src/renderer/sw_engine/tvgSwFill.cpp:128]: to do this I need to have access to `const Fill* fdata` inside `FillFetchSolid` fun (and also to the opacity). seems a bit strange to have access to this in tvgRaster, doesn't it? 
instead I can pass const Fill::ColorStop* and the size to the `rasterGradientShape` but then `fillFetchSolid` seems to be useless, and also the same question, whether passing them to tvgRaster is ok?

File: src/renderer/sw_engine/tvgSwFill.cpp [hunk 7]
```diff
@@ -841,6 +859,7 @@ void fillReset(SwFill* fill)
         fill->ctable = nullptr;
     }
     fill->translucent = false;
+    fill->useSolid = false;
```
hermet, line 862: see `translucent`, it doesn't say `useTranslucent`.

please try to keep the consistency. Words that are not particularly important can be omitted.

src/renderer/sw_engine/tvgSwFill.cpp
hermet: @mgrudzinska

> to do this I need to have access to `const Fill* fdata` inside `FillFetchSolid` fun (and also to the opacity). seems a bit strange to have access to this in tvgRaster, doesn't it?

If necessary, we can do it. The engine internals won't be reused other places and are inevitably already tightly coupled. 

> I can pass const Fill::ColorStop* and the size to the `rasterGradientShape` but then `fillFetchSolid` seems to be useless,

Maybe. I'm not sure whether i got your point. 

Please consider this style again: with no need to introduce an unlikely used `color` member.

```
if (auto color = FillFetchSolid(shape->fill)) {
   return raster...(c->r, c->g, c->b, MULTIPLY(c->a, opacity));
}
```

mgrudzinska:
samples for tests with opacity: 
[lin.svg.zip](https://github.com/user-attachments/files/16445004/lin.svg.zip)
[rad.svg.zip](https://github.com/user-attachments/files/16445005/rad.svg.zip)

hermet (approved):

PR #2590, opened 2024-07-23
sw_engine: ++skip radial gradient filling.

hermet:
Do not rendering if the filling has an invalid condition.

### DIFF
--- a/src/renderer/sw_engine/tvgSwCommon.h
+++ b/src/renderer/sw_engine/tvgSwCommon.h
@@ -155,6 +155,7 @@ struct SwFill
     FillSpread spread;
 
     bool translucent;
+    bool invalid;
 };
 
 struct SwStrokeBorder
--- a/src/renderer/sw_engine/tvgSwFill.cpp
+++ b/src/renderer/sw_engine/tvgSwFill.cpp
@@ -256,7 +256,12 @@ bool _prepareRadial(SwFill* fill, const RadialGradient* radial, const Matrix* tr
     auto fy = P(radial)->fy;
     auto fr = P(radial)->fr;
 
-    if (r < FLOAT_EPSILON) return true;
+    fill->invalid = false;
+
+    if (r < FLOAT_EPSILON) {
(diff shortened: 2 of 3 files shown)


mgrudzinska (changes requested):
`fill->incvalid` is a bool, so no need. to use tvg::zero
but the problem is, that according to svg std r<0 is an error, but r==0 is ok. introducing this change we can not fulfill svg std
and I don;t think this affects performance much, these are very rare cases when r==0

hermet:
> `fill->incvalid` is a bool, so no need. to use tvg::zero but the problem is, that according to svg std r<0 is an error, but r==0 is ok. introducing this change we can not fulfill svg std and I don;t think this affects performance much, these are very rare cases when r==0

@mgrudzinska This doesn't hurt the svg spec. thorvg doesn't do for radial when radius is zero. Still r == 0 is acceptable by thorvg and radius can be an animator factor thus it needs.

mgrudzinska:
> @mgrudzinska This doesn't hurt the svg spec. Regardless of this change, thorvg doesn't do for radial when radius is zero. Still r == 0 is acceptable by thorvg and radius can be an animation factor thus it needs.

before your change the sample below was rendered, so this change affects svg specs
```
<svg viewBox="0 0 10 10" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
  <defs>
    <radialGradient id="myGradient" cx="1" cy="0" r="0">
      <stop offset="0" stop-color="gold" />
      <stop offset="1" stop-color="red" />
    </radialGradient>
  </defs>

  <rect x="0" y="0" width="7" height="7" fill="url(#myGradient)" />
</svg>
```

but if you don't agree and still want to push this pr, then you should handle rle as well, not only rectangles

hermet:
> but if you don't agree and still want to push this pr, then you should handle rle as well, not only rectangles

Got it, my fault. Let's come up with the logic.

PR #2596, opened 2024-07-25
common: fix trimming

mgrudzinska:
For pairs of start/end values like -0.5 and 0.5
(also -0.4 and 0.6, etc), the entire stroke should be drawn, but instead, nothing is drawn.

```
        shape1->appendCircle(245, 125, 50, 120);
        shape1->appendCircle(245, 365, 50, 120);
        shape1->appendCircle(125, 245, 120, 50);
        shape1->appendCircle(365, 245, 120, 50);
        shape1->fill(0, 50, 155, 100);
        shape1->strokeFill(0, 0, 255);
        shape1->strokeCap(tvg::StrokeCap::Round);
        shape1->strokeWidth(12);
        shape1->strokeTrim(-0.4f, 0.6f, false);

        auto shape2 = tvg::cast<tvg::Shape>(shape1->duplicate());
        shape2->translate(300, 300);
        shape2->fill(0, 155, 50, 100);
        shape2->strokeFill(0, 255, 0);
        shape2->strokeTrim(-0.4f, 0.6f, true);

        canvas->push(std::move(shape1));
        canvas->push(std::move(shape2));
```

before:
Zrzut ekranu 2024-07-25 o 19 54 01

after:
Zrzut ekranu 2024-07-25 o 19 52 37

### DIFF
--- a/src/renderer/sw_engine/tvgSwShape.cpp
+++ b/src/renderer/sw_engine/tvgSwShape.cpp
@@ -244,10 +244,10 @@ static void _dashMoveTo(SwDashStroke& dash, uint32_t offIdx, float offset, const
 }
 
 
-static void _trimPattern(SwDashStroke* dash, const RenderShape* rshape, float length)
+static void _trimPattern(SwDashStroke* dash, const RenderShape* rshape, float length, float trimBegin, float trimEnd)
 {
-    auto begin = length * rshape->stroke->trim.begin;
-    auto end = length * rshape->stroke->trim.end;
+    auto begin = length * trimBegin;
+    auto end = length * trimEnd;
 
     //default
     if (end > begin) {
@@ -340,6 +340,8 @@ static SwOutline* _genDashOutline(const RenderShape* rshape, const Matrix& trans
     auto offset = 0.0f;
     dash.cnt = rshape->strokeDash((const float**)&dash.pattern, &offset);
     auto simultaneous = rshape->stroke->trim.simultaneous;
+    float trimBegin = 0.0f, trimEnd = 1.0f;
+    if (trimmed) rshape->stroke->strokeTrim(trimBegin, trimEnd);
 
     if (dash.cnt == 0) {
         if (trimmed) dash.pattern = (float*)malloc(sizeof(float) * 4);
@@ -371,7 +373,7 @@ static SwOutline* _genDashOutline(const RenderShape* rshape, const Matrix& trans
 
     //must begin with moveTo
     if (cmds[0] == PathCommand::MoveTo) {
-        if (trimmed) _trimPattern(&dash, rshape, _outlineLength(rshape, 0, 0, simultaneous));
+        if (trimmed) _trimPattern(&dash, rshape, _outlineLength(rshape, 0, 0, simultaneous), trimBegin, trimEnd);
         _dashMoveTo(dash, offIdx, offset, pts);
         cmds++;
         pts++;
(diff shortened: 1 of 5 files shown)


SergeyLebedkin (approved):

src/renderer/tvgShape.h
```diff
@@ -209,10 +209,9 @@ struct Shape::Impl
 
     void strokeTrim(float begin, float end, bool simultaneous)
     {
-        if (!rs.stroke) {
-            if (begin == 0.0f && end == 1.0f) return;
-            rs.stroke = new RenderStroke();
-        }
+        if ((begin == 0.0f && end == 1.0f) || tvg::zero(begin - end + 1.0f)) return;
```
hermet, line 212: I think this condition should be on line 213 as it is necessary to reset the value.

hermet (changes requested):
please check a comment.

src/renderer/tvgShape.h
mgrudzinska: please take a look at the sample
since trim is added as the last one, `rs.stroke` already exists, so this condition won't be checked and stroke won't be drawn, which is a bug

no matter whether stroke is created or not, if ```(begin == 0.0f && end == 1.0f) || tvg::zero(begin - end + 1.0f)``` there is no point to add trimming. Am I missing something?

hermet: @mgrudzinska Please consider resetting the case. When the stroke trim has valid values, reset it to zero. Your change ignores it.

hermet (changes requested):

src/renderer/tvgShape.h
mgrudzinska: @hermet thank you, I forgot about the case of overriding the trim.

Please take a look now. I am overriding the values provided by the user, which isn't optimal because if the user queries them with a getter, he gets different values than he provided. Do you think I should handle this differently?

mgrudzinska: Although in many cases, the values are adjusted, and the getter will return something different than what the user provided so it might not be a problem.

hermet:
@mgrudzinska Hello Mira, modifying the user input value on the renderer side doesn't seem ideal and isn't necessary. Why don't we handle it in the raster engine instead?

hermet:
This may be a separate issue from this patch, but strokeTrimp doc needs to address the range looping concept.
```
     * The values of the arguments @p begin, @p end, and @p offset are in the range of 0.0 to 1.0, representing the beginning of the path and the end, respectively.
```

mgrudzinska:
@hermet please check now, should be ok. only the last commit contains change in logic. in the first one I simply moved the calculations from shape to sw_engine

src/renderer/sw_engine/tvgSwShape.cpp
```diff
@@ -340,6 +370,8 @@ static SwOutline* _genDashOutline(const RenderShape* rshape, const Matrix* trans
     auto offset = 0.0f;
     dash.cnt = rshape->strokeDash((const float**)&dash.pattern, &offset);
     auto simultaneous = rshape->stroke->trim.simultaneous;
+    float trimBegin, trimEnd;
+    _recalculateTrim(rshape->stroke, trimBegin, trimEnd);
```
hermet, line 374: We can put the main update in the tvgRender.h so that other backend engines can commonly use it.
```
//valid == true, when trimming should work
auto valid = rshape->stroke->trim(trimBegin, trimEnd);  //return the values applyed the looping
```

mgrudzinska: @hermet done. please notice that we have strokeTrim in renderShape and adjustTrim in RenderStroke. 
adjustTrim returns bool like strokeTrim, but I didn;t want to unify them, because not always when we ask about validation we need to adjust the values.

src/renderer/tvgRender.h
```diff
@@ -153,6 +154,36 @@ struct RenderStroke
         trim = rhs.trim;
     }
 
+    bool adjustTrim(float& begin, float& end) const
```
hermet, line 157: Please use this name:
```
bool strokeTrim(float& begin, float&end) const
```

- I hope you know that ThorVG doesn't stick to using the verb-objective style. 
- Regardless of any internal adjustments. just knowing the valid return is enough.

hermet (approved):
@mgrudzinska appreciated!

Issue #2599, opened 2024-07-26
renderer/text: Text with composite glyphs issue


lpogic:
When setting text with a composite glyph, the application crashes:

```
Initializer::init(0);
auto canvas = SwCanvas::gen();
auto text = Text::gen();
text->font("Arial", 80);
text->text("\xc5\xbb"); // Composite glyph 'Z' with '.' above
canvas->push(std::move(text)); // Segmentation fault
```

The case of a composite glyph is not handled at all and is interpreted as a simple glyph.

PR #2600, opened 2024-07-26
ttf_loader: Basic support for composite glyphs loading

lpogic:
Adds the ability to load some composite glyphs and prevents an error when a composite glyph is used.

Implementation based on ttf glyf table documentation: https://learn.microsoft.com/en-us/typography/opentype/spec/glyf

There are still some missing features like scaling, parent glyph point based positioning etc. I think this is a topic for future work. Howerever, it looks like implemented features are enough for utf-8 latin subset in major fonts.

solves issue: #2599


hermet (approved):
Outstanding code. thanks!

hermet:
@lpogic could you please add yourself here? https://github.com/thorvg/thorvg/blob/main/AUTHORS

Issue #2530, opened 2024-07-06
sw_engine: incorrect radial grad result


mgrudzinska:
sample:
```
<svg
  viewBox="0 0 800 800"
  xmlns="http://www.w3.org/2000/svg"
  xmlns:xlink="http://www.w3.org/1999/xlink">
  <defs>
    <radialGradient id="myGradient" spreadMethod="pad"
      gradientUnits="userSpaceOnUse"
      r="30" cx="400" cy="400">
      <stop offset="0.0" stop-color="#000000" />
      <stop offset="1.0" stop-color="#FFFFFF" />
    </radialGradient>
  </defs>

  <circle cx="400" cy="400" r="400" fill="url(#myGradient)" />
</svg>
```

<img width="1156" alt="Zrzut ekranu 2024-07-6 o 02 34 25" src="https://github.com/thorvg/thorvg/assets/67589014/96ab4b4b-a7c3-4a24-9a8b-ae5be06c4b46">

hermet:
@mgrudzinska Please note that visual quality or differences will not be classified as software bugs.

mgrudzinska:
maybe changing pad -> reflect shows better the problem:

current:
<img width="400" alt="Zrzut ekranu 2024-07-6 o 10 52 13" src="https://github.com/thorvg/thorvg/assets/67589014/723fd163-c8ca-4c0d-9f20-5989650de999">

expected:
<img width="400" alt="Zrzut ekranu 2024-07-6 o 10 52 37" src="https://github.com/thorvg/thorvg/assets/67589014/9a94fbb0-99f4-4cc6-bce8-2ccc5e3d8d02">

PR #2601, opened 2024-07-26
sw_engine: fix radial coefficients calculations

mgrudzinska:
In the Taylor series expansion, there was a missing division by 2! in the term with the second derivative.

@Issue: https://github.com/thorvg/thorvg/issues/2530

sample:
```
<svg
  viewBox="0 0 100 100"
  xmlns="http://www.w3.org/2000/svg"
  <defs>
    <radialGradient id="myGradient" spreadMethod="reflect"
      gradientUnits="userSpaceOnUse"
      r="10" cx="50" cy="50">
      <stop offset="0.0" stop-color="#000000" />
      <stop offset="1.0" stop-color="#FFFFFF" />
    </radialGradient>
  </defs>

  <rect width="100" height="100" x="0" y="0" fill="url(#myGradient)" />
</svg>
```

before:
Zrzut ekranu 2024-07-27 o 01 27 23

after:
Zrzut ekranu 2024-07-27 o 01 27 06

### DIFF
--- a/src/renderer/sw_engine/tvgSwFill.cpp
+++ b/src/renderer/sw_engine/tvgSwFill.cpp
@@ -58,7 +58,7 @@ static void _calculateCoefficients(const SwFill* fill, uint32_t x, uint32_t y, f
     auto deltaDeltaRr = 2.0f * (radial->a11 * radial->a11 + radial->a21 * radial->a21) * radial->invA;
 
     det = b * b + (rr - radial->fr * radial->fr) * radial->invA;
-    deltaDet = 2.0f * b * deltaB + deltaB * deltaB + deltaRr + deltaDeltaRr;
+    deltaDet = 2.0f * b * deltaB + deltaB * deltaB + deltaRr + deltaDeltaRr * 0.5f;
     deltaDeltaDet = 2.0f * deltaB * deltaB + deltaDeltaRr;
 }
 


hermet:
@mgrudzinska https://github.com/thorvg/thorvg/issues/2530#issuecomment-2211679220

src/renderer/sw_engine/tvgSwFill.cpp
```diff
@@ -58,7 +58,7 @@ static void _calculateCoefficients(const SwFill* fill, uint32_t x, uint32_t y, f
     auto deltaDeltaRr = 2.0f * (radial->a11 * radial->a11 + radial->a21 * radial->a21) * radial->invA;
 
     det = b * b + (rr - radial->fr * radial->fr) * radial->invA;
-    deltaDet = 2.0f * b * deltaB + deltaB * deltaB + deltaRr + deltaDeltaRr;
+    deltaDet = 2.0f * b * deltaB + deltaB * deltaB + deltaRr + deltaDeltaRr * 0.5f;
```
hermet, line 61: Is this correct?
```
 * A = a  // fill->radial.a
 * B = 2 * (dr * fr + rx * dx + ry * dy)
 * C = fr^2 - rx^2 - ry^2
 ```

mgrudzinska: yes, this is correct. 
in the subsequent calculations I applied transformations to avoid repetitive divisions by 'a' in each term, etc. however the initial equations are correct.

hermet (approved):

Issue #2410, opened 2024-06-17
wg_engine: integrate with thorvg player


hermet:
- [x] Research and prototype the Emscripten build with WebGPU
- [x] Implement the necessary integration logic in the tvgWasm part
- [x] Support Multi-Canvases
- [ ] Integrate with the ThorVG player

hermet:
see also: 
- https://eliemichel.github.io/LearnWebGPU/getting-started/hello-webgpu.html
- https://github.com/gfx-rs/wgpu-native/tree/trunk/examples/capture

hermet:
```
    {
        WGPUSurfaceDescriptorFromCanvasHTMLSelector fromCanvasHTMLSelector;
        fromCanvasHTMLSelector.chain.next = NULL;
        fromCanvasHTMLSelector.chain.sType = WGPUSType_SurfaceDescriptorFromCanvasHTMLSelector;
        fromCanvasHTMLSelector.selector = "canvas";

        WGPUSurfaceDescriptor surfaceDescriptor;
        surfaceDescriptor.nextInChain = &fromCanvasHTMLSelector.chain;
        surfaceDescriptor.label = NULL;

        return wgpuInstanceCreateSurface(instance, &surfaceDescriptor);
    }  
```

capnm:
WGPU / DAWN – which backend is currently used in the development of the `wg_engine`?

SergeyLebedkin:
> WGPU / DAWN – which backend is currently used in the development of the `wg_engine`?

Currently we are using **wgpu_native** pre-builds
Please, see pre-buils
https://github.com/gfx-rs/wgpu-native/releases

tinyjin:
Hello, @SergeyLebedkin.

I'd like to discuss interface for WebGPU initialization part. Since current methods don't support browser to have multiple canvases.

According to wg_engine, WGPUAdapter and WGPUDevice are generated when calling WGContext::initialize. Which means those Adapter and Device could be generated in duplicate.

It's going to happen when we have multiple animations or try to resize more than once.

![CleanShot 2024-08-27 at 17 52 30@2x](https://github.com/user-attachments/assets/7ee95caa-2d02-4ae2-8fd8-9f6cb7087e78)

This function is called, following the steps:
- wasm::resize -> canvas::target -> renderer::target -> context::initialize -> _then the WGContext potentially tries to call adapter and device by multiple._


And this is basic WebGPU usage in multiple canvas, it's written in JavaScript, but you can understand the logic.

[...]

PR #2603, opened 2024-07-28
wg_engine: emscripten and browser support

SergeyLebedkin:
[issues 2410: emscripten support](https://github.com/thorvg/thorvg/issues/2410)
- Research and prototype the Emscripten build with WebGPU.

meson setup script:
`meson setup builddir -Ddefault_library=static -Dloaders=all -Dsavers="all" -Dbindings="wasm_beta" -Dengines="wg_beta" --cross-file ./cross/wasm_webgpu.txt`

image

tesing javascript:
```JS
<!DOCTYPE html>
<html>
<head>
  <title>Emscripten</title>
  <script src="../builddir/src/bindings/wasm/thorvg-wasm.js"></script>
</head>
<body>
  <div>WebGPU web player</div>
  <canvas id="canvasWebgpu" width="800" height="600">Hello</canvas>
  <script>
    let animator = null;
    let totalFrames = 0.0;
    let duration = 0.0;

    async function render(timestamp) {
      await animator.frame((timestamp / (duration * 1000) % 1) * totalFrames);
      await animator.update();
      await animator.render();
      window.requestAnimationFrame(render);
    }

    window.onload = async () => {
      // create thorvg module
      const module = await Module();
      // create animator
      animator = new module.TvgWgLottieAnimation();
      await animator.resize(800, 800);
      // load animation from file and get data
      let jsonFile = await (await fetch('../examples/resources/lottie/swinging.json')).text();
      await animator.load(jsonFile, "json");
      totalFrames = await animator.totalFrame();
      duration = await animator.duration();
      // request animation
      window.requestAnimationFrame(render);
    }
    window.onclose = () => animator.delete();
  </script>
</body>
</html>
```

### DIFF
--- a/src/renderer/wg_engine/tvgWgCommon.cpp
+++ b/src/renderer/wg_engine/tvgWgCommon.cpp
@@ -21,6 +21,9 @@
  */
 
 #include "tvgWgCommon.h"
+#ifdef __EMSCRIPTEN__
+#include <emscripten.h>
+#endif
 #include <iostream>
 
 //*****************************************************************************
@@ -50,12 +53,13 @@ void WgContext::initialize(WGPUInstance instance, WGPUSurface surface)
     };
     // request adapter
     wgpuInstanceRequestAdapter(instance, &requestAdapterOptions, onAdapterRequestEnded, &adapter);
+    #ifdef __EMSCRIPTEN__
+    while (!adapter) emscripten_sleep(10);
+    #endif
     assert(adapter);
 
-    // adapter enumerate features
+    // get feature names
     size_t featuresCount = wgpuAdapterEnumerateFeatures(adapter, featureNames);
-    wgpuAdapterGetProperties(adapter, &adapterProperties);
-    wgpuAdapterGetLimits(adapter, &supportedLimits);
 
     // request device
     WGPUDeviceDescriptor deviceDesc{};
@@ -76,6 +80,9 @@ void WgContext::initialize(WGPUInstance instance, WGPUSurface surface)
     };
     // request device
     wgpuAdapterRequestDevice(adapter, &deviceDesc, onDeviceRequestEnded, &device);
+    #ifdef __EMSCRIPTEN__
+    while (!device) emscripten_sleep(10);
+    #endif
     assert(device);
 
     // on device error function
@@ -535,8 +542,10 @@ WGPUShaderModule WgPipeline::createShaderModule(WGPUDevice device, const char* c
     WGPUShaderModuleDescriptor shaderModuleDesc{};
     shaderModuleDesc.nextInChain = &shaderModuleWGSLDesc.chain;
     shaderModuleDesc.label = label;
+    #ifndef __EMSCRIPTEN__
     shaderModuleDesc.hintCount = 0;
     shaderModuleDesc.hints = nullptr;
+    #endif
     return wgpuDeviceCreateShaderModule(device, &shaderModuleDesc);
 }
 
--- a/src/renderer/wg_engine/tvgWgBindGroups.cpp
+++ b/src/renderer/wg_engine/tvgWgBindGroups.cpp
@@ -33,11 +33,15 @@ WGPUBindGroupLayout WgBindGroupRadialGradient::layout = nullptr;
 WGPUBindGroupLayout WgBindGroupPicture::layout = nullptr;
 // composition and blending properties groups
 WGPUBindGroupLayout WgBindGroupTexture::layout = nullptr;
-WGPUBindGroupLayout WgBindGroupTextureStorageRgba::layout = nullptr;
-WGPUBindGroupLayout WgBindGroupTextureStorageBgra::layout = nullptr;
+WGPUBindGroupLayout WgBindGroupTextureStorageRgbaRO::layout = nullptr;
+WGPUBindGroupLayout WgBindGroupTextureStorageRgbaWO::layout = nullptr;
+WGPUBindGroupLayout WgBindGroupTextureStorageBgraRO::layout = nullptr;
+WGPUBindGroupLayout WgBindGroupTextureStorageBgraWO::layout = nullptr;
 WGPUBindGroupLayout WgBindGroupTextureSampled::layout = nullptr;
+WGPUBindGroupLayout WgBindGroupTexBlend::layout = nullptr;
+WGPUBindGroupLayout WgBindGroupTexBlendMask::layout = nullptr;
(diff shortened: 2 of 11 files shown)


src/bindings/wasm/tvgWgWasmLottieAnimation.cpp
```diff
@@ -0,0 +1,156 @@
...
+    void resize(uint32_t width, uint32_t height)
+    {
+        // instance
+        instance = wgpuCreateInstance(nullptr);
+
+        // surface
+        WGPUSurfaceDescriptorFromCanvasHTMLSelector canvasDesc{};
+        canvasDesc.chain.next = nullptr;
+        canvasDesc.chain.sType = WGPUSType_SurfaceDescriptorFromCanvasHTMLSelector;
+        canvasDesc.selector = "#canvasWebgpu";
```
tinyjin, line 53: `#thorvg-canvas` for current lottie-player element.

```html
<canvas id="thorvg-canvas" style="width: 100%; height: 100%;" />
```

src/bindings/wasm/tvgWgWasmLottieAnimation.cpp
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    ~TvgWgLottieAnimation()
+    {
+        wgpuSurfaceRelease(surface);
+        wgpuInstanceRelease(instance);
+        tvg::Initializer::term();
+     }
+
+    void resize(uint32_t width, uint32_t height)
```
tinyjin, line 44: For canvas size, Number type in JS is pretty enough, so I think we can simply go with `int` type.

CleanShot 2024-07-29 at 09 08 44@2x

src/bindings/wasm/tvgWgWasmLottieAnimation.cpp
```diff
@@ -0,0 +1,156 @@
...
+    {
```
tinyjin, line 45: We can avoid unnecessary computing by here checking:
```
if (this->width == width && this->height == height) return;
```

src/bindings/wasm/tvgWgWasmLottieAnimation.cpp
```diff
@@ -0,0 +1,156 @@
...
+        // instance
+        instance = wgpuCreateInstance(nullptr);
```
tinyjin, line 47: Do we need new instance creation every resize function called?

Is there another way without recreation? Currently lottie-player calls `resize` every frames.

src/bindings/wasm/meson.build
```diff
@@ -1,5 +1,11 @@
 if (cc.get_id() == 'emscripten')
-    source_file = files('tvgWasmLottieAnimation.cpp')
+
+    if get_option('bindings').contains('wasm_beta') == true
```
tinyjin, line 3: `if get_option('bindings').contains('wasm_beta')`?

src/renderer/wg_engine/tvgWgCommon.cpp
```diff
@@ -76,7 +79,9 @@ void WgContext::initialize(WGPUInstance instance, WGPUSurface surface)
     };
     // request device
     wgpuAdapterRequestDevice(adapter, &deviceDesc, onDeviceRequestEnded, &device);
-    assert(device);
```
tinyjin, line 79: Isn't it required by wgpu-native? I assume wgpu-native need to ensure whether device is valid.
We can put assertion after `emscripten_sleep` ilne, it would work in web as well.

tinyjin (commented):
Simply reviewed binding side. I'll revisit soon for compatibility with our lottie-player.

hermet:
@SergeyLebedkin quick feedback.
1. please separate commits: 
-wasm 
-wg_engine 
2. I think updating the existing `tvgWasmLottieAnimation` is a better idea to remove the duplicate logic rather than introducing a new animation.

src/renderer/wg_engine/tvgWgCommon.cpp
SergeyLebedkin: we can keep it.
fixed

SergeyLebedkin:
> @SergeyLebedkin quick feedback.
> 
> 1. please separate commits:
>    -wasm
>    -wg_engine
> 2. I think updating the existing `tvgWasmLottieAnimation` is a better idea to remove the duplicate logic rather than introducing a new animation.

OK... as we decide we keep single file for both rendering engines

src/bindings/wasm/meson.build
SergeyLebedkin: removed as non neccessury

src/bindings/wasm/tvgWgWasmLottieAnimation.cpp
SergeyLebedkin: removed as non neccessury

src/bindings/wasm/meson.build
SergeyLebedkin: removed as non neccessury

src/bindings/wasm/tvgWgWasmLottieAnimation.cpp
SergeyLebedkin: > For canvas size, Number type in JS is pretty enough, so I think we can simply go with `int` type.
> 
 Thanx for advice!

src/renderer/wg_engine/meson.build
```diff
@@ -19,7 +19,10 @@ source_file = [
     'tvgWgShaderTypes.cpp'
 ]
 
-wgpu_dep = dependency('wgpu_native')
+wgpu_dep = []
```
hermet, line 22: @SergeyLebedkin please do not touch wasm related  in this change.

cross/wasm_webgpu.txt
```diff
@@ -0,0 +1,17 @@
+[binaries]
```
hermet, line 1: I think this cross file is not proper in this PR.

hermet:
@SergeyLebedkin Please stabilizing wg_engine itself compatible to web environment regardless of wasm.

src/renderer/wg_engine/meson.build
SergeyLebedkin: > @SergeyLebedkin please do not touch wasm related in this change.

maybe I dont undertand what do ypu mean

SergeyLebedkin: @hermet  So, I can remove this change from this PR

hermet:
@tinyjin please approve this if this works for you. thanks.

tinyjin (approved):

PR #2604, opened 2024-07-29
renderer: code refactoring

hermet:
Replaced the transformation with
a strong associated data field.

This helps to reduce the binary size (-1k).


SergeyLebedkin (approved):
Nice idea to use references instead of pointers for transformation. transformations are always exists, so we don't need to check existence.

PR #2605, opened 2024-07-30
Interactive Animation Support

hermet:
https://github.com/user-attachments/assets/567980d2-3088-4361-a027-3dae35d0f270

https://github.com/thorvg/thorvg/issues/2055


src/bindings/capi/thorvg_capi.h
```diff
@@ -2067,6 +2070,21 @@ TVG_API Tvg_Result tvg_picture_set_size(Tvg_Paint* paint, float w, float h);
...
+* This function searches for a paint object within the Picture scene that matches the provided @p id.
+*
+* \param[in] id The Unique ID of the paint object.
+
+* \return A pointer to the paint object that matches the given identifier, or @c nullptr if no matching paint object is found.
+*
+* \see tvg_access_generate_id()
+* \note experimental API
+*/
+TVG_API const Tvg_Paint* tvg_picture_get_paint(Tvg_Paint* paint, uint32_t id);
```
hermet, line 2085: \+ @mgrudzinska

src/renderer/tvgAccessor.cpp
```diff
@@ -44,26 +45,46 @@ static bool accessChildren(Iterator* it, function<bool(const Paint* paint)> func
...
+        accessChildren(it, func, data);
         delete(it);
     }
-    return picture;
+    return Result::Success;
+}
+
+
+uint32_t Accessor::id(const char* name) noexcept
+{
```
mgrudzinska, line 86: +check for nullptr

inc/thorvg.h
```diff
@@ -389,15 +389,16 @@ class TVG_API Paint
...
      * @param[out] x The x-coordinate of the upper-left corner of the object.
      * @param[out] y The y-coordinate of the upper-left corner of the object.
      * @param[out] w The width of the object.
      * @param[out] h The height of the object.
-     * @param[in] transformed If @c true, the paint's transformations are taken into account, otherwise they aren't.
+     * @param[in] transformed If @c true, the paint's transformations are taken into account in the scene its belonged. Otherwise they aren't.
      *
+     * @note This is useful when you need to figure out the bounding box of the paint in the canvas space.
      * @note The bounding box doesn't indicate the actual drawing region. It's the smallest rectangle that encloses the object.
+     * @note The paint needs to be pushed into a canvas if @p transformed is true, Update the paint beforehand to get the correct values.
```
mgrudzinska, line 400: `If @p transformed is @c true, the paint needs to be pushed into a canvas and updated before this api is called.`

mgrudzinska (commented):
questions:
1.  do we want to have a C api to set the id to the Tvg_Paint object?
2.  in the svg loader we could set the node->id to the id (or it can be done in a separate pr)

src/renderer/tvgAccessor.cpp
hermet: not necessary.

hermet:
> * do we want to have a C api to set the id to the Tvg_Paint object?
> * in the svg loader we could set the node->id to the id (or it can be done in a separate pr)

For both questions, yes. probably. we can add them if it's necessary but I don't see any requirement yet.

PR #2618, opened 2024-08-08
wasm: Support WebGPU Binding

tinyjin:
Added build script & binding for WebGPU.

This is compatible for lottie-player in thorvg.web, doesn't affect to SW Renderer WASM.

Developers can choose one of ways for building ThorVG WASM.

```sh
# Enter ThorVG directory
cd ./thorvg

# If build_wasm exists, remove it
rm -rf ./build_wasm

# (Option 1) Build ThorVG for SW Renderer
./wasm_build.sh {absolute path}/emsdk/

# (Option 2) Build ThorVG for WebGPU capability [NEW]
./wasm_webgpu_build.sh {absolute path}/emsdk/
```

### DIFF
--- a/cross/wasm_webgpu.txt
+++ b/cross/wasm_webgpu.txt
@@ -1,0 +1,21 @@
+[binaries]
+cpp = 'EMSDK:upstream/emscripten/em++.py'
+ar = 'EMSDK:upstream/emscripten/emar.py'
+strip = '-strip'
+
+[properties]
+root = 'EMSDK:upstream/emscripten/system'
+shared_lib_suffix = 'js'
+static_lib_suffix = 'js'
+shared_module_suffix = 'js'
+exe_suffix = 'js'
+
+[built-in options]
+cpp_args = ['-Wshift-negative-value', '-flto', '-Os', '-fno-exceptions']
+cpp_link_args = ['-lembind', '-sMODULARIZE=1', '-sUSE_WEBGPU=1', '-sASYNCIFY=1', '-sFORCE_FILESYSTEM=1', '-sALLOW_MEMORY_GROWTH=1', '-sEXPORT_ES6=1', '-sSINGLE_FILE=1']
+
+[host_machine]
+system = 'emscripten'
+cpu_family = 'wasm32'
+cpu = 'wasm32'
+endian = 'little'
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -23,6 +23,9 @@
 #include <thorvg.h>
 #include <emscripten/bind.h>
 #include "tvgPicture.h"
+#ifdef THORVG_WG_RASTER_SUPPORT
+    #include <webgpu/webgpu.h>
+#endif
 
 using namespace emscripten;
 using namespace std;
@@ -36,6 +39,10 @@ public:
     ~TvgLottieAnimation()
     {
         free(buffer);
+        #ifdef THORVG_WG_RASTER_SUPPORT
+            wgpuSurfaceRelease(surface);
+            wgpuInstanceRelease(instance);
+        #endif
         Initializer::term();
     }
 
@@ -74,7 +81,7 @@ public:
     }
 
     // Render methods
-    bool load(string data, string mimetype, int width, int height, string rpath = "")
+    bool load(string data, string mimetype, int width, int height, string selector = "", string rpath = "")
     {
         errorMsg = NoError;
 
@@ -85,8 +92,20 @@ public:
             return false;
         }
 
-        //back up for saving
-        this->data = data;
+        this->data = data; //back up for saving
+
+        #ifdef THORVG_WG_RASTER_SUPPORT
+            this->selector = selector;
+            instance = wgpuCreateInstance(nullptr);
+            // surface
+            WGPUSurfaceDescriptorFromCanvasHTMLSelector canvasDesc{};
+            canvasDesc.chain.next = nullptr;
+            canvasDesc.chain.sType = WGPUSType_SurfaceDescriptorFromCanvasHTMLSelector;
+            canvasDesc.selector = this->selector.c_str();
+            WGPUSurfaceDescriptor surfaceDesc{};
+            surfaceDesc.nextInChain = &canvasDesc.chain;
+            surface = wgpuInstanceCreateSurface(instance, &surfaceDesc);
+        #endif
 
         canvas->clear(true);
 
@@ -184,9 +203,14 @@ public:
         this->width = width;
         this->height = height;
 
-        free(buffer);
-        buffer = (uint8_t*)malloc(width * height * sizeof(uint32_t));
-        canvas->target((uint32_t *)buffer, width, width, height, SwCanvas::ABGR8888S);
+        #ifdef THORVG_WG_RASTER_SUPPORT
+            static_cast<WgCanvas*>(canvas.get())->target(this->instance, this->surface, width, height);
+        #else
+            free(buffer);
+            buffer = (uint8_t*)malloc(width * height * sizeof(uint32_t));
+            static_cast<SwCanvas*>(canvas.get())->target((uint32_t *)buffer, width, width, height, SwCanvas::ABGR8888S);
+        #endif
+
 
         float scale;
         float shiftX = 0.0f, shiftY = 0.0f;
@@ -312,13 +336,22 @@ private:
     explicit TvgLottieAnimation()
     {
         errorMsg = NoError;
+        auto engine = CanvasEngine::Sw;
+        #ifdef THORVG_WG_RASTER_SUPPORT
+            engine = CanvasEngine::Wg;
+        #endif
 
-        if (Initializer::init(0) != Result::Success) {
+
+        if (Initializer::init(0, engine) != Result::Success) {
             errorMsg = "init() fail";
             return;
         }
 
-        canvas = SwCanvas::gen();
+        #ifdef THORVG_WG_RASTER_SUPPORT
+            canvas = WgCanvas::gen();
+        #else
+            canvas = SwCanvas::gen();
+        #endif
         if (!canvas) errorMsg = "Invalid canvas";
 
         animation = Animation::gen();
@@ -327,7 +360,7 @@ private:
 
 private:
     string                 errorMsg;
-    unique_ptr<SwCanvas>   canvas = nullptr;
+    unique_ptr<Canvas>   canvas = nullptr;
     unique_ptr<Animation>  animation = nullptr;
     string                 data;
     uint8_t*               buffer = nullptr;
@@ -335,6 +368,12 @@ private:
     uint32_t               height = 0;
     float                  psize[2];         //picture size
     bool                   updated = false;
+
+    #ifdef THORVG_WG_RASTER_SUPPORT
+    WGPUInstance instance{};
+    WGPUSurface surface{};
+    string selector;
+    #endif
 };
 
 EMSCRIPTEN_BINDINGS(thorvg_bindings)
--- a/wasm_webgpu_build.sh
+++ b/wasm_webgpu_build.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+#https://github.com/thorvg/thorvg/wiki/WebGPU-Raster-Engine-Development
+
+if [ ! -d "./build_wasm" ]; then
+    sed "s|EMSDK:|$1|g" ./cross/wasm_webgpu.txt > /tmp/.wasm_cross.txt
+    meson -Db_lto=true -Ddefault_library=static -Dstatic=true -Dloaders="all" -Dsavers="all" -Dthreads=false -Dbindings="wasm_beta" -Dengines="wg_beta" --cross-file /tmp/.wasm_cross.txt build_wasm
+fi
+
+ninja -C build_wasm/
+ls -lrt build_wasm/src/bindings/wasm/thorvg-wasm.*
--- a/src/renderer/wg_engine/meson.build
+++ b/src/renderer/wg_engine/meson.build
@@ -19,7 +19,10 @@ source_file = [
     'tvgWgShaderTypes.cpp'
 ]
 
-wgpu_dep = dependency('wgpu_native')
+wgpu_dep = []
+if not get_option('bindings').contains('wasm_beta')
+    wgpu_dep = dependency('wgpu_native')
+endif
 
 engine_dep += [declare_dependency(
     dependencies        : wgpu_dep,


SergeyLebedkin (approved):

cross/wasm_webgpu.txt
```diff
@@ -0,0 +1,21 @@
...
+shared_lib_suffix = 'js'
+static_lib_suffix = 'js'
+shared_module_suffix = 'js'
+exe_suffix = 'js'
+
+[built-in options]
+cpp_args = ['-Wshift-negative-value', '-flto', '-Os', '-fno-exceptions']
+cpp_link_args = ['-lembind', '-sMODULARIZE=1', '-sUSE_WEBGPU=1', '-sASYNCIFY=1', '-sFORCE_FILESYSTEM=1', '-sALLOW_MEMORY_GROWTH=1', '-sEXPORT_ES6=1', '-sSINGLE_FILE=1']
+
+[host_machine]
```
hermet, line 17: We can probably modify wasm_x86_i686 to be the same and rename it to:
```
wasm_x86_i686 -> wasm32_sw.txt
wasm_webgpu -> wasm32_wg.txt
```

wasm_webgpu_build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
hermet, line 1: let's use a single script wasm_build.sh  it could have a parameter:
```
./wasm_build.sh   //default: software
./wasm_builds.sh sw  //software engine
./wasm_build.sh wg //webgpu engine
```
something like:
```
if [ ! -d "./build_wasm" ]; then
    if webgpu
      sed "s|EMSDK:|$1|g" ./cross/wasm_webgpu.txt > /tmp/.wasm_cross.txt
      meson -Db_lto=true -Ddefault_library=static -Dstatic=true -Dloaders="all" -Dsavers="all" -Dthreads=false -Dbindings="wasm_beta" -Dengines="wg_beta" --cross-file /tmp/.wasm_cross.txt build_wasm
    else software
      sed "s|EMSDK:|$1|g" ./cross/wasm_x86_i686.txt > /tmp/.wasm_cross.txt
      meson -Db_lto=true -Ddefault_library=static -Dstatic=true -Dloaders="all" -Dsavers="all" -Dthreads=false -Dbindings="wasm_beta" -Dengines="sw" --cross-file /tmp/.wasm_cross.txt build_wasm
    fi
fi
```

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -327,14 +360,20 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
...
+    unique_ptr<Canvas>   canvas = nullptr;
     unique_ptr<Animation>  animation = nullptr;
     string                 data;
     uint8_t*               buffer = nullptr;
     uint32_t               width = 0;
     uint32_t               height = 0;
     float                  psize[2];         //picture size
     bool                   updated = false;
+
+    #ifdef THORVG_WG_RASTER_SUPPORT
```
hermet, line 372:
```
#ifdef THORVG_SW_RASTER_SUPPORT
    uint8_t*               buffer = nullptr;
    uint32_t               width = 0;
#endif
    WGPUInstance instance{};
    WGPUSurface surface{};
#endif

File: src/bindings/wasm/tvgWasmLottieAnimation.cpp [hunk 4]
```diff
@@ -312,13 +336,22 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
     explicit TvgLottieAnimation()
```
hermet, line 336: Time to add an engine specifier by users:

```
static TvgLottieAnimation* create(string engine = "sw")
{
  if (sw) engine = tvg::CanvasEngine::Sw
  if (wg) engine = tvg::CanvasEngine::Wg
}
```

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -74,7 +81,7 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
     }
 
     // Render methods
-    bool load(string data, string mimetype, int width, int height, string rpath = "")
+    bool load(string data, string mimetype, int width, int height, string selector = "", string rpath = "")
```
hermet, line 84: The selector is not compatible with other engines. Could you clarify what values the selector is supposed to receive from the user? Otherwise we can ignore it if it's an optional.

hermet (commented):
Thanks Please check comments.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
tinyjin: Done. however width, height are still needed for resize function to check whether size changed

src/bindings/wasm/tvgWasmLottieAnimation.cpp
tinyjin: I assume this interface doesn't make sense, because we'll have 2 WASM binaries(one is SW, the other is WG). Then the engine has to be selected by given WASM binary.

This interface might give users unmatched parameter case:
WebGPU binary imported, but users create SW Canvas.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
tinyjin: The value `selector` has to be HTML element selector.

This is required by `WGPUSurfaceDescriptorFromCanvasHTMLSelector`, to initialize browser based WGPU surface, which means this is not an optional.

In SW Engine, no need to have selector value. We can skip or pass and forget it.

hermet: @tinyjin 
1. We don't use the selector in the wgpu native systems.
2. What string value is the user supposed to send?

src/bindings/wasm/tvgWasmLottieAnimation.cpp
tinyjin: We can still choose and combine all engines with single package in this interface, but we'll have separated WASM binary.

hermet: > we'll have separated WASM binary.

It's inefficient. ThorVG already encapsulated the backend engine. Player binary doesn't need to be duplicated. User only need a single wasm player.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
tinyjin: 1. This is part of WGPU surface creation:

```cpp
WGPUSurfaceDescriptorFromCanvasHTMLSelector canvasDesc{};
canvasDesc.chain.next = nullptr;
canvasDesc.chain.sType = WGPUSType_SurfaceDescriptorFromCanvasHTMLSelector;
canvasDesc.selector = this->selector.c_str(); // This requires selector
WGPUSurfaceDescriptor surfaceDesc{};
surfaceDesc.nextInChain = &canvasDesc.chain;
surface = wgpuInstanceCreateSurface(instance, &surfaceDesc);
```

2. Target Canvas that we want to draw on

```html
<canvas class="thorvg" style="width: 100%; height: 100%;" />
```

With this canvas, users should pass `.thorvg` value, which is class selector.
Please Note, binding users should pass the data, but library users don't have to handle this value, because JavaScript side will cover the logic(canvas creation, passing selector).

src/bindings/wasm/tvgWasmLottieAnimation.cpp
tinyjin: OK, then we can have binaries like this.
- SW Binary
- WG Binary
- Combined Binary

Do we aim to support Combined binary on this patch? or should we introduce only one binary from the beginning?

hermet: Have a see this ultimate target:
+ Single Player Binary (tvgWasm + ThorVG Renderer + Lottie Loader + Image Loaders)
+ Dynamic Module (sw engine)
+ Dynamic Module (wg engine)

hermet: We will address the runtime loading for backend engines later.

tinyjin: I see, we can have Single Player Binary as a main module, and the others in side module. Emscripten supports compiling like this.

By the way, I never considered yet of this combining situation, thus current each binaries are compiled separately.
We should probably have PoC for binary separation first. so I'm wondering if this patch has to include this separation? Or we can introduce optimal binary structure in next optimization?

hermet: > By the way, I never considered yet of this combining situation, thus current each binaries are compiled separately.
> We should probably have PoC for binary separation first. so I'm wondering if this patch has to include this separation? Or we can introduce optimal binary structure in next optimization?

@tinyjin  No worries for now. Emscripten allows it, and we can explore it later. Our current main focus should be the API for those extensions without breaking the app usage.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -35,13 +38,24 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
...
+            }
+        #endif
         Initializer::term();
     }
 
-    static TvgLottieAnimation* create()
+    explicit TvgLottieAnimation(string engine = "sw", string selector = "")
     {
-        return new TvgLottieAnimation;
+        if (engine == "sw") this->engine = tvg::CanvasEngine::Sw;
```
hermet, line 53: Please check this:
```
if (engine =="sw") {
#ifdef THORVG_SW_RASTER_SUPPORT
    ...
#else 
    this->errorMsg = "Invalid engine";
#endif
} else if (engine == "wg") {
    same to sw.
}

**Review by hermet — COMMENTED** [review]
@tinyjin thanks, please check comments.

File: src/bindings/wasm/tvgWasmLottieAnimation.cpp [hunk 7]
```diff
@@ -35,13 +38,58 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
...
         Initializer::term();
     }
 
-    static TvgLottieAnimation* create()
+    explicit TvgLottieAnimation(string engine = "sw", string selector = "")
     {
-        return new TvgLottieAnimation;
+        errorMsg = NoError;
+
+        if (engine == "sw") {
```
hermet, line 59: 59 ~ 71 looks unnecessary.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -35,13 +38,58 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
...
+        this->selector = selector;
+        
+        if (Initializer::init(0, this->engine) != Result::Success) {
+            errorMsg = "init() fail";
+            return;
+        }
+
+        if (this->engine == tvg::CanvasEngine::Sw) {
+        #ifdef THORVG_SW_RASTER_SUPPORT
+            canvas = SwCanvas::gen();
```
hermet, line 82: add here instead of 59 ~ 71
```
this->engine = tvg::CanvasEngine::Sw;
```

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -35,13 +38,58 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
...
+        #endif
+        } else if (this->engine == tvg::CanvasEngine::Wg) {
+        #ifdef THORVG_WG_RASTER_SUPPORT
+            canvas = WgCanvas::gen();
```
hermet, line 86:
```
this->engine = tvg::CanvasEngine::Wg;
```

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -35,13 +38,58 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
...
+        } else this->errorMsg = "Invalid engine";
+
+        this->selector = selector;
+        
+        if (Initializer::init(0, this->engine) != Result::Success) {
+            errorMsg = "init() fail";
+            return;
+        }
+
+        if (this->engine == tvg::CanvasEngine::Sw) {
```
hermet, line 80: if (engine == "sw") {

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -85,8 +133,20 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
             return false;
         }
 
-        //back up for saving
-        this->data = data;
+        this->data = data; //back up for saving
+
+        if (engine == tvg::CanvasEngine::Wg) {
```
hermet, line 138: I think this WG initialization should be at creation time. Please double-check this.
```
explicit TvgLottieAnimation(string engine = "sw", string selector = "")
```

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -126,7 +186,13 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
 
         if (!canvas || !animation) return val(typed_memory_view<uint8_t>(0, nullptr));
 
-        if (!updated) return val(typed_memory_view(width * height * 4, buffer));
+        if (!updated) {
+            if (engine == tvg::CanvasEngine::Sw) {
```
hermet, line 190: let's make a private function
```
val output()
{
    if (engine == tvg::CanvasEngine::Sw) {
    #ifdef THORVG_SW_RASTER_SUPPORT
        return val(typed_memory_view(width * height * 4, buffer));
    #endif
    } 
    return val(typed_memory_view<uint8_t>(0, nullptr));
}
```

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -137,7 +203,11 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
 
         updated = false;
 
-        return val(typed_memory_view(width * height * 4, buffer));
+        if (engine == tvg::CanvasEngine::Sw) {
```
hermet, line 206:
```
return output();
```

src/bindings/wasm/tvgWasmLottieAnimation.cpp
tinyjin: @hermet Can we generate canvas before `init()`?

I was thiking `Initializer::init(0, this->engine)` can't be called if we move the line into canvas creation part.

hermet: Yes, it should be. You can move it. I believe even it's compact and simpler code.

hermet (approved):

PR #2627, opened 2024-08-09
lottie: handle edge case for rounded rect

mgrudzinska:
For rounded rectangles the roundness value should
be determined using to the formula:
r = max(min(size.x/2, r), min(size.y/2, r))
rather than the previous method:
r = min(size.x/2, size.y/2, r)

sample:
[roundedRect.json](https://github.com/user-attachments/files/16564597/roundedRect.json)

before:
Zrzut ekranu 2024-08-9 o 18 45 44

after:
Zrzut ekranu 2024-08-9 o 18 51 37

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -472,11 +472,8 @@ void LottieBuilder::updateRect(LottieGroup* parent, LottieObject** child, float 
     auto roundness = rect->radius(frameNo, exps);
     if (ctx->roundness > roundness) roundness = ctx->roundness;
 
-    if (roundness > ROUNDNESS_EPSILON) {
-        if (roundness > size.x * 0.5f)  roundness = size.x * 0.5f;
-        if (roundness > size.y * 0.5f)  roundness = size.y * 0.5f;
-    }
-
+    if (roundness > ROUNDNESS_EPSILON) roundness = std::min(roundness, std::max(size.x, size.y) * 0.5f);
+    
     if (!ctx->repeaters.empty()) {
         auto shape = rect->pooling();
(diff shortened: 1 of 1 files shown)


mgrudzinska:
discussion: https://github.com/lottie/lottie-spec/issues/82

hermet:
skottie: 
image

mgrudzinska:
@hermet I saw results from skottie, but I think theirs implementation is not perfect.
I prefer to compare with AE:

Zrzut ekranu 2024-08-10 o 14 16 37

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -534,11 +534,8 @@ static void _updateRect(LottieGroup* parent, LottieObject** child, float frameNo
     auto roundness = rect->radius(frameNo, exps);
     if (ctx->roundness > roundness) roundness = ctx->roundness;
 
-    if (roundness > ROUNDNESS_EPSILON) {
-        if (roundness > size.x * 0.5f)  roundness = size.x * 0.5f;
-        if (roundness > size.y * 0.5f)  roundness = size.y * 0.5f;
-    }
-
+    if (roundness > ROUNDNESS_EPSILON) roundness = std::max(std::min(roundness, size.x * 0.5f), std::min(roundness, size.y * 0.5f));
```
hermet, line 537: is this better?
```roundness = min(roundness, max(size.x, size.y) * 0.5f);```

hermet:
> I saw results from skottie, but I think theirs implementation is not perfect.
> I prefer to compare with AE:

@mgrudzinska Sure, it needs to be compared with AE. I've just left the problematic Skottie here for context and historical reference.

PR #2630, opened 2024-08-11
wg_engine: shaders refactoring

SergeyLebedkin:
* used precalculated gradient texture instead of per-pixel gradient map computation
* used HW wrap samples for fill spread setting
* used single shader module for composition instead of signle module per composition type
* used single shader module for blending for each of fill type (solid, gradient, image) instaed of signle module per blend type
* much easier to add new composition and blend equations
* get rided std::string uasge
* shaders code is more readable

Verified on windows/linux/macos/emsdk

Samples to verify:
* Blending
* MaskingMethods
* LinearGradient
* RadialGradient


mgrudzinska (approved):
@SergeyLebedkin  please keep the convention with 1-line commit title and after that commit msg

hermet (approved):

Issue #2593, opened 2024-07-24
lottie: fill alpha component support


hermet:
Filling with transparency isn't taken into account? 
It should be invisible since the opacity is zero.

[test.json](https://github.com/user-attachments/files/16359933/test.json)

mgrudzinska:
@hermet can you double check the file? in this one opacity is 100, after changing to 0 nothing is rendered, as expected

hermet:
@mgrudzinska Yes, it looks correct. @theashraf We reviewed this in ThorVG in detail. There is no data with zero opacity. It looks fine. Please let us know if something is wrong. \+ Note that Skottie has a bug related to this.

theashraf:
@hermet In the provided asset, the Fill shape color value has 0 opacity but is still rendered. I wasn't referring to the Fill shape opacity; I was referring to the Fill shape color value opacity.

hermet:
@theashraf The value is `s: [offset, r, g, b]`.  Opacity is separately handled in gradient fill.

theashraf:
@hermet Just to confirm, is the color value handled as [r,g,b] only in ThorVG, with the alpha being ignored?
https://lottiefiles.github.io/lottie-docs/concepts/#colors

hermet:
@theashraf It supports when opacity value is given s:[offset, a]

theashraf:
@hermet like this ? 

```json
 "c": {
            "a": 1,
            "k": [
              {
                "o": {
                  "x": 0.167,
                  "y": 0.167
                },
                "i": {
                  "x": 0.833,
                  "y": 0.833
                },
                "s": [
                  0, // offset
                  1, // red
                  0.5569, // green
                  0.9255, // blue
                  0 // opacity
                ],
                "t": 0
              },
              {
                "s": [
                  0, // offset
                  0.5569, // red
                  0.5647, // green
                  1, // blue
                  0 // opacity
                ],
                "t": 90
              }
            ]
          },
```

hermet:
@theashraf Ah, sorry, my brain dead. So this is not a gradient fill but a solid fill. It's `s:[r, g, b, a]` you're correct. Currently, the last component (a) is ignored by ThorVG. lottie-player web seems to be doing the same.

theashraf:
@hermet, There are no issues, I want to ensure we are aligned on the Lottie format and how ThorVG handles this case, given that ThorVG adheres strictly to the Lottie file format standards

hermet:
[...]

PR #2631, opened 2024-08-12
lottie: support alpha component of fill

hermet:
issue: https://github.com/thorvg/thorvg/issues/2593


hermet:
Mark as DRAFT. I think this is an unnecessary overspec.

PR #2637, opened 2024-08-14
common: spec out TexMap

hermet:
Spec out this incomplete experimental feature,
this is a still promising one, we will reintroduce this officially after 1.0 release

size: -2kb


SergeyLebedkin (approved):

hermet:
I’m not sure where this feature is currently used. It will be applied in 2 weeks unless there are any objections.

projectitis:
Just checking - so does this mean that TVG would not support textures any more? Or will there be an alternative replacement?

hermet:
> Just checking - so does this mean that TVG would not support textures any more? Or will there be an alternative replacement?

@projectitis Hi, we are preparing for the official v1 release and are currently cleaning out the unstable or incomplete features (tagged as Experimental). We can keep the Experimental APIs that are useful to users, but we will remove those that are not being used practically. As far as I know, the texture-mapping feature is only used in your POC and is not being utilized elsewhere.

I understand that this feature will be ultimately necessary for ThorVG, but we are not ready to support it at this time. Currently, I suggest reintroducing it in ThorVG v2 release when we are fully prepared for it.

projectitis:
ok. That makes sense.

hermet:
> It is used for the TVG Rive renderer (Rive Tizen renderer). However, that project has stalled.

Please note that the latest version, v0.14.5, which includes the texmap feature.

PR #2638, opened 2024-08-14
wg_engine: composition and blend optimization - PART 2

SergeyLebedkin:
wg_engine: aabb based masking optimization

* used fragment shaders for mask applience instead of compute shaders
* less render targets swithing
* aabb based mask application. Masks computes only on the area with shapes, but not full screen
* shape aabb based on transformed shape bbox

Ready for direct mask application (part 3)

Samples to verify:
* Masking
* MaskingMethods

### DIFF
--- a/src/renderer/wg_engine/tvgWgCompositor.cpp
+++ b/src/renderer/wg_engine/tvgWgCompositor.cpp
@@ -47,11 +47,17 @@ void WgCompositor::initialize(WgContext& context, uint32_t width, uint32_t heigh
     storageClipPath.initialize(context, width, height);
     storageInterm.initialize(context, width, height);
     storageDstCopy.initialize(context, width, height);
+    // composition and blend geometries
+    WgGeometryData geometryData;
+    geometryData.appendBlitBox();
+    meshData.update(context, &geometryData);
 }
 
 
 void WgCompositor::release(WgContext& context)
 {
+    // composition and blend geometries
+    meshData.release(context);
     // release intermediate render storages
     storageInterm.release(context);
     storageDstCopy.release(context);
@@ -252,6 +258,7 @@ void WgCompositor::blendImage(WgContext& context, WgRenderDataPicture* renderDat
 };
 
 
+// TODO: use direct mask applience
 void WgCompositor::composeShape(WgContext& context, WgRenderDataShape* renderData, WgRenderStorage* mask, CompositeMethod composeMethod)
 {
     assert(mask);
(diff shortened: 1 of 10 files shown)


hermet:
@SergeyLebedkin Hello, how % can be improved in your local test?

SergeyLebedkin:
> @SergeyLebedkin Hello, how % can be improved in your local test?

**From 170 fps to 200 fps on masking test (15%)**
The reason is that Scene Blending still works on compute shaders, but not on fragments and apllies for full screen
And also masking can be applied directly during shape rendering (any type), but this is a PART 3 stage
So, in the next step we will have 2 options:
1. Direct mask application
2. Scene Blending move to fragment shaders (based on aabb area)
It will also increase perofmance to the theoretical limit for out workflows

mgrudzinska:
@SergeyLebedkin the Lottie example crashed when built with thread sanitizer (on mac)

mgrudzinska:
> @SergeyLebedkin the Lottie example crashed when built with thread sanitizer (on mac)

correction - crash is present also on main

tinyjin:
CleanShot 2024-08-16 at 13 57 36@2x

Hello, sharing performance result in Web Browser.
This patch indicates around 65FPS of 120, gained 37%+ improvement than previous build (was 41FPS).

Memory Usage is the same.

src/renderer/wg_engine/tvgWgRenderData.cpp
```diff
@@ -306,6 +306,18 @@ void WgRenderDataShape::updateBBox(WgPoint pmin, WgPoint pmax)
 }
 
 
+void WgRenderDataShape::updateAABB(const Matrix& rt) {
+    WgPoint p0 = WgPoint(pMin.x, pMin.y).trans(rt);
+    WgPoint p1 = WgPoint(pMax.x, pMin.y).trans(rt);
+    WgPoint p2 = WgPoint(pMin.x, pMax.y).trans(rt);
+    WgPoint p3 = WgPoint(pMax.x, pMax.y).trans(rt);
+    aabb.x = std::min(p0.x, std::min(p1.x, std::min(p2.x, p3.x)));
```
mgrudzinska, line 314: ```aabb.x = std::min({p0.x, p1.x, p2.x, p3.x});```
same the rest

src/renderer/wg_engine/tvgWgCompositor.cpp
```diff
@@ -268,16 +275,22 @@ void WgCompositor::composeShape(WgContext& context, WgRenderDataShape* renderDat
     beginRenderPass(commandEncoder, &storageInterm, true);
     drawShape(context, renderData, WgPipelineBlendType::Custom);
     endRenderPass();
-    // run compisition
-    compose(commandEncoder, &storageInterm, mask, target, composeMethod);
     // restore current render pass
     beginRenderPass(commandEncoder, target, false);
+    // apply mask
```
mgrudzinska, line 280: I'm confused here. this function is used only for clipping, so why this comment? also the arg name 'mask'? should it be clip? composite? 
Could it just be a matter of nomenclature? is it consistent with the sw engine?

SergeyLebedkin: There are a different nameing convention and it will be handled in the future and resolved

mgrudzinska (approved):
these should be removed from commit msg 
```Samples to verify:
* Masking
* MaskingMethods```

File: src/renderer/wg_engine/tvgWgCompositor.cpp [hunk 3]
```diff
@@ -268,16 +275,22 @@ void WgCompositor::composeShape(WgContext& context, WgRenderDataShape* renderDat
...
+    wgpuRenderPassEncoderSetScissorRect(renderPassEncoder,0, 0, width, height);
```
hermet, line 281: This code snippet seems duplicated four times. It could be refactored into a common function for reuse.
```
WgCompositor::applyMask(...)
{
    wgpuRenderPassEncoderSetStencilReference(renderPassEncoder, 0);
    wgpuRenderPassEncoderSetBindGroup(renderPassEncoder, 0, storageInterm.bindGroupTexure, 0, nullptr);
    wgpuRenderPassEncoderSetBindGroup(renderPassEncoder, 1, mask->bindGroupTexure, 0, nullptr);
    wgpuRenderPassEncoderSetPipeline(renderPassEncoder, pipelines->sceneComp[(uint32_t)composeMethod]);
    meshData.drawImage(context, renderPassEncoder);
}
```

SergeyLebedkin: Make sence, Fixed

hermet (approved):

Issue #1496, opened 2023-06-14
Revise tvg::ComposteMethod Interface


hermet:
This enum class is messed up with different kind of composition.
It's time to revise the interface before 1.0 release....

We can separate the ClipPath from the CompositeMethod, with introducing clip().
```
**tvg::CompositeMethod::ClipPath**
-> tvg::Result Paint::clip(std::unique_ptr<Shape> clipper);
```

PR #2643, opened 2024-08-16
Introduced Paint::clip() API

hermet:
Separate clip function from the Composite() clipping and composition can be used together.

This helps avoid the introduction of nested scenes  when composition and clipping overlap.

Deprecated:
```
enum class CompositeMethod::ClipPath
enum Tvg_Composite_Method::TVG_COMPOSITE_METHOD_CLIP_PATH
```

Experimental API:
```
- Result Paint::clip(std::unique_ptr<Paint> clipper)
- Tvg_Result tvg_paint_set_clip(Tvg_Paint* paint, Tvg_Paint* clipper)
```

Issue: https://github.com/thorvg/thorvg/issues/1496

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -272,8 +272,7 @@ static void _applyComposition(SvgLoaderData& loaderData, Paint* paint, const Svg
             if (valid) {
                 Matrix finalTransform = _compositionTransform(paint, node, compNode, SvgNodeType::ClipPath);
                 comp->transform(finalTransform);
-
-                paint->composite(std::move(comp), CompositeMethod::ClipPath);
+                paint->clip(std::move(comp));
             }
 
             node->style->clipPath.applying = false;
@@ -714,7 +713,6 @@ static Matrix _calculateAspectRatioMatrix(AspectRatioAlign align, AspectRatioMee
 
 static unique_ptr<Scene> _useBuildHelper(SvgLoaderData& loaderData, const SvgNode* node, const Box& vBox, const string& svgPath, int depth, bool* isMaskWhite)
 {
-    unique_ptr<Scene> finalScene;
     auto scene = _sceneBuildHelper(loaderData, node, vBox, svgPath, false, depth + 1, isMaskWhite);
 
     // mUseTransform = mUseTransform * mTranslate
@@ -751,9 +749,7 @@ static unique_ptr<Scene> _useBuildHelper(SvgLoaderData& loaderData, const SvgNod
         mSceneTransform = mUseTransform * mSceneTransform;
         scene->transform(mSceneTransform);
 
-        if (node->node.use.symbol->node.symbol.overflowVisible) {
-            finalScene = std::move(scene);
-        } else {
+        if (!node->node.use.symbol->node.symbol.overflowVisible) {
             auto viewBoxClip = Shape::gen();
             viewBoxClip->appendRect(0, 0, width, height, 0, 0);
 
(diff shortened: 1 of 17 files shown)


hermet:
@SergeyLebedkin please review this, "CompositeMethod::ClipPath" is deprecated. wgpu must be updated as well.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -764,14 +763,10 @@ static unique_ptr<Scene> _useBuildHelper(SvgLoaderData& loaderData, const SvgNod
...
-            compositeLayer->composite(std::move(viewBoxClip), CompositeMethod::ClipPath);
-            compositeLayer->push(std::move(scene));
-
-            auto root = Scene::gen();
-            root->push(std::move(compositeLayer));
+            scene->clip(std::move(viewBoxClip));
 
-            finalScene = std::move(root);
+            finalScene = Scene::gen();
+            finalScene->push(std::move(scene));
```
JSUYA, line 769: I don't think `finalScene` is necessary since the composition work for the scene has been simplified.
How about remove `finalScene`? When I tested it, the result was visually fine.

```diff
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -713,7 +713,6 @@ static Matrix _calculateAspectRatioMatrix(AspectRatioAlign align, AspectRatioMee

 static unique_ptr<Scene> _useBuildHelper(SvgLoaderData& loaderData, const SvgNode* node, const Box& vBox, const string& svgPath, int depth, bool* isMaskWhite)
 {
-    unique_ptr<Scene> finalScene;
     auto scene = _sceneBuildHelper(loaderData, node, vBox, svgPath, false, depth + 1, isMaskWhite);

     // mUseTransform = mUseTransform * mTranslate
@@ -750,9 +749,7 @@ static unique_ptr<Scene> _useBuildHelper(SvgLoaderData& loaderData, const SvgNod
         mSceneTransform = mUseTransform * mSceneTransform;
         scene->transform(mSceneTransform);

-        if (node->node.use.symbol->node.symbol.overflowVisible) {
-            finalScene = std::move(scene);
-        } else {
+        if (!node->node.use.symbol->node.symbol.overflowVisible) {
             auto viewBoxClip = Shape::gen();
             viewBoxClip->appendRect(0, 0, width, height, 0, 0);

@@ -764,16 +761,12 @@ static unique_ptr<Scene> _useBuildHelper(SvgLoaderData& loaderData, const SvgNod
             viewBoxClip->transform(mClipTransform);

             scene->clip(std::move(viewBoxClip));
-
-            finalScene = Scene::gen();
-            finalScene->push(std::move(scene));
         }
     } else {
         scene->transform(mUseTransform);
-        finalScene = std::move(scene);
     }

-    return finalScene;
+    return scene;
 }

```

src/renderer/tvgPaint.cpp
```diff
@@ -433,15 +441,27 @@ Paint* Paint::duplicate() const noexcept
...
 
-Result Paint::composite(std::unique_ptr<Paint> target, CompositeMethod method) noexcept
+Result Paint::clip(std::unique_ptr<Paint> clipper) noexcept
 {
-    if (method == CompositeMethod::ClipPath && target && target->type() != Type::Shape) {
-        TVGERR("RENDERER", "ClipPath only allows the Shape!");
+    auto p = clipper.release();
+
+    if (p && p->type() != Type::Shape) {
+        TVGERR("RENDERER", "ClipPath only supports the Shape!");
```
mgrudzinska, line 449: maybe we could avoid using ClipPath name

mgrudzinska (commented):
1. just as a note: 
- ClipPath is still used in few places (tvgScene.h, tvgPicture.cpp, tvgShape.h, tvgSwRaster.cpp) without this comment "//TODO: remove. Keep this for the backward compatibility"
- also used in logs in tvgSwRenderer.cpp, tvgSwRle.cpp - these could be remove now

2. if rle clip goes beyond canvas it is ignored and everything is drawn like is was not used:
```
        shape->appendRect(0, 0, 800, 800);
        shape->fill(255, 0, 255);

        clip->appendCircle(900, 600, 100, 100); // ignored
        //clip->appendCircle(899, 600, 100, 100); // works

        shape->clip(std::move(clip));
```
but the same is on main, so maybe it should be opened as an issue

mgrudzinska (approved):
approved since earlier mentioned things were more comments than issues

hermet:
> but the same is on main, so maybe it should be opened as an issue

@mgrudzinska could you please add an issue of it?

hermet:
Perhaps we can consider promoting clip() to the Scene level.

PR #2644, opened 2024-08-16
examples: capi example ++

mgrudzinska:
Added loading animation from a memory.

### DIFF
--- a/examples/Capi.cpp
+++ b/examples/Capi.cpp
@@ -201,15 +201,35 @@ void contents()
     }
 
 //////6. Animation with a picture
+    //instead loading from a memory, an animation can be loaded directly from a file using:
+    //tvg_picture_load(pict_lottie, EXAMPLE_DIR"/lottie/sample.json")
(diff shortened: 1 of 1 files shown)


examples/Capi.cpp
```diff
@@ -201,16 +202,46 @@ void contents()
...
+    if (fread(data, 1, data_size, file) != (size_t)data_size) {
+        free(data);
+        fclose(file);
+        return;
+    }
+    data[data_size++] = '\0';
+
+    animation2 = tvg_animation_new();
+    pict_lottie = tvg_animation_get_picture(animation2);
+    if (tvg_picture_load_data(pict_lottie, data, data_size, nullptr, nullptr, false) != TVG_RESULT_SUCCESS) {
```
hermet, line 234: Suggestion: we could replace the original animation sample with this one and keep the API usage as a comment.
```
//You can directly load an animation from a file
//if (tvg_picture_load(pict_lottie, EXAMPLE_DIR"/lottie/sample.json") != TVG_RESULT_SUCCESS) {
```

hermet (approved):

PR #2645, opened 2024-08-16
docs: description++

mgrudzinska:


inc/thorvg.h
```diff
@@ -1286,7 +1286,7 @@ class TVG_API Picture final : public Paint
      * when the @p copy has @c false. This means that loading the same data again will not result in duplicate operations
      * for the sharable @p data. Instead, ThorVG will reuse the previously loaded picture data.
      *
-     * @param[in] data A pointer to a memory location where the content of the picture file is stored.
+     * @param[in] data A pointer to a memory location where the content of the picture file is stored, null-terminated for non-binary data.
```
hermet, line 1289: Let's carefully consider this. It likely applies only when copy == false.

mgrudzinska: in lottie for now we need that in both cases

hermet: @mgrudzinska Yes. I know that. just considering add `\0` thorvg inside in case of copy == true.

inc/thorvg.h
```diff
@@ -1286,7 +1286,7 @@ class TVG_API Picture final : public Paint
      * when the @p copy has @c false. This means that loading the same data again will not result in duplicate operations
      * for the sharable @p data. Instead, ThorVG will reuse the previously loaded picture data.
      *
-     * @param[in] data A pointer to a memory location where the content of the picture file is stored.
+     * @param[in] data A pointer to a memory location where the content of the picture file is stored (null-terminated for non-binary data if @p copy is @c false).
```
hermet, line 1289: I think we can clarify the sentence more. `A null-terminated string is expected for non-binary data if @p copy is @c false`

hermet (approved):

PR #2656, opened 2024-08-20
wasm: code refactoring

hermet:
Introduced the abstract EngineMethod to handle
different engines more effectively using polymorphism.


src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -84,11 +104,40 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
...
+    }
+
+    explicit TvgLottieAnimation(string engine = "sw", string selector = "")
+    {
+        errorMsg = NoError;
+
+        if (engine == "sw") this->engine = new TvgSwEngine;
+        else if (engine == "wg") this->engine = new TvgWgEngine;
+
+        canvas = this->engine->init(selector);
```
hermet, line 135: just noticed:
```
if (engine) {
   canvas = this->engine->init(selector);
}

Comment (tinyjin) [reply src/bindings/wasm/tvgWasmLottieAnimation.cpp:135]: How about this code?

```cpp
if (!engine) {
    errorMsg = "Invalid engine";
    return;
}
```

hermet: @tinyjin Let's simplify it, We can regards canvas / engine as one.

PR #2658, opened 2024-08-20
lottie: fix text update

mgrudzinska:
The occurrence of a 'carriage return' (13) or 'end of text' (3) caused the skipping of the next character check, immediately searching for it in the list of available characters. If the next character was also 13 or 3, it led to incorrect interpretation; however, if it was the last character in the sequence, a crash occurred.

modified sample (line 108: ```"t": "START\r\rSTART",```:
[text.json](https://github.com/user-attachments/files/16674490/text.json)

before (missing one \r):
Zrzut ekranu 2024-08-20 o 12 42 11

after:
Zrzut ekranu 2024-08-20 o 12 41 08

for line 108: ```"t": "START\r",``` crashed occured before this change

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -1055,6 +1055,7 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
             scene = Scene::gen();
             cursor.x = 0.0f;
             cursor.y = ++line * (doc.height / scale);
+            continue;
         }
         //find the glyph
         bool found = false;


tinyjin:
I have some resources to validate `updateText` logic. I'm going to check those files, then come back :)

tinyjin:
Have just finished test with some advanced features. (Range Selector, Tracking). Works well.

I see Mira corrected this logic. There is no need to get in the loop finds glyphs(:1062) for `\r`, `\n`, because Lottie's [character data](https://lottiefiles.github.io/lottie-docs/text/#character-data) doesn't have information about 'carriage return' (13) and 'end of text' (3).

In other words, both carriage return and end of text are not glpyh. They're just control characters, not printable.
Hence, this is correct logic to go.

Even it came with crash or incorrect rendering result by this point:

```cpp
// Here `p` pointer already increased when it's `\r` and `\n`
if (*p == 13 || *p == 3 || *p == '\0') {
  // ...
  if (*p == '\0') break;
  ++p;
  // ...
}

// For `\r` and `\n`, `found` is always `false`
// This logic increased `p` pointer by double
if (!found) {
    ++p; // ⚠️ doubled
    ++idx;
}
```

tinyjin (approved):

hermet (approved):

PR #2660, opened 2024-08-21
sw_engine: fix TexmapPolygon rastering

mgrudzinska:
The y range initialization was missing a check to ensure that the height is a positive value. This could lead to an attempt to call malloc with a negative argument, which cast to an unsigned value, caused a crash.

### DIFF
--- a/src/renderer/sw_engine/tvgSwRasterTexmap.h
+++ b/src/renderer/sw_engine/tvgSwRasterTexmap.h
@@ -53,12 +53,10 @@ static bool _arrange(const SwImage* image, const SwBBox* region, int& yStart, in
         regionBottom = image->rle->spans[image->rle->size - 1].y;
     }
 
-    if (yStart >= regionBottom) return false;
-
     if (yStart < regionTop) yStart = regionTop;
     if (yEnd > regionBottom) yEnd = regionBottom;
 
-    return true;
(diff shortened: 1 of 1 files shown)


src/renderer/sw_engine/tvgSwRasterTexmap.h
```diff
@@ -58,6 +58,7 @@ static bool _arrange(const SwImage* image, const SwBBox* region, int& yStart, in
     if (yStart < regionTop) yStart = regionTop;
     if (yEnd > regionBottom) yEnd = regionBottom;
 
+    if (yEnd - yStart <= 0) return false;
```
hermet, line 61: I think this condition is enough and we can remove the 56 line condition.
```
if (yEnd >= yStart) return false;
```

hermet: 1. 56 line is duplicated for this exceptional handling.
2. yEnd == yStart is a single line. No need to performance AA.

mgrudzinska: please check one more time, after a minute I push changes

mgrudzinska: l.56 removed and the current return is false for end=start, so should be fine

PR #2661, opened 2024-08-21
wasm: code refactoring

tinyjin:
Original patch from #2656 by @hermet 
- Added engine validating condition in constructor (As we talked [here](https://github.com/thorvg/thorvg/pull/2656#discussion_r1724426628))
- Enhanced destructor logic by adding explicit engine termination

---

Introduced the abstract EngineMethod to handle
different engines more effectively using polymorphism.

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -33,46 +33,66 @@ using namespace tvg;
 
 static const char* NoError = "None";
 
-class __attribute__((visibility("default"))) TvgLottieAnimation
+struct TvgEngineMethod
 {
-public:
-    ~TvgLottieAnimation()
+    virtual ~TvgEngineMethod() {}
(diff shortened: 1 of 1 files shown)


src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -33,46 +33,66 @@ using namespace tvg;
...
-        #endif
-        } else if (engine == tvg::CanvasEngine::Wg) {
         #ifdef THORVG_WG_RASTER_SUPPORT
             wgpuSurfaceRelease(surface);
             wgpuInstanceRelease(instance);
         #endif
-        }
-
-        Initializer::term();
+        Initializer::term(tvg::CanvasEngine::Wg);q
```
hermet, line 90: q typo?

tinyjin: Oops, it comes when i'm committing. Have just removed!

hermet (approved):

PR #2673, opened 2024-08-23
loader/ttf: Italic font of updated text

### LINKED ISSUE #2669: loader/ttf: Italic font of updated text
Last changed text font italic attribute is propagating on all texts using same loader.

Example (Modified examples/Text.cpp part):
```
        auto text = tvg::Text::gen();
        text->font("Arial", 80);
        text->text("THORVG Text");
        text->fill(255, 255, 255);

        auto text2 = tvg::Text::gen();
        text2->font("Arial", 30, "italic");
        text2->text("Font = \"Arial\", Size = 40, Style=Italic");
        text2->translate(0, 150);
        text2->fill(255, 255, 255);
        canvas->push(std::move(text2));

        canvas->update();
```

Output:

Expected output:

It looks like `italic` attribute of `TtfLoader` should be updated before every `TtfLoader::resize` call. Maybe move it to the resize method parameter?

lpogic:
The change prevents invalid input from occurring during text shape transformation.

I replaced the `resize` method with a `transform` because the text case does not fit the interface at all.

Closes #2669

### DIFF
--- a/src/loaders/ttf/tvgTtfLoader.cpp
+++ b/src/loaders/ttf/tvgTtfLoader.cpp
@@ -208,16 +208,20 @@ void TtfLoader::clear()
 /* External Class Implementation                                        */
 /************************************************************************/
 
-bool TtfLoader::transform()
+
+bool TtfLoader::resize(Paint* paint, float sx, TVG_UNUSED float sy)
 {
-    if (!shape) return false;
+    if (!paint) return false;
     auto shift = 0.0f;
+    auto dpi = 96.0f / 72.0f;   //dpi base?
+    scale = sx * dpi / reader.metrics.unitsPerEm;
     if (italic) shift = -scale * 0.18f;  //experimental decision.
     Matrix m = {scale, shift, -(shift * reader.metrics.minw), 0, scale, 0, 0, 0, 1};
-    shape->transform(m);
+    paint->transform(m);
 
     return true;
 }
+
 
 TtfLoader::TtfLoader() : FontLoader(FileType::Ttf)
 {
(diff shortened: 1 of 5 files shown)


src/loaders/ttf/tvgTtfLoader.cpp
```diff
@@ -208,21 +208,17 @@ void TtfLoader::clear()
...
-    scale = sx * dpi / reader.metrics.unitsPerEm;
     if (italic) shift = -scale * 0.18f;  //experimental decision.
     Matrix m = {scale, shift, -(shift * reader.metrics.minw), 0, scale, 0, 0, 0, 1};
-    paint->transform(m);
+    shape->transform(m);
 
     return true;
 }
 
-
```
hermet, line 225: tvg retains 2 lines space. please do not change this.

src/renderer/tvgText.h
```diff
@@ -98,13 +98,13 @@ struct Text::Impl
     {
         if (!loader) return false;
 
+        loader->request(shape, utf8, fontSize, italic);
```
hermet, line 101: I guess just only relocating this `request()` here from the 103 line will fix your problem. Could you please double-check it?

lpogic: That's right. I just made two steps too far. We can solve this issue this way. I will add another PR which makes cleaner why I decided to add `fontSize` to the `request()`.

hermet: Got it. Could u please update the pr? We can fix the issue anyway.

hermet (approved):

PR #2675, opened 2024-08-23
wg_engine: masking optimization - PART 3

SergeyLebedkin:
Masking workflows optimized:
* used hardware blending stage for scene blending
* used AABB for scene blending
* reduced number of offfscreen buffers coping
* reduced number of render pass switching
* used render pipelines abilities to convert offscreen pixel format to screen pixel format
* removed unused shaders

+15% performance boost in avarage in animations with masks

masking.json animation:
before - 200 fps, after - 235 fps

### DIFF
--- a/src/renderer/wg_engine/tvgWgCompositor.cpp
+++ b/src/renderer/wg_engine/tvgWgCompositor.cpp
@@ -87,6 +87,17 @@ WgPipelineBlendType WgCompositor::blendMethodToBlendType(BlendMethod blendMethod
 }
 
 
+RenderRegion WgCompositor::shrinkRenderRegion(RenderRegion& rect)
+{
+    // cut viewport to screen dimensions
+    int32_t xmin = std::max(0, std::min((int32_t)width, rect.x));
+    int32_t ymin = std::max(0, std::min((int32_t)height, rect.y));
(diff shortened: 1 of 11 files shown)


tinyjin:
CleanShot 2024-08-26 at 14 24 30@2x

Hello, from browser result.

With this patch, the masking animation consistently shows 120FPS.
Previously, was 65FPS, `+46%` improved.

tinyjin (approved):

src/renderer/wg_engine/tvgWgCompositor.cpp
```diff
@@ -258,7 +258,25 @@ void WgCompositor::blendImage(WgContext& context, WgRenderDataPicture* renderDat
 };
 
 
-// TODO: use direct mask applience
+void WgCompositor::blendScene(WgContext& context, WgRenderStorage* src, WgCompose* cmp)
+{
+    assert(currentTarget);
+    // cut viewport to screen dimensions
+    int32_t xmin = std::max(0, std::min((int32_t)width, cmp->aabb.x));
```
mgrudzinska, line 265: maybe some function? exactly the same part of the code to create render region is used in the composeScene()

mgrudzinska (commented):
can you please add some description to the commit msg?

mgrudzinska (approved):

PR #2677, opened 2024-08-24
renderer/text: Text resizing when font style changes

### LINKED ISSUE #2676: renderer/text: Size and font style is not updated
There is no effect when the user tries to update the text font size or style without changing the font file.

Example code:
```
auto text = tvg::Text::gen();
text->font("Arial", 80);
text->font("Arial", 40, "italic");
text->text("THORVG Text");
text->fill(255, 255, 255);
```

Output:

Excepted output:

lpogic:
Closes #2676 

Added `needResize` flag which is set to `true` when text resizing is actually required. The side effect is Canvas updating optimization, when the Text size & style doesn't change.

However, like my previous PR, this one could alternatively be replaced by changing the order of the code lines :)
Let me know if the reordering solution would be preferred.

### DIFF
--- a/src/renderer/tvgText.h
+++ b/src/renderer/tvgText.h
@@ -41,7 +41,9 @@ struct Text::Impl
     char* utf8 = nullptr;
     float fontSize;
     bool italic = false;
-    bool changed = false;
+    
+    bool needRead = false;
+    bool needResize = false;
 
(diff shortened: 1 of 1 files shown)


hermet:
@lpogic thanks, please check this: https://github.com/thorvg/thorvg/pull/2683

Issue #2678, opened 2024-08-25
HSL colors with out of range values are rendered incorrectly


MewPurPur:
https://github.com/user-attachments/assets/bcc1d347-45d6-4c50-acb3-62d4febc4082

As I understand it, only the hue shouldn't be clamped.

mgrudzinska:
@MewPurPur should be solved, if you still see some problem please reopen

PR #2679, opened 2024-08-26
svg_loader: Adjust hsl values out of range

JSUYA:
The range of saturation and brightness values is 0 ~ 100% and range of hue is 0 ~ 360.
If a value greater than 100% is loaded, it will be modified to be 100%.

issue: https://github.com/thorvg/thorvg/issues/2678

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -588,6 +588,11 @@ static bool _hslToRgb(float hue, float saturation, float brightness, uint8_t* re
     float _red = 0, _green = 0, _blue = 0;
     uint32_t i = 0;
 
+    while (hue < 0) hue += 360.0f;
+    hue = fmod(hue, 360.0f);
+    saturation = saturation > 0 ? std::min(saturation, 1.0f) : 0.0f;
+    brightness = brightness > 0 ? std::min(brightness, 1.0f) : 0.0f;
+
     if (tvg::zero(saturation))  _red = _green = _blue = brightness;
     else {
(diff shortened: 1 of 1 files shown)


hermet (approved):

JSUYA:
rebased :)

mgrudzinska (changes requested):
since we are changing this I'd handle all the edge cases:

``` 
saturation = std::max(0.0f, std::min(saturation, 1.0f));
brightness = std::max(0.0f, std::min(brightness, 1.0f));
```

also hue value can be improved. we can use fmod instead of casting into uint, since this like an angle and can be < 0 (l.717)

source: https://www.w3.org/TR/css-color-3/#hsl-color

mgrudzinska:
sample to test:
[hsl.svg.zip](https://github.com/user-attachments/files/16764079/hsl.svg.zip)


expected result:
Zrzut ekranu 2024-08-27 o 16 54 59

mgrudzinska (commented):
@JSUYA I should have left a comment instead of request changes, because this pr is correct, just can be improved, I think. sorry for that.

JSUYA:
> since we are changing this I'd handle all the edge cases:
> 
> ```
> saturation = std::max(0.0f, std::min(saturation, 1.0f));
> brightness = std::max(0.0f, std::min(brightness, 1.0f));
> ```
> 
> also hue value can be improved. we can use fmod instead of casting into uint, since this like an angle and can be < 0 (l.717)
> 
> source: https://www.w3.org/TR/css-color-3/#hsl-color

Thank you for detail. I modified PR. :)

mgrudzinska (approved):

PR #2683, opened 2024-08-27
renderer/text: fix a missing text update

hermet:
previously italic style had been ignored
even if the font style changes.

issue: https://github.com/thorvg/thorvg/issues/2676


lpogic (changes requested):
Works well for italic changes, but font size changes are still ignored:

```
auto text = tvg::Text::gen();
text->font("Arial", 80);
text->font("Arial", 40);
text->text("THORVG Text");
text->fill(255, 255, 255);
```

PR #2689, opened 2024-08-27
lottie: handle trim path edge cases

mgrudzinska:
According to the definition of trim path elements,
the begin and end values are specified as percentages
in the range of 0-100% (this is also confirmed by AE,
where it's not possible to exceed this range).
Added clamping to align with specs.

before:
before

after:
after

sample:
[part.json](https://github.com/user-attachments/files/16788631/part.json)

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -141,7 +141,9 @@ void LottieImage::prepare()
 void LottieTrimpath::segment(float frameNo, float& start, float& end, LottieExpressions* exps)
 {
     start = this->start(frameNo, exps) * 0.01f;
+    tvg::clamp(start, 0.0f, 1.0f);
     end = this->end(frameNo, exps) * 0.01f;
+    tvg::clamp(end, 0.0f, 1.0f);
 
     auto o = fmodf(this->offset(frameNo, exps), 360.0f) / 360.0f;  //0 ~ 1
 
--- a/src/common/tvgMath.cpp
+++ b/src/common/tvgMath.cpp
@@ -367,8 +367,7 @@ float Bezier::angle(float t) const
 uint8_t lerp(const uint8_t &start, const uint8_t &end, float t)
 {
     auto result = static_cast<int>(start + (end - start) * t);
-    if (result > 255) result = 255;
-    else if (result < 0) result = 0;
+    tvg::clamp(result, 0, 255);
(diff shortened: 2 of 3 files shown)


src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -141,7 +141,9 @@ void LottieImage::prepare()
 void LottieTrimpath::segment(float frameNo, float& start, float& end, LottieExpressions* exps)
 {
     start = this->start(frameNo, exps) * 0.01f;
+    start = start < 0.0f || start > 1.0f ? 0.0f : start;
```
hermet, line 144: Why the exceed values are exported from the get go....

please do not assign data unnecessarily.
```if (start < 0.0f || start > 1.0f ? 0.0f) start = 0.0f;```

File: src/loaders/lottie/tvgLottieBuilder.cpp [hunk 2]
```diff
@@ -865,6 +865,7 @@ void LottieBuilder::updateTrimpath(TVG_UNUSED LottieGroup* parent, LottieObject*
     if (P(ctx->propagator)->rs.stroke) {
         auto pbegin = P(ctx->propagator)->rs.stroke->trim.begin;
         auto pend = P(ctx->propagator)->rs.stroke->trim.end;
+        if (pbegin > pend) std::swap(pbegin, pend);
```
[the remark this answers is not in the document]
mgrudzinska: done in https://github.com/thorvg/thorvg/pull/2694

mgrudzinska:
I think the underlaying cause is something else, need to investigate

mgrudzinska:
I could not find any example that breaks the logic from this pr. it works

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -141,7 +141,11 @@ void LottieImage::prepare()
 void LottieTrimpath::segment(float frameNo, float& start, float& end, LottieExpressions* exps)
 {
     start = this->start(frameNo, exps) * 0.01f;
+    if (start > 1.0f) start = 1.0f;
```
hermet, line 144: Suggestion:
```
if () ...
else if() ...
```
Otherwise, probably we can introduce clamp() method in math general functions.
```
template <typename T>
static inline void clamp(T& v, T& min, T& max)
{
    if (v < min) v = min;
    else if (v > max) v = max;
}
```

hermet (approved):

PR #2695, opened 2024-08-28
wg_engine: external webgpu handles (web integration)

SergeyLebedkin:
move instance, adapter and device creation from renderer to application 
its necessary for web integration, because browser have its own mechanics to create hardware handles 
this changes makes webgpu canvas more universal to use in case of system and web applications

See https://github.com/thorvg/thorvg/issues/2410


src/renderer/tvgWgCanvas.cpp
```diff
@@ -50,11 +50,13 @@ WgCanvas::WgCanvas() : Canvas(nullptr), pImpl(nullptr)
 
 WgCanvas::~WgCanvas()
 {
-    delete pImpl;
+    #ifdef THORVG_WG_RASTER_SUPPORT
+    static_cast<WgRenderer*>(Canvas::pImpl->renderer)->surface = nullptr;
```
hermet, line 54: @SergeyLebedkin  Does this really necessary job?  it shouldn't be like this.

SergeyLebedkin: @hermet Yes. We need some how tell to renderer that surface already destroed. If not, it will raise an exception because of Canvas::Impl design: it calls renderer->sync() in disctuctor, but surface is invalid in this stage

```
struct Canvas::Impl
{
    ~Impl()
    {
        //make it sure any deferred jobs
        renderer->sync();

        clearPaints();

        if (renderer->unref() == 0) delete(renderer);
    }

```

examples/Example.h
```diff
@@ -421,6 +423,24 @@ struct WgWindow : Window
...
+
+        // request device
+        const WGPUDeviceDescriptor deviceDesc { .nextInChain = nullptr, .label = "The device", .requiredFeatureCount = featuresCount, .requiredFeatures = featureNames };
+        auto onDeviceRequestEnded = [](WGPURequestDeviceStatus status, WGPUDevice device, char const * message, void * pUserData) { *((WGPUDevice*)pUserData) = device; };
+        wgpuAdapterRequestDevice(adapter, &deviceDesc, onDeviceRequestEnded, &device);
+
+        // device uncaptured error callback
+        auto onDeviceError = [](WGPUErrorType type, char const* message, void* pUserData) { std::cout << message << std::endl; };
+        wgpuDeviceSetUncapturedErrorCallback(device, onDeviceError, nullptr);
+
```
tinyjin, line 443: @SergeyLebedkin 
What do you think about we make static helper function to initialize adapter and device?

To remove duplicated code lines and reduce binary size.

I think I'll just copy-paste that codes into WASM binding.

SergeyLebedkin: Yes, we can do this, if it will be hepfull for you

hermet (changes requested):
@SergeyLebedkin user should not destroy a surface, assigned in wg canvas. or they should use target() to unassign before destroying surface.

src/renderer/tvgWgCanvas.cpp
hermet: @SergeyLebedkin user should not destroy a surface, assigned in wg canvas. or they should use target() to unassign before destroying surface.

SergeyLebedkin: @hermet I think that we can use terget for un-usigning. Will fix now

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -271,14 +272,25 @@ bool WgRenderer::sync()
...
+
+bool WgRenderer::target(WGPUSurface surface, uint32_t w, uint32_t h) {
+        // store target surface properties
+    this->surface = surface;
     mTargetSurface.stride = w;
     mTargetSurface.w = w > 0 ? w : 1;
     mTargetSurface.h = h > 0 ? h : 1;
     
-    mContext.initialize(instance, surface);
+    if (!surface) return false;
```
hermet, line 293: If unassignment is also an expected user scenario, it should return true.

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -271,14 +272,25 @@ bool WgRenderer::sync()
...
+    initialize();
+    target(surface, w, h);
+    mRenderStoragePool.initialize(mContext, w, h);
+    mStorageRoot.initialize(mContext, w, h);
+    mCompositor.initialize(mContext, w, h);
+    return true;
+}
+
+bool WgRenderer::target(WGPUSurface surface, uint32_t w, uint32_t h) {
+        // store target surface properties
```
hermet, line 287: fix the indentation

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -245,10 +245,11 @@ bool WgRenderer::clear()
 
 bool WgRenderer::sync()
 {
+    if (!surface) return false;
     disposeObjects();
```
hermet, line 249: if the canvas is destroyed in the middle of the usage, no resource leaks?

inc/thorvg.h
```diff
@@ -1754,7 +1754,7 @@ class TVG_API WgCanvas final : public Canvas
      * @see Canvas::viewport()
      * @see Canvas::sync()
      */
-    Result target(void* instance, void* surface, uint32_t w, uint32_t h) noexcept;
+    Result target(void* instance, void* adapter, void* device, void* surface, uint32_t w, uint32_t h) noexcept;
```
hermet, line 1757: please update the api doc.

examples/Example.h
hermet: @SergeyLebedkin Please merge `createHandles()`; into `WgWindow()`.

hermet (changes requested):

src/renderer/wg_engine/tvgWgRenderer.cpp
SergeyLebedkin: Objects will be destoyed when renderer will be released.
But we can remove them in any case

SergeyLebedkin:
@hermet 
> Please merge createHandles(); into WgWindow().
Its already in WgWindow class, what do you mean "merge"

hermet:
> @hermet
> 
> > Please merge createHandles(); into WgWindow().
> > Its already in WgWindow class, what do you mean "merge"

@SergeyLebedkin Merge the code from createHandle() into WgWindow(). Rather than adding function call.

src/renderer/tvgWgCanvas.cpp
```diff
@@ -50,11 +50,14 @@ WgCanvas::WgCanvas() : Canvas(nullptr), pImpl(nullptr)
 
 WgCanvas::~WgCanvas()
 {
-    delete pImpl;
+    #ifdef THORVG_WG_RASTER_SUPPORT
```
hermet, line 53: please remove the indentation for `#ifdef ~ #endif `

examples/Example.h
```diff
@@ -435,20 +455,23 @@ struct WgWindow : Window
...
         //Set the canvas target and draw on it.
-        verify(canvas->target(instance, surface, width, height));
+        verify(canvas->target(instance, adapter, device, surface, width, height));
     }
 
     void refresh() override 
     {
         wgpuSurfacePresent(surface);
     }
+
```
hermet, line 474: -- unnecessary change

src/renderer/wg_engine/tvgWgRenderer.h
```diff
@@ -55,13 +56,15 @@ class WgRenderer : public RenderMethod
...
+    WGPUSurface surface{}; // external handle
+
 private:
     WgRenderer();
     ~WgRenderer();
     void initialize();
     void release();
     void disposeObjects();
-
+
```
hermet, line 67: -- white spaces

inc/thorvg.h
```diff
@@ -1754,7 +1756,7 @@ class TVG_API WgCanvas final : public Canvas
...
```
hermet, line 1759: Please double-check the `adapter` parameter here. I don't see any necessity, otherwise please remove it.
```
Result target(void* instance, void* surface, uint32_t w, uint32_t h, void* device = nullptr)
```

inc/thorvg.h
```diff
@@ -1742,6 +1742,8 @@ class TVG_API WgCanvas final : public Canvas
      * @brief Sets the drawing target for the rasterization.
      *
      * @param[in] instance WGPUInstance, context for all other wgpu objects.
+     * @param[in] adapter WGPUAdapter, handle for webgpu adapter
+     * @param[in] device WGPUDevice, handle for webgpu device
```
hermet, line 1746: ```@param[in] device WGPUDevice: desired handle for the WebGPU device. If it is nullptr, ThorVG will assign an appropriate device internally.```

hermet (changes requested):
@SergeyLebedkin Let's roll back this part. I think we can handle it in the WG renderer for user convenience, inheriting only when the device value is passed.
```
        // request adapter
        const WGPURequestAdapterOptions requestAdapterOptions { .nextInChain = nullptr, .compatibleSurface = surface, .powerPreference = WGPUPowerPreference_HighPerformance, .forceFallbackAdapter = false };
        auto onAdapterRequestEnded = [](WGPURequestAdapterStatus status, WGPUAdapter adapter, char const * message, void * pUserData) { *((WGPUAdapter*)pUserData) = adapter; };
        wgpuInstanceRequestAdapter(instance, &requestAdapterOptions, onAdapterRequestEnded, &adapter);

        // get adapter and surface properties
        WGPUFeatureName featureNames[32]{};
        size_t featuresCount = wgpuAdapterEnumerateFeatures(adapter, featureNames);

        // request device
        const WGPUDeviceDescriptor deviceDesc { .nextInChain = nullptr, .label = "The device", .requiredFeatureCount = featuresCount, .requiredFeatures = featureNames };
        auto onDeviceRequestEnded = [](WGPURequestDeviceStatus status, WGPUDevice device, char const * message, void * pUserData) { *((WGPUDevice*)pUserData) = device; };
        wgpuAdapterRequestDevice(adapter, &deviceDesc, onDeviceRequestEnded, &device);

        // device uncaptured error callback
        auto onDeviceError = [](WGPUErrorType type, char const* message, void* pUserData) { std::cout << message << std::endl; };
        wgpuDeviceSetUncapturedErrorCallback(device, onDeviceError, nullptr);
```

src/renderer/wg_engine/tvgWgRenderer.h
```diff
@@ -74,7 +77,7 @@ class WgRenderer : public RenderMethod
     WgContext mContext;
     WgPipelines mPipelines;
     WgCompositor mCompositor;
-
```
hermet, line 77: please remove white spaces.

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -271,14 +272,43 @@ bool WgRenderer::sync()
 
 
 // target for native window handle
-bool WgRenderer::target(WGPUInstance instance, WGPUSurface surface, uint32_t w, uint32_t h)
+bool WgRenderer::target(WGPUInstance instance, WGPUSurface surface, uint32_t w, uint32_t h, WGPUDevice device)
 {
+    WGPUDevice dev = device;
```
hermet, line 277: please make it code simple. i think you can directly use `device`.

src/renderer/tvgWgCanvas.cpp
```diff
@@ -50,11 +50,14 @@ WgCanvas::WgCanvas() : Canvas(nullptr), pImpl(nullptr)
 
 WgCanvas::~WgCanvas()
 {
-    delete pImpl;
+#ifdef THORVG_WG_RASTER_SUPPORT
+    auto renderer = static_cast<WgRenderer*>(Canvas::pImpl->renderer);
+    renderer->target(nullptr, 1, 1);
```
hermet, line 55: please pass 0, 0

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -271,14 +272,43 @@ bool WgRenderer::sync()
...
+    mStorageRoot.initialize(mContext, w, h);
+    mCompositor.initialize(mContext, w, h);
+    return true;
+}
+
+bool WgRenderer::target(WGPUSurface surface, uint32_t w, uint32_t h) {
     // store target surface properties
+    this->surface = surface;
     mTargetSurface.stride = w;
     mTargetSurface.w = w > 0 ? w : 1;
```
hermet, line 308: I think you should not change w/h value here.

examples/Example.h
```diff
@@ -421,6 +423,24 @@ struct WgWindow : Window
         surfaceDesc.label = "The surface";
         surface = wgpuInstanceCreateSurface(instance, &surfaceDesc);
 
+        // request adapter
```
hermet, line 426: no need to change Example code anymore.

examples/Example.h
hermet: @SergeyLebedkin please revert the example code all here and just pass null for device.

hermet (changes requested):

hermet (approved):

tinyjin (approved):
Confirmed that it works in the browser

PR #2696, opened 2024-08-28
loader/ttf: Spaces and invalid glyphs rendering fix

### LINKED ISSUE #2687: loader/text: Invalid space width
A space's width is incorrect when it appears after a glyph with kerning applied:

```
auto text = tvg::Text::gen();
text->font("Arial", 80);
text->text("  Te s   t \xe4\xb8\x8d !");
text->fill(255, 255, 255);
```

Output:

Expected output:

lpogic:
Whitespace is processed as valid glyph without an outline.
If for any reason an invalid glyph is found in text, it will now be skipped when rendering (instead of increasing offset by last valid glyph advance).

Closes #2687

### DIFF
--- a/src/loaders/ttf/tvgTtfReader.cpp
+++ b/src/loaders/ttf/tvgTtfReader.cpp
@@ -387,12 +387,11 @@ uint32_t TtfReader::glyph(uint32_t codepoint)
 uint32_t TtfReader::glyph(uint32_t codepoint, TtfGlyphMetrics& gmetrics)
 {
     auto glyph = this->glyph(codepoint);
-    if (glyph == INVALID_GLYPH) {
+    if (glyph == INVALID_GLYPH || !glyphMetrics(glyph, gmetrics)) {
         TVGERR("TTF", "invalid glyph id, codepoint(0x%x)", codepoint);
         return INVALID_GLYPH;
     }
-    
-    return glyphMetrics(glyph, gmetrics) ? glyph : INVALID_GLYPH;
+    return glyph;
 }
 
 bool TtfReader::glyphMetrics(uint32_t glyphIndex, TtfGlyphMetrics& gmetrics)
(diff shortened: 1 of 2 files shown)


src/loaders/ttf/tvgTtfReader.cpp
```diff
@@ -389,10 +389,11 @@ uint32_t TtfReader::glyph(uint32_t codepoint, TtfGlyphMetrics& gmetrics)
     auto glyph = this->glyph(codepoint);
     if (glyph == INVALID_GLYPH) {
         TVGERR("TTF", "invalid glyph id, codepoint(0x%x)", codepoint);
-        return INVALID_GLYPH;
+    } else if(!glyphMetrics(glyph, gmetrics)) {
```
hermet, line 392: Considering your first commit, it looks you don't need to change the code in glyph().
Could you please squash commits into one to make it clearly visible?

hermet (commented):
@lpogic Thanks. Could you please squash the commits and include the PR messages in the commit message so that we can track the history through the pure Git system?

lpogic:
Dear @hermet. I tried to implement the recommendations in the review, but I'm afraid I only made things worse. If so, I'll make a new PR and work on my git level.

lpogic:
I think it should be better now

src/loaders/ttf/tvgTtfReader.cpp
```diff
@@ -389,10 +389,11 @@ uint32_t TtfReader::glyph(uint32_t codepoint, TtfGlyphMetrics& gmetrics)
     auto glyph = this->glyph(codepoint);
     if (glyph == INVALID_GLYPH) {
         TVGERR("TTF", "invalid glyph id, codepoint(0x%x)", codepoint);
-        return INVALID_GLYPH;
```
hermet, line 392: @lpogic It doesn't seem necessary to change the `glyph()` function. I'd like to ask you to retain the original code unless this is needed for your fix. Could you please double-check this?

lpogic: This change is not essential and was just to make debugging easier. I will revert it soon.

hermet: Coud you please apply this code?
```
if (glyph == INVALID_GLYPH || !glyphMetrics(glyph, gmetrics)) {
        TVGERR("TTF", "invalid glyph id, codepoint(0x%x)", codepoint);
        return INVALID_GLYPH;
}
return glyph;
```

hermet (approved):

PR #2700, opened 2024-08-29
examples: clear canvas while updating

mgrudzinska:
An issue was observed while drawing without background.


hermet:
@mgrudzinska It doesn't need to clear buffer when it has a *intentional* bg.

hermet (changes requested):
It was intentional. Please don't change it; it's the best showcase for optimal usage.

PR #2702, opened 2024-09-01
sw_engine: handle small cubics

mgrudzinska:
During the stroke's outline calculation, the function handling small cubics set all angles to zero. Such cases should be ignored, as further processing caused errors - when the cubic was small but not zero, setting the angles to zero resulted in incorrect outlines.

@Issue: https://github.com/godotengine/godot/issues/96262

before:
Zrzut ekranu 2024-09-1 o 22 48 11

after:
Zrzut ekranu 2024-09-1 o 22 48 24

note:
To reproduce this issue no scaling can be applied to the svg file.


hermet (changes requested):
@mgrudzinska Please double-check, after your patch; it occasionally drops the Lottie performance.

image

mgrudzinska:
Since there is a problem with performance in this pr, the reported issue is solved in https://github.com/thorvg/thorvg/pull/2711

Although the underlying case should be investigated: https://github.com/thorvg/thorvg/issues/2712

mgrudzinska:
https://github.com/thorvg/thorvg/pull/2798

PR #2707, opened 2024-09-03
common: precision increase while matrix inversion

mgrudzinska:
A numerically motivated limit on the matrix determinant set at 1e-6 was not sufficient. The limit has been increased.

before:
Zrzut ekranu 2024-09-3 o 17 35 17

after:
Zrzut ekranu 2024-09-3 o 17 34 43

sample:
[icons.svg.zip](https://github.com/user-attachments/files/16851194/icons.svg.zip)

### DIFF
--- a/src/common/tvgMath.cpp
+++ b/src/common/tvgMath.cpp
@@ -120,9 +120,8 @@ bool inverse(const Matrix* m, Matrix* out)
                m->e12 * (m->e21 * m->e33 - m->e23 * m->e31) +
                m->e13 * (m->e21 * m->e32 - m->e22 * m->e31);
 
-    if (tvg::zero(det)) return false;
-
-    auto invDet = 1 / det;
+    auto invDet = 1.0f / det;
+    if (!std::isfinite(invDet)) return false;
 
     out->e11 = (m->e22 * m->e33 - m->e32 * m->e23) * invDet;
     out->e12 = (m->e13 * m->e32 - m->e12 * m->e33) * invDet;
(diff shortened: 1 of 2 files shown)


src/common/tvgMath.cpp
```diff
@@ -120,9 +120,15 @@ bool inverse(const Matrix* m, Matrix* out)
                m->e12 * (m->e21 * m->e33 - m->e23 * m->e31) +
                m->e13 * (m->e21 * m->e32 - m->e22 * m->e31);
 
-    if (tvg::zero(det)) return false;
+    auto scaled = false;
 
-    auto invDet = 1 / det;
```
hermet, line 125: how about just checking this?
```if (std::isnan(invDet)) return false;```

Comment (mgrudzinska) [reply src/common/tvgMath.cpp:125]: in order to handle nans and infs I used isfinite.

File: src/common/tvgMath.cpp [hunk 2]
```diff
@@ -21,6 +21,7 @@
  */
 
 #include "tvgMath.h"
+#include <cmath>
```
hermet, line 24: `tvgMath.h` also includes math. probably we can include one of them in `tvgMath.h`

src/common/tvgMath.cpp
hermet: hmm... inf is necessary to check here?

src/common/tvgMath.cpp
mgrudzinska: changed, on mac worked either way, waiting for android

hermet (approved):

Issue #2706, opened 2024-09-03
Regression in v0.13.8 for SVG with text node and no embedded font (take two)


akien-mga:
We found more SVG files which stopped working in ThorVG v0.13.8 because of the added text node support in 7e43cf402a4c5f585e96164661310a83f8d6eac4, and this problem is reproducible in v0.14.8.

Previous issue was #2468 and was fixed, but apparently this didn't solve the regression fully.

Godot bug report:
- https://github.com/godotengine/godot/issues/96491

Here's a set of SVG files which do not render anything with ThorVG 0.14.8 (since 7e43cf402 in 0.13.8), both in Godot and in the ThorVG Viewer (provided by @antimundo, from their MIT licensed open source game [poder-solar](https://github.com/antimundo/poder-solar)):

[thorvg-text-node-breaks-images.zip](https://github.com/user-attachments/files/16849750/thorvg-text-node-breaks-images.zip)

Here's one example from that set:

![power_plant](https://github.com/user-attachments/assets/3cc0d5d7-c17c-46a5-96a2-e57c09459414)

ThorVG v0.14.8 doesn't render it, while ThorVG v0.13.7 rendered it fine.

It does contain a text node, though I don't think it intends to display any text. This might have been added automatically by a plugin or copied over by mistake in a template SVG.

```xml
<text
       xml:space="preserve"
       style="font-size:37.3333px;font-family:Georgia;-inkscape-font-specification:Georgia;fill:#856944;fill-opacity:1;stroke:none;stroke-width:3;stroke-linejoin:round;stroke-dasharray:none;stroke-opacity:1;paint-order:stroke fill markers"
       x="982.45825"
       y="503.88559"
       id="text5"><tspan
         sodipodi:role="line"
         id="tspan5"
         x="982.45825"
         y="503.88559"
         style="font-style:normal;font-variant:normal;font-weight:normal;font-stretch:normal;font-family:Impact;-inkscape-font-specification:Impact" /></text><text
       xml:space="preserve"
       style="font-size:18.6667px;font-family:'Franklin Gothic Medium';-inkscape-font-specification:'Franklin Gothic Medium, ';text-align:center;letter-spacing:0px;text-anchor:middle;opacity:1;fill:#399351;fill-opacity:1;stroke:none;stroke-width:4;stroke-linecap:round;stroke-linejoin:miter;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;paint-order:stroke fill markers"
[...]

PR #2708, opened 2024-09-03
svg_loader: fix text nodes issue

mgrudzinska:
Text can be added in parts due to the presence
of the <tspan> tag. This requires that each
subsequent piece of text is appended rather than
overwriting the previous one.

before:
Zrzut ekranu 2024-09-10 o 11 24 40

after:
Zrzut ekranu 2024-09-10 o 11 22 17

sample:
[textAppend.svg.zip](https://github.com/user-attachments/files/16942598/textAppend.svg.zip)

already merged:
Since the text node doesn't get pushed onto the loader's stack, closing the text node incorrectly popped an element from the stack (if one was present). This could result in elements that were supposed to have that element as a parent being rendered incorrectly or not being rendered at all.

@Issue: https://github.com/thorvg/thorvg/issues/2706

before:
Zrzut ekranu 2024-09-4 o 00 05 08

after:
Zrzut ekranu 2024-09-4 o 00 05 33

samples from the issue: https://github.com/thorvg/thorvg/issues/2706

### DIFF
--- a/src/common/tvgStr.cpp
+++ b/src/common/tvgStr.cpp
@@ -219,6 +219,21 @@ char* strDuplicate(const char *str, size_t n)
     return (char *) memcpy(ret, str, n);
 }
 
+char* strAppend(char* lhs, const char* rhs, size_t n)
+{
+    if (!rhs) return lhs;
+    if (!lhs) return strDuplicate(rhs, n);
+
+    if (auto ret = (char*)malloc(strlen(lhs) + n + 1)) {
+        ret[0] = '\0';
+        strcat(ret, lhs);
+        free(lhs);
+        return strncat(ret, rhs, n);
+    }
+
+    return nullptr;
+}
+
 char* strDirname(const char* path)
 {
     const char *ptr = strrchr(path, '/');
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -2183,6 +2183,7 @@ static SvgNode* _createTextNode(SvgLoaderData* loader, SvgNode* parent, const ch
     //TODO: support the def font and size as used in a system?
     loader->svgParse->node->node.text.fontSize = 10.0f;
     loader->svgParse->node->node.text.fontFamily = nullptr;
+    loader->svgParse->node->node.text.text = nullptr;
 
     func(buf, bufLength, _attrParseTextNode, loader);
 
@@ -3400,8 +3401,7 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content, 
 static void _svgLoaderParserText(SvgLoaderData* loader, const char* content, unsigned int length)
 {
     auto text = &loader->svgParse->node->node.text;
-    if (text->text) free(text->text);
-    text->text = strDuplicate(content, length);
+    text->text = strAppend(text->text, content, length);
 }
 
 
--- a/src/common/tvgStr.h
+++ b/src/common/tvgStr.h
@@ -30,6 +30,7 @@ namespace tvg
 
 float strToFloat(const char *nPtr, char **endPtr);  //convert to float
 char* strDuplicate(const char *str, size_t n);      //copy the string
+char* strAppend(char* lhs, const char* rhs, size_t n);  //append the source to the target
 char* strDirname(const char* path);                 //return the full directory name
 
 }


capnm:
image ;-)

SVG [icon_looks_left_text.zip](https://github.com/user-attachments/files/16859274/icon_looks_left_text.zip)

:+1:

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3288,7 +3288,8 @@ static void _svgLoaderParserXmlClose(SvgLoaderData* loader, const char* content,
     for (unsigned int i = 0; i < sizeof(graphicsTags) / sizeof(graphicsTags[0]); i++) {
         if (!strncmp(tagName, graphicsTags[i].tag, sz)) {
             loader->currentGraphicsNode = nullptr;
-            loader->stack.pop();
+            if (!strncmp(tagName, "text", 4)) loader->openedTag = OpenedTagType::Other;
```
JSUYA, line 3291: There may be cases like this.
`<text> text <some tags.../>  txt </text> ` https://www.w3.org/TR/SVG2/text.html#TextElement
Since text also acts like a group, we can push it onto the stack.

```diff
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -3360,7 +3360,10 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content,
         else parent = loader->doc;
         node = method(loader, parent, attrs, attrsLength, simpleXmlParseAttributes);
         if (node && !empty) {
-            if (!strcmp(tagName, "text")) loader->openedTag = OpenedTagType::Text;
+            if (!strcmp(tagName, "text")) {
+                loader->openedTag = OpenedTagType::Text;
+                loader->stack.push(node);
+            }
             else {
                 auto defs = _createDefsNode(loader, nullptr, nullptr, 0, nullptr);
                 loader->stack.push(defs);

```

When I tested with the icon_looks_left_text.svg sample, the icon is displayed fine, but the text is not displayed, so I don't know what other problem is left.

capnm: IIRC, embedding the required fonts isn't yet implemented. You currently need to load an external font and give it the same name (family), e.g.:

```cpp
     // font
         // ---- Load from memory. ----
         {
            ifstream file(font_path, ios::binary);
            if (!file.is_open())
            {
               return false;
            }
            file.seekg(0, std::ios::end);
            std::streampos size = file.tellg();
            file.seekg(0, std::ios::beg);

            char *data = (char *)malloc(size);
            if (data && file.read(data, size))
            {
               tvg::Result result = tvg::Text::load(font_name.c_str(), data, size, "ttf", true);
...
```

JSUYA: > IIRC, embedding the required fonts isn't yet implemented. You currently need to load an external font and give it the same name (family), e.g.:

Thanks :) I loaded the fontfamily and tested it and I checked the text is displayed.

capnm:
```xml
      <text x="40" y="30" font-family="serif" fill="black" font-size="20" >
         LOOK <tspan fill="red" >?</tspan> LEFT
      </text>

// [L] SVG (../thorvg-git/src/loaders/svg/tvgSvgLoader.cpp 3395): Unsupported elements used [Elements: tspan]
```

Ideally `text` would skip the unsupported tags, now it outputs "LEFT". But this is, 
I think, an irrelevant issue to this one.

mgrudzinska:
> Ideally `text` would skip the unsupported tags, now it outputs "LEFT". But this is, I think, an irrelevant issue to this one.

thx for this comment :) applied.

no way to separate these two commits, since the second one doesn't work without the previous one.

mgrudzinska:
> the existing character spacing shouldn't be removed?

let's consider this as a separate issue (https://github.com/thorvg/thorvg/issues/2710). I need to check which characters should be ignored and which should not. For example, from what I've tested, spaces remain, but only single ones, etc

JSUYA (approved):

capnm:
I also tried "Unsupported elements used [elements: **textPath**]" and some other tags,
LGTM

> > the existing character spacing shouldn't be removed?
> 
> let's consider this as a separate issue (#2710). I need to check which characters should be ignored and which should not. For example, from what I've tested, spaces remain, but only single ones, etc

src/common/tvgStr.cpp
```diff
@@ -229,6 +229,26 @@ char* strDuplicate(const char *str, size_t n)
     return (char *) memcpy(ret, str, n);
 }
 
+char* strAppend(char* target, const char* source, size_t n)
+{
```
hermet, line 233: please check this code:
```
if ((ret = malloc(strlen(lhs)+n+1)){
    ret[0] = '\0';
    strcat(ret, lhs);
    strncat(ret, rhs, n);
}
```

hermet:
@mgrudzinska Hello, I applied only this [patch ](https://github.com/thorvg/thorvg/pull/2708/commits/acadfa420643a9fde7e0333916c9b70a119e44b7)for hotfix. Please update the second patch.

src/common/tvgStr.cpp
mgrudzinska: please check now

src/common/tvgStr.cpp
```diff
@@ -219,6 +219,21 @@ char* strDuplicate(const char *str, size_t n)
     return (char *) memcpy(ret, str, n);
 }
 
+char* strAppend(char* lhs, const char* rhs, size_t n)
+{
+    if (!rhs) return lhs;
```
hermet, line 224: `n` should be guaranteed, so let's remove 224; the exception is unnecessary.

src/common/tvgStr.cpp
```diff
@@ -219,6 +219,21 @@ char* strDuplicate(const char *str, size_t n)
...
+    if (!lhs) return strDuplicate(rhs, n);
+
+    if (auto ret = (char*)malloc(strlen(lhs) + n + 1)) {
```
hermet, line 227: these 3 lines are all needed for us?
```
if (!lhs) return strDuplicate(rhs, n);
lhs = (char*)realloc(strlen(lhs) + n + 1);
return strncat(lhs, rhs, n);
```

hermet:
> Any idea what could be wrong with the disabled font-support?
> (I just did a quick check in the Godot master repo):

@capnm I'm a bit confused about the context :) Does the current PR fix that issue?

capnm:
> > Any idea what could be wrong with the disabled font-support?
> > (I just did a quick check in the Godot master repo):
> 
> @capnm I'm a bit confused about the context :) Does the current PR fix that issue? or an another issue ThorVG needs to deal with it?

I added now a comment in the original bug report. Ideally, we should try to imitate Godot's SVG access in some ThorVG example. By running the debugger inside Godot, many TVG debug symbols are optimized out... :(
https://github.com/godotengine/godot/issues/96491#issuecomment-2333562665

capnm:
@hermet I can now test the bug in the TVG Viewer and have created a local issue. https://github.com/thorvg/thorvg/issues/2715

hermet:
@mgrudzinska, could you please pick any resource that addresses the bug fixed by this commit?

mgrudzinska:
> @mgrudzinska, could you please pick any resource that addresses the bug fixed by this commit?

added in the pr description.

mgrudzinska:
rebased

src/common/tvgStr.cpp
mgrudzinska: done

hermet:
https://github.com/thorvg/thorvg/commit/bf9f5ab1c5a7a56dcf0d46943dfe024c39aaaacd

PR #2725, opened 2024-09-12
wasm: Handle WebGPU part for integration

tinyjin:
Updated the WebGPU engine API to align with the latest interface.

Static/Single instance, adapter and device to support multiple canvases on the browser.

Moved the WGPU request part to `initWGInstance` helper function. This must only be called once, even when multiple players are involved.

Issue: #2695

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -21,6 +21,7 @@
  */
 
 #include <thorvg.h>
+#include <emscripten.h>
 #include <emscripten/bind.h>
 #include "tvgPicture.h"
 #ifdef THORVG_WG_RASTER_SUPPORT
@@ -32,6 +33,47 @@ using namespace std;
 using namespace tvg;
 
 static const char* NoError = "None";
+
+#ifdef THORVG_WG_RASTER_SUPPORT
+    static WGPUInstance instance{};
+    static WGPUAdapter adapter{};
+    static WGPUDevice device{};
+#endif
+
+void init()
+{
+#ifdef THORVG_WG_RASTER_SUPPORT
+    //Init WebGPU
+    if (!instance) instance = wgpuCreateInstance(nullptr);
+
+    // request adapter
+    if (!adapter) {
+        const WGPURequestAdapterOptions requestAdapterOptions { .nextInChain = nullptr, .powerPreference = WGPUPowerPreference_HighPerformance, .forceFallbackAdapter = false };
+        auto onAdapterRequestEnded = [](WGPURequestAdapterStatus status, WGPUAdapter adapter, char const * message, void * pUserData) { *((WGPUAdapter*)pUserData) = adapter; };
+        wgpuInstanceRequestAdapter(instance, &requestAdapterOptions, onAdapterRequestEnded, &adapter);
+        while (!adapter) emscripten_sleep(10);
+    }
+
+    // request device
+    WGPUFeatureName featureNames[32]{};
+    size_t featuresCount = wgpuAdapterEnumerateFeatures(adapter, featureNames);
+    if (!device) {
+        const WGPUDeviceDescriptor deviceDesc { .nextInChain = nullptr, .label = "The device", .requiredFeatureCount = featuresCount, .requiredFeatures = featureNames };
+        auto onDeviceRequestEnded = [](WGPURequestDeviceStatus status, WGPUDevice device, char const * message, void * pUserData) { *((WGPUDevice*)pUserData) = device; };
+        wgpuAdapterRequestDevice(adapter, &deviceDesc, onDeviceRequestEnded, &device);
+        while (!device) emscripten_sleep(10);
+    }
+#endif
+}
+
+void term()
+{
+#ifdef THORVG_WG_RASTER_SUPPORT
+    wgpuDeviceRelease(device);
+    wgpuAdapterRelease(adapter);
+    wgpuInstanceRelease(instance);
+#endif
+}
 
 struct TvgEngineMethod
 {
@@ -77,7 +119,6 @@ struct TvgSwEngine : TvgEngineMethod
 struct TvgWgEngine : TvgEngineMethod
 {
     #ifdef THORVG_WG_RASTER_SUPPORT
-        WGPUInstance instance{};
         WGPUSurface surface{};
     #endif
 
@@ -85,7 +126,6 @@ struct TvgWgEngine : TvgEngineMethod
     {
         #ifdef THORVG_WG_RASTER_SUPPORT
             wgpuSurfaceRelease(surface);
-            wgpuInstanceRelease(instance);
         #endif
         Initializer::term(tvg::CanvasEngine::Wg);
     }
@@ -93,9 +133,6 @@ struct TvgWgEngine : TvgEngineMethod
     unique_ptr<Canvas> init(string& selector) override
     {
         #ifdef THORVG_WG_RASTER_SUPPORT
-            //Init WebGPU
-            instance = wgpuCreateInstance(nullptr);
-
             WGPUSurfaceDescriptorFromCanvasHTMLSelector canvasDesc{};
             canvasDesc.chain.next = nullptr;
             canvasDesc.chain.sType = WGPUSType_SurfaceDescriptorFromCanvasHTMLSelector;
@@ -112,7 +149,7 @@ struct TvgWgEngine : TvgEngineMethod
 
     void resize(Canvas* canvas, int w, int h) override
     {
-        static_cast<WgCanvas*>(canvas)->target(instance, surface, w, h);
+        static_cast<WgCanvas*>(canvas)->target(instance, surface, w, h, device);
     }
 };
 
@@ -425,6 +462,9 @@ private:
 
 EMSCRIPTEN_BINDINGS(thorvg_bindings)
 {
+    emscripten::function("init", &init);
+    emscripten::function("term", &term);
+
     class_<TvgLottieAnimation>("TvgLottieAnimation")
         .constructor<string, string>()
         .function("error", &TvgLottieAnimation ::error, allow_raw_pointers())


src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -33,6 +34,38 @@ using namespace tvg;
...
+    static WGPUDevice device{};
+
+    void initWgInstance()
+    {
+        //Init WebGPU
+        if (!instance) instance = wgpuCreateInstance(nullptr);
+
+        // request adapter
+        if (!adapter) {
+            const WGPURequestAdapterOptions requestAdapterOptions { .nextInChain = nullptr, .powerPreference = WGPUPowerPreference_HighPerformance, .forceFallbackAdapter = false };
```
tinyjin, line 49: @SergeyLebedkin I found the `.compatibleSurface` parameter is not mandatory for browser support. Requiring surface on request adapter part makes browser initialization difficult.

Removed the param, please share your thought if you have any.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -33,6 +34,38 @@ using namespace tvg;
 
 static const char* NoError = "None";
 
+#ifdef THORVG_WG_RASTER_SUPPORT
+    static WGPUInstance instance{};
+    static WGPUAdapter adapter{};
+    static WGPUDevice device{};
```
tinyjin, line 40: @SergeyLebedkin 
Browser WebGPU will have bit different approach other than Native WebGPU.

WASM will ensure WGPUInstance, WGPUAdapter and WGPUDevice in singleton member. They will only be initialized once before used in player(s).

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -425,6 +456,10 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
 
 EMSCRIPTEN_BINDINGS(thorvg_bindings)
 {
+    #ifdef THORVG_WG_RASTER_SUPPORT
+        emscripten::function("initWgInstance", &initWgInstance);
+    #endif
```
tinyjin, line 461: @hermet I know avoiding helper function like this would be better. But without this interface, it is hard to make sure WGPU initialization called once in asynchronous and multiple players creation.

Furthermore if constructor has WGPU request part, emscripten forcibly makes constructor as a asynchronous function.

Then We have to call like this:
```ts
this._TVG = await (new _module.TvgLottieAnimation(engine, `#${this._canvas!.id}`));
```

Using `await` keyword on constructor is sort of unusual JavaScript constructor, it will make users confused.

Thus I introduced new helper function, this interface can provide APIs with better usage of WGPU initialization.

tinyjin (commented):
Pointed major changes on WASM for WebGPU. Feel free to review them plz.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -425,6 +456,10 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
 
 EMSCRIPTEN_BINDINGS(thorvg_bindings)
 {
+    #ifdef THORVG_WG_RASTER_SUPPORT
+        emscripten::function("initWgInstance", &initWgInstance);
```
hermet, line 460: This design looks weird to me. May I ask why this is needed? The `wgpu` initialization seems to happen in one place at a specific time. If you need to reconsider the calling position, the `wg` engine initialization might need to be rearranged accordingly together.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
hermet: @tinyjin If so, ThorVG may need to redesign the consistent engine initialization interfaces. Could you please suggest an example JS usage for `initWgInstance`?

src/bindings/wasm/tvgWasmLottieAnimation.cpp
hermet: You don't need to static them, please just adding a boolean (i.e: initialized) for avoiding the multiple initialization.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
SergeyLebedkin: @tinyjin 
`.compatibleSurface ` parameter is not mandatory for linux/macos. So if in browser its not mandatory, please, remove it

src/bindings/wasm/tvgWasmLottieAnimation.cpp
hermet: Somehow, this kind of usage might look better and be preferred.

```
   await _module.init();  //hide wg instances get ready in it.
   this._TVG = new _module.TvgLottieAnimation(engine, `#${this._canvas!.id}`);
```

```
    this._TVG = new _module.TvgLottieAnimation(engine, `#${this._canvas!.id}`);
    await _module.ready();  //hide wg instances get ready in it.
    ...
```

tinyjin: @hermet

I tried to move single initialization status to C++ binding side. To do that, we have to call `emscripten_sleep` function for asynchronous logic. However duplicated calls of `emscripten_sleep` won't be properly worked.

For example, when we have multiple players, we need condition to block for duplicated WG initialization logics in C++ side, then multiple `emscripten_sleep()` are definitely called at the same time.

But that point makes WG initialization fall into infinite loop, because C++ doesn't actually support asynchronous processing in single thread, conversely JavaScript does even in single thread.

While C++ waits for requesting WG adapter/device by calling `emscripten_sleep`, another `emscripten_sleep()` from the other players(to avoid duplicated calls) raises deadlock and they just keep waiting without running back for prior WG initialization. (_Meaning that we'll meet infinite loop in multiple players case_).

Thus in my opinion, **we should have condition for avoiding multiple WG initializations in JS side not binding.**

hermet: @tinyjin module.init() can be dealt with a global resource, it's expected to call just once in app, not multiple times.  What do you think?
```
   //must be called just once.
   await _module.init();  //hide wg instances get ready in it.
   
   ...
   //per player instances.
   this._TVG = new _module.TvgLottieAnimation(engine, `#${this._canvas!.id}`);  
   
```

tinyjin: @hermet It's possible, but when module loaded once in global, we don't know WG instance is required.
Some users could just import SW+WG combined binary and decide not to use WG. Then preloaded WG Instances will be occupied some resources.

If it's not critical, global initialization in out of component would be good.

hermet: @tinyjin Please take a look at the engine initialization of ThorVG. Ultimately, likewise, the user will decide the engine based on their preferences and requirements. The module init could have an engine types and initialize only with them.
```
await _module.init("default");  
await _module.init("software"); 
await _module.init("webgpu");
await _module.init("all"); //all possible engines
```

Please weigh the pros and cons of this approach compared to your suggestion. Which one is more difficult to understand and use from the WASM interfaces?

tinyjin: I agree with this way. Module initialization is lazy, some users might want to have their own strategy. In the end, we should let users choose options.

Currently, we don't have external interfaces other than component tag. I assume this update will come along with out of this patch.

So far, we can decide from two options:
1. Multiple attempt, blocking duplicated call from JS side
2. Global one-time initialization when module loaded

Option 1 is good for users. They won't waste resources. WG resources are only loaded when user has WG based animation.

Option 2 is good for us. Even if it could bring redundant resources, we can keep consistent and simpler interface.

If I had to decide, I'd go with Option 1 because user side benefit is priority, I believe.

hermet: > some users might want to have their own strategy

Ok, could you please suggest this you are considering?

> Currently, we don't have external interfaces other than component tag. I assume this update will come along with out of this patch.

May I ask what is `component tag` exactly? If it's necessary it's time to design the interface.

> Option 2 is good for us. Even if it could bring redundant resources, we can keep consistent and simpler interface.

@tinyjin I didn't suggest for us, but for developers who will implement directly using the WASM binding APIs. Now it's time to ensure there were 'redundant resources.' If it does, could you provide any measurable data?

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -33,6 +34,40 @@ using namespace tvg;
...
+    //Init WebGPU
+    if (!instance) instance = wgpuCreateInstance(nullptr);
+
+    // request adapter
+    if (!adapter) {
+        const WGPURequestAdapterOptions requestAdapterOptions { .nextInChain = nullptr, .powerPreference = WGPUPowerPreference_HighPerformance, .forceFallbackAdapter = false };
+        auto onAdapterRequestEnded = [](WGPURequestAdapterStatus status, WGPUAdapter adapter, char const * message, void * pUserData) { *((WGPUAdapter*)pUserData) = adapter; };
+        wgpuInstanceRequestAdapter(instance, &requestAdapterOptions, onAdapterRequestEnded, &adapter);
+        while (!adapter) emscripten_sleep(10);
+        assert(adapter);
```
hermet, line 55: please remove assert()

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -33,6 +34,40 @@ using namespace tvg;
...
+
+    // request device
+    WGPUFeatureName featureNames[32]{};
+    size_t featuresCount = wgpuAdapterEnumerateFeatures(adapter, featureNames);
+    if (!device) {
+        const WGPUDeviceDescriptor deviceDesc { .nextInChain = nullptr, .label = "The device", .requiredFeatureCount = featuresCount, .requiredFeatures = featureNames };
+        auto onDeviceRequestEnded = [](WGPURequestDeviceStatus status, WGPUDevice device, char const * message, void * pUserData) { *((WGPUDevice*)pUserData) = device; };
+        wgpuAdapterRequestDevice(adapter, &deviceDesc, onDeviceRequestEnded, &device);
+        while (!device) emscripten_sleep(10);
+        assert(device);
```
hermet, line 66: please remove assert()

hermet (approved):

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -77,13 +110,14 @@ struct TvgSwEngine : TvgEngineMethod
...
 {
     #ifdef THORVG_WG_RASTER_SUPPORT
-        WGPUInstance instance{};
         WGPUSurface surface{};
     #endif
 
     ~TvgWgEngine()
     {
         #ifdef THORVG_WG_RASTER_SUPPORT
+            wgpuDeviceRelease(device);
```
hermet, line 119: Hmm... this is a bad. initialize them in global init()  and terminate them by an instance...
A better is to init() - term() pairing.

tinyjin: @hermet Agree, instance/adapter/device should be released in global scope like `init()`. I'll leave surfaceRelease here and rearrange the others in new `term()` function.

tinyjin: Updated for this, please have a look. In binding, we'll need to call `init()` - `term()` at pair

hermet:
@tinyjin good to push ?

tinyjin:
> @tinyjin good to push ?

Yes so far looks ready

PR #2730, opened 2024-09-12
examples: updated lottie resources

hermet:
.


hermet:
perf-test sholud be updated together: + @tinyjin

Issue #2178, opened 2024-04-14
lottie: support missing Advanced Text feature


hermet:
- [x] Range Selector https://lottiefiles.github.io/lottie-docs/text/#text-range
- [x] Alignment Options https://lottiefiles.github.io/lottie-docs/text/#text-alignment-options

tinyjin:
The Text Advanced Feature aims to resolve broken texts in Lottie.

In the early stages of this process, text rendering was improved by fixing text grouping and newly supporting the Text Tracking feature.

See also:
- Text Tracking Support :#2254
- Text Alignment/Justification Correction : #2218

tinyjin:
The full support for Text Range Selector will be gradually added in the following steps.

### Phase 1
Initial & Partial Support
- Text Range Selector (Units)
- Text Style (Color, Opacity, Width, Transform)

### Phase 2
Full Support of Text Style
- Hue, Saturation, Brightness, Blur, Spacing

### Phase3
Full Support of Range Selector
- Easing, Randomize, Shape, Smoothness, Text Based, Max Amount

hermet:
@mgrudzinska, could you please share any resource that addresses the bug fixed by this commit?

hermet:
Separated the text path as it is uncommon: https://github.com/thorvg/thorvg/issues/2888

PR #2738, opened 2024-09-16
lottie/text: Support Text Randomize

tinyjin:
Issue: #2178 

[text-range-randomize.json](https://github.com/user-attachments/files/17008794/text-range-randomize.json)

https://github.com/user-attachments/assets/00003b27-efcd-4e32-87ca-acc7dfe1c14d

---

Applied Text Range randomization.

On each parse, if the `randomize` is enabled, the start and end of the Text Range are redefined with the same gap as the original range.

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -110,6 +110,25 @@ void LottieSlot::assign(LottieObject* target)
 }
 
 
+void LottieTextRange::range(float frameNo, size_t totalLen, float& start, float& end)
+{
+    float divisor = rangeUnit == Unit::Percent ? (100.0f / totalLen) : 1;
+    auto offset = this->offset(frameNo) / divisor;
+    start = nearbyintf(this->start(frameNo) / divisor) + offset;
+    end = nearbyintf(this->end(frameNo) / divisor) + offset;
+
+    if (start > end) std::swap(start, end);
+
+    if (random) {
+        auto range = end - start;
+        auto len = rangeUnit == Unit::Percent ? 100 : totalLen;
+
+        start = random % int(len - range);
+        end = start + range;
+    }
+}
+
+
 LottieImage::~LottieImage()
 {
     free(b64Data);
--- a/src/loaders/lottie/tvgLottieModel.h
+++ b/src/loaders/lottie/tvgLottieModel.h
@@ -184,8 +184,10 @@ struct LottieTextRange
     Based based = Chars;
     Shape shape = Square;
     Unit rangeUnit = Percent;
+    uint8_t random = 0;
     bool expressible = false;
-    bool randomize = false;
+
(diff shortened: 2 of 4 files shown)


src/loaders/lottie/tvgLottieModel.h
```diff
@@ -184,8 +184,12 @@ struct LottieTextRange
     Based based = Chars;
     Shape shape = Square;
     Unit rangeUnit = Percent;
+    uint8_t seed = 0;
```
hermet, line 187: Logically it's not a seed, just a random value.

Suggestion:  
```
uint8_t random = 0;  // allocate random = rand(sizeof(uint8_t)) in Parsing "rd"
bool randomize = false;  <--- remove
```
You can apply the random when it's value is not zero.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1123,7 +1123,9 @@ void LottieParser::parseTextRange(LottieText* text)
                     else if (KEY_AS("ne")) parseProperty<LottieProperty::Type::Float>(selector->minEase);
                     else if (KEY_AS("a")) parseProperty<LottieProperty::Type::Float>(selector->maxAmount);
                     else if (KEY_AS("b")) selector->based = (LottieTextRange::Based) getInt();
-                    else if (KEY_AS("rn")) selector->randomize = (bool) getInt();
+                    else if (KEY_AS("rn")) {
```
hermet, line 1126:
```
else if (KEY_AS("rn") && getInt()) selector->random = rand();
```

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -110,6 +110,25 @@ void LottieSlot::assign(LottieObject* target)
...
+void LottieTextRange::range(float frameNo, size_t totalLen, float& start, float& end)
+{
+    float divisor = rangeUnit == Unit::Percent ? (100.0f / totalLen) : 1;
+    auto offset = this->offset(frameNo) / divisor;
+    start = nearbyintf(this->start(frameNo) / divisor) + offset;
+    end = nearbyintf(this->end(frameNo) / divisor) + offset;
+
+    if (start > end) std::swap(start, end);
+
+    if (random) {
```
hermet, line 122:
```
if (random == 0) return;
```

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -110,6 +110,25 @@ void LottieSlot::assign(LottieObject* target)
...
+        auto range = end - start;
+        auto len = rangeUnit == Unit::Percent ? 100 : totalLen;
+
```
hermet, line 125: --space

hermet:
https://github.com/thorvg/thorvg/commit/9c5933a4e59159ff1b12c98630cc13bc4e262c65

Issue #2685, opened 2024-08-27
wg_engine: Lottie rendering compatibility


hermet:
This list shows the compatibility of Lottie rendering with the WebGPU backend.
https://github.com/thorvg/thorvg/tree/main/examples/resources/lottie

- [ ] 5317-fireworkds
- [ ] 5344-honey-sack-hud
- [ ] a_mountain
- [ ] birth_stone_logo
- [ ] insta_camera
- [ ] polystar
- [ ] starburst
- [ ] starstrips
- [ ] expressions/11272
- [ ] expressions/intelia_logo_animation
- [ ] expressions/jolly_walker
- [x] 11555
- [x] 1643-exploding-star
- [x] 27746-joypixels-partying-face-emoji-animation
- [x] abstract_circle
- [x] alien.json
- [x] anubis
- [x] balloons_with_string
- [x] batman
- [x] calculator
- [x] card_hover
- [x] cat_loader
- [x] coin
- [x] confetti
- [x] confetti2
- [x] customer
- [x] dancing_book
- [x] dancing_star
- [x] day_to_night
- [x] dodecahedron
- [x] down
- [x] dropball
- [x] duck
- [x] earth_hour
- [x] emoji
- [x] emoji_enjoying
- [x] error_404
- [x] fiery_skull
- [x] fleche
- [x] flipping_page
- [x] fly_in_beaker
- [x] focal_test
- [x] foodrating
- [x] frog_vr
- [x] fun_animation
- [x] funky_chicken
- [x] game_finished
- [x] geometric
- [x] glow_loading
- [x] gradient_background
- [x] gradient_infinite
- [x] gradient_sleepy_loader
- [x] gradient_smoke
- [x] growup
- [x] guitar
- [x] hamburger
- [x] happy_holidays
- [x] hola
- [x] happy_trio                                   
- [x] holdanimation
- [x] hourglass
- [x] isometric
- [x] kote
- [x] la_communaute
- [x] like
- [x] like_button
- [x] loading_rectangle
- [x] lolo
- [ ] lolo_walk
- [ ] loveface_emoji
- [ ] masking
- [ ] material_wave_loading
- [ ] merging_shapes
- [ ] message
- [ ] morphing_anim
- [ ] open_envelope
- [ ] personal_character
- [ ] property_market
- [ ] radar
- [ ] ripple_loading_animation
- [ ] rufo
- [ ] sad_emoji
- [ ] sample
- [ ] seawalk
- [ ] skullboy
[...]

Issue #2313, opened 2024-05-24
Refactoring the renderer - raster engines.


hermet:
Time to clean up the code. In order to maintain code, data structure consistency, it is necessary to unify the following separate definitions:

```
Point / GlPoint / WgPoint
```

There are two possible approaches:

1. Add the necessary math functions in `renderer/tvgMath.h` and retain them as `Point`.
2. Introduce `RenderPoint` inheriting `Point` in `tvgRender.h`, add the necessary point methods to it, and replace all `Point`/`GlPoint`/`WgPoint` usages inside ThorVG with `RenderPoint`

hermet:
cc @SergeyLebedkin @RuiwenTang

RuiwenTang:
I prefer the first option, it also reduces some binary size ：）

PR #2739, opened 2024-09-16
wg_engine: fix strokes triangulation artifacts

SergeyLebedkin:
Fixed visual arifacts:

* angular circles
* extra triangles
* wrongly closed polylines

Animations to verify:
* 5344-honey-sack-hud.json
* birth_stone_logo.json
* insta_camera.json
* polystar.json
* starstrips.json

see: https://github.com/thorvg/thorvg/issues/2685

### DIFF
--- a/src/renderer/wg_engine/tvgWgGeometry.cpp
+++ b/src/renderer/wg_engine/tvgWgGeometry.cpp
@@ -65,7 +65,10 @@ void WgPolyline::appendPoint(WgPoint pt)
 {
     if (pts.count > 0) {
         float distance = pts.last().dist(pt);
-        if (distance > 0) {
+        // adjust precision because of real user data points
+        // can be further than the accepted accuracy, 
+        // but still be considered tha same
+        if (!tvg::zero(distance*1e-1)) {
             // update min and max indexes
             iminx = pts[iminx].x >= pt.x ? pts.count : iminx;
             imaxx = pts[imaxx].x <= pt.x ? pts.count : imaxx;
@@ -217,13 +220,13 @@ void WgGeometryData::appendRect(WgPoint p0, WgPoint p1, WgPoint p2, WgPoint p3)
 }
 
 
-void WgGeometryData::appendCircle(WgPoint center, float radius)
+void WgGeometryData::appendCircle(WgPoint center, float radius, float scale)
 {
     uint32_t indexCenter = positions.pts.count;
     positions.appendPoint(center);
     uint32_t index = positions.pts.count;
     positions.appendPoint({ center.x + gMath->sinus[0] * radius, center.y + gMath->cosin[0] * radius });
-    uint32_t nPoints = (uint32_t)(radius * 2.0f);
+    uint32_t nPoints = (uint32_t)(scale * radius * 2.0f);
     nPoints = nPoints < 8 ? 8 : nPoints;
     const uint32_t step = gMath->sinus.count / nPoints;
     for (uint32_t i = step; i < gMath->sinus.count; i += step) {
@@ -279,7 +282,7 @@ void WgGeometryData::appendBlitBox()
 }
 
 
-void WgGeometryData::appendStrokeDashed(const WgPolyline* polyline, const RenderStroke *stroke)
+void WgGeometryData::appendStrokeDashed(const WgPolyline* polyline, const RenderStroke *stroke, float scale)
 {
     assert(stroke);
     assert(polyline);
@@ -308,7 +311,7 @@ void WgGeometryData::appendStrokeDashed(const WgPolyline* polyline, const Render
                 currentLength += stroke->dashPattern[dashIndex];
                 // append stroke if dash
                 if (dashIndex % 2 != 0) {
-                    appendStroke(&dashed, stroke);
+                    appendStroke(&dashed, stroke, scale);
                     dashed.clear();
                 }
             }
@@ -318,14 +321,14 @@ void WgGeometryData::appendStrokeDashed(const WgPolyline* polyline, const Render
         // draw last subline
         if (dashIndex % 2 == 0) {
             dashed.appendPoint(pts.last());
-            appendStroke(&dashed, stroke);
+            appendStroke(&dashed, stroke, scale);
             dashed.clear();
         }
     }
 }
 
 
-void WgGeometryData::appendStrokeJoin(const WgPoint& v0, const WgPoint& v1, const WgPoint& v2, StrokeJoin join, float halfWidth, float miterLimit)
+void WgGeometryData::appendStrokeJoin(const WgPoint& v0, const WgPoint& v1, const WgPoint& v2, StrokeJoin join, float halfWidth, float miterLimit, float scale)
 {
     WgPoint dir0 = (v1 - v0).normal();
     WgPoint dir1 = (v2 - v1).normal();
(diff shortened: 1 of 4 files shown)


src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -65,7 +65,7 @@ void WgPolyline::appendPoint(WgPoint pt)
 {
     if (pts.count > 0) {
         float distance = pts.last().dist(pt);
-        if (distance > 0) {
+        if (!tvg::zero(distance * 1e-1)) {
```
mgrudzinska, line 68: can you please explain why tvg::zero is not enough?

mgrudzinska: also you are using different factors in different places for values of different meanings. can you also explain please?

SergeyLebedkin: @mgrudzinska 
In the animations sometimes we can meet shapes with polylines like this:
```
    shape->moveTo(-1.74846e-06, -40);
    shape->lineTo(34.641, 20);
    shape->lineTo(-34.641, 20);
    shape->lineTo(4.76995e-07, -40);
    shape->close();
```
They are manually closed, but not close enough to detect it.

mgrudzinska: I don't think this is the way to solve this. 
try this:
```
        shape->moveTo(-1e-05, -40);         //try also: shape->moveTo(-1.1e-05, -40); 
        shape->lineTo(34.641, 20);
        shape->lineTo(-34.641, 20);
        shape->lineTo(0, -40);
        shape->close();
        shape->strokeFill(255, 0, 0);
        shape->strokeWidth(15);
        shape->translate(200,200);
```
the result is clearly not perfect for SW. and GL (but in exaclty the same way not perfect), but it works for WG. if we change this shape to start in ```shape->moveTo(-1.1e-05, -40); ``` then the results for SW and GL don't change (still not perfect in the same way), but in WG there is a different problem. 

I think that if such cases are present in  examples we have to consider changing the epsilon value, but for all engines in the same way. 
I don't understand why WG needs a different value of epsilon especially only in one place.

mgrudzinska: you are right (private correspondence) that non of the engines rely on the epsilon values used in tvg::zero. 

I think that you may need you own epsilon, but let me test some cases and I'll come back to this

SergeyLebedkin: Good idea to add own epsilon. 
Please, check the solution

src/renderer/wg_engine/tvgWgRenderData.cpp
```diff
@@ -375,21 +377,21 @@ void WgRenderDataShape::updateMeshes(WgContext &context, const RenderShape &rsha
         if (!rshape.stroke->strokeTrim(trimBegin, trimEnd)) { trimBegin = 0.0f; trimEnd = 1.0f; }
         if (rshape.stroke->trim.simultaneous) {
             for (uint32_t i = 0; i < polylines.count; i++)
```
mgrudzinska, line 379: this is not correct. please check insta_camera.json - rounded corners are not applied

mgrudzinska (dismissed):
please add some description in commit msg, especially since your are fixing more than one problem here. But also would be great, if you could split this commit into two separate commits. now one of the fixes is ok, but the other one is not, so the first one can not be applied.

src/renderer/wg_engine/tvgWgGeometry.cpp
mgrudzinska: now I like it!

and I found the problem, which was bothering me, why the result is not symmetrical - https://github.com/thorvg/thorvg/pull/2743

src/renderer/wg_engine/tvgWgRenderData.cpp
mgrudzinska: this comment should be next to l.354...

mgrudzinska: it is still a valid comment.
Lottie example, 1280x1280, NUM_PER_ROW 3, NUM_PER_COL 2:

current:
Zrzut ekranu 2024-09-18 o 01 28 01

expected:
Zrzut ekranu 2024-09-18 o 01 28 10

mgrudzinska:
@SergeyLebedkin please try to keep the convention in commit msg:
```
1 line summary

description 
```
please note the empty line between summary and description

mgrudzinska (approved):

src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -28,6 +28,8 @@
 #include "tvgRender.h"
 #include "tvgArray.h"
 
+#define WG_FLOAT_EPSILON 1e-4
```
hermet, line 31: --

src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -70,6 +72,8 @@ struct WgMath
 
     void initialize();
     void release();
+
+    static bool zero(float val) { return (fabsf(val) <= WG_FLOAT_EPSILON) ? true : false; }
```
hermet, line 76: --

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -65,7 +65,7 @@ void WgPolyline::appendPoint(WgPoint pt)
 {
     if (pts.count > 0) {
         float distance = pts.last().dist(pt);
-        if (distance > 0) {
+        if (!WgMath::zero(distance)) {
```
hermet, line 68: @SergeyLebedkin Please keep your original proposal:
` if (!tvg::zero(distance * 1e-1)) {`

If possible, you can add a comment with this special handling 'reason' there.

src/renderer/wg_engine/tvgWgGeometry.cpp
```diff
@@ -334,12 +334,12 @@ void WgGeometryData::appendStrokeJoin(const WgPoint& v0, const WgPoint& v1, cons
...
     WgPoint offset1 = nrm1 * halfWidth;
     if (join == StrokeJoin::Round) {
-        appendCircle(v1, halfWidth);
+        appendCircle(v1, halfWidth, scale);
     } else if (join == StrokeJoin::Bevel) {
         appendRect(v1 - offset0, v1 + offset0, v1 - offset1, v1 + offset1);
     } else if (join == StrokeJoin::Miter) {
         WgPoint nrm = (nrm0 + nrm1);
-        if (!tvg::zero(dir0.x * dir1.y -  dir0.y * dir1.x)) {
+        if (!WgMath::zero(dir0.x * dir1.y - dir0.y * dir1.x)) {
```
hermet, line 342: Please revert to `tvg::zero()`

hermet (commented):
@SergeyLebedkin Please check comments, it is deemed unnecessary because you are trying to hide an root problem with a fake zero. I think we should keep use one unified math function in ThorVG.  It is not good to keep spreading code fragments like this: https://github.com/thorvg/thorvg/issues/2313

hermet (approved):

Issue #2131, opened 2024-04-02
SVG stroke-width with a percentage uses the width instead of the diagonal over sqrt(2)


MewPurPur:
https://svgwg.org/svg2-draft/painting.html#StrokeWidth

![image](https://github.com/thorvg/thorvg/assets/85438892/1be3fed5-641d-4477-9ad2-8ca71d248d31)

Example: `<svg width="100" height="50" xmlns="http://www.w3.org/2000/svg"><rect x="15" y="15" width="70" height="20" fill="none" stroke="#def" stroke-width="10%"/><rect x="15" y="15" width="70" height="20" fill="none" stroke="#fac" stroke-opacity=".6" stroke-width="10"/></svg>`

Conforming to the SVG specification, 10% should amount to ~7.9px, as shown below - the two rects don't perfectly align. This isn't the case if you test this with ThorVG viewer right now. There may be more attributes with this flaw.

![image](https://github.com/thorvg/thorvg/assets/85438892/bf6f947d-1d4d-471a-9210-7186014d52a6)

hermet:
ThorVG:
![image](https://github.com/user-attachments/assets/3a753c67-e6b9-4ab7-87a1-55511463d3c7)

MewPurPur:
Yeah, this is wrong. The rectangles shouldn't align perfectly. You're taking 10% of the width, when stroke-width is not an x or y-based. The specification says that such attributes should be based on the diagonal length.

hermet:
@MewPurPur @JSUYA The fix will be in v0.15.0. Thanks.

PR #2750, opened 2024-09-20
svg_loader: Fix calculation when stroke-width unit is percentage

JSUYA:
When stroke-width unit is percentage, loader refer to the normalized diagonal of viewport.
+) Add Width type so as not to affect existing types.

https://svgwg.org/svg2-draft/painting.html#StrokeWidth
https://svgwg.org/svg2-draft/coords.html#Units

example)
```
<svg width="100" height="50" xmlns="http://www.w3.org/2000/svg">
   <rect x="15" y="15" width="70" height="20" fill="none" stroke="#F00" stroke-width="10%"/>
   <rect x="15" y="15" width="70" height="20" fill="none" stroke="#00F" stroke-opacity=".3" stroke-width="10"/>
</svg>

```

related issue: https://github.com/thorvg/thorvg/issues/2131

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -177,6 +177,7 @@ static float _toFloat(const SvgParser* svgParse, const char* str, SvgParserLengt
     else if (strstr(str, "%")) {
         if (type == SvgParserLengthType::Vertical) parsedValue = (parsedValue / 100.0f) * svgParse->global.h;
         else if (type == SvgParserLengthType::Horizontal) parsedValue = (parsedValue / 100.0f) * svgParse->global.w;
+        else if (type == SvgParserLengthType::Diagonal) parsedValue = (sqrtf(powf(svgParse->global.w, 2) + powf(svgParse->global.h, 2)) / sqrtf(2.0f)) * (parsedValue / 100.0f);
         else //if other than it's radius
         {
             float max = svgParse->global.w;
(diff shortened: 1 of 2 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1066,7 +1067,7 @@ static void _handleStrokeDashOffsetAttr(SvgLoaderData* loader, SvgNode* node, co
 static void _handleStrokeWidthAttr(SvgLoaderData* loader, SvgNode* node, const char* value)
 {
     node->style->stroke.flags = (node->style->stroke.flags | SvgStrokeFlags::Width);
-    node->style->stroke.width = _toFloat(loader->svgParse, value, SvgParserLengthType::Horizontal);
+    node->style->stroke.width = _toFloat(loader->svgParse, value, SvgParserLengthType::Width);
```
mgrudzinska, line 1070: I'd introduce a type "Diagonal" instead of "Width". what do you think?
no matter what the answer will be, I approved :)

mgrudzinska (approved):
left a comment, but feel free to resolve if you don;t agree

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: Diagonal is more appropriate. Thanks!! :) I updated PR.

MewPurPur:
I found that this should also apply for the "r" attribute.

PR #2757, opened 2024-09-23
svg_loader: fix circle radius if in percentages

mgrudzinska:
regarding https://github.com/thorvg/thorvg/pull/2750#issuecomment-2364333927 by @MewPurPur 

before:
Zrzut ekranu 2024-09-23 o 13 56 15

after:
Zrzut ekranu 2024-09-23 o 13 55 38

sample:
```
<svg width="1600" height="600" xmlns="http://www.w3.org/2000/svg">
     <rect x="10%" y="100" width="40%" height="400" fill="none" stroke="#00f" stroke-width="10%" />
     <circle cx="20%" cy="250" r="10%" opacity="0.5" />
</svg>
```

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1615,7 +1615,7 @@ static constexpr struct
 } circleTags[] = {
     {"cx", SvgParserLengthType::Horizontal, sizeof("cx"), offsetof(SvgCircleNode, cx)},
     {"cy", SvgParserLengthType::Vertical, sizeof("cy"), offsetof(SvgCircleNode, cy)},
-    {"r", SvgParserLengthType::Other, sizeof("r"), offsetof(SvgCircleNode, r)}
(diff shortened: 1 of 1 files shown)


mgrudzinska:
@JSUYA after we merge this I'd like to remove the ```SvgParserLengthType::Other```. there is a comment that it can be used by gradient, but it is not. do you remember any use case for this scenario?

JSUYA:
> @JSUYA after we merge this I'd like to remove the `SvgParserLengthType::Other`. there is a comment that it can be used by gradient, but it is not. do you remember any use case for this scenario?

I looked into it, the code in tvgSvgLoader.cpp lines 181~187 is part of a very old code (maybe 2017). It may be a very old spec, or a spec for a limited implementation. So I think Other can be removed.

JSUYA (approved):
Since `Other` type is only used here, I think it would be a good to also remove another code related to `Other` from this PR.

hermet:
\+ @MewPurPur

PR #2766, opened 2024-09-25
wg_engine: fix stroke trimming

mgrudzinska:
Handle very close values of begin and end.

### DIFF
--- a/src/renderer/wg_engine/tvgWgGeometry.h
+++ b/src/renderer/wg_engine/tvgWgGeometry.h
@@ -124,14 +124,14 @@ struct WgVertexBuffer {
     }
 
     // append source vertex buffer in index range from start to end (end not included)
-    void appendRange(const WgVertexBuffer& buff, size_t start_index, size_t end_index) {
-        if (start_index <= end_index)
-            for (size_t i = start_index; i < end_index; i++)
-                append(buff.vbuff[i]);
-        if (start_index > end_index) {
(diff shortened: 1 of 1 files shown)


src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -124,15 +124,15 @@ struct WgVertexBuffer {
     }
 
     // append source vertex buffer in index range from start to end (end not included)
-    void appendRange(const WgVertexBuffer& buff, size_t start_index, size_t end_index) {
-        if (start_index <= end_index)
-            for (size_t i = start_index; i < end_index; i++)
-                append(buff.vbuff[i]);
-        if (start_index > end_index) {
+    void appendRange(const WgVertexBuffer& buff, size_t start_index, size_t end_index, bool loop) {
```
SergeyLebedkin, line 127: make a loop value false by default

SergeyLebedkin, line 127: make a loop value false by default

SergeyLebedkin: So please, at least mention it in comments

mgrudzinska: @SergeyLebedkin please check https://github.com/thorvg/thorvg/pull/2771, should be more general

mgrudzinska:
https://github.com/thorvg/thorvg/pull/2771

Issue #2783, opened 2024-09-29
wg_engine: require code refactoring


hermet:
CompositeMethod::ClipPath will be replaced with the Scene::clip() (currently -> Paint::clip()) method. gl_engine needs to refactor to remove the enum value usage before proceeding with its removal.

![image](https://github.com/user-attachments/assets/7f0df78b-554b-4612-8b6e-7122e7bdd650)

see:
```
    /**
     * @brief Clip the drawing region of the paint object.
     *
     * This function restricts the drawing area of the paint object to the specified shape's paths.
     *
     * @param[in] clipper The shape object as the clipper.
     *
     * @retval Result::NonSupport If the @p clipper type is not Shape.
     *
     * @note @p clipper only supports the Shape type.
     * @note Experimental API
     */
    Result Paint::clip(std::unique_ptr<Paint> clipper);
```

SergeyLebedkin:
@hermet 
wg_engine did not use CompositeMethod::ClipPath enum value at all. 
wg_engine uses only clips count in RenderMethod::prepare stage.
So nothing to change in a case of enum values

hermet:
@SergeyLebedkin it does. Please check the screenshot.

PR #2784, opened 2024-09-29
wg_engine: CompositeMethod::ClipPath will be replaced with the Scene::clip()

SergeyLebedkin:
wg_engine needs to refactor to remove the enum value usage before proceeding with its removal.
Please, see TODO in tvgWgPipelines.h/cpp

see https://github.com/thorvg/thorvg/issues/2783

### DIFF
--- a/src/renderer/wg_engine/tvgWgPipelines.h
+++ b/src/renderer/wg_engine/tvgWgPipelines.h
@@ -65,7 +65,8 @@ public:
     WGPURenderPipeline radial[2]{};
     WGPURenderPipeline linear[2]{};
     WGPURenderPipeline image[2]{};
-    WGPURenderPipeline sceneComp[12];
+    WGPURenderPipeline sceneClip;
+    WGPURenderPipeline sceneComp[12]; // TODO: update to 11 after removing CompositeMethod::ClipPath enum value
     WGPURenderPipeline sceneBlend;
     WGPURenderPipeline blit{};
     WGPURenderPipeline clipPath{};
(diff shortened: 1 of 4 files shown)


src/renderer/wg_engine/tvgWgPipelines.h
```diff
@@ -65,7 +65,8 @@ class WgPipelines {
     WGPURenderPipeline radial[2]{};
     WGPURenderPipeline linear[2]{};
     WGPURenderPipeline image[2]{};
-    WGPURenderPipeline sceneComp[12];
+    WGPURenderPipeline sceneClip;
+    WGPURenderPipeline sceneComp[12]; // TODO: update to 11 after removing CompositeMethod::ClipPath enum value
```
hermet, line 69: I can't imagine the reason this. I think it would be better to remove it altogether.

SergeyLebedkin: @hermet 
> I can't think of a reason for this. The `Composite::ClipPath` command is never requested from the pipeline. it would be better to remove it altogether.

It was requested, but it was hiden in the routines. All other composite methods must be it this array, because pipelines request by composite method as an index.

hermet: @SergeyLebedkin  please use `[idx - 1]` no need to remove dead resource.

hermet (approved):

hermet (commented):
@SergeyLebedkin good, just two nit-pick please double check.

Issue #2794, opened 2024-10-01
svg: valgrind memory issue report


hermet:
![image](https://github.com/user-attachments/assets/1702e7a0-8565-4e28-a280-4fc10f3c3e89)
https://github.com/thorvg/thorvg/pull/2793

hermet:
This recently occurred, and it seems something caused this break. A clear case is needed to reproduce the issue(I couldn't produce this on my local machine yet). Or at least point out the commit which breaks this.

PR #2801, opened 2024-10-02
common: Prevent use of uninitialised values error

JSUYA:
Call memset after malloc to avoid "use of uninitialised values" errors on later.

https://github.com/thorvg/thorvg/issues/2794

### DIFF
--- a/src/common/tvgCompressor.cpp
+++ b/src/common/tvgCompressor.cpp
@@ -442,7 +442,7 @@ size_t b64Decode(const char* encoded, const size_t len, char** decoded)
     if (!decoded || !encoded || len == 0) return 0;
 
     auto reserved = 3 * (1 + (len >> 2)) + 1;
-    auto output = static_cast<char*>(malloc(reserved * sizeof(char)));
+    auto output = static_cast<char*>(calloc(reserved, sizeof(char)));
     if (!output) return 0;
     output[reserved - 1] = '\0';
 
@@ -490,3 +490,4 @@ unsigned long djb2Encode(const char* str)
 }
 
 }
+


src/common/tvgCompressor.cpp
```diff
@@ -444,6 +444,7 @@ size_t b64Decode(const char* encoded, const size_t len, char** decoded)
     auto reserved = 3 * (1 + (len >> 2)) + 1;
     auto output = static_cast<char*>(malloc(reserved * sizeof(char)));
```
hermet, line 445: 1. please use `calloc()`
2. Shouldn't _simpleXmlSkipWhiteSpace stop the iteration when the string ends `\0`?

JSUYA: 1. Changed.
2. As far as I know, `itrEnd` of `_simpleXmlSkipWhiteSpace` is set by strlen(). So, the end of `itrEnd` is probably `\0`. Is there any problem or suspicious point?

mgrudzinska (approved):

hermet:
@JSUYA You know, the `b64Decode` output is initialized with '\0' here, obviously, but the uninitialized memory violation has been detected by the SVG loader, which leads to incorrect access. Initializing the memory here incurs a performance cost. A more fundamental fix might be better implemented on the SVG side.

hermet:
Eventually, this introduces a cost burden to other situation. It might be better to determine whether accessing invalid memory (even with 0) is the correct behavior.

hermet (changes requested):
Eventually, this introduces a cost burden to other situation. It might be better to determine whether accessing invalid memory (even with 0) is the correct behavior. Otherwise, \0 initialization should be updated properly to prevent the invalid data access.

hermet:
@mgrudzinska Since it may affect performance on a 32-bit system, changing data types from 32-bit to 64-bit should be done carefully. We should probably also consider the system architecture and decide on the data type at build time.

hermet:
https://github.com/thorvg/thorvg/pull/2811

PR #2802, opened 2024-10-02
wg_engine: fix clip path visual artifacts

SergeyLebedkin:
Fixed wrongly applied clip path area.

see like.json
before/after
image


mgrudzinska (approved):

hermet:
Hold off on this until the v0.15 release.

src/renderer/wg_engine/tvgWgCompositor.cpp
```diff
@@ -299,7 +299,7 @@ void WgCompositor::drawShapeClipped(WgContext& context, WgRenderDataShape* rende
     endRenderPass();
     // restore current render pass
     beginRenderPass(commandEncoder, target, false);
-    RenderRegion rect = shrinkRenderRegion(renderData->aabb);
```
hermet, line 302: @SergeyLebedkin Just want to make it sure, partial region rendering is no more after this?

PR #2804, opened 2024-10-03
examples: fix mingw build error

mgrudzinska:
M_PI was not declared error solved


hermet (approved):
@vtorri any opinion?

vtorri:
`M_PI`is used in `deg2rad()`. There is already such function in `tvgMath.h`. Maybe it would be better to include tvgath.h to factorize code

hermet:
@vtorri `tvgMath.h` is used internally and cannot be used in examples. We could potentially introduce `tvg::M_PI`, and remove all  `#define _USE_MATH_DEFINES` but that's a separate problem to deal with.

Issue #2651, opened 2024-08-18
Using viewbox tricks to zoom in on very big SVGs causes a freeze.


MewPurPur:
I'm making a vector graphics application and while I can't reproduce this, Windows users have given me reports that they get freezes while they zoom into graphics.

In order to implement infinite zoom in my app, I re-render the visible part on each frame, using an adjusted width-height-viewbox. However, some users have reported issues with this. Of the four people who I tested this with (plus myself), the two Linux users weren't able to reproduce it, but the three Windows users were able. 

Test SVG: `<svg width="37.432" height="37.786" xmlns="http://www.w3.org/2000/svg" viewBox="431.158 533.026 37.432 37.786"><path d="M1000 0 0 1000"/></svg>`

One of the users tried to paste this SVG into the ThorVG viewer.

![image](https://github.com/user-attachments/assets/ac3c9716-460d-40e9-bcc5-e061355d0fd9)

Another user tried to load the SVG using Godot and got a permanent freeze:

```gdscript
func _ready() -> void:
    print("Starting")
    for _i in 3:
        await get_tree().process_frame
    var svg = """
    <svg width="37.432" height="37.786" xmlns="http://www.w3.org/2000/svg" viewBox="431.158 533.026 37.432 37.786"><path d="M1000 0 0 1000"/></svg>
    """
    var img = Image.new()
    var err = img.load_svg_from_string(svg, 22.627)  # 22.627 is a scaling factor that should result in a 700x700 image, nothing too intense.
    print("Finished")
```

hermet:
At least, freeze shouldn't be occurred.

Racer911dash1:
I would like to add some more info. I hope it helps:
- It scales with size. The larger the line becomes the sooner it will freeze/crash when zooming in
- There is a significant CPU usage when panning around (Both Linux and Windows)
- it specifically only happens with a diagonal path vector. Anything else seems to work fine

MewPurPur:
> There is a significant CPU usage when panning around (Both Linux and Windows)

In the case of my app, it's because the graphic is redrawn every frame when you pan around. But that seems to be a side effect of ThorVG being way slower than one would expect.

hermet:
> > There is a significant CPU usage when panning around (Both Linux and Windows)
> 
> In the case of my app, it's because the graphic is redrawn every frame when you pan around. But that might be a side effect of ThorVG being way slower than one would expect.

[...]

PR #2805, opened 2024-10-03
sw_engine: portability++

mgrudzinska:
Replace long and long long with int64_t to ensure
sufficient viriables size.

note:
the overflow occured in tvgSwRle.cpp in _lineTo() below line 564.
`prod`, `px`, `py` exceeded 1e9 (ignoring the sign here) and adding/subtracting them caused the result to exceed the maximum value of long

### DIFF
--- a/src/renderer/sw_engine/tvgSwCommon.h
+++ b/src/renderer/sw_engine/tvgSwCommon.h
@@ -33,8 +33,8 @@
 #define SW_ANGLE_2PI (SW_ANGLE_PI << 1)
 #define SW_ANGLE_PI2 (SW_ANGLE_PI >> 1)
 
-using SwCoord = signed long;
-using SwFixed = signed long long;
+using SwCoord = int64_t;
+using SwFixed = int64_t;
 
 
 static inline float TO_FLOAT(SwCoord val)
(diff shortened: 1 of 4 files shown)


mgrudzinska:
on my win machine this sample crashed:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 40 40">
    <path d="M0 0 1311 1311" stroke-width="10" stroke="#ff0000"/>
</svg>
```
after this pr, even bigger line worked
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 40 40">
    <path d="M0 0 10311 10311" stroke-width="10" stroke="#ff0000"/>
</svg>
```

@hermet I changed long/long long only were I needed to prevent this crash, but I'd consider to replace all of them by portable types

hermet:
@mgrudzinska Since it may affect performance on a 32-bit system, changing data types from 32-bit to 64-bit should be done carefully. We should probably also consider the system architecture and decide on the data type at build time.

mgrudzinska:
32/64 support is a bigger subject and should be treated separately. this issue was solved in a different way

PR #2812, opened 2024-10-03
sw_engine: prevent long overflow

mgrudzinska:
If long has 32 bits on the system, it is possible
to go beyond the range of variable values.
As a workaroung introduced counter to prevent infinite loop.

### DIFF
--- a/src/renderer/sw_engine/tvgSwRle.cpp
+++ b/src/renderer/sw_engine/tvgSwRle.cpp
@@ -558,7 +558,12 @@ static void _lineTo(RleWorker& rw, const SwPoint& to)
         /* The fundamental value `prod' determines which side and the  */
         /* exact coordinate where the line exits current cell.  It is  */
         /* also easily updated when moving from one cell to the next.  */
+        size_t iter = 0;
         do {
+            if (++iter > UINT32_MAX) {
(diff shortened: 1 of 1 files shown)


src/renderer/sw_engine/tvgSwRle.cpp
```diff
@@ -558,7 +558,12 @@ static void _lineTo(RleWorker& rw, const SwPoint& to)
         /* The fundamental value `prod' determines which side and the  */
         /* exact coordinate where the line exits current cell.  It is  */
         /* also easily updated when moving from one cell to the next.  */
+        uint32_t iter = 0;
         do {
+            if (sizeof(long) != 8 && ++iter > 1e6) {
```
hermet, line 563: please remove `sizeof(long) != 8`

src/renderer/sw_engine/tvgSwRle.cpp
```diff
@@ -558,7 +558,12 @@ static void _lineTo(RleWorker& rw, const SwPoint& to)
         /* The fundamental value `prod' determines which side and the  */
         /* exact coordinate where the line exits current cell.  It is  */
         /* also easily updated when moving from one cell to the next.  */
+        uint32_t iter = 0;
```
hermet, line 561: `size_t`

hermet:
https://github.com/thorvg/thorvg/pull/2815

Issue #2808, opened 2024-10-03
wg_engine: memory leak?


hermet:
Need to clarify the memory leaks report.

Test 'Shape' webgpu with this build:
```
meson setup build -Dexamples=true -Dloaders=all -Db_sanitize="address,undefined" -Dengines="wg_beta"
```

```
Direct leak of 88 byte(s) in 1 object(s) allocated from:
    #0 0x7f687ee39887 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7f687c15b5ac in wgpuSurfaceGetCurrentTexture (/usr/local/lib/libwgpu_native.so+0x1025ac)
    #2 0x7f687d4e137a in tvg::Canvas::Impl::sync() ../src/renderer/tvgCanvas.h:129
    #3 0x7f687d4e137a in tvg::Canvas::sync() ../src/renderer/tvgCanvas.cpp:86
    #4 0x5633152527bf in tvgexam::Window::ready() ../examples/Example.h:187
    #5 0x563315243506 in tvgexam::main(tvgexam::Example*, int, char**, unsigned int, unsigned int, unsigned int, bool) ../examples/Example.h:533
    #6 0x563315246a67 in main ../examples/Shape.cpp:56
    #7 0x7f687b5d6d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58

Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x7f687ee39887 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7f687c14cf2b in wgpuDeviceCreateBindGroupLayout (/usr/local/lib/libwgpu_native.so+0xf3f2b)

Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x7f687ee39887 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7f687c156096 in wgpuDeviceGetQueue (/usr/local/lib/libwgpu_native.so+0xfd096)
    #2 0x7f687d8a669a in WgRenderer::target(WGPUInstanceImpl*, WGPUSurfaceImpl*, unsigned int, unsigned int, WGPUDeviceImpl*) ../src/renderer/wg_engine/tvgWgRenderer.cpp:299
[...]

PR #2814, opened 2024-10-03
wg_engine: fixed resource leaks

hermet:
issue: https://github.com/thorvg/thorvg/issues/2808

### DIFF
--- a/src/renderer/wg_engine/tvgWgRenderer.cpp
+++ b/src/renderer/wg_engine/tvgWgRenderer.cpp
@@ -59,6 +59,7 @@ void WgRenderer::release()
         adapter = nullptr;
         gpuOwner = false;
     }
+    releaseSurfaceTexture();
 }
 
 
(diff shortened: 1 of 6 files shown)


src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -246,13 +247,24 @@ bool WgRenderer::clear()
...
+
+
 bool WgRenderer::sync()
 {
     disposeObjects();
     if (!surface) return false;
-    // get current texture
-    WGPUSurfaceTexture surfaceTexture{};
+
+    releaseSurfaceTexture();
```
SergeyLebedkin, line 264: Im not sure about this approach... Surface is owner of the screen buffer textures, and we are not _crreate_ them, but _get_ them from surface

hermet: it is required by the wgpu guidance. and due to this all the resources are certainly leaked.

SergeyLebedkin (approved):

tinyjin (approved):

Issue #2817, opened 2024-10-03
infra: shutdown the unnecessary valgrind report


hermet:
![image](https://github.com/user-attachments/assets/439a2768-c03d-43b9-a2b1-6306f5031d63)

\+ @JSUYA

PR #2823, opened 2024-10-04
infra: Improve valgrind result analysis syntax

JSUYA:
Previously, if "definitely lost:" was included, it was always decided an error.
Therefore, I modified it to recognize the size and number of errors.

https://github.com/thorvg/thorvg/issues/2817


hermet (approved):

hermet:
@JSUYA cool, thanks!

PR #2826, opened 2024-10-04
wg_engine: fix long path decoding crash

SergeyLebedkin:
fixed crash with shape have too big path.
update vertex array sizes

see PictureSvg example:
image


hermet (approved):

tinyjin (approved):
This patch requires 4MB stack size in browser for WASM

PR #2827, opened 2024-10-07
svg_loader: Support id attribute for tvg::Accessor

JSUYA:
Supports access with tvg::Accessor using the value set in the id attribute of SVG.
Please refer to examples/Accessor.cpp


JSUYA:
We've discussed accessing elements using id in svg before, but there was no real requirement or clear design.
https://github.com/thorvg/thorvg/issues/1276
This patch is inspired by examples/interaction.cpp.
There is still no requirement, but I created this PR because I thought this approach could provide functionality in a similar way to lottie.

hermet (approved):

examples/Accessor.cpp
```diff
@@ -58,6 +58,12 @@ struct UserExample : tvgexam::Example
             return true;
         };
 
+        if (auto paint = picture->paint(tvg::Accessor::id("star"))) {
+            auto shape = static_cast<tvg::Shape*>(const_cast<tvg::Paint*>(paint));
+            shape->strokeFill(255, 255, 0);
+            shape->strokeWidth(5);
+        }
+
         if (!tvgexam::verify(accessor->set(picture.get(), f, nullptr))) return false;
```
hermet, line 68: 1. Suggestion to move your addition after 67 line to not break the context.
2. Appreciate you if you adding a comment what you're doing here for the users who really want to learn from your code something like this:
```
// Try to retrieve the shape that corresponds to the SVG node with the unique ID "star".
```

JSUYA: I updated PR, Thanks!

mgrudzinska:
I wondered what should happen if we have 'use' element, because I found [this](https://www.w3.org/TR/SVG2/struct.html#UseAnimations) :
The ‘[id](https://www.w3.org/TR/SVG2/struct.html#IDAttribute)’ attribute is cloned, like any other attribute, from the [corresponding element](https://www.w3.org/TR/SVG2/struct.html#TermCorrespondingElement) to the [element instance](https://www.w3.org/TR/SVG2/struct.html#TermElementInstance); This does not conflict with the requirement for ‘[id](https://www.w3.org/TR/SVG2/struct.html#IDAttribute)’ to be unique, because the clone and the original are in distinct node trees.
but it seems to be completely different than normal use of the 'use' element.

LGTM

PR #2828, opened 2024-10-07
wasm: increase stack size

tinyjin:
Patch #2826 increased the stack size to 4MB


tinyjin:
@SergeyLebedkin You might want to increase larger size, if you have further plans to use more resources for optimization?

SergeyLebedkin:
> @SergeyLebedkin You might want to increase larger size, if you have further plans to use more resources for optimization?

Currently this size covers all of our existing exnimations and SVGs, so I dont think that we will increase the buffer size

SergeyLebedkin (approved):

PR #2839, opened 2024-10-08
sw_engine: split long lines to avoid overflow

mgrudzinska:

### DIFF
--- a/src/renderer/sw_engine/tvgSwRle.cpp
+++ b/src/renderer/sw_engine/tvgSwRle.cpp
@@ -235,6 +235,7 @@ struct RleWorker
     SwPoint pos;
 
     SwPoint bezStack[32 * 3 + 1];
+    SwPoint lineStack[32 + 1];
     int levStack[32];
 
     SwOutline* outline;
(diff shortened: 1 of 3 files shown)


mgrudzinska:
@hermet I tried to test performance, but the differences between runs are too big to find a difference between this pr and main. can you please double check?
also there is not so many changes - diff looks bad, but I didn't make changes in the inside logic of _lineTo, I just applied the approach that we already have in _cubicTo - if line to too big, split.

hermet:
> the differences between runs are too big to find a difference between this pr and main.

@mgrudzinska Could you please suggest the Lottie FPS? The value represents the average based on accumulation. You will see a stable value once the sampling exceeds a certain basis (i.e 1000)

image

mgrudzinska:
> > the differences between runs are too big to find a difference between this pr and main.
> 
> @mgrudzinska Could you please suggest the Lottie FPS? The value represents the average based on accumulation. You will see a stable value once the sampling exceeds a certain basis (i.e 1000)
> 
> image

I have 17/18 on main and on this branch as well. so I can not see any changes

src/renderer/sw_engine/tvgSwRle.cpp
```diff
@@ -513,103 +514,116 @@ static void _lineTo(RleWorker& rw, const SwPoint& to)
...
+                do {
+                    f2.y = 0;
+                    rw.cover += (f2.y - f1.y);
+                    rw.area += (f2.y - f1.y) * f1.x * 2;
+                    f1.y = ONE_PIXEL;
+                    --e1.y;
+                    _setCell(rw, e1);
+                } while(e1.y != e2.y);
+            }
+            //any other line
```
hermet, line 565: indentation

src/renderer/sw_engine/tvgSwRle.cpp
```diff
@@ -513,103 +514,116 @@ static void _lineTo(RleWorker& rw, const SwPoint& to)
...
+            if (diff.y > 0) {
+                do {
+                    f2.y = ONE_PIXEL;
+                    rw.cover += (f2.y - f1.y);
+                    rw.area += (f2.y - f1.y) * f1.x * 2;
+                    f1.y = 0;
+                    ++e1.y;
+                    _setCell(rw, e1);
+                } while(e1.y != e2.y);
+                //vertical line down
```
hermet, line 554: indentation

src/renderer/sw_engine/tvgSwRle.cpp
```diff
@@ -513,103 +514,116 @@ static void _lineTo(RleWorker& rw, const SwPoint& to)
...
+        e1 = TRUNC(line[1]);
+        e2 = TRUNC(line[0]);
+
+        auto f1 = line[1] - SUBPIXELS(e1);
+        SwPoint f2;
+
+        //inside one cell
+        if (e1 == e2) {
+            ;
+            //any horizontal line
```
hermet, line 539: indentation

hermet:
@tinyjin Could you please confirm that the crash is resolved in the ThorVG viewer with this PR?
sample: test

hermet (approved):

src/renderer/sw_engine/tvgSwRle.cpp
mgrudzinska: fix this and all other

tinyjin:
@hermet 
Yes, that sample has no crash

CleanShot 2024-10-10 at 14 20 26@2x

tinyjin (approved):

Issue #2721, opened 2024-09-09
Suggestion: Remove TVG Binary feature


hermet:
As we approach the 1.0 release, it is better to remove features that are not useful. In practice, the use cases for the TVG binary format have not been sufficient, and support for the TVG binary feature has been lacking as well. Instead of adding this feature in the 1.0 release, it would be better to aim for designing a better format for version 2.0 if the opportunity arises. The removal to the main branch will be applied after v0.15 release.

Relevant task: https://github.com/thorvg/thorvg/issues/1372

hermet:
Please feel free to leave any opinions if you really expect to use this feature.

hermet:
See as well:
- https://github.com/thorvg/thorvg/issues/2527
- https://github.com/thorvg/thorvg/issues/2506
- https://github.com/thorvg/thorvg/issues/2455
- https://github.com/thorvg/thorvg/issues/2348
- https://github.com/thorvg/thorvg/issues/1284

hermet:
@mgrudzinska Let's remove it.

hermet:
It's on my end.

PR #2849, opened 2024-10-12
spec out tvg binary format

hermet:
issue: https://github.com/thorvg/thorvg/issues/2721


hermet:
@tinyjin the viewer/web also needs to be updated.

PR #2853, opened 2024-10-13
wg_engine: MSAA support, part 1 - move blending to fragment shaders

SergeyLebedkin:
Apply custom blending using fragment shaders instead of compute shaders.

This reduces the number of context switches (from 4 to 2) when using a custom blend
Also removes artifacts on surface boundaries when using analiasing (will be added in part 2)

image


SergeyLebedkin:
closed due to multiple merge-conflicts. will be recreated

Issue #2831, opened 2024-10-07
API: promote the composite() method to Scene from Paint


hermet:
The composite() function is too general-purpose in thorvg now, making it somewhat over-engineered for practical use. It might not be a bad idea to restrict composite() to a Scene-level scope and optimize the data structure and resources for better efficiency.

`Paint::composite()` -> `Scene::composite()`

PR #1130, opened 2022-01-02
svg_loader: mask-type attribute introduced 

mgrudzinska:
In an svg file the mask-type attribute can be specified.
It takes one of the two values: Luminosity or Alpha.
After the LumaMask is introduced into TVG, this attribute can be
properly read by the svg loader.

Please be aware that this pr contains 2 commits - I'll change it after #1128 is merged. 

before:
03_before

after:
03_after

code:
- test the attribute:
```
<svg viewBox="0 0 200 200">
<defs>
<mask id="ml1" mask-type="Luminance">
<rect x="10" y="10" width="80" height="80" fill="red" fill-opacity="0.7"/>
</mask>
<mask id="ml2" mask-type="Luminance">
<rect x="110" y="10" width="80" height="80" fill="blue" fill-opacity="0.5"/>
</mask>
<mask id="ma1" mask-type="Alpha">
<rect x="10" y="110" width="80" height="80" fill="green" fill-opacity="0.7"/>
</mask>
<mask id="ma2" mask-type="Alpha">
<rect x="110" y="110" width="80" height="80" fill="blue" fill-opacity="0.5"/>
</mask>
</defs>
<circle cx="50" cy="50" r="50" mask="url(#ml1)"/>
<circle cx="150" cy="50" r="50" mask="url(#ml2)"/>
<circle cx="50" cy="150" r="50" mask="url(#ma1)"/>
<circle cx="150" cy="150" r="50" mask="url(#ma2)"/>
</svg>
```
- test the attribute used in a style attr
```
<svg viewBox="0 0 200 200">
<defs>
<mask id="ml1" style="mask-type:Luminance">
<rect x="10" y="10" width="80" height="80" fill="red" fill-opacity="0.7"/>
</mask>
<mask id="ml2" style="mask-type:Luminance">
<rect x="110" y="10" width="80" height="80" fill="blue" fill-opacity="0.5"/>
</mask>
<mask id="ma1" style="mask-type:Alpha">
<rect x="10" y="110" width="80" height="80" fill="green" fill-opacity="0.7"/>
</mask>
<mask id="ma2" style="mask-type:Alpha">
<rect x="110" y="110" width="80" height="80" fill="blue" fill-opacity="0.5"/>
</mask>
</defs>
<circle cx="50" cy="50" r="50" mask="url(#ml1)"/>
<circle cx="150" cy="50" r="50" mask="url(#ml2)"/>
<circle cx="50" cy="150" r="50" mask="url(#ma1)"/>
[...]

PR #2858, opened 2024-10-14
svg: wrapping a paint with a scene

hermet:
Apply scene when a clipper or masking was required.

This is a prerequisite task for migrating
the clip()/mask() features from Paint to Scene.

issue: https://github.com/thorvg/thorvg/issues/2831

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -223,78 +223,80 @@ static Matrix _compositionTransform(Paint* paint, const SvgNode* node, const Svg
 }
 
 
-static void _applyComposition(SvgLoaderData& loaderData, Paint* paint, const SvgNode* node, const Box& vBox, const string& svgPath)
+static Paint* _applyComposition(SvgLoaderData& loaderData, Paint* paint, const SvgNode* node, const Box& vBox, const string& svgPath)
 {
-    /* ClipPath */
     /* Do not drop in Circular Dependency for ClipPath.
        Composition can be applied recursively if its children nodes have composition target to this one. */
-    if (node->style->clipPath.applying) {
-        TVGLOG("SVG", "Multiple Composition Tried! Check out Circular dependency?");
-    } else {
-        auto compNode = node->style->clipPath.node;
-        if (compNode && compNode->child.count > 0) {
-            node->style->clipPath.applying = true;
+    if (node->style->clipPath.applying || node->style->mask.applying) {
+        TVGLOG("SVG", "Multiple composition tried! Check out circular dependency?");
+        return paint;
+    }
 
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -844,34 +834,22 @@ static unique_ptr<Scene> _sceneBuildHelper(SvgLoaderData& loaderData, const SvgN
...
                 } else if ((*child)->type == SvgNodeType::Text) {
-                    auto text = _textBuildHelper(loaderData, *child, vBox, svgPath);
-                    if (text) {
+                    if (auto text = _textBuildHelper(loaderData, *child, vBox, svgPath)) {
                         if ((*child)->id) text->id = djb2Encode((*child)->id);
                         scene->push(std::move(text));
                     }
                 } else if ((*child)->type != SvgNodeType::Mask) {
-                    auto shape = _shapeBuildHelper(loaderData, *child, vBox, svgPath);
-                    if (shape) {
```
hermet, line 866: @JSUYA @mgrudzinska Please confirm this logic.

JSUYA: I remember it was for performance reasons.
(`isMaskWhite` comes from this PR. https://github.com/thorvg/thorvg/pull/1124)
I think it's okay to remove it if it's an unnecessary condition. (I didn't check the performance.)
I confirmed that it works normally when I tested the sample of https://github.com/thorvg/thorvg/pull/1130.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -392,14 +390,6 @@ static void _applyProperty(SvgLoaderData& loaderData, SvgNode* node, Shape* vg,
 }
 
 
-static unique_ptr<Shape> _shapeBuildHelper(SvgLoaderData& loaderData, SvgNode* node, const Box& vBox, const string& svgPath)
-{
-    auto shape = Shape::gen();
-    if (_appendShape(loaderData, node, shape.get(), vBox, svgPath)) return shape;
-    else return nullptr;
-}
```
JSUYA, line 400: shapeAppend() -> shapeBuildHelper() doesn't have any special functions. That's why this refactoring looks efficient, but xxxBuildHelper() and append*Shape() have different internal structures. 

The reason why the append prefix was used is probably because shape->append*() APIs were used with _recognizeShape() (before the [Extract-Method](https://refactoring.guru/ko/extract-method) was applied).

Since append*Shape() became meaningless, I think _appendClipShape(), _appendClipUseNode() should also be _clipShapeBuildHelper(), _clipUseNodeBuildHelper().

JSUYA (approved):

mgrudzinska (approved):

PR #2864, opened 2024-10-15
wg_engine: fix miter gemotry generation in case of collinear segments

SergeyLebedkin:
simple check collinear vectors by cross product.
it prevent devision by zero in straight strokes

see svg2009.svg

### DIFF
--- a/src/renderer/wg_engine/tvgWgGeometry.h
+++ b/src/renderer/wg_engine/tvgWgGeometry.h
@@ -34,6 +34,7 @@ static inline float length2(const Point& a) { return a.x*a.x+a.y*a.y; };
 static inline float distance2(const Point& a, const Point& b) { return length2(a - b); };
 static inline float distance(const Point& a, const Point& b) { return length(a - b); };
 static inline float dot(const Point& a, const Point& b) { return a.x*b.x + a.y*b.y; };
+static inline float cross(const Point& a, const Point& b) { return a.x*b.y - a.y*b.x; };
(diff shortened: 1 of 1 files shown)


src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -329,15 +330,14 @@ struct WgVertexBufferInd {
     void appendMiter(const Point& v0, const Point& v1, const Point& v2, float dist1, float dist2, float halfWidth, float miterLimit) {
         Point sub1 = v1 - v0;
         Point sub2 = v2 - v1;
+        if (tvg::zero(cross(sub1, sub2))) return; // collinear
```
mgrudzinska, line 333: please check:
```
        shape->moveTo(100, 100);
        shape->lineTo(400,100);
        shape->lineTo(100,100.1);
        shape->strokeFill(255, 0, 0);
        shape->strokeWidth(10);
        shape->strokeJoin(tvg::StrokeJoin::Miter);
```
not exactly collinear (even when I check normalized vectors: `cross(nrm1, nrm2)`) 
but sin and cos are zero

SergeyLebedkin: @mgrudzinska I think, that this solusion is nice, but it is not precise enough. Lets keep existing solution and close the PR

SergeyLebedkin:
Let's leave the current solution. Closed due to insufficient accuracy.

PR #2871, opened 2024-10-17
example: include mask extension in personal_character.json

mgrudzinska:
test after #2875 and  #2872 are merged


hermet:
@mgrudzinska Thanks, could you please adjust the effect to enlarge the circle a bit more to show the person's entire head?

mgrudzinska:
@hermet I didn't cut off his head :) but to see the correct result you have to apply first #2875 (which is more a fix than a refactor)

left - masked, right - org
Zrzut ekranu 2024-10-18 o 11 47 47

but now I enlarged it even more:
Zrzut ekranu 2024-10-18 o 11 57 03

hermet (approved):

PR #2874, opened 2024-10-17
wg_engine: multisampling support part 2

SergeyLebedkin:
Full multisampling support including custom blend and compositions.

Must be verified on web and 4K resolution for performance issues

before/after
Blend
BlendMS
before/after
Threads
ThreadsMS

### DIFF
--- a/src/renderer/wg_engine/tvgWgCommon.cpp
+++ b/src/renderer/wg_engine/tvgWgCommon.cpp
@@ -111,18 +111,18 @@ WGPUTexture WgContext::createTexture(uint32_t width, uint32_t height, WGPUTextur
 }
 
 
-WGPUTexture WgContext::createTexStorage(uint32_t width, uint32_t height, WGPUTextureFormat format, uint32_t sc)
+WGPUTexture WgContext::createTexStorage(uint32_t width, uint32_t height, WGPUTextureFormat format)
 {
     const WGPUTextureDescriptor textureDesc {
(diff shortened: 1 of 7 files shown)


tinyjin:
@SergeyLebedkin 
**[Before]**
CleanShot 2024-10-18 at 11 46 53

**[After]**
CleanShot 2024-10-18 at 11 49 11

SergeyLebedkin:
In browser environment:

After/Before
image

SergeyLebedkin:
@tinyjin 
Can you plese, also, share your performance tests?

tinyjin:
@SergeyLebedkin Hello I checked again the quality and performance.

In the animation quality perspective, I don't see big changes between main and this. (I also tested without DPR, looked same)

Performance side, `threads.json` animation showed consistent result(FPS, Memory Usage), however `frog_vr` with this patch has difference on FPS.

**[Before]**

120FPS, 35MB

CleanShot 2024-10-22 at 17 33 27@2x

**[After]**

44-50FPS, 36MB

CleanShot 2024-10-22 at 17 36 58@2x

SergeyLebedkin:
@tinyjin 

Are you sure that screenshots are correct? I thing they are swaped.

[After] 44-50FPS, 36MB - must be with MSAA
[Before] 120FPS, 35MB - must be without MSAA

Quality in "After" is much better that on screenshot "Before"

But in any case, we can see peformance drop with MSAA.

PR #2875, opened 2024-10-17
lottie: offset does not depend on shape direction

mgrudzinska:
According to tests in AE, the offset direction should not depend on the direction of the shape. A positive offset value expands the shape, while a negative value contracts it. Fixed.

note: 
Comparing the results with Lottie-web was misleading.

### DIFF
--- a/src/loaders/lottie/tvgLottieModifier.cpp
+++ b/src/loaders/lottie/tvgLottieModifier.cpp
@@ -79,7 +79,20 @@ static bool _intersect(const Line& line1, const Line& line2, Point& intersection
 static Line _offset(const Point& p1, const Point& p2, float offset)
 {
     auto scaledNormal = normal(p1, p2) * offset;
-    return {p1 - scaledNormal, p2 - scaledNormal};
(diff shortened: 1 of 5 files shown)


src/loaders/lottie/tvgLottieModifier.cpp
```diff
@@ -79,7 +79,18 @@ static bool _intersect(const Line& line1, const Line& line2, Point& intersection
...
-    return {p1 - scaledNormal, p2 - scaledNormal};
+    return {p1 + scaledNormal, p2 + scaledNormal};
+}
+
+
+static bool _clockwise(const Point* pts, uint32_t n)
+{
+    auto area = 0.0f;
+
+    for (auto i = 0; i < n - 1; i++) area += pts[i].x * pts[i + 1].y - pts[i + 1].x * pts[i].y;
```
hermet, line 90: Suggestion:
```
for (...) {
 ........
}
```

Let's add a cross product in `tvgMath.h`.
```
area += cross(pt1, pt2);
```

hermet (approved):

Issue #2863, opened 2024-10-15
PNG image color differs for GL and Software renderers


alex-dorokhov:
I'm using the PicturePng example with my own image file (
![img_0_gimp](https://github.com/user-attachments/assets/dc01a8c0-8eb7-4895-949e-9fd85feed1a3)
).

The result on Mac OS is different depending on the renderer (GL - left, SW - right):

<img width="1601" alt="Screenshot 2024-10-15 at 17 04 41" src="https://github.com/user-attachments/assets/4d0946ce-3b18-44c6-b9e5-ae277e89ce64">

As you may see, the left (GL) picture is more pinky.

I'm using commit with the hash c9960fcc

hermet:
@alex-dorokhov the fix will be in 1.0.0-pre3. Thanks.

hermet:
@alex-dorokhov if you need an official release (v0.15.2), please let me know. thanks.

PR #2880, opened 2024-10-18
gl_engine: correct the ColorSpace when render Image

RuiwenTang:
Need handle ColorSpace correct when prepare Image texture. If premultiplied is not marked. Needs to change the ColorSpace.

related issue: #2863

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -1194,6 +1194,10 @@ RenderData GlRenderer::prepare(RenderSurface* image, RenderData data, const Matr
         sdata->texId = _genTexture(image);
         sdata->opacity = opacity;
         sdata->texColorSpace = image->cs;
+        if (!image->premultiplied) {
+            if (image->cs == ColorSpace::ABGR8888) sdata->texColorSpace = ColorSpace::ABGR8888S;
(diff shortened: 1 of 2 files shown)


src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1281,6 +1281,10 @@ RenderData GlRenderer::prepare(RenderSurface* image, RenderData data, const Matr
         sdata->texId = _genTexture(image);
         sdata->opacity = opacity;
         sdata->texColorSpace = image->cs;
+        if (!image->premultiplied) {
+            if (image->cs == ColorSpace::ABGR8888) sdata->texColorSpace = ColorSpace::ABGR8888S;
```
hermet, line 1285: Hmm... I think the image loader might need to assign the proper colorspace value.

hermet:
Quick test, and a bit confusing since it looks fine to me on linux.

image

RuiwenTang:
> Quick test, and a bit confusing since it looks fine to me on linux.
> 
> image

I can only reproduce this issue on My old laptop. Is it caused by different implementations of the png library?

hermet:
@RuiwenTang could you please confirm that fix the issue? https://github.com/thorvg/thorvg/pull/2883

Issue #2860, opened 2024-10-15
api: support RadialGradient focal parameters


hermet:
RadialGradient needs an API revision to expose focal parameters.

Result RadialGradient::RadialGradient::radial(float cx, float cy, float r)
 -> Result RadialGradient::RadialGradient::radial(float cx, float cy, float r, float fx, float fy, float fr)

mgrudzinska:
@hermet do we want to keep both or only the new one? 
I can not set any def values to fx and fy since if not specified they should be equal to cx and cy

PR #2886, opened 2024-10-19
api: add focal parameters to the radial apis

mgrudzinska:
@Issue: https://github.com/thorvg/thorvg/issues/2860

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -767,31 +767,43 @@ public:
     ~RadialGradient();
 
     /**
-     * @brief Sets the radial gradient bounds.
+     * @brief Sets the radial gradient attributes.
      *
-     * The radial gradient bounds are defined as a circle centered in a given point (@p cx, @p cy) of a given radius.
+     * The radial gradient is defined by the end circle with a center (@p cx, @p cy) and a radius @p r and
+     * the start circle with a center/focal point (@p fx, @p fy) and a radius @p fr.
+     * The gradient will be rendered such that the gradient stop at an offset of 100% aligns with the edge of the end circle
+     * and the stop at an offset of 0% aligns with the edge of the start circle.
      *
-     * @param[in] cx The horizontal coordinate of the center of the bounding circle.
-     * @param[in] cy The vertical coordinate of the center of the bounding circle.
-     * @param[in] radius The radius of the bounding circle.
+     * @param[in] cx The horizontal coordinate of the center of the end circle.
+     * @param[in] cy The vertical coordinate of the center of the end circle.
+     * @param[in] r The radius of the end circle.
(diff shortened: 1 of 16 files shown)


inc/thorvg.h
```diff
@@ -767,31 +767,40 @@ class TVG_API RadialGradient final : public Fill
...
+     * @param[in] fr The radius of the start circle.
      *
-     * @retval Result::InvalidArguments in case the @p radius value is zero or less.
+     * @retval Result::InvalidArguments in case the radius @p r or @p fr value is negative.
      *
-     * @note In case the @p radius is zero, an object is filled with a single color using the last color specified in the colorStops().
+     * @note In case the radius @p r is zero, an object is filled with a single color using the last color specified in the colorStops().
+     * @note If you choose not to specify a focal point, ensure that @p fx and @p fy are the same as @p cx and @p cy, and set @p fr to 0.
      */
-    Result radial(float cx, float cy, float radius) noexcept;
```
hermet, line 782: We can likely specify the usage of the focal:
```
By adjusting the position and size of the focal point, you can create various visual effects, such as focusing the gradient closer to one edge or generating more complex shading patterns. If you don't need a focal effect, set the focal point values to be the same as the center with a zero radius.
```

test/testFill.cpp
```diff
@@ -151,24 +151,31 @@ TEST_CASE("Radial Filling", "[tvgFill]")
...
-    REQUIRE(fill->radial(0, 0, -1) == Result::InvalidArguments);
-    REQUIRE(fill->radial(nullptr, nullptr, nullptr) == Result::Success);
-    REQUIRE(fill->radial(100, 120, 50) == Result::Success);
+    REQUIRE(fill->radial(0, 0, -1, 0, 0, 0) == Result::InvalidArguments);
+    REQUIRE(fill->radial(0, 0, 0, 0, 0, -1) == Result::InvalidArguments);
+    REQUIRE(fill->radial(nullptr, nullptr, nullptr, nullptr, nullptr, nullptr) == Result::Success);
+    REQUIRE(fill->radial(100, 120, 50, 10, 20, 5) == Result::Success);
 
-    REQUIRE(fill->radial(&cx, nullptr, &radius) == Result::Success);
+    REQUIRE(fill->radial(&cx, nullptr, &r, nullptr, nullptr, nullptr) == Result::Success);
```
hermet, line 161: let's remove the nullptr passes for focal.

mgrudzinska: please check now, removed also in l.158

inc/thorvg.h
mgrudzinska: please check now, changed in thorvg.h and capi version as well.

PR #2891, opened 2024-10-21
lottie: fix rotation in text range selector

mgrudzinska:
Since the translate API was used while text updating, the subsequent range selector transformations gets overwritten when updating the shape (scale and rotate, adding another translation will persist). This caused unexpected results. Fixed by using the transform API when additional transformations are needed.

before:
Zrzut ekranu 2024-10-21 o 22 06 39

after:
Zrzut ekranu 2024-10-21 o 22 04 40

sample:
[textRS.json](https://github.com/user-attachments/files/17463027/textRS.json)

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -1060,41 +1060,48 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
                     shape->strokeFill(doc.stroke.color.rgb[0], doc.stroke.color.rgb[1], doc.stroke.color.rgb[2]);
                 }
 
-                //text range process
-                for (auto s = text->ranges.begin(); s < text->ranges.end(); ++s) {
-                    float start, end;
-                    (*s)->range(frameNo, float(totalChars), start, end);
+                if (!text->ranges.empty()) {
+                    Point scaling = {1.0f, 1.0f};
+                    auto rotation = 0.0f;
+                    auto translation = cursor;
 
-                    auto basedIdx = idx;
-                    if ((*s)->based == LottieTextRange::Based::CharsExcludingSpaces) basedIdx = idx - space;
-                    else if ((*s)->based == LottieTextRange::Based::Words) basedIdx = line + space;
-                    else if ((*s)->based == LottieTextRange::Based::Lines) basedIdx = line;
+                    //text range process
+                    for (auto s = text->ranges.begin(); s < text->ranges.end(); ++s) {
+                        float start, end;
+                        (*s)->range(frameNo, float(totalChars), start, end);
 
-                    if (basedIdx < start || basedIdx >= end) continue;
-                    auto& matrix = shape->transform();
+                        auto basedIdx = idx;
+                        if ((*s)->based == LottieTextRange::Based::CharsExcludingSpaces) basedIdx = idx - space;
+                        else if ((*s)->based == LottieTextRange::Based::Words) basedIdx = line + space;
+                        else if ((*s)->based == LottieTextRange::Based::Lines) basedIdx = line;
 
-                    shape->opacity((*s)->style.opacity(frameNo));
+                        if (basedIdx < start || basedIdx >= end) continue;
 
-                    auto color = (*s)->style.fillColor(frameNo);
-                    shape->fill(color.rgb[0], color.rgb[1], color.rgb[2], (*s)->style.fillOpacity(frameNo));
+                        translation = translation + (*s)->style.position(frameNo);
+                        scaling = scaling * (*s)->style.scale(frameNo) * 0.01f;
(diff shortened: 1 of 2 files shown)


tinyjin (approved):

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1071,21 +1071,26 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
...
 
                     if (basedIdx < start || basedIdx >= end) continue;
-                    auto& matrix = shape->transform();
 
                     shape->opacity((*s)->style.opacity(frameNo));
 
                     auto color = (*s)->style.fillColor(frameNo);
                     shape->fill(color.rgb[0], color.rgb[1], color.rgb[2], (*s)->style.fillOpacity(frameNo));
 
+                    Matrix matrix;
```
hermet, line 1080: Please use this:
```
auto& matrix = shape->transform();
```

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1071,21 +1071,26 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
...
+                    identity(&matrix);
```
hermet, line 1081: The shape's transform will be reset on every iteration. This means that during the text range iteration process (line 1064), any previously applied transform to the shape might be overwritten, results in a meaningless job. It seems the text-range logic is getting messed up. Could you please double-check this?

hermet (changes requested):

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: yes, this is intentional. if we don;t overwrite previously applied transformations, this is what is happening:
 
rotBefore

this should be a static image, but each time we update text, one extra transformation is added to the previous one. 

I could use shape->rotate/scale/translate so that I don;t need to create matrix object, but than we need to update scale api to expect x and y scaling factors, not only one value. this will overwrite the transform matrix internally.

hermet: This is what I see the code.
```
shape = gen();
for(; ; ;) {
  matrix m;
  shape->transform(m);   //useless effect during iteration.
}
scene->push(shape);
```

mgrudzinska: @hermet please check now. you were right, that combining multiple transformations while iterating over ranges was not handled properly. 
as I tested in AE, all parameters (like rotation, scaling and translation) are cumulated first, and then applied. 

sample: 
[multiRanges.json](https://github.com/user-attachments/files/17491121/multiRanges.json)

result (rotate anchor point is not perfect now - that's why slight differences are still visible)
Zrzut ekranu 2024-10-23 o 19 13 16

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1060,41 +1060,48 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
                     shape->strokeFill(doc.stroke.color.rgb[0], doc.stroke.color.rgb[1], doc.stroke.color.rgb[2]);
                 }
 
-                //text range process
-                for (auto s = text->ranges.begin(); s < text->ranges.end(); ++s) {
-                    float start, end;
-                    (*s)->range(frameNo, float(totalChars), start, end);
+                if (!text->ranges.empty()) {
```
hermet, line 1063:
```
//default (see: 1053 ~ 1056)
if (text->ranges.empty()) {
   shape->fill(doc.color.rgb[0], doc.color.rgb[1], doc.color.rgb[2]);
   shape->translate(cursor.x, cursor.y);
   shape->opacity(255);
//override with the text range selector
} else {
   ..
}
```

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1060,41 +1060,48 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
...
-                        shape->strokeWidth((*s)->style.strokeWidth(frameNo) / scale);
-                        shape->strokeFill(strokeColor.rgb[0], strokeColor.rgb[1], strokeColor.rgb[2], (*s)->style.strokeOpacity(frameNo));
+                        auto spacing = (*s)->style.lineSpacing(frameNo);
+                        if (spacing > lineSpacing) lineSpacing = spacing;
                     }
-                    cursor.x += (*s)->style.letterSpacing(frameNo);
-
-                    auto spacing = (*s)->style.lineSpacing(frameNo);
-                    if (spacing > lineSpacing) lineSpacing = spacing;
+                    Matrix matrix;
```
hermet, line 1099: auto& matrix = shape->transform();

hermet (approved):

Issue #1367, opened 2023-04-16
svg: support Filter effect


hermet:
2229 - feColorMatrix filter
2684 - feGaussianBlur filter

[Example.zip](https://github.com/thorvg/thorvg/files/11241202/Example.zip)

mgrudzinska:
#374 #951 are also about filters

hermet:
@JSUYA @mgrudzinska can anyone help for feGaussianBlur filter support? 
See example: https://github.com/thorvg/thorvg/blob/main/examples/SceneEffect.cpp

mgrudzinska:
@hermet I think [this one](https://github.com/thorvg/thorvg/pull/2858) would be useful for me, since filters can be applied in svg to shapes so sometimes I need to build a scene

```
<svg width="230" height="120" xmlns="http://www.w3.org/2000/svg">
  <filter id="gauss">
    <feGaussianBlur stdDeviation="5" />
  </filter>
  <circle cx="170" cy="60" r="50" fill="blue" filter="url(#gauss)" />
</svg>
```

hermet:
@mgrudzinska changed to the review status, please confirm it, thanks.

PR #2893, opened 2024-10-21
svg: added filter node support

mgrudzinska:
Zrzut ekranu 2024-10-22 o 00 45 24

```
<svg width="230" height="120" xmlns="http://www.w3.org/2000/svg">
  <filter id="blurMe">
    <feGaussianBlur stdDeviation="5"/>
  </filter>

  <circle cx="60" cy="60" r="50" fill="green" />

  <g filter="url(#blurMe)">
    <circle cx="170" cy="60" r="50" fill="green"  />
  </g>
</svg>
```
issue: https://github.com/thorvg/thorvg/issues/1367

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1088,6 +1088,17 @@ static void _handleMaskAttr(TVG_UNUSED SvgLoaderData* loader, SvgNode* node, con
 }
 
 
+static void _handleFilterAttr(TVG_UNUSED SvgLoaderData* loader, SvgNode* node, const char* value)
+{
+    SvgStyleProperty* style = node->style;
+    int len = strlen(value);
+    if (len >= 3 && !strncmp(value, "url", 3)) {
+        if (style->filter.url) free(style->filter.url);
+        style->filter.url = _idFromUrl((const char*)(value + 3));
+    }
+}
+
+
 static void _handleMaskTypeAttr(TVG_UNUSED SvgLoaderData* loader, SvgNode* node, const char* value)
 {
     node->node.mask.type = _toMaskType(value);
@@ -1166,7 +1177,8 @@ static constexpr struct
     STYLE_DEF(mask, Mask, SvgStyleFlags::Mask),
     STYLE_DEF(mask-type, MaskType, SvgStyleFlags::MaskType),
     STYLE_DEF(display, Display, SvgStyleFlags::Display),
-    STYLE_DEF(paint-order, PaintOrder, SvgStyleFlags::PaintOrder)
+    STYLE_DEF(paint-order, PaintOrder, SvgStyleFlags::PaintOrder),
+    STYLE_DEF(filter, Filter, SvgStyleFlags::Filter)
 };
 
 
@@ -1240,6 +1252,8 @@ static bool _attrParseGNode(void* data, const char* key, const char* value)
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
         _handleMaskAttr(loader, node, value);
+    } else if (!strcmp(key, "filter")) {
+        _handleFilterAttr(loader, node, value);
     } else {
         return _parseStyleAttr(loader, key, value, false);
     }
@@ -1339,6 +1353,108 @@ static bool _attrParseSymbolNode(void* data, const char* key, const char* value)
         return _attrParseGNode(data, key, value);
     }
 
+    return true;
+}
+
+
+static constexpr struct
+{
+    const char* tag;
+    SvgParserLengthType type;
+    int sz;
+    size_t offset;
+} boxTags[] = {
+    {"x", SvgParserLengthType::Horizontal, sizeof("x"), offsetof(Box, x)},
+    {"y", SvgParserLengthType::Vertical, sizeof("y"), offsetof(Box, y)},
+    {"width", SvgParserLengthType::Horizontal, sizeof("width"), offsetof(Box, w)},
+    {"height", SvgParserLengthType::Vertical, sizeof("height"), offsetof(Box, h)}
+};
+
+
+static bool _parseBox(const char* key, const char* value, Box* box, bool (&isPercentage)[4])
+{
+    auto array = (unsigned char*)box;
+    int sz = strlen(key);
+    for (unsigned int i = 0; i < sizeof(boxTags) / sizeof(boxTags[0]); i++) {
+        if (boxTags[i].sz - 1 == sz && !strncmp(boxTags[i].tag, key, sz)) {
+            *(float*)(array + boxTags[i].offset) = _gradientToFloat(nullptr, value, isPercentage[i]);
+            return true;
+        }
+    }
+    return false;
+}
+
+
+static void _recalcBox(const SvgLoaderData* loader, Box* box, bool (&isPercentage)[4])
+{
+    auto array = (unsigned char*)box;
+    for (unsigned int i = 0; i < sizeof(boxTags) / sizeof(boxTags[0]); i++) {
+        if (!isPercentage[i]) continue;
+        if (boxTags[i].type == SvgParserLengthType::Horizontal) *(float*)(array + boxTags[i].offset) *= loader->svgParse->global.w;
+        else *(float*)(array + boxTags[i].offset) *= loader->svgParse->global.h;
+    }
+}
+
+
+static bool _attrParseFilterNode(void* data, const char* key, const char* value)
+{
+    SvgLoaderData* loader = (SvgLoaderData*)data;
+    SvgNode* node = loader->svgParse->node;
+    SvgFilterNode* filter = &node->node.filter;
+
+    _parseBox(key, value, &filter->box, filter->isPercentage);
+
+    if (!strcmp(key, "id")) {
+        if (node->id && value) free(node->id);
+        node->id = _copyId(value);
+    } else if (!strcmp(key, "primitiveUnits")) {
+        if (!strcmp(value, "objectBoundingBox")) filter->primitiveUserSpace = false;
+    } else if (!strcmp(key, "filterUnits")) {
+        if (!strcmp(value, "userSpaceOnUse")) filter->filterUserSpace = true;
+    }
+
+    return true;
+}
+
+
+static void _parseGaussianBlurStdDeviation(const char** content, float* x, float* y)
+{
+    auto str = *content;
+    char* end = nullptr;
+    float deviation[2] = {0, 0};
+    int n = 0;
+
+    while (*str && n < 2) {
+        str = _skipComma(str);
+        auto parsedValue = strToFloat(str, &end);
+        if (parsedValue < 0.0f) break;
+        deviation[n++] = parsedValue;
+        str = end;
+    }
+
+    *x = deviation[0];
+    *y = n == 1 ? deviation[0] : deviation[1];
+}
+
+
+static bool _attrParseGaussianBlurNode(void* data, const char* key, const char* value)
+{
+    SvgLoaderData* loader = (SvgLoaderData*)data;
+    SvgNode* node = loader->svgParse->node;
+    SvgGaussianBlurNode* gaussianBlur = &node->node.gaussianBlur;
+
+    if (_parseBox(key, value, &gaussianBlur->box, gaussianBlur->isPercentage)) gaussianBlur->hasBox = true;
+
+    if (!strcmp(key, "id")) {
+        if (node->id && value) free(node->id);
+        node->id = _copyId(value);
+    } else if (!strcmp(key, "stdDeviation")) {
+        _parseGaussianBlurStdDeviation(&value, &gaussianBlur->stdDevX, &gaussianBlur->stdDevY);
+    } else if (!strcmp(key, "edgeMode")) {
+        if (!strcmp(value, "wrap")) gaussianBlur->edgeModeWrap = true;
+    } else {
+        return _parseStyleAttr(loader, key, value, false);
+    }
     return true;
 }
 
@@ -1504,6 +1620,48 @@ static SvgNode* _createSymbolNode(SvgLoaderData* loader, SvgNode* parent, const 
 }
 
 
+static SvgNode* _createGaussianBlurNode(SvgLoaderData* loader, SvgNode* parent, const char* buf, unsigned bufLength, parseAttributes func)
+{
+    loader->svgParse->node = _createNode(parent, SvgNodeType::GaussianBlur);
+    if (!loader->svgParse->node) return nullptr;
+    SvgGaussianBlurNode& gaussianBlur = loader->svgParse->node->node.gaussianBlur;
+
+    loader->svgParse->node->style->display = false;
+
+    gaussianBlur.stdDevX = 0.0f;
+    gaussianBlur.stdDevY = 0.0f;
+    gaussianBlur.box = {0.0f, 0.0f, 1.0f, 1.0f};
+    for (auto& p : gaussianBlur.isPercentage) p = false;
+    gaussianBlur.hasBox = false;
+    gaussianBlur.edgeModeWrap = false;
+
+    func(buf, bufLength, _attrParseGaussianBlurNode, loader);
+
+    return loader->svgParse->node;
+}
+
+
+static SvgNode* _createFilterNode(SvgLoaderData* loader, SvgNode* parent, const char* buf, unsigned bufLength, parseAttributes func)
+{
+    loader->svgParse->node = _createNode(parent, SvgNodeType::Filter);
+    if (!loader->svgParse->node) return nullptr;
+    SvgFilterNode& filter = loader->svgParse->node->node.filter;
+
+    loader->svgParse->node->style->display = false;
+
+    filter.box = {-0.1f, -0.1f, 1.2f, 1.2f};
+    for (auto& p : filter.isPercentage) p = false;
+    filter.filterUserSpace = false;
+    filter.primitiveUserSpace = true;
+
+    func(buf, bufLength, _attrParseFilterNode, loader);
+
+    if (filter.filterUserSpace) _recalcBox(loader, &filter.box, filter.isPercentage);
+
+    return loader->svgParse->node;
+}
+
+
 static bool _attrParsePathNode(void* data, const char* key, const char* value)
 {
     SvgLoaderData* loader = (SvgLoaderData*)data;
@@ -1520,6 +1678,8 @@ static bool _attrParsePathNode(void* data, const char* key, const char* value)
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
         _handleMaskAttr(loader, node, value);
+    } else if (!strcmp(key, "filter")) {
+        _handleFilterAttr(loader, node, value);
     } else if (!strcmp(key, "id")) {
         if (value) free(node->id);
         node->id = _copyId(value);
@@ -1582,6 +1742,8 @@ static bool _attrParseCircleNode(void* data, const char* key, const char* value)
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
         _handleMaskAttr(loader, node, value);
+    } else if (!strcmp(key, "filter")) {
+        _handleFilterAttr(loader, node, value);
     } else if (!strcmp(key, "id")) {
         if (value) free(node->id);
         node->id = _copyId(value);
@@ -1649,6 +1811,8 @@ static bool _attrParseEllipseNode(void* data, const char* key, const char* value
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
         _handleMaskAttr(loader, node, value);
+    } else if (!strcmp(key, "filter")) {
+        _handleFilterAttr(loader, node, value);
     } else {
         return _parseStyleAttr(loader, key, value, false);
     }
@@ -1698,6 +1862,8 @@ static bool _attrParsePolygonNode(void* data, const char* key, const char* value
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
         _handleMaskAttr(loader, node, value);
+    } else if (!strcmp(key, "filter")) {
+        _handleFilterAttr(loader, node, value);
     } else if (!strcmp(key, "id")) {
         if (value) free(node->id);
         node->id = _copyId(value);
@@ -1785,6 +1951,8 @@ static bool _attrParseRectNode(void* data, const char* key, const char* value)
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
         _handleMaskAttr(loader, node, value);
+    } else if (!strcmp(key, "filter")) {
+        _handleFilterAttr(loader, node, value);
     } else {
         ret = _parseStyleAttr(loader, key, value, false);
     }
@@ -1850,6 +2018,8 @@ static bool _attrParseLineNode(void* data, const char* key, const char* value)
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
         _handleMaskAttr(loader, node, value);
+    } else if (!strcmp(key, "filter")) {
+        _handleFilterAttr(loader, node, value);
     } else {
         return _parseStyleAttr(loader, key, value, false);
     }
@@ -1923,6 +2093,8 @@ static bool _attrParseImageNode(void* data, const char* key, const char* value)
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
         _handleMaskAttr(loader, node, value);
+    } else if (!strcmp(key, "filter")) {
+        _handleFilterAttr(loader, node, value);
     } else if (!strcmp(key, "transform")) {
         node->transform = _parseTransformationMatrix(value);
     } else {
@@ -2103,6 +2275,8 @@ static bool _attrParseTextNode(void* data, const char* key, const char* value)
         _handleClipPathAttr(loader, node, value);
     } else if (!strcmp(key, "mask")) {
         _handleMaskAttr(loader, node, value);
+    } else if (!strcmp(key, "filter")) {
+        _handleFilterAttr(loader, node, value);
     } else if (!strcmp(key, "id")) {
         if (value) free(node->id);
         node->id = _copyId(value);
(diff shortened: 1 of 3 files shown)


mgrudzinska:
before:
Zrzut ekranu 2024-11-13 o 22 23 04

after:
Zrzut ekranu 2024-11-13 o 22 22 46

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1343,6 +1357,125 @@ static bool _attrParseSymbolNode(void* data, const char* key, const char* value)
...
+
+    PARSE_FILTER_REGION(key, value, loader, SvgFilterNode, filter);
+
+    if (!strcmp(key, "id")) {
+        if (node->id && value) free(node->id);
+        node->id = _copyId(value);
+    } else if (!strcmp(key, "filterUnits")) {
+        if (!strcmp(value, "userSpaceOnUse")) filter->userSpaceFilter = true;
+    } else if (!strcmp(key, "primitiveUnits")) {
+        if (!strcmp(value, "objectBoundingBox")) filter->userSpaceFilter = false;
```
JSUYA, line 1407: Is it okay to use the same variable for filterUnits and [primitiveUnits](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/primitiveUnits)? Both attributes can have the same format of value, but it seems like they are applied to different targets.

If `filterUnits = "userSpaceOnUse", primitiveUnits = "objectBoundingBox"` is set, the user space of the filter is changed to objectBoundingBox.

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -265,6 +268,14 @@ enum class AspectRatioMeetOrSlice
     Slice
 };
 
+enum class GaussianBlurEdgeMode
+{
+    None = 0,
+    Duplicate,
+    Wrap,
+    Mirror
```
JSUYA, line 276: Is there a mirror mode? It's not in the spec.
https://www.w3.org/TR/SVG11/filters.html#feConvolveMatrixElementEdgeModeAttribute

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: Oh I saw that there is a userSpacePrimitive. This looks a typo.
But additionally, it seems that these values ​​are not actually applied.
Wouldn't it be better to not parse it so that it outputs that it is an unsupported attribute?
(Or how about comment it out?)

mgrudzinska: there is few attribs that are not applied, will remove them

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: latest draft: https://drafts.fxtf.org/filter-effects/#element-attrdef-feconvolvematrix-edgemode
but I can remove since we don;t support this in the renderer

mgrudzinska:
@JSUYA updated, only edge mode and sigma are left. the rest I'll add in the next iteration

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -265,6 +268,13 @@ enum class AspectRatioMeetOrSlice
     Slice
 };
 
+enum class GaussianBlurEdgeMode
```
hermet, line 271: ThorVG APIs already provide the necessary functions and guidance. The SVG loader only needs such an `int gaussianBlurEdgeMode` to convert the SVG input to ThorVG.

mgrudzinska:
> and one more note - there is a discrepancy between stdDev in Svg and sigma in the renderer. rg1024_eggs.svg looks good, but gallardo.svg is too much blurred. I'll investigate this

this seems to be also a problem for lottie files, so should be handled as a separate issue probably on the renderer side

hermet:
> > and one more note - there is a discrepancy between stdDev in Svg and sigma in the renderer. rg1024_eggs.svg looks good, but gallardo.svg is too much blurred. I'll investigate this
> 
> this seems to be also a problem for lottie files, so should be handled as a separate issue probably on the renderer side
> 

You might wanna recalc the sigma based on the view scale factor.

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -383,6 +386,16 @@ struct SvgTextNode
     float fontSize;
 };
 
+struct SvgGaussianBlurNode
+{
+    float stdDevX, stdDevY;
+    bool edgeModeWrap;
+};
+
+struct SvgFilterNode
```
hermet, line 395: ?

mgrudzinska: This node should enclose other filter nodes and can also contain attributes independently of other filters, such as 'x', 'y', 'width', 'height', and 'units...'.

example:
```
  <filter id="dropShadow" x="2" y="14">
    <feGaussianBlur in="SourceAlpha" stdDeviation="3" />
    <feOffset dx="2" dy="4" />
    <feMerge>
      <feMergeNode />
      <feMergeNode in="SourceGraphic" />
    </feMerge>
  </filter>
  ```

Comment (hermet) [reply src/loaders/svg/tvgSvgLoaderCommon.h:395]: What does 'enclose' mean here?  It seems an empty struct and isn't actually used in the logic. If you're adding this for potential future use, please add it only when it's actually needed.

**Comment by hermet** [comment]
@mgrudzinska A crash occurred in the SVG example. This issue needs to be clarified before proceeding, as it appears critical. No issues were detected by sanitizers or gdb; the crash only occurs at runtime. :( 
image

File: src/loaders/svg/tvgSvgSceneBuilder.cpp [hunk 5]
```diff
@@ -283,6 +283,27 @@ static Paint* _applyComposition(SvgLoaderData& loaderData, Paint* paint, const S
...
+    Scene* scene = Scene::gen();
+    scene->push(paint);
+
+    auto child = filter->child.data;
+    for (uint32_t i = 0; i < filter->child.count; ++i, ++child) {
+        if ((*child)->type == SvgNodeType::GaussianBlur) {
+            auto& gauss = (*child)->node.gaussianBlur;
+            auto direction = gauss.stdDevX > 0.0f ? (gauss.stdDevY > 0.0f ? 0 : 1) : (gauss.stdDevY > 0.0f ? 2 : -1);
+            if (direction == -1) continue;
+            scene->push(SceneEffect::GaussianBlur, direction == 2 ? gauss.stdDevY : gauss.stdDevX, direction, gauss.edgeModeWrap);
```
hermet, line 300: @mgrudzinska could you please confirm here multiple gaussian filters are valid input?
image

hermet: please pass the quality value as well. The crash is due to it.
`scene->push(SceneEffect::GaussianBlur, direction == 2 ? gauss.stdDevY : gauss.stdDevX, direction, gauss.edgeModeWrap, 100);`

hermet:
a note with `gaussian1.svg`

tvg:
image

expected:
image

hermet (changes requested):
let's prevent crash at least.

mgrudzinska:
```
==48760==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x00012a695554 at pc 0x00010f77a4f8 bp 0x7ff7b1174260 sp 0x7ff7b1174258
WRITE of size 4 at 0x00012a695554 thread T0
    #0 0x10f77a4f7 in rasterXYFlip(unsigned int*, unsigned int*, int, int, int, SwBBox const&, bool) tvgSwRaster.cpp:1845
    #1 0x10f75c359 in effectGaussianBlur(SwCompositor*, SwSurface*, tvg::RenderEffectGaussianBlur const*, bool) tvgSwPostEffect.cpp:205
    #2 0x10f785e7b in tvg::SwRenderer::effect(tvg::RenderCompositor*, tvg::RenderEffect const*, bool) tvgSwRenderer.cpp:661
    #3 0x10f731310 in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:145
    #4 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
    #5 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
    #6 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
    #7 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
    #8 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
    #9 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
    #10 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
    #11 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
    #12 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
    #13 0x10f73add6 in tvg::Picture::Impl::render(tvg::RenderMethod*) tvgPicture.cpp:84
    #14 0x10f730075 in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
    #15 0x10f7256d7 in tvg::Canvas::Impl::draw() tvgCanvas.h:122
    #16 0x10f7252da in tvg::Canvas::draw() tvgCanvas.cpp:61
    #17 0x10ed91b13 in tvgexam::Window::draw() Example.h:173
    #18 0x10ed8e8c6 in tvgexam::Window::show() Example.h:251
    #19 0x10ed8dc29 in tvgexam::main(tvgexam::Example*, int, char**, unsigned int, unsigned int, unsigned int, bool) Example.h:523
    #20 0x10ed8ecc0 in main Svg.cpp:117
    #21 0x7ff8126c5344 in start+0x774 (dyld:x86_64+0xfffffffffff5c344)
```

hermet:
> ==48760==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x00012a695554 at pc 0x00010f77a4f8 bp 0x7ff7b1174260 sp 0x7ff7b1174258
> WRITE of size 4 at 0x00012a695554 thread T0
>     #0 0x10f77a4f7 in rasterXYFlip(unsigned int*, unsigned int*, int, int, int, SwBBox const&, bool) tvgSwRaster.cpp:1845
>     #1 0x10f75c359 in effectGaussianBlur(SwCompositor*, SwSurface*, tvg::RenderEffectGaussianBlur const*, bool) tvgSwPostEffect.cpp:205
>     #2 0x10f785e7b in tvg::SwRenderer::effect(tvg::RenderCompositor*, tvg::RenderEffect const*, bool) tvgSwRenderer.cpp:661
>     #3 0x10f731310 in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:145
>     #4 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
>     #5 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
>     #6 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
>     #7 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
>     #8 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
>     #9 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
>     #10 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
>     #11 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
>     #12 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
>     #13 0x10f73add6 in tvg::Picture::Impl::render(tvg::RenderMethod*) tvgPicture.cpp:84
>     #14 0x10f730075 in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
>     #15 0x10f7256d7 in tvg::Canvas::Impl::draw() tvgCanvas.h:122
>     #16 0x10f7252da in tvg::Canvas::draw() tvgCanvas.cpp:61
>     #17 0x10ed91b13 in tvgexam::Window::draw() Example.h:173
>     #18 0x10ed8e8c6 in tvgexam::Window::show() Example.h:251
>     #19 0x10ed8dc29 in tvgexam::main(tvgexam::Example*, int, char**, unsigned int, unsigned int, unsigned int, bool) Example.h:523
>     #20 0x10ed8ecc0 in main Svg.cpp:117
>     #21 0x7ff8126c5344 in start+0x774 (dyld:x86_64+0xfffffffffff5c344)

@mgrudzinska Let's address this issue separately.

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: > @mgrudzinska could you please confirm here multiple gaussian filters are valid input? image

yes, the filter contains two gaussian filters, as it is in the svg file - so it is as expected. 
for now we do not support 'in' and 'result' attribs, so they are applied to the whole scene.

mgrudzinska:
> a note with `gaussian1.svg`
> 

double checked, the results are exactly the same for svg and using apis. 


[gaussian_part.svg.zip](https://github.com/user-attachments/files/17846953/gaussian_part.svg.zip)

```
        if (!canvas) return false;

        shape->appendRect(0, 0, 1500, 1500);
        shape->fill(255, 255, 255);
        canvas->push(shape);


        shape4->appendRect(200, 100, 50, 50);
        shape4->fill(255, 0, 0);
        scene->push(shape4);
        scene->push(tvg::SceneEffect::GaussianBlur, 64.0f, 0, 0, 1);
        canvas->push(scene);

        scene = tvg::Scene::gen();
        shape4 = tvg::Shape::gen();
        shape4->appendRect(700, 1800, 400, 400);
        shape4->fill(255, 0, 0);
        shape4->transform(tvg::Matrix{.25,0,0,0,.25,0,0, 0, 1});
        scene->push(shape4);
        scene->push(tvg::SceneEffect::GaussianBlur, 64.0f, 0, 0, 1);
        canvas->push(scene);


        shape = tvg::Shape::gen();
        shape->appendRect(200, 100, 50, 50);
        shape->strokeFill(0, 0, 0);
        shape->strokeWidth(2);
        canvas->push(shape);

        shape = tvg::Shape::gen();
        shape->appendRect(700, 1800, 400, 400);
        shape->strokeFill(0, 0, 0);
        shape->strokeWidth(2);
        shape->transform(tvg::Matrix{.25,0,0,0,.25,0,0, 0, 1});
        canvas->push(shape);
        
        return true;
```

results (black borders for sample with apis):

Zrzut ekranu 2024-11-21 o 14 56 29

Zrzut ekranu 2024-11-21 o 14 56 27

and the expected result for svg:
Zrzut ekranu 2024-11-21 o 15 00 30

so it's not this pr, the issue is in the engine

for now I applied factos as they are in the lottie.

hermet:
@mgrudzinsk ThorVG currently does not clip the region automatically unless a clipper is applied, and this behavior is intentional so far because users may expect the blurred region to expand seamlessly. ThorVG could consider providing an option for this, but we may first need to confirm how seamless expansion of the blurred region is handled in SVG.

mgrudzinska:
Notes:
In the current gaussian blur implementation, there are two issues:

1. Clipping
To comply with SVG standards regarding the filter region and filter primitive subregion, clipping must be applied to the blurred image. However, the current implementation has some issues, which is why there are discrepancies in the samples below (the blurring effect are incorrect for some examples, but the sizes and positions should be identical):

[sample1](https://github.com/user-attachments/files/18391739/gauss_units_1.svg.zip)
Zrzut ekranu 2025-01-13 o 00 39 54

[sample1](https://github.com/user-attachments/files/18391741/gauss_units_2.svg.zip)
Zrzut ekranu 2025-01-13 o 00 40 14

[sample1](https://github.com/user-attachments/files/18391742/gauss_units_3.svg.zip)
Zrzut ekranu 2025-01-13 o 00 40 47

[sample1](https://github.com/user-attachments/files/18391743/gauss_units_4.svg.zip)
Zrzut ekranu 2025-01-13 o 00 41 08


2. Rescaling
All svgs were rendered using the Svg.cpp sample with removed scaling. 
Scaling affects the ```stdDev``` value, which is currently handled on the loader side. However, it does not account for potential transformations applied at higher-level parent nodes - marked as TODO.


rg1024_eggs.svg:
Zrzut ekranu 2025-01-13 o 00 32 34

gallardo.svg:
Zrzut ekranu 2025-01-13 o 00 33 26

gaussian1.svg
Zrzut ekranu 2025-01-13 o 01 26 50

- Green rectangle: highlights the most noticeable difference between the expected effect and the obtained result. This discrepancy is due to an issue with clipping on the renderer's side.
- Blue rectangle: highlights the most visible difference in the stdDev value. The scaling factor from the parent node's transformation matrix was significant enough that its omission impacts the intensity of the blur.
An extreme sample showing the same issue: 
[intensity.svg.zip](https://github.com/user-attachments/files/18391769/intensity.svg.zip):
Zrzut ekranu 2025-01-13 o 01 53 29

hermet:
> * Green rectangle: highlights the most noticeable difference between the expected effect and the obtained result. This discrepancy is due to an issue with clipping on the renderer's side.

@mgrudzinska Hello, to clarify this, Are these two rectangles children of a scene which has a blur effect applied?
image

The question here is, because, other shapes look having blur effects applied individually.
image

mgrudzinska:
> @mgrudzinska Hello, to clarify this, Are these two rectangles children of a scene which has a blur effect applied? image
> 
> The question here is, because, other shapes look having blur effects applied individually. image

I'm afraid I don't fully understand the question. Each rectangle has it's own blur effect. the small one has extra clipping. 

samples showing issues with clipping:

hermet:
> Green rectangle: highlights the most noticeable difference between the expected effect and the obtained result. This discrepancy is due to an issue with clipping on the renderer's side.

@mgrudzinska What I see here doesn’t look clipped. I don’t understand how this is a renderer clipping issue.

image

hermet:
> samples showing issues with clipping:

@mgrudzinska Hmm, it looks clipped as expected. I don’t think it’s a rendering issue, as it aligns with the demanded clipping observed so far.

mgrudzinska:
Sorry for confusion. The shape is clipped correctly. What I meant is that there are issues with **post effects**, and that one occurs when clipping is applied to a blurred shape. However, the problem is not with the clipping itself but with the post effects.

hermet:
> What I meant is that there are issues with post effects, and that one occurs when clipping is applied to a blurred shape. However, the problem is not with the clipping itself but with the post effects.

You applied both blur and clipping. The rendering engine blurred the image and clipped it to the specified region as expected. Does SVG ignore any subsequent clippings for post-effects?

mgrudzinska:
> @mgrudzinska You applied both blur and clipping. The rendering engine blurred the image and clipped it to the specified region as expected. Does SVG ignore any subsequent clippings for post-effects? Please clearly describe the expected specifications so that we can figure out the proper solution for this,

1. blurred image without clipping:
```

        shape1->appendRect(100, 100, 600, 600);
        shape1->fill(255, 0, 255, 160);
        scene->push(shape1);
        scene->push(tvg::SceneEffect::GaussianBlur, 450.0f, 0, 0, 10);

        canvas->push(scene);
```

result:
Zrzut ekranu 2025-01-13 o 16 20 34

2. I want to clip this blurred image:
```

        shape1->appendRect(100, 100, 600, 600);
        shape1->fill(255, 0, 255, 160);
        scene->push(shape1);

        auto clipShape = tvg::Shape::gen();
        clipShape->appendRect(100, 100, 100, 100);
        scene->clip(clipShape);

        scene->push(tvg::SceneEffect::GaussianBlur, 50.0f, 0, 0, 10);

        canvas->push(scene);
```

result:

Zrzut ekranu 2025-01-13 o 16 25 10

3. this is what I'd expect:
Zrzut ekranu 2025-01-13 o 16 25 10

I don't see how to achieve this effect in TVG. Am I missing something?

hermet:
> and regarding svg specs: [The filter region acts as a hard clip clipping rectangle on the filter primitive’s input image(s). The filter primitive subregion acts as a hard clip clipping rectangle on the filter primitive result.](https://drafts.fxtf.org/filter-effects/#FilterPrimitiveSubRegion)
> 
> so it is expected to apply clipping after a filtter is applied.

This looks applying clipping after the filter. May I ask what the problem in your sample?

image

hermet:
@JSUYA There seems to be an issue with blur rendering, but if there are no problems from the perspective of SVG implementation, it can likely be addressed as a separate issue.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1504,6 +1620,50 @@ static SvgNode* _createSymbolNode(SvgLoaderData* loader, SvgNode* parent, const
 }
 
 
+static SvgNode* _createGaussianBlurNode(SvgLoaderData* loader, SvgNode* parent, const char* buf, unsigned bufLength, parseAttributes func)
+{
+    loader->svgParse->node = _createNode(parent, SvgNodeType::GaussianBlur);
+    if (!loader->svgParse->node) return nullptr;
+
+    loader->svgParse->node->style->display = false;
```
hermet, line 1628: Here some zero-based values are unnecessarily initialized?

mgrudzinska: We use the same approach for clip nodes as well. In the `_sceneBuildHelper`, we check this value, and if it is `false`, we stop processing the node. For clips, the results will be incorrect if we do not set `display = false`. For filter/gaussian nodes, the results will remain unchanged because during processing, the `_recognizeShape` function will not detect any shapes inside filter/gaussian nodes (as there are none), and node processing will effectively do nothing (no paint will be pushed into the scene). However, there is no need to proceed that far, as we can stop earlier. Therefore, this value is set to `false`.

hermet: @mgrudzinska What I'm saying here is that those values are already initialized with calloc() during creation, and you are setting them zero to initialize again.
image

hermet:
An apparent visual drawback of the unstable blur effect among our samples.

image

JSUYA:
(Sorry for the late review.)

The results is different from Chrome.
```
<svg width="230" height="120" xmlns="http://www.w3.org/2000/svg">
  <filter id="blurMe">
    <feGaussianBlur stdDeviation="5"/>
  </filter>

  <circle cx="60" cy="60" r="50" fill="green" />

  <g filter="url(#blurMe)">
    <circle cx="170" cy="60" r="50" fill="green"  />
  </g>
</svg>
```
(This sample modified stdDeviation to 5 from the  PR message's sample.)
Chrome looks blurrier.

> An apparent visual drawback of the unstable blur effect among our samples. It might not yet be the right stage for us to proceed. It seems like an issue that requires more careful consideration.

I thought this problem occurred as the output image got smaller.
However, according to the SVG specfication, I was able to find this about stdDeviation:
(https://www.w3.org/TR/filter-effects-1/#element-attrdef-fegaussianblur-stddeviation)

```
This filter primitive performs a Gaussian blur on the input image.

The Gaussian blur kernel is an approximation of the normalized convolution:
```
I think, we need additional consideration for stdDeviation.
What do you think?

hermet:
https://github.com/thorvg/thorvg/pull/3137 

Two svgs are still not perfect, Others look good, 

image

hermet:
@mgrudzinska  if this invisible issue is addressed properly, I think we can proceed with pushing this feature.

image

mgrudzinska:
> @mgrudzinska if this invisible issue is addressed properly, I think we can proceed with pushing this feature.
> 
> image

The missing pict is now visible, but I need to test the scaling because I can see from the Gaussian1.svg sample that it's misaligned

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: > @mgrudzinska What I'm saying here is that those values are already initialized with calloc() during creation, and you are setting them zero to initialize again. image

few lines below this all def values are set:
```   
    //Default display is true("inline").
    node->style->display = true;
```

hermet:
> I can see from the Gaussian1.svg sample that it's misaligned

@mgrudzinska Please share what you observed.

mgrudzinska:
> > I can see from the Gaussian1.svg sample that it's misaligned
> 
> @mgrudzinska Please share what you observed. Maybe it’s related to the scale factor of the blur scene itself.

my mistake, sorry
seems that everything is good now:

Zrzut ekranu 2025-01-19 o 20 01 50

Zrzut ekranu 2025-01-19 o 20 08 15
Zrzut ekranu 2025-01-19 o 20 19 49

the result of blur.svg sample without unsupported features:
z

mgrudzinska:
@JSUYA can you please add samples and results you got.
I checked your sample:
Zrzut ekranu 2025-01-19 o 20 03 41
seems to be close enough for me (chrome on mac)

JSUYA:
Hi @mgrudzinska 
I confirmed that the issue was fixed after PR #3137.

before


after

src/loaders/svg/tvgSvgLoader.cpp
hermet: @mgrudzinska please confirm others.

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -26,6 +26,27 @@
 #include "tvgCommon.h"
 #include "tvgArray.h"
 
+struct Box
```
hermet, line 29: Maybe we can reuse this:

`struct RenderRegion  // tvgRender.h(83)`

or maybe with the refactoring `template<typename T = int32_t> Geometry //tvgMath.h`

hermet (approved):
Two minor suggestions, but they're optional for you.

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: after templating the RenderRegion the binary file  increased by 4k

hermet: @mgrudzinska got it, let's forget it.

PR #2898, opened 2024-10-22
lottie: range selector enhancement

mgrudzinska:
contains #2891 

two moving points are the effect of easing
Oct-25-2024 01-23-21

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -32,7 +32,6 @@
 /************************************************************************/
 
 
-
 /************************************************************************/
 /* External Class Implementation                                        */
 /************************************************************************/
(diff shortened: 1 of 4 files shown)


mgrudzinska:
@hermet @tinyjin I'll split this pr into smaller, after we merge 2891

everything here works, but rev won't be nice with so many different changes

before:
Zrzut ekranu 2024-10-25 o 00 16 04

after (translate in text range fixed, effect of easing visible):
Zrzut ekranu 2024-10-25 o 00 16 36

sample:
[easing.json](https://github.com/user-attachments/files/17511215/easing.json)

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -31,7 +31,33 @@
...
 
+    Point in{0.0f, 0.0f};
+    Point out{1.0f, 1.0f};
+
+    if (minEase > 0.0f) in.x = minEase * 0.01f;
+    else in.y = -minEase * 0.01f;
+    if (maxEase > 0.0f) out.x = 1.0 - maxEase * 0.01f;
+    else out.y = 1.0 + minEase * 0.01f;
+
+    interpolator->set("", in, out);
```
mgrudzinska, line 58: caching interpolator doesn't make sense, since there are too many of them. so the only idea I have now, is to introduce inside the interpolator struct another progress(value, in, out) fun, so that the samples[] values are not precalculated, since they will be used only once.

PR #2900, opened 2024-10-23
lottie/text: alignment options support

tinyjin:
- [x] alignment options parser
- [ ] 🚧 rotation transforming based on group alignment 
- [ ] text grouping
  - currently, `chars` only supported (separated grouping)
  - `chars` and `all` seem same (targeting individual chars, need to review)
  - [ ] Word
  - [ ] Line


tinyjin:
Grouping mechanism has been changed, it could bring side effects on text rendering.

I've tested with existing text test files, looks fine.

CleanShot 2024-10-24 at 18 47 11@2x

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1071,21 +1094,43 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
...
-                    rotate(&matrix, (*s)->style.rotation(frameNo));
-
                     auto glyphScale = (*s)->style.scale(frameNo) * 0.01f;
                     tvg::scale(&matrix, glyphScale.x, glyphScale.y);
 
                     auto position = (*s)->style.position(frameNo);
                     translate(&matrix, position.x, position.y);
 
+                    // NOTE: duplicated with Mira's PR, should be removed
+                    shape->transform(matrix);
```
tinyjin, line 1132: Please ignore this changes, it's workaround to fix current rotation bug. Which will be closed in #2891 patch. After all, I'll remove this commit and line.

Alternatively, you can review after I rebase.

tinyjin:
@hermet I am afraid if new text grouping could affect to performance. This grouping approach will occupy as many tvgScenes as characters exist in the Text, when text-grouping is `Chars/All`. (Meaning more tvgScenes will be generated than before)

Let me know whether this should be changed, or if you have any better idea.

hermet:
@mgrudzinska any of your feedback is welcome.

> @hermet I am afraid if new text grouping could affect to performance. This grouping approach will occupy as many tvgScenes as characters exist in the Text, when text-grouping is `Chars/All`. (Meaning more tvgScenes will be generated than before)
> 
> Let me know whether this should be changed, or if you have any better idea.

@mgrudzinska any opinion?

mgrudzinska:
@tinyjin I tried to remove as many scenes as possible from text and group letters into one shape if possible, since for now each letter is handled as a separate shape. 
https://github.com/mgrudzinska/thorvg/commit/1db56daf6a9c33ff07eb9bd29ea2f3a6d9d6738d
I thought that I'm close, but after tests I see that I didn't manage to get a working solution.
the main idea was to transform the path points asap (l.1048) and then use shape's matrix to apply transformations that are now applied to the scene. but even there, each letter in a range selector I represent as a separate shape. don;t see how to optimize this. 
but since I failed it's hard for me to advice here

src/loaders/lottie/tvgLottieBuilder.cpp
hermet: @tinyjin could you please rebase PR?

tinyjin:
@mgrudzinska Have just rebased. The examples above work well again as each animations shows in the screenshot.

However, given example in your comment. ThorVG renders like this. (When rotated, position is little different from lottie-web)

Seems like ThorVG and lottie-web, they take different anchor point. I assume this is the issue out of current patch.

CleanShot 2024-10-29 at 16 19 24@2x

tinyjin:
> @tinyjin I tried to remove as many scenes as possible from text and group letters into one shape if possible, since for now each letter is handled as a separate shape. [mgrudzinska@1db56da](https://github.com/mgrudzinska/thorvg/commit/1db56daf6a9c33ff07eb9bd29ea2f3a6d9d6738d) I thought that I'm close, but after tests I see that I didn't manage to get a working solution. the main idea was to transform the path points asap (l.1048) and then use shape's matrix to apply transformations that are now applied to the scene. but even there, each letter in a range selector I represent as a separate shape. don;t see how to optimize this. but since I failed it's hard for me to advice here

Thank you for sharing your thoughts :)

Yes, this isn't easy part. Only concern from my end is affecting performance. Even though we introduce more scenes, it will be fine if this doesn't harm performance.

We can simply test this with animation that has many characters, that group by chars. I will soon bring testing result.

tinyjin:
@mgrudzinska 
In my local build, it renders well?

If you mean the title of the box, that's because animation is modified (Someone removed text property in the file)

CleanShot 2024-10-30 at 13 00 43@2x

CleanShot 2024-10-30 at 13 00 26@2x

mgrudzinska:
@jinny I must check on the old branch, sorry. these examples are working great!
although I found other issue:

[words.json](https://github.com/user-attachments/files/17568644/words.json)

Zrzut ekranu 2024-10-30 o 15 42 16

[lines.json](https://github.com/user-attachments/files/17568647/lines.json)

Zrzut ekranu 2024-10-30 o 15 43 11

I used rotation but I know we have small issue with the anchor, that is not what I mean here. there is other issue

tinyjin:
@mgrudzinska Thanks for this! This issue comes from Text Based + Text Grouping.

Just fixed for this case. I might need to create the other combination samples, want to test more.
How about we fix this in separated PR? (It doesn't only come from Text Alignment Options, but from Text Based)

CleanShot 2024-10-31 at 14 24 38@2x

tinyjin:
I've pushed for this. @mgrudzinska 

> [!NOTE]  
> tldr; Some of glyphs in text group is not always selected, specifically in the case Text Based has smaller unit than Text Grouping


When unit of Text Based is smaller than Text Grouping, it had problem.
(ex_ Text Based = Char, Text Grouping = Line)

It happened like
- Text Group contains glyphs
- Some of glyphs isn't selected
- Still transformed by Text Grouping

So I made patch like
- When glyphs isn't selected
- Exclude them from Text Grouping
- Then Range Selector would only perform for valid glyphs that selected


---
I used and made samples including files you attached. Text Alignment Options did transform character out of range because of the text group.

Now it does only for characters in the range :)


[based_word+grouping_line.json](https://github.com/user-attachments/files/17584969/based_word%2Bgrouping_line.json)

CleanShot 2024-10-31 at 17 13 49@2x

[based_char+grouping_word.json](https://github.com/user-attachments/files/17584990/based_char%2Bgrouping_word.json)

CleanShot 2024-10-31 at 17 14 50@2x




[based_char+grouping_line.json](https://github.com/user-attachments/files/17584994/based_char%2Bgrouping_line.json)



CleanShot 2024-10-31 at 17 15 48@2x


Note: Some difference in spacing and anchor point is from the other issue we've already known.

mgrudzinska (approved):

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -219,6 +219,14 @@ struct LottieTextRange
 };
 
 
+enum LottieTextGrouping : uint8_t { Chars = 1, Word = 2, Line = 3, All = 4 };
```
hermet, line 222: It looks this data structure belongs entirely to LottieText; thus:
```
struct LottieText
{
    ...  
    struct AlignOp or AlignOption {  
        enum Group : uint8_t { ....... };
        LottiePoint anchor or origin;   //Used to change the origin point for transformations
        Group group;
   } align or alignOp or alignOption;   //alignment option
};
```

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -253,6 +253,13 @@ struct LottieMarker
 
 struct LottieText : LottieObject, LottieRenderPooler<tvg::Shape>
 {
+    struct AlignOption
+    {
+        enum Group : uint8_t { Chars = 1, Word = 2, Line = 3, All = 4 };
+        Group grouping = Chars;
+        LottiePoint groupAlignment = Point{0.0f, 0.0f};
```
hermet, line 260: I suggest a name "anchor" / "origin" / "pivot", these sounds much clearer than "groupAlignment"

hermet (commented):
@tinyjin + minor nit-pick.

hermet (approved):
@tinyjin could you please squash 2 commits?

tinyjin:
@hermet Done :)

PR #2903, opened 2024-10-24
svg: prevent runtime error

mgrudzinska:
runtime error: applying non-zero offset 1 to null pointer Observed for Bespoke-leather-belt-2016012857.svg


mgrudzinska:
error observed with sanitizers asan&undef

JSUYA (approved):

Issue #2837, opened 2024-10-08
WebGL2 support


hermet:
ThorVG OpenGL is in pretty good shape. It's time to integrate it with WebGL2.

PR #2906, opened 2024-10-25
gl_engine: seperate the index buffer from vertex buffer

RuiwenTang:
WebGL has a strict rule that does not allow the same GLBuffer to be bound to both ARRAY_BUFFER and ELEMENT_ARRAY_BUFFER at the same time. (https://registry.khronos.org/webgl/specs/latest/1.0/#5.14.5)

To support WebGL in the future, a separate GLBuffer is used to store index data.

prepare for #2837

### DIFF
--- a/src/renderer/gl_engine/tvgGlGpuBuffer.cpp
+++ b/src/renderer/gl_engine/tvgGlGpuBuffer.cpp
@@ -72,7 +72,7 @@ void GlGpuBuffer::unbind(Target target)
     GL_CHECK(glBindBuffer(static_cast<uint32_t>(target), 0));
 }
 
-GlStageBuffer::GlStageBuffer() : mVao(0), mGpuBuffer(new GlGpuBuffer)
+GlStageBuffer::GlStageBuffer() : mVao(0), mGpuBuffer(), mGpuIndexBuffer()
 {
     GL_CHECK(glGenVertexArrays(1, &mVao));
 }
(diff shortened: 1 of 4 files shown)


src/renderer/gl_engine/tvgGlGpuBuffer.cpp
```diff
@@ -102,37 +102,56 @@ uint32_t GlStageBuffer::push(void *data, uint32_t size, bool alignGpuOffset)
     return offset;
 }
 
+uint32_t GlStageBuffer::pushIndex(void *data, uint32_t size)
+{
+    uint32_t offset = mIndexBuffer.count;
+
+    if (this->mIndexBuffer.reserved - this->mIndexBuffer.count < size) {
```
hermet, line 109: this (109 ~ 111) looks not necessary. 
At reserve you want to use? `reserve()` instead of `grow()`

hermet:
@RuiwenTang What do you think about the performance downside? If so, perhaps we could use a VBO with index data by default and provide an option to use a separate buffer object specifically for WebGL.

src/renderer/gl_engine/tvgGlGpuBuffer.cpp
RuiwenTang: I just want to make sure that we need to resize the container

RuiwenTang:
> @RuiwenTang What do you think about the performance downside? If so, perhaps we could use a BO with index data by default and provide an option to use a separate IBO specifically for WebGL. I guess `mGpuIndexBuffer` can abstract it.

The impact on performance should be small. The design here is to avoid frequent swapping of VertexBuffer, IndexBuffer and UniformBuffer between DrawElements.

hermet:
> VertexBuffer, IndexBuffer

@RuiwenTang I think you are introducing this for WebGL spec not for other benefits. Just considered, we're introducing an additional buffer object for the index here. Imagine a highly complex, fine-grained shapes with thousands of polygons (i.e 8K vector image map). Fetching data from a single buffer and fetching data from two buffers would not yield the same performance. Just my guess, that's why I suggested the optional choice.  We can go with this now, just wanted to clarify it.

hermet (approved):

PR #2914, opened 2024-10-29
lottie: text range selector++

mgrudzinska:
Change in the algorithm for selecting characters included in the range selector. This is the first step towards adding support for maxAmount, smoothness, and easing.
Smoothness and maxAmount added.

@Issue: https://github.com/thorvg/thorvg/issues/2178

AE vs this pr (not sync)
sm

[smoothness.json](https://github.com/user-attachments/files/17581496/smoothness.json)

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -1091,15 +1091,13 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
 
                     //text range process
                     for (auto s = text->ranges.begin(); s < text->ranges.end(); ++s) {
-                        float start, end;
-                        (*s)->range(frameNo, float(totalChars), start, end);
-
                         auto basedIdx = idx;
                         if ((*s)->based == LottieTextRange::Based::CharsExcludingSpaces) basedIdx = idx - space;
                         else if ((*s)->based == LottieTextRange::Based::Words) basedIdx = line + space;
                         else if ((*s)->based == LottieTextRange::Based::Lines) basedIdx = line;
 
-                        if (basedIdx < start || basedIdx >= end) continue;
+                        auto f = (*s)->factor(frameNo, float(totalChars), (float)basedIdx);
+                        if (tvg::zero(f)) continue;
                         needGroup = true;
 
                         translation = translation + (*s)->style.position(frameNo);
@@ -1141,13 +1139,13 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
                         //center pivoting
                         textGroupMatrix.e13 += (pivotX * textGroupMatrix.e11 + pivotX * textGroupMatrix.e12);
                         textGroupMatrix.e23 += (pivotY * textGroupMatrix.e21 + pivotY * textGroupMatrix.e22);
-                        
+
                         textGroup->transform(textGroupMatrix);
                     }
 
                     auto& matrix = shape->transform();
                     identity(&matrix);
-                    translate(&matrix, (translation.x / scale + cursor.x) - textGroupMatrix.e13, (translation.y / scale + cursor.y) - textGroupMatrix.e23);
+                    translate(&matrix, translation.x / scale + cursor.x - textGroupMatrix.e13, translation.y / scale + cursor.y - textGroupMatrix.e23);
                     tvg::scale(&matrix, scaling.x, scaling.y);
                     shape->transform(matrix);
                 }
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -1083,13 +1083,15 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
                     shape->strokeFill(doc.stroke.color.rgb[0], doc.stroke.color.rgb[1], doc.stroke.color.rgb[2]);
                 }
 
-                bool needGroup = false;
+                auto needGroup = false;
+                //text range process
                 if (!text->ranges.empty()) {
                     Point scaling = {1.0f, 1.0f};
                     auto rotation = 0.0f;
                     Point translation = {0.0f, 0.0f};
+                    auto color = doc.color;
+                    auto strokeColor = doc.stroke.color;
 
-                    //text range process
                     for (auto s = text->ranges.begin(); s < text->ranges.end(); ++s) {
                         auto basedIdx = idx;
                         if ((*s)->based == LottieTextRange::Based::CharsExcludingSpaces) basedIdx = idx - space;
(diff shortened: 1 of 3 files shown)


mgrudzinska:
No samples to test are included, since no changes should be observed

tinyjin (approved):
LGTM

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1067,15 +1067,13 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
...
-                        float start, end;
-                        (*s)->range(frameNo, float(totalChars), start, end);
-
                         auto basedIdx = idx;
                         if ((*s)->based == LottieTextRange::Based::CharsExcludingSpaces) basedIdx = idx - space;
                         else if ((*s)->based == LottieTextRange::Based::Words) basedIdx = line + space;
                         else if ((*s)->based == LottieTextRange::Based::Lines) basedIdx = line;
 
-                        if (basedIdx < start || basedIdx >= end) continue;
+                        auto f = (*s)->factor(frameNo, float(totalChars), (float)basedIdx);
```
hermet, line 1075: 1. Not sure the subsequent changes you mind now, probably, you can also consider use:
```
if ((*s)->factor(frameNo, float(totalChars), (float)basedIdx)) continue;
```
2. The name `factor` is too ambiguous in this context. I couldn't find any beneficial to change it from the original.

hermet (changes requested):
`factor` is bad name here,  I couldn't find any benefical to change it from the original. Please confirm this point.

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: please take a look here: https://github.com/thorvg/thorvg/pull/2898/files
I can push the other changes here if it's more readable.
this is a scaling factor that will be used to scale all text range styling properties. calling it a 'range' doesn't make a sense any more.

hermet: @mgrudzinska If this patch alone doesn’t provide any benefit, please combine it with a patch that has a direct improvement or impact and submit them together in a single PR.

mgrudzinska:
@tinyjin since now there is more changes in this pr, please rev one more time

mgrudzinska:
+ rebase

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -31,7 +31,16 @@
 /* Internal Class Implementation                                        */
 /************************************************************************/
 
+static inline float _applySmoothness(float value, float smoothness)
```
hermet, line 34: If the logic of this function is not very lengthy and it won’t be called from multiple places, there’s no need to separate it into its own function. I think you can merge it in `factor()` with comment `//apply smoothness`

mgrudzinska: done

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1067,33 +1067,43 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
...
-                        rotation += (*s)->style.rotation(frameNo);
+                        translation = translation + f * (*s)->style.position(frameNo);
+                        scaling = scaling * (f * ((*s)->style.scale(frameNo) * 0.01f - Point{1.0f,1.0f}) + Point{1.0f,1.0f});
+                        rotation += f * (*s)->style.rotation(frameNo);
 
                         shape->opacity((*s)->style.opacity(frameNo));
 
                         auto color = (*s)->style.fillColor(frameNo);
-                        shape->fill(color.rgb[0], color.rgb[1], color.rgb[2], (*s)->style.fillOpacity(frameNo));
+                        if (f == 1.0f) shape->fill(color.rgb[0], color.rgb[1], color.rgb[2], (*s)->style.fillOpacity(frameNo));
```
hermet, line 1085: you might wanna use `tvg::equal(f, 1.0f)`

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1067,33 +1067,43 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
...
+                        else {
+                            auto r = lerp<uint8_t>(doc.color.rgb[0], color.rgb[0], f);
+                            auto g = lerp<uint8_t>(doc.color.rgb[1], color.rgb[1], f);
+                            auto b = lerp<uint8_t>(doc.color.rgb[2], color.rgb[2], f);
+                            shape->fill(r, g, b, (*s)->style.fillOpacity(frameNo));
+                        }
```
tinyjin, line 1091: Isn't this, factor for fill color, this should multiply color by `f`? This looks like factor replaces RGB data instead.

mgrudzinska: I don't understand what do you mean. 
so I double checked the blended colors, look fine

result AE vs this pr:
Zrzut ekranu 2024-11-3 o 10 31 30

sample:
[smoothnessABC.json](https://github.com/user-attachments/files/17609131/smoothnessABC.json)

mgrudzinska:
@tinyjin please take a look whether I rebased correctly.

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: @hermet please check now - color is now interpolated taking into account the pervious color, if was set

before last changes:
Zrzut ekranu 2024-11-4 o 11 22 42

after last change:
Zrzut ekranu 2024-11-4 o 12 16 27

sample:
[smoothness2Ranges.json](https://github.com/user-attachments/files/17614017/smoothness2Ranges.json)

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1083,42 +1083,67 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
...
+                                    color.rgb[2] = P(shape)->rs.stroke->color[2];
+                                }
+                                auto r = lerp<uint8_t>(color.rgb[0], rangeColor.rgb[0], f);
+                                auto g = lerp<uint8_t>(color.rgb[1], rangeColor.rgb[1], f);
+                                auto b = lerp<uint8_t>(color.rgb[2], rangeColor.rgb[2], f);
+                                shape->strokeFill(r, g, b, (*s)->style.strokeOpacity(frameNo));
+                            }
                         }
-                        cursor.x += (*s)->style.letterSpacing(frameNo);
+                        setColor = true;
```
tinyjin, line 1142: setColor represents the condition for both stroke and fill?

I understand this value is used to ensure interpolating from previous color, then probably boolean value should be devided:
- setFillColor
- setStrokeColor

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -110,21 +110,40 @@ void LottieSlot::assign(LottieObject* target)
...
+    if (idx >= std::floor(start)) {
+        auto diff = idx - start;
+        f = diff < 0.0f ? std::min(end, 1.0f) + diff : end - idx;
+        clamp(f, 0.0f, 1.0f);
+    }
+
+    //apply smoothness
+    auto smoothness = this->smoothness(frameNo);
+    if (smoothness < 100.0f) {
+        smoothness *= 0.01f;
```
tinyjin, line 141: if we always use smoothness multiplied by `0.01f`, how about we do in parsing step?

```cpp
// parser
else if (KEY_AS("sm")) {
  parseProperty<LottieProperty::Type::Float>(selector->smoothness);
  selector->smoothness *= 0.01f;
}

// model
auto smoothness = this->smoothness(frameNo);
if (smoothness < 1.0f) {
```

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -110,21 +110,40 @@ void LottieSlot::assign(LottieObject* target)
...
+        f = (f - (1.0f - smoothness) * 0.5f) / smoothness;
+        clamp(f, 0.0f, 1.0f);
+    }
 
-    auto range = end - start;
-    auto len = (rangeUnit == Unit::Percent) ? 100.0f : totalLen;
-    start = static_cast<float>(random % int(len - range));
-    end = start + range;
+    return f * this->maxAmount(frameNo) * 0.01f;
```
tinyjin, line 146: Same opinion as smoothness

```cpp
if (KEY_AS("a")) {
  parseProperty<LottieProperty::Type::Float>(selector->maxAmount);
  selector->maxAmount *= 0.01f
}
```

tinyjin (commented):
Rebase is good. Leaving some comments

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: for now doesn't matter, since we don;t have flags for colors. as you noticed before, there is an issue with colors, we'll be solved in a separate pr.

src/loaders/lottie/tvgLottieModel.cpp
mgrudzinska: but this value can be animated, this is LottieProperty::Type::Float, same with maxAmount

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1083,42 +1083,67 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
...
+                        rotation += f * (*s)->style.rotation(frameNo);
 
                         shape->opacity((*s)->style.opacity(frameNo));
 
-                        auto color = (*s)->style.fillColor(frameNo);
-                        shape->fill(color.rgb[0], color.rgb[1], color.rgb[2], (*s)->style.fillOpacity(frameNo));
+                        auto rangeColor = (*s)->style.fillColor(frameNo);
+                        if (tvg::equal(f, 1.0f)) shape->fill(rangeColor.rgb[0], rangeColor.rgb[1], rangeColor.rgb[2], (*s)->style.fillOpacity(frameNo));
+                        else {
+                            auto color = doc.color;
```
hermet, line 1113: 1.  What about opacity?

2. Suggestion logic:
```
auto color = doc.color;
auto strokeColor = doc.stroke.color;
for (auto s = text->ranges.begin(); s < text->ranges.end(); ++s) {
   ...
   color = interpolate (color, rangeColor, f);
   ...
}
shape->fill(color);
if (doc.stroke.render) shape->strokeFill(strokeColor);
```

src/loaders/lottie/tvgLottieBuilder.cpp
hermet: @mgrudzinska Please leave comments in the code (e.g., FIXME, TODO) for any known issues or tasks to do.

mgrudzinska:
maxAmount sample: 
[maxAmount.json](https://github.com/user-attachments/files/17633662/maxAmount.json)

AE and this pr (shifted...)
maxA

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: sample for opacity - works
[opacity.json](https://github.com/user-attachments/files/17633984/opacity.json)

mgrudzinska:
rebased

tinyjin (approved):
LGTM

src/loaders/lottie/tvgLottieBuilder.cpp
hermet: @mgrudzinska We cannot confirm with this sample, as it always has a factor of 1.0f, which means opacity with interpolation also works as expected.

hermet: By default, interpolating other properties, including color, while excluding only opacity does not seem to align with lottie spec. It would be better to validate cases where issues arise when interpolating opacity.

hermet: By default, interpolating other properties, including color, while excluding only opacity does not seem to align with lottie spec. It would be better to suggest cases where issues arise when interpolating opacity.

hermet (approved):

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: @hermet please check #2949

Issue #2592, opened 2024-07-24
wg_engine: SVG rendering compatibility


hermet:
This list shows the broken compatibility of SVG rendering with the WebGPU backend.
https://github.com/thorvg/thorvg/tree/main/examples/resources/svg

- [ ] svg2009 //bridge strings not clear (have no ideas how to fix)
- [x] patch //dash line
- [x] consecutive_lines //line cap
- [x] rg1024_Ufo_in_metalic_style //cap color
- [x] vnu //corner quality is too poor

hermet:
Pass rate: 71%

SergeyLebedkin:
rendering problems arise when applying a gradient transformation.
See https://github.com/thorvg/thorvg/issues/2620

hermet:
Pass rate: 89%

hermet:
Pass rate: 94%

hermet:
@SergeyLebedkin edge blending error?
![image](https://github.com/user-attachments/assets/b95b8827-b1ac-43c9-bd18-648f11041093)

SergeyLebedkin:
[...]

PR #2916, opened 2024-10-30
wg_engine: fix equals gradient offsets

SergeyLebedkin:
In case if gradient offsets are equal the last color are used instead of first

f.e
```
//Gradient Color Stops
tvg::Fill::ColorStop colorStops[3];
colorStops[1] = {0, 128, 128, 128, 255};
colorStops[2] = {1, 255, 255, 255, 255};
```
(128, 128, 128) will be used instaed of (0, 0, 0)

See _rg1024_Ufo_in_metalic_style.svg_
https://github.com/thorvg/thorvg/issues/2592

Reference:
image

Before:
image

After:
image

### DIFF
--- a/src/renderer/wg_engine/tvgWgShaderTypes.cpp
+++ b/src/renderer/wg_engine/tvgWgShaderTypes.cpp
@@ -146,6 +146,8 @@ void WgShaderTypeGradient::updateTexData(const Fill::ColorStop* stops, uint32_t 
     for (uint32_t i = 1; i < stopCnt; i++)
         if (sstops.last().offset < stops[i].offset)
             sstops.push(stops[i]);
+        else if (sstops.last().offset == stops[i].offset)
+            sstops.last() = stops[i];
     // head
     uint32_t range_s = 0;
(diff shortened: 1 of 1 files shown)


src/renderer/wg_engine/tvgWgShaderTypes.cpp
```diff
@@ -146,6 +146,8 @@ void WgShaderTypeGradient::updateTexData(const Fill::ColorStop* stops, uint32_t
     for (uint32_t i = 1; i < stopCnt; i++)
         if (sstops.last().offset < stops[i].offset)
             sstops.push(stops[i]);
+        else if (sstops.last().offset == stops[i].offset)
```
hermet, line 149: I think the addition can be done outside the for loop?

SergeyLebedkin: Nope.. thats becouse not onlu first or last offsets can be the same, but in a middle of the list. last() == current() of the filtered list

hermet (approved):

mgrudzinska (approved):

PR #2917, opened 2024-10-30
wg_engine: fix cubic splines and circles tesselation quality

SergeyLebedkin:
Tesseletion factor now depends on scale matix of the shape

See _vnu.svg_
https://github.com/thorvg/thorvg/issues/2592

Before:
image

After:
image

### DIFF
--- a/src/renderer/wg_engine/tvgWgRenderData.cpp
+++ b/src/renderer/wg_engine/tvgWgRenderData.cpp
@@ -381,8 +381,12 @@ void WgRenderDataShape::updateMeshes(WgContext &context, const RenderShape &rsha
     releaseMeshes(context);
     strokeFirst = rshape.stroke ? rshape.stroke->strokeFirst : false;
 
+    // get object scale
+    float scale = std::max(std::min(length(Point{tr.e11 + tr.e12,tr.e21 + tr.e22}), 8.0f), 1.0f);
+
     // path decoded vertex buffer
     WgVertexBuffer pbuff;
+    pbuff.reset(scale);
     // append shape without strokes
     if (!rshape.stroke) {
         pbuff.decodePath(rshape, false, [&](const WgVertexBuffer& path_buff) {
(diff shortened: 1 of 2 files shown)


src/renderer/wg_engine/tvgWgRenderData.cpp
```diff
@@ -381,8 +381,12 @@ void WgRenderDataShape::updateMeshes(WgContext &context, const RenderShape &rsha
     releaseMeshes(context);
     strokeFirst = rshape.stroke ? rshape.stroke->strokeFirst : false;
 
+    // get object scale
+    float scale = length(Point{1.0, 1.0} * tr) / MATH_SQRT2;
```
hermet, line 385: It's enough to use `sqrt(2)`. let's remove a new macro.

hermet (changes requested):

hermet (approved):

mgrudzinska (commented):
I have doubts about whether the way "scale" is calculated in tvgRenderData is the best approach. It works for the vnu sample because the default value of 8 is used, but that might not be sufficient. 
One issue is that the number of points will depend on the shape's position (translation is included into scaling factor), which seems incorrect. 
The other problem is this sample (modified vnu sample):
[vnu_modified.svg.zip](https://github.com/user-attachments/files/17587876/vnu_modified.svg.zip)

Zrzut ekranu 2024-10-31 o 19 37 29

I rescaled everything in this svg file, but I saw many times issues with attached 'small' svgs, so seems like a reasonable example

SergeyLebedkin:
Made a changes in scale factor calculations.

> I have doubts about whether the way "scale" is calculated in tvgRenderData is the best approach. It works for the vnu sample because the default value of 8 is used, but that might not be sufficient. One issue is that the number of points will depend on the shape's position (translation is included into scaling factor), which seems incorrect. The other problem is this sample (modified vnu sample): [vnu_modified.svg.zip](https://github.com/user-attachments/files/17587876/vnu_modified.svg.zip)
> 
> Zrzut ekranu 2024-10-31 o 19 37 29
> I rescaled everything in this svg file, but I saw many times issues with attached 'small' svgs, so seems like a reasonable example

I think we will handle 'small' svgs later.

SergeyLebedkin:
Translation is removed now

PR #2923, opened 2024-10-31
wg_engine: fix close command logic

SergeyLebedkin:
On a close path command creates a new object started form the closed point

This change handle cases like this:

```
shape->moveTo(10, 10);
shape->lineTo(510, 10);
shape->lineTo(510, 510);
shape->lineTo(10, 510);
shape->close();
shape->lineTo(510, 510);
```

See: _consecutive_lines.svg_
Issue https://github.com/thorvg/thorvg/issues/2592

Before:
image

After:
image

### DIFF
--- a/src/renderer/wg_engine/tvgWgGeometry.h
+++ b/src/renderer/wg_engine/tvgWgGeometry.h
@@ -198,8 +198,7 @@ struct WgVertexBuffer
             if (cmd == PathCommand::MoveTo) {
                 // after path decoding we need to update distances and total length
                 if (update_dist) updateDistances();
-                if ((onPolyline) && (vcount != 0))
-                    onPolyline(*this);
+                if ((onPolyline) && (vcount > 0)) onPolyline(*this);
                 reset(tscale);
                 append(rshape.path.pts[pntIndex]);
                 pntIndex++;
@@ -208,15 +207,28 @@ struct WgVertexBuffer
                 pntIndex++;
             } else if (cmd == PathCommand::Close) {
                 close();
+                // proceed path if close command is not the last command and next command is LineTo or CubicTo
+                if ((cmdIndex + 1 < rshape.path.cmds.count) &&
+                    ((rshape.path.cmds[cmdIndex + 1] == PathCommand::LineTo) ||
+                     (rshape.path.cmds[cmdIndex + 1] == PathCommand::CubicTo))) {
+                    // proceed current path
+                    if (update_dist) updateDistances();
+                    if ((vcount > 0) && (onPolyline)) onPolyline(*this);
+                    // append closing point of current path as a first point of the new path
+                    Point last_pt = last();
+                    reset(tscale);
+                    append(last_pt);
+                }
             } else if (cmd == PathCommand::CubicTo) {
+                // append tesselated cubic spline with tscale param
                 appendCubic(vbuff[vcount - 1], rshape.path.pts[pntIndex + 0], rshape.path.pts[pntIndex + 1], rshape.path.pts[pntIndex + 2]);
                 pntIndex += 3;
             }
         }
         // after path decoding we need to update distances and total length
         if (update_dist) updateDistances();
-        if ((onPolyline) && (vcount != 0))
-            onPolyline(*this);
+        if ((vcount > 0) && (onPolyline)) onPolyline(*this);
+        reset(tscale);
     }
 };
 


hermet:
@SergeyLebedkin As per our contribution guide, please include the relevant issue link in the commit message.

https://github.com/thorvg/thorvg/blob/main/CONTRIBUTING.md
image

reference: https://github.com/thorvg/thorvg/commit/752e9419e390149eac88e2cfc2b2dc37c77814b6

SergeyLebedkin:
> @SergeyLebedkin As per our contribution guide, please include the relevant issue link in the commit message.
> 
> reference: [752e941](https://github.com/thorvg/thorvg/commit/752e9419e390149eac88e2cfc2b2dc37c77814b6)

fixed

src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -208,7 +207,14 @@ struct WgVertexBuffer
                 pntIndex++;
             } else if (cmd == PathCommand::Close) {
                 close();
+                // close command creates the new object, started form the closing point
+                if (update_dist) updateDistances();
```
mgrudzinska, line 211: but it makes sense only if there is a next point and it is not a 'move to', right? maybe check for that?

SergeyLebedkin: fixed

src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -208,15 +207,22 @@ struct WgVertexBuffer
                 pntIndex++;
             } else if (cmd == PathCommand::Close) {
                 close();
+                // close command creates the new object, started form the closing point
+                if ((vcount > 0) && (update_dist)) updateDistances();
+                if ((vcount > 0) && (onPolyline)) onPolyline(*this);
+                Point last_pt = last();
+                reset(tscale);
+                if (cmdIndex + 1 < rshape.path.cmds.count) append(last_pt);
```
mgrudzinska, line 215: this is what I meant:
               ```
               close();
               if (cmdIndex == rshape...count || rshape...cmds[cmdIndex+1] != MoveTo) ccontinue;
               //rest of the code
                ```

```vcount > 0``` this can be done once

File: src/renderer/wg_engine/tvgWgGeometry.h [hunk 3]
```diff
@@ -208,15 +207,24 @@ struct WgVertexBuffer
...
             } else if (cmd == PathCommand::Close) {
                 close();
+                // close command creates the new object, started form the closing point
+                if (update_dist) updateDistances();
+                if ((vcount > 0) && (onPolyline)) onPolyline(*this);
+                Point last_pt = last();
+                reset(tscale);
+                // append new point in a case if close command is not the last command and next command is LineTo
+                if ((cmdIndex < rshape.path.cmds.count - 1) && (rshape.path.cmds[cmdIndex + 1] == PathCommand::LineTo))
+                    append(last_pt);
```
mgrudzinska, line 217: @SergeyLebedkin from the beggining I have the same comment: why it can not be done as below?
```
if ((cmdIndex < rshape.path.cmds.count - 1) && (rshape.path.cmds[cmdIndex + 1] != PathCommand::MoveTo)) {
    if (update_dist) updateDistances();
    if ((vcount > 0) && (onPolyline)) onPolyline(*this);
    reset(tscale);
    append(last());
}
```
I understand that you want to handle cases different than 'Close MoveTo' (so either 'Close LineTo' or 'Close CubicTo'). For a standard case 'Close MoveTo' nothing has to be changed, right? it worked till now correct, so there is no need to call neither the updateDistances() nor the onPolyline()? Or these were missing and this pr solves 2 problems?

SergeyLebedkin: We need to handle cases, when after Close Command we have a command like LineTo or CubicTo.
We don`t handle this cases till now.
In this cases we need to proceed path (updateDist() and OnPolylone()) and restart new object path from the closed point

mgrudzinska: I understand this, that's why I'm asking why don;t you add a check 'if ((cmdIndex < rshape.path.cmds.count - 1) && (rshape.path.cmds[cmdIndex + 1] != PathCommand::MoveTo))' just after close() is called

SergeyLebedkin: @mgrudzinska so qustion is closed?

SergeyLebedkin: @mgrudzinska and you was right with a CubicTo case. So I also add this case

mgrudzinska (approved):

hermet (approved):

Issue #2907, opened 2024-10-25
gl_engine: rendering differences when running C API example with Software vs OpenGL


SuperAuguste:
See https://github.com/SuperAuguste/thorvg/tree/opengl-capi which adds OpenGL canvas support to the C API and merges in https://github.com/thorvg/thorvg/pull/2866.

With those changes, here's what you get when you run the C API example with Software (left) and OpenGL (right):

![image](https://github.com/user-attachments/assets/284ce5c9-0991-4a76-990b-4875414704fe)

hermet:
\+ @RuiwenTang

hermet:
@SuperAuguste hello, the fix will be in thorvg v0.15.4 / 1.0.0-pre6. thanks for your report.

RuiwenTang:
The dash line and mask part still looks different, I think we should reopen this issue

hermet:
After this: https://github.com/thorvg/thorvg/pull/3028, improved trimming but still it has an issue about dash
![image](https://github.com/user-attachments/assets/5f3cdf2e-fc12-413e-8d2e-ee5a95ac8dd8)

RuiwenTang:
dash is an other problem needs to be fixed :(

mgrudzinska:
Dashing is compatible now in GL and SW. 

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/a6174014-1769-4055-afca-9740e9991ba0" />

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/1ce2a123-7956-49e8-934c-2060668c33fa" />

there are still differences in masking, but this seems to be an issue on the sw_engine side: https://github.com/thorvg/thorvg/issues/3263

hermet:
@SuperAuguste Hello, all the gl rendering issue has been resolved here please check with [v1.0-pre14](https://github.com/thorvg/thorvg/releases/tag/v1.0-pre14), Thanks.

Issue #2728, opened 2024-09-12
wg_engine & gl_engine: focal point support in radial grad is missing


mgrudzinska:
https://www.w3.org/TR/SVG2/pservers.html#RadialGradientElementFXAttribute

```
<svg width="600" height="600" version="1.1" id="https://github.com/thorvg/thorvg/issues/2014" xmlns="http://www.w3.org/2000/svg">
   <defs>
      <radialGradient cx="500" cy="300" fx="550" gradientUnits="userSpaceOnUse" id="b" r="60" spreadMethod="reflect">
         <stop offset="0" stop-color="red" id="stop10" />
         <stop offset="1" stop-color="white" id="stop11" />
      </radialGradient>
   </defs>

   <path d="M 0,0 H 600 V 600 H -600 Z" fill="url(#b)" id="svg_4" />
</svg>
```

expected (SW):
<img width="263" alt="Zrzut ekranu 2024-09-12 o 15 24 38" src="https://github.com/user-attachments/assets/43918cb6-b980-41a3-99fb-77ad96240589">

current (WG, GL):
<img width="265" alt="Zrzut ekranu 2024-09-12 o 15 24 36" src="https://github.com/user-attachments/assets/2ed418ea-a5a6-4099-9860-a189672f85a7">

PR #2934, opened 2024-11-06
gl_engine: Support new radial gradient data struct and calculation

RuiwenTang:
This PR support the new RadialGradient API.
The mathematical calculation comes from https://skia.org/docs/dev/design/conical/

related issue: #2907 #2728 #2936

Now the gradient parts is rendering correct:

截屏2024-11-06 19 43 40

### DIFF
--- a/src/renderer/gl_engine/tvgGlShaderSrc.cpp
+++ b/src/renderer/gl_engine/tvgGlShaderSrc.cpp
@@ -198,7 +198,7 @@ std::string STR_LINEAR_GRADIENT_MAIN = TVG_COMPOSE_SHADER(
 std::string STR_RADIAL_GRADIENT_VARIABLES = TVG_COMPOSE_SHADER(
     layout(std140) uniform GradientInfo {                                                                   \n
         vec4  nStops;                                                                                       \n
-        vec2  centerPos;                                                                                    \n
+        vec4  centerPos;                                                                                    \n
         vec2  radius;                                                                                       \n
         vec4  stopPoints[MAX_STOP_COUNT / 4];                                                               \n
         vec4  stopColors[MAX_STOP_COUNT];                                                                   \n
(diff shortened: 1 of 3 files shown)


src/renderer/gl_engine/tvgGlShaderSrc.cpp
```diff
@@ -207,14 +207,107 @@ std::string STR_RADIAL_GRADIENT_VARIABLES = TVG_COMPOSE_SHADER(
 
 std::string STR_RADIAL_GRADIENT_MAIN = TVG_COMPOSE_SHADER(
     out vec4 FragColor;                                                                                     \n
+                                                                                                            \n
+    mat3 radial_matrix(vec2 p0, vec2 p1) {                                                                  \n
+        return mat3(0.0, -1.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 1.0) *                                         \n
+         inverse(mat3(p1.y - p0.y, p0.x - p1.x, 0.0, p1.x - p0.x, p1.y - p0.y,                              \n
+                      0.0, p0.x, p0.y, 1.0));                                                               \n
+    }                                                                                                       \n
```
SergeyLebedkin, line 215: can you please provide any refs, how did you get this transformation?
also,, are you sure, that we should compose this matrix every time for each pixel or we can compose it once in CPU side?

SergeyLebedkin (commented):
Please, see my questions

src/renderer/gl_engine/tvgGlShaderSrc.cpp
RuiwenTang: The matrix comes from [skia conical gradient](https://github.com/google/skia/blob/main/src/shaders/gradients/SkConicalGradient.cpp#L48). 
It can be done in CPU, but requires some other floating point precision judgment code I think it's too troublesome so I put it in Shader :(

SergeyLebedkin (approved):

hermet (approved):

PR #2940, opened 2024-11-08
gl_engine: Fix memory leak in the IndexBuffer object

RuiwenTang:
This PR fix a low-level error in GL backend :(

The CPU buffer cache in the IndexBuffer object needs to be cleared, otherwise it will cause memory leaks and reduce performance per frame


hermet (approved):

hermet (approved):
@RuiwenTang Please note that, `Lottie` example slow down still valid even after this PR.

Issue #2936, opened 2024-11-07
Inconsistent radial gradient fill output across raster engines


hermet:
Test: `examples/resources/svg/intertwingly.svg`

OpenGL and WebGPU output look identical but look different with software
![image](https://github.com/user-attachments/assets/dfe519ce-ff93-49e1-83ef-f7f6aac82058)

RuiwenTang:
I think the #2934  can fix this problem :)

PR #2942, opened 2024-11-08
wg_engine: support gradient focal point

SergeyLebedkin:
Implemented gradien focal points

Issue https://github.com/thorvg/thorvg/issues/2728 
Issue https://github.com/thorvg/thorvg/issues/2936

See https://skia.org/docs/dev/design/conical/
The implementetion provide general solution for the equation, solved by t
```
Setting Up the Equations
Interpolated Center:
Ct = (1-t) * C0 + t * C1
Interpolated Radius:
rt = (1-t) * r0 + t * r1
Distance from Ct to Pt:
|Pt - Ct| = rt
```

No any impact on performance

Before/After:
image

Before/After:
image


hermet (approved):
very nice!

Issue #2938, opened 2024-11-08
wg_engine: webgpu engine with test wouldn't be compilerable on MacOS


hermet:
Build option: `meson setup builddir -Dengines="wg", -Dtests=true`

![CleanShot 2024-10-29 at 16 53 20@2x](https://github.com/user-attachments/assets/2121edf9-f7b4-4542-9593-ec7adfa01b25)

_Originally posted by @tinyjin in https://github.com/thorvg/thorvg/pull/2910#pullrequestreview-2401032467_

hermet:
\+ @SergeyLebedkin

PR #2943, opened 2024-11-08
wg_engine: fix test build scripts and ambiguous declarations for macos

SergeyLebedkin:
Fixed conflict with ThorVG and Cocoa framework (required for webgpu on macos): Point is **ambiguous** type in a case of macos usage.

image

Conflics appears only in case of **examples + test + wg** compilation:
`meson setup builddir -Dexamples=true -Dtests=true -Dengines="wg"`

Tests must be compiled before examples, where Cocoa are included

Also fixed **try** issue by adding compilations flags to support exceptions in a case of macos usage
image

Issue https://github.com/thorvg/thorvg/issues/2938

_Alternatively_ we can do not change order of the compilation, but change variable declarations in testShape.cpp from Point to tvg::Point, that also removes ambiguous, f.e.

`const tvg::Point* pts2;`
but not
`const Point* pts2;`

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -1,9 +1,11 @@
+examples_compiler_flags = compiler_flags
+
 if lib_type == 'static'
-    compiler_flags += ['-DTVG_STATIC']
+    examples_compiler_flags += ['-DTVG_STATIC']
 endif
 
 if target_opengles
-    compiler_flags += '-DTHORVG_GL_TARGET_GLES=1'
+    examples_compiler_flags += '-DTHORVG_GL_TARGET_GLES=1'
 endif
 
 examples_dep = [dependency('sdl2')]
@@ -13,7 +15,7 @@ if all_engines or get_option('engines').contains('wg')
     if host_machine.system() == 'darwin'
         add_languages('objcpp')
         examples_dep += declare_dependency(link_args: ['-lobjc', '-framework', 'Cocoa', '-framework', 'CoreVideo', '-framework', 'IOKit', '-framework', 'QuartzCore'])
-        compiler_flags += ['-x', 'objective-c++']
+        examples_compiler_flags += ['-x', 'objective-c++']
     endif
 endif
 
@@ -86,7 +88,7 @@ foreach current_file : source_file
     executable(name, current_file,
         include_directories : headers,
         link_with : thorvg_lib,
-        cpp_args : compiler_flags,
+        cpp_args : examples_compiler_flags,
         dependencies : examples_dep)
 endforeach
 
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -140,14 +140,14 @@ subdir('inc')
 subdir('src')
 subdir('tools')
 
-if get_option('examples')
```
hermet, line 143: test/example both should be at the same level as the apps. How the examples are successfully buildable...

SergeyLebedkin: > test/example both should be at the same level as the apps. How the examples are successfully buildable...

I am not sure, if I understand what do you mean _at the same level as the apps_

hermet: @SergeyLebedkin The question is why the build order of 'example/test' needs to be switched; they shouldn't affect each other.

SergeyLebedkin: @hermet 
Its just a theory, but I think that problem is in _compiler_flags_ variable, that is the same for examples and test build script.
If examples builds before test, its meson.build script add **['-x', 'objective-c++']** flags, but for test they should not be there

SergeyLebedkin: @hermet 
please, review the new variant

SergeyLebedkin:
this variant keeps the original root build script, but provide independent compile flags for examples and tests

hermet (changes requested):
This seems to be going the wrong way. In my opinion, we need to clarify why the example build works, but the test build does not from you ex-approach.

examples/meson.build
```diff
@@ -1,9 +1,11 @@
+compiler_flags_examples = compiler_flags
+
 if lib_type == 'static'
-    compiler_flags += ['-DTVG_STATIC']
+    compiler_flags_examples += ['-DTVG_STATIC']
```
hermet, line 4: could you please use the name `examples_compiler_flags` ?

test/meson.build
```diff
@@ -1,5 +1,12 @@
+compiler_flags_test = compiler_flags
+
 if lib_type == 'static'
-    compiler_flags += ['-DTVG_STATIC']
+    compiler_flags_test += ['-DTVG_STATIC']
```
hermet, line 4: could you please use `test_compiler_flags` ?

hermet (commented):
@SergeyLebedkin thanks. could you please check some nitpicks?

test/meson.build
SergeyLebedkin: @hermet  Fixed and checked on macos

hermet (approved):

tinyjin (approved):
CleanShot 2024-11-12 at 22 48 45@2x

No build error, LGTM!

PR #2950, opened 2024-11-11
infra: prevent warning on windows

mgrudzinska:
"conversion from 'float' to 'int8_t', possible loss of data"


src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -282,7 +282,7 @@ void LottieParser::getValue(int8_t& val)
         //discard rest
         while (nextArrayValue()) getInt();
     } else {
-        val = getFloat();
+        val = (uint8_t)getFloat();
```
hermet, line 285: int8_t

hermet (approved):

PR #2951, opened 2024-11-11
lottie: easing support in text range selector

mgrudzinska:
@issue: #2178


src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1163,7 +1163,10 @@ void LottieParser::parseTextRange(LottieText* text)
                 enterObject();
                 while (auto key = nextObjectKey()) {
                     if (KEY_AS("t")) selector->expressible = (bool) getInt();
-                    else if (KEY_AS("xe")) parseProperty<LottieProperty::Type::Float>(selector->maxEase);
+                    else if (KEY_AS("xe")) {
+                        parseProperty<LottieProperty::Type::Float>(selector->maxEase);
+                        selector->interpolator = static_cast<LottieInterpolator*>(malloc(sizeof(LottieInterpolator)));
```
tinyjin, line 1168: Can't it be allocated like this?

```cpp
selector->interpolator = new LottieInterpolator();
```

mgrudzinska: it can. I didn't need to initialize it now, since it will be done later while set is called. better perf

hermet (approved):

Issue #2958, opened 2024-11-13
renderer/Scene: Paint transformation after scene change


lpogic:
If paint is rendered in one scene and then we want to move it to another, its transform will not be updated:

```CPP
auto scene1 = tvg::Scene::gen();
auto scene2 = tvg::Scene::gen();
scene2->translate(100, 100);

auto shape = tvg::Shape::gen();
shape->appendRect(0, 0, 100, 100, 10, 10);
shape->fill(0, 255, 0);

scene1->push(shape);
canvas->push(scene1);
canvas->push(scene2);
canvas->update();

scene1->clear(false);
scene2->push(shape);
canvas->update();
```

Output:
![2](https://github.com/user-attachments/assets/b01e6c7a-a950-4be2-a184-0322b666c032)

Expected output:
![1](https://github.com/user-attachments/assets/2e31bc86-2775-49ed-8571-5489615624f6)

hermet:
@lpogic Hello, besides this issue handling, I would like to recommend using ThorVG in a sequence like this for your optimal performance.

```
1.  paint preparation //paint create & set properties & organize scene tree
2.  canvas->push()  //update
3.  canvas->draw()
```

This is current your usage:
```
1. paint preparation  //paint create & set properties
2. canvas->push() //update
3. paint preparation again //reorganize the scene
4. canvas->update()  //update gain
5. canvas->draw()
```


Recommended code:

```
auto scene1 = tvg::Scene::gen();
auto scene2 = tvg::Scene::gen();
scene2->translate(100, 100);

auto shape = tvg::Shape::gen();
shape->appendRect(0, 0, 100, 100, 10, 10);
shape->fill(0, 255, 0);

scene1->push(shape);

scene1->clear(false);
scene2->push(shape);

canvas->push(scene1);
canvas->push(scene2);
```

lpogic:
@hermet Thank you for suggestion. In my case I am working in an infinite loop:
```
loop {
  # create/modify some paints
  # update nessesary paint tree parts
}
```
The error occurs when I try to transfer paint that has been previously updated between scenes.

In the example, I wanted to more clearly illustrate the bug.

hermet:
@lpogic the fix will be in v0.15.5 thanks.

PR #2962, opened 2024-11-15
renderer: Fix paint transformation after scene change

lpogic:
This PR fixes a bug when paint was moved between scenes and its transformations wasn't updated.

Resolve: #2958

### DIFF
--- a/src/renderer/tvgScene.cpp
+++ b/src/renderer/tvgScene.cpp
@@ -70,7 +70,9 @@ Type Scene::type() const noexcept
 Result Scene::push(Paint* paint) noexcept
 {
     if (!paint) return Result::MemoryCorruption;
-    PP(paint)->ref();
+    auto pPaint = PP(paint);
+    pPaint->ref();
+    pPaint->renderFlag |= RenderUpdateFlag::Transform;
     pImpl->paints.push_back(paint);
 
     return Result::Success;


src/renderer/tvgScene.cpp
```diff
@@ -70,7 +70,9 @@ Type Scene::type() const noexcept
 Result Scene::push(Paint* paint) noexcept
 {
     if (!paint) return Result::MemoryCorruption;
-    PP(paint)->ref();
+    auto pPaint = PP(paint);
+    pPaint->ref();
+    pPaint->renderFlag |= RenderUpdateFlag::Transform;
```
hermet, line 75: I suggest updating the scene directly.
```
    PP(pImpl->scene)->renderFlag |= RenderUpdateFlag::All;
```

hermet (commented):
@lpogic Thanks for you contribution. Please check my comment.

src/renderer/tvgScene.cpp
lpogic: It seems to me that in this case, each time I push paint to the scene it will update all paints in that scene. Are there any situations where such a broad update would be required?

hermet: On second thought, you have a point. We can go with an individual update. Just I've also considered 2 cases more.

1. In case of canvas push.  (we can deal with this separately)
2. To the same scene push/clear repeatedly without any practical updates.

lpogic: 1. With Canvas push the problem did not occur, probably due to the [update call with the force flag](https://github.com/thorvg/thorvg/blob/511495f6d82b20acbe4ac3983a2aae9a3786b76b/src/renderer/tvgCanvas.h#L70). 

2. Maybe a second argument in the push method to raise the update flag? Or a separate method to do this? Automating flag raising can be difficult because the scene may be transformed between clear and push calls.

hermet:
@lpogic by conflict, directly cherry-picked thanks. https://github.com/thorvg/thorvg/commit/75cef8d3a00f3c2d70008b78c1bc01afd57ae2cd

Issue #2964, opened 2024-11-18
lottie/slot: incorrect image slot schema


tinyjin:
Image Slot schema is incorrect, valid slot format to be:

```json
{
  "my_img": {
    "p": {
      "w": 200,
      "h": 300,
      "u": "images/",
      "p": "logo.png",
      "e": 0
    }
  }
}
```

PR #2967, opened 2024-11-18
lottie/slot: Fix incorrect image slot schema

tinyjin:
Issue: #2964


src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1508,7 +1508,12 @@ bool LottieParser::apply(LottieSlot* slot)
             break;
         }
         case LottieProperty::Type::Image: {
-            obj = parseAsset();
+            while (auto key = nextObjectKey()) {
+                if (KEY_AS("p")) {
+                    enterObject();
```
hermet, line 1513: please move enterObject() into parseAsset() again.

hermet (approved):

PR #2977, opened 2024-11-19
gl_engine: support cross compile GL backend into WASM

RuiwenTang:
This PR support cross compile the GL backend code into WASM. The code needs WebGL 2.0 API so, the compile flags contains `MAX_WEBGL_VERSION` and `FULL_ES3` Also add binding code to initialize the WebGL context and GLCanvas.

With this PR, the wasm binary contains WebGL backend rendering, the rest of the work will need to be supported by the thorvg.web project.

issue: #2837

### DIFF
--- a/cross/wasm32_gl.txt
+++ b/cross/wasm32_gl.txt
@@ -1,0 +1,21 @@
+[binaries]
+cpp = 'EMSDK:upstream/emscripten/em++.py'
+ar = 'EMSDK:upstream/emscripten/emar.py'
+strip = '-strip'
+
+[properties]
+root = 'EMSDK:upstream/emscripten/system'
+shared_lib_suffix = 'js'
+static_lib_suffix = 'js'
+shared_module_suffix = 'js'
+exe_suffix = 'js'
+
+[built-in options]
+cpp_args = ['-Wshift-negative-value', '-flto', '-Os', '-fno-exceptions']
+cpp_link_args = ['-Wshift-negative-value', '-flto', '-Os', '-fno-exceptions', '--bind', '-sWASM=1', '-sALLOW_MEMORY_GROWTH=1', '-sEXPORT_ES6=1', '-sFORCE_FILESYSTEM=1', '-sMODULARIZE=1', '-sEXPORTED_RUNTIME_METHODS=FS', '-sMAX_WEBGL_VERSION=2', '-sFULL_ES3']
+
+[host_machine]
+system = 'emscripten'
+cpu_family = 'wasm32'
+cpu = 'wasm32'
+endian = 'little'
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -27,6 +27,9 @@
 #ifdef THORVG_WG_RASTER_SUPPORT
     #include <webgpu/webgpu.h>
 #endif
+#ifdef THORVG_GL_RASTER_SUPPORT
+    #include <emscripten/html5_webgl.h>
+#endif
 
 using namespace emscripten;
 using namespace std;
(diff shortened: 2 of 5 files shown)


src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -155,6 +158,56 @@ struct TvgWgEngine : TvgEngineMethod
...
+        attrs.premultipliedAlpha = true;
+        attrs.failIfMajorPerformanceCaveat = false;
+        attrs.majorVersion = 2;
+        attrs.minorVersion = 0;
+        attrs.enableExtensionsByDefault = true;
+
+
+        context = emscripten_webgl_create_context(selector.c_str(), &attrs);
+
+        if (context == 0) return {};
```
tinyjin, line 191: How about `return nullptr` for clarity?

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -155,6 +158,56 @@ struct TvgWgEngine : TvgEngineMethod
...
+
+        emscripten_webgl_make_context_current(context);
+
+        if(Initializer::init(0, tvg::CanvasEngine::Gl) != Result::Success) return {};
+
+        return GlCanvas::gen();
+    #endif
```
tinyjin, line 198: How about put the line 195, 197 after `#endif` line.
It'd be better to ensure function return regardless of macro. Also for code consistency with the other engines.

Reference: line 149, 150

tinyjin (commented):
Thank you @RuiwenTang, please check comments :)

tinyjin:
## Browser Compatibility Testing Result

I've proceeded browser level testing with WebGL. cc. @hermet 

### Lottie

I noticed some rendering compatibility issues in WebGL. Others in example folder appear to work fine.

1. abstract_circle.json

CleanShot 2024-11-25 at 09 22 10

2. anubis.json

CleanShot 2024-11-25 at 09 22 23

3. card_hover.json

CleanShot 2024-11-25 at 09 22 42

4. cat_loader.json

CleanShot 2024-11-25 at 09 23 35

5. kote.json

CleanShot 2024-11-25 at 09 23 58

6. swinging.json

CleanShot 2024-11-25 at 09 24 14

### SVG

I manually tested all the svg files in the example folder, and they all look good.

CleanShot 2024-11-25 at 09 25 10

cross/wasm32_gl.txt
```diff
@@ -0,0 +1,21 @@
...
+[properties]
+root = 'EMSDK:upstream/emscripten/system'
+shared_lib_suffix = 'js'
+static_lib_suffix = 'js'
+shared_module_suffix = 'js'
+exe_suffix = 'js'
+
+[built-in options]
+cpp_args = ['-Wshift-negative-value', '-flto', '-Os', '-fno-exceptions']
+cpp_link_args = ['-Wshift-negative-value', '-flto', '-Os', '-fno-exceptions', '--bind', '-sWASM=1', '-sALLOW_MEMORY_GROWTH=1', '-sEXPORT_ES6=1', '-sFORCE_FILESYSTEM=1', '-sMODULARIZE=1', '-sEXPORTED_RUNTIME_METHODS=FS', '-sMAX_WEBGL_VERSION=2', '-sFULL_ES3', '-sSTACK_SIZE=4MB']
```
tinyjin, line 15: We used STACK_SIZE option in WebGPU to address stack overflow issue. According to Emscripten, this value must be large enough to avoid such cases, but it should not be unnecessarily increased.

In my opinion, we can go with the default STACK_SIZE. (Remove parameter if there's not specific reason to use)

hermet:
@RuiwenTang @tinyjin Thanks for this. let's address the rendering issue separately -> https://github.com/thorvg/thorvg/issues/2997

cross/wasm32_gl.txt
RuiwenTang: This is copied from the compilation of webgpu. I am not sure if this has any effect.

hermet (approved):

PR #2990, opened 2024-11-22
wg_engine: reset context in examples

SergeyLebedkin:
Fix examples termination process: reset rendering context while window releasing before webgpu instance deleted
Issue https://github.com/thorvg/thorvg/issues/2745

### DIFF
--- a/examples/Example.h
+++ b/examples/Example.h
@@ -425,6 +425,8 @@ struct WgWindow : Window
 
     virtual ~WgWindow()
     {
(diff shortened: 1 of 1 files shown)


hermet (commented):
@SergeyLebedkin No need change in wg renderer side?
No, it will remove all wgpu handles, including objects and context entities while instance is alive

SergeyLebedkin:
> @SergeyLebedkin No need change in wg renderer side?
No, we its will remove all objects in the renderer before the instance will be released

Issue #2989, opened 2024-11-22
lottie: broken animation on thorvg viewer


hermet:
Looks a regression bug, trimpath animation look incorrect on the viewer(1.0-pre5), the local test looks good.

[11483.json](https://github.com/user-attachments/files/17866962/11483.json)

hermet:
+ @tinyjin

PR #2995, opened 2024-11-25
expressions++

mgrudzinska:
Solves the issue for cases where an expression refers to a value that was also calculated using an expression.

Issue: #2989


mgrudzinska:
@hermet It's a draft because in tvgLottieExpressions this is currently applied only to the trim path. This approach can be applied wherever needed, but I'd appreciate your feedback first. I need to store an additional value alongside 'value' - now called 'curValue' and I was wondering whether there is some cheaper sollution

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -257,8 +257,9 @@ struct LottieGenericProperty : LottieProperty
     //Property has an either keyframes or single value.
     Array<LottieScalarFrame<T>>* frames = nullptr;
     T value;
```
hermet, line 259: It might be able to reuse T value.

mgrudzinska: no, value can be used like here:
```
              "o": {
                "a": 0,
                "k": -40,   // this is the value
                "ix": 3,
                "x": "var $bm_rt;\n$bm_rt = $bm_sum($bm_mul(content('Group 1').content('trim control').end, -1.8), value);"
              },
```
as we discussed, I'trry ti move this extra value to expressions

mgrudzinska: @hermet can you please suggest the place where to move this value? I considered 'LottieExpression' but don't succeed. I tried not to template `LottieExpression` - do you think can I consider such an approach?

hermet: @mgrudzinska It’s probably something like this:
```
struct LottieExpression
{
   ...
    union {
        RGB24 color;
        Point point;
        float number;
        uint8_t opacity;
    } output;
    ...
```

```
    T operator()(float frameNo, LottieExpressions* exps)
    {
        if (exps && exp) {
            T& out = reinterpret_cast<T&>(exp->output);
            frameNo = _loop(frames, frameNo, exp);
            if (exps->result<LottieGenericProperty<T>>(frameNo, out, exp)) return out;
        }
        return operator()(frameNo);
    }
```

Ultimately, I don’t think this caching method will serve as a fundamental solution. We need to evaluate the values of the expressions as they are executed in a chain.

mgrudzinska:
@hermet I removed the last changes, since they were wrong. I tried to use union, but I had problem to access the value from expressions.

hermet:
> @hermet I removed the last changes, since they were wrong. I tried to use union, but I had problem to access the value from expressions.

@mgrudzinska It would be nicer, if you can provide a logical explanation of what the problem is. Eitherway, i think the approach seems not good as I mentioned the above.

PR #3002, opened 2024-11-26
wg_engine: separate pipelines ans bind group layouts (refactoring)

SergeyLebedkin:
The main reason of refactoring is to separate bind group sets and pipelines, and change owning of pipelines to compositor
It will be helpfull for resizing (intermidiate changes)
Should not impact to performance and web player
see: Svg example

### DIFF
--- a/src/renderer/wg_engine/tvgWgBindGroups.h
+++ b/src/renderer/wg_engine/tvgWgBindGroups.h
@@ -23,7 +23,7 @@
 #ifndef _TVG_WG_BIND_GROUPS_H_
 #define _TVG_WG_BIND_GROUPS_H_
 
-#include "tvgWgCommon.h"
+#include <webgpu/webgpu.h>
 
 class WgBindGroupLayouts {
 private:
@@ -53,8 +53,8 @@ public:
     void releaseBindGroup(WGPUBindGroup& bindGroup);
     void releaseBindGroupLayout(WGPUBindGroupLayout& bindGroupLayout);
 public:
-    void initialize(WgContext& context);
-    void release(WgContext& context);
+    void initialize(WGPUDevice device);
+    void release();
 };
 
 #endif // _TVG_WG_BIND_GROUPS_H_
--- a/src/renderer/wg_engine/tvgWgBindGroups.cpp
+++ b/src/renderer/wg_engine/tvgWgBindGroups.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "tvgWgBindGroups.h"
+#include <cassert>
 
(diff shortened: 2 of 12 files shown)


tinyjin (commented):
CleanShot 2024-11-27 at 11 21 48

I am still encountering a resizing issue.

hermet:
@tinyjin this is just a refactoring. any regressions (performance & functionality) by this?

tinyjin (approved):
CleanShot 2024-11-27 at 11 46 56@2x

Ah misunderstood. No, looks good

src/renderer/wg_engine/tvgWgBindGroups.h
```diff
@@ -23,7 +23,7 @@
 #ifndef _TVG_WG_BIND_GROUPS_H_
 #define _TVG_WG_BIND_GROUPS_H_
 
-#include "tvgWgCommon.h"
```
hermet, line 26: how about putting the header `#include <webgpu/webgpu.h>` in `tvgWgCommon.h` ?

hermet (approved):

src/renderer/wg_engine/tvgWgBindGroups.h
SergeyLebedkin: @hermet 
_tvgWgBindGroups.h_ also reguired _#include <webgpu/webgpu.h>_
But you still have an access to all webgpu instances if you will include _tvgWgCommon.h_ somewhere 

As _alternalive_, we can get rid tvgWgBindGroups.h and move BindGroup declaration into _tvgWgCommon.h_
its also make a sence

SergeyLebedkin: > @SergeyLebedkin I mean you can include `tvgWgCommon.h`, which could be a common and general header for wg implementation if it include the `webgpu.h`

tvgWgCommon.h use tvgWgBindGroups.h, because tvgWgCommon containe Context, that use BindGroups instance. So context must know about BindGroups, so they must be declared before. We can use forward declatation, but I want to use instalnce access but not pointer to the layouts in context

hermet: @SergeyLebedkin You can merge them as `tvgWgCommon.h`

SergeyLebedkin: > @SergeyLebedkin You can merge them as `tvgWgCommon.h`

Yes, its also make a sence...

hermet: @SergeyLebedkin And it's optional, you can submit patch.

PR #3012, opened 2024-11-28
sw_engine: replaced longjmp with returns

dragostis:
Since longjmp is not yet fully supported on Wasm (emscripten uses JS to make the jumps and requires slow jumps), but also because longjmp is hard to reason about, this patch replaces it with return values. The logic stays exactly the same.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,3 +36,5 @@ Josh Soref <jsoref@gmail.com>
 Elliott Sales de Andrade <quantum.analyst@gmail.com>
 Łukasz Pomietło <oficjalnyadreslukasza@gmail.com>
 Kelly Loh <kelly@lottiefiles.com>
+Dragoș Tiselice <dragos@lottiefiles.com>
+
--- a/src/renderer/sw_engine/tvgSwRle.cpp
+++ b/src/renderer/sw_engine/tvgSwRle.cpp
@@ -188,7 +188,6 @@
 *    http://www.freetype.org
 */
 
-#include <setjmp.h>
 #include <limits.h>
 #include <memory.h>
 #include "tvgSwCommon.h"
@@ -242,8 +241,6 @@ struct RleWorker
 
     int bandSize;
     int bandShoot;
-
-    jmp_buf jmpBuf;
 
     void* buffer;
     long bufferSize;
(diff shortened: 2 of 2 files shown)


mgrudzinska (approved):

dragostis:
Thanks for the review! Since it's my first PR, could you please enable he workflows to run? Also, is there anything else I can do?

hermet (changes requested):
Thanks but some broken occurred after this patch.

Examples/StrokeTrim:
image

Normal:
image

hermet:
> I've tried to reproduce this locally, but it renders fine. Does this reproduce at any resolution or is it resolution-dependent?

@dragostis Hi, I think the issue can be reproduced using the ThorVG example if the size is left unchanged. The problem might be related to system differences if you can't reproduce it. I confirmed that the issue is reproducible on both Linux and Windows but haven’t tested it on macOS.

mgrudzinska:
@dragostis The same happens on mac, but I found the problem.
Please make sure that all `_lineTo` calls are enclosed within an `if` statement: see lines 702 and 740 (the latter is responsible for the issue in the StrokeTrim example).

mgrudzinska:
@dragostis After addressing the above, could you please doublecheck whether the condition `if (rw.cellsCnt >= rw.maxCells)` evaluates to `true` the same number of times before and after the changes?

dragostis:
@mgrudzinska, thanks for the observation! I have no idea how I managed to miss a bunch of calls twice, but I double-checked now and I think it's fine.

hermet (approved):

mgrudzinska:
> @dragostis After addressing the above, could you please doublecheck whether the condition `if (rw.cellsCnt >= rw.maxCells)` evaluates to `true` the same number of times before and after the changes?

@dragostis did you double check this as well?

hermet:
@dragostis Thanks for your awesome contribution, before submitting this, 2 requests here:
1. Add yourself "Dragoș Tiselice <dragos@lottiefiles.com>" to `throvg/AUTHORS` file
image

2. Keep the consistency of the commit header message like:
```
    sw_engine: replaced longjmp with returns

    Since longjmp is not yet fully supported on Wasm (emscripten uses
    JS to make the jumps and requires slow jumps), but also because
    longjmp is hard to reason about, this patch replaces it with
    return values. The logic stays exactly the same.
```

mgrudzinska:
> > @dragostis After addressing the above, could you please doublecheck whether the condition `if (rw.cellsCnt >= rw.maxCells)` evaluates to `true` the same number of times before and after the changes?
> 
> @dragostis did you double check this as well?

sorry for bothering you with this, I checked, it's exactly the same number, perfect!

dragostis:
@hermet, I added myself as an author and fixed the commit title.

@mgrudzinska, no worries at all!

Thank you for helping me land this! :)

Issue #3008, opened 2024-11-28
renderer: prevent file system calls when the web is not allowed to use FS


hermet:
WebAssembly has a 'FILESYSTEM' option for flexible user system environrment, when this option is disabled, ThorVG should avoid making filesystem I/O calls; otherwise, a crash may occur.

PR #3015, opened 2024-11-29
renderer: make the file io configurable.

hermet:
certain systems, may not support file I/O operations. ThorVG should provide users with an option to configure builds according to their requirements.

This ensures that file I/O calls are avoided,
preventing potential crashes.

Please use the meson '-Dfile=true/false' option for this.

issue: https://github.com/thorvg/thorvg/issues/3008

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -116,6 +116,12 @@ if get_option('log')
     config_h.set10('THORVG_LOG_ENABLED', true)
 endif
 
+#File IO
+if get_option('file') == true
+    config_h.set10('THORVG_FILE_IO_SUPPORT', true)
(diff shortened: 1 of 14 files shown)


JSUYA (commented):
TtfLoader still has the file api.

meson.build
```diff
@@ -116,6 +116,12 @@ if get_option('log')
     config_h.set10('THORVG_LOG_ENABLED', true)
 endif
 
+#File IO
+if get_option('file') == true
```
QuLogic, line 120: As noted in #2650, there's no need for the `== true`.

PR #3018, opened 2024-12-02
lottie: add caps support

mgrudzinska:
before:
Zrzut ekranu 2024-12-3 o 00 44 25

after:
Zrzut ekranu 2024-12-3 o 00 44 52

AE:
CAP_AE

sample:
[CAPS_COMPARISON.json](https://github.com/user-attachments/files/17985153/CAPS_COMPARISON.json)

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -1051,12 +1051,24 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
             }
         }
 
+        /* all lowercase letters are converted to uppercase in the "t" text field, making the "ca" value irrelevant, thus AllCaps is nothing to do.
+           So only convert lowercase letters to uppercase (for 'SmallCaps' an extra scaling factor applied) */
+        auto code = p;
+        auto capScale = 1.0f;
+        if (char(p[0]) < 0x80 && doc.caps) {
+            if (*p >= 'a' && *p <= 'z') {
+                char capCode = *p + 'A' - 'a';
+                code = &capCode;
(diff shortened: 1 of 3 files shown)


mgrudzinska:
comment regarding resources:
- AllCaps - AE changes the 'text' field and the 'ca' value is not important any more ('some text' -> 'SOME TEXT')
- SmallCaps - in the exported file only the lower case characters are included, so there are no data of the upper case chars in the file (unless the file is prepared as in the example, in which the text contains already the necessary chars)

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1053,12 +1053,21 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
             }
         }
 
+        auto pp = *p;
```
hermet, line 1056: updateText() is horrible now. At least, it is necessary to add appropriate comments in the right places so that readers can understand the brief context flow.

mgrudzinska: I'd rather avoid refactoring updateText in this pr if possible. but I know it's has to be done at some point. added comments

src/loaders/lottie/tvgLottieCommon.h
```diff
@@ -68,6 +68,7 @@ struct TextDocument
     float size;
     float tracking = 0.0f;
     uint8_t justify;
+    uint8_t caps;
```
hermet, line 71: It seems to need an initial value of 0?

hermet:
@mgrudzinska boom, there seems a side effect. Please review this:

[7725.json](https://github.com/user-attachments/files/18008351/7725.json)

before / after
image

PR #3023, opened 2024-12-03
wasm: fix webgl context switching issue in multi-canvas

tinyjin:
Previously, rendering multiple canvases with separate WebGL contexts caused unexpected errors due to incorrect context activation. This patch ensures that `emscripten_webgl_make_context_current` is called before every draw operation.

[Main]
There are a few issues in rendering due to invalid context.

CleanShot 2024-12-04 at 03 16 06
CleanShot 2024-12-04 at 03 16 14

[Patched]
CleanShot 2024-12-04 at 03 06 22

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -83,6 +83,7 @@ struct TvgEngineMethod
     virtual ~TvgEngineMethod() {}
     virtual Canvas* init(string&) = 0;
     virtual void resize(Canvas* canvas, int w, int h) = 0;
+    virtual void update() = 0;
     virtual val output(int w, int h)
     {
         return val(typed_memory_view<uint8_t>(0, nullptr));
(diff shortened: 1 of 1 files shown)


RuiwenTang (approved):
LGTM

hermet:
@RuiwenTang I think ThorVG GL needs to hold gl context through the target() api, and the GL renderer must ensure that the context is made current inside before it begins the drawing context.

hermet:
@tinyjin please hold this

RuiwenTang:
> @RuiwenTang I think ThorVG GL needs to hold gl context through the target() api, and the GL renderer must ensure that the context is made current inside before it begins the drawing context.

If we compare it to WebGPU, the GlCanvas::target function only determines the Surface that the Canvas needs to draw, which is more like WGPUSurface. GLContext is similar to creating WGPUAdaptor and WGPUDevice, and the initialization logic of each platform is different. I think it is more suitable to be placed in the binding layer.

hermet:
@RuiwenTang Imagine applications, it's not safe. The user has no understanding of the context requirements under the GL canvas. Canvas can abstracts it.

hermet:
https://github.com/thorvg/thorvg/pull/3031

Issue #3009, opened 2024-11-28
lottie: Memory Access Out of Bounds


theashraf:
This issue occurs in dotLottie-web when rendering on a canvas with specific dimensions. In this case, the canvas has a width of 275 and a height of 137, but it can also occur with other canvas sizes.

Related issue: https://github.com/LottieFiles/dotlottie-web/issues/376

Error Log:
Attached below is the full error log with the stack trace:
```
dlmalloc.c:4562 Uncaught RuntimeError: memory access out of bounds
    at DotLottiePlayer.wasm.emscripten_builtin_malloc (http://127.0.0.1:5500/web-example/release/wasm/DotLottiePlayer.wasm:wasm-function[3057]:0xe3e49)
    at DotLottiePlayer.wasm.dlrealloc (http://127.0.0.1:5500/web-example/release/wasm/DotLottiePlayer.wasm:wasm-function[3060]:0xe562b)
    at DotLottiePlayer.wasm._horizLine(RleWorker&, long, long, long, long) (http://127.0.0.1:5500/web-example/release/wasm/DotLottiePlayer.wasm:wasm-function[4923]:0x1165d8)
    at DotLottiePlayer.wasm.rleRender(SwRle*, SwOutline const*, SwBBox const&, bool) (http://127.0.0.1:5500/web-example/release/wasm/DotLottiePlayer.wasm:wasm-function[4805]:0x10d9e7)
    at DotLottiePlayer.wasm.SwShapeTask::run(unsigned int) (http://127.0.0.1:5500/web-example/release/wasm/DotLottiePlayer.wasm:wasm-function[4812]:0x10dde1)
    at DotLottiePlayer.wasm.tvg::TaskScheduler::request(tvg::Task*) (http://127.0.0.1:5500/web-example/release/wasm/DotLottiePlayer.wasm:wasm-function[4791]:0x10cb92)
    at DotLottiePlayer.wasm.tvg::SwRenderer::prepareCommon(SwTask*, tvg::Matrix const&, tvg::Array<void*> const&, unsigned char, tvg::RenderUpdateFlag) (http://127.0.0.1:5500/web-example/release/wasm/DotLottiePlayer.wasm:wasm-function[4731]:0x105a8e)
[...]

Issue #2984, opened 2024-11-21
sw_engine: heap-buffer-overflow by effects


hermet:
on macOS (intel) 
scimitar-anim.svg 1200x1200

> ==48760==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x00012a695554 at pc 0x00010f77a4f8 bp 0x7ff7b1174260 sp 0x7ff7b1174258
> WRITE of size 4 at 0x00012a695554 thread T0
>     #0 0x10f77a4f7 in rasterXYFlip(unsigned int*, unsigned int*, int, int, int, SwBBox const&, bool) tvgSwRaster.cpp:1845
>     #1 0x10f75c359 in effectGaussianBlur(SwCompositor*, SwSurface*, tvg::RenderEffectGaussianBlur const*, bool) tvgSwPostEffect.cpp:205
>     #2 0x10f785e7b in tvg::SwRenderer::effect(tvg::RenderCompositor*, tvg::RenderEffect const*, bool) tvgSwRenderer.cpp:661
>     #3 0x10f731310 in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:145
>     #4 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
>     #5 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
>     #6 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
>     #7 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
>     #8 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
>     #9 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
>     #10 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
>     #11 0x10f7311ff in tvg::Scene::Impl::render(tvg::RenderMethod*) tvgScene.h:137
>     #12 0x10f7301dc in tvg::Paint::Impl::render(tvg::RenderMethod*) tvgPaint.cpp:228
>     #13 0x10f73add6 in tvg::Picture::Impl::render(tvg::RenderMethod*) tvgPicture.cpp:84
[...]

PR #3026, opened 2024-12-04
common: improve the rendering pipeline

hermet:
enhanced the rendering composition target to better support features such as alpha blending, blending, masking, post effects, and more. This allows rasterizers to prepare the composition context with more precise control.

Additionally, resolved a crash in the software engine's post-effect process caused by a buffer size mismatch during XY buffer flipping.

issue: https://github.com/thorvg/thorvg/issues/3009
issue: https://github.com/thorvg/thorvg/issues/2984


hermet:
@mgrudzinska could you please confirm that crash is fixed with this? https://github.com/thorvg/thorvg/issues/2984

mgrudzinska:
> @mgrudzinska could you please confirm that crash is fixed with this? #2984

yes, I confirm that this resolved all asan error for effects

PR #3034, opened 2024-12-09
wasm: remove ASYNCIFY to reduce binary size

tinyjin:
The binary size recently increased due to the ASYNCIFY option, which was required for WebGPU initialization. 

To prevent unnecessary binary size growth, ThorVG will no longer depend on asynchronous processes such as `emscripten_sleep` (ASYNCIFY or JSPI).

As a result, the combined WASM binary size has been significantly reduced to less than 1MB.

Size comparison: 1559KB → 998KB (-36%)

CleanShot 2024-12-09 at 19 59 51@2x

CleanShot 2024-12-09 at 19 55 19@2x

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -43,12 +43,15 @@ static const char* NoError = "None";
     static WGPUDevice device{};
     static bool adapterRequested = false;
     static bool deviceRequested = false;
+    static bool initializationFailed = false;
 #endif
 
 // 0: success, 1: fail, 2: wait for async request
 int init()
 {
 #ifdef THORVG_WG_RASTER_SUPPORT
+    if (initializationFailed) return 1;
+
     //Init WebGPU
     if (!instance) instance = wgpuCreateInstance(nullptr);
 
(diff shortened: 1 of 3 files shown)


hermet (commented):
@tinyjin What is the user supposed to do when init() returns false?

hermet:
https://github.com/thorvg/thorvg.web/pull/80

tinyjin:
> @tinyjin What is the user supposed to do when init() returns false?

Call `wait(100)` function in JS side instead of `emscripten_sleep(100)`

hermet:
> > @tinyjin What is the user supposed to do when init() returns false?
> 
> Call `wait(100)` function in JS side instead of `emscripten_sleep(100)`
> 
> `false` means initialization is not yet ready -> should wait until it gets done.

@tinyjin User should use an infinite loop for initialization. How to handle if adapter or device acquisition fails?

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -41,32 +41,44 @@ static const char* NoError = "None";
     static WGPUInstance instance{};
     static WGPUAdapter adapter{};
     static WGPUDevice device{};
+    static bool adapterRequested = false;
+    static bool deviceRequested = false;
 #endif
 
-void init()
+bool init()
```
hermet, line 48: int // 0: success, 1: fail, 2: try again

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -41,32 +41,60 @@ static const char* NoError = "None";
...
 
-void init()
+// 0: success, 1: fail, 2: wait for async request
+int init()
 {
 #ifdef THORVG_WG_RASTER_SUPPORT
     //Init WebGPU
     if (!instance) instance = wgpuCreateInstance(nullptr);
 
+    if (initializationFailed) return 1;
```
hermet, line 56: this should be at 53 line?

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -41,32 +41,60 @@ static const char* NoError = "None";
...
+
     // request adapter
     if (!adapter) {
+        if (adapterRequested) return 2;
+
         const WGPURequestAdapterOptions requestAdapterOptions { .nextInChain = nullptr, .powerPreference = WGPUPowerPreference_HighPerformance, .forceFallbackAdapter = false };
-        auto onAdapterRequestEnded = [](WGPURequestAdapterStatus status, WGPUAdapter adapter, char const * message, void * pUserData) { *((WGPUAdapter*)pUserData) = adapter; };
```
hermet, line 55: coding convention:
-> `char const* message, void* pUserData`

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -41,32 +41,60 @@ static const char* NoError = "None";
...
+        auto onAdapterRequestEnded = [](WGPURequestAdapterStatus status, WGPUAdapter adapter, char const * message, void * pUserData) {
```
hermet, line 63: coding convention:
-> `char const* message, void* pUserData`

hermet (approved):
please double-check the comments and push yourself. thanks.

PR #3036, opened 2024-12-09
API: Replace the Canvas::clear() API with Canvas::remove() & draw()

hermet:
Added a `clear` parameter to Canvas::draw(), allowing users to decide whether to clear the target buffer before drawing.

To remove the paints from a canvas, please use Canvas::remove()

C++ API Removals:
- Result Canvas::clear(bool paints, bool buffer)

C++ API Modifications:
- Result Canvas::draw() -> Result Canvas::draw(bool clear)

C API Removals:
- Tvg_Result tvg_canvas_clear(bool paints, bool buffer)

C API Modifications:
- Tvg_Result tvg_canvas_draw(Tvg_Canvas* canvas) -> Tvg_Result tvg_canvas_draw(Tvg_Canvas* canvas, bool clear)

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.h
+++ b/src/renderer/gl_engine/tvgGlRenderer.h
@@ -133,7 +133,7 @@ private:
         Key key;
     } mDisposed;
 
-    bool mClearBuffer = true;
+    bool mClearBuffer = true;  //FIXME: clear buffer should be optional (default is false)
 
     BlendMethod mBlendMethod = BlendMethod::Normal;
 };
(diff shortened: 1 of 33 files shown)


src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -239,6 +239,7 @@ const RenderSurface* WgRenderer::mainSurface()
 
 bool WgRenderer::clear()
 {
+    //TODO: clear the current target buffer only if clear() is called
```
hermet, line 242: @SergeyLebedkin

src/renderer/gl_engine/tvgGlRenderer.h
```diff
@@ -133,7 +133,7 @@ class GlRenderer : public RenderMethod
         Key key;
     } mDisposed;
 
-    bool mClearBuffer = true;
+    bool mClearBuffer = true;  //FIXME: clear buffer should be optional (default is false)
```
hermet, line 136: @RuiwenTang It is not always necessary to clear the buffer if the scene has solid background content. The user can choose whether to do so.

PR #3048, opened 2024-12-13
lottie: handle image data while loaders are inactive

mgrudzinska:
Handle cases where the lottie file contains image data, but the appropriate loaders are not active - the image object is removed to exclude it from further rendering process.

Note:
This issue became noticeable after commit 0ebbc61. Previously,  if at least one element of the canvas was successfully rendered (```paint->pImpl->render(renderer)```), the operation returned success. Currently, for the scene to return success (```PP(scene)->render(renderer)```), all its elements must be rendered successfully.

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -977,6 +977,7 @@ void LottieBuilder::updateSolid(LottieLayer* layer)
 
 void LottieBuilder::updateImage(LottieGroup* layer)
 {
+    if (layer->children.empty()) return;
     auto image = static_cast<LottieImage*>(layer->children.first());
     layer->scene->push(image->pooling(true));
 }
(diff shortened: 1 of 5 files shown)


hermet:
@mgrudzinska can you suggest a sample to reproduce the issue?

hermet:
> fun_animation.json from our samples

@mgrudzinska  Hmm.. I couldn't see any issue. Could you please point out the problem?

https://github.com/user-attachments/assets/73808e8b-ae34-44a5-b75e-9ec09aa2d1b3

mgrudzinska:
> > fun_animation.json from our samples
> 
> @mgrudzinska Hmm.. I couldn't see any issue. Could you please point out the problem?

you build with the png loader, right? the problem is when you disable it

hermet (changes requested):
Ok, it seems to be a matter of how to handle exceptional cases — whether to fail during loading or to proceed with rendering in an incomplete state. Eventually, the ultimate solution is to enable the necessary loaders. I don't think this handling doesn't useful for users. If there are any practical benefits or helpful aspects that I’m missing, please let me know.

PR #3050, opened 2024-12-13
sw_engine: prevent runtime error

mgrudzinska:
The sqrt function occasionally received a negative value as an argument, resulting in a nan value being propagated to further calculations.
This is now prevented by adding a check, taking advantage of the rarity of negative values for branch prediction optimization.

Note:
error occurred on mac os(intel) GradientTransform sample


hermet:
Adding exception handling for each pixel is costly. Need to carefully find a better solution.

hermet:
@mgrudzinska let's keep this if the issue is valid.

Issue #2854, opened 2024-10-14
Stroke Trimming: Refactoring


hermet:
It is necessary to review the current implementation of Stroke Trimming. Through refactoring, we can determine whether it can be improved into a unified logic. The initial direction was to implement each backend engine in the most optimal way on its own, without damaging the input `RenderShape` data. 

We can consider the following options:

1. Spec out the **_Stroke Trimming_** in the canvas engine and place the logic within the Lottie loader, similar to other Shape Modifiers.
2. Generalize **_Stroke Trimming_** at the common level as a helper function, allowing lower backend engines to utilize it.

hermet:
@RuiwenTang  @SergeyLebedkin The next figure demonstrates the second approach. Please count how many times the path data will be copied for gl/wg engines (2 or 3?). If more than 2 copies occur, we might wanna discard this approach.

![image](https://github.com/user-attachments/assets/652b6a28-a536-4cf0-996b-9204b7506750)

RuiwenTang:
At least in GLEngine, only need to read the last Path data.
Therefore, I feel it is more convenient for me to provide a function like 
```
some_function(const Array<PathCommand>& cmd, const Array<Point>>& p, float start, float end) -> ( Array<PathCommand>, Array<Point> )
``` 
Or can tvgShape complete the work of PathTrim and directly pass the Trimmed RenderShape to the rendering backend?

SergeyLebedkin:
> @RuiwenTang @SergeyLebedkin The next figure demonstrates the second approach. Let's count how many times the path data will be copied for gl/wg/sw engines (2 or 3?). If more than 2 copies occur, we might wanna discard the no.2 approach.

It will be nice to same approach and realization for path triming (simult and not simult) and share this for all render apis.
It can be handled in case, if we will separate Path as a standalone entity

SergeyLebedkin:
> If more than 2 copies occur

And also dashing, but thats optional

hermet:
> Therefore, I feel it is more convenient for me to provide a function like

Please count how many data copy is necessary at the end of gl engine.

> Or can tvgShape complete the work of PathTrim and directly pass the Trimmed RenderShape to the rendering backend?

RenderShape should not modify the input data. It should retain the user's input as origin. Approach No.1 could addresses this, as the LottieLoader will modify the path data in advance and then push it as input.

mgrudzinska:
[...]

Issue #2384, opened 2024-06-10
lottie: support for simultaneously trimmed and dashed stroke


mgrudzinska:
https://github.com/LottieFiles/dotlottie-web/issues/239

hermet:
see also: https://github.com/thorvg/thorvg/issues/1643

hermet:
@mgrudzinska please assign yourself when you start the task.

PR #3056, opened 2024-12-17
lottie: support filled/dashed trim path

mgrudzinska:
@issue: https://github.com/thorvg/thorvg/issues/2384
@issue: https://github.com/thorvg/thorvg/issues/2854

ball

fyrtki

TODO: improve connection point (top middle point in the animation below)
dashedTrim2

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -858,18 +858,21 @@ void LottieBuilder::updateTrimpath(TVG_UNUSED LottieGroup* parent, LottieObject*
 {
     auto trimpath = static_cast<LottieTrimpath*>(*child);
 
-    float begin, end;
-    trimpath->segment(frameNo, begin, end, exps);
+    float start, end;
+    if (!trimpath->segment(frameNo, start, end, exps)) return;
+    bool simultaneous = trimpath->type == LottieTrimpath::Type::Simultaneous;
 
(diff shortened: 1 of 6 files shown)


mgrudzinska:
1. now the logic is moved to the lottie loader - if we agree on that, should the strokeTrim api remain valid?
2. for now aa won;t be applied for fill+stroke, since the sw engine does not know about trimming
3. for ```trim type = individual``` it is possible to apply trim while the shape is created, like other shape modifiers, but for ```trim type = simultaneous``` it has to be done after all shapes are created (need to know the total length) - can be done but first we ahve to agree whether the logic will be moved to the lottie loaderr

hermet:
> * now the logic is moved to the lottie loader - if we agree on that, should the strokeTrim api remain valid?
> * for now aa won;t be applied for fill+stroke, since the sw engine does not know about trimming
> * for `trim type = individual` it is possible to apply trim while the shape is created, like other shape modifiers, but for `trim type = simultaneous` it has to be done after all shapes are created (need to know the total length) - can be done but first we ahve to agree whether the logic will be moved to the lottie loaderr

1. Not recommended. Trim Path can be useful for general motion graphics, not just for Lottie. We can only consider it as a last resort if no other solutions are available.
2. I think it's not a big deal,
3. Question: how does the current stroking trim path implementation determine the total length simultaneously? In other words, why can't the fill approach achieve this? Please propose a logical explanation.

mgrudzinska:
https://github.com/thorvg/thorvg/issues/3118

PR #3059, opened 2024-12-18
lottie: revise the masking logic

hermet:
Allow the masking data even though they were mask None mode. Those will be used by the layer stroke effect.


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1224,63 +1224,65 @@ void LottieBuilder::updateMaskings(LottieLayer* layer, float frameNo)
...
             pShape = shape;
-            pMethod = method;
         }
+
+        //Default Masking
+        if (expand == 0.0f) {
+            mask->pathset(frameNo, SHAPE(pShape)->rs.path.cmds, SHAPE(pShape)->rs.path.pts, nullptr, nullptr, nullptr, exps);
+        //Masking with Expansion (Offset)
+        } else {
+            //TODO: Once path direction support is implemented, ensure that the direction is ignored here
```
hermet, line 1270: @mgrudzinska please confirm this, the offset could be applied to all subsequent masking shapes. (before you only applied it to the first masking shape)

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1224,63 +1224,64 @@ void LottieBuilder::updateMaskings(LottieLayer* layer, float frameNo)
...
+            auto offset = LottieOffsetModifier(mask->expand(frameNo));
```
mgrudzinska, line 1270: ```auto offset = LottieOffsetModifier(expand);```

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: exactly the same :)
before only one of the masks was expanding

AE:
multiMask_AE

PR:
multimask_PR

sample:
[mask.json](https://github.com/user-attachments/files/18181394/mask.json)

mgrudzinska (approved):

PR #3062, opened 2024-12-19
common: filled/dashed trimming support

mgrudzinska:
Dec-19-2024 03-12-25
Zrzut ekranu 2024-12-19 o 03 14 33

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ The following list shows primitives that are supported by ThorVG: <br />
 <br />
  
 - **Lines & Shapes**: rectangles, circles, and paths with coordinate control
-- **Filling**: solid colors, linear & radial gradients, and path clipping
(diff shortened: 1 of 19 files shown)


hermet:
@mgrudzinska Please consider the potential drawbacks or downsides of having separate trim path functions for fill and stroke. Alternatively, a common implementation that supports both could be more useful for a wider range of user scenarios.

hermet:
@mgrudzinska Please consider to update the GL/WG engine as well with separate commits in the same PR.

hermet:
@mgrudzinska Let's apply the enhancement incrementally. Could you please apply the refactoring first?
-> `Generalize Stroke Trimming at the common level as a helper function, allowing render backends to utilize it.`

mgrudzinska:
https://github.com/thorvg/thorvg/issues/3118

Issue #3065, opened 2024-12-19
Renderer: Ensure short paths with stroke width are rendered


baszczewski:
ThorVG currently skips rendering paths with very short or zero-length segments. While this behavior may sometimes be justified, it should not occur when attributes such as ```stroke-width``` and ```stroke-linecap``` are applied.

This omission causes ThorVG to display SVG files or Lottie animations differently from other widely used tools, such as Google Chrome, Firefox, Inkscape, Figma, lottie-web, or Image Viewer (from the GNOME Desktop Environment).

### Observations

- The issue occurs across different rendering engines, including SW and GL. I have not tested it with WG, but similar behavior is likely expected.
- Using zero-length paths is a common and desirable technique in animation. Combined with stroke-width and stroke-linecap (e.g., round), it enables effects where a line is reduced to a dot at specific points in an animation.

### Example Use Case

In our ```Lordicon``` project, hundreds of icons rely on this technique. Removing this functionality would significantly complicate the animation preparation process, which is not a viable option.

### Attachments

Simply compare how the files below are rendered in browsers or in lottie-web versus in ThorVG.

**SVG:**

![simple](https://github.com/user-attachments/assets/0b46649d-998f-49d1-bed5-d331d45a08d0)

**Test SVG:**

![test](https://github.com/user-attachments/assets/2b265d99-d529-494b-a069-c34c61dcfe6c)

**Test Lottie:**

[test.json](https://github.com/user-attachments/files/18196150/test.json)

hermet:
\+ @RuiwenTang

hermet:
@baszczewski Thanks for reporting. Just for your information, you can quickly test resources using [thorvg viewer](https://www.thorvg.org/viewer)

![image](https://github.com/user-attachments/assets/bff97a3e-227a-469d-90cd-74fdcfeab2c4)

hermet:
related: https://github.com/thorvg/thorvg/issues/2094

hermet:
@baszczewski @mgrudzinska Thanks, the fix will be in v0.15.8.

PR #3066, opened 2024-12-19
sw_engine: Fix for rendering short paths

baszczewski:
I updated ```tvgSwStroke.cpp``` to ensure it does not terminate prematurely for paths with a step of 0 or exceptionally small values when a valid stroke-width is present.

As part of the examples, I introduced a program called ```./StrokeShort```, which demonstrates the correct rendering of lines in such cases with various ```stroke-cap``` settings.

image

I also added an SVG file (stroke-short.svg) to the resources, which can be viewed using the ```./Svg``` program. It now renders correctly.

Issues: https://github.com/thorvg/thorvg/issues/3065

### DIFF
--- a/src/renderer/sw_engine/tvgSwStroke.cpp
+++ b/src/renderer/sw_engine/tvgSwStroke.cpp
@@ -374,8 +374,8 @@ static void _lineTo(SwStroke& stroke, const SwPoint& to)
 {
     auto delta = to - stroke.center;
 
-    //a zero-length lineto is a no-op; avoid creating a spurious corner
-    if (delta.zero()) return;
+    //a zero-length lineto is a no-op if stroke width is 0; avoid creating a spurious corner
+    if (delta.zero() && stroke.width == 0) return;
 
     /* The lineLength is used to determine the intersection of strokes outlines.
        The scale needs to be reverted since the stroke width has not been scaled.
@@ -451,8 +451,8 @@ static void _cubicTo(SwStroke& stroke, const SwPoint& ctrl1, const SwPoint& ctrl
             continue;
         }
 
-        //ignoreable size
-        if (valid < 0 && arc == bezStack) {
+        //ignoreable size if stroke width is 0
+        if (valid < 0 && arc == bezStack && stroke.width == 0) {
             stroke.center = to;
             return;
         }
--- a/examples/StrokeShort.cpp
+++ b/examples/StrokeShort.cpp
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2020 - 2024 the ThorVG project. All rights reserved.
+
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "Example.h"
+
+/************************************************************************/
+/* ThorVG Drawing Contents                                              */
+/************************************************************************/
+
+struct UserExample : tvgexam::Example
+{
+    bool content(tvg::Canvas* canvas, uint32_t w, uint32_t h) override
+    {
+        if (!canvas) return false;
+
+        //Line 1 (0 length and available stroke width - draw circle 25x25)
+        auto line1 = tvg::Shape::gen();
+        line1->moveTo(150, 150);
+        line1->lineTo(150, 150);
+        line1->strokeFill(255, 255, 255);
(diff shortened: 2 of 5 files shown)


hermet:
@baszczewski Thank you for your contribution. That would be nicer if you can send a PR with only engine fix in this case. The examples in ThorVG are not specifically designed for feature-testing but rather serve as a user guide, providing hints on how to use ThorVG through various samples.

baszczewski:
@hermet  Of course, I have removed the unnecessary example. I hope this small commit can be helpful.

src/renderer/sw_engine/tvgSwStroke.cpp
```diff
@@ -374,8 +374,8 @@ static void _lineTo(SwStroke& stroke, const SwPoint& to)
 {
     auto delta = to - stroke.center;
 
-    //a zero-length lineto is a no-op; avoid creating a spurious corner
-    if (delta.zero()) return;
+    //a zero-length lineto is a no-op if stroke width is 0; avoid creating a spurious corner
+    if (delta.zero() && stroke.width == 0) return;
```
mgrudzinska, line 378: In this place (also in l.454), ```stroke.width == 0``` is never true because it was already checked [here](https://github.com/thorvg/thorvg/blob/285d1427ad823e91c7f3915629a50c4376d84759/src/renderer/sw_engine/tvgSwRenderer.cpp#L96)

What do you think about adding the following check in the ```strokeParseOutline()``` function:
```if (!nonZeroLen && stroke->cap != StrokeCap::Butt)```  
for each contour, i.e., each segment starting with M - (after 'close:' in l.879)
If such a condition is met, it would mean that the cap should be drawn. 
I did some tests by calling the ```_firstSubPath()``` function if the above is true, but it has to be double checked

Comment (mgrudzinska) [reply src/renderer/sw_engine/tvgSwStroke.cpp:378]: ```nonZeroLen``` is the logical sum of the booleans returned from _lineTo and _cubicTo :
```
nonZeroLen |= _lineTo(*stroke, *pt);
nonZeroLen |= _cubicTo(...)
```

Comment (baszczewski) [reply src/renderer/sw_engine/tvgSwStroke.cpp:378]: Could you please suggest how exactly to achieve this? Unfortunately, I’m not familiar enough with the project to accomplish it without issues.

The proposed _lineTo function in tvgSwStroke.cpp does not return a bool, while its variant that does is located in tvgSwRle.cpp. However, it requires a parameter of type RleWorker, which, unfortunately, I’m not familiar with.

Comment (baszczewski) [reply src/renderer/sw_engine/tvgSwStroke.cpp:378]: If I try this kind of hacky solution to determine nonZeroLength:

```c++
close:  
    auto nonZeroLength = start != stroke->center;  
    if (!nonZeroLength && stroke->cap != StrokeCap::Butt)  
        _firstSubPath(*stroke, stroke->angleIn, 0);  
    if (!stroke->firstPt)  
        _endSubPath(*stroke);  
    first = last + 1;  
}
```

Some minimal examples display correctly, while others do not.

image
image

mgrudzinska: Can you try this?:
```
        auto drawn = false;
        while (pt < limit) {
            //emit a single line_to
            if (types[0] == SW_CURVE_TYPE_POINT) {
                ++pt;
                ++types;
                drawn |= _lineTo(*stroke, *pt);
            //types cubic
            } else {
                pt += 3;
                types += 3;
                if (pt <= limit) drawn |= _cubicTo(*stroke, pt[-2], pt[-1], pt[0]);
                else if (pt - 1 == limit) drawn |= _cubicTo(*stroke, pt[-2], pt[-1], start);
                else goto close;
            }
        }
    close:
        if (!drawn && stroke->cap != StrokeCap::Butt) _firstSubPath(*stroke, 0, 0);
        if (!stroke->firstPt) _endSubPath(*stroke);
        first = last + 1;
```
this is part of the ```strokeParseOutline``` fun. 

you also have to change _lineTo and _cubicTo (from the tvgSwStroke.cpp) to return bool - in the same places you edited before:

in _cubicTo:
```
        //ignoreable size
        if (valid < 0 && arc == bezStack) {
            stroke.center = to;
            return false;
        }
```
otherwise return true

and in _lineTo:
```
    if (delta.zero()) return false;
```
otherwise return true

please let me know if this doesn't help and send a sample if possible that is problematic (dashed and trimmed stroke is a different story, I can solve it after we handle this issue)

baszczewski: WOW. After these changes, it seems that the problem is solved. Thank you so much for the support.
Can we thoroughly test the implemented changes and merge them into the main branch of the project?

mgrudzinska (approved):

src/renderer/sw_engine/tvgSwStroke.cpp
```diff
@@ -370,12 +370,13 @@ void _firstSubPath(SwStroke& stroke, SwFixed startAngle, SwFixed lineLength)
...
 
 
-static void _lineTo(SwStroke& stroke, const SwPoint& to)
+static bool _lineTo(SwStroke& stroke, const SwPoint& to)
 {
     auto delta = to - stroke.center;
 
     //a zero-length lineto is a no-op; avoid creating a spurious corner
-    if (delta.zero()) return;
+    if (delta.zero())
```
hermet, line 378: We can provide the de-coupled logic like this. (same to cubic to):
```
    if (delta.zero()) {
        //Round and square caps are expected to be drawn as a dot even for zero-length lines.
        if (stroke.firstPt && stroke.cap != StrokeCap::Butt) _firstSubPath(stroke, 0, 0);
        return;
    }
```

src/renderer/sw_engine/tvgSwStroke.cpp
```diff
@@ -861,25 +866,28 @@ bool strokeParseOutline(SwStroke* stroke, const SwOutline& outline)
...
+                else if (pt - 1 == limit) drawn |= _cubicTo(*stroke, pt[-2], pt[-1], start);
                 else goto close;
             }
         }
     close:
+        if (!drawn && stroke->cap != StrokeCap::Butt) _firstSubPath(*stroke, 0, 0);
         if (!stroke->firstPt) _endSubPath(*stroke);
         first = last + 1;
     }
+
```
hermet, line 890: please do not change unnecessary lines.

hermet:
@baszczewski hello, many thanks. could you please check these? (otherwise, you can leave them to me.)
1. you can double-check my comments. 
2. add yourself -> https://github.com/thorvg/thorvg/blob/main/AUTHORS
3. squash commits to one.

src/renderer/sw_engine/tvgSwStroke.cpp
baszczewski: Very good idea. Thank you.

baszczewski:
@hermet Thank you for your support on this matter. I have implemented your suggestions.

hermet (approved):

PR #3068, opened 2024-12-20
gl_engine: Fix for rendering short paths

baszczewski:
As part of this request, I updated ```tvgGlTessellator.cpp``` and ```tvgGlTessellator.h``` to ensure they do not terminate prematurely for paths with a step of 0 or exceptionally small values when a valid ```stroke-width``` is present.

In my opinion, the optimal approach was to separate vertex generation into dedicated methods: ```strokeRoundPoint``` and ```strokeSquarePoint```. My update supports two different ```stroke-cap``` styles. I have also tested it with various files (JSON, SVG) as well as a small example application similar to the one included in the previous pull request (https://github.com/thorvg/thorvg/pull/3066).

Issues: https://github.com/thorvg/thorvg/issues/3065

### DIFF
--- a/src/renderer/gl_engine/tvgGlTessellator.cpp
+++ b/src/renderer/gl_engine/tvgGlTessellator.cpp
@@ -1655,38 +1655,42 @@ void Stroker::doStroke(const PathCommand *cmds, uint32_t cmd_count, const Point 
     mResGlPoints->reserve(pts_count * 4 + 16);
     mResIndices->reserve(pts_count * 3);
 
+    auto validStrokeCap = false;
+
     for (uint32_t i = 0; i < cmd_count; i++) {
         switch (cmds[i]) {
             case PathCommand::MoveTo: {
-                if (mStrokeState.hasMove) {
+                if (validStrokeCap) { // check this, so we can skip if path only contains move instruction
                     strokeCap();
-                    mStrokeState.hasMove = false;
+                    validStrokeCap = false;
                 }
-                mStrokeState.hasMove = true;
                 mStrokeState.firstPt = *pts;
                 mStrokeState.firstPtDir = {0.0f, 0.0f};
                 mStrokeState.prevPt = *pts;
                 mStrokeState.prevPtDir = {0.0f, 0.0f};
                 pts++;
+                validStrokeCap = false;
             } break;
             case PathCommand::LineTo: {
+                validStrokeCap = true;
                 this->strokeLineTo(*pts);
                 pts++;
             } break;
             case PathCommand::CubicTo: {
+                validStrokeCap = true;
                 this->strokeCubicTo(pts[0], pts[1], pts[2]);
                 pts += 3;
             } break;
             case PathCommand::Close: {
                 this->strokeClose();
 
-                mStrokeState.hasMove = false;
+                validStrokeCap = false;
             } break;
             default:
                 break;
         }
     }
-    strokeCap();
+    if (validStrokeCap) strokeCap();
 }
 
 
@@ -1707,15 +1711,20 @@ void Stroker::doDashStroke(const PathCommand *cmds, uint32_t cmd_count, const Po
 
 void Stroker::strokeCap()
 {
-    if (mStrokeState.firstPt == mStrokeState.prevPt) return;
     if (mStrokeCap == StrokeCap::Butt) return;
 
     if (mStrokeCap == StrokeCap::Square) {
-        strokeSquare(mStrokeState.firstPt, {-mStrokeState.firstPtDir.x, -mStrokeState.firstPtDir.y});
-        strokeSquare(mStrokeState.prevPt, mStrokeState.prevPtDir);
+        if (mStrokeState.firstPt == mStrokeState.prevPt) strokeSquarePoint(mStrokeState.firstPt);
+        else {
+            strokeSquare(mStrokeState.firstPt, {-mStrokeState.firstPtDir.x, -mStrokeState.firstPtDir.y});
+            strokeSquare(mStrokeState.prevPt, mStrokeState.prevPtDir);
+        }
     } else if (mStrokeCap == StrokeCap::Round) {
-        strokeRound(mStrokeState.firstPt, {-mStrokeState.firstPtDir.x, -mStrokeState.firstPtDir.y});
-        strokeRound(mStrokeState.prevPt, mStrokeState.prevPtDir);
+        if (mStrokeState.firstPt == mStrokeState.prevPt) strokeRoundPoint(mStrokeState.firstPt);
+        else {
+            strokeRound(mStrokeState.firstPt, {-mStrokeState.firstPtDir.x, -mStrokeState.firstPtDir.y});
+            strokeRound(mStrokeState.prevPt, mStrokeState.prevPtDir);
+        }
     }
 }
 
@@ -1807,8 +1816,6 @@ void Stroker::strokeClose()
 
     // join firstPt with prevPt
     this->strokeJoin(mStrokeState.firstPtDir);
-
-    mStrokeState.hasMove = false;
 }
 
 
@@ -1887,6 +1894,33 @@ void Stroker::strokeRound(const Point &prev, const Point& curr, const Point& cen
 }
 
 
+void Stroker::strokeRoundPoint(const Point &p)
+{
+    // Fixme: just use bezier curve to calculate step count
+    auto count = _bezierCurveCount(_bezFromArc(p, p, strokeRadius())) * 2;
+    auto c = _pushVertex(mResGlPoints, p.x, p.y);
+    auto step = 2 * M_PI / (count - 1);
+
+    for (uint32_t i = 1; i <= static_cast<uint32_t>(count); i++) {
+        float angle = i * step;
+        Point dir = {cos(angle), sin(angle)};
+        Point out = p + dir * strokeRadius();
+        auto oi = _pushVertex(mResGlPoints, out.x, out.y);
+
+        if (oi > 1) {
+            mResIndices->push(c);
+            mResIndices->push(oi);
+            mResIndices->push(oi - 1);
+        }
+    }
+
+    mLeftTop.x = std::min(mLeftTop.x, p.x - strokeRadius());
+    mLeftTop.y = std::min(mLeftTop.y, p.y - strokeRadius());
+    mRightBottom.x = std::max(mRightBottom.x, p.x + strokeRadius());
+    mRightBottom.y = std::max(mRightBottom.y, p.y + strokeRadius());
+}
+
+
 void Stroker::strokeMiter(const Point& prev, const Point& curr, const Point& center)
 {
     auto pp1 = prev - center;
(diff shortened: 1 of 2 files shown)


RuiwenTang:
If the Path only contains multiple Move instructions, will RoundPoints be drawn now?

hermet:
> If the Path only contains multiple Move instructions, will RoundPoints be drawn now?

@RuiwenTang Only move operations should not be taken into account for drawing.

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -1707,15 +1707,26 @@ void Stroker::doDashStroke(const PathCommand *cmds, uint32_t cmd_count, const Po
...
 {
-    if (mStrokeState.firstPt == mStrokeState.prevPt) return;
     if (mStrokeCap == StrokeCap::Butt) return;
 
     if (mStrokeCap == StrokeCap::Square) {
-        strokeSquare(mStrokeState.firstPt, {-mStrokeState.firstPtDir.x, -mStrokeState.firstPtDir.y});
-        strokeSquare(mStrokeState.prevPt, mStrokeState.prevPtDir);
+        if (mStrokeState.firstPt == mStrokeState.prevPt) {
+            if (mStrokeWidth > 0) {
+                strokeSquarePoint(mStrokeState.firstPt);
```
RuiwenTang, line 1715: This may result in only containing move instructions resulting in drawing StrokeCaps.
Can you help confirm  the rendering result like this ?
```
auto path = tvg::Shape::gen();

path->moveTo(100, 100);
path->moveTo(150, 100);
path->moveTo(200, 100);
path->strokeWidth(10);
path->strokeJoin(tvg::StrokeJoin::Round);
```

baszczewski: @RuiwenTang Good point. In the proposed solution, caps were drawn even for move operations, which was not consistent with what the SW renderer does. The code above was drawing three points, which should not happen. I have updated my pull request with what I believe is a better handling of this scenario, ensuring it aligns with the behavior of other renderers.

RuiwenTang (approved):
LGTM

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -1722,11 +1722,22 @@ void Stroker::strokeCap()
 
 void Stroker::strokeLineTo(const Point& curr)
 {
+    // special case for line with zero length
```
hermet, line 1725:
```
    // special case for line with zero length
    if (mStrokeState.prevPt == curr) {
        if (mStrokeCap == StrokeCap::Butt && tvg::zero(mStrokeWidth)) return;
        if (mStrokeCap == StrokeCap::Square) strokeSquarePoint(curr);
        else if (mStrokeCap == StrokeCap::Round) strokeRoundPoint(curr);
    }
```

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -1778,6 +1789,18 @@ void Stroker::strokeLineTo(const Point& curr)
 
 void Stroker::strokeCubicTo(const Point& cnt1, const Point& cnt2, const Point& end)
 {
+    if (mStrokeState.prevPt == end) {
```
hermet, line 1792:
```
    // special case for line with zero length
    if (mStrokeState.prevPt == curr) {
        if (mStrokeCap == StrokeCap::Butt && tvg::zero(mStrokeWidth)) return;
        if (mStrokeCap == StrokeCap::Square) strokeSquarePoint(curr);
        else if (mStrokeCap == StrokeCap::Round) strokeRoundPoint(curr);
    }
```

hermet (commented):
@baszczewski thanks, please check my comments. I recommend bringing more slick and compact code as I suggested with comments. (Otherwise, you can also leave it to our devs)

Here is one request you can update:
1. squash commits and revise the commit message. probably, you can reuse the PR messages, like this:

```
gl_engine: Fix for rendering short paths

Ensure they do not terminate prematurely for paths 
with a step of 0 or exceptionally small values 
when a valid stroke-width is present.

In my opinion, the optimal approach was to separate vertex generation 
into dedicated methods: strokeRoundPoint and strokeSquarePoint. 

My update supports two different stroke-cap styles. 
I have also tested it with various files (JSON, SVG) 
as well as a small example application similar 
to the one included in the previous pull request (#3066).

issue: https://github.com/thorvg/thorvg/issues/3065
```

src/renderer/gl_engine/tvgGlTessellator.cpp
baszczewski: Here, we don't have curr. However, we likely need to compare prevPt and end instead.

src/renderer/gl_engine/tvgGlTessellator.cpp
baszczewski: Thank you for suggesting a more concise form. Regarding the additional ```strokeDot``` method, wouldn't it be better to remove it? I have already introduced two additional methods for generating vertices, and we are not using complex conditions when triggering them.

baszczewski:
@hermet Thank you for your help with this issue. I made some minor adjustments to my pull request and combined several iterations of changes into a single commit.

hermet:
@baszczewski @RuiwenTang Please note that there are regressions here. Needs some rectification. :)

1. new_design.json  (before -> after)
image

2. GradientStroke (before -> after)
image

RuiwenTang:
How about I modify this part of the code instead? 
I'll modify this part as soon as I finish the Blending changes in the GL backend.

baszczewski:
The issue lies in rendering caps simultaneously in strokeCubicTo and strokeLineTo. If a cap has already been drawn by strokeCubicTo, it should not be drawn again by strokeLineTo. Additionally, we should probably perform stricter checks on all points in strokeCubicTo. I'll try to propose a patch.

baszczewski:
@hermet 
I slightly changed my approach in the code and moved the location where I initialize point drawing. Is everything okay now? I didn't have access to the new_design.json file. I verified with my files and the GradientStroke example.

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -1670,10 +1670,20 @@ void Stroker::doStroke(const PathCommand *cmds, uint32_t cmd_count, const Point
                 pts++;
             } break;
             case PathCommand::LineTo: {
+                // check if the line is a point
+                if (mStrokeState.prevPt == *pts) {
```
RuiwenTang, line 1674: LineTo target to same point needs to do nothing, I think this logical should only trigged in handleStrokeCap
think about this:
```
path->moveTo(10, 10);
path->lineTo(100, 10);
path->lineTo(100, 10);
path->lineTo(10,20);
```

>BTW. The code for the stroke tessellation is a little bit mess :(

baszczewski: Did you specifically mean the strokeCap method? That’s where I initially started working on this issue, but it turned out not to be the best place, as we don’t want to draw caps when dealing solely with move operations.

In your example code, a cap would unnecessarily appear along with a lineTo to the same point, which shouldn’t happen.

baszczewski: @RuiwenTang 
What do you think about this solution?

```c++
void Stroker::doStroke(const PathCommand *cmds, uint32_t cmd_count, const Point *pts, uint32_t pts_count)
{
    mResGlPoints->reserve(pts_count * 4 + 16);
    mResIndices->reserve(pts_count * 3);

    auto allowZeroLengthCap = false;

    for (uint32_t i = 0; i < cmd_count; i++) {
        switch (cmds[i]) {
            case PathCommand::MoveTo: {
                if (mStrokeState.hasMove) {
                    strokeCap(allowZeroLengthCap);
                    mStrokeState.hasMove = false;
                }
                mStrokeState.hasMove = true;
                mStrokeState.firstPt = *pts;
                mStrokeState.firstPtDir = {0.0f, 0.0f};
                mStrokeState.prevPt = *pts;
                mStrokeState.prevPtDir = {0.0f, 0.0f};
                pts++;
                allowZeroLengthCap = false;
            } break;
            case PathCommand::LineTo: {
                if (mStrokeState.prevPt == *pts) allowZeroLengthCap = true;
                this->strokeLineTo(*pts);
                pts++;
            } break;
            case PathCommand::CubicTo: {
                if (mStrokeState.prevPt == pts[2]) allowZeroLengthCap = true;
                this->strokeCubicTo(pts[0], pts[1], pts[2]);
                pts += 3;
            } break;
            case PathCommand::Close: {
                this->strokeClose();

                mStrokeState.hasMove = false;
            } break;
            default:
                break;
        }
    }
    strokeCap(allowZeroLengthCap);
}
```

and

```c++
void Stroker::strokeCap(bool allowZeroLength)
{
    if (!allowZeroLength && mStrokeState.firstPt == mStrokeState.prevPt) return;
    if (mStrokeCap == StrokeCap::Butt) return;

    if (mStrokeCap == StrokeCap::Square) {
        if (allowZeroLength && mStrokeState.firstPt == mStrokeState.prevPt) strokeSquarePoint(mStrokeState.firstPt);
        else {
            strokeSquare(mStrokeState.firstPt, {-mStrokeState.firstPtDir.x, -mStrokeState.firstPtDir.y});
            strokeSquare(mStrokeState.prevPt, mStrokeState.prevPtDir);
        }
    } else if (mStrokeCap == StrokeCap::Round) {
        if (allowZeroLength && mStrokeState.firstPt == mStrokeState.prevPt) strokeRoundPoint(mStrokeState.firstPt);
        else {
            strokeRound(mStrokeState.firstPt, {-mStrokeState.firstPtDir.x, -mStrokeState.firstPtDir.y});
            strokeRound(mStrokeState.prevPt, mStrokeState.prevPtDir);
        }
    }
}
```

RuiwenTang: If so, I think only check some thing like `allowZeroLengthCap ` is enough.
Maybe the following code can work perfectly.
```
void Stroker::doStroke(const PathCommand *cmds, uint32_t cmd_count, const Point *pts, uint32_t pts_count)
{
    mResGlPoints->reserve(pts_count * 4 + 16);
    mResIndices->reserve(pts_count * 3);

    auto hasLineTo = false;

    for (uint32_t i = 0; i < cmd_count; i++) {
        switch (cmds[i]) {
            case PathCommand::MoveTo: {
                if (hasLineTo) { // check this, so we can skip if path only contains move instruction
                    strokeCap();
                    hasLineTo = false;
                }
                mStrokeState.hasMove = true;
                mStrokeState.firstPt = *pts;
                mStrokeState.firstPtDir = {0.0f, 0.0f};
                mStrokeState.prevPt = *pts;
                mStrokeState.prevPtDir = {0.0f, 0.0f};
                pts++;
                hasLineTo = false;
            } break;
            case PathCommand::LineTo: {
                hasLineTo = true;
                this->strokeLineTo(*pts);
                pts++;
            } break;
            case PathCommand::CubicTo: {
                 hasLineTo = true;
                this->strokeCubicTo(pts[0], pts[1], pts[2]);
                pts += 3;
            } break;
            case PathCommand::Close: {
                this->strokeClose();

                hasLineTo = false;
            } break;
            default:
                break;
        }
    }
    if (hasLineTo) strokeCap();
}

// strokeCap can handle if the path is zero length.
void Stroker::strokeCap()
{
    if (mStrokeCap == StrokeCap::Butt) return;

    if (mStrokeCap == StrokeCap::Square) {
        if (mStrokeState.firstPt == mStrokeState.prevPt) strokeSquarePoint(mStrokeState.firstPt);
        else {
            strokeSquare(mStrokeState.firstPt, {-mStrokeState.firstPtDir.x, -mStrokeState.firstPtDir.y});
            strokeSquare(mStrokeState.prevPt, mStrokeState.prevPtDir);
        }
    } else if (mStrokeCap == StrokeCap::Round) {
        if (mStrokeState.firstPt == mStrokeState.prevPt) strokeRoundPoint(mStrokeState.firstPt);
        else {
            strokeRound(mStrokeState.firstPt, {-mStrokeState.firstPtDir.x, -mStrokeState.firstPtDir.y});
            strokeRound(mStrokeState.prevPt, mStrokeState.prevPtDir);
        }
    }
}

```

baszczewski: @RuiwenTang This solution seems fine. Additionally, I noticed that in this case, the ```hasMove``` flag in the ```State``` structure is only updated but never read. Perhaps we should remove this value as well? Or do you have any specific plans for this flag?

RuiwenTang: Ok, it can be removed.

hermet:
> I didn't have access to the new_design.json file.
Here: https://github.com/thorvg/thorvg/blob/main/examples/resources/lottie/new_design.json

baszczewski:
@hermet @RuiwenTang 
Thank you for your support. Is the pull request in its current form now acceptable?

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -1655,38 +1655,42 @@ void Stroker::doStroke(const PathCommand *cmds, uint32_t cmd_count, const Point
     mResGlPoints->reserve(pts_count * 4 + 16);
     mResIndices->reserve(pts_count * 3);
 
+    auto hasLineTo = false;
```
hermet, line 1658: this doesn't mean only lineTo by curveTo. Suggestion:
```
auto validStrokeCap 
```

baszczewski: @hermet Done.

PR #3079, opened 2024-12-27
svg: remove styles from the clipPath

hermet:
Main purpose clippath is to cutout the
drawing region with a path. Otherwise
we need to use MaskMethod.

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -180,26 +180,21 @@ static RadialGradient* _applyRadialGradientProperty(SvgStyleGradient* g, const B
 }
 
 
-//The SVG standard allows only for 'use' nodes that point directly to a basic shape.
-static bool _appendClipUseNode(SvgLoaderData& loaderData, SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath)
+static bool _appendClipChild(SvgLoaderData& loaderData, SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath)
 {
-    if (node->child.count != 1) return false;
-    auto child = *(node->child.data);
+    //The SVG standard allows only for 'use' nodes that point directly to a basic shape.
+    if (node->type == SvgNodeType::Use) {
+        auto child = *(node->child.data);
 
-    Matrix finalTransform = {1, 0, 0, 0, 1, 0, 0, 0, 1};
-    if (node->transform) finalTransform = *node->transform;
-    if (node->node.use.x != 0.0f || node->node.use.y != 0.0f) {
-        finalTransform *= {1, 0, node->node.use.x, 0, 1, node->node.use.y, 0, 0, 1};
+        Matrix finalTransform = {1, 0, 0, 0, 1, 0, 0, 0, 1};
+        if (node->transform) finalTransform = *node->transform;
+        if (node->node.use.x != 0.0f || node->node.use.y != 0.0f) {
+            finalTransform *= {1, 0, node->node.use.x, 0, 1, node->node.use.y, 0, 0, 1};
+        }
+        if (child->transform) finalTransform = *child->transform * finalTransform;
+
+        return _appendClipShape(loaderData, child, shape, vBox, svgPath, identity((const Matrix*)(&finalTransform)) ? nullptr : &finalTransform);
     }
-    if (child->transform) finalTransform = *child->transform * finalTransform;
-
-    return _appendClipShape(loaderData, child, shape, vBox, svgPath, identity((const Matrix*)(&finalTransform)) ? nullptr : &finalTransform);
-}
-
-
-static bool _appendClipChild(SvgLoaderData& loaderData, SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath, bool clip)
-{
-    if (node->type == SvgNodeType::Use) return _appendClipUseNode(loaderData, node, shape, vBox, svgPath);
     return _appendClipShape(loaderData, node, shape, vBox, svgPath, nullptr);
 }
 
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -434,7 +434,6 @@ static bool _appendClipShape(SvgLoaderData& loaderData, SvgNode* node, Shape* sh
         }
     }
 
-    _applyProperty(loaderData, node, shape, vBox, svgPath, true);
```
hermet, line 437: This even properly handles the return value that could potentially a memory leak.

JSUYA (commented):
_applyProperty() doesn't apply only style properties (_applyComposition() is called after).
I made a sample to verify this, but I found that it has a bug unrelated to this PR. #3082 #3083 
I still think we need to call _applyComposition() though.

hermet:
> _applyProperty() doesn't apply only style properties (_applyComposition() is called after). I made a sample to verify this, but I found that it has a bug unrelated to this PR. #3082 #3083 I still think we need to call _applyComposition() though.

@JSUYA _applyComposition() should be addressed properly regardless of this PR. Until then, we cannot apply it (as I mentioned the memory leak), as the purpose of this PR does not cover that.

mgrudzinska:
@JSUYA I agree that _applyComposition() is important in this case, but after c14382f this is not handled properly. I'd remove this is as @hermet suggested. this issue should be handled in https://github.com/thorvg/thorvg/issues/3082

JSUYA (approved):
I checked that #3082 and #3083 are working good. thanks

+)
There is still an issue with empty clipPath not being handled, but I think it can be addressed in another patch.
```
<svg  viewBox="0 0 100 100"  xmlns="http://www.w3.org/2000/svg">
  <clipPath id="myClip">
  </clipPath>

  <path
    id="heart"
    d="M10,30 A20,20,0,0,1,50,30 A20,20,0,0,1,90,30 Q90,60,50,90 Q10,60,10,30 Z" clip-path="url(#myClip)" />
</svg>

```

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -180,26 +180,21 @@ static RadialGradient* _applyRadialGradientProperty(SvgStyleGradient* g, const B
...
+    //The SVG standard allows only for 'use' nodes that point directly to a basic shape.
+    if (node->type == SvgNodeType::Use) {
+        auto child = *(node->child.data);
 
+        Matrix finalTransform = {1, 0, 0, 0, 1, 0, 0, 0, 1};
+        if (node->transform) finalTransform = *node->transform;
+        if (node->node.use.x != 0.0f || node->node.use.y != 0.0f) {
+            finalTransform *= {1, 0, node->node.use.x, 0, 1, node->node.use.y, 0, 0, 1};
+        }
+        if (child->transform) finalTransform = *child->transform * finalTransform;
```
mgrudzinska, line 194: ```finalTransform *= *child->transform;```

sample:
```

  <defs>
    <circle id="m1" cx="30" cy="50" r="30" transform="matrix(1 0.44 1 1 0 10)"/>
  </defs>
  
<clipPath id="clip">
    <use href="#m1" x="10" y="12"/>
</clipPath>


<rect x="0" y="0" width="100" height="100"  fill="red" clip-path="url(#clip)"
      fill-opacity="0.5"/>

<use href="#m1" x="10" y="12" fill="blue" opacity="0.4"/>

  <rect x="47" y="52" width="20" height="20"  fill="red"
      fill-opacity="0.5"/>
</svg>
```

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -180,26 +180,21 @@ static RadialGradient* _applyRadialGradientProperty(SvgStyleGradient* g, const B
...
-    if (node->node.use.x != 0.0f || node->node.use.y != 0.0f) {
-        finalTransform *= {1, 0, node->node.use.x, 0, 1, node->node.use.y, 0, 0, 1};
-    }
-    if (child->transform) finalTransform = *child->transform * finalTransform;
-
-    return _appendClipShape(loaderData, child, shape, vBox, svgPath, identity((const Matrix*)(&finalTransform)) ? nullptr : &finalTransform);
-}
+    //The SVG standard allows only for 'use' nodes that point directly to a basic shape.
+    if (node->type == SvgNodeType::Use) {
+        auto child = *(node->child.data);
```
mgrudzinska, line 187: if the use node references a node that doesn;t exist, a crash will occure:
```
<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
  
<clipPath id="clip">
    <use href="#m1" x="0" fill="blue" />
</clipPath>

<rect x="0" y="0" width="100" height="100"  fill="red" clip-path="url(#clip)"
      fill-opacity="0.5"/>

</svg>
```

I removed from the sample:
```
  <defs>
    <circle id="m1" cx="30" cy="50" r="30" transform="matrix(1 0.44 1 1 0 10)"/>
  </defs>
```

mgrudzinska:
@hermet I left comments. Other than that, I have no further remarks. I tested other cases, and it works fine.

src/loaders/svg/tvgSvgSceneBuilder.cpp
hermet: @mgrudzinska Is there a bug with that sample? or you just want to suggest a better code? Please clarify the comment a bit more.

hermet: @mgrudzinska Please create a separate issue if you think it should be modified. This transform is part of the original logic, so it would be better to address it separately rather than modifying it in this PR.

Issue #3072, opened 2024-12-23
Questions about BlendMode


RuiwenTang:
I'm currently trying to refactor the advanced color blending logic of the GL backend.  I'm confused by some of Thorvg's behaviors on color blending.
The main problem is that I found that the behavior of blend mode is inconsistent when applied to a normal Shape or to a Scene.
For example:

---

```
//Normal
auto shape1 = tvg::Shape::gen();
shape1->appendRect(0, 0, 400, 400, 50, 50);
shape1->fill(0, 255, 255);
shape1->blend(tvg::BlendMethod::Normal);
canvas->push(shape1);

// direct draw shape with blend mode
auto shape = tvg::Shape::gen();
shape->appendRect(0, 0, 200, 200, 50, 50);
shape->fill(255, 255, 255, 250);
shape->blend(tvg::BlendMethod::ColorDodge);
canvas->push(shape);
```
This code get the following result:
<img width="428" alt="截屏2024-12-23 下午5 53 22" src="https://github.com/user-attachments/assets/02e9b650-d254-43d7-bcb0-21880753d8d2" />

---

But if put the shape inside a scene, even if the scene only contains one shape, like this:
```
//Normal
auto shape1 = tvg::Shape::gen();
shape1->appendRect(0, 0, 400, 400, 50, 50);
shape1->fill(0, 255, 255);
shape1->blend(tvg::BlendMethod::Normal);
canvas->push(shape1);


//ColorDodge On the scene
auto scene = tvg::Scene::gen();
scene->blend(tvg::BlendMethod::ColorDodge);

auto shape = tvg::Shape::gen();
shape->appendRect(0, 0, 200, 200, 50, 50);
shape->fill(255, 255, 255, 250);
scene->push(shape);

canvas->push(scene);

``` 
Got this result:
<img width="425" alt="截屏2024-12-23 下午5 57 10" src="https://github.com/user-attachments/assets/f8c9d779-1bfc-4135-bece-5ace6ddf57f7" />

> Some blend modes seem to draw nothing when DstColor is all zeros, and this behavior is different from skia or impeller or html & css.
> I'm going to align SKIA's Blend behavior in the GL backend by following this reference : https://github.com/google/skia/blob/main/include/core/SkBlendMode.h

RuiwenTang:
This is the result rendered by SKIA:
[...]

Issue #2799, opened 2024-10-01
gl_engine: Lottie rendering compatibility


hermet:
This list shows the broken compatibility of Lottie rendering with the OpenGL backend.
https://github.com/thorvg/thorvg/tree/main/examples/resources/lottie
\+ @RuiwenTang 

- [x] 11555  //flickering, dash line
- [x] birth_stone_logo //trimpath?
- [x] fly_in_beaker //trimpath?
- [x] frog_vr //trimpath?
- [x] cat_loader //flickering
- [x] happy_holidays  //trimpath
- [x] holdanimation //trimpath
- [x] isometric //trimpath
- [x] skullboy //trimpath
- [x] threads //trimpath
- [x] lolo //right arm layering
- [x] seawalk //right foot stroke disappeared
- [x] expressions/10151  //incorrect rotation
- [x] expressions/11272 //trimpath
- [x] expressions/jolly_walker //stroke cap?
- [x] expressions/world_locations //flickering, trimpath
- [x] polystar //incorrect join
- [x] starstrips //incorrect join
- [ ] guitar //blending
- [ ] expressions/10444  //blending
- [ ] expressions/intelia_logo_animation   //blending

hermet:
Pass rate: 60%

hermet:
Pass rate: 70%

sid-kumar-figure:
Hi, I have just started looking into using ThorVG for rendering lottie files using the gl_engine, and I noticed that running the example I see a significant amoutn slow down during rendering some of the examples :

RuiwenTang:
> Hi, I have just started looking into using ThorVG for rendering lottie files using the gl_engine, and I noticed that running the example I see a significant amount of slow down during rendering some of the examples :
> 
> 1. 1643-exploding-start
> 2. 5317-fireworks
> 
> I was wondering if this is related to this issue ?
> 
> For example here is a small snippet of the rendering time from the Lottie Example for 5317-fireworks (screen is 60fps) :
> 
> ```
> Lottie: experimental/thorvg_test/content/lottie/5317-fireworkds.json
> Refresh took 7ms
> Draw took 1ms
> Refresh took 1ms
> [    1]: elapsed time = 19ms (19ms)
> Update took 0ms
> Draw took 1ms
> Refresh took 5ms
> [    2]: elapsed time = 6ms (12ms)
> Update took 0ms
> Draw took 1ms
> Refresh took 2ms
> [    3]: elapsed time = 3ms (9ms)
[...]

PR #3081, opened 2024-12-27
gl_engine: rewrite the blending shader

RuiwenTang:
Completely rewrite the GL backend shader to support advance blending and correct the blending behavior.
Now the GL backend has the same blending result as SKIA and CanvasRenderingContext2D in browers did.

The formula is referenced from https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators

Note this change make the `Blending` example rendered different with other backend, but the result is consistent with SKIA, which I believe is correct.

The current rendering result can reference this issue #3072
It also fix the blending issue in Lottie animation: #2799


hermet:
@RuiwenTang Please check this:

image

hermet:
image

hermet:
@RuiwenTang I'd like to recommend separating the incorrect equation (ColorBurn, ColorEdge) from the destination alpha issue. We can apply the equation correction commit first.

sw:
image

gl:
image

RuiwenTang:
@hermet 
image
This is what I have been saying. I think this PR fixes all the problems of GLEngine in Blending, and I think that all of the result is correct.

hermet:
> @hermet image This is what I have been saying. I think this PR fixes all the problems of GLEngine in Blending, and I think that all of the result is correct.

@RuiwenTang 
1. Separating commits per problems. That is our basic policy.
2. You know, I didn't say that is the wrong result. I requested you to see me further the layer level scene blending first to put them together fix. You are not working alone. Please humor me.

hermet:
@SergeyLebedkin Please review this, 

+ `Animation` looks leaking gpu resources after this PR.

Issue #3039, opened 2024-12-10
lottie: thorvg v1 prerelease rendering regression 


theashraf:
in [dotLottie-rs](https://github.com/LottieFiles/dotlottie-rs), i'm trying to migrate to thorvg v1 and the only change I introduced was modifying the render cycle to clear the buffer before drawing. However, I’m not sure why the animation is getting clipped.

Here’s the updated render cycle code:  
```rs
fn render(&mut self) -> Result<(), LottieRendererError> {
        self.renderer.update().map_err(into_lottie::<R>)?; // calls tvg_canvas_update c api

        #[cfg(feature = "thorvg-v1")]
        self.renderer.clear(false, true).map_err(into_lottie::<R>)?; // calls tvg_canvas_clear c api

        self.renderer.draw().map_err(into_lottie::<R>)?; // calls tvg_canvas_draw c api
        self.renderer.sync().map_err(into_lottie::<R>)?; // calls tvg_canvas_sync c api

        Ok(())
   }
```

Behavior with ThorVG v0.15.6:  
<img src="https://github.com/user-attachments/assets/8b767fd3-f955-4d1c-a00e-08c7efddada1" width="100" />

Behavior with ThorVG v1.0.0-pre7:  
<img src="https://github.com/user-attachments/assets/2bcba599-0933-42c3-920c-0c9f0e8ae305" width="100" />

> This issue is not specific to a certain animation

hermet:
@theashraf could you please share the resource as well?

theashraf:
> @theashraf could you please share the resource as well?

@hermet  [test-lottie.json](https://github.com/user-attachments/files/18089432/test-lottie.json)

hermet:
curprit: https://github.com/thorvg/thorvg/commit/367ac5f331383eea978f07131c64e54fee3f8e58

tinyjin:
Here is difference the commit mentioned and v0.15.x.

The regression issue is come from when use `picture->size` function instead of `scale` after applying mentioned commit.

**[After commit]**
https://github.com/thorvg/thorvg/commit/367ac5f331383eea978f07131c64e54fee3f8e58

When the animation(canvas) size is bigger than given picture size, rendering result is clipped.

![CleanShot 2024-12-27 at 19 14 02@2x](https://github.com/user-attachments/assets/2881bed0-c122-4e36-9a2a-7744d9e96e90)

**[v0.15.x]**

No issue in same condition.

[...]

PR #3084, opened 2024-12-27
lottie: fix viewport clipping issue when resize

tinyjin:
After resizing animation, base clipper for viewport(of root scene) didn't change. This causes issue that animation always clipped by initial psize.

This also brings v1 regression issue after `Result Picture::size(float w, float h)` function called:
Issue: #3039 

### Example:
_psize is 500 by 500, and resized to 700 by 700_

**[v1.0-pre]**

Even if resized to bigger, animations shows in initial size(500 x 500)

CleanShot 2024-12-27 at 23 22 20@2x

**[v1.0-pre + this patch]**

CleanShot 2024-12-27 at 23 24 47@2x

### DIFF
--- a/src/loaders/lottie/tvgLottieLoader.cpp
+++ b/src/loaders/lottie/tvgLottieLoader.cpp
@@ -260,9 +260,8 @@ bool LottieLoader::resize(Paint* paint, float w, float h)
     paint->transform(m);
 
     //apply the scale to the base clipper
-    const Paint* clipper;
-    paint->mask(&clipper);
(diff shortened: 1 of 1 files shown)


hermet (changes requested):
please check with this:
```
    //apply the scale to the base clipper
    auto clipper = PAINT(paint)->clipper;
    if (clipper) clipper->transform(m);
```

tinyjin:
CleanShot 2024-12-30 at 11 56 19@2x

@hermet It works too!

hermet (approved):
@tinyjin good job, thanks.

PR #3098, opened 2025-01-03
sw_engine: replaced the line with cubic

hermet:
A trial for size and maintainence efficiency.
The size can be reduced by 1kb.


hermet:
Not quite sure that efficiency.

Issue #3102, opened 2025-01-07
lottie: Runtime crash by heap-buffer-overflow


tinyjin:
Animation File: [sample.json](https://github.com/user-attachments/files/18327865/sample.json)

Crashes on landscape resolution (w > h)

Related: https://github.com/LottieFiles/dotlottie-ios/issues/49

![CleanShot 2025-01-07 at 13 09 37@2x](https://github.com/user-attachments/assets/c6a8adb9-a5fb-4334-9f34-535586a85997)

![CleanShot 2025-01-07 at 13 09 51@2x](https://github.com/user-attachments/assets/534694f6-2f3f-4435-ba05-1c47cba6097c)

![CleanShot 2025-01-07 at 13 10 04@2x](https://github.com/user-attachments/assets/1ad974ef-9518-4229-b87d-424bd7a30bf2)

hermet:
@tinyjin Could you please test with this [PR](https://github.com/thorvg/thorvg/pull/3105)? I couldn't reproduce the crash on my macOS.

tinyjin:
![CleanShot 2025-01-08 at 17 30 14@2x](https://github.com/user-attachments/assets/b7524e84-be42-451e-9c09-4012f3cbebaa)

Still happenes

> @tinyjin Could you please test with this [PR](https://github.com/thorvg/thorvg/pull/3105)? I couldn't reproduce the crash on my macOS / linux.

hermet:
@tinyjin could you please share the following value at the crash point?

`vv`, `sw`, `uu` and `image->w`, `image->h`

elahav:
Could this be the same issue I reported today in the discussion section? See my patch there:

https://github.com/orgs/thorvg/discussions/3109

hermet:
@elahav I guess not. :-)

hermet:
> Could this be the same issue I reported today in the discussion section? See my patch there:
> 
> https://github.com/orgs/thorvg/discussions/3109

@elahav I guess not. :-)

tinyjin:
> @tinyjin could you please share the following value at the crash point?
> 
> `vv`, `sw`, `uu` and `image->w`, `image->h`

Sure, here is a log I've gotten.

![CleanShot 2025-01-10 at 16 48 48@2x](https://github.com/user-attachments/assets/7bee378c-73d4-4647-a702-90c7a9fd003f)

PR #3105, opened 2025-01-08
sw_engine: ++portability on macOS

hermet:
fixed a runtime error report of the sanitizer
at data casting.

issue: https://github.com/thorvg/thorvg/issues/3102


src/renderer/sw_engine/tvgSwRasterTexmap.h
```diff
@@ -569,7 +569,7 @@ static void _rasterPolygonImageSegment(SwSurface* surface, const SwImage* image,
                     u += _dudx;
                     v += _dvdx;
                     //range over?
-                    if ((uint32_t)v >= image->h) break;
+                    if ((uint32_t)(int32_t)v >= image->h) break;
```
mgrudzinska, line 572: this solve warning on mac. can we consider to do this in all places marked as "range over" ?
ex. Blending sample causes the same warning in l.382

hermet: Yes, makes sense.

PR #3107, opened 2025-01-08
Fix compilation warnings

faxe1008:
Hi,

The zephyrproject (which I am part of), wants to use lottie widget in LVGL which in turn uses your excellent library. Unfortunately our CI has quite a few strict compliation settings turned on. One of them being -Wdouble-promotion.  I thought the easiest way was to fix it at the source and therefore I combed through all the warnings that I found.


src/renderer/tvgLoader.cpp
```diff
@@ -368,7 +368,7 @@ LoadModule* LoaderMgr::loader(const char* data, uint32_t size, const char* mimeT
                 }
                 return loader;
             } else {
-                TVGLOG("LOADER", "Given mimetype \"%s\" seems incorrect or not supported.", mimeType);
+                TVGLOG("LOADER", "Given mimetype seems incorrect or not supported.");
```
hermet, line 371: At 361 line should be `mimeType` instead of `!mimeType`. Seems a wrong logic.

hermet (commented):
@faxe1008 Thanks, Please check a comment (or you can leave us to fix it), I applied only them for now, https://github.com/thorvg/thorvg/commit/9a8173cc22dfc4323f326f9171479ff0ca94846a https://github.com/thorvg/thorvg/commit/1d537de1cda0b109a615f28adcadf2911d2feccf

Could you please rebase again? The other patches seem to be somewhat controversial, so we'll need to review them one by one after that.

src/renderer/tvgLoader.cpp
faxe1008: Done it :^)

meson.build
```diff
@@ -1,6 +1,6 @@
 project('thorvg',
         'cpp',
-        default_options : ['buildtype=debugoptimized', 'b_sanitize=none', 'werror=false', 'optimization=3', 'cpp_std=c++14', 'strip=true'],
+        default_options : ['buildtype=debugoptimized', 'b_sanitize=none', 'werror=false','c_args=-Wdouble-promotion', 'cpp_args=-Wdouble-promotion', 'optimization=3', 'cpp_std=c++14', 'strip=true'],
```
hermet, line 3: We may need to further review whether to add 'double-promotion.' Alongside that, I'd like to suggest adding these arguments in `src/meson.build`. Also, since ThorVG is primarily a C++ project, I think `c_args` may not be strictly necessary.

src/renderer/tvgLoader.cpp
hermet: @faxe1008 awesome, thanks!

meson.build
faxe1008: Yeah I used this locally to not miss any warnings. Just put them into the PR because having a strict set of rules is quite useful in CI I think

faxe1008: Yeah I used this locally to not miss any warnings. Just put them into the PR because having a strict set of rules is quite useful in CI I think.

hermet: I'm not sure if adding only this setting here will be of practical help to the CI. I think, it might be better to aim for consistency.

hermet:
some resolved by: https://github.com/thorvg/thorvg/pull/3212

hermet:
@faxe1008 hopefully, https://github.com/thorvg/thorvg/pull/3212 fixes all your integration. thanks.

faxe1008:
This looks great @hermet . Thanks for addressing this. I will only now if everything works as intended after LVGL updates their ThorVG version. Will come back here if there are still things to fix. :^)

PR #3112, opened 2025-01-10
wasm: fix build issues

tinyjin:
Fixed wasm build issues from latest webgl changes.

CleanShot 2025-01-10 at 16 31 27@2x

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -188,9 +188,7 @@ struct TvgWgEngine : TvgEngineMethod
 
 struct TvgGLEngine : TvgEngineMethod
 {
-#ifdef THORVG_GL_RASTER_SUPPORT
-    EMSCRIPTEN_WEBGL_CONTEXT_HANDLE context = 0;
-#endif
+    intptr_t context = 0;
     ~TvgGLEngine() override
     {
     #ifdef THORVG_GL_RASTER_SUPPORT
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -188,7 +188,9 @@ struct TvgWgEngine : TvgEngineMethod
 
 struct TvgGLEngine : TvgEngineMethod
 {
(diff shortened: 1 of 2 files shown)


src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -803,8 +803,6 @@ bool GlRenderer::target(void* context, int32_t id, uint32_t w, uint32_t h)
     //assume the context zero is invalid
     if (!context || id == GL_INVALID_VALUE || w == 0 || h == 0) return false;
 
-    currentContext();
-
     flush();
```
hermet, line 806: this call should be guanrateed with currentContext...

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -188,7 +188,9 @@ struct TvgWgEngine : TvgEngineMethod
 
 struct TvgGLEngine : TvgEngineMethod
 {
+#ifdef THORVG_GL_RASTER_SUPPORT
     EMSCRIPTEN_WEBGL_CONTEXT_HANDLE context = 0;
```
hermet, line 192: Suggest just use `int`  without EMSCRIPTEN dependency.

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -803,8 +803,6 @@ bool GlRenderer::target(void* context, int32_t id, uint32_t w, uint32_t h)
     //assume the context zero is invalid
     if (!context || id == GL_INVALID_VALUE || w == 0 || h == 0) return false;
 
-    currentContext();
```
hermet, line 806: if (mContext) currentContext();

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -188,7 +188,7 @@ struct TvgWgEngine : TvgEngineMethod
 
 struct TvgGLEngine : TvgEngineMethod
 {
-    EMSCRIPTEN_WEBGL_CONTEXT_HANDLE context = 0;
+    uintptr_t context = 0;
```
hermet, line 191: `intptr_t` ?
image

hermet (approved):

PR #2659, opened 2024-08-20
sw_engine: fix image rastering

mgrudzinska:
In the case of _rasterPolygonImageSegment the mapping algorithm could go out beyond the buffer range, causing crash. A check has been added to prevent this.

PR #3113, opened 2025-01-10
sw_engine: Fix buffer overflow in texture mapping rasterizer

tinyjin:
Fix heap buffer overflow in texture mapping rasterizer by adding proper bounds checking for texture coordinates. This prevents accessing memory outside of the allocated image buffer during texture sampling and interpolation.

---

**Situation:**
At the ` _rasterPolygonImageSegment`, `vv` value gets `-1` then crahses.

image

**Sample file:** 
[Lottie Runtime Crash Issue 3102.json](https://github.com/user-attachments/files/18373071/Lottie.Runtime.Crash.Issue.3102.json)

**Crash Point:**
Drawing `image_0`

image

**Fixes**

Added a defense logic for preventing the values get negative value, but this need to be reviewed properly.

Issues: #3102

### DIFF
--- a/src/renderer/sw_engine/tvgSwRasterTexmap.h
+++ b/src/renderer/sw_engine/tvgSwRasterTexmap.h
@@ -85,9 +85,8 @@ static void _rasterBlendingPolygonImageSegment(SwSurface* surface, const SwImage
     float _xa = xa, _xb = xb, _ua = ua, _va = va;
     auto sbuf = image->buf32;
     auto dbuf = surface->buf32;
-    int32_t sw = static_cast<int32_t>(image->stride);
-    int32_t sh = image->h;
-    int32_t dw = surface->stride;
+    int32_t sw = static_cast<int32_t>(image->w);
+    int32_t sh = static_cast<int32_t>(image->h);
     int32_t x1, x2, x, y, ar, ab, iru, irv, px, ay;
     int32_t vv = 0, uu = 0;
     int32_t minx = INT32_MAX, maxx = 0;
@@ -144,7 +143,7 @@ static void _rasterBlendingPolygonImageSegment(SwSurface* surface, const SwImage
             u = _ua + dx * _dudx;
             v = _va + dx * _dvdx;
 
-            buf = dbuf + ((y * dw) + x1);
+            buf = dbuf + ((y * surface->stride) + x1);
 
             x = x1;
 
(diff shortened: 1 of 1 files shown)


mgrudzinska (approved):
long time ago I had some other problem and this was in the same part of the code:
https://github.com/thorvg/thorvg/pull/2659/files
I added then 
```
            if (u <= -1.0f) u = 0.0f;
            if (v <= -1.0f) v = 0.0f;
```
as soon as possible just to skip all values that can be skipped. You can check whether it is worth it.

hermet (changes requested):
This is the worst scenario you are adding the useless operations `16,156,800` per frame. It's not ignoreable. Let's clarify which scenario could reach to the value less than zero. 

It would be helpful to understand by firstly comparing the normal case and the abnormal case at the certain mac system to identify which parts are causing this different result.

hermet:
It would be helpful to understand by firstly comparing the normal case and the abnormal case of the certain mac system to identify which parts are causing this different result.

tinyjin:
This is test between Mac(Arm) and Windows(Intel).

At the line 348
```cpp
px = *(sbuf + (vv * sw) + uu);
```

In some case(given size of animation), result of `(vv * sw) + uu` is negative value, then accesses `sbuf` with negative index.

However two machines have different result at this point:

Mac(Arm)
- `*(sbuf + (vv * sw) + uu)` caused heap-buffer-overflow

Windows(Arm)
- `*(sbuf + (vv * sw) + uu)` returned `0`

This out-of-bounds memory access with negative offsets results seems to have different behaviors depending on the architecture.


Thus, these two points make potential crash and different behavior.
1. The value falls into negative with specific image and size.
2. Arm/Intel have different result in negative indexing with `uint32_t*` variable.

Regarding to the point No. 1:
- The buffer overflow occurs when scaling a square image (e.g. 846x846) to a rectangle canvas (e.g. 500x350).
- This seems to happen because the aspect ratio scaling causes UV mapping coordinates to exceed the image bounds.

hermet:
> In some case(given size of animation), result of `(vv * sw) + uu` is negative value, then accesses `sbuf` with negative index.

@tinyjin I couldn't find the value has a negative value. Could you please double-check it? 

image

tinyjin:
@hermet 

I still have a negative value in the case.

Can you verify again with this branch?

**[Mac/Arm]**
CleanShot 2025-01-16 at 12 00 18@2x

**[Windows/Intel]**
image

tinyjin:
@hermet After the new push, the issue has gone! 👍

hermet (approved):

PR #2866, opened 2024-10-16
gl_engine: dynamic load GL functions with glad

RuiwenTang:
This PR shows how to dynamic load GL functions with [GLAD](https://github.com/dav1dde/glad-web) (one of the [OpenGL Loader Library](https://www.khronos.org/opengl/wiki/OpenGL_Loading_Library)).

I choose `GLAD`  because it only contains a single source file, and does not need to link OpenGL library at compile time.

Issue: https://github.com/thorvg/thorvg/issues/2453 https://github.com/thorvg/thorvg/issues/2562

duplicate: #2574

PR #3114, opened 2025-01-11
support glad

std-microblock:
is it possible to support glad? this is simple and this pr can be a sample implementation.

### DIFF
--- a/src/renderer/gl_engine/tvgGlCommon.h
+++ b/src/renderer/gl_engine/tvgGlCommon.h
@@ -25,6 +25,9 @@
 
 #include <assert.h>
 #include <memory>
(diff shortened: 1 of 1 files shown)


std-microblock:
```lua
    io.replace("src/renderer/gl_engine/tvgGlCommon.h", "#include <GL/gl.h>",
        "#include <glad/glad.h>\n#define M_PI 3.1415926535897", { plain = true })

    io.replace("meson.build", "include_directories('.')", "include_directories('.'), include_directories('" ..
        package:dep("glad"):installdir():gsub("\\", "/")
        .. "/include')", { plain = true })
```

some changes I made in xmake.lua to make it work

vtorri:
duplicate of #2866

Issue #3118, opened 2025-01-13
common: TrimPath enhancement


hermet:
This issue consolidates current TrimPath issues in one place.

- [x] TrimPath refactoring for unify the implementation across the render backends.
https://github.com/thorvg/thorvg/issues/2854
- [x] Add Lottie TrimPath for the Shape fill Support
- [ ] TrimPath BugFix & Stabilization
https://github.com/thorvg/thorvg/issues/3192
https://github.com/thorvg/thorvg/issues/3133
https://github.com/thorvg/thorvg/issues/2693
https://github.com/thorvg/thorvg/issues/2691
https://github.com/thorvg/thorvg/issues/2690
https://github.com/thorvg/thorvg/issues/2384
https://github.com/thorvg/thorvg/issues/3074
https://github.com/thorvg/thorvg/issues/3053
https://github.com/thorvg/thorvg/issues/3076

hermet:
Leave the bug fixes as separate issues.

Issue #3053, opened 2024-12-14
gl_engine: incorrect stroke trimming 


hermet:
```
        auto shape1 = tvg::Shape::gen();
        shape1->appendCircle(245, 225, 50, 120);
        shape1->appendCircle(245, 565, 50, 120);
        shape1->fill(0, 50, 155, 100);
        shape1->strokeFill(0, 0, 255);
        shape1->strokeJoin(tvg::StrokeJoin::Round);
        shape1->strokeCap(tvg::StrokeCap::Round);
        shape1->strokeWidth(12);
        shape1->strokeTrim(0.5f, 0.75f, false);

        canvas->push(std::move(shape1));
```

sw:
![image](https://github.com/user-attachments/assets/36e13b7f-cada-4ae6-bcca-1ec0f62e9632)

gl:
![image](https://github.com/user-attachments/assets/9d7abaf9-9eb7-44c3-9667-6cbce5834f28)

hermet:
\+ @RuiwenTang

PR #3123, opened 2025-01-14
common: trimming refactoring

mgrudzinska:
Zrzut ekranu 2025-01-15 o 01 46 12

issue: https://github.com/thorvg/thorvg/issues/3118 https://github.com/thorvg/thorvg/issues/2854 https://github.com/thorvg/thorvg/issues/3053

### DIFF
--- a/src/renderer/sw_engine/tvgSwShape.cpp
+++ b/src/renderer/sw_engine/tvgSwShape.cpp
@@ -341,7 +341,7 @@ static SwOutline* _genDashOutline(const RenderShape* rshape, const Matrix& trans
     dash.cnt = rshape->strokeDash((const float**)&dash.pattern, &offset);
     auto simultaneous = rshape->stroke->trim.simultaneous;
     float trimBegin = 0.0f, trimEnd = 1.0f;
-    if (trimmed) rshape->stroke->strokeTrim(trimBegin, trimEnd);
+    if (trimmed) rshape->stroke->trim.get(trimBegin, trimEnd);
 
     if (dash.cnt == 0) {
         if (trimmed) dash.pattern = (float*)malloc(sizeof(float) * 4);
--- a/src/renderer/sw_engine/tvgSwShape.cpp
+++ b/src/renderer/sw_engine/tvgSwShape.cpp
@@ -226,14 +226,6 @@ static void _dashClose(SwDashStroke& dash, const Matrix& transform)
 }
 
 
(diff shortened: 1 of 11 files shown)


hermet:
Only for the refactoring + sw_engine replace increased binary size 2 (size)  ~ 4kb (o3).

src/renderer/tvgRender.h
```diff
@@ -105,12 +106,7 @@ struct RenderStroke
     StrokeCap cap = StrokeCap::Square;
     StrokeJoin join = StrokeJoin::Bevel;
     bool strokeFirst = false;
-
-    struct {
-        float begin = 0.0f;
-        float end = 1.0f;
-        bool simultaneous = true;
-    } trim;
+    TrimPath trim;
```
hermet, line 109: Let's insert this between 105 and 106 for memory packing.

hermet (commented):
Suggestion for the smooth refactoring.
1. Please introduce the TrimPath class with the demanded features (methods) first.
2. After that, simply replace the SwEngine with no.1

mgrudzinska:
@hermet I applied your comments, size should be fine now. turned to draft, since I've tested perf and it dropped significantly.

src/renderer/sw_engine/tvgSwShape.cpp
```diff
@@ -244,112 +236,26 @@ static void _dashMoveTo(SwDashStroke& dash, uint32_t offIdx, float offset, const
...
-    } else {
-        dash->pattern[0] = 0;     //zero dash to start with a space.
-        dash->pattern[1] = begin;
-        dash->pattern[2] = end - begin;
-        dash->pattern[3] = length - end;
+    PathCommand* cmds = nullptr;
+    Point* pts = nullptr;
+    uint32_t cmdCnt, ptsCnt;
+
+    if (trimmed) rshape->stroke->trim.trim(rshape->path.cmds, rshape->path.pts, cmds, cmdCnt, pts, ptsCnt);
```
hermet, line 245: I looks the trimming is performed twice for the shape and stroke.... I think we can reduce to 1?

hermet:
> since I've tested perf and it dropped significantly.

@mgrudzinska Or maybe,  If gl/wgpu shows no degradation, we can apply it to them first and later software.

mgrudzinska:
https://github.com/thorvg/thorvg/pull/3151 contains these changes and more

PR #3134, opened 2025-01-17
sw_engine: fix bug introduced in 07e73a9

mgrudzinska:
The value of the variable 'i' was not modified
after the refactor, resulting in a reference to
an incorrect array element.

from svg samples:
before:
Zrzut ekranu 2025-01-17 o 02 08 52

after:
Zrzut ekranu 2025-01-17 o 02 14 21

### DIFF
--- a/src/renderer/sw_engine/tvgSwStroke.cpp
+++ b/src/renderer/sw_engine/tvgSwStroke.cpp
@@ -844,6 +844,7 @@ bool strokeParseOutline(SwStroke* stroke, const SwOutline& outline)
     ARRAY_FOREACH(p, outline.cntrs) {
         auto last = *p;           //index of last point in contour
         auto limit = outline.pts.data + last;
+        ++i;
 
         //Skip empty points
         if (last <= first) {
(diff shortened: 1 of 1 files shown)


src/renderer/sw_engine/tvgSwStroke.cpp
```diff
@@ -860,7 +861,7 @@ bool strokeParseOutline(SwStroke* stroke, const SwOutline& outline)
         if (type == SW_CURVE_TYPE_CUBIC) return false;
         ++types;
 
-        auto closed =  outline.closed.data ? outline.closed.data[i]: false;
+        auto closed =  outline.closed.data ? outline.closed.data[i - 1]: false;
```
hermet, line 864: Thanks, may I ask why you changed the logic? I think this should remain unchanged to preserve the original logic.

mgrudzinska: Because I added `++i` at the very beginning of the loop, so the first value of i is 1. This is because we no longer increment `i` in the for statement but inside the loop body. Since there is a continue statement, I didn’t want to repeat `++i` in two places. Instead, I ensured it always happens by placing it at the very beginning of the loop, but as a result, I had to adjust this line.
do you think it will be more clear to keep as it was and repeat `++i` in two places?

hermet (approved):

PR #3139, opened 2025-01-20
wg_engine: fix bug introduced in 07e73a9

mgrudzinska:
Fixed the for loop condition to properly check
the iterator against the container's end, ensuring it terminates correctly.


src/renderer/wg_engine/tvgWgCompositor.cpp
```diff
@@ -671,9 +671,9 @@ void WgCompositor::renderClipPath(WgContext& context, WgRenderDataPaint* paint)
     wgpuRenderPassEncoderSetPipeline(renderPassEncoder, pipelines.copy_stencil_to_depth);
     renderData0->meshDataBBox.drawFan(context, renderPassEncoder);
     // merge clip pathes with AND logic
-    for (auto p = paint->clips.begin() + 1; paint->clips.end(); ++p) {
+    for (auto clip = paint->clips.begin() + 1; clip < paint->clips.end(); ++clip) {
```
mgrudzinska, line 674: after ARRAY_FOREACH is introduced 'p' is also used in l.682

hermet (commented):
@mgrudzinska Could you please share the issue you observed? Any nested scrope (683) variable has higher priority.

mgrudzinska:
> @mgrudzinska Could you please share the issue you observed? Any nested scrope (683) variable is allowed, with a higher priority.

I know... that's why in the first version I didn't apply this change... now I'm not able to reproduce, so I removed this part. If I have this problem again I'll open an issue. sorry

SergeyLebedkin (approved):

hermet (approved):

PR #3145, opened 2025-01-21
gl_engine: add support for stroke/fill ordering

mgrudzinska:
sample:
```
        if (!canvas) return false;

        //Shape 1
        shape1->appendRect(50, 50, 200, 200);
        shape1->fill(250, 0, 0);
        shape1->strokeFill(0, 255, 0);
        shape1->strokeWidth(30);                       
        shape1->order(true);

        canvas->push(shape1);

        //Shape 2
        shape2->appendRect(200, 150, 200, 200);
        shape2->fill(0, 0, 255);
        shape2->strokeFill(0, 255, 0);
        shape2->strokeWidth(30);
        shape2->order(false);

        canvas->push(shape2);
```

before (left - sw, right - gl):
Zrzut ekranu 2025-01-21 o 01 54 04

after (left - sw, right - gl):
Zrzut ekranu 2025-01-21 o 01 53 43

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -1093,24 +1093,35 @@ bool GlRenderer::renderShape(RenderData data)
 
     if (flags & (RenderUpdateFlag::Stroke | RenderUpdateFlag::GradientStroke)) drawDepth2 = currentPass()->nextDrawDepth();
 
-
     if (!sdata->clips.empty()) drawClip(sdata->clips);
 
-    if (flags & (RenderUpdateFlag::Color | RenderUpdateFlag::Gradient))
-    {
-        auto gradient = sdata->rshape->fill;
-        if (gradient) drawPrimitive(*sdata, gradient, RenderUpdateFlag::Gradient, drawDepth1);
-        else if (sdata->rshape->color.a > 0) drawPrimitive(*sdata, sdata->rshape->color, RenderUpdateFlag::Color, drawDepth1);
-    }
-
-    if (flags & (RenderUpdateFlag::Stroke | RenderUpdateFlag::GradientStroke))
-    {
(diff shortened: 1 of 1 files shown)


mgrudzinska:
@RuiwenTang I was wondering if simply setting the depth value passed to ```drawPrimitive``` should be sufficient. However, I tested it, and it doesn’t work. 
Could you please comment on whether the changes I made are actually necessary in this form, or if it should be enough to just set ```drawDepth1``` and ```drawDepth2``` properly?

RuiwenTang:
@mgrudzinska If you wish to use the depth value, the depth writable nees to turn on.
 I closed it in begin render pass:https://github.com/thorvg/thorvg/blob/main/src/renderer/gl_engine/tvgGlRenderTask.cpp#L188C9-L188C34

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1083,34 +1083,39 @@ bool GlRenderer::renderShape(RenderData data)
...
 
-    int32_t drawDepth1 = 0, drawDepth2 = 0;
-
     size_t flags = static_cast<size_t>(sdata->updateFlag);
 
     if (flags == 0) return false;
 
-    if (flags & (RenderUpdateFlag::Gradient | RenderUpdateFlag::Color)) drawDepth1 = currentPass()->nextDrawDepth();
-
-    if (flags & (RenderUpdateFlag::Stroke | RenderUpdateFlag::GradientStroke)) drawDepth2 = currentPass()->nextDrawDepth();
```
RuiwenTang, line 1094: BTW, the normal color depth value needs to be smaller than clip depth. So it needs to be specified before drawClip

mgrudzinska: fixed, thx

RuiwenTang (approved):

hermet (approved):

PR #3147, opened 2025-01-21
chore(build): exclude webgpu_native for emscripten build

theashraf:
- Add platform check to exclude webgpu_native in meson.build
- Avoid requiring -Dbindings=wasm_beta for Emscripten builds


tinyjin (approved):

hermet:
@SergeyLebedkin @tinyjin we developed `wg_engine` not only for web platform but also for native platforms, so we can test it on the native systems.

hermet:
sorry my misunderstanding of the code.

PR #3149, opened 2025-01-21
svg_loader: loc--

mgrudzinska:
Removed unnecessary initialization of variables
to 0/false in memory allocated by calloc.

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1475,36 +1475,22 @@ static SvgNode* _createNode(SvgNode* parent, SvgNodeType type)
         return nullptr;
     }
 
-    //Update the default value of stroke and fill
-    //https://www.w3.org/TR/SVGTiny12/painting.html#SpecifyingPaint
-    node->style->fill.paint.none = false;
-    //Default fill opacity is 1
-    node->style->fill.opacity = 255;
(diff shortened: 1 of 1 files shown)


JSUYA (approved):
I think you can do the same for lines that initialize to 0.f values, like 1616 ~ 1617 and 1633 ~ 1634.

mgrudzinska:
> I think you can do the same for lines that initialize to 0.f values, like 1616 ~ 1617 and 1633 ~ 1634.

I didn't do it because of:
[Initialization to all bits zero does not guarantee that a floating-point or a pointer would be initialized to 0.0 and the null pointer value, respectively (although that is true on all common platforms).](https://en.cppreference.com/w/c/memory/calloc)

but it seems very unrealistic that ThorVG will be used on a system that do not support IEEE754 std. 
@hermet what do you think?

hermet:
> very unrealistic

I'm sold on @JSUYA.

mgrudzinska:
I removed the lines @JSUYA mentioned + nullptr:
```
    loader->svgParse->node->node.text.fontFamily = nullptr;
    loader->svgParse->node->node.text.text = nullptr;
```

also the def vaules in `_createNode` - separate commit (for readability while reviewing - I'd like to squash them)

hermet (approved):
@mgrudzinska please submit.

Issue #3116, opened 2025-01-13
api: revise ThorVG 1.0 official APIs


hermet:
Here are the proposed candidates to be included in the v1.0 specification.

**General**
- [ ] Add a Canvas parameter to all paints gen()
- [ ] Result Initializer::flush()  //clear internal caches
- [ ] Result Paint::clip(Paint* clipper) -> Result Scene::clip(Paint* clipper)
- [ ] Result Paint::mask(Paint* target, MaskMethod method) -> Result Scene::mask(Paint* target, MaskMethod method) https://github.com/thorvg/thorvg/issues/2831
- [ ] MaskMethod Paint::mask(const Paint** target) const -> MaskMethod Scene::mask(const Paint** target) const https://github.com/thorvg/thorvg/issues/2831
- [ ] Documentation Installation https://github.com/thorvg/thorvg/issues/2075
- [ ] Revert appendArc?  https://github.com/thorvg/thorvg/issues/2632#issuecomment-2736309716
- [x] Revise the Lottie Slot APIs https://github.com/thorvg/thorvg/issues/2591
- [x] GL runtime binding https://github.com/thorvg/thorvg/pull/2866
- [x] Result Shape::fillColor(uint8_t* r, uint8_t* g, uint8_t* b, uint8_t* a = nullptr) const -> Result Shape::fill(uint8_t* r, uint8_t* g, uint8_t* b, uint8_t* a = nullptr) const
- [x] Result Shape::appendRect(float x, float y, float w, float h, float rx = 0, float ry = 0) -> Result Shape::appendRect(float x, float y, float w, float h, float rx = 0, float ry = 0, bool clockwise = true) https://github.com/thorvg/thorvg/issues/3179
- [x] Result Shape::appendCircle(float cx, float cy, float rx, float ry) -> Result Shape::appendCircle(float cx, float cy, float rx, float ry, bool clockwise = true) https://github.com/thorvg/thorvg/issues/3179
- [x] Remove SwCanvas::MempoolPolicy
- [x] Result Paint::bounds(float* x, float* y, float* w, float* h, bool transformed) -> Result Paint::bounds(float* x, float* y, float* w, float* h), Result Paint::bounds(Point* pt4) https://github.com/thorvg/thorvg/issues/3290
- [x] Add Result Paint::origin(float x, float y);

**Path**
- [ ] Introduce a Path Class https://github.com/thorvg/thorvg/issues/3121
[...]

PR #3150, opened 2025-01-21
api: revised the animation segment specification

hermet:
Changed the unit of the segment from a normalized value to frame numbers, ensuring alignment with other frame control interfaces.

Note that This change may break backward compatibility.

issue: https://github.com/thorvg/thorvg/issues/3116


hermet:
@theashraf Hello, could you please confirm if the dotLottie player uses the tvg_animation_xxx_segment() API? If so, is this change critical for it? Wanna confirm if this could affect to the dotlottie user backward compatibility.

image

PR #3151, opened 2025-01-23
common: trimmed fill

mgrudzinska:
refactoring + https://github.com/thorvg/thorvg/issues/3053 fixed

### DIFF
--- a/src/renderer/tvgTrimPath.cpp
+++ b/src/renderer/tvgTrimPath.cpp
@@ -249,23 +249,8 @@ static void _trim(const PathCommand* inCmds, uint32_t inCmdsCnt, const Point* in
 }
 
 
-/************************************************************************/
-/* External Class Implementation                                        */
-/************************************************************************/
-
-
-bool TrimPath::valid() const
+static void _get(float& begin, float& end)
 {
-    if (begin == 0.0f && end == 1.0f) return false;
-    return true;
-}
-
-
-bool TrimPath::get(float& begin, float& end) const
-{
-    begin = this->begin;
-    end = this->end;
-
     auto loop = true;
 
     if (begin > 1.0f && end > 1.0f) loop = false;
@@ -278,14 +263,25 @@ bool TrimPath::get(float& begin, float& end) const
     if (end < 0.0f) end += 1.0f;
 
     if ((loop && begin < end) || (!loop && begin > end)) std::swap(begin, end);
+}
+
+
+/************************************************************************/
+/* External Class Implementation                                        */
+/************************************************************************/
+
+
+bool TrimPath::valid() const
+{
+    if (begin == 0.0f && end == 1.0f) return false;
     return true;
 }
 
 
 bool TrimPath::trim(const RenderPath& in, RenderPath& out) const
 {
-    float begin, end;
-    get(begin, end);
+    float begin = this->begin, end = this->end;
+    _get(begin, end);
 
     if (in.pts.count < 2 || tvg::zero(begin - end)) return false;
 
--- a/src/renderer/tvgTrimPath.cpp
+++ b/src/renderer/tvgTrimPath.cpp
@@ -1,0 +1,338 @@
+/*
+ * Copyright (c) 2025 the ThorVG project. All rights reserved.
+
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+
+#include "tvgTrimPath.h"
+#include "tvgMath.h"
+#include "tvgRender.h"
+
+#define EPSILON 1e-4f
+
+/************************************************************************/
+/* Internal Class Implementation                                        */
+/************************************************************************/
+
+static float _pathLength(const PathCommand* cmds, uint32_t cmdsCnt, const Point* pts, uint32_t ptsCnt)
+{
+    if (ptsCnt < 2) return 0.0f;
+
+    auto start = pts;
+    auto totalLength = 0.0f;
+
+    while (cmdsCnt-- > 0) {
+        switch (*cmds) {
+            case PathCommand::Close: {
+                totalLength += length(pts - 1, start);
+                break;
+            }
+            case PathCommand::MoveTo: {
+                start = pts;
+                ++pts;
+                break;
+            }
+            case PathCommand::LineTo: {
+                totalLength += length(pts - 1, pts);
+                ++pts;
+                break;
+            }
+            case PathCommand::CubicTo: {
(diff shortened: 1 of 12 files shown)


mgrudzinska:
@hermet @RuiwenTang at this point for shapes with trimmed stroke and fill, the trimming in gl_engine is done twice, once for the stroke and once for the fill. I will optimize this, but after the review for sw_engine and wg_engine.

hermet:
@mgrudzinska Thanks, could you please separate the PR for TrimPath unification and TrimPath fill support? We need to carefully review and adopt the unification refactoring first.

Also, here is the test result, applied only with the first 4 unification commits:

Size Diff: `-loaders=all` and `-Os` options:
```
Sw Engine: 616,681 -> 618,291  (+1.6kb)
All Engines:: 777,524 -> 775,068 (-2.5kb)
```

FPS: No diff, within the margin of error.

The overall result is good. We can probably review the size during the code review. Thanks.

mgrudzinska:
@hermet I removed the commits with trimming applied to the fill.

previously I tested the size only on mac, and it is the same for sw_engine (main vs pr).
now I tested also on linux vm and I see the difference in size, so I can try to fix this

src/renderer/tvgTrimPath.cpp
```diff
@@ -205,8 +206,8 @@ static void _trimPath(const PathCommand* inCmds, uint32_t inCmdsCnt, const Point
     bool start = true;
     for (uint32_t i = 0; i < inCmdsCnt; ++i) {
         auto dLen = _length();
-        //skip the segment
-        if (len + dLen <= trimStart || len > trimEnd) {}
+        //skip the segment (also skip segments with len < EPSILON
```
hermet, line 209: Please add a descriptive comment about epsilon at the code.
```
Due to the finite precision of Bezier curve subdivision and length calculation (1e-2), 
trimming may produce very short segments that would effectively have zero length with higher computational accuracy.
This helps to prevent unwanted artifacts during dashing.
```

src/renderer/tvgTrimPath.cpp
```diff
@@ -0,0 +1,337 @@
...
+#include "tvgMath.h"
+
+
+/************************************************************************/
+/* Internal Class Implementation                                        */
+/************************************************************************/
+
+static float _pathLength(const PathCommand* inCmds, uint32_t inCmdsCnt, const Point* inPts, uint32_t inPtsCnt)
+{
+    if (inCmdsCnt <= 0 || inPtsCnt <= 0) return 0.0f;
```
hermet, line 35: Unmatched commands and points are not allowed here. I think this condition is sufficient.
```
    if (inPtsCnt < 2) return 0.0f;
```

src/renderer/tvgTrimPath.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+    auto start = inPts;
+    auto totalLength = 0.0f;
+
+    uint32_t i = 0;
+    while (i < inCmdsCnt) {
```
hermet, line 41: suggest `for()` for less LOC.

src/renderer/tvgTrimPath.cpp
```diff
@@ -0,0 +1,337 @@
...
+                break;
+        }
+        ++cmds;
+    };
+
+    bool start = true;
+    for (uint32_t i = 0; i < inCmdsCnt; ++i) {
+        auto dLen = _length();
+        //skip the segment
+        if (len + dLen <= trimStart || len > trimEnd) {}
```
hermet: Please unify the condition
```
        if (len + dLen <= trimStart || len > trimEnd) {}
        else if (len <= trimStart) {
```

src/renderer/tvgTrimPath.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+/************************************************************************/
+/* External Class Implementation                                        */
+/************************************************************************/
+
+
+bool TrimPath::valid() const
+{
+    if (begin == 0.0f && end == 1.0f) return false;
+    if (fabsf(end - begin) >= 1.0f) return false;
```
hermet, line 253: I guess this exception could be handled at the StrokeTrim API level, rather than here.

src/renderer/tvgTrimPath.cpp
```diff
@@ -0,0 +1,337 @@
...
+    return true;
+}
+
+
+bool TrimPath::get(float& begin, float& end) const
+{
+    begin = this->begin;
+    end = this->end;
+
+    if (fabsf(end - begin) >= 1.0f) {
```
hermet, line 263: same "is this a valid exception?"

src/renderer/tvgTrimPath.cpp
```diff
@@ -0,0 +1,337 @@
...
+        begin = 0.0f;
+        end = 1.0f;
+        return false;
+    }
+
+    auto loop = true;
+
+    if (begin > 1.0f && end > 1.0f) loop = false;
```
hermet, line 271: Please double-check if out-of-range trimming exceptions are really necessary here.

hermet (commented):
Please unify two commits:
- common: introduce epsilon to trim path calculations
- common: introduce TrimPath struct

src/renderer/tvgTrimPath.cpp
mgrudzinska: Didn't mention dashing, as it's not related.
as I wrote in the previous commit msg a similar approach is implemented in the sw engine for dashing (https://github.com/thorvg/thorvg/pull/1932), but these pr doesn't affect anything related to dashing.

src/renderer/tvgTrimPath.cpp
mgrudzinska: I changed to `while (cmdsCnt-- > 0)`

src/renderer/tvgTrimPath.cpp
mgrudzinska: handled in api now, so removed from here

src/renderer/tvgTrimPath.cpp
mgrudzinska: yes, because in lottie there is offset value (and in api we let for all possible values of begin and end)

[sample.json](https://github.com/user-attachments/files/18679959/sample.json)
tvg vs lottie-web:
tvg_trim
lottie_web_trim

in the file:
- begin: 90%
- end: 0%-100%
- offset: 180%
so as the result during this animation `begin` changes in the 0.5-1.4 range and `end` changes in the 1.4-1.5 range

src/renderer/tvgTrimPath.cpp
mgrudzinska: @hermet please check now

src/renderer/tvgTrimPath.h
```diff
@@ -0,0 +1,47 @@
...
+{
+    float begin = 0.0f;
+    float end = 1.0f;
+    bool simultaneous = true;
+
+    TrimPath() = default;
+    TrimPath(float b, float e, bool sim) : begin(b), end(e), simultaneous(sim) {}
+    bool valid() const;
+    bool get(float& begin, float& end) const;
+    bool trim(const Array<PathCommand>& inCmds, const Array<Point>& inPts, Array<PathCommand>& outCmds, Array<Point>& outPts) const;
```
hermet, line 41: Please consider this prototype if this is compatible (other functions as well)
```
bool trim(const RenderPath& in, RenderPath& out) const;
_trimPath(..., Array<PathCommand>& outCmds, Array<Point>& outPts)  -> _trimPath(..., RenderPath& out)
```

src/renderer/tvgTrimPath.h
```diff
@@ -0,0 +1,47 @@
...
+namespace tvg
+{
+
+struct TrimPath
+{
+    float begin = 0.0f;
+    float end = 1.0f;
+    bool simultaneous = true;
+
+    TrimPath() = default;
```
hermet, line 38: Please remove unnecessary constructors (looks they are).

hermet:
@SergeyLebedkin please confirm the wgpu changes, thanks.

hermet:
@SergeyLebedkin Hello, please confirm the wgpu changes, thanks.
@RuiwenTang Hello, I'd appreciate it if you could confirm the GL changes. Thanks!

src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -161,56 +161,52 @@ struct WgVertexBuffer
...
+        } else {
+            cmds = rshape.path.cmds.data;
+            cmdCnt = rshape.path.cmds.count;
+            pts = rshape.path.pts.data;
+        }
+
         size_t pntIndex = 0;
-        ARRAY_FOREACH(p, rshape.path.cmds) {
-            auto cmd = *p;
+        for (uint32_t i = 0; i < cmdCnt; i++) {
```
SergeyLebedkin, line 192: ARRAY_FOREACH?

mgrudzinska: no, since now we can walk either through `rshape.path` or through `trimmedPath`

src/renderer/tvgTrimPath.h
mgrudzinska: so you prefer to add def values to the other constructor?

SergeyLebedkin (approved):

PR #3152, opened 2025-01-23
loader: WebP dynamic loader++

tinyjin:
```sh
meson . build -Dexamples="true" -Dloaders="all" -Dsavers="all" -Dmodule=true
```


hermet:
@tinyjin I think `WebpLoader` should be a layer for the static/dynamic linking. You don't need to wrapping around the webp library itself.

PR #3155, opened 2025-01-23
wg_engine: fix segmentation fault on windows with SVG examples

SergeyLebedkin:
Some SVG files fails during stoke dashing on windows.
Fixed

### DIFF
--- a/src/renderer/wg_engine/tvgWgGeometry.h
+++ b/src/renderer/wg_engine/tvgWgGeometry.h
@@ -241,6 +241,8 @@ struct WgVertexBufferInd
     size_t vcount = 0;
     size_t icount = 0;
     float tscale = 1.0f;
+    // intermediate buffer for stroke dashing
+    WgVertexBuffer dashed;
 
     // reset buffer
     void reset(float scale)
@@ -284,7 +286,6 @@ struct WgVertexBufferInd
     void appendStrokesDashed(const WgVertexBuffer& buff, const RenderStroke* rstroke)
     {
         // dashed buffer
-        WgVertexBuffer dashed;
         dashed.reset(tscale);
         // ignore single points polyline
         if (buff.vcount < 2) return;


src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -284,7 +284,7 @@ struct WgVertexBufferInd
     void appendStrokesDashed(const WgVertexBuffer& buff, const RenderStroke* rstroke)
     {
         // dashed buffer
-        WgVertexBuffer dashed;
+        static WgVertexBuffer dashed;
```
hermet, line 287: Looks odd... Moreover, you can probably find a better solution than using a static declaration in a method...

SergeyLebedkin: another solution provided

hermet:
@SergeyLebedkin I guess that exceeded the stack memory limit. It's better to use an array instead of a stack memory.

image

SergeyLebedkin:
> @SergeyLebedkin I think this exceeded the stack memory limit or ran out of the memory pool. It's better to use an array instead of a fixed size buffer.
The main idia of static array usage is to avoid system memory reallocations and range cheks for each vertex in a tesseletion process for each shape. Dynamic arrays do both and destroy streaming idea without any additinal operation

hermet:
> > @SergeyLebedkin I think this exceeded the stack memory limit or ran out of the memory pool. It's better to use an array instead of a fixed size buffer.
> 
> The main idia of static array usage is to avoid system memory reallocations and range cheks for each vertex in a tesseletion process for each shape. Dynamic arrays do both and destroy streaming idea without any additinal operation

@SergeyLebedkin sure, here is a question: where the size came from and how to guarantee the size is safe to use?

SergeyLebedkin:
> @SergeyLebedkin sure, here is a question: where the size came from and how to guarantee the size is safe to use?

Currently we have a static array size selected in a two ways: default stack size (for windows and linux they are different) and the extrimely long strokes in some SVG or Lottie thet hits to the array size. We can not predict reguired array size to prevent reallocations

PR #3157, opened 2025-01-23
lottie: handle parsing error

mgrudzinska:
Exporting from After Effects adds an extra array element to the tint effect: object with the effect value of type "Ignored". All other object types support animated values, meaning they are enclosed in parentheses. The "Ignored" object, however, expects a single number as a value. This caused a parsing error, which has been fixed by skipping all other objects after reading the expected 3.

sample:
[tint_effect_swapted.json](https://github.com/user-attachments/files/18528107/tint_effect_swapted.json)
[tint_effect.json](https://github.com/user-attachments/files/18528108/tint_effect.json)

before:
Zrzut ekranu 2025-01-24 o 00 05 45

after:
Zrzut ekranu 2025-01-24 o 00 06 11

note:
lotti-web and skottie are ignoring this extra segment. Importing a lottie file into AE ignores all effects, so no way to check the expected behavior.

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -1242,6 +1242,8 @@ void LottieParser::parseTint(LottieFxTint* effect)
                 ++idx;
             } else skip();
         }
+        //handle extra AE-generated "Ignored" effect value
+        if (idx > 2) skip();
     }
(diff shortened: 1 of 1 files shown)


hermet:
thanks, please see this: https://github.com/thorvg/thorvg/pull/3160

Issue #3054, opened 2024-12-14
gl_engine: Support SceneEffect


hermet:
Make samples work:

GL:
- [x] EffectGaussianBlur
- [x] EffectDropShadow
- [x] Fill
- [x] Tint
- [x] Trinton

WG:
- [x] EffectGaussianBlur
- [x] EffectDropShadow
- [x] Fill
- [x] Tint
- [x] Trinton

hermet:
Please note that GaussianBlur and DropShadow have quality options in the SceneEffect interface. These are optional parameters designed to balance quality and performance based on user demand and are not mandatory to support. This is particularly relevant if GPU processing could be efficient enough to maintain animations exceeding 120fps.

hermet:
@SergeyLebedkin awesome, thanks for your great effort!

PR #3158, opened 2025-01-24
engine: introduce scene effect dispose render method

SergeyLebedkin:
to correctly delete RenderData from effects it must be disposed in the renderer 
a new method has been added to the RenderMethod interface

https://github.com/thorvg/thorvg/issues/3054

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -968,6 +968,12 @@ bool GlRenderer::render(TVG_UNUSED RenderCompositor* cmp, TVG_UNUSED const Rende
 }
 
 
+void GlRenderer::dispose(TVG_UNUSED RenderEffect* effect)
+{
+    //TODO: dispose the effect
+}
+
+
 ColorSpace GlRenderer::colorSpace()
 {
     return ColorSpace::Unknown;
--- a/src/renderer/sw_engine/tvgSwRenderer.cpp
+++ b/src/renderer/sw_engine/tvgSwRenderer.cpp
@@ -691,7 +691,7 @@ bool SwRenderer::render(RenderCompositor* cmp, const RenderEffect* effect, bool 
         TVGERR("SW_ENGINE", "Not supported grayscale Gaussian Blur!");
         return false;
     }
-
+    
     switch (effect->type) {
         case SceneEffect::GaussianBlur: {
             return effectGaussianBlur(p, request(surface->channelSize, true), static_cast<const RenderEffectGaussianBlur*>(effect));
@@ -716,6 +716,13 @@ bool SwRenderer::render(RenderCompositor* cmp, const RenderEffect* effect, bool 
         }
         default: return false;
     }
+}
+
+
+void SwRenderer::dispose(RenderEffect* effect) 
+{
+    free(effect->rd);
+    effect->rd = nullptr;
 }
 
 
--- a/src/renderer/sw_engine/tvgSwRenderer.h
+++ b/src/renderer/sw_engine/tvgSwRenderer.h
@@ -65,6 +65,7 @@ public:
     void prepare(RenderEffect* effect, const Matrix& transform) override;
     bool region(RenderEffect* effect) override;
     bool render(RenderCompositor* cmp, const RenderEffect* effect, bool direct) override;
+    void dispose(RenderEffect* effect) override;
 
     static SwRenderer* gen();
     static bool init(uint32_t threads);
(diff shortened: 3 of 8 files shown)


src/renderer/tvgScene.h
```diff
@@ -299,7 +299,10 @@ struct Scene::Impl : Paint::Impl
     Result resetEffects()
     {
         if (effects) {
-            ARRAY_FOREACH(p, *effects) delete(*p);
+            ARRAY_FOREACH(p, *effects) {
+                if ((*p)->rm) ((*p)->rm)->dispose(*p);
```
hermet, line 303: Please consider this
`
renderer->dispose(*p);
`

src/renderer/tvgRender.h
```diff
@@ -258,15 +258,13 @@ struct RenderShape
 
 struct RenderEffect
 {
+    RenderMethod* rm = nullptr;
```
hermet, line 261: not necessarily

src/renderer/sw_engine/tvgSwRenderer.cpp
```diff
@@ -691,7 +692,7 @@ bool SwRenderer::render(RenderCompositor* cmp, const RenderEffect* effect, bool
         TVGERR("SW_ENGINE", "Not supported grayscale Gaussian Blur!");
         return false;
     }
-
+
```
hermet, line 695: --

src/renderer/sw_engine/tvgSwRenderer.cpp
```diff
@@ -670,6 +670,7 @@ void SwRenderer::prepare(RenderEffect* effect, const Matrix& transform)
         case SceneEffect::Tritone: effectTritoneUpdate(static_cast<RenderEffectTritone*>(effect)); break;
         default: break;
     }
+    effect->rm = this;
```
hermet, line 673: --

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -968,6 +968,12 @@ bool GlRenderer::render(TVG_UNUSED RenderCompositor* cmp, TVG_UNUSED const Rende
 }
 
 
+void GlRenderer::dispose(RenderEffect* effect)
```
hermet, line 971: `TVG_UNUSED RenderEffect* effect`

src/renderer/tvgRender.h
```diff
@@ -258,15 +258,13 @@ struct RenderShape
...
     RenderData rd = nullptr;
     RenderRegion extend = {0, 0, 0, 0};
     SceneEffect type;
     bool valid = false;
 
-    virtual ~RenderEffect()
```
hermet, line 266: please consider the original logic. freeing the member data here seems more intuitive.

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -483,6 +483,12 @@ bool WgRenderer::render(TVG_UNUSED RenderCompositor* cmp, TVG_UNUSED const Rende
 }
 
 
+void WgRenderer::dispose(RenderEffect* effect)
```
hermet, line 486: `void WgRenderer::dispose(TVG_UNUSED RenderEffect* effect)`

src/renderer/sw_engine/tvgSwRenderer.h
```diff
@@ -65,6 +65,7 @@ class SwRenderer : public RenderMethod
     void prepare(RenderEffect* effect, const Matrix& transform) override;
     bool region(RenderEffect* effect) override;
     bool render(RenderCompositor* cmp, const RenderEffect* effect, bool direct) override;
+    void dispose(RenderEffect* effect) override;
```
hermet, line 68: Please leave as empty method here since software wouldn't use this dispose() method.
```
void dispose(TVG_UNUSED RenderEffect* effect) override {}
```

src/renderer/sw_engine/tvgSwRenderer.cpp
```diff
@@ -719,6 +720,12 @@ bool SwRenderer::render(RenderCompositor* cmp, const RenderEffect* effect, bool
 }
 
 
+void SwRenderer::dispose(RenderEffect* effect) 
+{
+    free(effect->rd);
```
hermet, line 725: I think free the rd is better inside the `RenderEffect` destructor

hermet (changes requested):
@SergeyLebedkin please check comments

src/renderer/sw_engine/tvgSwRenderer.cpp
SergeyLebedkin: No... Logic is that rd pointer is renderer depender data structure. f.e. for sw_renderer is SwGaussianBlur and it creates in sw_engine, so it must be freed by sw_renderer.
same for others renderers. it will containe API dependedent handles and must be desposed inside renderer, but not as a general data structure

PR #3163, opened 2025-01-28
wg_engine/sw_engine: fix color burn/dodge edge cases

mgrudzinska:
samples:
[blurColorBurn.json](https://github.com/user-attachments/files/18572504/blurColorBurn.json)
[blurColorDodge.json](https://github.com/user-attachments/files/18572505/blurColorDodge.json)

before:
Zrzut ekranu 2025-01-28 o 12 22 09

after:
Zrzut ekranu 2025-01-28 o 12 00 41
Zrzut ekranu 2025-01-28 o 12 06 50

AE:
Zrzut ekranu 2025-01-28 o 12 20 06
Zrzut ekranu 2025-01-28 o 12 20 21

note: skottie results are as from AE, lottie-web are a bit different

### DIFF
--- a/src/renderer/sw_engine/tvgSwCommon.h
+++ b/src/renderer/sw_engine/tvgSwCommon.h
@@ -448,9 +448,9 @@ static inline uint32_t opBlendColorDodge(uint32_t s, uint32_t d, TVG_UNUSED uint
 {
     // d / (1 - s)
     s = 0xffffffff - s;
-    auto c1 = (C1(s) == 0) ? C1(d) : std::min(C1(d) * 255 / C1(s), 255);
-    auto c2 = (C2(s) == 0) ? C2(d) : std::min(C2(d) * 255 / C2(s), 255);
-    auto c3 = (C3(s) == 0) ? C3(d) : std::min(C3(d) * 255 / C3(s), 255);
+    auto c1 = C1(d) == 0 ? 0 : (C1(s) == 0 ? 255 : std::min(C1(d) * 255 / C1(s), 255));
+    auto c2 = C2(d) == 0 ? 0 : (C2(s) == 0 ? 255 : std::min(C2(d) * 255 / C2(s), 255));
+    auto c3 = C3(d) == 0 ? 0 : (C3(s) == 0 ? 255 : std::min(C3(d) * 255 / C3(s), 255));
     return JOIN(255, c1, c2, c3);
 }
 
@@ -458,9 +458,9 @@ static inline uint32_t opBlendColorBurn(uint32_t s, uint32_t d, TVG_UNUSED uint8
 {
     // 1 - (1 - d) / s
     auto id = 0xffffffff - d;
-    auto c1 = (C1(s) == 0) ? C1(d) : 255 - std::min(C1(id) * 255 / C1(s), 255);
-    auto c2 = (C2(s) == 0) ? C2(d) : 255 - std::min(C2(id) * 255 / C2(s), 255);
-    auto c3 = (C3(s) == 0) ? C3(d) : 255 - std::min(C3(id) * 255 / C3(s), 255);
+    auto c1 = C1(d) == 255 ? 255 : (C1(s) == 0 ? 0 : 255 - std::min(C1(id) * 255 / C1(s), 255));
+    auto c2 = C2(d) == 255 ? 255 : (C2(s) == 0 ? 0 : 255 - std::min(C2(id) * 255 / C2(s), 255));
+    auto c3 = C3(d) == 255 ? 255 : (C3(s) == 0 ? 0 : 255 - std::min(C3(id) * 255 / C3(s), 255));
 
     return JOIN(255, c1, c2, c3);
 }
--- a/src/renderer/gl_engine/tvgGlShaderSrc.cpp
+++ b/src/renderer/gl_engine/tvgGlShaderSrc.cpp
@@ -671,9 +671,9 @@ const char* COLOR_DODGE_BLEND_FRAG = COMPLEX_BLEND_HEADER R"(
         vec4 dstColor = texture(uDstTexture, vUV);
 
         FragColor = vec4(
-            1.0 - srcColor.r > 0.0 ? dstColor.r / (1.0 - srcColor.r) : dstColor.r,
-            1.0 - srcColor.g > 0.0 ? dstColor.g / (1.0 - srcColor.g) : dstColor.g,
-            1.0 - srcColor.b > 0.0 ? dstColor.b / (1.0 - srcColor.b) : dstColor.b,
+            srcColor.r < 1.0 ? dstColor.r / (1.0 - srcColor.r) : (dstColor.r > 0.0 ? 1.0 : 0.0),
+            srcColor.g < 1.0 ? dstColor.g / (1.0 - srcColor.g) : (dstColor.g > 0.0 ? 1.0 : 0.0),
+            srcColor.b < 1.0 ? dstColor.b / (1.0 - srcColor.b) : (dstColor.b > 0.0 ? 1.0 : 0.0),
             1.0
         );
     }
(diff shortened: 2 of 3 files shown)


mgrudzinska:
> @mgrudzinska Please confirm the equation in the document. If the equation is incorrect, feel free to correct it. However, if it is accurate, it should be respected. I wonder if you are modifying the composition rule but not the equation itself (see: #3072)
> 
> ```
> ColorDodge,        ///< Divides the bottom layer by the inverted top layer. D / (255 - S)
> ColorBurn,         ///< Divides the inverted bottom layer by the top layer, and then inverts the result. 255 - (255 - D) / S
> ```

the equation itself is correct. the implementation didn't cover the edge cases for D=0 and D=255, that's all

hermet:
> edge cases for S=0 and S=255

This absolutely changes the equation that alters the color; it doesn't seem like just an exception handling issue like 0 division. Please suggest more clues about where this exception is coming from.
image

src/renderer/sw_engine/tvgSwCommon.h
```diff
@@ -446,21 +446,29 @@ static inline uint32_t opBlendLighten(uint32_t s, uint32_t d, TVG_UNUSED uint8_t
 
 static inline uint32_t opBlendColorDodge(uint32_t s, uint32_t d, TVG_UNUSED uint8_t a)
 {
-    // d / (1 - s)
+    // - general formula: d / (1 - s)
+    // - with edge cases
+    // if (d == 0) return 0
+    // else if (s == 1) return 1
+    // else return min(1, d / (1 - s))
```
mgrudzinska, line 453: @hermet please check this comment. 
the original formula is correct and valid. the edge cases have to be handled separately, that's all.

for `s=1` we get `d/0` which is `inf`. since we are in the 0-255 range, so the infinity corresponds to the biggest number, so 1 (or 255 in our case).

also a special case is when `s=1` and `d=0` in the same time, since we get `0/0`. in such case the result is 0 (confirmed in AE). 

similar logic is applied to color burn - same as here, the formula is correct, the edge cases have to be handled correctly.

hermet: Thanks. Here is my humble suggestion: 

You’re only providing before/after images without any explanation of what the issue was in the logic or how you fixed it (Please review again your commit and PR messages). The thing is, it's not a simple to figure out the root cause from just result screen shots. Reviewers have to reverse-engineer the code to understand the root problem and how it was resolved. If you describe the brief context of the issue you faced -like the comment you added here or just saying like 0 division issue or something like it—Reviewers can save huge time by avoiding unnecessary investigation for the problem context from your change and from the back-and-forth communication. This isn't just for me; it’s to help you as well.

src/renderer/sw_engine/tvgSwCommon.h
```diff
@@ -446,21 +446,29 @@ static inline uint32_t opBlendLighten(uint32_t s, uint32_t d, TVG_UNUSED uint8_t
...
     s = 0xffffffff - s;
-    auto c1 = (C1(s) == 0) ? C1(d) : std::min(C1(d) * 255 / C1(s), 255);
-    auto c2 = (C2(s) == 0) ? C2(d) : std::min(C2(d) * 255 / C2(s), 255);
-    auto c3 = (C3(s) == 0) ? C3(d) : std::min(C3(d) * 255 / C3(s), 255);
+    auto c1 = C1(d) == 0 ? 0 : (C1(s) == 0 ? 255 : std::min(C1(d) * 255 / C1(s), 255));
```
hermet, line 455: `C1(s) == 0 ? 255` does this necessary? Otherwise let's simplify the logic.

src/renderer/sw_engine/tvgSwCommon.h
```diff
@@ -446,21 +446,29 @@ static inline uint32_t opBlendLighten(uint32_t s, uint32_t d, TVG_UNUSED uint8_t
...
+    // - general formula: 1 - (1 - d) / s
+    // - with edge cases:
+    // if (d == 1) return 1
+    // else if (s == 0) return 0
+    // else return 1 - min(1, ( 1 - d) / s)
     auto id = 0xffffffff - d;
-    auto c1 = (C1(s) == 0) ? C1(d) : 255 - std::min(C1(id) * 255 / C1(s), 255);
-    auto c2 = (C2(s) == 0) ? C2(d) : 255 - std::min(C2(id) * 255 / C2(s), 255);
-    auto c3 = (C3(s) == 0) ? C3(d) : 255 - std::min(C3(id) * 255 / C3(s), 255);
+    auto c1 = C1(d) == 255 ? 255 : (C1(s) == 0 ? 0 : 255 - std::min(C1(id) * 255 / C1(s), 255));
```
hermet, line 469: In your logic, `C1(s) == 0 ? 0` sholudn't be `C1(s) == 0 ? 255` ?

hermet (commented):
1. Please check comments
2. Is GL engine ok?

src/renderer/sw_engine/tvgSwCommon.h
mgrudzinska: without this we have division by 0: ```C1(d) * 255 / C1(s)```
and runtime error:
```
../src/renderer/sw_engine/tvgSwCommon.h:455:54: runtime error: division by zero
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/renderer/sw_engine/tvgSwCommon.h:455:54 in 
AddressSanitizer:DEADLYSIGNAL
```

mgrudzinska:
@hermet gl included

hermet:
> @hermet gl included, but the sample (files included in the pr description) is not working - no blur is applied (same on main, so the issue lies elsewhere).

@mgrudzinska Thanks! One last confirmation—could you please check by reproducing the same problem using simple shape 
 (without any scenes) blending in our example?

mgrudzinska:
> @mgrudzinska Thanks. One last confirmation—could you please check by reproducing the same problem using simple shape (without any scenes) blending in our example?

@hermet ofc, please check the modified content of the _Blending.cpp_ sample:
```
    void blender(tvg::Canvas* canvas, const char* name, tvg::BlendMethod method, float x, float y, uint32_t* data)
    {
        tvg::Fill::ColorStop colorStops1[2];
        colorStops1[0] = {0.3, 0, 0, 0, 255};
        colorStops1[1] = {0.7, 255, 255, 255, 255};

        tvg::Fill::ColorStop colorStops2[4];
        colorStops2[0] = {0.2, 0, 0, 0, 255};
        colorStops2[1] = {0.3, 200, 0, 0, 255};
        colorStops2[2] = {0.7, 0, 0, 200, 255};
        colorStops2[3] = {0.8, 255, 255, 255, 255};

        {
            auto bottom = tvg::Shape::gen();
            bottom->appendRect(20.0f + x, 25.0f + y, 100.0f, 100.0f);
            bottom->fill(200, 100, 50);
            canvas->push(bottom);

            auto top = tvg::Shape::gen();
            top->appendCircle(70.0f + x, 75.0f + y, 45.0f, 45.0f);
            auto fillTop = tvg::LinearGradient::gen();
            fillTop->linear(20.0f + x, 25.0f + y, 20.0f + x + 100.0f, 25.0f + y);
            fillTop->colorStops(colorStops1, 2);
            top->fill(fillTop);

            top->blend(method);
            canvas->push(top);
        }

        {
            auto bottom = tvg::Shape::gen();
            bottom->appendRect(220.0f + x, 25.0f + y, 100.0f, 100.0f);
            fill->linear(220.0f + x, 25.0f + y, 220.0f + x + 100.0f, 25.0f + y);
            fill->colorStops(colorStops2, 4);
            bottom->fill(fill);
            canvas->push(bottom);

            auto top = tvg::Shape::gen();
            top->appendCircle(270.0f + x, 75.0f + y, 45.0f, 45.0f);
            top->fill(255, 255, 255);

            top->blend(method);
            canvas->push(top);
        }

        {
            auto bottom = tvg::Shape::gen();
            bottom->appendRect(420.0f + x, 25.0f + y, 100.0f, 100.0f);
            fill->linear(420.0f + x, 25.0f + y, 420.0f + x + 100.0f, 25.0f + y);
            fill->colorStops(colorStops2, 4);
            bottom->fill(fill);
            canvas->push(bottom);

            auto top = tvg::Shape::gen();
            top->appendCircle(470.0f + x, 75.0f + y, 45.0f, 45.0f);
            top->fill(0, 0, 0);

            top->blend(method);
            canvas->push(top);
        }

    }


    {
        if (!canvas) return false;

        if (!tvgexam::verify(tvg::Text::load(EXAMPLE_DIR"/font/Arial.ttf"))) return false;

        auto bg = tvg::Shape::gen();
        bg->appendRect(0, 0, 800, 800);
        bg->fill(70, 70, 70);
        canvas->push(bg);

        blender(canvas, "Normal", tvg::BlendMethod::Normal, 0.0f, 0.0f, nullptr);
        blender(canvas, "ColorDodge", tvg::BlendMethod::ColorDodge, 0.0f, 120.0f, nullptr);
        blender(canvas, "ColorBurn", tvg::BlendMethod::ColorBurn, 0.0f, 240.0f, nullptr);

        return true;
    }
};
```

before:
Zrzut ekranu 2025-02-1 o 14 23 08

after:
Zrzut ekranu 2025-02-1 o 14 24 10


1. column:
covers scenarios with _the source_ is **black** (the edge cos for _ColorBurn_) and **white** (the edge cos for _ColorDodge_).
in the previous approach for such cases _the destination_  pixel value was returned, which was not in the agreement with the equation

2. column
a **white** _source_ + **black** _destination_ (some part of it) - the other edge case for _ColorDodge_
result from AE (I don't see why for completely white circle the result is as it is, but one can see the there are some artifacts so maybe the result is not trustworthy):
dodge

3.  column
a **black** _source_ + **white** _destination_ (some part of it) - the other edge case for _ColorBurn_
result from AE:
burn

notes:
theory I found:
https://drafts.fxtf.org/compositing/#blendingcolordodge
https://drafts.fxtf.org/compositing/#blendingcolorburn

hermet (approved):

hermet:
Regardless of this PR, GL incorrectly blends the anti-aliased parts.

image

hermet (approved):

PR #3167, opened 2025-01-29
Support Lottie Frame Tweening

hermet:
frame tweening allows user to interpolate two frames
over a speicified duration. This Tweening functionality
can be particularly powerful when combined with state-based
(a.k.a. Marker) animation playback in Lottie.

For example, apps support state machine based animations,
where the transition sequences between states are not linear
and can occur in highly irregular directions.

Experimental APIs:
- Result LottieAnimation::tween(float from, float to, float progress)
- Tvg_Result tvg_lottie_animation_get_marker(Tvg_Animation* animation, float from, float to, float progress)

https://github.com/user-attachments/assets/48e389b2-3fd4-45bb-b4bf-b41191ac68d3

https://github.com/user-attachments/assets/5fdafc60-3226-4a0c-858d-f9c8c55aa82a

https://github.com/user-attachments/assets/427a137b-9e6a-4c46-a95e-ea6e0edb14ec


hermet:
size increased by 3kb (0.5%)

Issue #3148, opened 2025-01-21
core: issue with scaling


mgrudzinska:
sample: [bl.svg.zip](https://github.com/user-attachments/files/18491159/bl.svg.zip)

result of the Svg.cpp example (canvas size 800x800) with settings:

1) visible 'border' - unexpected
- ```picture->scale(0.176f);  //Svg.cpp l.66```

<img width="515" alt="Image" src="https://github.com/user-attachments/assets/89a45523-21dd-42fc-a039-d4062dab9a3b" />

2) no border (as expected)
- ```picture->scale(0.2f);  //Svg.cpp l.66``` (similar results for other scaling values)

<img width="518" alt="Image" src="https://github.com/user-attachments/assets/b7c5d9f1-b4cd-4389-a76e-fd6f414e2684" />

mgrudzinska:
```
        if (!canvas) return false;

        auto global = tvg::Scene::gen();

        // scene with a red rectangle - clipped to top left: (50,50) bottom right: (450,450)
        auto scene = tvg::Scene::gen();

        auto shape1 = tvg::Shape::gen();
        shape1->appendRect(10, 10, 480, 480);
        shape1->fill(255, 0, 0);
        scene->push(shape1);

        auto clip = tvg::Shape::gen();
        clip->appendRect(50, 50, 400, 400);
        scene->clip(clip);

        global->push(scene);

        // translucent rect with exactly the same size as clip - top left: (50,50) bottom right: (450,450)
        auto mat = tvg::Shape::gen();
        mat->appendRect(50, 50, 400, 400);
        mat->fill(255, 255, 255, 200);
        global->push(mat);

        global->scale(0.43f); //different results for different scaling factors

        canvas->push(global);
```

Since both the clip and the overlaid semi-transparent rectangle have identical dimensions, the expected result is a uniform appearance—a red rectangle with a white semi-transparent layer on top. 

`global->scale(0.43f);`

<img width="222" alt="Image" src="https://github.com/user-attachments/assets/0d2b34fe-5f76-4a8e-845b-f9af99a187f5" />

`global->scale(0.44f);`

<img width="212" alt="Image" src="https://github.com/user-attachments/assets/554c4509-fadd-45ae-a30c-95b946531899" />


`global->scale(0.45f);`

<img width="219" alt="Image" src="https://github.com/user-attachments/assets/a2910bb6-20e3-4a41-92c7-718ef1474683" />

Issue #916, opened 2021-10-14
Svg rendering issue. 


mihashco:
The "l" letter is not rendered in the same way as other letters. 

1. ThorVG Viewer
![Thorvg Viewer](https://user-images.githubusercontent.com/4451022/137304630-f14bbf62-323a-413f-b7e5-417ad3297761.png)

2. Ubuntu Images Viewer
![Ubuntu Image Viewer](https://user-images.githubusercontent.com/4451022/137304723-2422bc23-8371-41ae-9995-6f30ef32dcc7.png)

PR #3172, opened 2025-02-01
common: more accurate clipped viewport rounding

mgrudzinska:
Replaced ceil and direct casts with nearbyint to improve viewport size calculation. This ensures consistent approach with the fastTrack case in sw_engine.

@Issue: https://github.com/thorvg/thorvg/issues/3148

### DIFF
--- a/src/renderer/tvgPaint.cpp
+++ b/src/renderer/tvgPaint.cpp
@@ -119,10 +119,10 @@ static Result _compFastTrack(RenderMethod* renderer, Paint* cmpTarget, const Mat
         if (v1.x > v2.x) std::swap(v1.x, v2.x);
         if (v1.y > v2.y) std::swap(v1.y, v2.y);
 
-        after.x = static_cast<int32_t>(v1.x);
-        after.y = static_cast<int32_t>(v1.y);
-        after.w = static_cast<int32_t>(ceil(v2.x - after.x));
-        after.h = static_cast<int32_t>(ceil(v2.y - after.y));
+        after.x = static_cast<int32_t>(nearbyint(v1.x));
+        after.y = static_cast<int32_t>(nearbyint(v1.y));
+        after.w = static_cast<int32_t>(nearbyint(v2.x)) - after.x;
+        after.h = static_cast<int32_t>(nearbyint(v2.y)) - after.y;
 
         if (after.w < 0) after.w = 0;
         if (after.h < 0) after.h = 0;


mgrudzinska:
long time ago we had such an [issue](https://github.com/thorvg/thorvg/issues/916) with the render region size while rendering rectangles (_fastTrack_ case). Before the [fix](https://github.com/thorvg/thorvg/pull/939) (+some [explanation](https://github.com/thorvg/thorvg/pull/926#issuecomment-950004296)) this was the result we got from ThorVG:
image
Clearly the letter "I" is too big.

We solved this problem by applying a slightly different approach to calculating the bounding box in the _fastTrack_ case. The difference is no greater than 1 pixel, but that 1 pixel was the issue.
Today this part of the code is in the  **tvgSwMath.cpp** in the **mathUpdateOutlineBBox()** function:
```
    if (fastTrack) {
        renderRegion.min.x = static_cast<SwCoord>(nearbyint(xMin / 64.0f));
        renderRegion.max.x = static_cast<SwCoord>(nearbyint(xMax / 64.0f));
        renderRegion.min.y = static_cast<SwCoord>(nearbyint(yMin / 64.0f));
        renderRegion.max.y = static_cast<SwCoord>(nearbyint(yMax / 64.0f));
    } else {
        renderRegion.min.x = xMin >> 6;
        renderRegion.max.x = (xMax + 63) >> 6;
        renderRegion.min.y = yMin >> 6;
        renderRegion.max.y = (yMax + 63) >> 6;
    }
```

An important aspect of the above code, in the context of this PR, is that in the _fastTrack_ case, we don’t perform regular casting but use rounding instead (_nearbyint_). We need to apply the same approach to rectangular clips to ensure their size is calculated exactly the same way as other rectangles in the code.

mgrudzinska:
```
        auto global = tvg::Scene::gen();

        // scene with a clipped red rectangle

        shape1->appendRect(0, 0, 800, 800);
        shape1->fill(255, 0, 0);
        scene->push(shape1);

        clip->appendRect(50, 50, 400.6, 400.6);
        scene->clip(clip);

        global->push(scene);

        // black rect of the size of the above clip - so it should cover the whole red rectangle
        auto mat = tvg::Shape::gen();
        mat->appendRect(50, 50, 400.6, 400.6);
        mat->fill(0, 0, 0);
        global->push(mat);

        global->scale(0.99f);

        canvas->push(global);
```

before:
Zrzut ekranu 2025-02-1 o 13 42 11

after:
Zrzut ekranu 2025-02-1 o 13 40 44

hermet (approved):

PR #3193, opened 2025-02-07
trimming fill & stroke

mgrudzinska:
support added for all engines

@issue: https://github.com/thorvg/thorvg/issues/3118 

before:
Zrzut ekranu 2025-02-12 o 21 48 28

after:
Zrzut ekranu 2025-02-12 o 21 46 27

sample:
```
        float dashPattern1[2] = {20, 10};

        //Dashed stroke + fill
        shape1->appendCircle(120, 120, 75, 75);
        shape1->appendRect(200, 50, 150, 150);
        shape1->fill(0, 50, 155, 100);
        shape1->strokeFill(0, 0, 255);
        shape1->strokeCap(tvg::StrokeCap::Round);
        shape1->strokeDash(dashPattern1, 2);
        shape1->strokeWidth(12);
        shape1->trimpath(0.2f, 0.8f, false);

        canvas->push(shape1);

        //Dashed stroke only
        shape1 = tvg::Shape::gen();
        shape1->appendCircle(120, 120, 75, 75);
        shape1->appendRect(200, 50, 150, 150);
        shape1->strokeFill(0, 0, 255);
        shape1->strokeCap(tvg::StrokeCap::Round);
        shape1->strokeDash(dashPattern1, 2);
        shape1->strokeWidth(12);
        shape1->trimpath(0.2f, 0.8f, false);
        shape1->translate(400, 0);

        canvas->push(shape1);

        //stroke + fill
        shape1 = tvg::Shape::gen();
        shape1->appendCircle(120, 120, 75, 75);
        shape1->appendRect(200, 50, 150, 150);
        shape1->fill(0, 50, 155, 100);
        shape1->strokeFill(0, 0, 255);
        shape1->strokeCap(tvg::StrokeCap::Round);
        shape1->strokeWidth(12);
        shape1->trimpath(0.2f, 0.8f, false);
        shape1->translate(0, 300);

        canvas->push(shape1);

        //stroke only
        shape1 = tvg::Shape::gen();
        shape1->appendCircle(120, 120, 75, 75);
        shape1->appendRect(200, 50, 150, 150);
        shape1->strokeFill(0, 0, 255);
        shape1->strokeCap(tvg::StrokeCap::Round);
        shape1->strokeWidth(12);
        shape1->trimpath(0.2f, 0.8f, false);
        shape1->translate(400, 300);

        canvas->push(shape1);

        //fill only
        shape1 = tvg::Shape::gen();
        shape1->appendCircle(120, 120, 75, 75);
        shape1->appendRect(200, 50, 150, 150);
        shape1->fill(0, 50, 155, 100);
        shape1->trimpath(0.2f, 0.8f, false);
        shape1->translate(0, 600);

        canvas->push(shape1);
```

### DIFF
--- a/src/renderer/gl_engine/tvgGlTessellator.cpp
+++ b/src/renderer/gl_engine/tvgGlTessellator.cpp
@@ -1544,7 +1544,7 @@ void Stroker::stroke(const RenderShape *rshape)
     Point *pts, *trimmedPts = nullptr;
     uint32_t cmdCnt = 0, ptsCnt = 0;
 
-    if (rshape->strokeTrim()) {
+    if (rshape->trimpath()) {
         RenderPath trimmedPath;
         if (!rshape->stroke->trim.trim(rshape->path, trimmedPath)) return;
 
--- a/src/renderer/gl_engine/tvgGlTessellator.cpp
+++ b/src/renderer/gl_engine/tvgGlTessellator.cpp
@@ -850,16 +850,11 @@ Tessellator::~Tessellator()
 }
 
 
-bool Tessellator::tessellate(const RenderShape *rshape, bool antialias)
+bool Tessellator::tessellate(const RenderShape *rshape, const RenderPath& path, bool antialias)
 {
-    auto cmds = rshape->path.cmds.data;
-    auto cmdCnt = rshape->path.cmds.count;
-    auto pts = rshape->path.pts.data;
-    auto ptsCnt = rshape->path.pts.count;
-
     this->fillRule = rshape->rule;
 
-    this->visitShape(cmds, cmdCnt, pts, ptsCnt);
+    this->visitShape(path.cmds.data, path.cmds.count, path.pts.data, path.pts.count);
     this->buildMesh();
     this->mergeVertices();
 
@@ -888,12 +883,14 @@ void Tessellator::tessellate(const Array<const RenderShape *> &shapes)
     this->fillRule = FillRule::NonZero;
 
     for (uint32_t i = 0; i < shapes.count; i++) {
-        auto cmds = shapes[i]->path.cmds.data;
-        auto cmdCnt = shapes[i]->path.cmds.count;
-        auto pts = shapes[i]->path.pts.data;
-        auto ptsCnt = shapes[i]->path.pts.count;
+        const RenderPath* path = nullptr;
+        RenderPath trimmedPath;
+        if (shapes[i]->trimpath()) {
+            if (!shapes[i]->stroke->trim.trim(shapes[i]->path, trimmedPath)) continue;
+            path = &trimmedPath;
+        } else path = &shapes[i]->path;
 
-        this->visitShape(cmds, cmdCnt, pts, ptsCnt);
+        this->visitShape(path->cmds.data, path->cmds.count, path->pts.data, path->pts.count);
     }
 
     this->buildMesh();
@@ -1527,7 +1524,7 @@ Stroker::Stroker(GlGeometryBuffer* buffer, const Matrix& matrix) : mBuffer(buffe
 }
 
 
-void Stroker::stroke(const RenderShape *rshape)
+void Stroker::stroke(const RenderShape *rshape, const RenderPath& path)
 {
     mMiterLimit = rshape->strokeMiterlimit();
     mStrokeCap = rshape->strokeCap();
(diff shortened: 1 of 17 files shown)


hermet (commented):
common: support trimmed fill in common 
 -> renderer: support trimmed fill in common

src/renderer/tvgRender.h
```diff
@@ -160,6 +160,7 @@ struct RenderShape
     RenderColor color{};
     RenderStroke *stroke = nullptr;
     FillRule rule = FillRule::NonZero;
+    bool fillTrim = false;
```
hermet, line 163: Hmm.. I wonder how to figure out the fill for 3 cases: stroke only, fill only, both.
Otherwise, probably, we can put this variable in`TrimPath` like this:
```
enum Target : uint8_t { Stroke, Fill, Both };
```

src/renderer/tvgShape.h
```diff
@@ -209,7 +209,7 @@ struct Shape::Impl : Paint::Impl
         renderFlag |= RenderUpdateFlag::Stroke;
     }
 
-    void strokeTrim(float begin, float end, bool simultaneous)
+    void strokeTrim(float begin, float end, bool simultaneous, bool fillTrim = false)
```
hermet, line 212: Probably, you can add like this:
```
void trimpath(TrimPath& trimpath)
{
}
```

src/renderer/sw_engine/tvgSwShape.cpp
```diff
@@ -348,13 +348,13 @@ static bool _axisAlignedRect(const SwOutline* outline)
 }
 
 
-static SwOutline* _genOutline(SwShape* shape, const RenderShape* rshape, const Matrix& transform, SwMpool* mpool, unsigned tid, bool hasComposite, bool trimmed = false)
+static SwOutline* _genOutline(SwShape* shape, const RenderShape* rshape, const Matrix& transform, SwMpool* mpool, unsigned tid, bool hasComposite, bool strokeTrim = false)
```
hermet, line 351: `trimmed` looks already good. This is a general path outline. Please do not mix the stroke concept here.

hermet (changes requested):
Commit message is too weak. I believe that readers will have a hard time understanding 'trimmed fill' in this context. Please address the concept of the "trimmed fill" in the commit message. Thanks.

image

src/renderer/tvgShape.h
mgrudzinska: I can change the name, but I don't see what is the TrimPath that I should use as an arg. here begin/end and simultaneous are values passed by the api.

src/renderer/tvgRender.h
mgrudzinska: Do we want to support only fill trimming without trimming the stroke at the same time? via the API? The Lottie standard requires trimming both. so what are the use cases?

Please note that, for now, TrimPath is part of RenderStroke. This was motivated by the fact that in lotties fill-only trimming (for shapes without a stroke) doesn’t seem very common, whereas trimming both the stroke and fill-or just the stroke-  is more frequent (I can check numbers tmr). However, if we are considering support for fill-only trimming, I think TrimPath should be moved from RenderStroke to RenderShape. In that case, it should likely be a pointer rather than an object. But then if the stroke is trimmed, two objects will be created-RenderStroke in RenderShape and TrimPath in RenderShape.

src/renderer/sw_engine/tvgSwShape.cpp
mgrudzinska: but this is really only info about stroke... I'll try to change this after you comment whether we want to support fill only trimming (I asked this in the comment below)

mgrudzinska:
@hermet please be aware that the sw_engine is the only one that performs trimming for stroke and fill separately - the worst case scenario it is done twice

hermet:
@mgrudzinska please note that it's rebased.

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -2170,12 +2157,12 @@ BWTessellator::BWTessellator(Array<float>* points, Array<uint32_t>* indices): mR
 }
 
 
-void BWTessellator::tessellate(const RenderShape *rshape, const Matrix& matrix)
+void BWTessellator::tessellate(TVG_UNUSED const RenderShape *rshape, const RenderPath& path, const Matrix& matrix)
```
hermet, line 2160: let's remove `rshape` if it's not used.

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -889,10 +889,17 @@ void Tessellator::tessellate(const Array<const RenderShape *> &shapes)
...
+            if (!shapes[i]->stroke->trim.trim(shapes[i]->path, trimmedPath)) continue;
+            path = &trimmedPath;
+        } else path = &shapes[i]->path;
+
+        auto cmds = path->cmds.data;
+        auto cmdCnt = path->cmds.count;
+        auto pts = path->pts.data;
+        auto ptsCnt = path->pts.count;
 
         this->visitShape(cmds, cmdCnt, pts, ptsCnt);
```
hermet, line 904: Let's go simpler:
`
this->visitShape(path->cmds.data, path->cmds.count, path->pts.data, path->pts.count);
`

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -851,12 +851,12 @@ Tessellator::~Tessellator()
...
+bool Tessellator::tessellate(const RenderShape *rshape, const RenderPath& path, bool antialias)
 {
-    auto cmds = rshape->path.cmds.data;
-    auto cmdCnt = rshape->path.cmds.count;
-    auto pts = rshape->path.pts.data;
-    auto ptsCnt = rshape->path.pts.count;
+    auto cmds = path.cmds.data;
+    auto cmdCnt = path.cmds.count;
+    auto pts = path.pts.data;
+    auto ptsCnt = path.pts.count;
```
hermet, line 860: let's go simpler:
`
this->visitShape(path->cmds.data, path->cmds.count, path->pts.data, path->pts.count);
`

hermet (commented):
Some nitpicks, please check them. thanks.

hermet:
@mgrudzinska Please double-check with thorvg.test-suite; it seems to have crashed after this PR. Thanks.

mgrudzinska:
> @mgrudzinska Please double-check with thorvg.test-suite; it seems to have crashed after this PR. Thanks.

checked all samples, now is ok. Also I checked in AE and trim is not applied to the mask:
[mask_trimmed_pr.json](https://github.com/user-attachments/files/18908293/mask_trimmed_pr.json)
AE:
mask_trimmed_AE

thorvg before (after fill trimming is added):
mask_trimmed_before

final version of this pr:
trimed_mask_after

hermet (approved):

PR #3194, opened 2025-02-07
lottie: fix & enhance parsing

mgrudzinska:
The object type key does not need to be provided
first. Previously, content was skipped until the
"ty" type key was found. If the key was placed at
the end, the entire object was ignored.
Now, before parsing subsequent elements, finding
the "ty" key is enforced. Since in-situ parsing
does not support backtracking, the key is searched with an internal function.

Introducing support for the "ty" key placed anywhere highlighted the use of redundant logic for parsing the list of shapes — this duplicate logic has been removed.

before:
Zrzut ekranu 2024-05-30 o 01 42 45

after:
Zrzut ekranu 2024-05-30 o 01 27 31

previous version with discussion:
https://github.com/thorvg/thorvg/pull/2330


hermet:
Parsing back and forth in a haphazard way is highly inefficient, and it's not just an issue of covering 'ty' - it needs a general method.

hermet (approved):

Issue #3192, opened 2025-02-07
common: trim path ignored although it should affect the dash


mgrudzinska:
original posted by @hermet :

[18859.json](https://github.com/user-attachments/files/18704591/18859.json)

dash line should be moving:

![Image](https://github.com/user-attachments/assets/9ab9572c-5666-4aef-a692-d5b03fb0a5d3)

mgrudzinska:
In this example, the `begin` and `end` values of the trim path are 0 and 1, respectively, so it is ignored. However, the `offset` value of the trim path should affect the dashing.

mgrudzinska:
TVG after #3198 and #3199:
![Image](https://github.com/user-attachments/assets/0c5013a8-810a-48ff-bb80-346402d2c432)

AE:
![Image](https://github.com/user-attachments/assets/e59b66c7-e8e0-433d-9665-4087c037b618)

PR #3198, opened 2025-02-10
renderer: handle edge cases for trimming with dashing

mgrudzinska:
In cases where the begin and end values of trimming did not result in trimming (end - begin > 1), trimming was ignored. However, when the stroke is also dashed, this case affects the shift of the point where the first dash becomes visible.

@Issue: https://github.com/thorvg/thorvg/pull/3192

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -236,15 +236,16 @@ void LottieTrimpath::segment(float frameNo, float& start, float& end, Tween& twe
         end = 0.0f;
         return;
     }
-    if (tvg::equal(diff, 1.0f) || tvg::equal(diff, 2.0f)) {
+
+    //Even if the start and end values do not cause trimming, an offset > 0 can still affect dashing starting point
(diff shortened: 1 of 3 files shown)


mgrudzinska:
pyt_tvg1

SergeyLebedkin (approved):
Verified on wsl and windows
svg and lottie exaples

hermet:
@mgrudzinska could you please double check this part?

After the PR:

https://github.com/user-attachments/assets/9ed96c70-6874-4f2e-b970-2adca8ed5df4

lottie-web:

https://github.com/user-attachments/assets/83d25afb-3155-4b2d-b974-a69426d9e2a4

mgrudzinska:
> @mgrudzinska @SergeyLebedkin Could you please double check this part?
> 
> I applied both #3198 and #3199
> 
> It looks a different result you shared: [#3192 (comment)](https://github.com/thorvg/thorvg/issues/3192#issuecomment-2646757993)
> 
>  2025-02-11-12-01-17.mp4

Yes, because merging them automatically doesn’t apply the changes from 3199 to the new parts of the code in 3198.
I rebased 3199 to include both commits

mgrudzinska:
applied in #3199 3199

PR #3199, opened 2025-02-10
renderer: handle trimming cases passing through 0.0 point

mgrudzinska:
Trimming that passes through the curve's start point (like -0.1:0.1) requires passing through the curve twice. The points obtained from both passes must be joined - visible when dashing
Cases for simultaneous = true have been handled.

@Issue: https://github.com/thorvg/thorvg/issues/3192

sample:
```
        if (!canvas) return false;

        float dashPattern[2] = {80, 20};

        shape1->appendCircle(200, 200, 150, 150);
        shape1->fill(55, 0, 255, 100);
        shape1->strokeFill(0, 0, 255, 200);
        shape1->strokeJoin(tvg::StrokeJoin::Bevel);
        shape1->strokeCap(tvg::StrokeCap::Butt);
        shape1->strokeWidth(22);
        shape1->strokeTrim(-0.25, 0.25, true);
        shape1->strokeDash(dashPattern, 2, 40);
        canvas->push(shape1);
```

before:
Zrzut ekranu 2025-02-10 o 02 45 01

after:
Zrzut ekranu 2025-02-10 o 02 45 16

note:
now includes also https://github.com/thorvg/thorvg/pull/3198

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -236,15 +236,16 @@ void LottieTrimpath::segment(float frameNo, float& start, float& end, Tween& twe
         end = 0.0f;
         return;
     }
-    if (tvg::equal(diff, 1.0f) || tvg::equal(diff, 2.0f)) {
+
+    //Even if the start and end values do not cause trimming, an offset > 0 can still affect dashing starting point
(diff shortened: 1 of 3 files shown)


SergeyLebedkin (approved):
Verified on wsl and windows
svg and lottie exaples and various trim settings

PR #3201, opened 2025-02-10
webgpu_gaussian_optimization

SergeyLebedkin:
wg_engine: optimize gaussian blur effect for webgpu renderer

Issue: https://github.com/thorvg/thorvg/issues/3054

+10-15% FPS expected on hi resolutions

Single shader instance used for vertical and horizontal blur
Local workgroup memory used for caching texture data for local threads group
Threads configuration changed from 2d to 1d topology
before:
image
after:
gaussian_line


hermet (changes requested):
After this PR, a blur grid appears, which is not a desirable result we can not allow.

image

SergeyLebedkin:
> After this PR, a blur grid appears, which is not a desirable result we can allow.

before/after
image

SergeyLebedkin:
+10% increase
Before:
image

After:
optim

hermet (approved):

PR #3208, opened 2025-02-12
doc: update quick start demo

xland:


hermet:
@xland thanks for contribution, but the current doc in `main` is correct. the enum type has been updated. :)

Issue #3205, opened 2025-02-11
renderer: improve the dash-stroke


mgrudzinska:
The `strokeDash` API description explicitly states that values should be provided in pairs:
[@param[in] dashPattern The array of consecutive pair values of the dash length and the gap length.](https://github.com/thorvg/thorvg/blob/cc4c18d6c681432e3378f6664f3e1dcb59dc4a40/inc/thorvg.h#L1057)

Both the Lottie standard:
[Dash and gap entries MUST all be in a continuous order and alternate between dash and gap, starting with dash. If there are an odd number of dashes + gaps, the sequence will repeat with dashes and gaps reversed.](https://lottiefiles.github.io/lottie-docs/shapes/#stroke-dash)
and the Svg standard:
[If the list has an odd number of values, then it is repeated to yield an even number of values](https://www.w3.org/TR/SVG2/painting.html#StrokeDashing)
specify that if an odd number of values is provided, they should be duplicated. 
None of the loaders verify the count; they simply pass the data to the API.

The SW and GL implementations correctly handle an odd number of values internally. However, the WG implementation follows the API description and expects an even number of values.

Issues: 
- [x] https://github.com/thorvg/thorvg/issues/3217
- [x] https://github.com/thorvg/thorvg/issues/3191
- [x] https://github.com/thorvg/thorvg/issues/2729
- [x] https://github.com/thorvg/thorvg/issues/3222
- [x] https://github.com/thorvg/thorvg/issues/3223
- [ ] https://github.com/thorvg/thorvg/issues/2925
- [x] https://github.com/thorvg/thorvg/issues/3231

hermet:
Summarize this:
1. improve the doc to specify clearly for even/odd dash values are allowed so that our api is compatible with svg/lottie.
2. improve the engines to allow zero values.
3. bug fixes and make its behavior identical across all engines.

hermet:
Relative Issues: 
- [ ] https://github.com/thorvg/thorvg/issues/3217
- [ ] https://github.com/thorvg/thorvg/issues/3191

mgrudzinska:
@hermet tested on mac only, let me double check tmr at linux

PR #3221, opened 2025-02-13
wg_engine: handle properly odd numbers of dashes/gaps

mgrudzinska:
If the provided list has an odd number of values, then it should be repeated to yield an even number of values.

+ typo fixed

@Issue: https://github.com/thorvg/thorvg/issues/3205


mgrudzinska:
Zrzut ekranu 2025-02-13 o 15 51 04

SergeyLebedkin (approved):

PR #3232, opened 2025-02-14
docs: clarification of the strokeDash() description

mgrudzinska:
Clarification of the strokeDash API behavior according to the SVG and Lottie standards. Appropriate changes have been made at the API level, and they have been removed from the loaders.

@Issue: https://github.com/thorvg/thorvg/issues/3205

[dash_zeros.svg.zip](https://github.com/user-attachments/files/18792022/dash_zeros.svg.zip)
Zrzut ekranu 2025-02-14 o 01 47 05

[dash_zeros.json](https://github.com/user-attachments/files/18792067/dash_zeros.json)
Zrzut ekranu 2025-02-14 o 02 07 52


hermet:
@mgrudzinska Please improve the commit msg so that it properly reflects which part of Dash is being revised.

mgrudzinska:
> @mgrudzinska Please improve the commit msg so that it properly reflects which part of Dash is being revised. (I think this is not about doc change.)

@hermet please check now.

src/renderer/tvgShape.h
```diff
@@ -297,13 +297,16 @@ struct Shape::Impl : Paint::Impl
 
     Result strokeDash(const float* pattern, uint32_t cnt, float offset)
     {
-        if ((cnt == 1) || (!pattern && cnt > 0) || (pattern && cnt == 0)) {
+        if ((!pattern && cnt > 0) || (pattern && cnt == 0)) {
             return Result::InvalidArguments;
         }
 
+        auto len = 0.0f;
```
hermet, line 304: I think this is no point. it would be better to remove the exceptional handling for 304 ~ 309 here and just let it result in a broken visual.

mgrudzinska: So I need to add handling for such cases in each engine individually. For now, SW and GL will hang.

hermet: @mgrudzinska If it's simple to do, then let's go for it. Adding iteration for unusual exception handling is excessive, IHO.

mgrudzinska: @hermet Please check now. I'm not making any additional iterations, just using the existing ones.

src/renderer/tvgShape.h
```diff
@@ -319,9 +320,19 @@ struct Shape::Impl : Paint::Impl
                 rs.stroke->dashPattern = static_cast<float*>(malloc(sizeof(float) * cnt));
                 if (!rs.stroke->dashPattern) return Result::FailedAllocation;
             }
+            auto len = 0.0f;
             for (uint32_t i = 0; i < cnt; ++i) {
+                if (pattern[i] < 0.0f) {
```
mgrudzinska, line 325: this check has to be done somewhere, if not here, then in the svg/lottie loaders.

hermet: I think the engine is not obligated to handle those exceptions. Could you please address the reason?

src/renderer/tvgShape.h
```diff
@@ -319,9 +320,19 @@ struct Shape::Impl : Paint::Impl
...
+                    resetDash();
+                    return Result::InvalidArguments;
+                }
+                len += pattern[i];
                 rs.stroke->dashPattern[i] = pattern[i];
```
hermet, line 330: I think this is enough:
```
rs.stroke->dashPattern[i] = (pattern[i] < FLT_EPSILON) ? 0.0f : pattern[i];
```

mgrudzinska: This is up to us—how we want the API to behave. If we decide to handle such cases this way, it's fine, but an additional check needs to be added in both the SVG loader and the TVG loader.

In SVG, it's specified that if any value is < 0, dashing should be ignored. As far as I checked, Lottie Web follows the same behavior, but we need to confirm the specs with LAC.

src/renderer/tvgShape.h
mgrudzinska: please check my comment below

src/renderer/tvgShape.h
hermet: > This is up to us—how we want the API to behave.

let's do it.

hermet: @mgrudzinska please note If you want to make a strong case, please present the pros and cons. We can make a decision based on that. Thanks.

mgrudzinska: I’ve handled the SVG/Lottie standard on the loader side.

As for the API, I’ve made the changes as you suggested. If I understand correctly, you don’t want to perform a check to see if the sum of all dashes/gaps is greater than 0. So, we have two options: draw nothing, or draw the full stroke, ignoring the dash pattern (this is the standard for SVG, and while it's not yet specified for Lottie, this is how Lottie Web behaves).

1. If we decide to ignore the dash pattern – then this check must happen somewhere, either at the API level or individually in each engine. Otherwise, we won’t be able to distinguish the case where only some dash patterns are zero (the behavior differs depending on whether it’s a dash or a gap) compared to the case where all of them are zero.

2. If we decide that nothing is drawn, this check must still happen somewhere, because otherwise, we won’t move along the stroke while calculating its dashed outline. This non-movement is the expected behavior when selected dash pattern values are zero, but it causes an infinite loop if all values are zero. We could add a loop counter to break the loop, but that seems like a poor solution.

my conclusion:
I would keep the check to see if the sum of all elements in the dashPattern is greater than 0, and if not, I would draw the entire stroke.

mgrudzinska: changed to draft since I don't know whether we want to support option 1 or 2.

hermet: > changed to draft since I don't know whether we want to support option 1 or 2.

This is my suggestion:

1. Please add exception at the svg parsing level if that is necessary. It would be better efficent.
2. Please don't do anything for lottie, it's pointless so far. (Even you mentioned it has no spec and negative dash value is not expected in lottie)

hermet:
@mgrudzinska The last thing we need to make sure that  whether our rendering engine can draw a solid line when all dash values are set to 0.  Please this confirm.

mgrudzinska:
note: 
will be handled after LAC confirms the expected behavior:
https://github.com/lottie/lottie-spec/pull/138

[dash_zeros.json](https://github.com/user-attachments/files/18792067/dash_zeros.json)

hermet:
@mgrudzinska could you please rebase this?

hermet:
@mgrudzinska could you please double-check this?
before ->after:
image

mgrudzinska:
sample:
```
<svg viewBox="0 0 400 400" xmlns="http://www.w3.org/2000/svg">

    <path d="M 10 11 h 200 " fill="none" stroke="red" stroke-width="2" stroke-dasharray="4 4" stroke-linecap="butt"/>
    <path d="M 10 14 h 200 " fill="none" stroke="red" stroke-width="2" stroke-dasharray="5 5" stroke-linecap="butt"/>
    <path d="M 10 17 h 200 " fill="none" stroke="red" stroke-width="2" stroke-dasharray="4 0 5" stroke-linecap="butt"/>
    <path d="M 10 20 h 200 " fill="none" stroke="red" stroke-width="2" stroke-dasharray="4 0 5" stroke-linecap="round"/>

</svg>
```

before:
Zrzut ekranu 2025-03-11 o 12 42 37

after:
Zrzut ekranu 2025-03-11 o 12 48 26

all engines after:
Zrzut ekranu 2025-03-11 o 13 17 00

hermet:
@mgrudzinska Still small size values still occur issues. dead hang. 

```
        float pattern[2] = {FLT_EPSILON, FLT_EPSILON};
        shape->strokeDash(dashPattern, 2);
```

src/renderer/sw_engine/tvgSwShape.cpp
```diff
@@ -179,7 +179,7 @@ static void _dashCubicTo(SwDashStroke& dash, const Point* ctrl1, const Point* ct
     } else {
         while ((len - dash.curLen) > 0.0001f) {
             Bezier left, right;
-            if (dash.curLen > 0) {
+            if (dash.curLen > 0 || addZero) {
```
hermet, line 182: Eventually, this is bad for performance, so we need to think about it carefully.

mgrudzinska:
1) samples:
[dash_zero_test_butt.json](https://github.com/user-attachments/files/19566510/dash_zero_test_butt.json)
[dash_zero_test_round.json](https://github.com/user-attachments/files/19566512/dash_zero_test_round.json)
[dash_zero_test_square.json](https://github.com/user-attachments/files/19566513/dash_zero_test_square.json)

 dashes


2) static example:
Zrzut ekranu 2025-04-2 o 11 51 46
Zrzut ekranu 2025-04-2 o 11 51 42
Zrzut ekranu 2025-04-2 o 11 51 41


 ```
         tvg::StrokeCap caps[3] = {tvg::StrokeCap::Round, tvg::StrokeCap::Square, tvg::StrokeCap::Butt};

        float dy;

        //EPSILON
        // expected - lines
        dy = 0.0f;
        for (auto c: caps) {
            shape1->moveTo(50, 100 + dy);
            shape1->lineTo(350, 250 + dy);
            shape1->strokeFill(255, 0, 0);
            shape1->strokeWidth(25);
            shape1->strokeCap(c);

            auto shape2 = static_cast<tvg::Shape*>(shape1->duplicate());
            shape2->strokeFill(0, 0, 255);
            shape2->strokeWidth(2);

            float dashPattern1[2] = {50, FLT_EPSILON};
            shape1->strokeDash(dashPattern1, 2);
            canvas->push(shape1);
            canvas->push(shape2);

            dy += 50;
        }


        // expected - dots for Round/Square, nothing for Butt
        dy = 200.0f;
        for (auto c: caps) {
            shape1->moveTo(50, 100 + dy);
            shape1->lineTo(350, 250 + dy);
            shape1->strokeFill(255, 0, 0);
            shape1->strokeWidth(25);
            shape1->strokeCap(c);

            auto shape2 = static_cast<tvg::Shape*>(shape1->duplicate());
            shape2->strokeFill(0, 0, 255);
            shape2->strokeWidth(2);

            float dashPattern1[2] = {FLT_EPSILON, 50};
            shape1->strokeDash(dashPattern1, 2);
            canvas->push(shape1);
            canvas->push(shape2);

            dy += 50;
        }


        // expected - lines
        dy = 400.0f;
        for (auto c: caps) {
            shape1->moveTo(50, 100 + dy);
            shape1->lineTo(350, 250 + dy);
            shape1->strokeFill(255, 0, 0);
            shape1->strokeWidth(25);
            shape1->strokeCap(c);

            auto shape2 = static_cast<tvg::Shape*>(shape1->duplicate());
            shape2->strokeFill(0, 0, 255);
            shape2->strokeWidth(2);

            float dashPattern1[2] = {FLT_EPSILON, FLT_EPSILON};
            shape1->strokeDash(dashPattern1, 2);
            canvas->push(shape1);
            canvas->push(shape2);

            dy += 50;
        }

        //ZEROS
        // expected - lines
        dy = 0.0f;
        for (auto c: caps) {
            shape1->moveTo(400, 100 + dy);
            shape1->lineTo(750, 250 + dy);
            shape1->strokeFill(255, 0, 0);
            shape1->strokeWidth(25);
            shape1->strokeCap(c);

            auto shape2 = static_cast<tvg::Shape*>(shape1->duplicate());
            shape2->strokeFill(0, 0, 255);
            shape2->strokeWidth(2);

            float dashPattern1[2] = {50, 0};
            shape1->strokeDash(dashPattern1, 2);
            canvas->push(shape1);
            canvas->push(shape2);

            dy += 50;
        }

        // expected - dots for Round/Square, nothing for Butt
        dy = 200.0f;
        for (auto c: caps) {
            shape1->moveTo(400, 100 + dy);
            shape1->lineTo(750, 250 + dy);
            shape1->strokeFill(255, 0, 0);
            shape1->strokeWidth(25);
            shape1->strokeCap(c);

            auto shape2 = static_cast<tvg::Shape*>(shape1->duplicate());
            shape2->strokeFill(0, 0, 255);
            shape2->strokeWidth(2);

            float dashPattern1[2] = {0, 50};
            shape1->strokeDash(dashPattern1, 2);
            canvas->push(shape1);
            canvas->push(shape2);

            dy += 50;
        }

        // expected - lines
        dy = 400.0f;
        for (auto c: caps) {
            shape1->moveTo(400, 100 + dy);
            shape1->lineTo(750, 250 + dy);
            shape1->strokeFill(255, 0, 0);
            shape1->strokeWidth(25);
            shape1->strokeCap(c);

            auto shape2 = static_cast<tvg::Shape*>(shape1->duplicate());
            shape2->strokeFill(0, 0, 255);
            shape2->strokeWidth(2);

            float dashPattern1[20] = {0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1};
            shape1->strokeDash(dashPattern1, 20);
            canvas->push(shape1);
            canvas->push(shape2);

            dy += 50;
        }
 ```

src/renderer/sw_engine/tvgSwShape.cpp
mgrudzinska: I moved this part to 'else'.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -338,18 +338,19 @@ static void _parseDashArray(SvgLoaderData* loader, const char *str, SvgDash* das
         str = _skipComma(str);
         auto parsedValue = toFloat(str, &end);
         if (str == end) break;
-        if (parsedValue <= 0.0f) break;
+        if (parsedValue < 0.0f) {
+            dash->array.reset();
+            return;
```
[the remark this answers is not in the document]
mgrudzinska: api does not have to be aligned with the svg std

src/renderer/wg_engine/tvgWgGeometry.h
```diff
@@ -466,7 +468,10 @@ struct WgIndexedVertexBuffer
 
     void appendSquare(Point v0, Point v1, float dist, float halfWidth)
     {
-        if(tvg::zero(dist)) return;
+        if(tvg::zero(dist)) {
```
SergeyLebedkin, line 471: Are you shure that we need to apped geometry in a case of close (same) points?

mgrudzinska: @SergeyLebedkin yes, it's exactly the same like with round cap - even for 0 length lines, caps have to be added:
issue_3065

Zrzut ekranu 2025-06-19 o 01 37 51


also sample from here https://github.com/thorvg/thorvg/pull/3232#issuecomment-2772178164:

Zrzut ekranu 2025-06-19 o 01 33 24

src/renderer/sw_engine/tvgSwShape.cpp
```diff
@@ -508,7 +523,17 @@ bool shapeGenStrokeRle(SwShape* shape, const RenderShape* rshape, const Matrix&
     auto ret = true;
 
     //Dash style with/without trimming
-    if (rshape->stroke->dash.count > 0) {
+    if (rshape->stroke->dash.length > DASH_PATTERN_THRESHOLD) {
+        //check if the length of the drawn dashes is negligibly small => if yes, then it's a valid shape without a stroke
+        if (rshape->stroke->cap == StrokeCap::Butt) {
+            auto len = 0.0f;
+            for (uint32_t i = 0; i < rshape->stroke->dash.count; i += 2) len += rshape->stroke->dash.pattern[i];
+            if (len < DASH_PATTERN_THRESHOLD) {
```
hermet: I think you can just use `dash.length`.

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -611,7 +611,18 @@ void DashStroke::dashLineTo(const Point& to)
                     }
                     this->lineTo(left.pt2);
                 }
-            } else right = curr;
+            } else {
+                if (drawPoint && !mCurOpGap) {
+                    if (drawPoint && !mCurOpGap) {
```
SergeyLebedkin: Double check?

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -674,7 +685,16 @@ void DashStroke::dashCubicTo(const Point& cnt1, const Point& cnt2, const Point&
...
+            } else {
+                if (drawPoint && !mCurOpGap) {
+                    if (mMove || mDashPattern[mCurrIdx] < FLOAT_EPSILON) {
+                        this->moveTo(cur.start);
+                        mMove = false;
+                    }
+                    this->lineTo(cur.start);
+                }
+                right = cur;
+            }
```
SergeyLebedkin, line 694: Seems like this code is duplicates and we can generalize it...

mgrudzinska: @SergeyLebedkin  introduced drawPaoint() fun & changed a name of a boolen (drawPoint->validPoint)

src/renderer/sw_engine/tvgSwShape.cpp
mgrudzinska: this is added to handle cases like this:
```
        shape->moveTo(100,100);
        shape->lineTo(600,100);
        shape->lineTo(600,600);
        shape->strokeFill(255, 0, 0);
        shape->fill(0, 0, 255);
        shape->strokeWidth(15);
        shape->strokeCap(tvg::StrokeCap::Butt);

        float dashPattern1[2] = {0, 10}; //anything with dash values < 0.001 will case the same issue
        shape->strokeDash(dashPattern1, 2);
        canvas->push(shape);
```

dash is ```{0,10}``` so ```dash.length > DASH_PATTERN_THRESHOLD```, but since the dashing part is 0 and no caps needs to be drawn (```StrokeCap::Butt```), the ```mathUpdateOutlineBBox``` will return false since ```strokeOutline->pts.count``` is zero. As a result  the  ```shapeGenStrokeRle()``` fun returns false and since that nothing is drawn, even though this is a valid shape with a blue fill

hermet: If there are no other concerns, let's remove the entire exception logic here; cleaner code is preferable to introducing complexity for such ignoreable exceptions.

src/renderer/sw_engine/tvgSwShape.cpp
```diff
@@ -291,10 +303,11 @@ static SwOutline* _genDashOutline(const RenderShape* rshape, const Matrix& trans
         pts++;
     }
 
+    auto validPoint = rshape->stroke->cap != StrokeCap::Butt;
```
hermet, line 307: This is an appropriate place to address the useful comment, such as:
`// Corner area should be drawn regardless of the dash size.`

gl, wg as well.

mgrudzinska: done

src/renderer/sw_engine/tvgSwShape.cpp
mgrudzinska: done

PR #3245, opened 2025-02-18
wg_engine: Fix undefined error in task scheduler usage

tinyjin:
CleanShot 2025-02-18 at 17 33 11@2x


hermet:
@tinyjin Could you please build the GL engine as well? If it fails on your machine, we should add the header inclusion in the GL engine as well.

PR #3246, opened 2025-02-18
lottie/text: add default fallback font

tinyjin:
Embedded default font directly into the binary.

Fallback to the default font when the given text cannot be composed (e.g., missing glyphs or fonts).

---

[sample.json](https://github.com/user-attachments/files/18842804/sample.json)

https://github.com/user-attachments/assets/90f1052a-f95a-4d1b-bd89-fec305c5aab6


hermet (changes requested):
Let's consider this at the `tvgWasmAnimation` level. You can embed the font data and load the font during the initialization step.

hermet:
see this as well: https://github.com/thorvg/thorvg/pull/3252

hermet:
let's replace with this: https://github.com/thorvg/thorvg/pull/3252

PR #3247, opened 2025-02-18
sw_engine: fix internal types of rasterXYFlip

pharuxtan:
Before the C++ compiler couldn't decide the template of std::min (that happen if `int` type is not an `int32_t`), so explicitly changing all types to int32_t fixes this.


src/renderer/sw_engine/tvgSwRaster.cpp
```diff
@@ -1819,11 +1819,11 @@ void rasterXYFlip(uint32_t* src, uint32_t* dst, int32_t stride, int32_t w, int32
     }
 
     #pragma omp parallel for
-    for (int x = 0; x < w; x += BLOCK) {
+    for (int32_t x = 0; x < w; x += BLOCK) {
         auto bx = std::min(w, x + BLOCK) - x;
         auto in = &src[x];
         auto out = &dst[x * stride];
-        for (int y = 0; y < h; y += BLOCK) {
+        for (int32_t y = 0; y < h; y += BLOCK) {
```
mgrudzinska, line 1826: I'd change also in l.1830 and l.1831 int -> int32_t,  just for consistency

Issue #1897, opened 2024-01-09
Implement (partial) support for the SVG text element


capnm:
+ https://svgwg.org/svg2-draft/text.html

+ It seems to be possible to embed the required fonts 
  (but can lead to an impractical file size and latency):
  [Data URI scheme](https://en.wikipedia.org/wiki/Data_URI_scheme)  
  [Data URI Generator](https://dopiaza.org/tools/datauri/index.php)
```svg
<svg>
    <defs>
        <style>
            @font-face {
                font-family: Inter;
                src: url("data:application/font-woff;base64,AAEAAAATAQAABAAwR0RFRv4pBjw....")
            }
        </style>
    </defs>

    <!-- The rest of your SVG content goes here -->
</svg>
```

* The new api 1.x parameter isn't usable for system resources without direct filesystem access:
    `resourcePath`: absolute directory path where resource directory is included.
    https://github.com/thorvg/thorvg/pull/1802

* Add some threadsafe API to query the required external font files (parse <-> query -> render FSM)?

capnm:
```xml
<svg viewBox="0 0 300 150" xmlns="http://www.w3.org/2000/svg">
  <defs
     id="defs1">
    <style
       type="text/css"
       id="style1">
      @font-face {
        font-family: 'MyFont';
[...]

PR #3249, opened 2025-02-19
svg_loader: handle embedded fonts

mgrudzinska:
It is possible to embed fonts directly into an SVG file. Support for parsing and loading embedded fonts has been added.

@Issue: https://github.com/thorvg/thorvg/issues/1897

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -377,6 +377,27 @@ static char* _idFromUrl(const char* url)
 }
 
 
+static char* _srcFromUrl(const char* url)
+{
+    auto open = strchr(url, '(');
+    auto close = strchr(url, ')');
+    if (!open || !close || open >= close) return nullptr;
+
+    open = strchr(open, '\'');
+    if (!open || open >= close) return nullptr;
+    ++open;
+
+    close = strchr(open, '\'');
+    if (!close || close == open) return nullptr;
+    --close;
+
+    while (open < close && *open == ' ') ++open;
+    while (open < close && *close == ' ') --close;
+
+    return strDuplicate(open, (close - open + 1));
+}
+
+
 static unsigned char _parseColor(const char* value, char** end)
 {
     float r;
@@ -2084,6 +2105,35 @@ static SvgNode* _createImageNode(SvgLoaderData* loader, SvgNode* parent, const c
 }
 
 
+static bool _attrParseFontFace(void* data, const char* key, const char* value)
+{
+    if (!key || !value) return false;
+
+    key = _skipSpace(key, nullptr);
+    value = _skipSpace(value, nullptr);
+
+    auto loader = (SvgLoaderData*)data;
+    auto& embeddedFont = loader->fonts.last();
+
+    if (!strcmp(key, "font-family")) {
+        if (embeddedFont.name) tvg::free(embeddedFont.name);
+        embeddedFont.name = strdup(value);
+    } else if (!strcmp(key, "src")) {
+        if (embeddedFont.src) tvg::free(embeddedFont.src);
+        embeddedFont.src = _srcFromUrl(value);
+    }
+
+    return true;
+}
+
+
+static void _createEmbeddedFont(SvgLoaderData* loader, const char* buf, unsigned bufLength, parseAttributes func)
+{
+    loader->fonts.push(EmbeddedFont());
+    func(buf, bufLength, _attrParseFontFace, loader);
+}
+
+
 static SvgNode* _getDefsNode(SvgNode* node)
 {
     if (!node) return nullptr;
@@ -3517,6 +3567,8 @@ static void _svgLoaderParserXmlCssStyle(SvgLoaderData* loader, const char* conte
             TVGLOG("SVG", "Unsupported elements used in the internal CSS style sheets [Elements: %s]", tag);
         } else if (STR_AS(tag, "all")) {
             if ((node = _createCssStyleNode(loader, loader->cssStyle, attrs, attrsLength, simpleXmlParseW3CAttribute))) node->id = _copyId(name);
+        } else if (!strcmp(tag, "@font-face")) { //css at-rule specifying font
+            _createEmbeddedFont(loader, attrs, attrsLength, simpleXmlParseW3CAttribute);
         } else if (!isIgnoreUnsupportedLogElements(tag)) {
             TVGLOG("SVG", "Unsupported elements used in the internal CSS style sheets [Elements: %s]", tag);
         }
@@ -3869,6 +3921,12 @@ void SvgLoader::clear(bool all)
 
     ARRAY_FOREACH(p, loaderData.images) tvg::free(*p);
     loaderData.images.reset();
+
+    ARRAY_FOREACH(p, loaderData.fonts) {
+        Text::unload(p->name);
+        tvg::free(p->name); //p->src is freed right after loading the font
+    }
+    loaderData.fonts.reset();
 
     if (copy) tvg::free((char*)content);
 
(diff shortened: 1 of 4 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -961,6 +961,41 @@ static void _updateInvalidViewSize(const Scene* scene, Box& vBox, float& w, floa
...
+        if (!src) {
+            TVGLOG("SVG", "The embedded font named \"%s\" not loaded properly.", font->name);
+            tvg::free(font->src);
+            font->src = nullptr;
+            continue;
+        }
+        src += sizeof("ttf;base64,") - 1;
+
+        char *decoded = nullptr;
+        auto size = b64Decode(src, strlen(src), &decoded);
```
JSUYA, line 988: Should we consider utf-8 decoding?

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -568,6 +568,12 @@ struct SvgNodeIdPair
     char *id;
 };
 
+struct EmbeddedFont
+{
+    char* name;
+    char* src;
```
JSUYA, line 574: https://www.w3.org/TR/css-fonts-3/#at-font-face-rule
We only support ttf type, but font-face can have multiple src. It is not necessary now, but it seems that we should consider cases where multiple src are applied.

src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -568,6 +568,12 @@ struct SvgNodeIdPair
     char *id;
 };
 
+struct EmbeddedFont
```
JSUYA, line 571: font-face seems to allow src to contain a local path.
We support loading local ttf fonts. So it seems that font-face is not necessarily used only in the embedded font case.
Therefore, I think struct FontFace is more appropriate.

JSUYA (commented):
Overall it's very good. But I think it would be good if more notify was added for exception cases.

JSUYA:
The rendering of O character in https://github.com/thorvg/thorvg/issues/1897's sample is strange. this need to be reviewed as a different issue.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -961,6 +961,41 @@ static void _updateInvalidViewSize(const Scene* scene, Box& vBox, float& w, floa
...
+static void _loadFonts(Array<EmbeddedFont>& fonts)
+{
+    if (fonts.empty()) return;
+
+    TaskScheduler::async(false);
+    for (auto font = fonts.begin(); font < fonts.end(); ++font) {
+        if (!font->src || !font->name) return;
+
+        auto src = font->src;
+        if (strncmp(src, "data:", sizeof("data:") - 1)) continue;
```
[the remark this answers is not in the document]
mgrudzinska: please check now, I tried to unify your comment with Hermet's comment - I do all checked in one place.

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: But font data is binary - base64 allows embedding this data inside a text based file. 
utf8 encodes unicode text, so it won't work for binary font data. Do you agree?

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: I don't think it's about what types we support. It's true that src can contain multiple urls (including local sources, etc.):
```
@font-face {
            font-family: 'MyFont';
            src: url('data:font/ttf;base64,...data....') format('truetype'),       
                   url('data:font/ttf;base64,...data....') format('truetype');
            }
```
However, we currently don't validate whether a given font supports the specified text.
We also don't support `font-family` as a prioritized list of font family names - only a single name is supported.

I'd prefer to handle this after a real-case scenario arises. Is it ok with you?

src/loaders/svg/tvgSvgLoaderCommon.h
mgrudzinska: done

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -961,6 +961,41 @@ static void _updateInvalidViewSize(const Scene* scene, Box& vBox, float& w, floa
...
+        src += sizeof("data:") - 1;
+
+        //handling: 1) data:font/ttf;base64, 2) data:application/font-ttf;base64,
+        src = strstr(src, "ttf;base64,");
+        if (!src) {
```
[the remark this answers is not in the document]
mgrudzinska: I merged this check with the one above. please check now. I can remove the log, but I wanted to free `src` asap.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -963,6 +963,34 @@ static void _updateInvalidViewSize(const Scene* scene, Box& vBox, float& w, floa
...
+        auto src = p->src;
+        //handling: 1) data:font/ttf;base64, 2) data:application/font-ttf;base64,
+        if (strncmp(src, "data:", sizeof("data:") - 1) || !(src = strstr(src + sizeof("data:") - 1, "ttf;base64,"))) {
+            TVGLOG("SVG", "The embedded font \"%s\" data not loaded properly.", p->name);
+            tvg::free(p->src);
+            continue;
+        }
+        src += sizeof("ttf;base64,") - 1;
+
+        char *decoded = nullptr;
```
hermet, line 983: Seems it needs free() for decoded?

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -377,6 +377,27 @@ static char* _idFromUrl(const char* url)
...
+    ++open;
+
+    close = strchr(open, '\'');
+    if (!close || close == open) return nullptr;
+    --close;
+
+    while (open < close && *open == ' ') ++open;
+    while (open < close && *close == ' ') --close;
+
+    return strDuplicate(open, (close - open + 1));
```
hermet, line 397: Hmm... you might wanna consider to avoid data copy...

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2084,6 +2105,43 @@ static SvgNode* _createImageNode(SvgLoaderData* loader, SvgNode* parent, const c
...
+{
+    if (!key || !value) return false;
+
+    key = _skipSpace(key, nullptr);
+    value = _skipSpace(value, nullptr);
+
+    auto loader = (SvgLoaderData*)data;
+    auto& font = loader->fonts.last();
+
+    if (!strcmp(key, "font-family")) {
```
hermet, line 2126: please use `STR_AS`

src/loaders/svg/tvgSvgLoaderCommon.h
JSUYA: > I'd prefer to handle this after a real-case scenario arises. Is it ok with you?

Sure, let's handle when a real case to arise.

src/loaders/svg/tvgSvgSceneBuilder.cpp
mgrudzinska: please check now

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: We are in the middle of parsing at this stage. I might not know the font name, for example, so I am unable to load the font yet. so I don't quite see how - could you please elaborate? All char* data that requires it, such as IDs or URLs, is copied.

hermet: What you need is the pointer and length, rather than a data copy. Based on that, you can decode the base64.

image

PR #3252, opened 2025-02-20
Support the default font fallback

hermet:
- Allow the unnamed font and added lottie, svg font fallback mechanism
- Load the default font in wasm.
- Support the default font in wasm player.


hermet:
@tinyjin Please add a commit that load the default font in `tvgWasmAnimation` with the default font addition in this PR

tinyjin:
@hermet Have just added them in init/term process for web. Please take a look :)

CleanShot 2025-02-20 at 15 12 31@2x

hermet:
@tinyjin could you please add the font data in wasm side?

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -99,6 +110,9 @@ int init()
 
 void term()
 {
+    // Release loaded font
+    Text::unload(fontName);
```
hermet, line 114: not necessary.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -49,7 +51,16 @@ static const char* NoError = "None";
 // 0: success, 1: fail, 2: wait for async request
 int init()
 {
+    // Load default font
+    Text::load(fontName, reinterpret_cast<const char*>(default_font), default_font_size, "ttf", false);
+
 #ifdef THORVG_WG_RASTER_SUPPORT
+    // Check if WebGPU is supported in the browser
```
hermet, line 58: This doesn't seem related to font support. Please separate this change into an additional commit.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -49,7 +51,16 @@ static const char* NoError = "None";
 // 0: success, 1: fail, 2: wait for async request
 int init()
 {
+    // Load default font
+    Text::load(fontName, reinterpret_cast<const char*>(default_font), default_font_size, "ttf", false);
```
hermet, line 55: Text load should be in the next of the engine Init.
ie)
```
if (Initializer::init(0, tvg::CanvasEngine::Gl) != Result::Success) return nullptr;
Text::load();
```
 + sw, wgpu also

tinyjin:
@hermet done

PR #3244, opened 2025-02-18
common: introduced custom memory allocators

hermet:
Pleaes note that thorvg would allow the only custom memory allocators

malloc -> tvg::malloc
calloc -> tvg::calloc
realloc -> tvg::realloc
free -> tvg::free

issue: https://github.com/thorvg/thorvg/issues/2652

PR #3253, opened 2025-02-20
wasm: align with custom memory allocators

tinyjin:
see: https://github.com/thorvg/thorvg/pull/3244


hermet (changes requested):
Wasm doesn't need to do this since it's memory allocation mechanism should be aligned with the web system.

PR #3257, opened 2025-02-20
wasm: optimize default font size

tinyjin:
Enable default font compression to reduce WASM size.

The [LZSS](https://en.wikipedia.org/wiki/Lempel%E2%80%93Ziv%E2%80%93Storer%E2%80%93Szymanski) algorithm(lossless) is a simple compression method with advantages in decompression performance.

Font data is optimized without external compression dependencies, reducing the embedded font size to 9KB (-40%).

---

**[WASM Size Reduction]**

> 4456 bytes reduced

Before
before

After
after

**[Loading Time Diff]**

> No loading time bottleneck related to this change.

Before
image

After
CleanShot 2025-02-20 at 19 11 23@2x

**[Compression Rate]**  
LZSS may have a weaker compression rate than zip(Deflate), but with the current system font we use, the resulting data size is comparable to zipped data.

CleanShot 2025-02-20 at 19 19 47@2x


tinyjin:
[DMSans Regular.otf.zip](https://github.com/user-attachments/files/18885907/DMSans.Regular.otf.zip)

Sharing font data i've tested with, just in case

src/bindings/wasm/tvgWasmDefaultFont.h
```diff
@@ -97,10 +97,42 @@
...
 };
 
+inline const char* get_default_font() {
```
hermet, line 107: 1. please double check conventions:
` get_default_font()` -> ` getDefaultFont()` 

But there is the only one font, no other fonts existed:
`getDefaultFont()` -> `getFont()`

if  `set/get` is not quite mandatory (we can figure out with the parameters and return types)
`getFont()` -> `font()`

My suggestion:
`font()`

2. Please see other wrong convention naming in the next.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -118,7 +118,7 @@ struct TvgEngineMethod
     }
 
     void loadFont() {
-        Text::load("default", reinterpret_cast<const char*>(DEFAULT_FONT), DEFAULT_FONT_SIZE, "ttf", false);
+        Text::load("default", get_default_font(), DEFAULT_FONT_SIZE, "ttf", false);
```
hermet, line 121: Please confirm if decompression can be performed globally only once across multiple instances.

src/bindings/wasm/tvgWasmDefaultFont.h
```diff
@@ -97,10 +97,42 @@
 
 #include <cstddef>
 
+constexpr size_t COMPRESSED_FONT_SIZE = 9721;
 constexpr size_t DEFAULT_FONT_SIZE = 14852;
 
 constexpr const unsigned char DEFAULT_FONT[] = {
```
hermet, line 103: Rather say:
`
COMPRESSED_FONT[]
`

src/bindings/wasm/tvgWasmDefaultFont.h
```diff
@@ -97,10 +97,45 @@
...
 };
+unsigned char* _fontData = nullptr;
+
+inline const char* font() {
+    if (_fontData) return reinterpret_cast<const char*>(_fontData);
+
+    const unsigned char* input = COMPRESSED_FONT;
+    _fontData = new unsigned char[DEFAULT_FONT_SIZE];
```
hermet, line 112: memory leak? nowhere free it.

hermet (approved):

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -130,6 +130,7 @@ struct TvgSwEngine : TvgEngineMethod
     {
         std::free(buffer);
         Initializer::term(tvg::CanvasEngine::Sw);
+        retrieveFont();
```
hermet, line 133: @tinyjin renamed the functions to match `requestFont()` / `retrieveFont()`

PR #3267, opened 2025-02-24
renderer: fixing cached font selection

mgrudzinska:
Using a function (strstr) to check if a given font name appeared as a substring in the list of existing names caused errors when similar font names were used.
Fixed by switching to an exact comparison.

samples - problem observed when drawn using Lottie.cpp sample
[text_embeded_fontName1.json](https://github.com/user-attachments/files/18933545/text_embeded_myPath1.json)
[text_embeded_fontName2.json](https://github.com/user-attachments/files/18933546/text_embeded_myPath2.json)

expected:
Zrzut ekranu 2025-02-23 o 22 33 45

### DIFF
--- a/src/renderer/tvgLoader.cpp
+++ b/src/renderer/tvgLoader.cpp
@@ -232,6 +232,22 @@ static LoadModule* _findFromCache(const char* data, uint32_t size, const char* m
 }
 
 
+static bool _match(const char* path, const char* fontName)
+{
+    auto name = (const char*)strrchr(path, '/');
+#ifdef _WIN32
+    if (name) {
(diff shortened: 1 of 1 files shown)


mgrudzinska:
@hermet seems that this is intentional, right? 
isn't it a bit confusing?
I was testing embedded fonts and got some errors, since I used my own names of fonts

hermet:
> @hermet seems that this is intentional, right? isn't it a bit confusing? I was testing embedded fonts and got some errors, since I used my own names of fonts

@mgrudzinska You can improve this implementation, but please ensure that the test cases properly.

hermet:
https://github.com/thorvg/thorvg/pull/3313

Issue #3168, opened 2025-01-30
lottie: support expressions in slot overriding


theashraf:
I’m trying to replicate the effect shown in the attached video, where an expression is passed to a specific property (rotation, scale, and position) of a shape in a Lottie animation. Below is an example of the slots object I’ve used, which contains expressions for a rectangular shape’s rotation, scale, and position:

```json
{
  "rect_rotation": {
    "p": {
      "x": "var $bm_rt = time * 360;"
    }
  },
  "rect_scale": {
    "p": {
      "x": "var $bm_rt = [];\n$bm_rt[0] = value[0] + Math.cos(2 * Math.PI * time) * 100;\n$bm_rt[1] = value[1];"
    }
  },
  "rect_position": {
    "p": {
      "x": "var $bm_rt = [];\n$bm_rt[0] = value[0] + Math.cos(2 * Math.PI * time) * 100;\n$bm_rt[1] = value[1];"
    }
  }
}
```

https://github.com/user-attachments/assets/fd9777ec-8b7a-4ab3-b9cc-4a974fad4333

Lottie Asset
[test.json](https://github.com/user-attachments/files/18599255/test.json)

PR #3269, opened 2025-02-24
lottie/slot: Support expressions overriding

tinyjin:
Enabling Lottie Expressions Slot Overriding.

issue: #3168 

---

https://github.com/user-attachments/assets/4717d45f-80e5-418a-b67c-1d53cf22db64

---

# Demo

## 1. Default Slot

[Expressions Default Slot.json](https://github.com/user-attachments/files/18939439/Expressions.Default.Slot.json)

https://github.com/user-attachments/assets/6d96a6a1-f68a-4a72-87a6-8677ffc76e55

## 2. Override(add) Expressions

[no expressions file.json](https://github.com/user-attachments/files/18939725/no.expressions.file.json)

| Original  | Override | Reset |
| ------------- | -------------  | ------------- |
| CleanShot 2025-02-24 at 18 01 15@2x | CleanShot 2025-02-24 at 18 00 44 | CleanShot 2025-02-24 at 18 01 05@2x

## 3. Override existing Expressions

[original expressions to slot one.json](https://github.com/user-attachments/files/18939554/original.expressions.to.slot.one.json)

```cpp
const char* slotJson = R"({"rect_rotation":{"p":{"x":"var $bm_rt = time * 360;"}}})";
animation->override(slotJson);
```

| Original  | Override | Reset |
| ------------- | -------------  | ------------- |
| CleanShot 2025-02-24 at 17 48 52@2x  | CleanShot 2025-02-24 at 17 49 56  | CleanShot 2025-02-24 at 17 50 08@2x  |


src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -356,6 +356,20 @@ struct LottieGenericProperty : LottieProperty
 
     void copy(const LottieGenericProperty<Frame, Value, Scalar>& rhs, bool shallow = true)
     {
+        // Only override the expression if one exists
+        if (rhs.exp) {
+            if (shallow) {
+                exp = rhs.exp;
```
mgrudzinska, line 362: Is it a possible scenario that exp already has a value? For example, if I try to call override twice for the same sid?
If so, who is responsible for freeing the previous exp?

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -915,10 +915,15 @@ struct LottieSlot
...
+    struct Context {
+        LottieLayer* layer;
+        LottieObject* parent;
+    };
+
     void assign(LottieObject* target, bool byDefault);
     void reset();
 
-    LottieSlot(char* sid, LottieObject* obj, LottieProperty::Type type) : sid(sid), type(type)
+    LottieSlot(char* sid, LottieObject* obj, LottieProperty::Type type, Context context) : sid(sid), type(type), context(context)
```
mgrudzinska, line 926: I'd pass context as a const ref :
```
LottieSlot(char* sid, LottieObject* obj, LottieProperty::Type type, const Context& context)
```

src/loaders/lottie/tvgLottieProperty.h
tinyjin: @mgrudzinska 
The exp slot with existing property values is recently discussed. It should cover the case you mentioned, so the `return` statement in end of the condition is wrong. Will update for it soon, thanks.

tinyjin: The original value of exp is backed up in `pair->prop` for slot reset, and is freed at destructor of `LottieSlot`. However, if we try to override twice without reset, there is memory leak because overridden current value from first attemption would not be preserved, since it's not an original value.

This is not just an issue with the Expressions Slot but a general problem with Slots. I’ll address it in a separate PR. Thank you!

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -916,10 +916,15 @@ struct LottieSlot
...
+    struct Context {
+        LottieLayer* layer;
+        LottieObject* parent;
+    };
+
     void assign(LottieObject* target, bool byDefault);
     void reset();
 
-    LottieSlot(char* sid, LottieObject* obj, LottieProperty::Type type) : sid(sid), type(type)
+    LottieSlot(char* sid, LottieObject* obj, LottieProperty::Type type, const Context context) : sid(sid), type(type), context(context)
```
mgrudzinska, line 927: const Context context - `&` is missing

mgrudzinska (approved):
I left a small comment, otherwise ok

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -357,6 +357,20 @@ struct LottieGenericProperty : LottieProperty
...
+                const_cast<LottieGenericProperty<Frame, Value, Scalar>&>(rhs).exp = nullptr;
+            } else {
+                exp = new LottieExpression;
```
hermet, line 366: looks the `LottieExpression::code` is still shallow copied.

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -357,6 +357,20 @@ struct LottieGenericProperty : LottieProperty
 
     void copy(const LottieGenericProperty<Frame, Value, Scalar>& rhs, bool shallow = true)
     {
+        // Only override the expression if one exists
+        if (rhs.exp) {
```
hermet, line 361: This should be applied to all properties, not only for generic `LottieGenericProperty` but also `LottiePathSet`, etc...

Thus, rather, `LottieProperty::copy(...)` might be a better place handle this exp copy and all derivated classes just call the `LottieProperty::copy()`.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1502,41 +1502,37 @@ bool LottieParser::apply(LottieSlot* slot, bool byDefault)
 
     //OPTIMIZE: we can create the property directly, without object
     LottieObject* obj = nullptr;  //slot object
+    context.layer = slot->context.layer;
```
hermet, line 1505: one liner.
`context = {slot->context.layer, slot->context.parent};`

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -935,6 +940,9 @@ struct LottieSlot
     Array<Pair> pairs;
     LottieProperty::Type type;
     bool overridden = false;
+
+    //parsing context, used for expression
+    Context context;
```
hermet, line 945: Please put this in 941 ~ 942 for memory packing.
```
struct Context {
    LottieLayer* layer;
    LottieObject* parent;
} context;
```

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -916,10 +916,15 @@ struct LottieSlot
...
+    struct Context {
+        LottieLayer* layer;
+        LottieObject* parent;
+    };
+
     void assign(LottieObject* target, bool byDefault);
     void reset();
 
-    LottieSlot(char* sid, LottieObject* obj, LottieProperty::Type type) : sid(sid), type(type)
+    LottieSlot(char* sid, LottieObject* obj, LottieProperty::Type type, const Context& context) : sid(sid), type(type), context(context)
```
hermet, line 927: separate: `context -> layer, parent`

hermet (commented):
@tinyjin thanks, please check my comments

src/loaders/lottie/tvgLottieProperty.h
hermet: -1 for this.

hermet (changes requested):

PR #3278, opened 2025-02-25
wg_engine: Introduce fill effect for webgpu renderer

SergeyLebedkin:
Issue: https://github.com/thorvg/thorvg/issues/3054

Supported color param
See SceneEffects example

image

### DIFF
--- a/src/renderer/wg_engine/tvgWgShaderTypes.cpp
+++ b/src/renderer/wg_engine/tvgWgShaderTypes.cpp
@@ -197,52 +197,53 @@ void WgShaderTypeGradient::updateTexData(const Fill::ColorStop* stops, uint32_t 
 }
 
 //************************************************************************
-// WgShaderTypeGaussianBlur
+// WgShaderTypeEffectParams
 //************************************************************************
 
-void WgShaderTypeGaussianBlur::update(float sigma, const Matrix& transform)
-{
-    const float scale = std::sqrt(transform.e11 * transform.e11 + transform.e12 * transform.e12);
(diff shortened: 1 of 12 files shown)


src/renderer/wg_engine/tvgWgShaderTypes.cpp
```diff
@@ -197,52 +197,84 @@ void WgShaderTypeGradient::updateTexData(const Fill::ColorStop* stops, uint32_t
...
+    params[5] = tint->white[1] / 255.0f;
+    params[6] = tint->white[2] / 255.0f;
+    params[7] = 0.0f;
+    params[8] = tint->intensity / 255.0f;
+};
+
+
+void WgShaderTypeEffectParams::update(const RenderEffectTritone* tritone)
+{
+    params[0] = tritone->shadow[0] / 255.0f;
```
hermet, line 268: tint/tritone should not be in this commit.

hermet (changes requested):
1. Please ensure that refactoring changes and the addition of fill effects are handled in separate commits. This will make it easier to review and track changes.
2. If this commit is specifically for adding color effects, please avoid mixing partial implementations of tritone and tint logic. Each effect should be implemented fully and independently to maintain clarity and prevent incomplete features from being merged.

src/renderer/wg_engine/tvgWgShaderTypes.cpp
SergeyLebedkin: tint/tritone removed.
all other stuff is necessary for implementation

hermet (approved):

PR #3280, opened 2025-02-25
lottie: more precise culling for inverse mattes

hermet:
don't do rendering further when the inverse mattes has
a fully translucent. This could improve rendering performance
in that case.


hermet:
at a certain animation performance can be improved by this more than 10% (tested at 1900x1900)

PR #3281, opened 2025-02-26
common: Support .lot extension

tinyjin:
Add support for the .lot extension to recognize Lottie animation files, as this introduces an additional extension for Lottie.

issue: #3248

### DIFF
--- a/examples/Lottie.cpp
+++ b/examples/Lottie.cpp
@@ -43,7 +43,7 @@ struct UserExample : tvgexam::Example
 
         //ignore if not lottie.
         const char *ext = path + strlen(path) - 4;
-        if (strcmp(ext, "json")) return;
+        if (strcmp(ext, "json") && strcmp(ext, "lot")) return;
 
         //Animation Controller
         auto animation = tvg::Animation::gen();
--- a/examples/LottieExpressions.cpp
+++ b/examples/LottieExpressions.cpp
@@ -43,7 +43,7 @@ struct UserExample : tvgexam::Example
 
         //ignore if not lottie.
         const char *ext = path + strlen(path) - 4;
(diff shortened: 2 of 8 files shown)


hermet (commented):
could you please update the api doc as well?

```

    /**
     * @brief Loads a picture data from a memory block of a given size.
     *
     * ThorVG efficiently caches the loaded data using the specified @p data address as a key
     * when the @p copy has @c false. This means that loading the same data again will not result in duplicate operations
     * for the sharable @p data. Instead, ThorVG will reuse the previously loaded picture data.
     *
     * @param[in] data A pointer to a memory location where the content of the picture file is stored. A null-terminated string is expected for non-binary data if @p copy is @c false.
     * @param[in] size The size in bytes of the memory occupied by the @p data.
     * @param[in] mimeType Mimetype or extension of data such as "jpg", "jpeg", "lottie", "svg", "svg+xml", "png", etc. In case an empty string or an unknown type is provided, the loaders will be tried one by one.
     * @param[in] rpath A resource directory path, if the @p data needs to access any external resources.
     * @param[in] copy If @c true the data are copied into the engine local buffer, otherwise they are not.
     *
     * @retval Result::InvalidArguments In case no data are provided or the @p size is zero or less.
     * @retval Result::NonSupport When trying to load a file with an unknown extension.
     *
     * @warning It's the user responsibility to release the @p data memory.
     *
     * @note If you are unsure about the MIME type, you can provide an empty value like @c nullptr, and thorvg will attempt to figure it out.
     * @since 0.5
     */
    Result load(const char* data, uint32_t size, const char* mimeType, const char* rpath = nullptr, bool copy = false) noexcept;
```
+ capi as well.

hermet (approved):

Issue #3063, opened 2024-12-19
gl_engine: allow the stroke clipper (spec update)


hermet:
Clipping with a stroke is useful for various visual effects. We can support this approach as it offers better efficiency compared to alpha masking.

If a valid stroke is defined in the clipper shape, the renderer prioritizes clipping with the stroke over the shape's fill (See the right bottom sample)

```
            auto shape1 = tvg::Shape::gen();
            shape1->appendRect(500, 420, 250, 250, 20, 20);
            shape1->fill(255, 0, 255, 160);

            // color/alpha/opacity are ignored for a clip object - no need to set them
            auto clipShape = tvg::Shape::gen();
            clipShape->appendCircle(600, 550, 150, 150);
            clipShape->strokeWidth(20);

            //Clipping shape1 to clipShape
            shape1->clip(clipShape);

            canvas->push(shape1);
```

![image](https://github.com/user-attachments/assets/8f2bdfd7-ce48-4487-90f2-99b0ca8ea0b4)

Please leave any feedback for this behavior updates if you have any.

PR #3282, opened 2025-02-26
wg_engine: Introduce the stroke clipper

SergeyLebedkin:
Issue: https://github.com/thorvg/thorvg/issues/3063
See Clipping example

Before/After
image


hermet (approved):

hermet (changes requested):
Please double-check regressions with the Lottie example:

image

expected:
image

SergeyLebedkin:
> Please double-check regressions with the Lottie example:

We should ignore strokes with width==0 
Fixed

hermet (approved):

Issue #3266, opened 2025-02-21
lottie: text alignment issue


hermet:
texts are not center aligned

thorvg(top) / expected(bottom) [27578.json](https://github.com/user-attachments/files/18903139/27578.json)
![Image](https://github.com/user-attachments/assets/3b075f5e-f507-4851-bfa4-904573e48d2c)

thorvg(top) / expected(bottom) [7924.json](https://github.com/user-attachments/files/18903141/7924.json)
![Image](https://github.com/user-attachments/assets/a36fc523-a8fa-48a6-884f-610ec300a3de)

PR #3295, opened 2025-03-03
lottie/text: fix justification in embedded font rendering

tinyjin:
issue: #3266

Embedded font rendering was recently introduced but lacked support for text justification.  This adds the necessary justification logic to ensure proper alignment.

| Before | After |
|--------|-------|
| Before | After |
| Before | After |


hermet (changes requested):
Please double-check the indentation. Thanks.

hermet (approved):

Issue #3301, opened 2025-03-04
lottie: local font text line wrapping?


hermet:
Current:
![Image](https://github.com/user-attachments/assets/47a105c8-b44f-4332-ab87-8f97e5bb9696)

Expected:
![Image](https://github.com/user-attachments/assets/62389cfd-e6bb-4cd9-bbbb-b2ff256fe576)

[23240.json](https://github.com/user-attachments/files/19065480/23240.json)

PR #3310, opened 2025-03-05
lottie/text: render carriage return with local font

tinyjin:
Added missing carriage return support in local font case.

issue: #3301

CleanShot 2025-03-05 at 16 54 05@2x

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -912,26 +912,42 @@ void LottieBuilder::updateImage(LottieGroup* layer)
 }
 
 
+//TODO: unify with the updateText() building logic
 static void _fontText(LottieText* text, Scene* scene, float frameNo, LottieExpressions* exps)
 {
     auto& doc = text->doc(frameNo, exps);
     if (!doc.text) return;
 
+    auto delim = "\r\n";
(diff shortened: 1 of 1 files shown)


hermet:
Textbuilder might need a smart apporach to avoid the full duplicate the implementation between local font glyph & embedded glyph, Let's think about it.

image

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -917,21 +917,33 @@ static void _fontText(LottieText* text, Scene* scene, float frameNo, LottieExpre
...
     auto size = doc.size * 75.0f; //1 pt = 1/72; 1 in = 96 px; -> 72/96 = 0.75
-    auto txt = Text::gen();
-    if (txt->font(doc.name, size) != Result::Success) {
-        //fallback to any available font
-        txt->font(nullptr, size);
-    }
-    txt->text(doc.text);
-    txt->fill(doc.color.rgb[0], doc.color.rgb[1], doc.color.rgb[2]);
+    auto lineHeight = doc.size * 100.0f;
+    char* token = std::strtok(strDuplicate(doc.text), delim);
```
hermet, line 923: strDuplicate() leads memory leak.

hermet (changes requested):
Please check a comment, thanks.

Issue #3321, opened 2025-03-11
svg: stop elements in gradient changed when meshgradient present in a file


mgrudzinska:
originally posted https://github.com/godotengine/godot/issues/102005#issuecomment-2715550903

PR #3322, opened 2025-03-11
svg: fix incorrect assignment of <stop> elements

mgrudzinska:
If the svg contained an unsupported element with <stop> elements, they were incorrectly assigned to the last loaded gradient. Now fixed.

@Issue: https://github.com/thorvg/thorvg/issues/3321

before:
Zrzut ekranu 2025-03-11 o 23 30 29

after:
Zrzut ekranu 2025-03-11 o 23 30 14

[sample.svg.zip](https://github.com/user-attachments/files/19198907/421553147-22eaf24a-a9a7-4a35-abac-df5a54ac5b19.svg.zip)


JSUYA (approved):
LGTM

Issue #2591, opened 2024-07-24
lottie: Slot overriding enhancement


hermet:
These lists show the lack of slot overriding coverage, which needs improvement

- [x] File > Image
- [x] Transform → Position
- [x] Transform → Scale
- [x] Transform → Radius
- [x] Transform → Opacity
- [x] Expressions
- [ ] Modifier > Rounded Corner

theashraf:
@hermet do you have a plan for gradient override support ?

hermet:
@theashraf we already support Gradient filings.

hermet:
see: https://lottie.github.io/lottie-spec/latest/specs/helpers/#slottable-property

hermet:
https://github.com/thorvg/thorvg/issues/3168

hermet:
Considering slot interfaces for better optimal resusage:
```
int32_t slot = lottie->genSlot("xxxxx");    //Return unique slot id. return 0 when failed.
lottie->applySlot(slot);                             //Apply the created slot.  To reset to default, call -> lottie->override(0);
lottie->delSlot(slot);                                 //Delete the created slot. thorvg engine need to return to the default if the current deleting slot is applied.
```
@tinyjin please check this, and feel free to address better approaches for slot enhancement. Thanks.
@theashraf please confirm & leave any considerations for dotlottie player, if you have any. Thanks.

PR #3323, opened 2025-03-12
lottie/slot: Introduce reusable Slot API

tinyjin:
Introduce Slot APIs for efficient slot data reuse

New APIs:
- uint32_t LottieAnimation::gen(const char* slot)
- Result LottieAnimation::apply(uint32_t id)
- Result LottieAnimation::del(uint32_t id)

Removed API:
- Result override(const char* slot)

issue: #2591 

---

CleanShot 2025-03-17 at 15 44 14

# Usage

## Slot Generation

```cpp
const char* slotJson = R"({ /* Slot Data */  })";
auto slotId = animation->gen(slotJson); // retrieve slotId
```

## Apply specified Slot

```cpp
animation->apply(slotId);
```

## Reset applied Slots

```cpp
animation->apply(0); // Animation resets to default
```

## Slot Reusage

```cpp
auto slotId = animation->gen(slotJson);
animation->apply(slotId);
animation->apply(0);

// Apply slot data using slotId, without duplicated parsing
animation->apply(slotId);
```

## Delete Slot (Free)

```cpp
animation->del(slotId); // Also applied slot will be reset
```

### DIFF
--- a/src/loaders/lottie/thorvg_lottie.h
+++ b/src/loaders/lottie/thorvg_lottie.h
@@ -23,17 +23,6 @@ public:
     ~LottieAnimation() override;
 
     /**
-     * @brief Override Lottie properties using slot data.
-     *
-     * @param[in] slot The Lottie slot data in JSON format to override, or @c nullptr to reset.
-     *
-     * @retval Result::InsufficientCondition In case the animation is not loaded.
-     *
-     * @since 1.0
-     */
-    Result override(const char* slot) noexcept;
-
-    /**
     * @brief Specifies a segment by marker. 
     * 
     * Markers are used to control animation playback by specifying start and end points, 
@@ -109,6 +98,56 @@ public:
      * @note Experimental API
      */
     Result assign(const char* layer, uint32_t ix, const char* var, float val);
+
+    /**
+     * @brief Creates a new slot based on the given Lottie slot data.
+     *
+     * This function parses the provided JSON-formatted slot data and generates
+     * a new slot for animation control. The returned slot ID can be used to apply
+     * or delete the slot later.
+     *
+     * @param[in] slot A JSON string representing the Lottie slot data.
+     *
+     * @return A unique, non-zero slot ID on success. Returns @c 0 if the slot generation fails.
+     *
+     * @see apply(uint32_t id)
+     * @see del(uint32_t id)
+     *
+     * @since 1.0
+     */
+    uint32_t gen(const char* slot) noexcept;
+
+    /**
+     * @brief Applies a previously generated slot to the animation.
+     *
+     * This function applies the animation parameters defined by a slot.
+     * If the provided slot ID is 0, all previously applied slots will be reset.
+     *
+     * @param[in] id The ID of the slot to apply. Use 0 to reset all slots.
+     *
+     * @retval Result::InvalidArguments If the animation is not loaded or the slot ID is invalid.
+     *
+     * @see gen(const char* slot)
+     *
+     * @since 1.0
+     */
+    Result apply(uint32_t id) noexcept;
+
+    /**
+     * @brief Deletes a previously generated slot.
+     *
+     * This function removes a slot by its ID.
+     *
+     * @param[in] id The ID of the slot to delete. Retrieve the ID from gen().
+     *
+     * @retval Result::InvalidArguments If the animation is not loaded or the slot ID is invalid.
+     *
+     * @note This function should be paired with gen.
+     * @see gen(const char* slot)
+     *
+     * @since 1.0
+     */
+    Result del(uint32_t id) noexcept;
 
     /**
      * @brief Creates a new LottieAnimation object.
--- a/src/loaders/lottie/tvgLottieAnimation.cpp
+++ b/src/loaders/lottie/tvgLottieAnimation.cpp
@@ -30,15 +30,39 @@ LottieAnimation::LottieAnimation() = default;
 LottieAnimation::~LottieAnimation() = default;
 
 
-Result LottieAnimation::override(const char* slot) noexcept
+uint32_t LottieAnimation::gen(const char* slot) noexcept
+{
+    auto loader = PICTURE(pImpl->picture)->loader;
+    if (!loader) return 0;
+
+    return static_cast<LottieLoader*>(loader)->gen(slot);
+}
+
+
+Result LottieAnimation::apply(uint32_t id) noexcept
 {
     auto loader = PICTURE(pImpl->picture)->loader;
     if (!loader) return Result::InsufficientCondition;
 
-    if (static_cast<LottieLoader*>(loader)->override(slot)) {
+    if (static_cast<LottieLoader*>(loader)->apply(id)) {
         PAINT(pImpl->picture)->mark(RenderUpdateFlag::All);
         return Result::Success;
     }
+
+    return Result::InvalidArguments;
+}
+
+
+Result LottieAnimation::del(uint32_t id) noexcept
+{
+    auto loader = PICTURE(pImpl->picture)->loader;
+    if (!loader) return Result::InsufficientCondition;
+
+    if (static_cast<LottieLoader*>(loader)->del(id)) {
+        PAINT(pImpl->picture)->mark(RenderUpdateFlag::All);
+        return Result::Success;
+    }
+
     return Result::InvalidArguments;
 }
 
--- a/src/loaders/lottie/tvgLottieLoader.cpp
+++ b/src/loaders/lottie/tvgLottieLoader.cpp
@@ -25,10 +25,19 @@
 #include "tvgLottieModel.h"
 #include "tvgLottieParser.h"
 #include "tvgLottieBuilder.h"
+#include "tvgCompressor.h"
 
 /************************************************************************/
 /* Internal Class Implementation                                        */
 /************************************************************************/
+
+LottieCustomSlot::~LottieCustomSlot()
+{
+    ARRAY_FOREACH(p, props) {
+        delete(p->prop);
+    }
+}
+
 
 void LottieLoader::run(unsigned tid)
 {
@@ -44,7 +53,9 @@ void LottieLoader::run(unsigned tid)
             comp = parser.comp;
         }
         if (parser.slots) {
-            override(parser.slots, true);
+            auto slotcode = gen(parser.slots, true);
+            apply(slotcode, true);
+            del(slotcode, true);
             parser.slots = nullptr;
         }
         builder->build(comp);
@@ -288,42 +299,88 @@ Paint* LottieLoader::paint()
 }
 
 
-bool LottieLoader::override(const char* slots, bool byDefault)
+bool LottieLoader::apply(uint32_t slotcode, bool byDefault)
 {
     if (!ready() || comp->slots.count == 0) return false;
 
-    //override slots
-    if (slots) {
-        //Copy the input data because the JSON parser will encode the data immediately.
-        auto temp = byDefault ? slots : duplicate(slots);
+    auto applied = false;
 
-        //parsing slot json
-        LottieParser parser(temp, dirName, builder->expressions());
-        parser.comp = comp;
-
-        auto idx = 0;
-        auto succeed = false;
-        while (auto sid = parser.sid(idx == 0)) {
-            auto applied = false;
-            ARRAY_FOREACH(p, comp->slots) {
-                if (strcmp((*p)->sid, sid)) continue;
-                if (parser.apply(*p, byDefault)) succeed = applied = true;
-                break;
-            }
-            if (!applied) parser.skip();
-            ++idx;
+    // Reset all slots if slotcode is 0
+    if (slotcode == 0) {
+        ARRAY_FOREACH(p, comp->slots) {
+            (*p)->reset();
         }
-        tvg::free((char*)temp);
-        rebuild = succeed;
-        overridden |= succeed;
-        return rebuild;
-    //reset slots
-    } else if (overridden) {
-        ARRAY_FOREACH(p, comp->slots) (*p)->reset();
-        overridden = false;
-        rebuild = true;
+        applied = true;
+    } else {
+        INLIST_FOREACH(this->slots, slot) {
+            if (slot->code != slotcode) continue;
+            ARRAY_FOREACH(p, slot->props) {
+                p->target->apply(p->prop, byDefault);
+            }
+            applied = true;
+            break;
+        }
+    }
+    overridden = (slotcode != 0);
+    if (applied) rebuild = true;
+
+    return applied;
+}
+
+
+bool LottieLoader::del(uint32_t slotcode, bool byDefault)
+{
+    if (comp->slots.empty() || slotcode == 0 || !ready()) return false;
+
+    // Search matching value and remove
+    INLIST_SAFE_FOREACH(this->slots, slot) {
+        if (slot->code != slotcode) continue;
+        if (!byDefault) {
+            ARRAY_FOREACH(p, slot->props) {
+                p->target->reset();
+            }
+            rebuild = true;
+        }
+        this->slots.remove(slot);
+        delete(slot);
+        break;
     }
     return true;
+}
+
+
+uint32_t LottieLoader::gen(const char* slots, bool byDefault)
+{
+    if (!slots || !ready() || comp->slots.empty()) return 0;
+
+    //parsing slot json
+    auto temp = byDefault ? slots : duplicate(slots);
+    LottieParser parser(temp, dirName, builder->expressions());
+    parser.comp = comp;
+    
+    auto idx = 0;
+    auto custom = new LottieCustomSlot(djb2Encode(slots));
+
+    while (auto sid = parser.sid(idx == 0)) {
+        ARRAY_FOREACH(p, comp->slots) {
+            if (strcmp((*p)->sid, sid)) continue;  //find target
+            if (auto prop = parser.parse(*p)) {
+                custom->props.push({prop, *p});
+            }
+            break;
+        }
+        ++idx;
+    }
+
+    tvg::free((char*)temp);
+
+    if (custom->props.count > 0) {
+        this->slots.back(custom);
+        return custom->code;
+    }
+
+    delete(custom);
+    return 0;
 }
 
 
--- a/src/loaders/lottie/tvgLottieLoader.h
+++ b/src/loaders/lottie/tvgLottieLoader.h
@@ -24,11 +24,32 @@
 #define _TVG_LOTTIE_LOADER_H_
 
 #include "tvgCommon.h"
+#include "tvgInlist.h"
 #include "tvgFrameModule.h"
 #include "tvgTaskScheduler.h"
 
 struct LottieComposition;
 struct LottieBuilder;
+struct LottieProperty;
+struct LottieSlot;
+
+
+struct LottieCustomSlot
+{
+    INLIST_ITEM(LottieCustomSlot);
+
+    struct Override
+    {
+        LottieProperty* prop;
+        LottieSlot* target;
+    };
+    Array<Override> props;
+    uint32_t code;
+
+    LottieCustomSlot(uint32_t code) : code(code) {}
+    ~LottieCustomSlot();
+};
+
 
 class LottieLoader : public FrameModule, public Task
 {
@@ -41,9 +62,11 @@ public:
 
     LottieBuilder* builder;
     LottieComposition* comp = nullptr;
+    Inlist<LottieCustomSlot> slots;     //user custom slot list
 
     Key key;
     char* dirName = nullptr;            //base resource directory
+
     bool copy = false;                  //"content" is owned by this loader
     bool overridden = false;            //overridden properties with slots
     bool rebuild = false;               //require building the lottie scene
@@ -56,7 +79,11 @@ public:
     bool resize(Paint* paint, float w, float h) override;
     bool read() override;
     Paint* paint() override;
-    bool override(const char* slot, bool byDefault = false);
+
+    //Slot Methods
+    uint32_t gen(const char* slot, bool byDefault = false);
+    bool apply(uint32_t slotcode, bool byDefault = false);
+    bool del(uint32_t slotcode, bool byDefault = false);
 
     //Frame Controls
     bool frame(float no) override;
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -203,7 +203,7 @@ void LottieSlot::reset()
 }
 
 
-void LottieSlot::assign(LottieObject* target, bool byDefault)
+void LottieSlot::apply(LottieProperty* prop, bool byDefault)
 {
     auto copy = !overridden && !byDefault;
     auto shallow = pairs.count == 1 ? true : false;
@@ -214,22 +214,22 @@ void LottieSlot::assign(LottieObject* target, bool byDefault)
         switch (type) {
             case LottieProperty::Type::Float: {
                 if (copy) pair->prop = new LottieFloat(static_cast<LottieTransform*>(pair->obj)->rotation);
-                pair->obj->override(&static_cast<LottieTransform*>(target)->rotation, shallow, !copy);
+                pair->obj->override(prop, shallow, !copy);
                 break;
             }
             case LottieProperty::Type::Scalar: {
                 if (copy) pair->prop = new LottieScalar(static_cast<LottieTransform*>(pair->obj)->scale);
-                pair->obj->override(&static_cast<LottieTransform*>(target)->scale, shallow, !copy);
+                pair->obj->override(prop, shallow, !copy);
                 break;
             }
             case LottieProperty::Type::Vector: {
                 if (copy) pair->prop = new LottieVector(static_cast<LottieTransform*>(pair->obj)->position);
-                pair->obj->override(&static_cast<LottieTransform*>(target)->position, shallow, !copy);
+                pair->obj->override(prop, shallow, !copy);
                 break;
             }
             case LottieProperty::Type::Color: {
                 if (copy) pair->prop = new LottieColor(static_cast<LottieSolid*>(pair->obj)->color);
-                pair->obj->override(&static_cast<LottieSolid*>(target)->color, shallow, !copy);
+                pair->obj->override(prop, shallow, !copy);
                 break;
             }
             case LottieProperty::Type::Opacity: {
@@ -237,27 +237,28 @@ void LottieSlot::assign(LottieObject* target, bool byDefault)
                     if (pair->obj->type == LottieObject::Type::Transform) pair->prop = new LottieOpacity(static_cast<LottieTransform*>(pair->obj)->opacity);
                     else pair->prop = new LottieOpacity(static_cast<LottieSolid*>(pair->obj)->opacity);
                 }
-                pair->obj->override(&static_cast<LottieSolid*>(target)->opacity, shallow, !copy);
+                pair->obj->override(prop, shallow, !copy);
                 break;
             }
             case LottieProperty::Type::ColorStop: {
                 if (copy) pair->prop = new LottieColorStop(static_cast<LottieGradient*>(pair->obj)->colorStops);
-                pair->obj->override(&static_cast<LottieGradient*>(target)->colorStops, shallow, !copy);
+                pair->obj->override(prop, shallow, !copy);
                 break;
             }
             case LottieProperty::Type::TextDoc: {
                 if (copy) pair->prop = new LottieTextDoc(static_cast<LottieText*>(pair->obj)->doc);
-                pair->obj->override(&static_cast<LottieText*>(target)->doc, shallow, !copy);
+                pair->obj->override(prop, shallow, !copy);
                 break;
             }
             case LottieProperty::Type::Image: {
                 if (copy) pair->prop = new LottieBitmap(static_cast<LottieImage*>(pair->obj)->data);
-                pair->obj->override(&static_cast<LottieImage*>(target)->data, shallow, !copy);
+                pair->obj->override(prop, shallow, !copy);
                 break;
             }
             default: break;
         }
     }
+
     if (!byDefault) overridden = true;
 }
 
--- a/src/loaders/lottie/tvgLottieModel.h
+++ b/src/loaders/lottie/tvgLottieModel.h
@@ -26,6 +26,7 @@
 #include "tvgCommon.h"
 #include "tvgStr.h"
 #include "tvgCompressor.h"
+#include "tvgInlist.h"
 #include "tvgRender.h"
 #include "tvgLottieProperty.h"
 #include "tvgLottieRenderPooler.h"
@@ -1058,7 +1059,8 @@ struct LottieSlot
         LottieProperty* prop;
     };
 
-    void assign(LottieObject* target, bool byDefault);
+    void add(uint32_t slotcode, LottieProperty* prop);
+    void apply(LottieProperty* prop, bool byDefault = false);
     void reset();
 
     LottieSlot(LottieLayer* layer, LottieObject* parent, char* sid, LottieObject* obj, LottieProperty::Type type) : context{layer, parent}, sid(sid), type(type)
@@ -1079,7 +1081,7 @@ struct LottieSlot
     } context;
 
     char* sid;
-    Array<Pair> pairs;
+    Array<Pair> pairs;    // Object-property pairs that can be overridden by this slot
     LottieProperty::Type type;
 
     bool overridden = false;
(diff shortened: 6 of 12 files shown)


hermet:
@tinyjin could you please rebase this?

tinyjin:
@hermet Done!

mgrudzinska (commented):
The issue I see is when a slot id is identical for more than one slot - not due to hash collisions, but when the user provides two identical slots:
modified transformation slots:
```
        //slot (transform: scale)
        {
            slot7 = std::unique_ptr<tvg::LottieAnimation>(tvg::LottieAnimation::gen());
            auto picture = slot7->picture();
            if (!tvgexam::verify(picture->load(EXAMPLE_DIR"/lottie/extensions/slotsample7.json"))) return false;

            const char* slotJson = R"({"scale_id":{"p":{"a":1,"k":[{"i":{"x":0.833,"y":0.833},"o":{"x":0.167,"y":0.167},"s":[0,0],"t":0},{"s":[100,100],"t":100}]}}})";
            auto slotId = slot7->genSlot(slotJson);
            if (!tvgexam::verify(slot7->applySlot(slotId))) return false;

            const char* slotJson2 = R"({"scale_id":{"p":{"a":1,"k":[{"i":{"x":0.833,"y":0.833},"o":{"x":0.167,"y":0.167},"s":[0,0],"t":0},{"s":[100,100],"t":100}]}}})";
            auto slotId2 = slot7->genSlot(slotJson2);
            if (!tvgexam::verify(slot7->applySlot(slotId2))) return false;

            sizing(picture, 7);

            canvas->push(picture);
        }
```
result - the red dot should be scaled:
slot

mgrudzinska:
while closing the sample:

```

../src/loaders/lottie/tvgLottieModel.h:594:23: runtime error: member access within null pointer of type 'LottieProperty'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/loaders/lottie/tvgLottieModel.h:594:23 in 
AddressSanitizer:DEADLYSIGNAL
=================================================================
==98643==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000010 (pc 0x00010b76cc83 bp 0x7ff7b5cf81a0 sp 0x7ff7b5cf8160 T0)
==98643==The signal is caused by a READ memory access.
==98643==Hint: address points to the zero page.
    #0 0x10b76cc83 in LottieTransform::override(LottieProperty*, bool, bool) tvgLottieModel.h:594
    #1 0x10b70437f in LottieSlot::reset() tvgLottieModel.cpp:168
    #2 0x10b6f72f8 in LottieLoader::delSlot(unsigned int, bool) tvgLottieLoader.cpp:339
    #3 0x10b67b256 in tvg::LottieAnimation::delSlot(unsigned int) tvgLottieAnimation.cpp:59
    #4 0x10a21ecfa in UserExample::~UserExample() LottieExtension.cpp:340
    #5 0x10a21d4ec in UserExample::~UserExample() LottieExtension.cpp:335
    #6 0x10a20ebbd in tvgexam::Window::~Window() Example.h:168
    #7 0x10a20ee9c in tvgexam::SwWindow::~SwWindow() Example.h:285
    #8 0x10a20aec1 in tvgexam::main(tvgexam::Example*, int, char**, bool, unsigned int, unsigned int, unsigned int, bool) Example.h:568
    #9 0x10a20dec8 in main LottieExtension.cpp:355
    #10 0x7ff8075f5344 in start+0x774 (dyld:x86_64+0xfffffffffff5c344)

==98643==Register values:
rax = 0x0000000000000000  rbx = 0x0000000000000000  rcx = 0x0000100000000002  rdx = 0x0000000000000000  
rdi = 0x0000000000000000  rsi = 0x0000000000000000  rbp = 0x00007ff7b5cf81a0  rsp = 0x00007ff7b5cf8160  
 r8 = 0x0000000000000001   r9 = 0x0000000000000000  r10 = 0x00007ff807943dde  r11 = 0x0000000000000206  
r12 = 0x0000000000000001  r13 = 0x0000000000000010  r14 = 0x00006120000541c0  r15 = 0x9ddfea08eb382d69  
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV tvgLottieModel.h:594 in LottieTransform::override(LottieProperty*, bool, bool)
==98643==ABORTING
zsh: abort      ./LottieExtension
```

hermet:
No rush, this change must align with the the .lottie project.

tinyjin:
@hermet Have just rebased!

src/loaders/lottie/thorvg_lottie.h
```diff
@@ -110,6 +99,39 @@ class TVG_API LottieAnimation final : public Animation
      */
     Result assign(const char* layer, uint32_t ix, const char* var, float val);
 
+    /**
```
hermet, line 102:
```
/**
 * @brief Creates a new slot based on the given Lottie slot data.
 *
 * This function parses the provided JSON-formatted slot data and generates
 * a new slot for animation control. The returned slot ID can be used to apply
 * or delete the slot later.
 *
 * @param[in] slot A JSON string representing the Lottie slot data.
 *
 * @return A unique, non-zero slot ID on success. Returns @c 0 if the slot generation fails.
 *
 * @see applySlot(uint32_t slotId)
 * @see delSlot(uint32_t slotId)
 *
 * @since 1.0
 */
 ```

File: src/loaders/lottie/thorvg_lottie.h [hunk 2]
```diff
@@ -110,6 +99,39 @@ class TVG_API LottieAnimation final : public Animation
...
+     *
+     * @param[in] slot The Lottie slot data in JSON format.
+     *
+     * @retval The generated slot ID when successful, 0 otherwise.
+     *
+     * @since 1.0
+     */
+    uint32_t genSlot(const char* slot) noexcept;
+
+    /**
```
hermet, line 120:
```
/**
 * @brief Applies a previously generated slot to the animation.
 *
 * This function applies the animation parameters defined by a slot.
 * If the provided slot ID is 0, all previously applied slots will be reset.
 *
 * @param[in] slotId The ID of the slot to apply. Use 0 to reset all slots.
 *
 * @retval Result::InsufficientCondition If the animation is not loaded or the slot ID is invalid.
 *
 * @see genSlot(const char* slot)
 *
 * @since 1.0
 */
 ```

File: src/loaders/lottie/thorvg_lottie.h [hunk 3]
```diff
@@ -110,6 +99,39 @@ class TVG_API LottieAnimation final : public Animation
...
+     *
+     * @param[in] slotId The ID of the slot to apply, or 0 to reset all slots.
+     *
+     * @retval Result::InsufficientCondition In case the animation is not loaded or the slot ID is invalid.
+     *
+     * @since 1.0
+     */
+    Result applySlot(uint32_t slotId) noexcept;
+
+    /**
```
hermet, line 136:
```
/**
 * @brief Deletes a previously generated slot.
 *
 * This function removes a slot by its ID. If 0 is passed, all slots will be deleted.
 *
 * @param[in] slotId The ID of the slot to delete. Use 0 to delete all slots.
 *
 * @retval Result::InsufficientCondition If the animation is not loaded or the slot ID is invalid.
 *
 * @see genSlot(const char* slot)
 *
 * @since 1.0
 */
```

src/loaders/lottie/thorvg_lottie.h
```diff
@@ -110,6 +99,39 @@ class TVG_API LottieAnimation final : public Animation
...
+     * @brief Applies a previously generated slot to the animation.
+     *
+     * @param[in] slotId The ID of the slot to apply, or 0 to reset all slots.
+     *
+     * @retval Result::InsufficientCondition In case the animation is not loaded or the slot ID is invalid.
+     *
+     * @since 1.0
+     */
+    Result applySlot(uint32_t slotId) noexcept;
```
hermet: suggestion:
`slotId` -> `sid`

src/loaders/lottie/thorvg_lottie.h
```diff
@@ -110,6 +99,39 @@ class TVG_API LottieAnimation final : public Animation
...
+    /**
+     * @brief Deletes a previously generated slot.
+     *
+     * @param[in] slotId The ID of the slot to delete, or 0 to delete all slots.
+     *
+     * @retval Result::InsufficientCondition In case the animation is not loaded or the slot ID is invalid.
+     *
+     * @since 1.0
+     */
+    Result delSlot(uint32_t slotId) noexcept;
```
hermet: suggestion:
`slotId` -> `sid`

src/loaders/lottie/thorvg_lottie.h
```diff
@@ -110,6 +99,39 @@ class TVG_API LottieAnimation final : public Animation
...
+     * @since 1.0
+     */
+    Result applySlot(uint32_t slotId) noexcept;
+
+    /**
+     * @brief Deletes a previously generated slot.
+     *
+     * @param[in] slotId The ID of the slot to delete, or 0 to delete all slots.
+     *
+     * @retval Result::InsufficientCondition In case the animation is not loaded or the slot ID is invalid.
```
hermet: the api returns `InvalidArgument`.

src/loaders/lottie/thorvg_lottie.h
```diff
@@ -110,6 +99,39 @@ class TVG_API LottieAnimation final : public Animation
...
+     * @since 1.0
+     */
+    uint32_t genSlot(const char* slot) noexcept;
+
+    /**
+     * @brief Applies a previously generated slot to the animation.
+     *
+     * @param[in] slotId The ID of the slot to apply, or 0 to reset all slots.
+     *
+     * @retval Result::InsufficientCondition In case the animation is not loaded or the slot ID is invalid.
```
hermet: the api returns `InvalidArgument`.

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -44,7 +45,9 @@ void LottieLoader::run(unsigned tid)
             comp = parser.comp;
         }
         if (parser.slots) {
```
hermet, line 55: I think it's good to place this slot application part inside `parser.parse()` at the end of the function:
```
        if (parser.slots) {
            override(parser.slots, true);
            auto slotId = genSlot(parser.slots);
            applySlot(slotId, true);
            delSlot(slotId, true);
            parser.slots = nullptr;
        }
```

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -288,45 +291,133 @@ Paint* LottieLoader::paint()
...
+uint32_t LottieLoader::genSlot(const char* slots)
+{
+    if (!slots || !ready() || comp->slots.count == 0) return 0;
+
+    auto slotId = djb2Encode(slots);
+    INLIST_FOREACH(comp->slotDatas, p) {
+        if (p->id == slotId) return slotId;
+    }
+
+    auto temp = duplicate(slots);
```
hermet: Previously, that slots is skipped copying in the `byDefault` case because the text is hold by the parser. Actually, after the slot data is generated and the slot text can be flushed out immediately. Please double check this.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1547,7 +1547,7 @@ const char* LottieParser::sid(bool first)
 }
 
 
-bool LottieParser::apply(LottieSlot* slot, bool byDefault)
+LottieProperty* LottieParser::slotData(LottieSlot* slot)
```
hermet: behavior name is better.
`slotData()` -> `apply()`, `gen()`,  `parse()` ...

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -1143,6 +1169,7 @@ struct LottieComposition
     Array<LottieFont*> fonts;
     Array<LottieSlot*> slots;
     Array<LottieMarker*> markers;
+    Inlist<LottieSlotData> slotDatas;
```
hermet: Additional data set looks very inefficient. Couldn't it be associated with `Array<LottieSlot*> slots` ?

src/loaders/lottie/tvgLottieLoader.cpp
tinyjin: @hermet 
Hello, I agree with you that the logics are to parse default slots., thus would be good to arrange this into parser.
However this is also used as public API, which means APIs are defined at tvgLottieLoader. And parser doesn't have direct access to loader currently. I think that is also reason why the logic has been in loader side.

To let the parser handle those logics, we might need to add loader pointer to parser constructor so parser can handle its own default slot processing. I am not sure its good access, but that's the only way for now in my looking.

Perhaps you have better approach or idea?

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -288,45 +291,108 @@ Paint* LottieLoader::paint()
...
+    return true;
+}
+
+
+uint32_t LottieLoader::genSlot(const char* slots, bool byDefault)
+{
+    if (!slots || !ready() || comp->slots.count == 0) return 0;
+
+    auto slotId = djb2Encode(slots);
+    ARRAY_FOREACH(p, comp->slots) {
```
hermet: For general benefits, duplicate check looks not quite essential...

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -1058,7 +1059,23 @@ struct LottieSlot
...
+        Value(uint32_t id, LottieProperty* prop) : id(id), prop(prop) {}
+
+        ~Value()
+        {
+            if (prop) delete(prop);
+        }
+    };
+
+    LottieProperty* data(LottieObject* target);
+    void addValue(uint32_t id, LottieProperty* prop);
```
hermet: addValue() -> add()

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -1058,7 +1059,23 @@ struct LottieSlot
...
     };
 
-    void assign(LottieObject* target, bool byDefault);
+    struct Value {
+        INLIST_ITEM(Value);
+
+        uint32_t id;
+        LottieProperty* prop;
+
+        Value(uint32_t id, LottieProperty* prop) : id(id), prop(prop) {}
```
hermet: not necessary.

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -1058,7 +1059,23 @@ struct LottieSlot
...
+
+        ~Value()
+        {
+            if (prop) delete(prop);
```
hermet: if (prop) delete(prop) -> delete(prop);

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -1080,6 +1098,7 @@ struct LottieSlot
 
     char* sid;
     Array<Pair> pairs;
+    Inlist<Value> values;
```
hermet: 1100, 1101 looks ambiguous.  They need descriptive comments for their purposes.

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -203,7 +203,68 @@ void LottieSlot::reset()
 }
 
 
-void LottieSlot::assign(LottieObject* target, bool byDefault)
+LottieProperty* LottieSlot::data(LottieObject* target)
```
hermet: Suggestion, either:
1.get()  
2.property()

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -288,45 +291,108 @@ Paint* LottieLoader::paint()
...
+        INLIST_FOREACH((*p)->values, q) {
+            if (q->id != sid) continue;
+            (*p)->apply(q->prop, byDefault);
+            break;
+        }
+    }
+
+    overridden = true;
+    rebuild = true;
     return true;
```
hermet: this needs to return false when given sid is not valid.

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -288,45 +291,108 @@ Paint* LottieLoader::paint()
...
-        while (auto sid = parser.sid(idx == 0)) {
-            auto applied = false;
-            ARRAY_FOREACH(p, comp->slots) {
-                if (strcmp((*p)->sid, sid)) continue;
-                if (parser.apply(*p, byDefault)) succeed = applied = true;
-                break;
-            }
-            if (!applied) parser.skip();
-            ++idx;
+    if (sid == 0) {
```
hermet: good to add comment

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -288,45 +291,108 @@ Paint* LottieLoader::paint()
...
+        ARRAY_FOREACH(p, comp->slots) {
```
hermet, line 310: good to add comment

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -288,45 +291,108 @@ Paint* LottieLoader::paint()
...
+                rebuild = true;
+            }
+            (*p)->values.free();
+        }
+
+        overridden = false;
+        return true;
+    }
+
+    ARRAY_FOREACH(p, comp->slots) {
```
hermet: good to add comment

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -288,45 +291,108 @@ Paint* LottieLoader::paint()
...
+    rebuild = true;
     return true;
 }
 
 
+bool LottieLoader::delSlot(uint32_t sid, bool byDefault)
+{
+    if (!ready() || comp->slots.count == 0) return false;
+
+    if (sid == 0) {
```
hermet: 0 is not valid sid. Request to delete 0 sid looks weird...

`LottieAnimation::delSlot(uint32_t sid)` might be able to handle this.

src/loaders/lottie/tvgLottieAnimation.cpp
```diff
@@ -30,15 +30,39 @@ LottieAnimation::LottieAnimation() = default;
...
-Result LottieAnimation::override(const char* slot) noexcept
+uint32_t LottieAnimation::genSlot(const char* slot) noexcept
+{
+    auto loader = PICTURE(pImpl->picture)->loader;
+    if (!loader) return 0;
+
+    return static_cast<LottieLoader*>(loader)->genSlot(slot);
+}
+
+
```
hermet, line 41: applySlot -> apply

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -288,45 +291,108 @@ Paint* LottieLoader::paint()
...
+            break;
+        }
+        if ((*p)->values.empty()) (*p)->overridden = false;
+    }
+
+    return true;
+}
+
+
+uint32_t LottieLoader::genSlot(const char* slots, bool byDefault)
```
hermet: genSlot -> gen

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -288,45 +291,108 @@ Paint* LottieLoader::paint()
...
+        }
+    }
+
+    overridden = true;
+    rebuild = true;
     return true;
 }
 
 
+bool LottieLoader::delSlot(uint32_t sid, bool byDefault)
```
hermet: delSlot -> del

hermet (commented):
1. Please maintain naming consistency. `slotId` -> `sid` among all code and apis.
2. override() is not deprecated, it's removed. Please check the commit msg.

src/loaders/lottie/tvgLottieLoader.cpp
tinyjin: @hermet 0(zero) is called to delete all slots. It's the same to `apply(0)`.

tinyjin:
@hermet Here is a consideration regarding to the naming (slotId, sid).

- `sid` is actually from Lottie spec which is `char*`, the string key.
- `slotId` is here a different key, is a key generated by `djb2Encode` to return unique slot Id. `uint32_t` type.

I think those two IDs are different purpose of key. So good to have different key I guess. Also there is case two keys are used at same context, merging the names will conflict.

CleanShot 2025-07-22 at 11 44 45@2x

If the slotId given to `apply`, `del` and `gen` functions sounds weird, could you suggest better naming? or you still think to merge the names?

src/loaders/lottie/tvgLottieLoader.cpp
tinyjin: @hermet I was thinking the same, but thought early-exit for the condition `sid == 0` is much valid at the time. I will unify them. Thank you.

src/loaders/lottie/tvgLottieLoader.cpp
hermet: > 0(zero) is called to delete all slots.

The spec/behavior is essential for users? explict id to delete is more clear.

hermet:
> @hermet Here is a consideration regarding to the naming (slotId, sid).
> 
> * `sid` is actually from Lottie spec which is `char*`, the string key.
> * `slotId` is here a different key, is a key generated by `djb2Encode` to return unique slot Id. `uint32_t` type.
> 
> I think those two IDs are different purpose of key. So good to have different key I guess. Also there is case two keys are used at same context, merging the names will conflict.
> 
> CleanShot 2025-07-22 at 11 44 45@2x
> If the slotId given to `apply`, `del` and `gen` functions, could you suggest better naming? or you still think to merge the names?

@tinyjin Gotcha, `sid`, `slotId` is a bit confused using together. How about instead of slotId?

API -> `id` 
Internal -> `slotcode`

tinyjin:
> > @hermet Here is a consideration regarding to the naming (slotId, sid).
> > 
> > * `sid` is actually from Lottie spec which is `char*`, the string key.
> > * `slotId` is here a different key, is a key generated by `djb2Encode` to return unique slot Id. `uint32_t` type.
> > 
> > I think those two IDs are different purpose of key. So good to have different key I guess. Also there is case two keys are used at same context, merging the names will conflict.
> > CleanShot 2025-07-22 at 11 44 45@2x
> > If the slotId given to `apply`, `del` and `gen` functions, could you suggest better naming? or you still think to merge the names?
> 
> @tinyjin Gotcha, `sid`, `slotId` is a bit confused using together. How about instead of slotId?
> 
> API -> `id` Internal -> `slotcode`

That sounds good!

src/loaders/lottie/tvgLottieLoader.cpp
tinyjin: @hermet We originally have `override(nullptr)` to clear all. You mean we don't need that corresponding way?

src/loaders/lottie/tvgLottieLoader.cpp
hermet: Please remove this. This is not a general case. Users don't need to call gen() multiple times for the same code. You can leave a note about this at API doc.

hermet: @tinyjin I think we can remove this. It's costful and not a general case. Rather gen() - del() calls could be paired. You can leave the behavior note at api doc.

tinyjin: @hermet Probably, `apply(0)` is also removable part?

hermet: @tinyjin No. it needs to unset the exisiting slot.

tinyjin: @hermet Removed `del(0)` use-parameter

hermet:
```

=================================================================
==18410==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 48 byte(s) in 1 object(s) allocated from:
    #0 0x7f8f139cf548 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:95
    #1 0x7f8f12aada69 in LottieSlot::property(LottieObject*) ../src/loaders/lottie/tvgLottieModel.cpp:234
    #2 0x7f8f12b9fa53 in LottieParser::parseSlot(LottieSlot*) ../src/loaders/lottie/tvgLottieParser.cpp:1604
    #3 0x7f8f12a7d34e in LottieLoader::gen(char const*, bool) ../src/loaders/lottie/tvgLottieLoader.cpp:368
    #4 0x7f8f12a7ef84 in LottieLoader::run(unsigned int) ../src/loaders/lottie/tvgLottieLoader.cpp:48
    #5 0x7f8f125bdaf4 in tvg::Task::operator()(unsigned int) ../src/renderer/tvgTaskScheduler.h:71
    #6 0x7f8f125bdaf4 in tvg::TaskSchedulerImpl::run(unsigned int) ../src/renderer/tvgTaskScheduler.cpp:144
    #7 0x7f8f116ccdb3  (/lib/x86_64-linux-gnu/libstdc++.so.6+0xecdb3) (BuildId: ca77dae775ec87540acd7218fa990c40d1c94ab1)
    #8 0x7f8f1392fa41 in asan_thread_start ../../../../src/libsanitizer/asan/asan_interceptors.cpp:234
    #9 0x7f8f10e24aa3 in start_thread nptl/pthread_create.c:447

Indirect leak of 64 byte(s) in 1 object(s) allocated from:
    #0 0x7f8f139ce9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x7f8f12ab076e in realloc<LottieScalarFrame<RGB32>*> ../src/renderer/tvgCommon.h:109
    #2 0x7f8f12ab076e in tvg::Array<LottieScalarFrame<RGB32> >::reserve(unsigned int) ../src/common/tvgArray.h:78
    #3 0x7f8f12ab076e in tvg::Array<LottieScalarFrame<RGB32> >::operator=(tvg::Array<LottieScalarFrame<RGB32> > const&) ../src/common/tvgArray.h:100
    #4 0x7f8f12ab076e in LottieGenericProperty<LottieScalarFrame<RGB32>, RGB32, (LottieProperty::Type)6, true>::copy(LottieGenericProperty<LottieScalarFrame<RGB32>, RGB32, (LottieProperty::Type)6, true>&, bool) ../src/loaders/lottie/tvgLottieProperty.h:420
    #5 0x7f8f12ab076e in LottieSlot::property(LottieObject*) ../src/loaders/lottie/tvgLottieModel.cpp:235
    #6 0x7f8f12b9fa53 in LottieParser::parseSlot(LottieSlot*) ../src/loaders/lottie/tvgLottieParser.cpp:1604
    #7 0x7f8f12a7d34e in LottieLoader::gen(char const*, bool) ../src/loaders/lottie/tvgLottieLoader.cpp:368
    #8 0x7f8f12a7ef84 in LottieLoader::run(unsigned int) ../src/loaders/lottie/tvgLottieLoader.cpp:48
    #9 0x7f8f125bdaf4 in tvg::Task::operator()(unsigned int) ../src/renderer/tvgTaskScheduler.h:71
    #10 0x7f8f125bdaf4 in tvg::TaskSchedulerImpl::run(unsigned int) ../src/renderer/tvgTaskScheduler.cpp:144
    #11 0x7f8f116ccdb3  (/lib/x86_64-linux-gnu/libstdc++.so.6+0xecdb3) (BuildId: ca77dae775ec87540acd7218fa990c40d1c94ab1)
    #12 0x7f8f1392fa41 in asan_thread_start ../../../../src/libsanitizer/asan/asan_interceptors.cpp:234
    #13 0x7f8f10e24aa3 in start_thread nptl/pthread_create.c:447

Indirect leak of 16 byte(s) in 1 object(s) allocated from:
    #0 0x7f8f139cf548 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:95
    #1 0x7f8f12aade88 in LottieGenericProperty<LottieScalarFrame<RGB32>, RGB32, (LottieProperty::Type)6, true>::copy(LottieGenericProperty<LottieScalarFrame<RGB32>, RGB32, (LottieProperty::Type)6, true>&, bool) ../src/loaders/lottie/tvgLottieProperty.h:419
    #2 0x7f8f12aade88 in LottieSlot::property(LottieObject*) ../src/loaders/lottie/tvgLottieModel.cpp:235
    #3 0x7f8f12b9fa53 in LottieParser::parseSlot(LottieSlot*) ../src/loaders/lottie/tvgLottieParser.cpp:1604
    #4 0x7f8f12a7d34e in LottieLoader::gen(char const*, bool) ../src/loaders/lottie/tvgLottieLoader.cpp:368
    #5 0x7f8f12a7ef84 in LottieLoader::run(unsigned int) ../src/loaders/lottie/tvgLottieLoader.cpp:48
    #6 0x7f8f125bdaf4 in tvg::Task::operator()(unsigned int) ../src/renderer/tvgTaskScheduler.h:71
    #7 0x7f8f125bdaf4 in tvg::TaskSchedulerImpl::run(unsigned int) ../src/renderer/tvgTaskScheduler.cpp:144
    #8 0x7f8f116ccdb3  (/lib/x86_64-linux-gnu/libstdc++.so.6+0xecdb3) (BuildId: ca77dae775ec87540acd7218fa990c40d1c94ab1)
    #9 0x7f8f1392fa41 in asan_thread_start ../../../../src/libsanitizer/asan/asan_interceptors.cpp:234
    #10 0x7f8f10e24aa3 in start_thread nptl/pthread_create.c:447

```
+ @tinyjin

tinyjin:
@hermet Sorry for this! I've just fixed them up!

tinyjin:
@hermet While testing, I've found a bug where the slot reusage cannot override properly. The bug happens by shallow-copying the properties.

Previously every time we override, LottieParser remakes instant LottieProperty → we shallow-copy that property, and then override. (_It has been nomatter that just parsed slot property loses ownership of own datas, because that was instantly used_ → new override() <> new LottieProperty)

In this improvement, we want to use that property multiple times. We once generate slot as well as property, then even if we overrode once, the slot datas should not be cleared, that should be freed when the LottieSlot is deleted or del() is called.

--
As a result, when we apply the slot multiple times. From the second attempts, the property is overridden with empty value.

```cpp
const char* slotJson = R"({"gradient_fill":{"p":{"p":2,"k":{"k":[0,0.1,0.1,0.2,1,1,0.1,0.2,0,0,1,1]}}}})";
auto slotId = slot1->gen(slotJson);

// Case 1
slot1->apply(slotId);
slot1->apply(slotId);

// Case 2
slot1->apply(slotId);
slot1->apply(0);
slot1->apply(slotId);
```

| Case                          | Screenshot                                                                                  | Description                       |
|------------------------------|----------------------------------------------------------------------------------------------|-----------------------------------|
| **Original**                 | Original | 초기 상태                         |
| **Expected** | Expected | Fine with first overriding |
| **Result**     | Result | From the 2nd overriding |--

I've just modified copying and overriding logic. Particulary, colorStop and populate logic are quite changed.
Would appreciate your review, Thank you so much.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1598,14 +1598,13 @@ bool LottieParser::apply(LottieSlot* slot, bool byDefault)
 
     if (!obj || Invalid()) {
         delete(obj);
-        return false;
+        return nullptr;
     }
 
-    slot->assign(obj, byDefault);
-
+    auto prop = slot->property(obj);
```
hermet: It seems you just need a property after the parsing. Seems wrong introduction property() for a data copy.

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -203,7 +203,68 @@ void LottieSlot::reset()
 }
 
 
-void LottieSlot::assign(LottieObject* target, bool byDefault)
+LottieProperty* LottieSlot::property(LottieObject* target)
+{
+    LottieProperty* prop = nullptr;
+    //apply slot object to all targets
+    ARRAY_FOREACH(pair, pairs) {
```
hermet: This seems wrong. You iterates the logic here and overwrite the property without any memory handling. It seems you just copied the apply() here.

hermet:
@tinyjin I don’t understand why `property()` was introduced. It already looks like incorrect logic in terms of memory handling.

src/loaders/lottie/tvgLottieModel.cpp
tinyjin: you're right. I've mistaken. Removed unnecessary loop.

tinyjin:
@hermet Thanks. I agree with your point. I've corrected for those.

hermet (approved):

src/loaders/lottie/tvgLottieLoader.h
```diff
@@ -24,11 +24,32 @@
...
 #include "tvgFrameModule.h"
 #include "tvgTaskScheduler.h"
 
 struct LottieComposition;
 struct LottieBuilder;
+struct LottieProperty;
+struct LottieSlot;
+
+
+struct LottieCustomSlot
```
hermet, line 37: @tinyjin Please check, introduced LottieCustomSlot with your original design.

Issue #3327, opened 2025-03-16
svg2png: certain svg files with embedded png crash


beicause:
Reproducible in 0.14.2, 0.15.x, main branch
Not reproducible in 0.14.1

Use svg2png on these svg files will crash: [Specular_.zip](https://github.com/user-attachments/files/19272319/Specular_.zip)

![Image](https://github.com/user-attachments/assets/6cfc7f02-1795-45c2-86df-dbfc18116287)
![Image](https://github.com/user-attachments/assets/ca3a6ca2-1c61-45a1-bff2-cff7f53f9e70)

This was also reported in https://github.com/godotengine/godot/issues/103876

JSUYA:
I found that this issue occurs when specifying odd value for size.

hermet:
curlprit: https://github.com/thorvg/thorvg/pull/2552

hermet:
@JSUYA @mgrudzinska Please note that the SVG result does not work properly even after the crash is resolved

![Image](https://github.com/user-attachments/assets/3baee309-b43e-4478-847b-fe66c84fa97b)

hermet:
@beicause Hello, the fix will be in thorvg v0.15.12. Thanks.

hermet:
cc: @mgrudzinska

PR #3329, opened 2025-03-17
sw_engine: Prevent invalid access when rendering a fast tracked image

JSUYA:
When rendering an image with fast track, if the shifted image is shifted more than the spans area, an invalid access to the image buffer occured. So to prevent this,the area shifted further out of the buffer is not rendered.

related issue : https://github.com/thorvg/thorvg/issues/3327

### DIFF
--- a/src/renderer/sw_engine/tvgSwRaster.cpp
+++ b/src/renderer/sw_engine/tvgSwRaster.cpp
@@ -811,6 +811,7 @@ static bool _rasterDirectMattedRleImage(SwSurface* surface, const SwImage* image
     auto alpha = surface->alpha(surface->compositor->method);
 
     for (uint32_t i = 0; i < image->rle->size; ++i, ++span) {
+        if (span->y + image->oy < 0 || span->x + image->ox < 0) continue;
         auto dst = &surface->buf32[span->y * surface->stride + span->x];
         auto cmp = &cbuffer[(span->y * surface->compositor->image.stride + span->x) * csize];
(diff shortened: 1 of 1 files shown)


hermet:
This looks not good. I think the span region needs to be properly addressed before the rasterization step.

JSUYA:
https://github.com/thorvg/thorvg/blob/02a28bf9c32646c735e88b33af732cbbdad3aa2e/src/renderer/sw_engine/tvgSwImage.cpp#L81-L82

I have checked that the converted(nearbtint()) ox, oy are 1 px different from renderRegion.
If I remove nearbyint from ox, oy, the error does not occur, but I think the value is missing during the type conversion of renderRegion in the other place.
I think this part should be considered together with other examples, so I don't think it can be handled in this PR.
so i will closing this PR. I hope the above helps resolve this issue.

hermet:
> https://github.com/thorvg/thorvg/blob/02a28bf9c32646c735e88b33af732cbbdad3aa2e/src/renderer/sw_engine/tvgSwImage.cpp#L81-L82
> 
> I have checked that the converted(nearbtint()) ox, oy are 1 px different from renderRegion. If I remove nearbyint from ox, oy, the error does not occur, but I think the value is missing during the type conversion of renderRegion in the other place. I think this part should be considered together with other examples, so I don't think it can be handled in this PR. so i will closing this PR. I hope the above helps resolve this issue.

@JSUYA Thanks for your info. Let us thoroughly double-check the optimal fix.

PR #3340, opened 2025-03-20
lottie/text: Enable advanced text features with local font

tinyjin:
Previously, local fonts had their own rendering process, which did not support advanced features such as Range Selector, Alignment Options and Follow Path.

To enable these features and unify the logic, local fonts are now rendered glyph by glyph.

---

## Local Font existing Test set

| Test Case 1 | Test Case 2 | Test Case 3 |
|-------------|-------------|-------------|
| CleanShot 2025-03-20 at 20 29 49@2x | CleanShot 2025-03-20 at 20 29 30@2x | CleanShot 2025-03-20 at 20 29 16@2x |

---

## Advanced Feature with Local Font

| Text Range Selector | Text Follow Path | Text Alignment Options |
|---------------------|------------------|-------------------------|
| CleanShot 2025-03-20 at 20 21 21@2x | CleanShot 2025-03-20 at 20 13 07 | CleanShot 2025-03-20 at 20 26 50@2x |

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -26,7 +26,7 @@
 #include "tvgLottieModel.h"
 #include "tvgLottieBuilder.h"
 #include "tvgLottieExpressions.h"
-
+#include "tvgText.h"
 
 /************************************************************************/
 /* Internal Class Implementation                                        */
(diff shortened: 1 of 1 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1027,167 +984,240 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
...
+                txt->bounds(nullptr, nullptr, &twoCharWidth, nullptr);
+
+                // Get width of next char alone
+                char nextChar[2] = {*(p + 1), '\0'};
+                txt->text(nextChar);
+                float nextCharWidth;
+                txt->bounds(nullptr, nullptr, &nextCharWidth, nullptr);
+
+                // Calculate spacing between chars
+                letterSpacing = twoCharWidth - nextCharWidth; // Caching might help performance?
```
tinyjin, line 1057: Need a review for multiple calls of bounds, might affect to performance

hermet (changes requested):
Please properly refactor the logic. I can't accept buildText becoming such a monstrous function anymore — it's too much. Please review at the change in function LOC (255 -> 324).

image

Issue #3356, opened 2025-03-27
build wasm, error: use of undeclared identifier 'device'


ankye:
build wasm

./wasm_build.sh sw /mnt/c/bin/emsdk/

than build failed

`ninja: Entering directory `build_wasm/'
[63/208] Compiling C++ object src/bindings/wasm/thorvg-wasm.js.p/tvgWasmLottieAnimation.cpp.o
FAILED: src/bindings/wasm/thorvg-wasm.js.p/tvgWasmLottieAnimation.cpp.o
[...]

PR #3357, opened 2025-03-27
wasm: fix build error when WebGPU is disabled (SW/GL only)

tinyjin:
Fixed a build error that occurred when the WebGPU feature was disabled due to a missing feature check condition. The issue affected building with the software and OpenGL only backends.

issue: #3356

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -191,7 +191,9 @@ struct TvgWgEngine : TvgEngineMethod
 
     void resize(Canvas* canvas, int w, int h) override
     {
+    #ifdef THORVG_WG_RASTER_SUPPORT
         if (canvas) static_cast<WgCanvas*>(canvas)->target(device, instance, surface, w, h, ColorSpace::ABGR8888S);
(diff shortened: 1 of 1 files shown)


src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -191,7 +191,9 @@ struct TvgWgEngine : TvgEngineMethod
 
     void resize(Canvas* canvas, int w, int h) override
     {
+    #ifdef THORVG_WG_RASTER_SUPPORT
```
hermet, line 194: @tinyjin You might wanna add an #ifdef at line 265; then you can wrap the entire `TvgEngineMethod` with a single shot.

hermet (approved):

src/bindings/wasm/tvgWasmLottieAnimation.cpp
tinyjin: Agree, lets do this in another PR. GL part also needs to change

hermet:
@tinyjin A single refactoring can resolve the issue, making this fix unnecessary and helping us avoid an extra commit. Please think the optimal change.

PR #3363, opened 2025-03-28
examples: Add default font fallback to LottieExpressions demo

tinyjin:
Text Doc Expressions sample requires default font to show.

CleanShot 2025-03-28 at 16 14 44@2x


hermet:
@tinyjin could you please add the lottie sample as well?

PR #3370, opened 2025-03-31
Add Fedora package link

bkmgit:


hermet (approved):

hermet:
@bkmgit awesome, thanks for your contribution!

bkmgit:
Welcome.  Fedora has a build infrastructure https://packit.dev/ if it would be helpful to test commits on that. Can also use a container to run a Fedora build within GitHub actions.  Can make pull requests to add one of these if that would be helpful.

hermet:
Note: https://bodhi.fedoraproject.org/updates/?packages=thorvg

hermet:
> Can make pull requests to add one of these if that would be helpful.

@bkmgit Thanks for the info. We might need to consider this further, as the main branch is used for dev, doesn’t guarantee the backward compatibility.

bkmgit:
It takes about a day for new packages to be added to development branch (rawhide). Another week or so to release branches (f42 and f41)

bkmgit:
> > Can make pull requests to add one of these if that would be helpful.
> 
> @bkmgit Thanks for the info. We might need to consider this further, as the main branch is used for dev, doesn’t guarantee the backward compatibility.

Ok.  Prefer to package releases. Finding problems early is helpful though, so perhaps running an extended set of tests once a week or once a month or just before releases can be helpful.  Fedora tends to get newer versions of GCC earlier than other distributions.

PR #3373, opened 2025-04-01
lottie: fix stroke cap/join values

mgrudzinska:
For both enums, values 1 and 2 correspond to the same thing in ThorVG and in the Lottie file format. However, in ThorVG, StrokeCap::Square and StrokeJoin::Bevel correspond to 0, whereas in the Lottie file format, they correspond to value 3.

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -636,8 +636,8 @@ LottieSolidStroke* LottieParser::parseSolidStroke()
         else if (KEY_AS("c")) parseProperty(stroke->color, stroke);
         else if (KEY_AS("o")) parseProperty(stroke->opacity, stroke);
         else if (KEY_AS("w")) parseProperty(stroke->width, stroke);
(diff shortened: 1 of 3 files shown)


src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -636,8 +636,8 @@ LottieSolidStroke* LottieParser::parseSolidStroke()
         else if (KEY_AS("c")) parseProperty(stroke->color, stroke);
         else if (KEY_AS("o")) parseProperty(stroke->opacity, stroke);
         else if (KEY_AS("w")) parseProperty(stroke->width, stroke);
-        else if (KEY_AS("lc")) stroke->cap = (StrokeCap) getInt();
-        else if (KEY_AS("lj")) stroke->join = (StrokeJoin) getInt();
+        else if (KEY_AS("lc")) stroke->cap = (StrokeCap) (getInt() % 3);
```
hermet, line 639: Changing enum order, we can consider this as well and `getInt()++;`
```
Butt = 0, Round, Square
```
```
Miter = 0, Round, Bevel
```

hermet (approved):

PR #3379, opened 2025-04-03
lottie/text: added range selector flags

tinyjin:
| Text Style | Text Doc |
|------------|--------|
| CleanShot 2025-04-03 at 15 39 51@2x | CleanShot 2025-04-03 at 15 38 49@2x |

Duplicated text document properties (fill color, stroke color, and stroke width) to the text range.  
If these properties are not defined in the range selector, ThorVG incorrectly overrides them with default values.  
The correct behavior is to preserve the original values. The added flags address this issue.

---

### Stroke Width

| lottie-web | ThorVG (Before) | ThorVG (Patched) |
|------------|------------------|-------------------|
| CleanShot 2025-04-03 at 15 42 09@2x | Warp 2025-04-03 15 29 08 | CleanShot 2025-04-03 at 15 42 42@2x |

---

### Range Color

| lottie-web | ThorVG (Before) | ThorVG (Patched) |
|------------|------------------|-------------------|
| CleanShot 2025-04-03 at 15 45 06@2x | CleanShot 2025-04-03 at 15 45 47@2x | CleanShot 2025-04-03 at 15 45 10@2x |

---

**[Test Samples]**

- [stroke-width.json](https://github.com/user-attachments/files/19581352/stroke-width.json)  
- [36504.json](https://github.com/user-attachments/files/19581354/36504.json)

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -1082,25 +1082,13 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
                         opacity = (uint8_t)(opacity - f * (opacity - range->style.opacity(frameNo, tween, exps)));
                         shape->opacity(opacity);
 
-                        auto rangeColor = range->style.fillColor(frameNo, tween, exps); //TODO: use flag to check whether it was really set
-                        if (tvg::equal(f, 1.0f)) color = rangeColor;
-                        else {
-                            color.rgb[0] = tvg::lerp<uint8_t>(color.rgb[0], rangeColor.rgb[0], f);
-                            color.rgb[1] = tvg::lerp<uint8_t>(color.rgb[1], rangeColor.rgb[1], f);
-                            color.rgb[2] = tvg::lerp<uint8_t>(color.rgb[2], rangeColor.rgb[2], f);
-                        }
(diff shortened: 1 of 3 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -929,6 +929,20 @@ static void _fontText(TextDocument& doc, Scene* scene)
 }
 
 
+static void _rangeColor(LottieColor& color, RGB24& baseColor, float frameNo, Tween& tween, LottieExpressions* exps, float factor)
```
mgrudzinska, line 932: I think like this would be a little bit simpler?

```
static void _rangeColor(const RGB24& rangeColor, RGB24& color, float factor)
{
    if (tvg::equal(factor, 1.0f)) {
        color = rangeColor;
        return;
    }

    color.rgb[0] = tvg::lerp<uint8_t>(color.rgb[0], rangeColor.rgb[0], factor);
    color.rgb[1] = tvg::lerp<uint8_t>(color.rgb[1], rangeColor.rgb[1], factor);
    color.rgb[2] = tvg::lerp<uint8_t>(color.rgb[2], rangeColor.rgb[2], factor);
}
```

mgrudzinska (approved):
I made a small suggestion, but otherwise, I tested it and it works as expected.

hermet (approved):

PR #3389, opened 2025-04-08
common: fix incorrect aabb of scenes

BenjaminHalko:
- If you rotate a scene that contains a path, the AABB is incorrect.
  - The AABB of the scene is "wrapped" around the OBB for the child path, instead of being wrapped around the AABB for the child path.
  
Screenshot 2025-04-08 104437

Now after these changes:

image


hermet (approved):

hermet:
@BenjaminHalko thanks!

PR #3394, opened 2025-04-10
loaders: explicit variable scope to improve portability

hermet:
issue: https://github.com/godotengine/godot/pull/105093


hermet:
@mgrudzinska the fix must be included in v0.15.13

PR #3396, opened 2025-04-10
gl_engine: introduce dropshadow effect for opengl renderer

SergeyLebedkin:
issue: https://github.com/thorvg/thorvg/issues/3054

see EffectDropShadow example
image

### DIFF
--- a/src/renderer/gl_engine/tvgGlCommon.h
+++ b/src/renderer/gl_engine/tvgGlCommon.h
@@ -208,6 +208,11 @@ struct GlGaussianBlur {
     float extend{};
 };
 
+struct GlDropShadow: GlGaussianBlur {
+    float color[4];
+    float offset[2];
+};
+
 struct GlEffectParams {
     // fill:          [0..3]: color
     // tint:          [0..2]: black,  [4..6]: white,   [8]: intensity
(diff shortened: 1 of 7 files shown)


mgrudzinska (commented):
please note the difference in the color of the shadow (gl and wg have the same issue)
left - gl;    right - sw 
drop
screen:
Zrzut ekranu 2025-04-12 o 00 00 19


also I played with parameters (EffectDropShadow.cpp third sample):
```
        scene3->push(tvg::SceneEffect::DropShadow, 0, 0, 255, 125, (double)(360.0f * progress), 20.0, 4.0, 1);
```
one can see discrepancy (note the stroke) between gl and sw
left - gl;    right - sw 
Zrzut ekranu 2025-04-12 o 00 00 41

SergeyLebedkin:
the reason for the difference in the color of the tlt shadow shape is objectively not quite the same results of the blur mechanisms in different renderers.

at the moment, I do not know of any mechanics that would give identical results for different types of renderers

mgrudzinska:
so I think these discrepancies can be considered as issues - don't seem to be critical enough to hold off on the merge

mgrudzinska (approved):
so I think these discrepancies can be considered as issues - don't seem to be critical enough to hold off on the merge

hermet (approved):

Issue #3404, opened 2025-04-15
renderer: nullptr as clip argument


lpogic:
There is one more small issue related to clipping. Passing nullptr as an argument to Paint::clip method crashes the application.

hermet:
@lpogic the fix will be in v0.15.13. thanks for your contribution. :)

PR #3405, opened 2025-04-15
renderer: Added null check in Paint::Impl:clip

lpogic:
Prevents field accessing on `nullptr`.

Resolve: #3404


hermet (approved):

hermet:
@lpogic thanks!

Issue #2909, opened 2024-10-26
wg_engine: upgrade wgpu-native library


hermet:
Sync up to update the engine work.

The latest version: 22  -> https://github.com/gfx-rs/wgpu-native/releases
Current version: 19

See: https://github.com/thorvg/thorvg/wiki/WebGPU-Raster-Engine-Development

hermet:
@SergeyLebedkin Could you please figure out the new version is good for us? Thanks.
https://github.com/gfx-rs/wgpu-native/releases/tag/v24.0.0.2

PR #3406, opened 2025-04-15
webgpu: replace readonly strorages to texture2d (v24 migration)

SergeyLebedkin:
Since webgpu_native v24 _StorageTextureAccess_ReadOnly_ are not alowed:
```
Validation Error

Caused by:
  In wgpuDeviceCreateBindGroupLayout
    Binding 0 entry is invalid
      Read-write and read-only storage textures are not allowed by baseline webgpu
```
solution: https://github.com/gfx-rs/wgpu/issues/2735
bind it as a normal texture (not a storage texture) and use textureLoad on it

Code is still compatible with v19

issue: https://github.com/thorvg/thorvg/issues/2909


mgrudzinska (commented):
tested v24.0.3.1 - I can not build on macOS intel:
```
../src/renderer/wg_engine/tvgWgPipelines.h:116:15: error: unknown type name 'WGPUColorWriteMaskFlags'; did you mean 'WGPUColorWriteMask'?
```
in `webgpu/webgpu.h` the WGPUColorWriteMaskFlags typedef is no longer there.
after solving that I'm getting many other errors... can you please double check?

SergeyLebedkin:
> tested v24.0.3.1 - I can not build on macOS intel:

Sorry, please, test it on v19, but not on v24!
Current code is still based on v19 version, but not on v24
This is just minor changes to start migration process, but is not a migration at all
We are still on v19

mgrudzinska (approved):
tested on v0.19.4.1, no problems there

hermet (approved):

PR #3410, opened 2025-04-16
wg_engine/gl_engine: fix drop shadow blended color 

mgrudzinska:
|before|after|
|-|-|
|bbb|bbb|
|bbb|bbb|


SergeyLebedkin (approved):
great catch! thank you

Issue #3214, opened 2025-02-12
Can the Impl-pattern be dropped to (among others) ease debugging and to reduce code and (a small amount) of used memory?


J-u-n-o:
The impl-pattern hides the private members, but also add two pointers to the current implementation:
one pointing to the impl object and another pointing back.

It is a nice pattern hiding attributes, but adds a layer of complexity.
But should have been implemented by the C++ language.

Now this makes (among others) it harder to debug, both looking at variables and stepping through the code.

Also Macro's are added to hide the impl objects (partially):
like `PICTURE(pImpl->picture)->loader`  → `pImpl->picture->pImpl->loader`
non impl-pattern: `picture->loader`

Also when looking at PAINT_METHOD() for Paint::Impl,
it looks like this implements a virtual function calls depending on the type of the inheritance.
So better to use pure virtual functions and abstract classes, showing the actual intention.

PR #3415, opened 2025-04-18
renderer: revise the pImpl design with a better efficiency

hermet:
The following is a redesign that extends the class internally.

The main goal is to preserve the characteristics of the pImpl idiom for data encapsulation, while simultaneously reducing the memory allocation overhead typically associated with pImpl.

The stragegy is here:
Rather than alloc the impl memory inside of the thorvg engine, impl extends the API classes in order to consolidate the memory.

size has been decreased by -4kb with optimization=s

issue: https://github.com/thorvg/thorvg/issues/3214

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -801,8 +801,10 @@ public:
  *
  * Besides the APIs inherited from the Fill class, it enables setting and getting the linear gradient bounds.
  * The behavior outside the gradient bounds depends on the value specified in the spread API.
+ *
+ * @warning This class is not designed for inheritance.
  */
-class TVG_API LinearGradient final : public Fill
+class TVG_API LinearGradient : public Fill
 {
 public:
     /**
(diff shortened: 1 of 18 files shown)


hermet:
A reassessment of the risks of multiple inheritance is necessary.

hermet:
size has been decreased by -4kb with optimization=s

SergeyLebedkin (approved):

hermet:
size has been decreased by -4kb with optimization=s
perf: 15% improved (7.97s -> 6.93s)

```
        auto before = timestamp();

        for (int i = 0; i < 1000000; ++i) {

            auto p = tvg::Shape::gen();
            p->appendRect(0, 0, 200, 200);
            
            //LinearGradient
            fill->linear(0, 0, 400, 400);

            //Gradient Color Stops
            colorStops[1] = {1, 255, 255, 255, 255};
    
            p->fill(fill);
            canvas->push(p);
        }

        auto after = timestamp();
        printf("%f\n", after - before);
```

PR #3416, opened 2025-04-19
webgpu: introduce windows build scripts

SergeyLebedkin:
To simplify development and testing for windows platform introduce meson build stripts and cross file for windows platform

this is a part of migration to webgpu v24
should not impact to other platforms

issue: https://github.com/thorvg/thorvg/issues/2909

also, please see windows build instructions: 
https://github.com/thorvg/thorvg/wiki/WebGPU-Raster-Engine-Development#windows-mingw

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -8,10 +8,12 @@ if target_opengles
     examples_compiler_flags += '-DTHORVG_GL_TARGET_GLES=1'
 endif
 
-examples_dep = [dependency('sdl2')]
+# dependencies for sdl2 for windows are described in the corresponding cross file
+examples_dep = [dependency('sdl2', required : host_machine.system() not in ['emscripten', 'windows'])]
 
 if all_engines or get_option('engines').contains('wg')
-    examples_dep += dependency('wgpu_native')
+    # dependencies for webgpu for windows are described in the corresponding cross file
+    examples_dep += dependency('wgpu_native', required : host_machine.system() not in ['emscripten', 'windows'])
     if host_machine.system() == 'darwin'
         add_languages('objcpp')
         examples_dep += declare_dependency(link_args: ['-lobjc', '-framework', 'Cocoa', '-framework', 'CoreVideo', '-framework', 'IOKit', '-framework', 'QuartzCore'])
--- a/cross/win32_wg.txt
+++ b/cross/win32_wg.txt
@@ -1,0 +1,15 @@
+[binaries]
+c     = 'gcc'
+cpp   = 'g++'
+ar    = 'ar'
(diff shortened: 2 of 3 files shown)


examples/meson.build
```diff
@@ -8,10 +8,12 @@ if target_opengles
     examples_compiler_flags += '-DTHORVG_GL_TARGET_GLES=1'
 endif
 
-examples_dep = [dependency('sdl2')]
+# dependencies for sdl2 for windows are described in the corresponding cross file
+examples_dep = [dependency('sdl2', required : host_machine.system() not in ['emscripten', 'windows'])]
```
hermet, line 12: This will break the MSYS2 environment. It's easier to install essential packages such as SDL2 on Windows using the MSYS2 package manager rather than relying on a custom build.

SergeyLebedkin: OK. Let me check MSYS2 approach

SergeyLebedkin: @hermet 
Double check in MSYS2 environment (MINGW64) on this scripts.
Used packages:
```
pacman -Suy
pacman -S mingw-w64-x86_64-gcc
pacman -S mingw-w64-x86_64-SDL2
pacman -S mingw-w64-x86_64-meson
```
Result: meson and ninja works fine
image
Build instructions updated

SergeyLebedkin (commented):
MSYS2 works fine, meson and ninja as well
Build instuctions updated

examples/meson.build
SergeyLebedkin: @hermet 
MSYS2/MINGW64 env used
Installed packages:
```
pacman -Suy
pacman -Suy
pacman -S mingw-w64-x86_64-gcc
pacman -S mingw-w64-x86_64-gdb
pacman -S mingw-w64-x86_64-SDL2
pacman -S mingw-w64-x86_64-meson
```
image
Builded successfully

hermet: @SergeyLebedkin You're still using the custom build script, but it's not compatible with package managers. ThorVG can be built and installed on systems like vcpkg and MSYS2 without relying on it. I think we can reserve the custom script for WebGPU, since it's still in beta. However, in the case of SDL, we should ensure that it remains unaffected.

SergeyLebedkin:
@hermet 
Please, try to apply new build instuctions for [MSYS2/MINGW64](https://github.com/thorvg/thorvg/wiki/WebGPU-Raster-Engine-Development#windows-msys2mingw64) on main branch

If every thing is ok, lets close this PR without merging

hermet:
@SergeyLebedkin Fair enough, thanks!

Issue #3392, opened 2025-04-09
SVG clip-path not applied to embedded bitmap images (symbol via use)


lostminds:
It appears that when an svg clip-path is applied to a use tag displaying embedded bitmap data (or maybe any symbol content?) the clipping is not correctly applied.

As far as I can test embedding bitmap image data and displaying this using a `<use>` tag renders correctly in ThorVG
For example 

```
(svg wrapper and image data symbol def omitted)
<use id="internet.jpg" transform="translate(265.0,300.0)" x="-265.0" y="-300.0" width="530.0" height="600.0" href="#internet.jpg"/>
<ellipse stroke="rgb(0,0,0)" stroke-width="5.7" transform="translate(265.0,300.0)" ry="200.0" rx="200"/>
```
Produces something like this:
![Image](https://github.com/user-attachments/assets/657d5d7e-c533-47ca-a41e-9a491f2b9c29)

However, if you extend this to define a clip path (same as the circle)

```
(svg wrapper and image data symbol def omitted)
<clipPath id="clip_1">
<ellipse transform="translate(265.0,300.0)" ry="200.0" rx="200"/>
</clipPath>
<use id="internet.jpg" clip-path="url(#clip_1)" transform="translate(265.0,300.0)" x="-265.0" y="-300.0" width="530.0" height="600.0" href="#internet.jpg"/>
<ellipse stroke="rgb(0,0,0)" stroke-width="5.7" transform="translate(265.0,300.0)" ry="200.0" rx="200"/>
```
This should result in something like this:
![Image](https://github.com/user-attachments/assets/a47d4415-6a16-48cb-96a6-e58b8e4e1abf)
SVG file:
![Image](https://github.com/user-attachments/assets/c9606880-3f46-461a-a628-10a6c9325c50)

However, in the ThorVG viewer we get this instead:
![Image](https://github.com/user-attachments/assets/b65bcfc0-a38d-4fac-80cb-875784b650e3)
And in older ThorVG versions it seems so clipping is applied at all.

mgrudzinska:
@lostminds I see what the issue is, thx :) will solve this asap

hermet:
@lostminds hello, could you please help to reproduce your issue? (see the attach) i used your code snippet but couldn't reproduce it.  Thanks.
clip-test.zip](https://github.com/user-attachments/files/19860083/clip-test.zip)

hermet:
@lostminds the fix will be in v0.15.13, thanks.

PR #3420, opened 2025-04-22
svg_loader: fix clipping in use/symbol nodes

mgrudzinska:
Clipping for the <use> node when a <symbol> was used, was being overridden when 'overflowVisible' was 'true' (the default value), which caused the 'clip-path' applied to the <use> node to have no visible effect.
The clipping for the <use> node (without <symbol> node) was transformed in an incorrect order, resulting in a visually incorrect results.

@Issue: https://github.com/thorvg/thorvg/issues/3392

before:
Zrzut ekranu 2025-04-22 o 13 51 16

after (as expected):
Zrzut ekranu 2025-04-22 o 13 50 59

samples:
sym
use

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -807,12 +807,22 @@ static Scene* _useBuildHelper(SvgLoaderData& loaderData, const SvgNode* node, co
             }
             viewBoxClip->transform(mClipTransform);
 
-            scene->clip(viewBoxClip);
+            auto clippingLayer = Scene::gen();
+            clippingLayer->clip(viewBoxClip);
+            clippingLayer->push(scene);
+            return clippingLayer;
         }
-    } else {
-        scene->transform(mUseTransform);
+        return scene;
     }
 
+    if (auto clipper = PAINT(scene)->clipper) {
+        auto& clipTransform = clipper->transform();
+        Matrix inv;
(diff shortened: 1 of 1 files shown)


hermet:
@mgrudzinska For verifying your fix, It would be helpful if you could provide a sample. Thanks.

JSUYA:
Test SVG
https://github.com/JSUYA/jsuya.github.io/blob/master/test_embedded_image_with_clipper.svg

Test before
<image src="https://github.com/JSUYA/jsuya.github.io/blob/master/clip_fix_before.png?raw=true"/>

Test after
<image src="https://github.com/JSUYA/jsuya.github.io/blob/master/clip_fix_after.png?raw=true"/>

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -807,12 +807,22 @@ static Scene* _useBuildHelper(SvgLoaderData& loaderData, const SvgNode* node, co
...
         }
-    } else {
-        scene->transform(mUseTransform);
+        return scene;
+    }
+
+    if (auto clipper = PAINT(scene)->clipper) {
+        auto& clipTransform = clipper->transform();
+        Matrix inv;
+        if (node->transform && inverse(node->transform, &inv)) clipTransform = inv * clipTransform;
```
JSUYA, line 821: I think it's better to move `inv` inside the `if` since it won't be used if there's no `node->transform`.

JSUYA (approved):

hermet (approved):

Issue #2453, opened 2024-06-23
gl_engine: Improve portability by using dynamic GL APIs bindings


hermet:
The GL engine is maturing in terms of feature coverage. Currently it decides gl-link on build time, it needs to support dynamic symbol bindings to improve portability across systems.

RuiwenTang:
Can we use an [OpenGL Loader Library](https://www.khronos.org/opengl/wiki/OpenGL_Loading_Library) like #2574 ?

hermet:
@RuiwenTang It would likely be a simple wrapping process, so how about absorbing Glad into the ThorVG style? I think it might not be necessary to mention "glad" and its prefix within ThorVG.

PR #3423, opened 2025-04-23
gl_engine: dynamic load gl functions

SergeyLebedkin:
no any 3rd-party libs, such as glfw or glad did not used. code looks more clean
support windows, linux, macos and emsdk

issue: https://github.com/thorvg/thorvg/issues/2453

### DIFF
--- a/examples/Example.h
+++ b/examples/Example.h
@@ -327,6 +327,7 @@ struct GlWindow : Window
         if (!initialized) return;
 
 #ifdef THORVG_GL_TARGET_GLES
+        SDL_SetHint(SDL_HINT_OPENGL_ES_DRIVER, "1"); // important for windows
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
@@ -353,7 +354,6 @@ struct GlWindow : Window
         //Free in the reverse order of their creation.
         delete(canvas);
         canvas = nullptr;
-
         SDL_GL_DeleteContext(context);
     }
 
--- a/src/renderer/gl_engine/meson.build
+++ b/src/renderer/gl_engine/meson.build
@@ -1,4 +1,5 @@
 source_file = [
+   'tvgGl.h',
    'tvgGlCommon.h',
    'tvgGlGpuBuffer.h',
    'tvgGlList.h',
@@ -9,6 +10,7 @@ source_file = [
    'tvgGlRenderTask.h',
    'tvgGlShader.h',
    'tvgGlShaderSrc.h',
+   'tvgGl.cpp',
    'tvgGlGeometry.cpp',
    'tvgGlGpuBuffer.cpp',
    'tvgGlProgram.cpp',
@@ -27,9 +29,9 @@ if host_machine.system() == 'darwin'
     compiler_flags += ['-Wno-deprecated']     #for deprecated opengl
 elif cc.get_id() == 'emscripten'
     target_opengles = true
-    gl_dep = dependency('GLESv3', required: false)
+    gl_dep = dependency('GLESv2', required: false)
 else
-    #find a opengl library with fallbacks
+    # find a opengl library with fallbacks
     gl_dep = dependency('GL', required: false)
 
     if not gl_dep.found()
@@ -37,12 +39,12 @@ else
     endif
 
     if not gl_dep.found()
-        gl_dep = dependency('GLESv3', required: false)
+        gl_dep = dependency('GLESv2', required: false)
         target_opengles = true
     endif
 
     if not gl_dep.found()
-        gl_dep = meson.get_compiler('cpp').find_library('GLESv3')
+        gl_dep = meson.get_compiler('cpp').find_library('GLESv2')
         target_opengles = true
     endif
 endif
@@ -55,7 +57,6 @@ endif
 
 engine_dep += [declare_dependency(
     compile_args        : gl_target_profile,
-    dependencies        : gl_dep,
     include_directories : include_directories('.'),
     sources             : source_file,
 )]
--- a/src/renderer/gl_engine/tvgGl.cpp
+++ b/src/renderer/gl_engine/tvgGl.cpp
@@ -1,0 +1,769 @@
+/*
+ * Copyright (c) 2025 the ThorVG project. All rights reserved.
+
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "tvgGl.h"
+#include "tvgCommon.h"
+
+#ifndef __EMSCRIPTEN__
+// GL_VERSION_1_0
+PFNGLCULLFACEPROC               glCullFace = nullptr;
+PFNGLFRONTFACEPROC              glFrontFace = nullptr;
+// PFNGLHINTPROC                   glHint = nullptr;
+// PFNGLLINEWIDTHPROC              glLineWidth = nullptr;
+// PFNGLPOINTSIZEPROC              glPointSize = nullptr;
+// PFNGLPOLYGONMODEPROC            glPolygonMode = nullptr;
+PFNGLSCISSORPROC                glScissor = nullptr;
+// PFNGLTEXPARAMETERFPROC          glTexParameterf = nullptr;
+// PFNGLTEXPARAMETERFVPROC         glTexParameterfv = nullptr;
+PFNGLTEXPARAMETERIPROC          glTexParameteri = nullptr;
+// PFNGLTEXPARAMETERIVPROC         glTexParameteriv = nullptr;
+// PFNGLTEXIMAGE1DPROC             glTexImage1D = nullptr;
+PFNGLTEXIMAGE2DPROC             glTexImage2D = nullptr;
+PFNGLDRAWBUFFERPROC             glDrawBuffer = nullptr;
+PFNGLCLEARPROC                  glClear = nullptr;
+PFNGLCLEARCOLORPROC             glClearColor = nullptr;
+PFNGLCLEARSTENCILPROC           glClearStencil = nullptr;
+PFNGLCLEARDEPTHPROC             glClearDepth = nullptr;
+PFNGLCLEARDEPTHFPROC            glClearDepthf = nullptr; // GLES
+// PFNGLSTENCILMASKPROC            glStencilMask = nullptr;
+PFNGLCOLORMASKPROC              glColorMask = nullptr;
+PFNGLDEPTHMASKPROC              glDepthMask = nullptr;
+PFNGLDISABLEPROC                glDisable = nullptr;
+PFNGLENABLEPROC                 glEnable = nullptr;
+// PFNGLFINISHPROC                 glFinish = nullptr;
+// PFNGLFLUSHPROC                  glFlush = nullptr;
+PFNGLBLENDFUNCPROC              glBlendFunc = nullptr;
+// PFNGLLOGICOPPROC                glLogicOp = nullptr;
+PFNGLSTENCILFUNCPROC            glStencilFunc = nullptr;
+PFNGLSTENCILOPPROC              glStencilOp = nullptr;
+PFNGLDEPTHFUNCPROC              glDepthFunc = nullptr;
+// PFNGLPIXELSTOREFPROC            glPixelStoref = nullptr;
+// PFNGLPIXELSTOREIPROC            glPixelStorei = nullptr;
+// PFNGLREADBUFFERPROC             glReadBuffer = nullptr;
+// PFNGLREADPIXELSPROC             glReadPixels = nullptr;
+// PFNGLGETBOOLEANVPROC            glGetBooleanv = nullptr;
+// PFNGLGETDOUBLEVPROC             glGetDoublev = nullptr;
+PFNGLGETERRORPROC               glGetError = nullptr;
+// PFNGLGETFLOATVPROC              glGetFloatv = nullptr;
+PFNGLGETINTEGERVPROC            glGetIntegerv = nullptr;
+PFNGLGETSTRINGPROC              glGetString = nullptr;
+// PFNGLGETTEXIMAGEPROC            glGetTexImage = nullptr;
+// PFNGLGETTEXPARAMETERFVPROC      glGetTexParameterfv = nullptr;
+// PFNGLGETTEXPARAMETERIVPROC      glGetTexParameteriv = nullptr;
+// PFNGLGETTEXLEVELPARAMETERFVPROC glGetTexLevelParameterfv = nullptr;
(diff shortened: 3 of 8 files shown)


hermet:
https://github.com/thorvg/thorvg/pull/2866

hermet:
@SergeyLebedkin Instead of duplicate PR, could you please update https://github.com/thorvg/thorvg/pull/2866 directly?

SergeyLebedkin:
Added limit ourselves to only functions from the core, without extensions and only those that are actually used in the project.
It is also desirable to load functions using native OpenGL approaches for win, linux and macos

SergeyLebedkin:
- [ ] windows
- [ ] linux
- [ ] macos
- [ ] emsdk

examples/Example.h
```diff
@@ -337,6 +337,7 @@ struct GlWindow : Window
 #endif
         window = SDL_CreateWindow("ThorVG Example (OpenGL/ES)", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE);
         context = SDL_GL_CreateContext(window);
+        SDL_GL_SetSwapInterval(0);
```
hermet, line 340: May I ask why this needs here?

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -0,0 +1,657 @@
...
+        #elif defined(__linux__)
+        #include <GL/glx.h>
+        #define GetProcAddress(procName) glXGetProcAddress((const GLubyte *)procName);
+        #elif defined(__APPLE__) || defined(__MACH__)
+        #define GetProcAddress(procName) 
+        #endif
+    #endif
+#endif
+
+#if !defined(THORVG_GL_TARGET_GLES) && !defined(__APPLE__) && !defined(__MACH__) && !defined(__EMSCRIPTEN__)
```
hermet, line 39: This allows this runtime binding only for windows... this should be clarified , Our purpose is to support the runtime bindings for whole systems

hermet (changes requested):
@SergeyLebedkin please check my comments. Thanks.
+ Please verify that ThorVG requires a minimum of OpenGL version 3.0 and OpenGL ES version 3.0. Thanks.

SergeyLebedkin:
> * Please verify that ThorVG requires a minimum of OpenGL version 3.0 and OpenGL ES version 3.0. Thanks.

We have it. See _GlRenderer::gen_

src/renderer/gl_engine/tvgGl.cpp
SergeyLebedkin: No, not only for Windows, but for Linux and other platforms.
Apple, emsdk, GLES make automatic binding of OpenGL functions

examples/Example.h
SergeyLebedkin: This disables vertical sync, allowing for performance better than 17ms per frame.

src/renderer/gl_engine/tvgGl.cpp
SergeyLebedkin: Also note that Linux and Windows have different sets of automatic binding functions than Linux. Windows automatically loads version 1.1, and Linux loads up to 1.5.

hermet: 1. Got it. Let's forget about macOS for now. Linux systems might still require this, since there are various chipset vendors and not all of them guarantee full support. I recommend using runtime binding on Linux, as it's the standard practice.

3. This versioning check can be better placed in the `initOpenGL()`
image

4. Let's put the `initOpenGL()` call in the first boot.
```
    if (rendererCnt == -1) {
        initOpenGL();
    }
 ```

**Review by hermet — COMMENTED** [review]
Thanks, Please preserve the original author `Ruiwen` for this PR.
you can use the command for changing the author with:
`git commit --amend --author="RuiwenTang <tangruiwen1989@gmail.com>
" `
while you can add co-author in the commit message with yours:
`Co-Authored-By: Sergii Liebodkin <sergii@lottiefiles.com>`

Please double-check the `gl_engine/meson.build` whether gl dependency can be removed there for cases.

File: src/renderer/gl_engine/tvgGl.h [hunk 3]
```diff
@@ -0,0 +1,370 @@
...
+    #define GL_CHECK(stmt) stmt
+#else
+    #ifdef _DEBUG
+    #define GL_CHECK(stmt) stmt; assert(glGetError() == GL_NO_ERROR);
+    #else
+    #define GL_CHECK(stmt) stmt
+    #endif
+#endif
+
+#if !defined(THORVG_GL_TARGET_GLES) && !defined(__APPLE__) && !defined(__MACH__) && !defined(__EMSCRIPTEN__)
```
hermet, line 61: same to cpp

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -0,0 +1,670 @@
...
+        #define GetProcAddress(procName) wglGetProcAddress(procName)
+        #elif defined(__linux__)
+        #include <GL/glx.h>
+        #define GetProcAddress(procName) glXGetProcAddress((const GLubyte *)procName)
+        #endif
+    #endif
+#endif
+
+#if !defined(THORVG_GL_TARGET_GLES) && !defined(__APPLE__) && !defined(__MACH__) && !defined(__EMSCRIPTEN__)
+#if defined(_WIN32)
```
hermet, line 38: this _WIN32 scope looks weird to me. `glActiveTexture` and `glBlendEquation` are not window-specific features, could you please double-check it? otherwise it should be removed.

hermet (commented):
1. Please check more comments
2. it would be nicer adding a space between `description` and `co-authored-by`
image

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -0,0 +1,670 @@
...
+        #if defined(_WIN32)
+        #define GetProcAddress(procName) wglGetProcAddress(procName)
+        #elif defined(__linux__)
+        #include <GL/glx.h>
+        #define GetProcAddress(procName) glXGetProcAddress((const GLubyte *)procName)
+        #endif
+    #endif
+#endif
+
+#if !defined(THORVG_GL_TARGET_GLES) && !defined(__APPLE__) && !defined(__MACH__) && !defined(__EMSCRIPTEN__)
```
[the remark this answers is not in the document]
SergeyLebedkin: OpenGL **ES** no needs any dynamic binding

src/renderer/gl_engine/tvgGl.cpp
SergeyLebedkin: > this _WIN32 scope looks weird to me. `glActiveTexture` and `glBlendEquation` are not window-specific features, could you please double-check it? otherwise it should be removed.

Yes, it looks weird. But in practice, automatic function binding by opengl versions looks like this:
windows - 1.1
linux - 1.4
macos - 4.1

src/renderer/gl_engine/tvgGl.cpp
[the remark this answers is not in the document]
hermet: @SergeyLebedkin Got it, If you are sure, it's ok, otherwise, please check glad does this or not.

And Just for refererence:
image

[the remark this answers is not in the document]
SergeyLebedkin: the difference between glad and our implementation is that glad does not reference standard opengl header files at all, but completely redefines functions.
This solves the problem of incompatibility of header files on different platforms (for example, linux and windows have different sets of function prototypes in gl.h)
but glad's approach is to load functions directly from dynamic libraries, the names of which are strictly hardcoded.
our approach is to make maximum use of native platform function binding methods, including both GL and GLES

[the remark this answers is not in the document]
hermet: > This solves the problem of incompatibility of header files on different platforms (for example, linux and windows have different sets of function prototypes in gl.h)

@SergeyLebedkin One of the ideas behind dynamic binding is that chipset vendors may not fully support all GLES features. We are trying to runtime bind APIs and check whether they are supported or not. Thus gles is also expected working with runtime binding as well.

src/renderer/gl_engine/tvgGl.cpp
hermet: > Yes, it looks weird. But in practice, automatic function binding by opengl versions looks like this: windows - 1.1 linux - 1.4 macos - 4.1

@SergeyLebedkin could you please share where this information comes from?

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1527,15 +1527,11 @@ GlRenderer* GlRenderer::gen(TVG_UNUSED uint32_t threads)
...
     //initialize engine
     if (rendererCnt == -1) {
-        //TODO: GL minimum version check, should be replaced with the runtime linking in GlRenderer::init()
-        GLint vMajor, vMinor;
-        glGetIntegerv(GL_MAJOR_VERSION, &vMajor);
-        glGetIntegerv(GL_MINOR_VERSION, &vMinor);
-        if (vMajor < TVG_REQUIRE_GL_MAJOR_VER || (vMajor ==  TVG_REQUIRE_GL_MAJOR_VER && vMinor <  TVG_REQUIRE_GL_MINOR_VER)) {
-            TVGERR("GL_ENGINE", "OpenGL/ES version is not satisfied. Current: v%d.%d, Required: v%d.%d", vMajor, vMinor, TVG_REQUIRE_GL_MAJOR_VER, TVG_REQUIRE_GL_MINOR_VER);
+        auto init = initOpenGL();
+        if (!init) {
```
hermet, line 1531: `if (!initOpenGL()) {`

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -0,0 +1,670 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "tvgGl.h"
+
+// get opengl procedure address
+#if !defined(THORVG_GL_TARGET_GLES)
+    #ifndef GetProcAddress
```
hermet, line 27: Suggest to use `glGetProcAddress` as the name, following our naming convention.

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -0,0 +1,670 @@
...
+
+    #ifdef _WIN32
+    // // GL_VERSION_1_2
+    // glDrawRangeElements = (PFNGLDRAWRANGEELEMENTSPROC)GetProcAddress("glDrawRangeElements");
+    // glTexImage3D = (PFNGLTEXIMAGE3DPROC)GetProcAddress("glTexImage3D");
+    // glTexSubImage3D = (PFNGLTEXSUBIMAGE3DPROC)GetProcAddress("glTexSubImage3D");
+    // glCopyTexSubImage3D = (PFNGLCOPYTEXSUBIMAGE3DPROC)GetProcAddress("glCopyTexSubImage3D");
+    
+    // // GL_VERSION_1_3
+    glActiveTexture = (PFNGLACTIVETEXTUREPROC)GetProcAddress("glActiveTexture");
```
hermet, line 371: It seems we need a failure case for when the driver doesn't support the API, using a specific magic macro to cover all APIs here.
ie)
```
if (!glActiveTexture) {
    TVGERR("...."); // print that glActiveTexture is not supported.
    return false;
}
```

examples/Example.h
```diff
@@ -353,7 +353,8 @@ struct GlWindow : Window
         //Free in the reverse order of their creation.
         delete(canvas);
         canvas = nullptr;
-
+        // terminate engine before SDL context destruction
+        tvg::Initializer::term();
```
hermet, line 357: Please separate this if it’s not closely related to this API bindings

src/renderer/gl_engine/tvgGl.cpp
SergeyLebedkin: We are already check gl version and opengl have a strictly defined functions set for each version. So I think it's not necessary to check each function

examples/Example.h
SergeyLebedkin: You mean in separated commit?

hermet: both fine.

hermet:
Remark (TODO)
1. Load gl/gles symbols on runtime and binding necessary gl apis on runtime. use `dlopen()` and `LoadLibrary()`
2. Remove gl header inclusion instead define essential declarations in `tvgGl.h`
3. Remove GL library dependency in compile time.
4. Mac deprecated OpenGL support so we can exclude the runtime binding support for that platform.

src/renderer/gl_engine/meson.build
```diff
@@ -22,31 +24,6 @@ source_file = [
...
-    endif
-
-    if not gl_dep.found()
-        gl_dep = meson.get_compiler('cpp').find_library('GLESv3')
-        target_opengles = true
-    endif
-endif
-
 if target_opengles
     gl_target_profile = '-DTHORVG_GL_TARGET_GLES=1'
```
hermet, line 28: This looks like dead code. I think we can leave the build script above for confirming OpenGL/OpenGL ES.

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -0,0 +1,763 @@
...
+bool openglInitLibs() {
+    #ifndef THORVG_GL_TARGET_GLES
+    if (!libOpenGL) libOpenGL = dlopen("libGL.so", RTLD_LAZY);
+    if (!libOpenGL) libOpenGL = dlopen("libGL.so.0", RTLD_LAZY);
+    if (!libOpenGL) return false;
+    if (!glGetProcAddress) glGetProcAddress = (PFNGLGETPROCADDRESSPROC)dlsym(libOpenGL, "glXGetProcAddress");
+    if (!glGetProcAddress) glGetProcAddress = (PFNGLGETPROCADDRESSPROC)dlsym(libOpenGL, "glXGetProcAddressARB");
+    if (!glGetProcAddress) return false;
+    #else
+    if (!libOpenGL) libOpenGL = dlopen("libGLESv2.so", RTLD_LAZY);
```
hermet, line 390:
```
   libOpenGL = dlopen("libGLESv2.so", RTLD_NOW);
   if (!libOpenGL) _gles3_handle = dlopen("libGLESv2.so.2.0", RTLD_NOW);
   if (!libOpenGL) _gles3_handle = dlopen("libGLESv2.so.2", RTLD_NOW);
```

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -0,0 +1,763 @@
...
+// linux
+#elif defined(__linux__)
+#include <dlfcn.h>
+void* libOpenGL = nullptr;
+typedef void* (APIENTRYP PFNGLGETPROCADDRESSPROC)(const char*);
+PFNGLGETPROCADDRESSPROC glGetProcAddress = nullptr;
+// load opengl libraries
+bool openglInitLibs() {
+    #ifndef THORVG_GL_TARGET_GLES
+    if (!libOpenGL) libOpenGL = dlopen("libGL.so", RTLD_LAZY);
```
hermet, line 383: please check this, we might need to fallback for cases with `RTLD_NOW`
```
libOpenGL = dlopen("libGL.so", RTLD_NOW);
if (!libOpenGL) libOpenGL  = dlopen("libGL.so.4", RTLD_NOW);
if (!libOpenGL) libOpenGL  = dlopen("libGL.so.3", RTLD_NOW);
```

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -0,0 +1,763 @@
...
+}
+// load opengl proc address from glXGetProcAddress
+void* openglGetProcAddress(const char* procName) {
+    #ifndef THORVG_GL_TARGET_GLES
+    return glGetProcAddress(procName);
+    #else
+    return dlsym(libOpenGL, procName);
+    #endif
+}
+// macos
```
hermet, line 403: --

src/renderer/gl_engine/tvgGl.h
```diff
@@ -0,0 +1,1511 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+ #ifndef _TVG_GL_H_
+ #define _TVG_GL_H_
+
+  #ifdef __EMSCRIPTEN__
```
hermet, line 26: indentation

src/renderer/gl_engine/tvgGl.h
```diff
@@ -0,0 +1,1511 @@
...
+#ifndef __EMSCRIPTEN__
+#if defined (THORVG_GL_TARGET_GLES)
+    #define TVG_REQUIRE_GL_MAJOR_VER 3
+    #define TVG_REQUIRE_GL_MINOR_VER 0
+#else
+    #define TVG_REQUIRE_GL_MAJOR_VER 3
+    #define TVG_REQUIRE_GL_MINOR_VER 3
+#endif
+
+#ifdef _DEBUG
```
hermet, line 42: please use the proper indentation for readibility. (other lines as well)
```
#ifdef XXX
    -----------
#endif
```

hermet:
replaced with https://github.com/thorvg/thorvg/pull/3445

PR #3428, opened 2025-04-30
wg_engine: fixed target texture update

BenjaminHalko:
Allows the target texture to be updated without forcing the size to changed.


SergeyLebedkin (approved):
Make sence! Thanx

PR #3431, opened 2025-05-02
gl_engine: replace std::string with C-style strings for portability

tinyjin:
Replace `std::string` usage with C-style strings in GL engine to improve ABI stability and cross-language compatibility.

- C-style strings provide consistent binary interface for FFI
- Reduces potential issues where standard library is limited (e.g., WebAssembly, embedded systems)


SergeyLebedkin (approved):
The same approach used in the webgpu renderer for dynamic shader composition.

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -98,9 +98,21 @@ void GlRenderer::initShaders()
 {
     mPrograms.reserve((int)RT_None);
 
+    char linearGradientFragShader[8192];
+    strcpy(linearGradientFragShader, STR_GRADIENT_FRAG_COMMON_VARIABLES);
+    strcat(linearGradientFragShader, STR_LINEAR_GRADIENT_VARIABLES);
```
hermet, line 103: recommend to use `snprintf()` for optimal.

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -98,9 +98,25 @@ void GlRenderer::initShaders()
 {
     mPrograms.reserve((int)RT_None);
 
+    char linearGradientFragShader[8192];
```
hermet, line 101: `8192` looks unnecessarily large. I guess `2800` and `5300` looks enough to use.

Otherwise you might wanna check this:

```
#define STR_TOTAL_LENGTH (sizeof(STR_GRADIENT_FRAG_COMMON_VARIABLES) - 1 + \
                          sizeof(STR_RADIAL_GRADIENT_VARIABLES) - 1 + \
                          sizeof(STR_GRADIENT_FRAG_COMMON_FUNCTIONS) - 1 + \
                          sizeof(STR_RADIAL_GRADIENT_MAIN) - 1 + \
                          1)  // +1 for null terminator

char radialGradientFragShader[STR_TOTAL_LENGTH];
```

tinyjin: Shader strings are typed in const char*, `sizeof` returns the size of pointer. I used `strlen`

hermet (approved):

PR #3434, opened 2025-05-05
wg_engine: fixed target surface update

BenjaminHalko:
Allows the target **surface** to be updated without forcing the size to changed.


BenjaminHalko:
I had missed this in #3428

SergeyLebedkin (approved):

Issue #3425, opened 2025-04-29
web: remove emscripten dependency


hermet:
It's good to discuss removing the Emscripten dependency to improve portability.

See: https://schellcode.github.io/webassembly-without-emscripten

PR #3435, opened 2025-05-05
web: support disabling emscripten

BenjaminHalko:
By moving `__EMSCRIPTEN__` to `THORVG_USE_EMSCRIPTEN` this allows users to force disable Emscripten, even if `__EMSCRIPTEN__` is already defined.

By default, nothing will change, but if you `#define THORVG_USE_EMSCRIPTEN false` it will turn off Emscripten support.

I know the full removal of Emscripten is planned (#3425), but this PR serves to act as a work around in the meantime.
Feel free to close this if it's too "hacky", or if a better solution is created.

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -60,6 +60,10 @@ public: \
 #define _TVG_DECLARE_ACCESSOR(A) \
     friend A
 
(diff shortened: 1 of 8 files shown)


BenjaminHalko:
Honestly, this work around is too janky, even for me. So I'll close this PR.

PR #3437, opened 2025-05-06
gl_engine: move Emscripten WebGL context handling out of GL engine

tinyjin:
- Removed direct inclusion of `<emscripten/html5_webgl.h>` from gl engine.
- Moved Emscripten-specific WebGL context creation and management to own WASM binding layer.

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -63,19 +63,6 @@ void GlRenderer::flush()
 }
 
 
-void GlRenderer::currentContext()
-{
-#ifdef __EMSCRIPTEN__
-    auto targetContext = (EMSCRIPTEN_WEBGL_CONTEXT_HANDLE)mContext;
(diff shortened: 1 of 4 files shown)


src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -63,19 +63,6 @@ void GlRenderer::flush()
...
 
-void GlRenderer::currentContext()
-{
-#ifdef __EMSCRIPTEN__
-    auto targetContext = (EMSCRIPTEN_WEBGL_CONTEXT_HANDLE)mContext;
-    if (emscripten_webgl_get_current_context() != targetContext) {
-        emscripten_webgl_make_context_current(targetContext);
-    }
-#else
-    TVGERR("GL_ENGINE", "Maybe missing MakeCurrent() Call?");
```
hermet, line 74: We can implement it here for portability. ThorVG users typically don't have deep knowledge of the ThorVG engine internals, so it's not ideal to let them manage the context themselves.

PR #3439, opened 2025-05-06
lottie: interpret null as a float value of 0

mgrudzinska:
It may happen that a float value is saved as null. In such cases, it should be interpreted as 0 without raising an error.

Zrzut ekranu 2025-05-6 o 19 53 41


hermet:
Here is the question: In JSON, a number value can be null? Otherwise, the resource should be fixed to handle this correctly.

hermet:
@mgrudzinska please check this: https://github.com/thorvg/thorvg/pull/3442

mgrudzinska:
handled in #3442

PR #3440, opened 2025-05-06
lottie: enhance parsing of position's SeparateCoords

mgrudzinska:
The "s" key does not have to appear at the beginning of the position block. Previously, such cases would result in a parsing error — now they are handled correctly.

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.h
+++ b/src/loaders/lottie/tvgLottieModel.h
@@ -639,6 +639,12 @@ struct LottieTransform : LottieObject
         LottieFloat y = 0.0f;
     };
 
+    SeparateCoord* separateCoord()
+    {
+        if (!coords) coords = new SeparateCoord;
+        return coords;
+    }
+
     struct RotationEx
     {
         LottieFloat x = 0.0f;
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -566,11 +566,14 @@ LottieTransform* LottieParser::parseTransform(bool ddd)
             enterObject();
             while (auto key = nextObjectKey()) {
                 if (KEY_AS("k")) parsePropertyInternal(transform->position);
-                else if (KEY_AS("s") && getBool()) transform->coords = new LottieTransform::SeparateCoord;
-                //check separateCoord to figure out whether "x(expression)" / "x(coord)"
-                else if (transform->coords && KEY_AS("x")) parseProperty(transform->coords->x);
-                else if (transform->coords && KEY_AS("y")) parseProperty(transform->coords->y);
-                else if (KEY_AS("x") && expressions) transform->position.exp = getExpression(getStringCopy(), comp, context.layer, context.parent, &transform->position);
+                else if (KEY_AS("x")) {
+                    //check separateCoord to figure out whether "x(expression)" / "x(coord)"
+                    if (peekType() == kStringType) {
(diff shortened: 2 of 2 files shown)


src/loaders/lottie/tvgLottieModel.h
```diff
@@ -637,8 +637,15 @@ struct LottieTransform : LottieObject
     {
         LottieFloat x = 0.0f;
         LottieFloat y = 0.0f;
+        bool apply = true;
```
hermet, line 640: Please consider to remove this additional property. I don't see its really necessary.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -566,11 +566,15 @@ LottieTransform* LottieParser::parseTransform(bool ddd)
             enterObject();
             while (auto key = nextObjectKey()) {
                 if (KEY_AS("k")) parsePropertyInternal(transform->position);
-                else if (KEY_AS("s") && getBool()) transform->coords = new LottieTransform::SeparateCoord;
-                //check separateCoord to figure out whether "x(expression)" / "x(coord)"
-                else if (transform->coords && KEY_AS("x")) parseProperty(transform->coords->x);
-                else if (transform->coords && KEY_AS("y")) parseProperty(transform->coords->y);
-                else if (KEY_AS("x") && expressions) transform->position.exp = getExpression(getStringCopy(), comp, context.layer, context.parent, &transform->position);
+                else if (KEY_AS("s")) transform->separateCoord()->apply = getBool();
```
hermet, line 569: You removed `getBool()`, but isn't split supposed to be enabled only when its value is true?
I guess you can keep the status quo:
` else if (KEY_AS("s") && getBool()) transform->coords = new LottieTransform::SeparateCoord;`

mgrudzinska: it seems that 's' always is true:
Zrzut ekranu 2025-05-7 o 11 53 07

in AE I don;t see a possibility to change the 's' value to be 'false'. if I change this value manually the file imports to AE with an error. for lottie-web 's=false' is also not a correct data. I haven't seen a real example with 's' set to false. 

and it may happen that 'x' and 'y' are given before 's', so coords needs to be created independently from s value, I think:
```
        "p": {
          "x": {
            "a": 0,
            "k": 250,
            "ix": 3
          },
          "y": {
            "a": 0,
            "k": 100,
            "ix": 4
          },
          "s": true
        }
```

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -566,11 +566,15 @@ LottieTransform* LottieParser::parseTransform(bool ddd)
             enterObject();
             while (auto key = nextObjectKey()) {
                 if (KEY_AS("k")) parsePropertyInternal(transform->position);
-                else if (KEY_AS("s") && getBool()) transform->coords = new LottieTransform::SeparateCoord;
-                //check separateCoord to figure out whether "x(expression)" / "x(coord)"
-                else if (transform->coords && KEY_AS("x")) parseProperty(transform->coords->x);
-                else if (transform->coords && KEY_AS("y")) parseProperty(transform->coords->y);
-                else if (KEY_AS("x") && expressions) transform->position.exp = getExpression(getStringCopy(), comp, context.layer, context.parent, &transform->position);
+                else if (KEY_AS("s")) getBool(); //always true - ignored
```
hermet, line 569: in this condition, this line can be removable.

Issue #3452, opened 2025-05-09
The cropped edge of the rotated image is 1 pixel less


FASTSHIFT:
When rotating an image, it is found that one pixel is cropped off the edge of the image. The cropping is correct when the image is not rotated.

Reproduction code:
```c
using namespace tvg;

#define IMG_WIDTH 227
#define IMG_HEIGHT 149

static void tvg_draw_rect(Canvas* canvas, int x, int y, int w, int h)
{
    auto shape = Shape::gen();
    shape->appendRect(x, y, w, h, 0, 0);
    shape->fill(0xff, 0x00, 0x00);
    canvas->push(std::move(shape));
}

int tvg_image_test(void)
{
    Initializer::init(0);

    auto picture = Picture::gen();

    auto buf_size = IMG_WIDTH * IMG_HEIGHT * sizeof(uint32_t);
    auto img_buf = (uint32_t*)malloc(buf_size);
    memset(img_buf, 0xDD, buf_size);
    auto res = picture->load(img_buf, IMG_WIDTH, IMG_HEIGHT, ColorSpace::ARGB8888, true);

    if (res != Result::Success) {
        LV_LOG_ERROR("Failed to load the image");
        return -1;
    }

    float w, h;
    res = picture->size(&w, &h);
    if (res != Result::Success) {
        LV_LOG_ERROR("Failed to get the image size");
        return -1;
    }

    LV_LOG_USER("Image size: %f x %f", w, h);

    auto draw_buf = lv_draw_buf_create(480, 480, LV_COLOR_FORMAT_ARGB8888, LV_STRIDE_AUTO);

    if (!draw_buf) {
        LV_LOG_ERROR("Failed to create draw buffer");
        return -1;
    }

    lv_draw_buf_clear(draw_buf, NULL);

    auto shape = Shape::gen();
    shape->moveTo(139.000000, -48.000000);
    shape->lineTo(139.000000, 179.000000);
    shape->lineTo(-10.000000, 179.000000);
    shape->lineTo(-10.000000, -48.000000);
    shape->close();

    auto bounding_box = Shape::gen();
    bounding_box->appendRect(0, 0, 130, 130, 0, 0);
    shape->clip(std::move(bounding_box));

    picture->clip(std::move(shape));

    /* Rotate the picture 90 degrees around the center of the image */
    tvg::Matrix matrix = {0, -1, 139, 1, 0, -48, 0, 0, 1};
    picture->transform(matrix);

[...]

PR #3457, opened 2025-05-12
sw_engine: replacing a cast with rounding

mgrudzinska:
Casting a float to an int instead of rounding resulted in a visibly incorrect reduction of the rendered area.

@Issue: https://github.com/thorvg/thorvg/issues/3452

### DIFF
--- a/src/renderer/sw_engine/tvgSwRasterTexmap.h
+++ b/src/renderer/sw_engine/tvgSwRasterTexmap.h
@@ -113,8 +113,8 @@ static void _rasterBlendingPolygonImageSegment(SwSurface* surface, const SwImage
     y = yStart;
 
     while (y < yEnd) {
-        x1 = (int32_t)_xa;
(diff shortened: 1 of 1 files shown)


hermet:
@mgrudzinska could you please double-check with this? I couldn't see any improvement.

```    
    {
        if (!canvas) return false;

        //Open file manually
        ifstream file(EXAMPLE_DIR"/image/test.png", ios::binary);
        if (!file.is_open()) return false;
        auto begin = file.tellg();
        file.seekg(0, std::ios::end);
        auto size = file.tellg() - begin;
        auto data = (char*)malloc(size);
        file.seekg(0, std::ios::beg);
        file.read(data, size);
        file.close();

        if (!tvgexam::verify(picture->load(data, size, "png", "", true))) return false;
        free(data);
        picture->translate(600, 0);
        picture->rotate(90);
        canvas->push(picture);

        return true;
    }
```
sample resource:
test

PR #3459, opened 2025-05-13
lottie: handle time remapping with zero value correctly

tinyjin:
|                      | Previous                                                                 | Patched                                                                  |
|--------------------------|---------------------------------------------------------------------------|---------------------------------------------------------------------------|
| Rendering                | Rendering (Previous) | Rendering (Patched) |
| Difference (First 10frs) | Diff (Previous)     | Diff (Patched)     |

Example file: [1f3b2.json](https://github.com/user-attachments/files/20186838/1f3b2.json)

The time remapping logic had a regression bug where animations with zero value "tm"(Time Remap) were not being processed correctly.

This was happening because the previous implementation only applied time remapping when the value was non-zero. (A zero-value Time Remap should be applied, but is ignored)

CleanShot 2025-05-13 at 20 16 30@2x

The fix changes the default time remap value to `-1.0` (since frame values cannot be negative) and applies time remapping when the value is 0.0 or greater. This ensures that zero value time remapping is properly processed.


hermet:
image

@tinyjin This doesn't seem to be a regression bug. :-)

tinyjin:
> image
> 
> @tinyjin This doesn't seem to be a regression bug. :-)

@hermet Yes, I updated commit description!

hermet (approved):

PR #3465, opened 2025-05-14
lottie: update text using def font

mgrudzinska:
In cases where the JSON does not include glyphs
information, an attempts is made to update the text 
using the any loaded font, regardless of the specified 
font path origin.

@Issue: https://github.com/thorvg/thorvg.swift/issues/6

result:
Zrzut ekranu 2025-05-14 o 23 38 44

sample:
[org.json](https://github.com/user-attachments/files/20214642/org.json)

NOTE:
1. text is not correctly placed - this will be solved in a separate pr.
2. lottie specs are not clear about the default font path origin value - we have samples in which it is not specified for both, embedded and local

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -934,7 +934,7 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
 
     if (!p || !text->font) return;
 
-    if (text->font->origin != LottieFont::Origin::Embedded) {
+    if (text->font->origin != LottieFont::Origin::Embedded || text->font->chars.empty()) {
         _fontText(doc, layer->scene);
         return;
     }


monkeyswarm:
Note that the erratic text placement is likely within my sample lottie file itself, since it renders like that even in other programs! So thorvg placement is likely working just fine in this case. Thanks again for working on this.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -934,7 +934,7 @@ void LottieBuilder::updateText(LottieLayer* layer, float frameNo)
 
     if (!p || !text->font) return;
 
-    if (text->font->origin != LottieFont::Origin::Embedded) {
+    if (text->font->origin != LottieFont::Origin::Embedded || text->font->chars.empty()) {
```
hermet, line 937: This doesn't helpful, if any dummy glyph are existed.

hermet (changes requested):
The user resource omits the origin attribute, and expect to use the "local" font.

I think it should be to "local (0)" than "embedded" in thorvg side by default.

image

hermet:
To add an additional opinion, we could set 'local' as the default, but for local fonts, we need a way to check whether glyphs exist **per each character**. If they don't, engine fall back to using a system font character.

hermet:
see: https://github.com/thorvg/thorvg/pull/3467

Issue #3397, opened 2025-04-11
text: advanced feature support


hermet:
These are candidate features for text/font enhancement that ThorVG is currently missing. While not yet confirmed, they are worth investigating and potentially supporting going forward.

**Essential**
- [x] Text Alignment
- [x] Multiline Layout (character/word/auto/...)
- [x] support Line Feed (`\n`)
- [ ] Basic Glyph Metrics Provider and Cursor

**Decoration (Optional)**
- [x] Outline
- [ ] Underline
- [ ] Strikethrough

**Formats**
- [ ] TTC
- [ ] OTF / OTC
- [ ] AAT?
- [ ] WOFF / WOFF2 ?

**Advanced**
- [ ] Bidirectional Text
- [ ] Complex Script Shaping (ligatures/alternative characters/etc)
- [ ] Rich Text(markup/paragraph)
- [ ] Vertical Texts

**Optimization**
- [x] Glyph caching

**Relevants:**
- https://github.com/thorvg/thorvg/issues/2731
- https://github.com/thorvg/thorvg/issues/3258
- https://github.com/thorvg/thorvg/issues/2961
- https://github.com/thorvg/thorvg/issues/3307

hermet:
**Basic Glyph Metrics Provider and Cursor**

The most essential properties for the font metrics will be:
- Origin
- LSB(Left Side Bearing)  / TSB(Top Side Bearing)
- Ascent
- Descent
- Baseline
- Advanced Width

![Image](https://github.com/user-attachments/assets/fbe6223c-2de6-41ab-bdc1-3222f991e2d3)

We can start with writing a simple text document with thorvg canvas: ie)

![Image](https://github.com/user-attachments/assets/57010e85-7683-4c2c-868b-7faeaed4b173)

xland:
Hope to provide text measure Apis.
I've been waiting for a long time.

xland:
When will TTC and WOFF2 be supported?
Thanks.

hermet:
> When will TTC and WOFF2 be supported? Thanks.

@xland Hmm... no exact plan yet. do you expect to use it in your project?

xland:
Yes, I plan to develop a simple GUI framework based on Thorvg, which allows users to render text with system fonts. On the Windows platform, most system fonts are in TTC format（msyh.ttc）.
In addition, fonts in WOFF2 format are smaller in size and deliver better rendering quality than TTF fonts.

lpogic:
[...]

PR #3466, opened 2025-05-14
ttf_loader: changing the reference point for text

mgrudzinska:
After the change, the reference point for text is the left edge of the baseline.

Zrzut ekranu 2025-05-14 o 23 51 38

Zrzut ekranu 2025-05-15 o 00 02 11

TODO: some changes in the svg loader


hermet:
@mgrudzinska I think we should first focus on returning essential glyph metrics information. Without it, changing the reference point isn't helpful because users have no idea where to place the line origin. Please reach a consensus on the concept of the feature enhancement and its API usage before moving forward with implementation. This will help save your time. Thanks.

https://github.com/thorvg/thorvg/issues/3397

hermet (changes requested):
Changing paint geometry system looks not correct direction.

Issue #3469, opened 2025-05-16
Stop element after gradient affects the SVG


MewPurPur:
I saw this while demoing gradients in my app:
![Image](https://github.com/user-attachments/assets/e97abfc5-8f9f-4b9a-9136-453c7fc3c773)

```
<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
	<linearGradient id="b" spreadMethod="reflect" x2="12.1%" y2="5%">
		<stop stop-color="#ffe345" offset="0.15" />
		<stop stop-color="#ff4545" offset="0.5" stop-opacity="0.4" />
	</linearGradient>
	<stop offset="0.85" />
	<circle r="6" cx="8" cy="8" fill="url(#b)" />
</svg>
```

In fact, you can have multiple stops and they will all affect the gradient.

This doesn't affect production-ready SVGs, but in an SVG editor like mine it can confuse users.

hermet:
@MewPurPur The fix will be in v0.15.13 / v1.0.0-pre21. Thanks.

PR #3470, opened 2025-05-19
svg_loader: Fix <stop> being registered in closed latestGradient

JSUYA:
If stop is declared immediately after Gradient is closed, the stop is registered in latestGradient.
Change latestGradient to a Stack and add a stop only to the last added gradient.
When that gradient is closed, no more stops should be registered for that gradient.
And gradients do not allow nested declarations, so only the earliest declared Gradient is valid.

related issue: https://github.com/thorvg/thorvg/issues/3469

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -3401,6 +3401,13 @@ static void _svgLoaderParserXmlClose(SvgLoaderData* loader, const char* content,
         }
     }
 
+    for (unsigned int i = 0; i < sizeof(gradientTags) / sizeof(gradientTags[0]); i++) {
+        if (!strncmp(tagName, gradientTags[i].tag, sz)) {
+            loader->gradientStack.pop();
+            break;
+        }
+    }
+
     for (unsigned int i = 0; i < sizeof(graphicsTags) / sizeof(graphicsTags[0]); i++) {
         if (!strncmp(tagName, graphicsTags[i].tag, sz)) {
             loader->currentGraphicsNode = nullptr;
@@ -3472,7 +3479,6 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content, 
         if (node->type != SvgNodeType::Defs || !empty) {
             loader->stack.push(node);
         }
-        loader->latestGradient = nullptr;
     } else if ((method = _findGraphicsFactory(tagName))) {
         if (loader->stack.count > 0) parent = loader->stack.last();
         else parent = loader->doc;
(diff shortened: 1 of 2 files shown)


JSUYA:
test svg
```html
<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
	<linearGradient id="b" spreadMethod="reflect" x2="12.1%" y2="5%">
		<stop stop-color="#ffe345" offset="0.15" />
	        <linearGradient id="c" spreadMethod="reflect" x2="12.1%" y2="5%">
          		<stop stop-color="#00FFFF" offset="0.15" />
         		<stop stop-color="#00FF45" offset="0.5" stop-opacity="0.4" />
         	</linearGradient>
		<stop stop-color="#ff4545" offset="0.5" stop-opacity="0.4" />
	</linearGradient>
	<stop offset="0.85" />
	<circle r="3" cx="3" cy="3" fill="url(#b)" />
	<circle r="3" cx="8" cy="8" fill="url(#c)" />
</svg>

```

mgrudzinska (approved):
It works fine - it's even consistent with the `loader->stack` logic.
What bothers me a bit is that only one element ever goes onto the gradStack and we also have a separate mechanism for tracking the open tag: `OpenedTagType`.
I’d probably implement it in a way that’s also aligned with what we already have:
```
@@ -3401,6 +3401,13 @@ static void _svgLoaderParserXmlClose(SvgLoaderData* loader, const char* content,
+    for (unsigned int i = 0; i < sizeof(gradientTags) / sizeof(gradientTags[0]); i++) {
+        if (!strncmp(tagName, gradientTags[i].tag, sz)) {
+            loader->openedTag = OpenedTagType::Other;
+            break;
+        }
+    }
+

@@ -3499,8 +3506,9 @@ static void _svgLoaderParserXmlOpen(SvgLoaderData* loader, const char* content,
             loader->gradients.push(gradient);
         }
         loader->latestGradient = gradient;
+        if (!empty) loader->openedTag = OpenedTagType::Gradient;
     } else if (STR_AS(tagName, "stop")) {
-        if (!loader->latestGradient) {
+        if (!loader->latestGradient || loader->openedTag != OpenedTagType::Gradient) {
             TVGLOG("SVG", "Stop element is used outside of the Gradient element");
             return;
         }

+++ b/src/loaders/svg/tvgSvgLoaderCommon.h
@@ -580,7 +580,8 @@ enum class OpenedTagType : uint8_t
 {
     Other = 0,
     Style,
-    Text
+    Text,
+    Gradient
 };
```

At some point it might be worth unifying the logic of these approaches, but it's not the goal of this PR.

PR #3473, opened 2025-05-20
lottie: fix precomposition with masking

mgrudzinska:
A precomposition layer is clipped to its viewport. If the same layer also had a mask that was optimized using clipping, this clip was unintentionally overridden by the viewport clipping. This conflict is now fixed.

before:
Zrzut ekranu 2025-05-21 o 00 39 51

after (same as in AE, lottie web):
Zrzut ekranu 2025-05-21 o 00 40 49

sample:
[test.json](https://github.com/user-attachments/files/20357644/test.json)

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -1174,8 +1174,8 @@ void LottieBuilder::updateMasks(LottieLayer* layer, float frameNo)
 {
     if (layer->masks.count == 0) return;
 
-    //Introduce an intermediate scene for embracing the matte + masking
-    if (layer->matteTarget) {
+    //Introduce an intermediate scene for embracing matte + masking or precomp clipping + masking replaced by clipping
+    if (layer->matteTarget || layer->type == LottieLayer::Precomp) {
         auto scene = Scene::gen();
         scene->push(layer->scene);
         layer->scene = scene;


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1202,9 +1195,21 @@ void LottieBuilder::updateMasks(LottieLayer* layer, float frameNo)
             //Cheaper. Replace the masking with a clipper
             if (layer->masks.count == 1 && compMethod == MaskMethod::Alpha) {
                 layer->scene->opacity(MULTIPLY(layer->scene->opacity(), opacity));
+                //Introduce an intermediate scene for embracing the precomp clipping + masking replaced by clipping
```
hermet, line 1198: Please check in 1177:
```
    if (layer->matteTarget || layer->type == LottieLayer::Precomp) {
        auto scene = Scene::gen();
        scene->push(layer->scene);
        layer->scene = scene;
    }
```

mgrudzinska: the reasons why I didn't want to  combine these two are:
- matte + masking - an intermediate scene is _**not needed**_ if masking is replaced by clipping
- precomp + masking - an intermediate scene is _**needed**_ only if masking is replaced by clipping

but I understand that code is not so clear now, so I'll change this

PR #3475, opened 2025-05-21
ttf/api: add metrics() api

mgrudzinska:
Introduced the GlyphMetrics struct and the API:
  Result metrics(GlyphMetrics** metrics, uint32_t* size)
This allows retrieving glyph metrics for a given Text object. The font and utf8 content must be set beforehand using the font() and text() apis.

Memory for the metrics array is allocated internally by the API. It is the user's responsibility to free it by calling the same API with a valid metrics pointer and a nullptr passed as size.

@Issue: https://github.com/thorvg/thorvg/issues/3397

metrics

Zrzut ekranu 2025-05-22 o 07 57 13

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -284,6 +284,23 @@ struct Matrix
 
 
 /**
+ * @brief A data structure representing metrics for a single glyph.
+ *
+ * @since Experimental API
+ */
+struct GlyphMetrics
+{
+    Point kerning = {0.0f, 0.0f};
+
+    float advanceWidth;
+    float leftSideBearing;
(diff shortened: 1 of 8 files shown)


mgrudzinska:
draft because I need to double check the meaning of the values in GlyphMetrics. In this example, I applied both #3466 and the updated definition of yOffset

hermet:
@mgrudzinska Hello, I don't see any necessity to use glyph metrics for the sample you attached. I think you can demonstrate this with any practical example, such as a text document. See:  https://github.com/thorvg/thorvg/issues/3397#issuecomment-2900597109

PR #3479, opened 2025-05-22
Enable builds on Fedora using Packit

### LINKED ISSUE #3371: Infra: enhanced the github test process by integrating Fedora builds.
Welcome.  Fedora has a build infrastructure https://packit.dev/ if it would be helpful to test commits on that. Can also use a container to run a Fedora build within GitHub actions.  Can make pull requests to add one of these if that would be helpful.

_Originally posted by @bkmgit in https://github.com/thorvg/thorvg/issues/3370#issuecomment-2765252391_

bkmgit:
Closes https://github.com/thorvg/thorvg/issues/3371

To get it to work, someone with commit access will need to agree to make a Fedora account at:
https://accounts.fedoraproject.org/

and agree to Fedora Code of Conduct, then follow the steps at:

http://packit.dev/docs/guide#1-set-up-packit-integration

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -1,0 +1,17 @@
+# https://packit.dev/docs/configuration/
+
+specfile_path: buildtools/thorvg.spec
+
+notifications:
+  pull_request:
+    successful_build: true
+
+jobs:
+- job: copr_build
+  trigger: pull_request
+  targets:
+  - fedora-rawhide-aarch64
+  - fedora-rawhide-i386
+  - fedora-rawhide-ppc64le
+  - fedora-rawhide-s390x
+  - fedora-rawhide-x86_64
--- a/buildtools/thorvg.spec
+++ b/buildtools/thorvg.spec
@@ -1,0 +1,109 @@
+Name:           thorvg
+Version:        0.15.12
+Release:        %{autorelease}
+Summary:        Lightweight vector-based scenes and animation drawing library
+
+License:        MIT
+URL:            https://www.thorvg.org/
+Source:         https://github.com/thorvg/thorvg/archive/v%{version}/thorvg-%{version}.tar.gz
+# Give binaries names that are not generic to prevent conflicts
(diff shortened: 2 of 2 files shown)


bkmgit:
@hermet Is anything else needed from me for this?

hermet:
> @hermet Is anything else needed from me for this?

@bkmgit hi, thanks for contribution. Still there are 2 thing we might need a confident to apply this.

1. Adding a Fedora-specific configuration introduces a dependency on a particular Linux distribution, which could compromise ThorVG's platform neutrality. Therefore, it might be more appropriate to place such scripts under a dedicated directory like **thorvg/packaging/fedora/**, and include any necessary CI scripts there. Maybe you can update this.

2. This is the current concern for me. From ThorVG’s perspective, it’s unclear whether this script is truly valid or maintainable. We’re not sure if it might break due to updates elsewhere. In other words, we need to carefully consider the long-term maintenance responsibilities involved. (How to solve this?)

bkmgit:
> > @hermet Is anything else needed from me for this?
> 
> @bkmgit hi, thanks for contribution. Still there are 2 thing we might need a confident to apply this.
> 
>     1. Adding a Fedora-specific configuration introduces a dependency on a particular Linux distribution, which could compromise ThorVG's platform neutrality. Therefore, it might be more appropriate to place such scripts under a dedicated directory like **thorvg/packaging/fedora/**, and include any necessary CI scripts there. Maybe you can update this.
> 
>     2. This is the current concern for me. From ThorVG’s perspective, it’s unclear whether this script is truly valid or maintainable. We’re not sure if it might break due to updates elsewhere. In other words, we need to carefully consider the long-term maintenance responsibilities involved. (How to solve this?)

The spec file can be moved.  The packit.yaml file needs to be at the root of the repository though does not need to be packaged or distributed in releases.

Long term maintenance is unclear.  Expect if it becomes too burdensome, it should be easy to remove.  It need not block updates, so trying it should not be a big effort.

Issue #3489, opened 2025-05-28
wg_engine: geometry stage buffer


SergeyLebedkin:
Implement geometry stage buffer.

Accumulate geometry data into a stage buffer during frame rendering and flush it to the GPU in single call instead of flush data per each path.

hermet:
Further way to go: fill settings staged buffer application.

Issue #3455, opened 2025-05-12
wg_engine: task-based rendering


SergeyLebedkin:
Implement task besed rendering.

Rendering occurs in 2 stages:
1. Collecting data about objects in the current frame - **creating the tasks**
2. Running the actual rendering, based on full information about the frame configuration, including custom blends, compositions and scenes - **running the tasks**

This will reduce the number of context switches and will further help reduce the number of data flashes in the GPU

PR #3491, opened 2025-05-28
wg_engine: geometry stage buffers implementation

SergeyLebedkin:
Implemented task-based rendering and geometry stage buffers:
1. Get information about current frame objects
2. Accumulate geometry data into a stage buffer during frame rendering
3. Flush it to the GPU in single call
4. Run rendering process in post render stage

https://github.com/thorvg/thorvg/issues/3489
https://github.com/thorvg/thorvg/issues/3455

Lottie example performance:
```
main:    83 ms
staging: 67 ms
~25% boost
```

thorvg-wasm.wasm (linux, emsdk)
```
staging: 929864 bytes
main:    927727 bytes
```

1. Current Implementation: Per-Shape GPU Buffer Management
In the existing rendering pipeline, each geometric shape undergoes a distinct process of tessellation and subsequent GPU buffer allocation. Upon each geometry update, the tessellator processes the shape's path data, generating vertices and indices. These resulting data are then immediately written to dedicated GPU buffers—one vertex buffer and one index buffer per shape.

The workflow can be summarized as follows:

1.1. Prepare Stage (Current)
For each shape, the tessellation and GPU buffer flushing occur independently:

shape path 0 → tessellator → render data 0 (GPU vertex buffers flush)
shape path 1 → tessellator → render data 1 (GPU vertex buffers flush)
shape path 2 → tessellator → render data 2 (GPU vertex buffers flush)
... (and so on for all shapes)

This approach necessitates the creation of a unique set of vertex and index buffers on the GPU for every individual shape. Consequently, the system performs multiple write buffers calls, each targeting a distinct set of GPU buffers.

1.2. Render Stage (Current)
During the rasterization stage, each shape's dedicated GPU buffers are drawn sequentially:

render data 0 draw
render data 1 draw
render data 2 draw
... (and so on for all shapes)

1.3. Analysis of Current Approach
While straightforward, this method introduces considerable overhead. The repeated allocation of GPU memory for each shape and, more critically, the numerous individual write buffers calls, contribute to increased CPU-to-GPU communication latency. This frequent flushing of small data chunks can saturate the command queue and reduce overall rendering throughput.

2. Proposed Optimization: Consolidated GPU Buffer Management
The new, optimized approach aims to mitigate the inefficiencies of the current implementation by centralizing geometry data management. Instead of immediately flushing tessellated data to the GPU, the geometry of shapes is first stored in system (CPU) memory. During the rendering phase, the data for all shapes intended for drawing are aggregated into a single, common stage buffer before a consolidated flush operation to the GPU.

The optimized workflow is detailed below:

2.1. Prepare Stage (New)
In the preparation phase, tessellation still occurs per shape, but the resulting render data is retained in CPU memory:

```
shape path 0 → tessellator → render data 0 (CPU buffers)
shape path 1 → tessellator → render data 1 (CPU buffers)
shape path 2 → tessellator → render data 2 (CPU buffers)
... (and so on for all shapes)
```

2.2. Render Stage (New)
The key optimization occurs in the render stage. Geometry data from all shapes to be drawn are accumulated into a single, temporary common stage buffer. This buffer is then flushed to the GPU in a single, efficient operation:

```
render data 0 →
render data 1 → common stage buffer
render data 2 →
... (all accumulated)
common stage buffer flush
```

2.3. Raster Stage (New)
After the consolidated flush, the GPU can then draw the shapes from the now populated GPU buffers, which effectively contain the combined geometry data:

```
render data 0 draw
render data 1 draw
render data 2 draw
... (all drawn from the consolidated buffer)

```
2.4. Benefits of the New Approach
This optimized strategy offers several significant advantages:
- Reduced GPU Flushes: By accumulating data into a single stage buffer, the number of write buffers calls to the GPU is drastically reduced to just one per frame (or per batch of shapes). This minimizes CPU-to-GPU communication overhead.
- Improved Data Locality: Consolidating geometry data before transfer can lead to better cache utilization on the GPU side, as related data is transferred contiguously.
- Simplified GPU Memory Management: While still requiring GPU buffers, the management becomes simpler as a large, single buffer (or a few large buffers) can be used, rather than numerous small, fragmented ones.

### DIFF
--- a/src/renderer/wg_engine/meson.build
+++ b/src/renderer/wg_engine/meson.build
@@ -7,6 +7,7 @@ source_file = [
     'tvgWgRenderData.h',
     'tvgWgRenderer.h',
     'tvgWgRenderTarget.h',
+    'tvgWgRenderTask.h',
     'tvgWgShaderSrc.h',
     'tvgWgShaderTypes.h',
     'tvgWgBindGroups.cpp',
@@ -17,6 +18,7 @@ source_file = [
     'tvgWgRenderData.cpp',
     'tvgWgRenderer.cpp',
     'tvgWgRenderTarget.cpp',
+    'tvgWgRenderTask.cpp',
     'tvgWgShaderSrc.cpp',
     'tvgWgShaderTypes.cpp'
 ]
--- a/src/renderer/wg_engine/tvgWgCommon.cpp
+++ b/src/renderer/wg_engine/tvgWgCommon.cpp
@@ -269,3 +269,34 @@ void WgContext::releaseQueue(WGPUQueue& queue)
         queue = nullptr;
     }
 }
+
+
+WGPUCommandEncoder WgContext::createCommandEncoder()
+{
+    WGPUCommandEncoderDescriptor commandEncoderDesc{};
+    return wgpuDeviceCreateCommandEncoder(device, &commandEncoderDesc);
+}
+
+
+void WgContext::submitCommandEncoder(WGPUCommandEncoder commandEncoder)
+{
+    const WGPUCommandBufferDescriptor commandBufferDesc{};
+    WGPUCommandBuffer commandsBuffer = wgpuCommandEncoderFinish(commandEncoder, &commandBufferDesc);
+    wgpuQueueSubmit(queue, 1, &commandsBuffer);
+    wgpuCommandBufferRelease(commandsBuffer);
+}
+
+
+void WgContext::releaseCommandEncoder(WGPUCommandEncoder& commandEncoder)
+{
+    if (commandEncoder) {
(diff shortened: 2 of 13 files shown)


hermet:
This is what I see with this PR: 

**92 ms -> 83 ms (~10% boost)**

hermet (commented):
Thanks, Please review the code once more and remove all the nonsensical comments. Comments are not meant to describe what the code does — they are meant to provide context as to why the code is there. In many places, your comments are just restating what the code already makes obvious. For example:
```
// here a plus b  
int result = a + b;
```
This kind of comment is unhelpful for people and only adds unnecessary noise to the code.

hermet (approved):

PR #3507, opened 2025-06-06
gl_engine: fix always-true clear flag

Nor-s:
The clear flag specified in Canvas::draw is ignored when set to false, since GlRenderer::mClearBuffer is never explicitly reset to false.

This commit ensures that mClearBuffer is reset to its default (false) after being used, so that the clear operation behaves correctly per draw.

- Note: GlComposeTask: always clear since the bound framebuffer is selfFbo
    - `GlComposeTask::run`: `GL_CHECK(glBindFramebuffer(GL_FRAMEBUFFER, getSelfFbo()));`
    -  `GlBlitTask::run`: `GL_CHECK(glBindFramebuffer(GL_FRAMEBUFFER, getTargetFbo()));`
- Issue: #1779
- History: 
    - https://github.com/thorvg/thorvg/commit/548962f5f8868429b9ebf5ce6fb9005dcb9a606a#diff-9e1e59292fa7b422d5ff866d9b48e16bdd61ab18bed965fa9d1a6b809bca7ee0R110
    - https://github.com/thorvg/thorvg/commit/cc2fa23359b9c07c13cb6cedf11298386d80517c#diff-9e1e59292fa7b422d5ff866d9b48e16bdd61ab18bed965fa9d1a6b809bca7ee0L111

### reproduce

```cpp
// Example.h
void tempFunc(){}
void (*ClearColor)() = tempFunc;

// Example.h, Window::draw
    bool draw()
    {
        ClearColor();
        //...
    }
// Example.h, Window::ready
    bool ready()
    {
        if (!canvas) return false;

        if (!example->content(canvas, width, height)) return false;

        //initiate the first rendering before window pop-up.
        ClearColor();
        if (!verify(canvas->draw())) return false;
```

```cpp
// examples/Shapes.cpp
#include <OpenGL/gl3.h>
void GL_ClearColor()
{
    glClearColor(1.0f,0.5f,0.25f,1.0f);
    glClear(GL_COLOR_BUFFER_BIT);
}

// examples/Shapes.cpp
{
    ClearColor = GL_ClearColor;
```
|main branch| fixed branch|
|-|-|
|image|image |

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderTask.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderTask.cpp
@@ -195,23 +195,20 @@ void GlComposeTask::run()
 {
     GL_CHECK(glBindFramebuffer(GL_FRAMEBUFFER, getSelfFbo()));
 
-    // clear this fbo
-    if (mClearBuffer) {
-        // we must clear all area of fbo
-        GL_CHECK(glViewport(0, 0, mFbo->getWidth(), mFbo->getHeight()));
-        GL_CHECK(glScissor(0, 0, mFbo->getWidth(), mFbo->getHeight()));
-        GL_CHECK(glClearColor(0, 0, 0, 0));
-        GL_CHECK(glClearStencil(0));
-    #ifdef THORVG_GL_TARGET_GLES
-        GL_CHECK(glClearDepthf(0.0));
-    #else
-        GL_CHECK(glClearDepth(0.0));
-    #endif
-        GL_CHECK(glDepthMask(1));
+    // we must clear all area of fbo
+    GL_CHECK(glViewport(0, 0, mFbo->getWidth(), mFbo->getHeight()));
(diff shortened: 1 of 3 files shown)


src/renderer/gl_engine/tvgGlRenderTask.h
```diff
@@ -125,15 +125,17 @@ class GlComposeTask : public GlRenderTask
     void run() override;
 
     void setRenderSize(uint32_t width, uint32_t height) { mRenderWidth = width; mRenderHeight = height; }
+    void setClearBuffer(bool clearBuffer) { mClearBuffer = clearBuffer; }
```
hermet: Useless setter/getter is not preferred.
suggest:
```
public:
bool mClearBuffer
```

Nor-s: Updated the code.

hermet (approved):
@Nor-s could you please add yourself? https://github.com/thorvg/thorvg/blob/main/CONTRIBUTORS

PR #3509, opened 2025-06-07
jpg: removed setjmp usage for better portability

hermet:
Replaced non-portable setjmp/longjmp-based error handling with conventional return-based error propagation.
This approach improves compatibility with Rust and other environments that do not support stack manipulation via jump buffers.


BenjaminHalko (approved):
Incredible! LGTM!

PR #3378, opened 2025-04-02
lotttie: added pucker/bloat support

mgrudzinska:
DOES NOT BUILD without #3515 

issue: https://github.com/thorvg/thorvg/issues/2681

PR #3515, opened 2025-06-09
lottie: modifier chaining enhanced

mgrudzinska:
DONE:
- allowed modifiers in any order - previously, rounding was always applied first
- added support for LineTo command in the rounding algorithm - required when allowing other modifiers before rounding 

TODO:
- double check polystar
- double check offset alg (for rounding point after offset are important even if colinear)
- dd sample

pr needed for #3378


src/loaders/lottie/tvgLottieBuilder.h
```diff
@@ -174,7 +174,7 @@ struct LottieBuilder
     void updateRoundedCorner(LottieGroup* parent, LottieObject** child, float frameNo, Inlist<RenderContext>& contexts, RenderContext* ctx);
     void updateOffsetPath(LottieGroup* parent, LottieObject** child, float frameNo, Inlist<RenderContext>& contexts, RenderContext* ctx);
 
-    RenderPath buffer;   //resusable path
+    RenderPath buffer[2];   //resusable path
```
hermet: Please add descriptions as comments for [0] and [1], explaining the purpose of each.

src/loaders/lottie/tvgLottieModifier.h
```diff
@@ -31,44 +31,33 @@
...
-        //roundness -> offset
-        if (next->type == Roundness) {
-            next->next = this;
-            return next;
-        }
-
-        //just in the order.
-        this->next = next;
-        return this;
+        //prevent cyclic decoration: 1) self-decoration: a->decorate(a); 2) mutual decoration: a->decorate(b); b->decorate(a);
```
hermet, line 51: This sounds odd to me. Would ThorVG allow this scenario? 
1. Could Lottie produce this scenario?
2. So this situation occurs — is exception result the expected one?
3. if you are just adding just in case, what about? a -> b -> c -> d -> a ?

If this is a scenario that cannot occur under the Lottie specification, it would be better to simply remove it. But if it's an exceptional case and you have it by any chance, then adding a log would be more appropriate to properly handle it later.

src/loaders/lottie/tvgLottieModifier.h
```diff
@@ -31,44 +31,33 @@
...
+    bool chained = false;
 
+    LottieModifier(RenderPath* buffer) : buffer(buffer) {}
     virtual ~LottieModifier() {}
 
     virtual bool modifyPath(PathCommand* inCmds, uint32_t inCmdsCnt, Point* inPts, uint32_t inPtsCnt, Matrix* transform, RenderPath& out) = 0;
     virtual bool modifyPolystar(RenderPath& in, RenderPath& out, float outerRoundness, bool hasRoundness) = 0;
 
-    LottieModifier* decorate(LottieModifier* next)
+    LottieModifier* decorate(LottieModifier* prev)
```
hermet: I'd recommend keep the decorator concept here. The subject -being decorated with other(next)- can be oneself, not someone(prev) else.

src/loaders/lottie/tvgLottieBuilder.h
mgrudzinska: done

src/loaders/lottie/tvgLottieModifier.h
mgrudzinska: done, also in tvgLottieBuilder.h l.100

src/loaders/lottie/tvgLottieModifier.h
mgrudzinska: log added

1. yes, check the sample: [error.json](https://github.com/user-attachments/files/20704391/error.json)
Crashes on main because rounded corners are applied one after another (because in `decorate`, `this` is assigned to `this->next`)

2. For now ThorVG doesn't fully meet the spec.

Based on my testing (AE vs ThorVG):
- rounded corner: for a given object, only the first rounded corners modifier should be applied - ThorVG instead picks the one with the largest rounding value
- offset: all offset modifiers should be considered. If there are other modifiers in between, the order must be preserved - ThorVG only supports a single one.

These are things that need to be addressed, but not within the scope of this PR. The goal here is to prevent a crash.

3. the scenario will be handled as well. the last step, chaining 'a' will not happen, because a->chained is true

src/loaders/lottie/tvgLottieModifier.h
```diff
@@ -31,44 +31,37 @@
...
     LottieModifier* decorate(LottieModifier* next)
     {
-        /* TODO: build the decorative chaining here.
-           currently we only have roundness and offset. */
-
-        //roundness -> offset
-        if (next->type == Roundness) {
-            next->next = this;
-            return next;
+        //prevent cyclic decoration: 1) self-decoration: a->decorate(a); 2) mutual decoration: a->decorate(b); b->decorate(a);
```
hermet, line 51: when something handled in the future
"TODO: xxxxxx"

src/loaders/lottie/tvgLottieModifier.h
```diff
@@ -31,44 +31,37 @@
...
+        if (next->chained || next == this) {
+            TVGLOG("LOTTIE", "Decoration skipped to prevent cyclic chain with modifier: %p", next);
```
hermet: TVGERR("");

src/loaders/lottie/tvgLottieModifier.h
hermet: > 1. yes, check the sample: [error.json](https://github.com/user-attachments/files/20704391/error.json)
>    Crashes on main because rounded corners are applied one after another (because in `decorate`, `this` is assigned to `this->next`)

probably, we can add a issue spearately.

mgrudzinska: https://github.com/thorvg/thorvg/issues/3539

src/loaders/lottie/tvgLottieModifier.h
mgrudzinska: done

mgrudzinska:
> Please check comments: For optimal maintenance and size, how about replacing all LineTo usage with CubicTo? Then we wouldn't need additional line handling in certain cases.

@hermet please check now. I removed support for 'lineTo' everywhere I could. but since we have polygons and rectangles, it was not possible to remove all lineTo cases.

mgrudzinska:
> @hermet please check now. I removed support for 'lineTo' everywhere I could. but since we have polygons and rectangles, it was not possible to remove all lineTo cases.

the above is outdated already.
this pr contains changes only with chaining - as a result if #3556 is not applied, samples with the very first modifier other than the round corners may render wrong/not at all.

src/loaders/lottie/tvgLottieModifier.h
```diff
@@ -34,7 +34,11 @@ struct LottieModifier
     enum Type : uint8_t {Roundness = 0, Offset};
 
     LottieModifier* next = nullptr;
+    RenderPath* buffer;
     Type type;
+    bool chained = false;
```
hermet, line 39: This variable looks an unnecessary complexity. doesn't mean a valid `next` chained?

src/loaders/lottie/tvgLottieModifier.h
```diff
@@ -43,29 +47,27 @@ struct LottieModifier
...
-        /* TODO: build the decorative chaining here.
-           currently we only have roundness and offset. */
-
-        //roundness -> offset
-        if (next->type == Roundness) {
-            next->next = this;
-            return next;
+        //TODO: resolve the possible cyclic decoration by adding support for multiple modifiers of the same type in the RenderContext
+        //prevent cyclic decoration: 1) self-decoration: a->decorate(a); 2) mutual decoration: a->decorate(b); b->decorate(a);
+        if (next->chained || next == this) {
```
hermet, line 52: Looks wrong order.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -391,9 +391,9 @@ static void _repeat(LottieGroup* parent, Shape* path, RenderContext* ctx)
 }
 
 
-void LottieBuilder::appendRect(Shape* shape, Point& pos, Point& size, float r, bool clockwise, RenderContext* ctx)
+void LottieBuilder::appendRect(Shape* shape, Point& pos, Point& size, float r, bool clockwise, RenderContext* ctx, LottieModifier* modifier)
```
hermet: Please double-check if possible to use `ctx` rather than an additional parameter.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -417,38 +418,56 @@ void LottieBuilder::updateRect(LottieGroup* parent, LottieObject** child, float
     auto size = rect->size(frameNo, tween, exps);
     auto pos = rect->position(frameNo, tween, exps) - size * 0.5f;
     auto r = rect->radius(frameNo, tween, exps);
+    auto modifier = ctx->modifier;
```
hermet: --

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -417,38 +418,56 @@ void LottieBuilder::updateRect(LottieGroup* parent, LottieObject** child, float
...
 
     if (r == 0.0f)  {
-        if (ctx->roundness) ctx->roundness->modifyRect(size, r);
+        //roundness is the first modifier -> it can be applied before the shape's path is established
```
hermet: I think this can be delegated to the relying modifier itself, allowing it to properly handle the logic internally by receiving the current shape target.

ie:
//modifies could query the target type (rect, ellipse, path, polystar, etc)
`modifyPath(... LottieShape* target)`


you could also improve all the if-else condition other places with the above approach.

hermet (changes requested):

src/loaders/lottie/tvgLottieModifier.h
mgrudzinska: @hermet  can you please rephrase this sentence: "doesn't mean a valid `next` chained?"? 
I introduced it to handle cases ```a->b->...->a```

|current chain|add|inside decorate|
|-|-|-|
|a|b|```a->chaining = true; return b->a```|
|b->a|c|```b->chaining = true; return c->b->a```|
|c->b->a|a|```a->chaining == true; return c->b->a```|
without chaining check:
|c->b->a|a|```c->chaining = true; return a->c->b->a->....``` - CYCLE|

src/loaders/lottie/tvgLottieModifier.h
```diff
@@ -43,29 +47,27 @@ struct LottieModifier
...
+            TVGERR("LOTTIE", "Decoration skipped to prevent cyclic chain with modifier: %p", next);
+            return this;
         }
+        this->chained = true;
 
-        //just in the order.
```
hermet, line 54: Rather change the method concept, Please consider modifying the callers.

src/loaders/lottie/tvgLottieModifier.h
hermet: Got it,  Do you have any sample where the exception occurs at line 53?

mgrudzinska: [cycle.json](https://github.com/user-attachments/files/20871464/cycle.json)

Issue #3505, opened 2025-06-05
wg_engine: uniform stage buffers


SergeyLebedkin:
Implement stage buffers for diffferent types of rasterization settings
1. transforms
2. blend settings (color space and opacity)
3. solid color
4. linear/radial gradients

PR #3522, opened 2025-06-11
wg_engine: uniform stage buffers implementation

SergeyLebedkin:
Introduced stage buffer for uniforms to reduce number of memory shafles from cpu to gpu memory

**Refactor: Unified Uniform Buffer for Improved Performance**
This pull request introduces a stage buffer for uniforms, significantly reducing memory transfers calls between the CPU and GPU.

**Problem**
Previously, each shape managed its transformation settings, opacity, colors, and gradients in separate data structures. This necessitated the creation of individual uniform buffers on the GPU, leading to a high volume of data transfers calls from system memory to video memory.

**Solution**
Now, all shape-related data (fills and strokes settings) are consolidated into a single, shared data structure. During rendering, this data is accumulated on demand into a single buffer, which can then be written to video memory in a single call.

**WebGPU Considerations**
This approach leverages WebGPU's bind groups to link buffer data with named uniforms in shaders. There were a few key considerations:

 - Bind Group Creation Cost: Creating bind groups can be a moderately expensive operation. If the buffer's memory address changes, all referencing bind groups must be recreated, which can impact performance. To mitigate this, both the buffer and bind groups can now be pre-allocated.
 - Offset Alignment: WebGPU's _minUniformBufferOffsetAlignment_ (see spec 3.6.2. Limits) requires bind group offsets within the buffer to be a multiple of 256 bytes. Our current total data size for all necessary information is 192 bytes, leaving a 64-byte reserve to accommodate this alignment.

**Impact**
This change required modifications to the shaders, pipelines, and intra-shader data structures. Ultimately, by using a common buffer for shape properties, we've reduced data write requests to video memory in complex scenarios from **tens of thousands** to just **one**. This also allows us to **allocate fewer** buffer handles, further optimizing resource usage. This has resulted in a significant performance increase.

Tested on windows/macos/linux/emsdk
see: https://github.com/thorvg/thorvg/issues/3505

### DIFF
--- a/src/renderer/wg_engine/tvgWgBindGroups.cpp
+++ b/src/renderer/wg_engine/tvgWgBindGroups.cpp
@@ -104,8 +104,14 @@ WGPUBindGroup WgBindGroupLayouts::createBindGroupStrorage3RO(WGPUTextureView tex
 
 WGPUBindGroup WgBindGroupLayouts::createBindGroupBuffer1Un(WGPUBuffer buff)
 {
+    return createBindGroupBuffer1Un(buff, 0, wgpuBufferGetSize(buff));
+}
+
+
+WGPUBindGroup WgBindGroupLayouts::createBindGroupBuffer1Un(WGPUBuffer buff, uint64_t offset, uint64_t size)
+{
     const WGPUBindGroupEntry bindGroupEntrys[] = {
-        { .binding = 0, .buffer = buff, .size = wgpuBufferGetSize(buff) }
+        { .binding = 0, .buffer = buff, .offset = offset, .size = size }
     };
     const WGPUBindGroupDescriptor bindGroupDesc { .layout = layoutBuffer1Un, .entryCount = 1, .entries = bindGroupEntrys };
     return wgpuDeviceCreateBindGroup(device, &bindGroupDesc);
--- a/src/renderer/wg_engine/tvgWgBindGroups.h
+++ b/src/renderer/wg_engine/tvgWgBindGroups.h
@@ -49,6 +49,7 @@ public:
     WGPUBindGroup createBindGroupStrorage2RO(WGPUTextureView texView0, WGPUTextureView texView1);
     WGPUBindGroup createBindGroupStrorage3RO(WGPUTextureView texView0, WGPUTextureView texView1, WGPUTextureView texView2);
     WGPUBindGroup createBindGroupBuffer1Un(WGPUBuffer buff);
+    WGPUBindGroup createBindGroupBuffer1Un(WGPUBuffer buff, uint64_t offset, uint64_t size);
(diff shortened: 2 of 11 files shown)


SergeyLebedkin:
```
Lottie example

// windows 3060 Ti
main: 68
3522: 45
+34%

// macos (intel)
main: 695
3522: 133
x5
```

mgrudzinska:
macos intel + asan & undef 
```
../src/renderer/wg_engine/tvgWgRenderData.cpp:216:25: runtime error: downcast of address 0x6080000bd5a0 which does not point to an object of type 'RadialGradient'
0x6080000bd5a0: note: object is of type 'LinearGradientImpl'
 03 00 00 00  78 6e d2 0b 01 00 00 00  b0 d5 0b 00 80 60 00 00  48 6e d2 0b 01 00 00 00  40 60 15 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for 'LinearGradientImpl'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/renderer/wg_engine/tvgWgRenderData.cpp:216:25
```

perf 🔥 
main - 760 ms
this pr - 154 ms 💥

cpu mem - seems like there is some mem leak (maybe on main as well, but since it is so slow, it does not grow so much so fast):
Zrzut ekranu 2025-06-12 o 12 09 52
and growing:
Zrzut ekranu 2025-06-12 o 12 11 17

SergeyLebedkin:
> SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/renderer/wg_engine/tvgWgRenderData.cpp:216:25

fixed

SergeyLebedkin:
> cpu mem - seems like there is some mem leak (maybe on main as well, but since it is so slow, it does not grow so much so fast): 

Looks like we have mem leak... Try to investigate. Tnx!

SergeyLebedkin:
> > cpu mem - seems like there is some mem leak (maybe on main as well, but since it is so slow, it does not grow so much so fast):
> 
> Looks like we have mem leak... Try to investigate. Tnx!

Seems to all its not a problem of this PR, but general problem of the wg renderer. Also exists in main

mgrudzinska:
perf 🔥

1) macos intel:
main - 760 ms
this pr - 154 ms 💥💥💥💥

2) macos arm: 
main: 101 ms
this pr: 46 ms 💥

mgrudzinska (approved):
since the mem leak is not connected with tthis pr, rather simpler to observe, approved

SergeyLebedkin:
Linux (i5-1135G7/MX250, weak HW)
main: 550 ms
this pr: 550 ms
Looks like CPU is a bottle neck in this case

hermet:
Intel Core Ultra5 (Intel Arc): +45%
Size: +1.7kb

PR #3523, opened 2025-06-11
lottie: fix offsetting partially degenerated cubics 

mgrudzinska:
Until now, cases where a Bezier curve had `start == ctrl1` or
`ctrl2 == end` were incorrectly treated as linear segments.
This led to incorrect rendering, for example when offsetting
a polystar with outer roundness > 0 and inner roundness == 0.
Now such cases handled as proper curves with full cubic behavior.

before:
Zrzut ekranu 2025-06-11 o 23 34 48

after:
Zrzut ekranu 2025-06-11 o 23 35 10

AE (even here visible some artefacts):
Zrzut ekranu 2025-06-11 o 23 37 05
Zrzut ekranu 2025-06-11 o 23 37 01

samples:
[sample1.json](https://github.com/user-attachments/files/20698386/sample1.json)
[sample2.json](https://github.com/user-attachments/files/20698387/sample2.json)

### DIFF
--- a/src/loaders/lottie/tvgLottieModifier.cpp
+++ b/src/loaders/lottie/tvgLottieModifier.cpp
@@ -27,6 +27,12 @@
 /* Internal Class Implementation                                        */
 /************************************************************************/
 
+static bool _colinear(const Point* p)
+{
+    return tvg::zero(*p - *(p + 1)) && tvg::zero(*(p + 2) - *(p + 3));
+}
+
+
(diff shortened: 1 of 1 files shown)


mgrudzinska:
is ready, but I want to test more (more complicated cases can be tested after modifiers chaining is applied - more combinations of lineTo/cubicTo possible)

tinyjin:
https://github.com/thorvg/thorvg/pull/4248

PR #3529, opened 2025-06-13
wg_engine: prevent potential pipeline deadlock when resizing

tinyjin:
There are two points causing deadlock on WebGPU's animation rendering pipeline when resizing.

These occurred in specific case, resize() is fired while animation is being loaded.
1. renderPassEncoder is available, and causing runtime error by assert. → changed logic, ends render pass if it is being proceeded.

2. `stageBufferGeometry` isn't prepared when canvas->sync() called before preRender → Add condition to ensure stageBuffer isn't null


tinyjin:
#3530

PR #3531, opened 2025-06-13
wg_engine: remove dead condition in target method

tinyjin:
Remove redundant instance/device/target checks in the second condition, since they are already early-exited in the first condition.


SergeyLebedkin (approved):
nice catch!

PR #3537, opened 2025-06-16
capi: Redefine object interfaces to use opaque pointer types

hermet:
Redefined Tvg_XXX and related types as pointers
e.g., typedef struct _Tvg_Canvas* Tvg_Canvas)
to improve encapsulation and clarity in API usage.

issue: https://github.com/thorvg/thorvg/issues/3116

### DIFF
--- a/examples/Capi.cpp
+++ b/examples/Capi.cpp
@@ -32,29 +32,29 @@
 /* Capi Test Code                                                       */
 /************************************************************************/
 
-static Tvg_Canvas* canvas = NULL;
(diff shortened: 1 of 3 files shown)


hermet:
API breaks, but abstracts & simplifies the interface. Let's carefully confirm.

hermet:
cc: @theashraf @BenjaminHalko

vtorri:
what about adding a `_p` to the struct names to say that it is a pointer ?

PR #3540, opened 2025-06-17
lottie: use tvg::zero function for point comparison

mgrudzinska:
Previously, modifiers used a local _zero() function to avoid overly precise point comparisons. However, this approach could lead to division by zero in certain cases. The internal function has been removed in favor of the shared tvg::zero.
Additionally, shape construction (for polygon and star) was updated to ensure proper closure, preventing such comparison issues from arising.

sample: 
[rect_ro.json](https://github.com/user-attachments/files/20766165/rect_ro.json)

before:
Zrzut ekranu 2025-06-17 o 02 19 19

after:
Zrzut ekranu 2025-06-17 o 02 25 11


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -595,6 +602,8 @@ void LottieBuilder::updateStar(LottiePolyStar* star, float frameNo, Matrix* tran
         angle += dTheta * direction;
         longSegment = !longSegment;
     }
+    //ensure proper shape closure - may be important for modifiers
```
hermet: This comment is just ambuguous, not helpful.  "may be important for modifiers"

Please make it sure or remove them.

hermet:
> Previously, modifiers used a local _zero() function to avoid overly precise point comparisons. However, this approach could lead to division by zero in certain cases. The internal function has been removed in favor of the shared tvg::zero.

You can also apply the refactoring separately.

src/loaders/lottie/tvgLottieBuilder.cpp
mgrudzinska: changed but now present in https://github.com/thorvg/thorvg/pull/3560

hermet (approved):

hermet:
https://github.com/thorvg/thorvg/commit/2679880bc3c967d438da1f2bea62ee117ed77c58

PR #3549, opened 2025-06-17
lottie: avoid precision issues in radial gradient

mgrudzinska:
If the focal point is exactly on the edge of the gradient's end circle, numerical errors may occur due to rounding, including potential division by zero. To ensure stability and consistency across the engines the focal point is slightly moved inward.

Note: 
Focal point support in the SW and WG engines is consistent with the SVG 1.1 standard, whereas the GL engine aligns with the SVG 2.0 standard. This change ensures compatibility with AE regardless of the engine's internal implementation.

before:
Jun-17-2025 23-38-15

after:
Jun-17-2025 23-35-33

samples:
[focal_HL-100.json](https://github.com/user-attachments/files/20784600/focal_HL-100.json)
[focal_HL.json](https://github.com/user-attachments/files/20784601/focal_HL.json)
[focal_HL100.json](https://github.com/user-attachments/files/20784602/focal_HL100.json)

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -519,7 +519,7 @@ Fill* LottieGradient::fill(float frameNo, uint8_t opacity, Tween& tween, LottieE
         if (tvg::zero(progress)) {
             static_cast<RadialGradient*>(fill)->radial(s.x, s.y, r, s.x, s.y, 0.0f);
         } else {
-            if (tvg::equal(progress, 1.0f)) progress = 0.99f;
+            progress = tvg::clamp(progress, -0.99f, 0.99f);
             auto startAngle = rad2deg(tvg::atan2(e.y - s.y, e.x - s.x));
             auto angle = deg2rad((startAngle + this->angle(frameNo, tween, exps)));
             auto fx = s.x + cos(angle) * progress * r;


hermet:
Eventually, users who use the gradient directly will encounter this problem. Maybe, we could either handle the exception nicely at the API level.

mgrudzinska:
> Eventually, users who use the gradient directly will encounter this problem. Maybe, we could either handle the exception nicely at the API level.

@hermet I was thinking about a global solution. But the standard that will be represented globally across engines doesn’t necessarily have to match the Lottie/SVG specs. Lottie is aligned with SVG 1.1, but not with SVG 2. gl already supports SVG 2, while the other engines do not.

what if we eventually decide that all engines should move toward SVG 2?
if we apply an api-level solution, wouldn't it be a regression for the gl engine?

hermet:
Q: Currently 3 engines have the same problem and you applied a uniform exception. The exception handling needs the alignment of the engine? Otherwise, let's focus on the this exception handling here.

> what if we eventually decide that all engines should move toward SVG 2? if we apply an api-level solution, wouldn't it be a regression for the gl engine?

Regarding the future direction, Agree: https://github.com/thorvg/thorvg/issues/3559, but please note that current our official is Tiny SVG 1.2. No svg 1.1, No SVG 2. Even Lottie seems having no any spec about it.  

> if we apply an api-level solution, wouldn't it be a regression for the gl engine?

Well, I don't think any users really care about it. Rather current inconsistent behavior sounds a bug for thorvg. If we can confirm the change before 1.0 release, it will be the best.

hermet:
The issue of radial image corruption can occur not only in Lottie but also when using the API directly, so it's best to find a fundamental solution.

PR #3550, opened 2025-06-17
sw/wg_engine: clamp focal point to the edge

mgrudzinska:
If the radial gradient's focal point lies outside
the end circle, it's projected onto the edge.
A slight inward offset is applied to avoid numerical issues.

Note:
Focal point support in the sw and wg engines is
consistent with the SVG 1.1 standard, whereas the gl engine aligns with the SVG 2.0 standard.

before (gl for comparison):
Zrzut ekranu 2025-06-18 o 00 05 27

after (gl for comparison):
Zrzut ekranu 2025-06-17 o 23 58 46

samples:
[samples.zip](https://github.com/user-attachments/files/20784913/samples.zip)


src/renderer/wg_engine/tvgWgShaderSrc.cpp
```diff
@@ -166,9 +166,18 @@ fn vs_main(in: VertexInput) -> VertexOutput {
...
     // orignal data
     let d0 = in.vGradCoord.xy - uPaintSettings.gradient.coords.xy;
-    let d1 = uPaintSettings.gradient.coords.xy - uPaintSettings.gradient.focal.xy;
+    var d1 = uPaintSettings.gradient.coords.xy - uPaintSettings.gradient.focal.xy;
     let r0 = uPaintSettings.gradient.coords.z;
     let rd = uPaintSettings.gradient.focal.z - uPaintSettings.gradient.coords.z;
+
+    let dd = dot(d1, d1);
+    //push focal to the edge if outside (slightly inside to prevent precision errors)
+    if (dd >= r0 * r0) {
```
hermet: Please check this. inverseSqrt() could be faster than sqrt() + division
```
    // Clamp focal point if it's outside the radius
    if (dd >= r0 * r0) {
        d1 = d1 * (r0 * 0.99) * inverseSqrt(dd); 
    }
```
Note: In per-pixel operations, unnecessary computations should be minimized as much as possible.

hermet:
@mgrudzinska [sw fix](https://github.com/thorvg/thorvg/commit/f5d0dda2cdd067bfb39c209c7bbe5a9b7bfcedb7) was firstly pushed, thanks.

src/renderer/wg_engine/tvgWgShaderSrc.cpp
SergeyLebedkin: @hermet @mgrudzinska 
Looks like this operations we can move to **void WgShaderTypeGradSettings::update(const Fill* fill)** on a system level to avoid per pixel opeartions?

hermet: @SergeyLebedkin @mgrudzinska please make a right correction. Thanks.

mgrudzinska: @hermet I'd rather implement inverseSqrt as a separate pr. 

Also, since the entire shader needs to be reorganized to avoid computing everything separately for each fragment, I’d suggest keeping this fix here, merge it and then handling the full shader cleanup in a separate pr. is that ok?

src/renderer/wg_engine/tvgWgShaderSrc.cpp
```diff
@@ -162,13 +162,23 @@ fn vs_main(in: VertexInput) -> VertexOutput {
...
+    var d1 = uPaintSettings.gradient.coords.xy - uPaintSettings.gradient.focal.xy;
     let r0 = uPaintSettings.gradient.coords.z;
     let rd = uPaintSettings.gradient.focal.z - uPaintSettings.gradient.coords.z;
+
+    let dd = dot(d1, d1);
+    //push focal to the edge if outside (slightly inside to prevent precision errors)
+    if (r0 * r0 - dd <= PRECISION) {
+        let f = uPaintSettings.gradient.coords.xy - r0 * (1.0 - PRECISION) * d1 / sqrt(dd);
+        d1 = uPaintSettings.gradient.coords.xy - f;
+    }
```
SergeyLebedkin: d1 = r0 * (1.0 - PRECISION) * d1 / sqrt(dd);
?

SergeyLebedkin: or we can precise uPaintSettings.gradient.focal.xy in a system code?

src/renderer/wg_engine/tvgWgShaderSrc.cpp
hermet: > @hermet I'd rather implement inverseSqrt as a separate pr.
>

@mgrudzinska Any proper reason?

SergeyLebedkin (approved):

src/renderer/wg_engine/tvgWgShaderSrc.cpp
mgrudzinska: @SergeyLebedkin @hermet I moved the logic into the update fun

src/renderer/wg_engine/tvgWgShaderTypes.cpp
```diff
@@ -139,8 +139,20 @@ void WgShaderTypeGradSettings::update(const Fill* fill)
...
+        auto dx = coords.vec[0] - focal.vec[0]; //cx - fx
+        auto dy = coords.vec[1] - focal.vec[1]; //cy - fy
+        auto dr = coords.vec[2] - focal.vec[2]; //r - fr
+        auto dd = dx * dx + dy * dy;
+        auto a = dr * dr - dd;
+        constexpr float precision = 0.01f; //retract focal point slightly from edge to avoid numerical errors
+        if (a <= precision) {
+            auto dist = sqrtf(dd);
+            focal.vec[0] = coords.vec[0] - coords.vec[2] * (1.0f - precision) * dx / dist;
+            focal.vec[1] = coords.vec[1] - coords.vec[2] * (1.0f - precision) * dy / dist;
```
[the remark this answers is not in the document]
hermet: @mgrudzinska Please confirm this. Thanks.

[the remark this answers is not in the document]
mgrudzinska: @hermet this is now handled in https://github.com/thorvg/thorvg/pull/3568

PR #3552, opened 2025-06-18
loader/jpg: Fix uninitialized warning

JSUYA:
```
../src/loaders/jpg/tvgJpgd.cpp:670:10: note: ‘padding_flag’ was declared here
  670 |     bool padding_flag;
      |          ^~~~~~~~~~~~
../src/loaders/jpg/tvgJpgd.cpp:674:9: warning: ‘padding_flag’ may be used uninitialized [-Wmaybe-uninitialized]
  674 |         if (padding_flag) return 0xFF;
      |         ^~
```


src/loaders/jpg/tvgJpgd.cpp
```diff
@@ -667,7 +667,7 @@ inline void jpeg_decoder::stuff_char(uint8_t q)
 // Retrieves one character from the input stream, but does not read past markers. Will continue to return 0xFF when a marker is encountered.
 inline uint8_t jpeg_decoder::get_octet()
 {
-    bool padding_flag;
+    bool padding_flag = false;
```
hermet: I think there are 2 options:
1. just to shutdown, TVG_UNUSED may work for this
2. It's better to fix inside get_char().   see: 668 line.

JSUYA: > 2\. It's better to fix inside get_char().   see: 668 line.

Fixed, thanks!

hermet:
https://github.com/thorvg/thorvg/commit/d5416de8ac7ca9dc9de001362df4e1271341a107

PR #3555, opened 2025-06-18
gl_engine: remove the outdated tesseletion mechanics

SergeyLebedkin:
After introducing **Triangle Fan-like Triangulation** (BWTessellator) there is no necessary to have heavy and complex version of general tessellator and now its unused
All unused entities are removed


hermet:
@SergeyLebedkin Please check sort() functions in `tvgArray.h`, probably, this become dead code as well.

image

SergeyLebedkin:
@hermet 
> @SergeyLebedkin Please check sort() functions in `tvgArray.h`, probably, this become dead code as well.

removed

PR #3556, opened 2025-06-18
lottie: add lineTo support in round corner alg

mgrudzinska:
So far, there hasn't been a case where this support was necessary, but it will be once chaining is added- when the LottieRoundnessModifier::modifyPath() is called by a previous modifier.


hermet:
@mgrudzinska do we have any test sample?

mgrudzinska:
> @mgrudzinska do we have any test sample?

as described [here](https://github.com/thorvg/thorvg/pull/3556#issue-3157751897), for now not possible to test. only together with modifiers chaining

PR #3560, opened 2025-06-19
lottie: ensure proper shape closure

mgrudzinska:
Ensured proper closure of star and polygon shapes. The start and end points now match - in cases with degenerate bezier curves, the second-to-last point is also aligned.
Proper shape closure is necessary for modifiers like offset (future pucker bloat). If the start and end points aren’t equal (within the comparison function’s precision), the shape will be closed with a straight line - and during offsetting, that line will become visible, even though it’s not intended.


hermet:
https://github.com/thorvg/thorvg/commit/2679880bc3c967d438da1f2bea62ee117ed77c58

PR #3564, opened 2025-06-20
build: Use Meson-provided summaries

QuLogic:
Meson has provided a summary function since 0.53.0, so there's no need to generate a summary message manually. The Meson function also has support for separate sections and styled output for booleans.

It now looks like:
Screenshot From 2025-06-20 18-02-03

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -166,59 +166,73 @@ if get_option('tests')
    subdir('test')
 endif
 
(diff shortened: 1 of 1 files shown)


hermet (approved):

hermet:
@QuLogic thanks!

PR #3565, opened 2025-06-22
build: Export TVG_STATIC to parent projects

QuLogic:
On Windows, you need to consistently set `__declspec` on symbols. This is handled automatically in `thorvg.h` for shared libraries, and left out entirely for static libraries by the `TVG_STATIC` define.

When building as a subproject with static libraries, it is thus useful to export `TVG_STATIC` to the parent project, or else `thorvg.h` will attempt to set `__declspec` for a shared library (when the parent project uses it), and linkage will fail.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -74,6 +74,7 @@ thorvg_lib = library(
 )
 
 thorvg_dep = declare_dependency(
+  compile_args: (lib_type == 'static') ? ['-DTVG_STATIC'] : [],
   include_directories: thorvg_inc,
   link_with: thorvg_lib,
 )


src/meson.build
```diff
@@ -74,6 +74,7 @@ thorvg_lib = library(
 )
 
 thorvg_dep = declare_dependency(
+  compile_args: (lib_type == 'static') ? ['-DTVG_STATIC'] : [],
```
hermet, line 77: Thanks, You can't just use `compiler_flags` here? Just wanna be sure.

QuLogic: No, that includes all the build flags, like enabling AVX, etc., which the user doesn't need to copy.

QuLogic:
Just wondering what happened to this commit? GitHub says it's merged, but doesn't show the merge commit as being in any branches or tags, and the file on `main` does not have this change.

hermet:
@QuLogic Hi. pushed in main. https://github.com/thorvg/thorvg/commit/5c99fc507148c0dd191ca789e49bd7ec6c84f779

I guess there were something mistake during merging process. sorry for that.

PR #3568, opened 2025-06-23
renderer: improve radial gradient handling

mgrudzinska:
This change ensures at the api level that if the focal point lies outside the end circle, it is projected onto the edge of the end circle.
Additionally, if the start circle does not fully fit inside the end circle (after possible repositioning), its radius is reduced accordingly.

The modification aligns with the SVG 1.1 standard (for fr = 0).
 Cases with fr > 0 are not covered by the SVG 1.1, and edge
cases have been handled here to avoid numerical issues.

Note:
This update replaces previous behavior where gl handled the SVG 2.0 standard.

samples:
[Archiwum.zip](https://github.com/user-attachments/files/20866633/Archiwum.zip)

before:
Zrzut ekranu 2025-06-23 o 15 58 18

after:
Zrzut ekranu 2025-06-23 o 16 00 38

this also handles this:
https://github.com/thorvg/thorvg/pull/3563#issue-3162822909


SergeyLebedkin (approved):

src/renderer/tvgFill.h
```diff
@@ -128,7 +128,41 @@ struct RadialGradientImpl : RadialGradient
         if (fy) *fy = this->fy;
         if (fr) *fr = this->fr;
 
-        return Result::Success;
+        //TODO: remove the "internal" logic once SVG 2.0 is adopted by sw and wg engines (gl already supports it); lottie-specific handling will then be delegated entirely to the loader
+        if (!internal) return true;
+
+        constexpr float precision = 0.01f;
+
+        //solid fill - handled internaly in engines
```
hermet: Please use like this, Renderer guides a spec to engine. It has no idea about each engine.
`//a solid fill case. It can be handled by engine.`

src/renderer/tvgFill.h
```diff
@@ -119,7 +119,7 @@ struct RadialGradientImpl : RadialGradient
         return Result::Success;
     }
 
-    Result radial(float* cx, float* cy, float* r, float* fx, float* fy, float* fr) const
+    bool radial(float* cx, float* cy, float* r, float* fx, float* fy, float* fr, bool internal = false) const
```
hermet: Suggest to separate functions:
```
//Unusual case, Please add a description explaining what it means for the focal point to be 'correct'.
bool correct(fx, fy, fr) const;
{
   ...
}
```
Call the function by engines directly.
```
fill->radial(x, y, r, fx, fy, fr);
fill->correct(fx, fy, fr);
```

hermet:
@mgrudzinska the above before/after figures are correct? It looks reversed?

mgrudzinska:
> @mgrudzinska the above before/after figures are correct? It looks reversed?

ofc not, sorry. fixed

src/renderer/tvgFill.h
mgrudzinska: please check now. I added a short description, detailed is inside the fun to describe separately the cases

PR #3569, opened 2025-06-24
wasm: align wasm-opt binary with build EMSDK version

tinyjin:
Replace global wasm-opt with EMSDK's bundled version to ensure toolchain consistency and avoid version mismatches.


hermet:
Code review is not free. Please be more careful when submitting patches. Thanks.

PR #3573, opened 2025-06-24
lottie: ensuring rectangle closure

mgrudzinska:
A rectangle with roundness = 0 is the only shape where one of its sides is created using a close command. This exceptional case requires additional handling of the close command for every modifier. By adding an extra lineTo command to the rectangle path, its becomes consistent with other shapes, eliminating the special case inside modifiers algorithms.

Note: this change does not produce any visible effects.

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -392,12 +392,29 @@ static void _repeat(LottieGroup* parent, Shape* path, RenderContext* ctx)
 }
 
 
+static void _close(RenderPath& path, const Point& p, bool round, bool rect = false)
+{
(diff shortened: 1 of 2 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -392,12 +392,29 @@ static void _repeat(LottieGroup* parent, Shape* path, RenderContext* ctx)
 }
 
 
+static void _close(RenderPath& path, const Point& p, bool round, bool rect = false)
```
hermet, line 395: Rectangle itself doesn't need this logic. This kind of patchwork solution looks bad.

hermet (changes requested):

PR #3579, opened 2025-06-26
common: generalize bezier math

SergeyLebedkin:
added or moved new functionality for bezier curves:
1. generate bezier curve as an arc
2. estimate the number of points for optimal tessellation
3. check that bezier curve is close to a straight line 
This functions is useful for gl/wg renderers and are general for math lib atself

### DIFF
--- a/src/common/tvgMath.cpp
+++ b/src/common/tvgMath.cpp
@@ -293,6 +293,21 @@ void Line::split(float at, Line& left, Line& right) const
 }
 
 
+Bezier::Bezier(const Point& st, const Point& ed, float radius)
+{
+    // Calculate the angle between the start and end points
+    auto angle = tvg::atan2(ed.y - st.y, ed.x - st.x);
+
+    // Calculate the control points of the cubic bezier curve
+    auto c = radius * PATH_KAPPA;  // c = radius * (4/3) * tan(pi/8)
+ 
+    start = {st.x, st.y};
+    ctrl1 = {st.x + radius * cos(angle), st.y + radius * sin(angle)};
+    ctrl2 = {ed.x - c * cos(angle), ed.y - c * sin(angle)};
+    end = {ed.x, ed.y};
+}
+
+
 void Bezier::split(Bezier& left, Bezier& right) const
 {
     auto c = (ctrl1.x + ctrl2.x) * 0.5f;
@@ -447,6 +462,32 @@ void Bezier::bounds(Point& min, Point& max) const
 
     findMinMax(start.x, ctrl1.x, ctrl2.x, end.x, min.x, max.x);
     findMinMax(start.y, ctrl1.y, ctrl2.y, end.y, min.y, max.y);
+}
+
+bool Bezier::flatten() const
+{
+    float diff1_x = fabsf((ctrl1.x * 3.f) - (start.x * 2.f) - end.x);
+    float diff1_y = fabsf((ctrl1.y * 3.f) - (start.y * 2.f) - end.y);
+    float diff2_x = fabsf((ctrl2.x * 3.f) - (end.x * 2.f) - start.x);
+    float diff2_y = fabsf((ctrl2.y * 3.f) - (end.y * 2.f) - start.y);
+    if (diff1_x < diff2_x) diff1_x = diff2_x;
+    if (diff1_y < diff2_y) diff1_y = diff2_y;
+    return (diff1_x + diff1_y <= 0.5f);
+}
+
+
+uint32_t Bezier::segments() const
+{
+    if (flatten()) return 1;
+    Bezier left, right;
+    split(left, right);
+    return left.segments() + right.segments();
+}
+
+
+Bezier Bezier::operator*(const Matrix& m)
+{
+    return Bezier{start * m, ctrl1* m, ctrl2 * m, end * m};
 }
 
 }
--- a/src/common/tvgMath.h
+++ b/src/common/tvgMath.h
@@ -121,6 +121,12 @@ static inline constexpr const Matrix identity()
 }
 
 
+static inline float scaling(const Matrix& m)
+{
+    return sqrtf(m.e11 * m.e11 + m.e21 * m.e21);
+}
+
+
 static inline void scale(Matrix* m, const Point& p)
 {
     m->e11 *= p.x;
@@ -330,6 +336,21 @@ static inline Point operator-(const Point& a)
     return {-a.x, -a.y};
 }
 
+enum class Orientation
+{
+    Linear = 0,
+    Clockwise,
+    CounterClockwise,
+};
+
+
+static inline Orientation orientation(const Point& p1, const Point& p2, const Point& p3)
+{
+    auto val = cross(p2 - p1, p3 - p1);
+    if (zero(val)) return Orientation::Linear;
+    else return val > 0 ? Orientation::Clockwise : Orientation::CounterClockwise;
+}
+
 
 static inline void log(const Point& pt)
 {
@@ -362,6 +383,13 @@ struct Bezier
     Point ctrl2;
     Point end;
 
+    Bezier() {}
+    Bezier(const Point& p0, const Point& p1, const Point& p2, const Point& p3):
+        start(p0), ctrl1(p1), ctrl2(p2), end(p3) {}
+    // Constructor that approximates a quarter-circle segment of arc between 'start' and 'end' points 
+    // using a cubic Bezier curve with a given 'radius'.
+    Bezier(const Point& start, const Point& end, float radius);
+
     void split(float t, Bezier& left);
     void split(Bezier& left, Bezier& right) const;
     void split(float at, Bezier& left, Bezier& right) const;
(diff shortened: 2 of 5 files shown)


src/common/tvgMath.cpp
```diff
@@ -449,5 +449,26 @@ void Bezier::bounds(Point& min, Point& max) const
     findMinMax(start.y, ctrl1.y, ctrl2.y, end.y, min.y, max.y);
 }
 
+bool Bezier::isFlatten() const
```
hermet: -> `bool flatten() const`

src/common/tvgMath.cpp
```diff
@@ -449,5 +449,26 @@ void Bezier::bounds(Point& min, Point& max) const
...
+{
+    float diff1_x = fabs((ctrl1.x * 3.f) - (start.x * 2.f) - end.x);
```
hermet: `float` -> `auto`
`fabs` -> `fabsf`

src/common/tvgMath.cpp
```diff
@@ -449,5 +449,26 @@ void Bezier::bounds(Point& min, Point& max) const
...
+    float diff1_y = fabs((ctrl1.y * 3.f) - (start.y * 2.f) - end.y);
+    float diff2_x = fabs((ctrl2.x * 3.f) - (end.x * 2.f) - start.x);
+    float diff2_y = fabs((ctrl2.y * 3.f) - (end.y * 2.f) - start.y);
+    if (diff1_x < diff2_x) diff1_x = diff2_x;
+    if (diff1_y < diff2_y) diff1_y = diff2_y;
+    return (diff1_x + diff1_y <= 0.5f);
+}
+
+
+uint32_t Bezier::curveCount() const
```
hermet: suggestion: `segmentCnt()`

src/common/tvgMath.h
```diff
@@ -372,9 +393,28 @@ struct Bezier
     Point at(float t) const;
     float angle(float t) const;
     void bounds(Point& min, Point& max) const;
+    bool isFlatten() const;
+    uint32_t curveCount() const;
 };
 
 
+static inline Bezier bezierFromArc(const Point& start, const Point& end, float radius)
```
hermet: I think you can add this in constructor:
```
// Constructor that approximates an arc between 'start' and 'end' points 
// using a cubic Bezier curve with a given 'radius'. */
Bezier::Bezier(const Point& start, const Point& end, float radius) {...}

File: src/common/tvgMath.h [hunk 5]
```diff
@@ -330,6 +336,21 @@ static inline Point operator-(const Point& a)
     return {-a.x, -a.y};
 }
 
+enum class Orientation
+{
+    Linear,
```
hermet: `    Linear = 0,`

src/common/tvgMath.cpp
SergeyLebedkin: Just for my understanding.
What is the porpose to use fabsf instead of abs or std::abs?

src/common/tvgMath.h
```diff
@@ -330,6 +336,21 @@ static inline Point operator-(const Point& a)
...
 
+enum class Orientation
+{
+    Linear = 0,
+    Clockwise,
+    CounterClockwise,
+};
+
+
+static inline Orientation orientation(const Point& p1, const Point& p2, const Point& p3)
```
mgrudzinska, line 347: maybe we could have more general fun, so that it can replace the one used in modifiers (tvgLottieModifier.cpp):
```

static bool _clockwise(Point* pts, uint32_t n)
{
    auto area = 0.0f;

    for (uint32_t i = 0; i < n - 1; i++) {
        area += cross(pts[i], pts[i + 1]);
    }
    area += cross(pts[n - 1], pts[0]);;

    return area < 0.0f;
}
```

src/common/tvgMath.h
```diff
@@ -330,6 +336,21 @@ static inline Point operator-(const Point& a)
...
+{
+    auto val = cross(p2 - p1, p3 - p1);
+    if (std::abs(val) < 0.0001f) return Orientation::Linear;
```
mgrudzinska: tvg::zero ?

src/common/tvgMath.cpp
```diff
@@ -449,5 +464,26 @@ void Bezier::bounds(Point& min, Point& max) const
...
+    float diff1_y = fabsf((ctrl1.y * 3.f) - (start.y * 2.f) - end.y);
+    float diff2_x = fabsf((ctrl2.x * 3.f) - (end.x * 2.f) - start.x);
+    float diff2_y = fabsf((ctrl2.y * 3.f) - (end.y * 2.f) - start.y);
+    if (diff1_x < diff2_x) diff1_x = diff2_x;
+    if (diff1_y < diff2_y) diff1_y = diff2_y;
+    return (diff1_x + diff1_y <= 0.5f);
+}
+
+
+uint32_t Bezier::segmentCnt() const
```
mgrudzinska: I'd add a depth counter. 
this is a strange case, but still should be handled and now it causes stack-overflow:
```
    Bezier b{{-87206376.000000, -176699552.000000}, {-359726240.000000, -728884992.000000}, {-14.693240, -61.985207}, {-0.078390, -42.229122}};
    auto cnt = b.segmentCnt();
```
or alternatively flatten() can be enhanced to handle such strange cases?

src/common/tvgMath.cpp
```diff
@@ -293,6 +293,21 @@ void Line::split(float at, Line& left, Line& right) const
 }
 
 
+Bezier::Bezier(const Point& st, const Point& ed, float radius)
+{
+    // Calculate the angle between the start and end points
+    auto angle = tvg::atan2(ed.y - st.y, ed.x - st.x);
+
+    // Calculate the control points of the cubic bezier curve
+    auto c = radius * 0.552284749831f;  // c = radius * (4/3) * tan(pi/8)
```
mgrudzinska: already defined as PATH_KAPPA in tvgMath.h

src/common/tvgMath.h
```diff
@@ -362,6 +383,13 @@ struct Bezier
     Point ctrl2;
     Point end;
 
+    Bezier() {};
+    Bezier(const Point& p0, const Point& p1, const Point& p2, const Point& p3):
+        start(p0), ctrl1(p1), ctrl2(p2), end(p3) {};
+    // Constructor that approximates an arc between 'start' and 'end' points 
+    // using a cubic Bezier curve with a given 'radius'. */
+    Bezier(const Point& start, const Point& end, float radius);
```
mgrudzinska, line 391: this is a very specific case, creating a quarter-circle segment. I'd make it a separate fun instead of a constructor. 
ofc feel free to ignore this comment

SergeyLebedkin: I just modify the comment in the code and notice about this specific

src/common/tvgMath.h
SergeyLebedkin: we need to know also about co-linear cases, but not just about direction

mgrudzinska:
```
enum class Orientation
{
    Linear = 0,
    Clockwise,
    CounterClockwise,
};
```
this is not about the direction? in the fun I copy pasted from modifiers we do not handle 'linear', but this requires to add only one line, right?
```if (zero(val)) return Orientation::Linear;```

File: src/common/tvgMath.h [hunk 11]
```diff
@@ -372,8 +400,15 @@ struct Bezier
     Point at(float t) const;
     float angle(float t) const;
     void bounds(Point& min, Point& max) const;
+    bool flatten() const;
+    uint32_t segmentCnt() const;
 };
 
+static inline Bezier operator*(const Bezier& bez, const Matrix& m)
```
hermet: we can put this in Bezier struct:
```
    Bezier operator*(const Matrix& m)
    {
        return Bezier{start * m, ctrl1* m, ctrl2 * m, end * m};
    }    
```

src/common/tvgMath.h
```diff
@@ -362,6 +383,13 @@ struct Bezier
     Point ctrl2;
     Point end;
 
+    Bezier() {};
```
hermet: `;` not necessary

src/common/tvgMath.h
```diff
@@ -362,6 +383,13 @@ struct Bezier
...
+    Bezier(const Point& p0, const Point& p1, const Point& p2, const Point& p3):
+        start(p0), ctrl1(p1), ctrl2(p2), end(p3) {};
```
hermet: `;` not necessary

src/common/tvgMath.h
```diff
@@ -372,8 +400,15 @@ struct Bezier
     Point at(float t) const;
     float angle(float t) const;
     void bounds(Point& min, Point& max) const;
+    bool flatten() const;
+    uint32_t segmentCnt() const;
```
hermet: sorry for this, let's change the name to `segments()`

src/common/tvgMath.cpp
```diff
@@ -449,5 +464,26 @@ void Bezier::bounds(Point& min, Point& max) const
...
+    if (diff1_x < diff2_x) diff1_x = diff2_x;
+    if (diff1_y < diff2_y) diff1_y = diff2_y;
+    return (diff1_x + diff1_y <= 0.5f);
+}
+
+
+uint32_t Bezier::segmentCnt() const
+{
+    if (flatten()) return 1;
+    Bezier left{}, right{};
```
hermet: I think this doesn't need zero-initialization?
-> `Bezier left, right;`

hermet (approved):

Issue #3557, opened 2025-06-18
wg_engine: tesseletion optimization


SergeyLebedkin:
this ideas based on gl_engine tessetation mechanics
1. **do not split geometry** into separate buffers in a case of moveTo command, use single vertex buffer for shape
2. **directly fill vertex/index buffers** for shape without static mediator buffers
3. apply **stroke dashes for path** but not for decoded outline
4. use **more precise tesseletion** factor for cubic splines/circles
5. get rid common tri-fun index bufffer
6. draw shape using **single draw call**, but not for each sub-mesh (moveTo)

PR #3581, opened 2025-06-28
common: share stroke dasher

SergeyLebedkin:
Move stroke dasher to the common code space to create an abillity to use it on the cross API renderers (wg and gl).
Stroke dasher is a path-to-path operation, same as path trim, so can be placed to the common space.

issue: https://github.com/thorvg/thorvg/issues/3557

### DIFF
--- a/src/renderer/gl_engine/tvgGlGeometry.cpp
+++ b/src/renderer/gl_engine/tvgGlGeometry.cpp
@@ -28,18 +28,16 @@
 
 bool GlGeometry::tesselate(const RenderShape& rshape, RenderUpdateFlag flag)
 {
-    const RenderPath* path = nullptr;
-    RenderPath trimmedPath;
-    if (rshape.trimpath()) {
-        if (!rshape.stroke->trim.trim(rshape.path, trimmedPath)) return true;
-        path = &trimmedPath;
-    } else path = &rshape.path;
-
     if (flag & (RenderUpdateFlag::Color | RenderUpdateFlag::Gradient | RenderUpdateFlag::Transform | RenderUpdateFlag::Path)) {
         fill.clear();
 
         BWTessellator bwTess{&fill};
-        bwTess.tessellate(*path, matrix);
+        if (rshape.trimpath()) {
+            RenderPath trimmedPath;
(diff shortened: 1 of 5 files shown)


hermet:
`tvgRender` has been introduced as an interface similar to HAL or RHI, but This one is more of a helper function. In fact, a single function like `bool genDashPath(const RenderPath& in, RenderPath& out);` might be sufficient for this purpose. The question is: where would be the most appropriate place to include it...

hermet:
Additionally, perhaps we can also replace the software ver. with this as well.

SergeyLebedkin:
> `tvgRender` has been introduced as an interface similar to HAL or RHI, but This one is more of a helper function. In fact, a single function like `bool genDashPath(const RenderPath& in, RenderPath& out);` might be sufficient for this purpose. The question is: where would be the most appropriate place to include it...

thats a good point and good question.
we can hide inplementation in render.cpp file and make a function as a public interface like `bool genDashPath(const RenderPath& in, RenderPath& out, ...);`

SergeyLebedkin:
> Additionally, perhaps we can also replace the software ver. with this as well.

As I can see dashers for gl and sw is mostly similar. The difference is in output format.
gl version: path-to-path conversion (RenderPath to RenderPath)
sw version: path-to-outline conversion (RenderPath to SwOutline)wg
wg version (current): outline-to-outline conversion (WgVertexBuffer to WgVertexBuffer, will be get rided)

hermet:
@SergeyLebedkin Please consider this interface:

`bool RenderShape::strokeDash(RenderPath& out) const`

SergeyLebedkin:
> @SergeyLebedkin Please consider this interface:
> 
> `bool RenderShape::strokeDash(RenderPath& out) const`

whats do you mean consider? this dasher uses all settings for dashing

hermet:
@SergeyLebedkin I suggest defining a common internal interface (`strokeDash()`) for the Dash Stroker. The `RenderPath& out` parameter could be used to return the dash path data.

SergeyLebedkin:
OK. So now RenderShape can generate dashed path by himself.
Dashing mechanics fully hiden from the user but can be used in gl and wg renderers

hermet:
duplicated: https://github.com/thorvg/thorvg/pull/3624

Issue #1944, opened 2024-01-30
Lottie: blending doesn work properly.


hermet:
[blending_sample.json](https://github.com/thorvg/thorvg/files/14096264/blending_sample.json)

hermet:
The reason is the blending is performed on the zero canvas space (0x00000000)

baszczewski:
I wanted to ask to what extent ThorVG supports blending. Specifically, I'm interested in blend multiply, which, unfortunately, doesn't seem to work in any of the rendering engines.

Sample file: 

![Image](https://github.com/user-attachments/assets/9b544e60-66f7-41ff-be2f-edc84b572185)

ThorVG viewer (v1.0.0.pre17): 

![Image](https://github.com/user-attachments/assets/9c8ebc97-5dfb-427a-8cf4-01388bfa6739)

hermet:
> I wanted to ask to what extent ThorVG supports blending. Specifically, I'm interested in blend multiply, which, unfortunately, doesn't seem to work in any of the rendering engines.
> 
> Sample (SVG):
> 
> ![Image](https://github.com/user-attachments/assets/9b544e60-66f7-41ff-be2f-edc84b572185)
> 
> Sample (LOTTIE):
> 
> [wired-flat-20-love-heart-hover-heartbeat.json](https://github.com/user-attachments/files/19416015/wired-flat-20-love-heart-hover-heartbeat.json)
> 
> ThorVG viewer (v1.0.0.pre17):
> 
> ![Image](https://github.com/user-attachments/assets/9c8ebc97-5dfb-427a-8cf4-01388bfa6739)

@baszczewski  The fix will be in v0.15.14 & v1.0-pre23 thanks.

PR #3583, opened 2025-07-02
lottie: support group-level blending

hermet:
previously, only layer-level blending was supported. Now, blending has been extended to support group-level as well.


hermet:
issue: https://github.com/thorvg/thorvg/issues/1944

PR #3585, opened 2025-07-02
lottie: prevent mem leak

mgrudzinska:
Masks of pooled shapes were not unrefed, filling up the memory pool.

before:
Zrzut ekranu 2025-07-2 o 14 19 35
Zrzut ekranu 2025-07-2 o 14 24 53
Zrzut ekranu 2025-07-2 o 14 29 22

after:
Zrzut ekranu 2025-07-2 o 14 31 37

sample:
[mask.json](https://github.com/user-attachments/files/21018098/mask.json)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1204,6 +1204,7 @@ void LottieBuilder::updateMasks(LottieLayer* layer, float frameNo)
         if (!pShape) {
             pShape = layer->pooling();
             pShape->reset();
+            PAINT(pShape)->resetMaskClip();
```
hermet: I think you can just use `SHAPE(pShape)->reset();`
with removing 1225, 1226 lines.

mgrudzinska: please check now, ```pShape->reset();``` replaced with ```SHAPE(pShape)->reset();```

PR #3610, opened 2025-07-09
lottie: fix layer update

mgrudzinska:
If a layer’s inPoint and/or outPoint differs from
the overall animation range, it must continue to be updated past those points so that it correctly becomes invisible. Otherwise, the last updated frame remains stuck on screen.
Likewise, when a layer’s opacity suddenly drops to zero, it should still be redrawn as fully transparent.

samples:
[op.json](https://github.com/user-attachments/files/21143067/op.json)
[update.json](https://github.com/user-attachments/files/21143069/update.json)

before:
before

after:
after


hermet:
@mgrudzinska I can't reproduce this with the latest main or the ThorVG viewer. It seems to have been a temporary issue wrt partial rendering during main dev. and doesn't appear to be related to Lottie itself. Please double-check. Thanks.

PR #3614, opened 2025-07-12
sw_engine: rectifiy logic wrt blendings

hermet:
- minor optimization of the unpremultiply logic
- exception handling for the unpremultipy logic
- appropriate alpha pre/unpre multiplication is applied
- code clean++

issue: https://github.com/thorvg/thorvg/issues/1944


SergeyLebedkin:
As I caa see the resuts must be the same?
I mean we do not need to chnage gl/wg shaders for this cond of blendings?

hermet:
> As I caa see the resuts must be the same? I mean we do not need to chnage gl/wg shaders for this cond of blendings?

@SergeyLebedkin yes, need to change them & align the same result.

Issue #3615, opened 2025-07-13
Not rendering complex SVG's


Neodevils:
Hello, I am a Godot user and Godot uses ThorVG as far as I know. 
_I don't know this is duplicated. I have searched for use keyword but couldn't find any._

> ### Issue description
> 
> So far, I am trying to use extracted SVG files from a SWF file with using FFDec Flash Decompiler. It exports SVG files with `<use>` along with `defs` & `g` elements.
> 
> ### Example
> ```xml
> 
> <use ffdec:characterId="797" ffdec:characterName="WholeSword" height="79.45" transform="matrix(0.9871, -0.1571, 0.1571, 0.9871, -95.6324, -201.0661)" width="132.65" xlink:href="#sprite25"/>
> 
> ```
> 
> When I import this to Godot, it just shows a blank screen for SVG file. I have tried to SVG to SVG converter to escape from these elements but now it returns the file just as `<path>` which is making the dedicated sprite name unable to detect _since all become path_.
> 
> 1. Upload the file to Godot.
> 2. You can't see it is output.
> ---
> 1. Open it on any kind of browser.
> 2. It is being displayed on browser.
> 
> ![Image](https://github.com/user-attachments/assets/1e831e81-6416-4a35-a414-ea00ea833507)
>

hermet:
@Neodevils Hello, the fix will be in v0.15.14. Thanks.

PR #3616, opened 2025-07-14
svg: fix nested use nodes

mgrudzinska:
Previously, only single-level <use> references were supported. If a <use> node pointed to another element that itself contained a <use> node, the reference wasn’t resolved.
This has been fixed by replacing the array of postponed elements with a list. The list is traversed, and nodes aren’t cloned while they or any of their children remain unresolved. In such cases, the target element also gets added to the list, enabling recursive resolution of nested href references.

@Issue: https://github.com/thorvg/thorvg/issues/3615

samples:
[Archiwum.zip](https://github.com/user-attachments/files/21222689/Archiwum.zip)

before:
Zrzut ekranu 2025-07-14 o 23 29 10

after:
Zrzut ekranu 2025-07-14 o 23 28 52

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -894,12 +894,6 @@ error:
     }
 
 
-static void _postpone(Array<SvgNodeIdPair>& nodes, SvgNode *node, char* id)
-{
-    nodes.push({node, id});
-}
-
-
 static bool _attrParseSvgNode(void* data, const char* key, const char* value)
 {
     SvgLoaderData* loader = (SvgLoaderData*)data;
@@ -1157,7 +1151,7 @@ static void _handleCssClassAttr(SvgLoaderData* loader, SvgNode* node, const char
         cssCopyStyleAttr(node, cssNode);
     }
 
-    if (!cssClassFound) _postpone(loader->nodesToStyle, node, *cssClass);
+    if (!cssClassFound) loader->nodesToStyle.push({node, *cssClass});
 }
 
 
@@ -2155,6 +2149,23 @@ static SvgNode* _findParentById(SvgNode* node, char* id, SvgNode* doc)
 }
 
 
+static bool _checkPostponed(SvgNode* node, SvgNode* cloneNode, int depth)
+{
+    if (node == cloneNode) return true;
+
+    if (depth == 512) {
+        TVGERR("SVG", "Infinite recursive call - stopped after %d calls! Svg file may be incorrectly formatted.", depth);
+        return false;
+    }
+
+    ARRAY_FOREACH(p, node->child) {
+        if (_checkPostponed(*p, cloneNode, depth + 1)) return true;
+    }
+
+    return false;
+}
+
+
 static constexpr struct
 {
     const char* tag;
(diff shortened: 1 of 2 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2155,6 +2164,22 @@ static SvgNode* _findParentById(SvgNode* node, char* id, SvgNode* doc)
 }
 
 
+static bool _checkPostponed(SvgNode* node, SvgNode* cloneNode, int depth)
+{
+    if (depth == 8192) {
```
hermet: A recursion depth of 8192 seems excessively large.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -2155,6 +2164,22 @@ static SvgNode* _findParentById(SvgNode* node, char* id, SvgNode* doc)
...
+        TVGERR("SVG", "Infinite recursive call - stopped after %d calls! Svg file may be incorrectly formatted.", depth);
+        return false;
+    }
+
+    if (node == cloneNode) return true;
```
hermet: This success condition needs to be prioritized in the function.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -900,6 +900,15 @@ static void _postpone(Array<SvgNodeIdPair>& nodes, SvgNode *node, char* id)
 }
 
 
+static void _postpone(Inlist<SvgNodeIdPair>& nodes, SvgNode *node, char* id)
```
hermet: There’s no point in separating the function. `_postpone()` is neither self-explanatory nor reusable, and it adds unnecessary complexity with this change.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3356,22 +3394,39 @@ static void _cloneNode(SvgNode* from, SvgNode* parent, int depth)
...
+    while (nodeIdPair) {
+        auto defs = _getDefsNode(nodeIdPair->node);
+        auto nodeFrom = _findNodeById(defs, nodeIdPair->id);
+        if (!nodeFrom) nodeFrom = _findNodeById(doc, nodeIdPair->id);
+
+        if (!_findParentById(nodeIdPair->node, nodeIdPair->id, doc)) {
+            //Check if none of nodeFrom's children are in the cloneNodes list
+            auto postpone = false;
+            INLIST_FOREACH((*cloneNodes), pair) {
+                if (_checkPostponed(nodeFrom, pair->node, 1)) {
```
JSUYA: There are scenarios where `nodeFrom` can be null, please see the example below.

```html
<svg xmlns:xlink="http://www.w3.org/1999/xlink" height="729.7" width="639.65" xmlns="http://www.w3.org/2000/svg">
  <defs>

    <use id="sprite0" height="5.15"  width="4.45" href="#shape0"/>

    <use id="sprite0" height="5.15"  width="4.45" href="#invalidUrl"/>

    <path id="shape0" d="M0 0 h 200 v 200 " fill="#ff00ff" />

  </defs>

  <use height="36.05"  width="31.15" href="#sprite0"/>

</svg>

```

how about this?
```cpp
if (nodeFrom && _checkPostponed(nodeFrom, pair->node, 1)) {
```

src/loaders/svg/tvgSvgLoader.cpp
mgrudzinska: copied from other function in tvgSvgLoader. 
I changed it to 512

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3356,22 +3394,39 @@ static void _cloneNode(SvgNode* from, SvgNode* parent, int depth)
...
+                    postpone = true;
+                    cloneNodes->back(nodeIdPair);
```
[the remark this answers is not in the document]
mgrudzinska: nodeIdPair = cloneNodes->front(); removes the node

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3356,22 +3380,41 @@ static void _cloneNode(SvgNode* from, SvgNode* parent, int depth)
...
-            _cloneNode(nodeFrom, nodeIdPair.node, 0);
-            if (nodeFrom && nodeFrom->type == SvgNodeType::Symbol && nodeIdPair.node->type == SvgNodeType::Use) {
-                nodeIdPair.node->node.use.symbol = nodeFrom;
+    auto nodeIdPair = cloneNodes->front();
+    while (nodeIdPair) {
+        auto defs = _getDefsNode(nodeIdPair->node);
+        auto nodeFrom = _findNodeById(defs, nodeIdPair->id);
+        if (!nodeFrom) nodeFrom = _findNodeById(doc, nodeIdPair->id);
+
+        if (!_findParentById(nodeIdPair->node, nodeIdPair->id, doc)) {
```
JSUYA: It seems more efficient to have the code that gets nodeFrom inside the if-condition.

```suggestion
        if (!_findParentById(nodeIdPair->node, nodeIdPair->id, doc)) {
            auto defs = _getDefsNode(nodeIdPair->node);
            auto nodeFrom = _findNodeById(defs, nodeIdPair->id);
            if (!nodeFrom) nodeFrom = _findNodeById(doc, nodeIdPair->id);
```

JSUYA (approved):
LGTM

hermet (approved):

PR #3621, opened 2025-07-17
api: Add Composition blend mode

hermet:
The composition mode enables intermediate blending. When a scene
 uses Composition mode, it generates an intermediate buffer to
render the scene image first. This image can then be blended with the canvas afterward.

Please avoid to use this, unless you really need to composite the precomposite scene. This feature is relatively expensive at performance.

C++ API:
+BlendMethod::Composition

C API:
+Tvg_Blend_Method::TVG_BLEND_METHOD_COMPOSITION


hermet:
\+ @SergeyLebedkin gl/wgpu also need to support Composition

PR #3624, opened 2025-07-17
renderer: refactored to share gl stroke dasher among engines

SergeyLebedkin:
Move stroke dasher to the common code space to create an abillity to use it on the cross API renderers (wg and gl). 
Stroke dasher is a path-to-path operation, same as path trim, so can be placed to the common space.
Now RenderShape can generate dashed path by himself.
Dashing mechanics fully hiden from the user but can be used in gl and wg renderers.

issue: https://github.com/thorvg/thorvg/issues/3557


hermet:
binary size +5kb, even if sw engine only is built...

PR #3627, opened 2025-07-17
gl_engine: Fix GlRenderTarget::reset()

Nor-s:
Avoid deleting the framebuffer when `mFbo == GL_INVALID_VALUE`, as this is not a valid framebuffer object.

mFbo is initialized with GL_INVALID_VALUE as its default value.
- https://github.com/thorvg/thorvg/blob/e3e0ab8579792d67aa299ff3ee3f6a93c53610d7/src/renderer/gl_engine/tvgGlRenderTarget.h#L53
- https://github.com/thorvg/thorvg/blob/e3e0ab8579792d67aa299ff3ee3f6a93c53610d7/src/renderer/gl_engine/tvgGlRenderTarget.cpp#L87

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderTarget.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderTarget.cpp
@@ -77,7 +77,7 @@ void GlRenderTarget::init(uint32_t width, uint32_t height, GLint resolveId)
 
 void GlRenderTarget::reset()
 {
-    if (mFbo == 0) return;
+    if (mFbo == 0 || mFbo == GL_INVALID_VALUE) return;
     GL_CHECK(glBindFramebuffer(GL_FRAMEBUFFER, 0));
     GL_CHECK(glDeleteFramebuffers(1, &mFbo));
     GL_CHECK(glDeleteRenderbuffers(1, &mColorBuffer));


src/renderer/gl_engine/tvgGlRenderTarget.cpp
```diff
@@ -77,7 +77,7 @@ void GlRenderTarget::init(uint32_t width, uint32_t height, GLint resolveId)
 
 void GlRenderTarget::reset()
 {
-    if (mFbo == 0) return;
+    if (mFbo == 0 || mFbo == GL_INVALID_VALUE) return;
```
SergeyLebedkin, line 80: Official spec:

```
GL_INVALID_VALUE, 0x0501
Given when a value parameter is not a legal value for that function.
```
So its related to the func parameters but not for a handles

Nor-s: Thank you. I'm closing this PR.

hermet:
@SergeyLebedkin This patch looks valid exception. Please double-check again(Create GlCanvas -> destroy GlCanvas). It's initialized with GL_INVALID_VALUE.
image

SergeyLebedkin:
> @SergeyLebedkin This patch looks valid exception. Please double-check again(Create GlCanvas -> destroy GlCanvas). It's initialized with GL_INVALID_VALUE. 

No, it means, that we just use GL_INVALID_VALUE constans as a marker for invalid buffer handle.
The other question is why this constant? Just because of name? but its confusing

hermet:
@SergeyLebedkin Whether we use `GL_INVALID_VALUE` or another value is something we can decide. But it's important to prevent invalid values from being passed to the GL call. Please suggest if you have a clean fix. Thanks.

SergeyLebedkin:
opengl has no concept of invalid handle to buffer. However, we can introduce our own constant for this, but not in a scope of GL constants. F.e.

```
TVG_GL_INVALID_HANDLE = 0xFFFF
```

hermet:
@SergeyLebedkin What are the benefits of doing that compared to the current? and What would you do if FFFF is actually a valid value allowed by GL?

SergeyLebedkin:
> @SergeyLebedkin What are the benefits of doing that compared to the current? and What would you do if FFFF is actually a valid value allowed by GL?

Same question about GL_INVALID_VALUE 0x0501
The most relevant value for invalid handle of FBO is 0

hermet:
@SergeyLebedkin Please propose a proper fix in this engine and work toward improvement. Just referring to the OpenGL spec isn't very helpful.

PR #3632, opened 2025-07-20
gl_engine: fix FBO validation

SergeyLebedkin:
At the moment, we use the value GL_INVALID_VALUE to mark an invalid handle for FBO, which is not correct.
It is more correct to use the value 0, which signals that the buffer was not created. 
Canvas initialization processes have also been adjusted

See Lottie, MultiCanvas examples

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderTarget.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderTarget.cpp
@@ -31,7 +31,7 @@ GlRenderTarget::~GlRenderTarget()
 
 void GlRenderTarget::init(uint32_t width, uint32_t height, GLint resolveId)
 {
-    if (mFbo != GL_INVALID_VALUE || width == 0 || height == 0) return;
+    if (mFbo != 0 || width == 0 || height == 0) return;
     mWidth = width;
     mHeight = height;
 
@@ -84,7 +84,11 @@ void GlRenderTarget::reset()
     GL_CHECK(glDeleteRenderbuffers(1, &mDepthStencilBuffer));
     GL_CHECK(glDeleteFramebuffers(1, &mResolveFbo));
     GL_CHECK(glDeleteTextures(1, &mColorTex));
-    mFbo = GL_INVALID_VALUE;
+    mFbo = 0;
+    mColorBuffer = 0;
+    mDepthStencilBuffer = 0;
+    mResolveFbo = 0;
+    mColorTex = 0;
 }
 
 GlRenderTargetPool::GlRenderTargetPool(uint32_t maxWidth, uint32_t maxHeight): mMaxWidth(maxWidth), mMaxHeight(maxHeight), mPool() {}
--- a/src/renderer/gl_engine/tvgGlRenderTarget.h
+++ b/src/renderer/gl_engine/tvgGlRenderTarget.h
@@ -44,13 +44,13 @@ public:
     void setViewport(const RenderRegion& vp) { mViewport = vp; }
     const RenderRegion& getViewport() const { return mViewport; }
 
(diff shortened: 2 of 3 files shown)


SergeyLebedkin:
Altarnate to https://github.com/thorvg/thorvg/pull/3627

hermet:
Somehow, this has broken the original specification of the GL Canvas. Direct rendering is no longer valid. This needs to be reconsidered...

`* @param[in] id The GL target ID, usually indicating the FBO ID. A value of @c 0 specifies the main surface.`

```
    /**
     * @brief Sets the drawing target for rasterization.
     *
     * This function specifies the drawing target where the rasterization will occur. It can target
     * a specific framebuffer object (FBO) or the main surface.
     *
     * @param[in] context The GL context assigning to the current canvas rendering.
     * @param[in] id The GL target ID, usually indicating the FBO ID. A value of @c 0 specifies the main surface.
     * @param[in] w The width (in pixels) of the raster image.
     * @param[in] h The height (in pixels) of the raster image.
     * @param[in] cs Specifies how the pixel values should be interpreted. Currently, it only allows @c ColorSpace::ABGR8888S as @c GL_RGBA8.
     *
     * @retval Result::InsufficientCondition if the canvas is performing rendering. Please ensure the canvas is synced.
     * @retval Result::NonSupport In case the gl engine is not supported.
     *
     * @see Canvas::viewport()
     * @see Canvas::sync()
     *
     * @note Experimental API
    */
    Result target(void* context, int32_t id, uint32_t w, uint32_t h, ColorSpace cs) noexcept;
```

SergeyLebedkin:
> Somehow, this has broken the original specification of the GL Canvas. Direct rendering is no longer valid

No. Its did not break the currenе logic. In any case, the gl (and wg as well) renderer make rendering through an intermediate offscreen target that just blited to the providet external target in sync stage.
External target 0 - is still valid value as a direct native window surface. Intermidiate buffers are fully controled by as they are internal structures (not public), so it will not provide any handles conflicts.

hermet:
> In any case, the gl (and wg as well) renderer make rendering through an intermediate offscreen target that just blited to the providet external target in sync stage.
> External target 0 - is still valid value as a direct native window surface. Intermidiate buffers are fully controled by as they are internal structures (not public), so it will not provide any handles conflicts.

@SergeyLebedkin The FBO ID was originally provided by the user according to our API specification, but it is now entirely controlled by the engine. That’s a broken design. I'm not saying this PR caused the issue, but the current behavior and the spec are not aligned anymore. Time to clean it.

SergeyLebedkin:
> The FBO ID was originally provided by the user according to our API specification, but it is now entirely controlled by the engine

No, its not. We did not touch the originally provided ID. Its still controlled by user and valid for user.

SergeyLebedkin:
@Nor-s Thanks for bringing this up! It should have been resolved a long time ago! Thanks for reminding

hermet:
> No, its not. We did not touch the originally provided ID. Its still controlled by user and valid for user.

@SergeyLebedkin It is, the idea was to support direct drawing to the main surface without surface composition for performance when id is zero. I designed this and now it doesn't work like that...

SergeyLebedkin:
> @SergeyLebedkin It is, the idea was to support direct drawing to the main surface without surface composition for performance when id is zero. I designed this and now it doesn't work like that...

Its understandable, but it was changed a long time ago and not by this PR. 
We can back to review of this logic in some time, but its not related to this PR

hermet:
> I'm not saying this PR caused the issue, but the current behavior and the spec are not aligned anymore. Time to clean it.

@SergeyLebedkin What i said.

Issue #3099, opened 2025-01-06
lottie: Slot Optimization refactoring


hermet:
Currently, the overriding property is accessed through its object. This is not an ideal solution, as the target/source should be at the property level. We can revise the design and implementation to make it much flexible and simple logic.

![image](https://github.com/user-attachments/assets/4940814c-4470-49cf-9d7e-dd99c013b043)

See:
![image](https://github.com/user-attachments/assets/3575c113-058c-4b03-8754-22526e6a8d71)

hermet:
closed by https://github.com/thorvg/thorvg/pull/3323

PR #3634, opened 2025-07-21
lottie/slot: refactor override mechanism to property level

tinyjin:
CleanShot 2025-07-22 at 01 55 28@2x

Previously, the Slot overriding were accessed through their parent objects, which wasn't a best solution.

This change moves the target/source access to the property level directly, making the logic more flexible and simple.

issue: https://github.com/thorvg/thorvg/issues/3099


hermet:
@tinyjin could you please rebase this? thanks.

tinyjin:
@hermet I think it's already applied in https://github.com/thorvg/thorvg/pull/3323

Issue #3631, opened 2025-07-19
Duplicating SVG picture loses filter effects


lmdsp:
I very slightly modified the Duplicate.cpp example at lines 106 and 108 to load an SVG (`2684.svg`) with Gaussian blur:

```cpp
//Duplicate Picture - svg
{
  auto picture1 = tvg::Picture::gen();
  if (!tvgexam::verify(picture1->load(EXAMPLE_DIR"/svg/2684.svg"))) return false; // SVG with blur
  picture1->translate(350, 200);
  picture1->scale(4); // Changed scale

 auto picture2 = picture1->duplicate();
 picture2->translate(550, 250);
```

As can be seen on the screenshot below the blur is lost on the picture2 copy (right)

<img width="569" height="170" alt="Image" src="https://github.com/user-attachments/assets/9177993a-148a-4d25-99e4-32a0040c4608" />

I think this might be related to #3630

tinyjin:
<img width="998" height="154" alt="Image" src="https://github.com/user-attachments/assets/2e2f3b2c-c288-437f-8706-d232703f620c" />

This is because we don't support effects duplication currently. Need to support them.

hermet:
@lmdsp The fix will be in v0.15.15 & v1.0-pre25. thanks

hermet:
@lmdsp https://github.com/thorvg/thorvg/releases/tag/v0.15.15

PR #3638, opened 2025-07-23
renderer: support duplicate effects

tinyjin:
CleanShot 2025-07-23 at 17 19 17@2x

Implement effect duplication functionality for all RenderEffect types to enable proper Scene duplication with effects.

issue: https://github.com/thorvg/thorvg/issues/3631

### DIFF
--- a/src/renderer/tvgScene.h
+++ b/src/renderer/tvgScene.h
@@ -272,7 +272,40 @@ struct SceneImpl : Scene
             dup->paints.push_back(cdup);
         }
 
-        if (effects) TVGERR("RENDERER", "TODO: Duplicate Effects?");
+        if (effects) {
+            dup->effects = new Array<RenderEffect*>;
+            ARRAY_FOREACH(p, *effects) {
+                RenderEffect* ret = nullptr;
+                switch ((*p)->type) {
+                    case SceneEffect::GaussianBlur: {
+                        ret = new RenderEffectGaussianBlur(*(RenderEffectGaussianBlur*)(*p));
+                        break;
+                    }
+                    case SceneEffect::DropShadow: {
+                        ret = new RenderEffectDropShadow(*(RenderEffectDropShadow*)(*p));
+                        break;
+                    }
+                    case SceneEffect::Fill: {
+                        ret = new RenderEffectFill(*(RenderEffectFill*)(*p));
+                        break;
+                    }
+                    case SceneEffect::Tint: {
+                        ret = new RenderEffectTint(*(RenderEffectTint*)(*p));
+                        break;
+                    }
+                    case SceneEffect::Tritone: {
+                        ret = new RenderEffectTritone(*(RenderEffectTritone*)(*p));
+                        break;
+                    }
+                    default: break;
+                }
+                if (ret) {
+                    ret->rd = nullptr;
+                    ret->valid = false;
+                    dup->effects->push(ret);
+                }
+            }
+        }
 
         return scene;
     }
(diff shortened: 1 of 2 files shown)


src/renderer/tvgRender.h
```diff
@@ -438,6 +439,17 @@ struct RenderEffectGaussianBlur : RenderEffect
         inst->type = SceneEffect::GaussianBlur;
         return inst;
     }
+
+    RenderEffectGaussianBlur* duplicate() const override
+    {
+        auto inst = new RenderEffectGaussianBlur;
```
hermet: There’s no need to write code unnecessarily verbosely. One-line assignment is enough for copy structure data...

src/renderer/tvgScene.h
```diff
@@ -272,7 +272,25 @@ struct SceneImpl : Scene
...
+                newEffect = (*effect)->duplicate();
+
+                if (newEffect) {
+                    newEffect->extend = (*effect)->extend;
+                    newEffect->valid = (*effect)->valid;
+                    dup->effects->push(newEffect);
+                }
+            }
+
+            if (dup->effects->empty()) {
```
hermet: Pointless check: don't try to copy if the effects are empty from the start.

src/renderer/tvgScene.h
```diff
@@ -272,7 +272,25 @@ struct SceneImpl : Scene
             dup->paints.push_back(cdup);
         }
 
-        if (effects) TVGERR("RENDERER", "TODO: Duplicate Effects?");
+        if (effects) {
+            dup->effects = new Array<RenderEffect*>;
+
+            ARRAY_FOREACH(effect, *effects) {
+                RenderEffect* newEffect = nullptr;
+                newEffect = (*effect)->duplicate();
```
hermet: auto effect->duplicate();

src/renderer/tvgScene.h
```diff
@@ -272,7 +272,25 @@ struct SceneImpl : Scene
...
+
+                if (newEffect) {
```
hermet: -- condition.

src/renderer/tvgScene.h
```diff
@@ -272,7 +272,25 @@ struct SceneImpl : Scene
...
+                    newEffect->extend = (*effect)->extend;
```
hermet: --

src/renderer/tvgScene.h
```diff
@@ -272,7 +272,25 @@ struct SceneImpl : Scene
...
+                    newEffect->valid = (*effect)->valid;
```
hermet: --

src/renderer/tvgRender.h
```diff
@@ -419,6 +419,7 @@ struct RenderEffect
     bool valid = false;
 
     virtual ~RenderEffect() {}
+    virtual RenderEffect* duplicate() const { return nullptr; };
```
hermet: `virtual RenderEffect* duplicate() const = 0;`
Not usre yet, ultimately `switch` could be more efficient than polymorphism here...

hermet (changes requested):

src/renderer/tvgRender.h
tinyjin: Yes, the copy logic is simple, using a `switch` looks efficient. Still, there is benefit of polymorphism, even single line can get little wordy with 5 cases of switch, and having each implementation would appears clearer. It also leaves room if some additional logic needs to be added to `duplicate` later.

I'd still like to go with this, but let me know if it's bad choice for current logic.

hermet: @tinyjin Polymorphism is inefficient in that it increases the virtual memory table. Since effects are simply for storing data, it seems more appropriate to focus on performance efficiency rather than maintenance. The most definitive criterion would probably be the difference in binary size.

Issue #3602, opened 2025-07-08
lottie: drop shadow spec mismatch


hermet:
The rendering result for 0 blur is not consistent across engines.
All engines need to treat a 0-blur shadow the same as the raw shadow image, without any blur applied.

ThorVG:
<img width="939" height="727" alt="Image" src="https://github.com/user-attachments/assets/1bef47a1-89f2-42fa-b184-a566f1636d74" />

Expected:

<img width="719" height="569" alt="Image" src="https://github.com/user-attachments/assets/d45f73d6-50f0-434b-832b-d936310b96dc" />

[16251.json](https://github.com/user-attachments/files/21118468/16251.json)

PR #3639, opened 2025-07-23
gl/wg engines: zero size blur dropshadow support

SergeyLebedkin:
gl/wg engines: zero size blur dropshadow support
in case of sigma == 0 there is no any sense to apply blur original image used as the shadow source
EffectDropShadow example changed to reproduce zero sigma shadow

before (with sigma 0)
before
after (with sigma 0)
after

https://github.com/thorvg/thorvg/issues/3602


hermet (changes requested):
please polish the commit msg.

image

hermet:
@SergeyLebedkin could you please double-check the wgpu version? left border shadow shouldn't be visible

left: wgpu (wrong) 
right: gl (correct)
image

SergeyLebedkin:
> @SergeyLebedkin could you please double-check the wgpu version? left border shadow shouldn't be visible

 Looks like its a rounding problem. Fixed

image

hermet (approved):

PR #3645, opened 2025-07-24
sw_engine: prevent NaN in gradient pixel calculations

tinyjin:
### Test file : [radial-basic-v2.json](https://github.com/user-attachments/files/21415205/radial-basic-v2.json)

CleanShot 2025-07-25 at 01 25 10@2x

---

Add std::isfinite() check in _pixel() to prevent undefined behavior when NaN values are cast to int32_t. This prevents potential crashes during gradient color table indexing.

NaN can occur from:
- Division by zero in radial gradient calculations
- Negative values passed to sqrtf() in radial gradient determinant


hermet:
This could also hurt performance. We need to understand where this exception originates from.

hermet:
@tinyjin please add an issue if any unexpected rendering result were observable. thanks..

PR #3646, opened 2025-07-24
lottie/parser: fix uint8_t overflow in getValue function

tinyjin:
## Test file :  [text-radialGradient-stroke.json](https://github.com/user-attachments/files/21415559/text-radialGradient-stroke.json)

CleanShot 2025-07-25 at 01 53 21@2x

---

Fix runtime error when float values exceed uint8_t range (0-255). The getValue(uint8_t&) function was directly casting float values multiplied by 2.55f to uint8_t without range validation.

This caused overflow errors when input values exceeded ~100% (e.g. opacity value is greater than 100).

Added tvg::clamp() to ensure values stay within valid uint8_t range (0-255), preventing runtime overflow errors.

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -264,11 +264,11 @@ bool LottieParser::getValue(uint8_t& val)
 {
     if (peekType() == kArrayType) {
         enterArray();
-        if (nextArrayValue()) val = (uint8_t)(getFloat() * 2.55f);
+        if (nextArrayValue()) val = (uint8_t)(tvg::clamp(getFloat() * 2.55f, 0.0f, 255.0f));
         //discard rest
         while (nextArrayValue()) getFloat();
     } else {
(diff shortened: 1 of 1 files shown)


tinyjin:
> This looks wrong handling, if the resource violates the spec.

Okay, one question. Even if the assets violated spec, we'd not need defensive logic?

hermet:
> > This looks wrong handling, if the resource violates the spec.
> 
> Okay, one question. Even if the assets violated spec, we'd not need defensive logic?

@tinyjin Basically, no. There's no magic that can make software behave as users expect when the specification is violated. Instead, libraries or engines should help users understand the problem situation and firstly fix their software properly rather hiding the issue. So here is my question, what is the practical enhancement (user-perspective but not about sanitizer report) with this fix?

And please note that the reason I’m rejecting this PR is because the function is clearly designed for property values in the 0–255 range. If a property needs to support larger values, then the property type should be modified, rather than changing the method itself. By making exceptions for rare cases, you harm the general use cases.

Issue #3314, opened 2025-03-06
gl_engine: initial boot screen is not synced with window appearance


hermet:
When launching a sample app, the first boot screen appears with black. It seems that precise synchronization is needed between the window display and the sample's content preparation calls. This may require modifications within the sample or the gl engine.

https://github.com/user-attachments/assets/64c939fa-f028-4a7d-9ed0-88b890a83e42

hermet:
@SergeyLebedkin need your attention.

hermet:
Still a valid issue. Need to re-evaluate it.

hermet:
It's unfortunate that Windows systems behave inconsistently across different machines. There’s no reliable way today to guarantee perfect behavior through the generic SDL interface. Only WSL and some older macOS machines (which are not our primary targets) show this issue, so I’ve decided to give up on fully resolving it.

PR #3650, opened 2025-07-25
examples: fix first frame sync with window appearance

Nor-s:
In some examples, the first frame is black. This goes away if `refresh` before `SDL_ShowWindow `.

reproduce:
- run Blending example (GL,SW)

issue: #3314

|Previous | Patched |
|-|-|
|Jul-25-2025 22-55-30| Jul-25-2025 22-37-41|

### DIFF
--- a/examples/Example.h
+++ b/examples/Example.h
@@ -198,13 +198,14 @@ struct Window
         if (!verify(canvas->draw())) return false;
         if (!verify(canvas->sync())) return false;
 
+        refresh();
+
         return true;
     }
 
     void show()
     {
         SDL_ShowWindow(window);
-        refresh();
 
         //Mainloop
         SDL_Event event;


examples/Example.h
```diff
@@ -198,13 +198,14 @@ struct Window
         if (!verify(canvas->draw())) return false;
         if (!verify(canvas->sync())) return false;
```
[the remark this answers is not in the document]
hermet: @Nor-s could you please add a descriptive comment (copilot suggestion) as well?

examples/Example.h
```diff
@@ -198,13 +198,14 @@ struct Window
...
 
+        refresh();
+
         return true;
     }
 
     void show()
     {
         SDL_ShowWindow(window);
-        refresh();
```
hermet, line 207: I think we can leave this refresh call here. After removing this, examples are not visible at at on my end.

hermet (changes requested):
broken examples after this change.

examples/Example.h
Nor-s: I will check to make sure it looks good in other environments. Thank you for checking.

hermet:
https://github.com/thorvg/thorvg/pull/3655

hermet:
@Nor-s Let's forget this issue. Reviwed again, this looks a false issue which is only observable on wsl. Thanks for your effort.

Nor-s:
> @Nor-s Let's forget this issue. Reviwed again, this looks a false issue which is only observable on wsl. Thanks for your effort.

Thanks for the confirmation.
At one point, I was considering drawing to both the front and back buffers before showing the window.
Looking back, that would’ve been a terrible idea.

Here’s what I had in mind:
1. draw()
2. refresh()
3. draw()
4. showWindow()
5. refresh()

PR #3658, opened 2025-07-29
gl_engine: neat code++

hermet:


SergeyLebedkin:
Also it will be nice to populate it on wg stroker

PR #3660, opened 2025-07-30
renderer: support a feature for paint hit-detection

hermet:
The new APIs tests whether the specified rectangular region, given as individual integer values (`x`, `y`, `w`, `h`), intersects
the interior (fill area) of the shape. The intersection test considers the shape's fill rule (winding or even-odd) and applies to
the filled area including the stroke.

C++ API:
 + bool Paint::intersects(int32_t x, int32_t y, int32_t w, int32_t h)

C API:
 + bool tvg_paint_intersects(Tvg_Paint* paint, int32_t x, int32_t y, int32_t w, int32_t h)

Please note that this is an experimental support,
only sw canvas is supported. API/Spec can be updated later.

https://github.com/user-attachments/assets/4d562300-b084-41b7-991e-9aca6a0099e9

https://github.com/user-attachments/assets/4194b897-f38d-4a39-8c06-95b41f42a0da


hermet:
\+ @BenjaminHalko

BenjaminHalko (commented):
@Hermet Hey this looks great! Out of curiosity, will you be adding an API for single point hit detection as well?

hermet:
@BenjaminHalko hi.

> Hey this looks great! Out of curiosity, will you be adding an API for single point hit detection as well?

I suppose you can set w/h =1 for single point hit detection.

> Also, how hard is it to add support for Picture and Text as well? It doesn't need to be as precise, just their OBB. If it is too difficult no need to implement them right now.

I was thinking of promoting the API from the Shape to the Paint level, so it can be more broadly used. However, this API is specifically intended for precise hit-testing. It will try hit-test for the actual visual region of Text glyphs once the api were promoted.

Please let us consider about this further, while you can use the existing bounds() method for AABB or OBB checks. Thanks.

BenjaminHalko (approved):
LGTM

BenjaminHalko (approved):
Thanks @hermet These changes LGTM!

hermet:
https://github.com/user-attachments/assets/44324847-47b0-47a6-9561-8faeb59011e1

https://github.com/user-attachments/assets/0c2a887e-110f-4187-ae29-d8d6fef747a9

https://github.com/user-attachments/assets/f615976e-2dab-4efe-8f85-d925cfe08a74

https://github.com/user-attachments/assets/532d13ae-b7a2-406e-8bcc-a93640334977

PR #3663, opened 2025-07-30
renderer: apply engine-based AABB support

hermet:
Use bounding box information provided by the engine to ensure precise support for the bounding box,
including stroke visuals.

Please note that gl/wg engines currently doesn't compatible with this update.


hermet:
\+ @BenjaminHalko

PR #3678, opened 2025-08-02
wasm: add TypeScript bindings for the WASM module

kunyoungparkk:
This PR improves TypeScript bindings for the WASM module by:

- Enables `.d.ts` generation via `--emit-tsd`
- Defines custom `val`-based types using `EMSCRIPTEN_DECLARE_VAL_TYPE`

These changes are related to [`thorvg.web#16`](https://github.com/thorvg/thorvg.web/issues/16), which requests better TypeScript support for the WASM output.

This PR focuses on updating the C++ binding layer. A follow-up change may be submitted to `thorvg.web` to utilize these bindings in the front-end.

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -30,6 +30,9 @@ using namespace emscripten;
 using namespace std;
 using namespace tvg;
 
+EMSCRIPTEN_DECLARE_VAL_TYPE(Uint8Array);
+EMSCRIPTEN_DECLARE_VAL_TYPE(Float32Array);
+
 static const char* NoError = "None";
 
 struct TvgEngineMethod
@@ -37,9 +40,9 @@ struct TvgEngineMethod
     virtual ~TvgEngineMethod() {}
     virtual Canvas* init(string&) = 0;
     virtual void resize(Canvas* canvas, int w, int h) = 0;
-    virtual val output(int w, int h)
+    virtual Uint8Array output(int w, int h)
     {
-        return val(typed_memory_view<uint8_t>(0, nullptr));
+        return Uint8Array(val(typed_memory_view<uint8_t>(0, nullptr)));
     }
 
     void loadFont() {
@@ -74,9 +77,9 @@ struct TvgSwEngine : TvgEngineMethod
         static_cast<SwCanvas*>(canvas)->target((uint32_t *)buffer, w, w, h, ColorSpace::ABGR8888S);
     }
 
-    val output(int w, int h) override
+    Uint8Array output(int w, int h) override
     {
-        return val(typed_memory_view(w * h * 4, buffer));
+        return Uint8Array(val(typed_memory_view(w * h * 4, buffer)));
     }
 };
 
@@ -278,9 +281,9 @@ public:
         return errorMsg;
     }
 
-    val size()
+    Float32Array size()
     {
-        return val(typed_memory_view(2, psize));
+        return Float32Array(val(typed_memory_view(2, psize)));
     }
 
     float duration()
@@ -345,17 +348,17 @@ public:
         return true;
     }
 
-    val render()
+    Uint8Array render()
     {
         errorMsg = NoError;
 
-        if (!canvas || !animation) return val(typed_memory_view<uint8_t>(0, nullptr));
+        if (!canvas || !animation) return Uint8Array(val(typed_memory_view<uint8_t>(0, nullptr)));
 
         if (!updated) return engine->output(width, height);
 
         if (canvas->draw(true) != Result::Success) {
             errorMsg = "draw() fail";
-            return val(typed_memory_view<uint8_t>(0, nullptr));
+            return Uint8Array(val(typed_memory_view<uint8_t>(0, nullptr)));
         }
 
         canvas->sync();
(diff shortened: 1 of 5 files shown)


tinyjin:
Thank you @kunyoungparkk 

Does this patch itself add type support for `thorvg.web` in glue code usage?
I believe the corresponding patch on the thorvg.web side is required and should be paired with this PR.

kunyoungparkk:
@tinyjin Thank you for your feedback.

As you pointed out, I've now submitted a related PR to thorvg.web to apply the new TypeScript bindings accordingly.

tinyjin:
@kunyoungparkk 

In my opinion, TypeScript should not be forced on every user.
- JavaScript & WASM: Provide the basic runtime and ensure compatibility.
- TypeScript Definitions: Offer type stability and development convenience.

We can add a separate script to generate TypeScript definitions so that users aren’t blocked by extra requirements and can choose TypeScript if it fits their project.

CleanShot 2025-08-04 at 01 00 25@2x

kunyoungparkk:
@tinyjin Thanks for the feedback! I agree with your opinion. I created a separate script(wasm_build_with_types.sh) for generating TypeScript definitions.

tinyjin:
> @tinyjin Thanks for the feedback! I agree with your opinion. I created a separate script(wasm_build_with_types.sh) for generating TypeScript definitions.

Thank you. Can we just make `wasm_build_with_types.sh` build for TS definition only? I think type definition doesn't need to be separated by sw, gl or wg. We can export common type definition and use them.

kunyoungparkk:
@tinyjin As you may already know, Emscripten doesn’t support .d.ts generation without building the wasm and glue js files.
However, I can configure the wasm_build_with_types.sh script to keep only the generated .d.ts file and remove wasm and glue js files.
Please let me know if this approach works for you.

wasm_build_with_types.sh
```diff
@@ -0,0 +1,30 @@
...
+      sed "s|EMSDK:|$EMSDK|g; s|'--bind'|'--bind', '--emit-tsd=thorvg-wasm.d.ts'|g" ./cross/wasm32_sw.txt > /tmp/.wasm_cross.txt
+      meson setup -Db_lto=true -Ddefault_library=static -Dstatic=true -Dloaders="all" -Dsavers="all" -Dthreads=false -Dbindings="wasm_beta" -Dpartial=false --cross-file /tmp/.wasm_cross.txt build_wasm
+    elif [[ "$BACKEND" == "gl" ]]; then
+      sed "s|EMSDK:|$EMSDK|g; s|'--bind'|'--bind', '--emit-tsd=thorvg-wasm.d.ts'|g" ./cross/wasm32_gl.txt > /tmp/.wasm_cross.txt
+      meson setup -Db_lto=true -Ddefault_library=static -Dstatic=true -Dloaders="all" -Dsavers="all" -Dthreads=false -Dbindings="wasm_beta" -Dpartial=false -Dengines="gl" --cross-file /tmp/.wasm_cross.txt build_wasm
+    else
+      sed "s|EMSDK:|$EMSDK|g; s|'--bind'|'--bind', '--emit-tsd=thorvg-wasm.d.ts'|g" ./cross/wasm32.txt > /tmp/.wasm_cross.txt
+      meson setup -Db_lto=true -Ddefault_library=static -Dstatic=true -Dloaders="all" -Dsavers="all" -Dthreads=false -Dbindings="wasm_beta" -Dpartial=false -Dengines="all" --cross-file /tmp/.wasm_cross.txt build_wasm
+    fi
+fi
```
tinyjin: Rather than we compromise with toolchain's default behavior, I'd like to focus on the way we need and simplification.

In `tvgWasmLottieAnimation`, we provide 3 engine types that are abstracted through TvgEngineMethod. This means we don't need duplicated type definitions for each engine. Therefore, generating separate types for each backend engine would be an invalid approach.

> TvgSwEngine, TvgGLEngine, TvgWgEngine implement the same interface.

We can avoid building unnecessary WASM binaries that were already built in the basic process. It should only be responsible for generating types. Even if Emscripten forces creating WASM files to have type def, we can simply ignore them and extract only the `d.ts` file as you mentioned. If we leave them as is, then users can confuse what's the binary to go.

tinyjin (commented):
@kunyoungparkk Please check comment. This is my idea about you pointed above.

wasm_build_with_types.sh
kunyoungparkk: Thank you. d.ts generation without rebuilding wasm would be nice, but due to the current Emscripten behavior, I isolated the process in a temp dir to avoid affecting main output.
Let me know if this looks okay or needs to be done differently.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -30,16 +30,19 @@ using namespace emscripten;
 using namespace std;
 using namespace tvg;
 
+EMSCRIPTEN_DECLARE_VAL_TYPE(Uint8ArrayView);
+EMSCRIPTEN_DECLARE_VAL_TYPE(Float32ArrayView);
```
tinyjin: `Uint8ArrayView` and `Float32ArrayView` represent `Uint8Array` and `Float32Array`, is there any reason we have different name here other than commonly used name?

kunyoungparkk: There wasn’t a strong reason. I just wanted to distinguish them from the JS types, but now I realize using the same names improves readability. I have updated the names and pushed the changes!

tinyjin:
#3707

PR #3680, opened 2025-08-03
wg_engine: hsl blending support

SergeyLebedkin:
Introduced HSL blending support for webgpu: Hue, Saturation, Color, Luminosity
Supported for all blending targets: solid, gradient, image, scene

sw/wg
image
image

### DIFF
--- a/src/renderer/wg_engine/tvgWgPipelines.cpp
+++ b/src/renderer/wg_engine/tvgWgPipelines.cpp
@@ -20,8 +20,8 @@
  * SOFTWARE.
  */
 
-#include "tvgWgPipelines.h"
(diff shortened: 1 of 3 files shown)


hermet:
@SergeyLebedkin could you please review the anti-aliasing part? This must be addressed together.

image

SergeyLebedkin:
> @SergeyLebedkin could you please review the anti-aliasing part? This must be addressed together.

As I can see sw and wg have the same rendering results for shape borders 

image

hermet:
@SergeyLebedkin That's why I'm requesting it. Please review it carefully to address it properly. No need to follow up the wrong one.

Issue #3671, opened 2025-08-01
renderer: `SwRenderer` is not destructed when using `Saver::background()`


jiggyjiggy:
#### Summary
- When un-commenting the `TEST_CASE("Save a lottie into gif", "[tvgSavers]")` block in `testSavers.cpp`, the test fails due to `Initializer::term()` returning `Result::InsufficientCondition (2)` instead of `Result::Success (0)`. This happens because `SwRenderer` is not destructed after the second `sync()` call.

#### Steps to Reproduce
1. Remove `#if 0` guard in `testSavers.cpp`
2. Run the unit test

#### Expected Behavior
`Initializer::term()` should return `Result::Success (0)` and all resources including `SwRenderer` should be released properly.

#### Actual Behavior
- `rendererCnt` remains at 1 (expected 0)
- `SwRenderer` destructor is not called after second `sync()`
- Test fails with:
  ```c++
  REQUIRE(Initializer::term() == Result::Success) // returns 2
  ```

#### Suspected Cause
ref/unref mismatch
- Saver::background(bg) increases ref count without corresponding unref

hermet:
@jiggyjiggy Thanks, but conceptually, this is a expected behavior. The engine must be terminated only after all resources used by the user have been freed.

hermet:
@jiggyjiggy please try this way.
```
    REQUIRE(Initializer::init() == Result::Success);
    {
           .... 
    }
    REQUIRE(Initializer::term() == Result::Success);
```

jiggyjiggy:
Is the following code what you meant by that usage?

```c++
#if defined(THORVG_GIF_SAVER_SUPPORT) && defined(THORVG_LOTTIE_LOADER_SUPPORT)

TEST_CASE("Save a lottie into gif", "[tvgSavers]")
{
    REQUIRE(Initializer::init(0) == Result::Success);
    {
        auto animation = Animation::gen();
        REQUIRE(animation);

        auto picture = animation->picture();
        REQUIRE(picture);
        REQUIRE(picture->load(TEST_DIR"/test.json") == Result::Success);
        REQUIRE(picture->size(100, 100) == Result::Success);

        auto saver =  unique_ptr<Saver>(Saver::gen());
        REQUIRE(saver);
        REQUIRE(saver->save(animation, TEST_DIR"/test.gif") == Result::Success);
        REQUIRE(saver->sync() == Result::Success);

        //with a background
        auto animation2 = Animation::gen();
[...]

PR #3683, opened 2025-08-04
savers/gif: Unref bg after push to fix ref count

jiggyjiggy:
https://github.com/thorvg/thorvg/issues/3671#issuecomment-3143892584

1
The test failure in test/testSavers.cpp was due to an incorrect ref count of `bg`. 
Added unref in `GifSaver::run` to resolve it.

2
Added brace-format usage for Initializer

### DIFF
--- a/src/savers/gif/tvgGifSaver.cpp
+++ b/src/savers/gif/tvgGifSaver.cpp
@@ -44,8 +44,8 @@ void GifSaver::run(unsigned tid)
 
     if (bg) {
         bg->unref();
+        bg = nullptr;
     }
-    bg = nullptr;
 
     canvas->push(animation->picture());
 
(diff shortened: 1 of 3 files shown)


src/savers/gif/tvgGifSaver.cpp
```diff
@@ -41,6 +41,10 @@ void GifSaver::run(unsigned tid)
     buffer = tvg::realloc<uint32_t*>(buffer, sizeof(uint32_t) * w * h);
     canvas->target(buffer, w, w, h, ColorSpace::ABGR8888S);
     canvas->push(bg);
+
+    if (bg) {
```
hermet:
```
if (bg) {
   bg->unref();
   bg = nullptr;
}
```

hermet: @jiggyjiggy Since on this, could you please move this bg assignment logic at the end of the function (Ln79)? Thanks.

image

jiggyjiggy: Thanks for the review.
Applied the changes

Issue #3677, opened 2025-08-01
wg/gl_engine: effects regressions


hermet:
inaccurate border clamping is observable at v1.0-pre25

wgpu:

<img width="1069" height="1118" alt="Image" src="https://github.com/user-attachments/assets/ea72a602-2180-4e2a-bc80-bd02c5a86316" />

gl:

<img width="1063" height="1065" alt="Image" src="https://github.com/user-attachments/assets/fc4ffbe9-d0a1-4ef5-ace1-77b94f77f8cb" />

PR #3686, opened 2025-08-04
gl_engine: fix gauss border clamping

SergeyLebedkin:
fixed observable border clamping for gauss filter in opengl

by design, the gl renderer creates temporary off-screen buffers for rendering scenes, 
including for applying effects. the size of the off-screen buffer depends on the scene bbox 
and it is offset relative to the screen to the beginning of the bbox. 
in this case, the off-screen buffer may extend beyond the screen boundaries, 
but the scene will only be rendered in the area of intersection of the screen and the buffer itself. in
this case, we do not have enough data to apply the gaussian filter, 
since the source of data for it may be neighboring pixels that were initially cut off 
and are not in the area of intersection of the screen viewport and the off-screen buffer. 
we can cut off such pixels when calculating the gaussian filter by first passing the active viewport 
to the shader and not taking into account the pixels that are not included in it

https://github.com/thorvg/thorvg/issues/3677

image

### DIFF
--- a/src/renderer/gl_engine/tvgGlEffect.cpp
+++ b/src/renderer/gl_engine/tvgGlEffect.cpp
@@ -77,7 +77,9 @@ GlRenderTask* GlEffect::render(RenderEffectGaussianBlur* effect, GlRenderTarget*
     auto dstCopyFbo1 = blendPool[1]->getRenderTarget(vp);
 
     // add uniform data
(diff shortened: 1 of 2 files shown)


hermet:
@SergeyLebedkin Thanks for your fix but this is a very strange. If this issue is caused by a recent change, we need to clarify the connection & fundamental reason between the two. If we can avoid handling the viewport like before it worked, that would be the best option.

PR #3690, opened 2025-08-05
gl_engine: hsl blending support

SergeyLebedkin:
Introduced HSL blending support: Hue, Saturation, Color, Luminosity 
Supported for all blending solid, gradient and image targets
Scenes custom blend does not supported, will be added as separate PR

image


hermet:
@SergeyLebedkin Could you please separate commits 1, 2? 

`1. Introduced HSL blending support: Hue, Saturation, Color, Luminosity`
`2. Supported for all blending solid, gradient and image targets`

Also, what is the definition of the scene custom blend? I suppose the fix commit can be applied in this PR.

hermet:
see: https://github.com/thorvg/thorvg/pull/3081

hermet (changes requested):
please check comments thanks.

SergeyLebedkin:
> Also, what is the definition of the scene custom blend?

It means that gl renderer dous not support any other blend except normal blend. It will be fixed in this PR

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -148,16 +151,33 @@ void GlRenderer::initShaders()
...
-    mPrograms.push(new GlProgram(MASK_VERT_SHADER, SCREEN_BLEND_FRAG));
-    mPrograms.push(new GlProgram(MASK_VERT_SHADER, OVERLAY_BLEND_FRAG));
-    mPrograms.push(new GlProgram(MASK_VERT_SHADER, COLOR_DODGE_BLEND_FRAG));
-    mPrograms.push(new GlProgram(MASK_VERT_SHADER, COLOR_BURN_BLEND_FRAG));
-    mPrograms.push(new GlProgram(MASK_VERT_SHADER, HARD_LIGHT_BLEND_FRAG));
-    mPrograms.push(new GlProgram(MASK_VERT_SHADER, SOFT_LIGHT_BLEND_FRAG));
-    mPrograms.push(new GlProgram(MASK_VERT_SHADER, DIFFERENCE_BLEND_FRAG));
-    mPrograms.push(new GlProgram(MASK_VERT_SHADER, EXCLUSION_BLEND_FRAG));
+        // custom blend shaders
+    const char* customBlendShaders[17] {
```
hermet: Except for 'normal',  other blend methods are not commonly used or never used in apps. Can you try initialize them in a deferred way?

SergeyLebedkin: RAII? I think its a good idea. Yes, we can do that. Will fix it now

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -103,6 +105,7 @@ void GlRenderer::initShaders()
 #if 1  //for optimization
     #define LINEAR_TOTAL_LENGTH 2770
     #define RADIAL_TOTAL_LENGTH 5272
+    #define BLEND_TOTAL_LENGTH 8192
```
hermet, line 106: please update the else case  (113) as well.

src/renderer/gl_engine/tvgGlRenderer.h
```diff
@@ -132,6 +121,9 @@ class GlRenderer : public RenderMethod
     GlRenderTarget mRootTarget;
     GlEffect mEffect;
     Array<GlProgram*> mPrograms;
```
hermet, line 160: One single array seems enough for managing programs for compact size.

SergeyLebedkin: > One single array seems enough for managing programs for compact size.

From other side access by index faster than if-else or switch statment while selecting of the proper blend shader

hermet: @SergeyLebedkin if you know the starting idx for program segments, that's it. no any slow down.

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -586,20 +584,59 @@ void GlRenderer::endBlendingCompose(GlRenderTask* stencilTask, const Matrix& mat
...
+        EXCLUSION_BLEND_FRAG,
+        HUE_BLEND_FRAG,
+        SATURATION_BLEND_FRAG,
+        COLOR_BLEND_FRAG,
+        LUMINOSITY_BLEND_FRAG,
+        ADD_BLEND_FRAG
+    };
+    // custom blend shaders helpers
+    static const char* customBlendHelpers[17] {
+        "", "", "", "", "", "", "", "", "", "", "", "",
```
hermet: could you please remove the garbage memory?

SergeyLebedkin: I will try to update this mechanics, without empty strings

hermet:
added `RuiwenTang <tangruiwen1989@gmail.com>` as co-author.

hermet (approved):
@SergeyLebedkin good job. thanks!

Issue #3481, opened 2025-05-23
Support Web Lottie Players Preset


hermet:
Provide prepared presets of Lottie web players for user selection. We can consider to create `thorvg.lottie` project.

### Standard version
- lottie-sw (+sw, +lottie, +png, +webp, +jpg, +ttf -file)
- lottie-gl (+gl, +lottie, +png, +webp, +jpg +ttf  -file)

### Lite version
- lottie-lite-sw (+sw, +lottie, +png  -file -expressions)
- lottie-lite-gl (+gl, +lottie, +png -file -expressions)

See: https://cdnjs.com/libraries/lottie-web

hermet:
Font support: https://fontsource.org/

tinyjin:
Ideally ~~we can have common glue code~~, and separate WASM binaries in required types as gluecode size diff is not so considerable on each configurations. Also this would help users seamlessly load preset binaries in runtime without import statement change.

> Sharing a single Emscripten JS glue code across different .wasm files isn't good idea for now.
Each .wasm binary may have differences in internal tables, memory layout, and export sets, which can cause unexpected runtime inconsistencies when using a shared glue file.
It is more reasonable to generate a separate JS glue file per variant and apply size optimizations to each glude codes individually.

Also we might need to remove `wasm` postfix in it's name next to the extension. Just in case binary name is too long.

- lottie-player.js (common glue code)
- thorvg-sw-wasm.wasm
- thorvg-gl-wasm.wasm
- thorvg-sw-lite-wasm.wasm
- thorvg-gl-lite-wasm.wasm

tinyjin:
Related: https://github.com/thorvg/thorvg.web/issues/108

tinyjin:
When the TTF loader is disabled for the lite version, tvgWasmDefaultFont still increases the binary size unnecessarily, even though the font cannot be used. This should be handled with appropriate conditional build.

### Lite version
<img width="3829" height="476" alt="Image" src="https://github.com/user-attachments/assets/c24151ea-5946-403a-ad6c-8adc534b0597" />

### Lite version, --WASM default font
<img width="1932" height="250" alt="Image" src="https://github.com/user-attachments/assets/0467e78d-bb51-4d6f-939b-b44dac8f38d7" />

tinyjin:
## WASM Binary Size comparison

### SW (640KB)
[...]

PR #3697, opened 2025-08-07
wasm: apply conditional TTF loader support

tinyjin:
Add conditional compilation for the TTF loader in the WASM binding. This allows proper WASM builds even when the TTF loader is disabled, preventing unnecessary binary size increase from unused font additions. Default font functionality is preserved when the loader is enabled.

This helps maintain an optimal binary size when building a lightweight version of the player.

related: https://github.com/thorvg/thorvg/issues/3481

### DIFF
--- a/src/bindings/wasm/tvgWasmDefaultFont.h
+++ b/src/bindings/wasm/tvgWasmDefaultFont.h
@@ -95,6 +95,8 @@
 #ifndef _TVG_WASM_DEFAULT_FONT_H_
 #define _TVG_WASM_DEFAULT_FONT_H_
 
+#ifdef THORVG_TTF_LOADER_SUPPORT
+
 #include <cstddef>
 #include <atomic>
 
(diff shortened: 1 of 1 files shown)


src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -43,7 +45,15 @@ struct TvgEngineMethod
     }
 
     void loadFont() {
+#ifdef THORVG_TTF_LOADER_SUPPORT
```
hermet: I think if-def is better in `tvgWasmDefaultFont.h`, not here. Please make clean the main logic.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -43,7 +43,8 @@ struct TvgEngineMethod
     }
 
     void loadFont() {
-        Text::load("default", requestFont(), DEFAULT_FONT_SIZE, "ttf", false);
```
hermet: Please keep the original code. No need to change here.

PR #3700, opened 2025-08-07
examples: fix OpenGL black screen on initial render

effozen:
## Description

Fixed the issue where OpenGL examples display a black screen initially before content appears. This was caused by OpenGL's double buffering synchronization issue between window display and content preparation.

## Problem
- **Bug Type**: Rendering issue with OpenGL backend
- **Steps to Reproduce**:
  1. Build ThorVG with GL engine enabled: `meson setup builddir
-Dengines=gl -Dexamples=true`
  2. Run any example with GL: `./Lottie gl` or `./Animation gl`
  3. Observe black screen appears first before content is loaded

## Root Cause
When content is rendered in `ready()` before the window is shown, it only renders to the back buffer while the front buffer remains uninitialized (containing garbage/black pixels). The issue is more noticeable with examples that load files (Lottie, Animation) due to the loading time.

## Solution Applied
1. **Force buffer clear on initial render**: Changed `canvas->draw()` to `canvas->draw(true)` to ensure the buffer is cleared
2. **Add buffer swap in GlWindow::ready()**: Added `SDL_GL_SwapWindow()` after initial render to sync both front and back buffers
3. **Remove duplicate refresh()**: Removed redundant `refresh()` call after `SDL_ShowWindow()` to prevent unwanted buffer swap
4. **Explicit double buffering**: Added `SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1)` for clarity

## Changes
- Modified `examples/Example.h`
  - Made `Window::ready()` virtual to allow override
  - Added `GlWindow::ready()` override with buffer swap
  - Removed duplicate `refresh()` in `show()` method
  - Added explicit double buffer attribute setting

## Testing
- Tested with all GL examples (Lottie, Animation, Shapes, etc.)
- Verified SW renderer still works correctly
- No regression in other rendering backends

## Related Issues
#3314 

## Reference Video

https://github.com/user-attachments/assets/934e77da-c7b4-4459-afe8-7f5d59eeeb34

### DIFF
--- a/examples/Example.h
+++ b/examples/Example.h
@@ -188,14 +188,14 @@ struct Window
         return false;
     }
 
-    bool ready()
+    virtual bool ready()
     {
         if (!canvas) return false;
 
         if (!example->content(canvas, width, height)) return false;
 
         //initiate the first rendering before window pop-up.
-        if (!verify(canvas->draw())) return false;
+        if (!verify(canvas->draw(true))) return false;  //Force clear buffer on initial render
         if (!verify(canvas->sync())) return false;
 
         return true;
@@ -204,8 +204,7 @@ struct Window
     void show()
     {
         SDL_ShowWindow(window);
-        refresh();
-
+        
         //Mainloop
         SDL_Event event;
         auto running = true;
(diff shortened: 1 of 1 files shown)


effozen:
## Update: SW renderer fix added

Found that SW renderer had a similar issue after testing the GL fix.

**Additional changes:**
- Added `SwWindow::ready()` to clear and update surface before display
- SW uses single buffer with `SDL_GetWindowSurface()` that needs explicit initialization
- Different from GL's double buffering issue but same visual symptom

**Current status:**
- Both GL and SW renderers now work without initial black screen

https://github.com/user-attachments/assets/d6b6d258-a8f2-4284-902f-c55eda2f55cc

examples/Example.h
```diff
@@ -371,6 +390,79 @@ struct GlWindow : Window
         return true;
     }
 
+    void show() override
```
hermet, line 393: This is the wrong approach. ThorVG canvases have designed for abstraction, the example must show a single backbone main loop for all canvases.

hermet (changes requested):
Thanks for your contribution, but a generalized approach is essential. Thanks.

examples/Example.h
effozen: Thank you for quick review. I'll check more deeply, and try to solve by general approach. Thank you!

effozen:
As I studied related this issues, my approach is not good for this issues. 
To prevent conflict, I close this PR.

PR #3702, opened 2025-08-08
sw_engine: fine-tuning of render region calculation

hermet:
Compute the bounding box of the render region during path processing. Previously, this required an additional iteration over the points to perform the calculation.


hermet:
https://github.com/thorvg/thorvg/pull/3708

PR #3708, opened 2025-08-11
revise the engine based bbox support

hermet:
- Refactored to use bounding box information provided by the rendering engine
to ensure accurate AABB/OBB support, including stroke visuals.
Note that this requires paints to be properly pushed into the canvas,
and may affect API behavior.

- Updated Lottie/SVG loaders to use internal AABB functions,
which now exclude stroke bounds. The previous implementation
was also inaccurate in this regard.

- Adjusted example & tests accordingly.

Note:
- GL and WG render engines do not yet support these updates.
- This change may break backward compatibility in existing applications.

Before:
image

After:
image

### DIFF
--- a/examples/BoundingBox.cpp
+++ b/examples/BoundingBox.cpp
@@ -30,11 +30,16 @@ struct UserExample : tvgexam::Example
 {
     void bbox(tvg::Canvas* canvas, tvg::Paint* paint)
     {
-        //aabb
-        {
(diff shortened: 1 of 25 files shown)


hermet:
\+ @BenjaminHalko

BenjaminHalko (approved):
LGTM! Thanks @hermet !

Issue #3698, opened 2025-08-07
gl_engine: enhancement plan


SergeyLebedkin:
### Shaders 
Localize shader managment: add/remove/modify existing shaders and manage RAII (differed creation)

- [ ] declare and implement GlProgram and GlShader in the same cpp/h file
- [ ] move shaders to a separate structure-storage f.e. GlShaders
- [ ] request Unifirm/UBO/Samplers locations at the shader creation, instead of each frame by the name on shader binding

### Render Targets
Simplify render targets viewport/scissor and pooling logic

- [ ] make offscreen buffers the same size as the screen buffer, so we will have same coordinate system (offset and size) for all of them, so we do not need to recalculate viewports for scenes/shapes for each buffer. it will helps in composition/blending/effects
- [ ] it will also simplify the logic of render targets pools, because we din`t need to create new pool for the new target size

### Task based renderer
More tiny context swithing control, increase performance

- [ ] simplify the logic of the tasks. use just a tasks for shape/image and scene. it will help for proper control in context switching and concentrate all -information about current frame objects/targets

### Implement general Context instance
Universilize global entities creation and managment

- [ ] It will hold information about current GL context, screen information, containe shared objects will localize vertex/index/uniform buffers and textures creations/reallocation

### Gradient fill 
Cleanup gradient shaders and increase performance, reduce shader code size

- [ ] Use pre-calculated 1d texture for the gradient color map. It will reduce presure on the fragment shaders in case of gradient fill

### Sampler entity
Simplify texture creations and access type (repeat, clamp, mirror, filtration)

- [ ] start to use sampler entity instead of each texture local settings.

### Use generalized compositor
Hold general rasterization mechnics and entities at the same place

- [ ] use the single class for rasterization in different scenarios (cliping, blending, masking) for different object types (shapes/images/scenes)
- [ ] it will be holder for view matrix, temporary render targets, shaders, current render target and render pipeline state
- [ ] it will handle resize mechanics rasterization entities

### Batch draw calls to speed up 
( Updated by @wenjieshen  December 05, 2025)
- [ ] Batch convex shapes. (WIP)
- [ ] Use a texture instead of a stencil buffer to enable the batching of complex shapes.
- [ ] Merge the coverage map into a large image using a texture atlas, depending on hardware bandwidth.

###  Analytic AA
( Updated by @wenjieshen  December 05, 2025)
[...]

PR #3710, opened 2025-08-11
gl_engine: use full screen intermediate render targets

SergeyLebedkin:
by design the gl renderer created offscreen buffers of different sizes based on the scene box size. Because of this different offscreen buffers had their own coordinate system and had to deal with the alignment of offscreen buffers and the window buffer. It also meant having multiple buffer pools for different sizes. This complicates the logic and creates the need to recalculate alignment, viewports and scissors every frame, as well as texture coordinates for shaders.

Now all offscreen buffers have the same size as the window buffer, which simplifies the pool logic, as well as the blending/compositing/effects logic. The number of created buffers is equal to the depth of the scene graph, including temporary buffers for copies of the window buffer

https://github.com/thorvg/thorvg/issues/3698 - Render Targets

### DIFF
--- a/src/renderer/gl_engine/tvgGlEffect.cpp
+++ b/src/renderer/gl_engine/tvgGlEffect.cpp
@@ -67,14 +67,14 @@ void GlEffect::update(RenderEffectGaussianBlur* effect, const Matrix& transform)
 }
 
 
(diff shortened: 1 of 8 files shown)


hermet:
@tinyjin Please confirm any regression on the web. Thanks.

SergeyLebedkin:
> It Looks good.
@tinyjin 
Could you please share some performance changes?

For a quick check in a single animation, we can use _game_finished.json_ and _guitar.json_ animation from our standtard animations set. Run web player on the largest resolution as we can and compare performance

Thanx

Issue #3716, opened 2025-08-14
wg_engine: spec alignment


hermet:
if multiple call of `Canvas::update()`, before `Canvas::render()`, the rendering is broken. Need to investigate if we can make the engine safety that situation.

See the `BoundingBox` example:

<img width="1144" height="1179" alt="Image" src="https://github.com/user-attachments/assets/305bcea1-4872-469e-8f5b-874111f0a8d3" />

'StrokeLine` example

<img width="1000" height="1197" alt="Image" src="https://github.com/user-attachments/assets/650b692b-3ed4-4d13-8e4b-6e9e728b3445" />

relavent change: 
```
commit 08ea9a6039d56b600ac6cc21f6d4f9ed7eaaf760 (HEAD)
Author: Hermet Park <hermet@lottiefiles.com>
Date:   Thu Aug 7 00:42:10 2025 +0900

    api: revise canvas behavior for improved flexibility and efficiency

    This update adjusts the ThorVG canvas specification to
    allow more flexible and efficient usage patterns.

    Key changes include:

    a. push behavior:
     calling canvas push no longer triggers an automatic update
     on the paint object.  This avoids redundant updates in common
     scenarios where the user explicitly calls canvas update afterwards.

    b. implicit update on draw:
     if canvas draw is called without a prior canvas update,
     an implicit update will be performed. This ensures compatibility
     with the updated push behavior and prevents missing updates.

    c. relaxed success conditions
     the functions canvas update, sync, draw are now more tolerant
     of empty or idle canvas states. If no paint objects are present,
     they will return success instead of failure.

    Note: These changes may break backward compatibility in cases
    where stricter failure checks were previously relied upon.

    issue: https://github.com/thorvg/thorvg/issues/3116
```

SergeyLebedkin:
```
        tvg::Point pts[4] = {{158.205078f, 125.980759f}, {244.807617f, 175.980759f}, {234.807617f, 193.30127f}, {148.205078f, 143.30127f}};
        
        auto bound = tvg::Shape::gen();
        bound->moveTo(pts[0].x, pts[0].y);
        bound->lineTo(pts[1].x, pts[1].y);
        bound->lineTo(pts[2].x, pts[2].y);
        bound->lineTo(pts[3].x, pts[3].y);
        bound->close();
        bound->strokeWidth(2.0f);
        float dash[] = {3.0f, 10.0f};
        bound->strokeDash(dash, 2);
[...]

PR #3718, opened 2025-08-14
wg_engine: fix default shape and strokes visibility

SergeyLebedkin:
Shape or Strokes geometry must be marked as skipped by default to prevent draw calls for empty mesh

issue: https://github.com/thorvg/thorvg/issues/3716


hermet:
@SergeyLebedkin one more "DirectUpdate"
image

src/renderer/wg_engine/tvgWgRenderData.cpp
```diff
@@ -165,6 +165,7 @@ void WgRenderDataShape::updateMeshes(const RenderShape &rshape, RenderUpdateFlag
 
     // update fill shapes
     if (flag & (RenderUpdateFlag::Color | RenderUpdateFlag::Gradient | RenderUpdateFlag::Transform | RenderUpdateFlag::Path)) {
+        renderSettingsShape.skip = (rshape.color.a == 0) && (!rshape.fill);
```
hermet: you might wanna multiply `opacity` with `color.a`.

hermet (commented):
Please check them. everywhere you change the values and the skip conditions are all different. :( 

```
void WgRenderSettings::update(WgContext& context, const Fill* fill)
{
   ...
    skip = false;
};


void WgRenderSettings::update(WgContext& context, const RenderColor& c)
{
   ...
    skip = (c.a == 0);
};
```

You need to introduce coherent, manageable core logic:
```
update():
  //start with skip = true; 
  //turn off skip only if it turns out the visible object.
```
Not like your spaghetti code.

src/renderer/wg_engine/tvgWgRenderData.h
```diff
@@ -47,7 +47,7 @@ struct WgRenderSettings
     WgImageData gradientData;
     WgRenderSettingsType fillType{};
     WgRenderRasterType rasterType{};
-    bool skip{};
+    bool skip = true;
```
hermet: not necessary

src/renderer/wg_engine/tvgWgRenderData.cpp
```diff
@@ -214,6 +219,8 @@ void WgRenderDataShape::updateMeshes(const RenderShape &rshape, RenderUpdateFlag
 
 void WgRenderDataShape::releaseMeshes()
 {
+    renderSettingsShape.skip = true;
```
hermet: not necessary

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -146,6 +146,7 @@ RenderData WgRenderer::prepare(const RenderShape& rshape, RenderData data, const
 
     // setup fill settings
     renderDataShape->viewport = vport;
+    renderDataShape->updateVisibility(rshape, opacity);
     if (flags & RenderUpdateFlag::Gradient && rshape.fill) renderDataShape->renderSettingsShape.update(mContext, rshape.fill);
```
hermet: Question, does it need to update them if the visibility is false?

SergeyLebedkin: I think it won't be a waste

Issue #3722, opened 2025-08-18
gl_engine: broken blending visual


hermet:
After scene blending [4445f97be92ab3be10fde9ad3741cc706a2ea654](https://github.com/thorvg/thorvg/pull/3712), some visual contents are missing.

See: `examples/lottie/expressisons/intelia_logo_animation.json`

**Current:**
<img width="1003" height="1017" alt="Image" src="https://github.com/user-attachments/assets/c9438e50-2776-41ce-9597-aa33c2526fae" />

**Before:**  
<img width="1003" height="1017" alt="Image" src="https://github.com/user-attachments/assets/e430c8b7-b342-4eb1-8f66-77e518b41788" />

**Normal:**
<img width="1003" height="1017" alt="Image" src="https://github.com/user-attachments/assets/7d8b1a3c-d8ec-4d60-a290-6b40476d11c8" />

hermet:
@SergeyLebedkin could you please review this as well?

gl:
<img width="318" height="274" alt="image" src="https://github.com/user-attachments/assets/ee741796-e295-41e1-8616-75dcd3c63329" />

wg:
<img width="310" height="316" alt="image" src="https://github.com/user-attachments/assets/28b0a612-ff65-4a90-b7e4-7a95d623770c" />

sample:
```
    bool content(tvg::Canvas* canvas, uint32_t w, uint32_t h) override
    {
        auto bg = Shape::gen();
        bg->appendRect(0, 0, 1200, 800);
        bg->fill(255, 200, 200, 255);

        canvas->push(bg);

        auto unblendedScene = Scene::gen();

        auto blendedScene = Scene::gen();
        blendedScene->blend(BlendMethod::Add);

        auto rect = Shape::gen();
        rect->appendRect(100, 100, 100, 100);
        rect->fill(0, 255, 0);
        rect->strokeWidth(20);
        rect->strokeFill(255, 255, 255);
        blendedScene->push(rect);

        auto normalScene = Scene::gen();

        auto rect2 = Shape::gen();
        rect2->appendRect(50, 50, 100, 100, 0, 0, true);
        rect2->fill(0, 0, 255);
        rect2->strokeWidth(20);
        rect2->strokeFill(255, 255, 255);
        normalScene->push(rect2);

        unblendedScene->push(normalScene);
        unblendedScene->push(blendedScene);

        unblendedScene->blend(BlendMethod::Composition);

        canvas->push(unblendedScene);

        return true;
    }
```

PR #3730, opened 2025-08-19
gl engine: fix broken blending visual (sub-scenes alignment)

SergeyLebedkin:
Align intermediate fbo with source buffer instead of destination buffer. 
Align text coordinates for reading source and intermediate fbos.

image

https://github.com/thorvg/thorvg/issues/3722


hermet:
@SergeyLebedkin good job!

PR #3733, opened 2025-08-20
wasm: rename player binary for web presets support

tinyjin:
This PR is to align with `thorvg.web` preset-player support: https://github.com/thorvg/thorvg.web/pull/137

issue: https://github.com/thorvg/thorvg/issues/3481

### DIFF
--- a/src/bindings/wasm/meson.build
+++ b/src/bindings/wasm/meson.build
@@ -6,7 +6,7 @@ if cc.get_id() == 'emscripten'
     thorvg_wasm_dep = declare_dependency(include_directories
                                          : include_directories('.'), sources
                                          : source_file)
-    executable('thorvg-wasm',
+    executable('thorvg',
         [],
         include_directories : headers,
         dependencies : [thorvg_lib_dep, thorvg_wasm_dep],


hermet:
@tinyjin Suggestion: let's put new scripts in `thorvg.web`!

tinyjin:
@hermet I introduced new `wasm_build.sh` at thorvg.web. Also leaving original script here as is (I thought we better keep original one until we move total web binding to thorvg.web)

see: https://github.com/thorvg/thorvg.web/pull/137/commits/b2f39384911b69e927cd24d7fdbf217acaf38957#diff-3424d8015efb263b21fef6efcfe9d86b8d518e41e120d6dae9164562bdf56431

hermet:
> @hermet I introduced new `wasm_build.sh` at thorvg.web. Also leaving original script here as is (I thought we better keep original one until we move total web binding to thorvg.web)
> 
> see: [thorvg/thorvg.web@b2f3938#diff-3424d8015efb263b21fef6efcfe9d86b8d518e41e120d6dae9164562bdf56431](https://github.com/thorvg/thorvg.web/commit/b2f39384911b69e927cd24d7fdbf217acaf38957#diff-3424d8015efb263b21fef6efcfe9d86b8d518e41e120d6dae9164562bdf56431)

I think we'd prefer to unify around 'thorvg' rather than 'lottie' for future variants. Please consider to use `thorvg` in thorvg.web as well. Thanks.

PR #3736, opened 2025-08-21
wasm: apply conditional GIF saver support

tinyjin:
Add conditional compilation for the GIF saver functionality in the WASM binding. This allows proper WASM builds even when the GIF saver is disabled, preventing unnecessary binary size for preset variants.

related: https://github.com/thorvg/thorvg/issues/3481

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -438,6 +438,7 @@ public:
 
     bool save2Gif(string data)
     {
+#ifdef THORVG_GIF_SAVER_SUPPORT
         errorMsg = NoError;
 
         if (data.empty()) {
@@ -496,6 +497,10 @@ public:
         saver->sync();
 
         return true;
+#else
+        errorMsg = "GIF saver is not supported";
+        return false;
+#endif
     }
 
     // TODO: Advanced APIs wrt Interactivity & theme methods...


src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -540,6 +542,9 @@ EMSCRIPTEN_BINDINGS(thorvg_bindings)
 
     class_<TvgLottieAnimation>("TvgLottieAnimation")
         .constructor<string, string>()
+#ifdef THORVG_GIF_SAVER_SUPPORT
```
hermet: No point. we'd rather retain the interfaces for compatibility.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -429,6 +429,7 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
         updated = true;
     }
 
+#ifdef THORVG_GIF_SAVER_SUPPORT
```
hermet: Please consider this.
```
save2Gif(..)
{
#THORVG_GIF_SAVER_SUPPORT
   ...
#endif
   return false;
}


errorMsg = "GIF Save is not supported"

hermet (changes requested):

hermet (approved):

PR #3738, opened 2025-08-22
lottie/slot: hotfix multiple overriding case

tinyjin:
Fix incorrect overriding when reusing slots; use deep-copy for slot values to prevent unintended data clearing.

---

In the new slot API, when reusing a slot for overriding, a problem occurs due to shallow copying, causing properties not to override correctly.

Previously, each override created a new LottieProperty instance via LottieParser, which was then shallow-copied and overridden. This approach worked fine because the parsed property immediately lost ownership of its data—override() always operated on a fresh LottieProperty.

With the new design, we aim to reuse the same slot and property multiple times. Once a slot and its property are created, the slot’s data should remain valid even after overrides, and should only be freed when the LottieSlot is deleted or del() is called.

However, with this approach, applying the slot multiple times results in the property being overridden with an empty value starting from the second attempt.

---

### Examples

So basically when we reuse the slot after one proceeded, the result should remain the same.

1. Image Overriding

```cpp
const char* slotJson = R"({"path_img":{"p":{"id":"image_0","w":200,"h":300,"u":"images/","p":"logo.png","e":0}}})";
slotId3 = slot3->gen(slotJson);
if (!tvgexam::verify(slot3->apply(slotId3))) return false;
if (!tvgexam::verify(slot3->apply(0))) return false;
if (!tvgexam::verify(slot3->apply(slotId3))) return false; // Reuse the slot -> it should be the same as first `apply`
```

| Result | Expected |
|--------|----------|
| CleanShot 2025-08-22 at 16 10 32@2x  | CleanShot 2025-08-22 at 16 10 47@2x | 

2. Gradient Overriding

```cpp
const char* slotJson = R"({"gradient_fill":{"p":{"p":2,"k":{"k":[0,0.1,0.1,0.2,1,1,0.1,0.2,0,0,1,1]}}}})";
slotId1 = slot1->gen(slotJson);
if (!tvgexam::verify(slot1->apply(slotId1))) return false;
if (!tvgexam::verify(slot1->apply(0))) return false;
if (!tvgexam::verify(slot1->apply(slotId1))) return false;
```

| Result | Expected |
|--------|----------|
| CleanShot 2025-08-22 at 16 13 22@2x | CleanShot 2025-08-22 at 16 13 10@2x

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -192,10 +192,8 @@ void LottieSlot::reset()
 {
     if (!overridden) return;
 
-    auto shallow = pairs.count == 1 ? true : false;
-
     ARRAY_FOREACH(pair, pairs) {
-        pair->obj->override(pair->prop, shallow, true);
+        pair->obj->override(pair->prop, true);
         delete(pair->prop);
         pair->prop = nullptr;
     }
@@ -206,7 +204,6 @@ void LottieSlot::reset()
 void LottieSlot::apply(LottieProperty* prop, bool byDefault)
 {
     auto copy = !overridden && !byDefault;
-    auto shallow = pairs.count == 1 ? true : false;
 
     //apply slot object to all targets
     ARRAY_FOREACH(pair, pairs) {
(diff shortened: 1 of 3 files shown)


hermet:
@tinyjin you changed the colorstop much, Is it ok if we don't use shallow copy?

hermet:
@tinyjin could you please add this case to thorvg unit test?

tinyjin:
> @tinyjin could you please add this case to thorvg unit test?

That's great idea!

hermet:
> @tinyjin you changed the colorstop much, Is it ok if we don't use shallow copy?

If then we'd rather not using shallow copy (also remove that code around it)

tinyjin:
> If then we'd rather not using shallow copy (also remove that code around it)

Hello, shallow copy is still used LottieProperty's constructor and dash attribute.
I think we should not remove shallow copy in this reason.

Instead, I removed `shallow` parameter from `override` to clean this up. When calling that function, shallow-copy is no longer needed.

### 1. tvgLottieModel.h `:62`

CleanShot 2025-08-26 at 23 27 57@2x

### 2. tvgLottieProperty.h

CleanShot 2025-08-26 at 23 27 36@2x

tinyjin:
> @tinyjin could you please add this case to thorvg unit test?

Unit test already covered this scenario. This issue comes up in visual output, we still need manual testing to verify this.
So I didn't add new unit test, thank you.

CleanShot 2025-08-26 at 23 33 13@2x

hermet:
https://github.com/thorvg/thorvg/pull/3758

PR #3760, opened 2025-08-28
wasm: remove unused rpath parameter from load()

kunyoungparkk:
This patch removes the unused `rpath` parameter from the load() function.
This simplifies the WASM binding while keeping the API aligned with current usage in thorvg.web.

For context, my initial intent was to handle default parameters via overload, but since rpath is not used, I simplified the API as requested.

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -304,7 +304,7 @@ public:
         return animation->curFrame();
     }
 
-    bool load(string data, string mimetype, int width, int height, string rpath = "")
+    bool load(string data, string mimetype, int width, int height)
     {
         errorMsg = NoError;
 
@@ -326,7 +326,7 @@ public:
             filetype = "lottie+json";
         }
 
-        if (animation->picture()->load(data.c_str(), data.size(), filetype.c_str(), rpath.c_str(), false) != Result::Success) {
+        if (animation->picture()->load(data.c_str(), data.size(), filetype.c_str()) != Result::Success) {
             errorMsg = "load() fail";
             return false;
         }


src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -545,7 +550,8 @@ EMSCRIPTEN_BINDINGS(thorvg_bindings)
         .function("totalFrame", &TvgLottieAnimation ::totalFrame)
         .function("curFrame", &TvgLottieAnimation ::curFrame)
         .function("render", &TvgLottieAnimation::render)
-        .function("load", &TvgLottieAnimation ::load)
+        .function("load", emscripten::select_overload<bool(std::string, std::string, int, int)>(&TvgLottieAnimation::load))
+        .function("load", emscripten::select_overload<bool(std::string, std::string, int, int, std::string)>(&TvgLottieAnimation::load))
```
tinyjin: namespace `emscripten::` doesn't look necessary

kunyoungparkk:
I changed the code.
This code binds two actual C++ overloads using select_overload.
Embind automatically resolves which overload to call based on the number of arguments, so there’s no ambiguity in the JavaScript interface, and the generated .d.ts properly reflects both signatures.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
kunyoungparkk: Thank you! I have changed that.

kunyoungparkk:
@tinyjin 
First, I tested whether the overloaded load function calls work correctly in thorvg.web, and confirmed that they do without any issues.
However, I was also curious about how exactly this works internally, so I looked into it further.

According to the official Emscripten documentation, select_overload only works when the number of arguments differs. overloading based on just argument types is not supported. Still, the examples there don’t actually show binding multiple overloads under the same JavaScript function name, so I found it a bit ambiguous.
Screenshot 2025-08-29 at 10 46 01 AM


To clarify, I checked the emscripten/src/lib/libembind.js source, and verified that if multiple overloads are bound to the same JS function name, each overload gets registered in an overload table. At runtime, the dispatcher function uses this table to select the correct overload based  on arguments.length.
Screenshot 2025-08-29 at 10 17 26 AM
Screenshot 2025-08-29 at 10 44 25 AM

And, I found similar usage examples in the following links. Those references describe the issue that arises when trying to overload the same JS function with the same number of arguments but different types. In other words, conversely, for functions with a different number of arguments, they were actually leveraging overloading under the same JS function name.

[https://github.com/emscripten-core/emscripten/issues/20117](https://github.com/emscripten-core/emscripten/issues/20117)

[https://stackoverflow.com/questions/76911248/not-all-overloads-in-a-bound-c-class-are-available-in-javascript-using-emscri](https://stackoverflow.com/questions/76911248/not-all-overloads-in-a-bound-c-class-are-available-in-javascript-using-emscri)

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -305,7 +305,7 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
         return animation->curFrame();
     }
 
-    bool load(string data, string mimetype, int width, int height, string rpath = "")
+    bool load(string data, string mimetype, int width, int height, string rpath)
```
hermet: we might not need the rpath support. if there were no specific requirements, let's remove the `rpath`

tinyjin: @hermet I can confirm `thorvg.web` no longer uses `rpath`. Not sure whether another users use rpath with this binding. (I assume there might not be usecase using from our user end)

hermet: @kunyoungparkk @tinyjin please remove this, no need overload.

kunyoungparkk: thank you.
I’ll update this PR to remove the `rpath` parameter and clean up the overload accordingly.

Issue #3766, opened 2025-09-01
lottie/slot: unexpected crash/memory leak on overriding animated ColorStops


tinyjin:
## 1. Crash with overriding animated gradient

```cpp
const char* slotJson = R"({"gradient_fill":{"p":{"k":{"a":1,"k":[{"t":0,"s":[0.0,1.0,0.0,1.0,1.0,0.0,1.0,0.0,0.0,1.0,1.0,0.2],"i":{"x":0.2,"y":0.2},"o":{"x":0.8,"y":0.8}},{"t":30,"s":[0.0,1.0,1.0,0.0,1.0,0.0,0.0,1.0,0.0,1.0,1.0,0.5],"i":{"x":0.4,"y":0.4},"o":{"x":0.6,"y":0.6}},{"t":60,"s":[0.0,0.0,1.0,1.0,1.0,1.0,0.0,0.0,0.0,0.8,1.0,1.0],"i":{"x":0.3,"y":0.3},"o":{"x":0.7,"y":0.7}}]},"p":2}}})";
slotId1 = slot1->gen(slotJson);
if (!tvgexam::verify(slot1->apply(slotId1))) return false;
```

<img width="3010" height="1250" alt="Image" src="https://github.com/user-attachments/assets/71ba2b7b-63c3-43a0-97f5-c1480526e8ba" />

## 2. Memory Leak in animated gradient

<img width="2514" height="1434" alt="Image" src="https://github.com/user-attachments/assets/1d2321ec-fda4-4cc6-9859-ceb931557175" />

## 3. Crash with multiple overriding case with animated gradient

<img width="2260" height="1004" alt="Image" src="https://github.com/user-attachments/assets/2c68a9b1-9880-4867-bbee-47889fe1731a" />

tinyjin:
Closing as completed!

<img width="490" height="494" alt="Image" src="https://github.com/user-attachments/assets/f0123166-4e2e-4ce2-bc2c-bc9e6267d0bc" />

PR #3771, opened 2025-09-01
lottie/slot: handle colorStops input deep-copy properly

tinyjin:
There is a missing copy/free pair in frame handling.

Resolve potential memory leak and runtime crash caused by missing copy/free handling in animated value and in cases with multiple overriding.

| Memory Leak | Runtime Crash |
| --- | --- |
| image | image |

issue: https://github.com/thorvg/thorvg/issues/3766


hermet:
replaced with https://github.com/thorvg/thorvg/pull/3780

PR #3763, opened 2025-08-29
wasm: fix WG context registration issue with remove() timing

tinyjin:
Since the last change (a5be644b50415d37947adf41e3ee18b0c319e38a), the `canvas::remove()` function now sets the status to Painting.

CleanShot 2025-09-02 at 01 16 13@2x

When remove() function changes status to Painting mode, WebGPU instance and adapter may not be properly targeted by the target method if WG context is not ready.

To prevent incorrect state transitions, canvas.remove() is now only called when an existing animation is present, ensuring WG context is prepared (after resize) before any status changes that could affect context registration.

This resolves the issue where WG context registration fails due to premature status changes during animation lifecycle management.

PR #3773, opened 2025-09-02
renderer: revise canvas context handling in target method

tinyjin:
This update refines the management of the canvas context during initialization and operation. By improving context handling, it resolves a potential issue where the target method might not function correctly in certain cases.

related: #3763


hermet (changes requested):
1. please check copilot comment,  `||` is correct.
2. Please update the title — there is no deadlock issue in ThorVG itself. The deadlock is entirely due to the WASM binding, caused by a naughty implementation of context handling. You've now revised the concept of the canvas context, so the title should be updated accordingly.

PR #3775, opened 2025-09-02
renderer: supplement text boundingbox for layout

hermet:
Just missed this case, from the newly introduced text layout support.


hermet:
+ @BenjaminHalko

PR #3778, opened 2025-09-03
loader: Support `img/` tag for embedded image

hinohie:
Most of browser support "img/" tag for decode embedded image at svg.

Let we also allow to thorvg support it.

resvg didnt support "img/" tag also.
So, it will make thorvg is more powerful svg rendering engine rather than resvg

Example :
embedded_rgbywb_image
Image with "image/" (standard svg)

embedded_rgbywb_img
Imge with "img/" (not standard, but all of web browser support)


hinohie:
Browser support not only "img/" but also "asdf/" or "aaaaaaaaaaa/". It looks that they didn't check "image/" tag anyway.

We'd better found more good way to support it :(

PR #3784, opened 2025-09-04
lottie: repeater fine-tune with memory pooling

hermet:


hermet:
fps: ~2.3%+ at Lottie

Issue #3834, opened 2025-09-24
SVG gradients fail to render if fill url reference surrounded by quotes


lostminds:
This simple [Mozilla SVG gradient example](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/linearGradient) does not render at all in ThorVG (in the web viewer 1.0.0-pre28, as well as current embedded version in Godot 4.5):

```
<svg
  viewBox="0 0 10 10"
  xmlns="http://www.w3.org/2000/svg"
  xmlns:xlink="http://www.w3.org/1999/xlink">
  <defs>
    <linearGradient id="myGradient" gradientTransform="rotate(90)">
      <stop offset="5%" stop-color="gold" />
      <stop offset="95%" stop-color="red" />
    </linearGradient>
  </defs>

  <!-- using my linear gradient -->
  <circle cx="5" cy="5" r="4" fill="url('#myGradient')" />
</svg>
```

After a little experimenting and comparing with examples that did render correctly in ThorVG it seems it's simply the formatting of the fill url property. It appears (based on the Mozilla examples) that it's valid SVG syntax to define url references with quotes around them. Like in the example `fill="url('#myGradient')"` or `fill='url("#myGradient")'` however, in ThorVG it only renders the gradient fill if you set the url without the quotes inside the url() `fill="url(#myGradient)"`

If this is a url parsing issue it might be that this also affects other embedded symbols etc references by url, but I haven't tested that.

![Image](https://github.com/user-attachments/assets/2a44b04e-a961-4bd9-b99a-db80a08247a7)

hermet:
@lostminds @JSUYA Hello, the fix will be in the next release (v0.15.16 / v1.0-pre30). Thanks.

PR #3843, opened 2025-09-26
svg_loader: When trimming id url string, trim it with single quotes as well

JSUYA:
IdFromUrl starts with the # symbol.(Therefore, the single quote preceding # is ignored.)
Previously, only whitespace was trimmed. If the string ends with a single quote, it is trimmed as well.

related issue : https://github.com/thorvg/thorvg/issues/3834

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -367,8 +367,8 @@ static char* _idFromUrl(const char* url)
     ++open;
     --close;
 
-    //trim the rest of the spaces if any
-    while (open < close && *close == ' ') --close;
+    //trim the rest of the spaces and the quote marks if any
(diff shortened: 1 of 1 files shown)


lostminds:
Should probably trim double-quotes as well, as in `fill='url("#myGradient")'` which also seems to be accepted by other svg renderers if nested inside property value using single quotes.

hermet:
> Should probably trim double-quotes as well, as in `fill='url("#myGradient")'` which also seems to be accepted by other svg renderers if nested inside property value using single quotes.

@JSUYA any opinion?

JSUYA:
> Should probably trim double-quotes as well, as in `fill='url("#myGradient")'` which also seems to be accepted by other svg renderers if nested inside property value using single quotes.

Thank you for your feedback. I modified it to work with double-quotes too.

PR #3863, opened 2025-10-01
lottie: properly invert chained masking logic

tinyjin:
Improved handling of inverted root masks (InvAlpha, InvLuma). When the root mask is inverted, subsequent mask modes are now remapped (e.g., Subtract → Add, Add → Subtract).

This resolves the issue where only the first inverted mask was applied in masks in the chain.

file: [sample.json](https://github.com/user-attachments/files/22647078/sample.json)

| Before (Second masking didn’t apply) | After (Correctly chained masks) |
|--------------------------------------|---------------------------------|
| Before | After |

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -72,17 +72,19 @@ LottieEffect* LottieParser::getEffect(int type)
 }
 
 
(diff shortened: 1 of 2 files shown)


hermet:
This will break other cases; ThorVG needs a masking enhancement from the redesign.

PR #3878, opened 2025-10-11
Add random transform stress test to reproduce tvgSwRasterTexmap::_apply() dst overflow

WangSimiao2000:
This PR adds a **minimal reproducible example** that triggers a `dst` buffer overflow inside
`src/renderer/sw_engine/tvgSwRasterTexmap.h::_apply()` under certain transformation matrices.

The test continuously applies random transformations (**translate**, **scale**, **rotate**, **shear**)
to a picture each frame, which helps to expose corner cases in texture mapping and matrix-based rasterization.

When using specific affine transform matrices with large shear and rotation factors,
the `dst` pointer will overflow in `_apply()`.

Example crash matrix:

```
Matrix:
[ 0.572866, -4.431353, 336.605835 ]
[ 5.198910, -0.386219, 30.710693 ]
[ 0.000000, 0.000000, 1.000000 ]
```


hermet:
@WangSimiao2000 Thanks for your contribution, It would be nicer if you send a patch to `thorvg/test/testSwEngine.cpp`

WangSimiao2000:
@hermet 
create a new PR：
https://github.com/thorvg/thorvg/pull/3880

Issue #3881, opened 2025-10-13
API: runtime memory allocator customization


hermet:
Consider to support the next Interface for user custom memory allocator:

```   
//register a user memory allocators
struct Initializer::Allocator
{
        void* (*alloc)(size_t);
        void* (*calloc)(size_t, size_t);
        void* (*realloc)(size_t, size_t);
        void (*free)(void*);
};

//initialize
static Result Initializer::init(uint32_t threads = 0, const Allocator& allocator);
```

Why use this structure?

**Customization enables deeper optimization**
High-performance engines and renderers often replace malloc() with custom allocators (e.g., memory pools) to avoid bottlenecks. ThorVG can similarly benefit from tailored memory strategies for complex animations and real-time UI rendering.

**Platform-specific allocator optimization**
Different platforms benefit from different allocators (e.g., ashmem on Android, objc_alloc on iOS, arena allocators on embedded systems).

**Improved debugging and memory tracking**
Custom allocators allow precise tracking of memory usage, detecting leaks, and logging allocations for analysis.

Maybe 1–2 additional instructions are required per memory call, though they are likely negligible. A performance comparison before and after is needed to confirm.

high-red:
Since operator new/delete is currently overloaded globally, it affects other modules that are statically linked together with thorvg.
Please also consider removing the global overloading of operator new/delete.

huming2207:
I don't think it will? 

As per what I've done here: https://github.com/thorvg/thorvg/pull/3892

It will only override the the new/delete under thorvg namespace. I've tested it and it seems working for me. I might should test it again some time later. Right now I'm busy on some other higher priority projects.

high-red:
If I’m mistaken due to a lack of understanding, I apologize.
As far as I know, C++ does not provide a way to make the global operator new/delete apply only within a specific namespace.
This isn’t an issue with dynamic linking, but it becomes problematic with static linking.

Could it be that the environment where the issue is not occurring uses separate static linkage units for the application and ThorVG—in other words, they are dynamically linked?
If the application and the ThorVG library are statically linked, then the application's new/delete operators would also end up calling the global operator new/delete defined by the ThorVG library.

Although ThorVG targets embedded systems, dynamic linking is generally unavailable in embedded environments (especially bare-metal or RTOS systems), so global new/delete operator overload is a critical issue.

Makis42:
[...]

PR #3883, opened 2025-10-13
common: separate memory allocators

hermet:
Make the memory allocators more visible so that
users can easily replace them with a custom file.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -48,3 +48,4 @@ SoonGeon Noh <nors.nsg@gmail.com>
 Giseong Ji <jiggyjiggy0323@gmail.com>
 KunYoung Park <rjsdud3263@gmail.com>
 Wang SiMiao 王思淼 <wangsimiao1@xiaomi.com>
+Jackson Hu <huming2207@gmail.com>
--- a/src/common/tvgAllocator.h
+++ b/src/common/tvgAllocator.h
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2025 the ThorVG project. All rights reserved.
+
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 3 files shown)


huming2207:
Hi @hermet 

Thanks for this new PR. I'm not quite sure how are you going to let user to override the this `tvgAllocator.h`?  Did you forget to put in the meson config, or is there any other alternative meson stuff that allows me to override?

Jackson

hermet:
> Hi @hermet
> 
> Thanks for this new PR. I'm not quite sure how are you going to let user to override the this `tvgAllocator.h`? Did you forget to put in the meson config, or is there any other alternative meson stuff that allows me to override?
> 
> Jackson

@huming2207 Hi, file replacement is hacky approach. I don't think it's a proper solution for ThorVG. You know what it does, so you can replace the allocator file through your custom build script. This is what I can suggest for you now.

huming2207:
> You know what it does, so you can replace the allocator file through your custom build script. This is what I can suggest for you now. 

Hmm are you saying I should just patch the ThorVG's `tvgAllocator.h` manually for now?

I think the better way of doing this is to hold a singleton/global struct of alloc/free functions somewhere, as I've also seen you purposed here: https://github.com/thorvg/thorvg/issues/3881 , and allow user to override this struct. That's the safest way.

hermet:
> I think the better way of doing this is to hold a singleton/global struct of alloc/free functions somewhere, as I've also seen you purposed here: #3881 , and allow user to override this struct. That's the safest way.

@huming2207 Yes, you're right. That is the ideal approach for ThorVG. As a compromise solution, I suggest that you replace the allocator file. Since you are building ThorVG manually, I hope you can freely modify its build script on your environment.

huming2207:
> As a compromise solution, I suggest that you replace the allocator file. Since you are building ThorVG manually, I hope you can freely modify its build script on your environment.

Okay understood, I think it's fine for the short term anyway.

huming2207:
Meanwhile we've just tested some complex SVGs on ESP32-S3 that have embedded `<image>` tags with PNGs inside. This may often crash the microcontroller as the PNG decoder is still trying to allocate via the default `malloc()`. Therefore I think we need to let the 3rd-party decoders to use the customised allocator as well.

If you want, I will work on another PR for this matter, and I will try doing it properly with a global allocator struct. Because this is one of our major roadblocks...😅

hermet:
> Meanwhile we've just tested some complex SVGs on ESP32-S3 that have embedded `<image>` tags with PNGs inside. This may often crash the microcontroller as the PNG decoder is still trying to allocate via the default `malloc()`. Therefore I think we need to let the 3rd-party decoders to use the customised allocator as well.

Which PNG loader of ThorVG are you using? The static PNG loader (-Dstatic=true) should comply with ThorVG’s memory allocator. 

> If you want, I will work on another PR for this matter, and I will try doing it properly with a global allocator struct. Because this is one of our major roadblocks...

Thanks. ThorVG is currently focused on the v1 release, and I believe the memory allocator shouldn't be part of this target. While the review and merge might happen after v1, we're open to your contribution now. What I can suggest is, You can apply your solution in your own projects and submit a PR. Once it's merged upstream after v1, you can sync it.

huming2207:
> What I can suggest is, You can apply your solution in your own projects and submit a PR. Once it's merged upstream after v1, you can sync it.

Got it, I will submit a draft PR if I have spare time, and then merge back later after we both finish.

Issue #3604, opened 2025-07-08
wg/gl: shader optimization


hermet:
Shader code is loaded and interpreted at runtime. Neat, compact code is required for better performance and binary size. Even simple code clean up helps this:

- Remove unnecessary comments in the shaders.
- Make code shorter and compact.

hermet:
Find the best minifier tool, integrate it with thorvg build process. ie) https://www.ctrl-alt-test.fr/glsl-minifier/

PR #3886, opened 2025-10-14
gl_engine: revise shaders

SergeyLebedkin:
- remove unnecessary comments in the shaders
- shaders are separated into a separate entity
- shaders for blends and compositions are created only on the first request

https://github.com/thorvg/thorvg/issues/3604
https://github.com/thorvg/thorvg/issues/3698 - shaders


hermet:
@SergeyLebedkin You might wanna separate a PR with 3 commits:

- remove unnecessary comments in the shaders
- shaders are separated into a separate entity
- shaders for blends and compositions are created only on the first request

PR #3890, opened 2025-10-15
gl_engine: revise shader sources

SergeyLebedkin:
remove unnecessary comments in the shaders, replaced macro-based shader composition with cleaner R-string literals and regroup shader sources to make it more readable

https://github.com/thorvg/thorvg/issues/3604
https://github.com/thorvg/thorvg/issues/3698 - shaders

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -101,8 +101,8 @@ void GlRenderer::initShaders()
     mPrograms.reserve((int)RT_None);
 
 #if 1  //for optimization
-    #define LINEAR_TOTAL_LENGTH 2770
-    #define RADIAL_TOTAL_LENGTH 5272
+    #define LINEAR_TOTAL_LENGTH 8192
+    #define RADIAL_TOTAL_LENGTH 8192
     #define BLEND_TOTAL_LENGTH 8192
 #else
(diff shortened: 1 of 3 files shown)


wenjieshen (approved):
I recorded and compared the execution of all examples before and after this change, and merged the videos side by side for verification.
I didn’t observe any visible differences in rendering.
The theoretical memory increase is only about 9 KB, which is acceptable.
Approving the change. ✅
Screenshot 2025-10-19 at 12 22 38 AM

hermet:
@SergeyLebedkin This increases a binary size, approximately 1.3kb. Practically, this harms products... Do you intend this?

SergeyLebedkin:
@hermet @wenjieshen 

I've observed a regress in binary size after switching the shader definitions from the old TVG_COMPOSE_SHADER macro (which used stringification and line continuation) to C++11 Raw String Literals (R"(...)").

Observation and Root Cause
The compiled executable is noticeably larger.

Old Macro (#shader with \ and \n): The previous approach, while less readable, seems to have benefited from some compiler or build-system optimization that effectively stripped unnecessary characters (like the excess whitespace used for indentation in the C++ file) before embedding the string into the binary.

Raw String Literals (R"..."): This modern approach includes all whitespace and indentation used for formatting the shader within the C++ file (e.g., leading spaces on each line) as literal characters in the resulting string. Since these characters are not stripped, the string is longer, directly increasing the size of the data segment in the compiled file.

hermet:
@SergeyLebedkin Please feel free to reopen this, if it's valid. Thanks.

Issue #2884, opened 2024-10-19
gl_engine: binary shader support


hermet:
Compiling shaders every time is inefficient. Let's store the initially baked shaders in a designated ThorVG folder and reuse them. 

- Accessinhg the application's designated local folder is needed.
- Update binary shaders is required. (After checking the ThorVG version, it will be necessary to deprecate the existing binary format.)

hermet:
\+ @RuiwenTang

vtorri:
why not provide the binary shader directly ? you compile it before a release and add it in the tarball

hermet:
It's not compatible. Tvg needs to consider different vendors and systems working on.

hermet:
Reference (may not be accurate)
```
GLuint program = glCreateProgram();

// Attach shaders
glAttachShader(program, vertexShader);
glAttachShader(program, fragmentShader);

// Tell OpenGL to make the binary retrievable
glProgramParameteri(program, GL_PROGRAM_BINARY_RETRIEVABLE_HINT, GL_TRUE);

// Link the program
glLinkProgram(program);

// Check link status
GLint linked = GL_FALSE;
glGetProgramiv(program, GL_LINK_STATUS, &linked);
if (!linked) {
    // Handle error
}

// Get the binary
GLint length = 0;
glGetProgramiv(program, GL_PROGRAM_BINARY_LENGTH, &length);

GLenum format;
std::vector<GLubyte> binary(length);
glGetProgramBinary(program, length, nullptr, &format, binary.data());

// Save to file
FILE* file = fopen("shader.bin", "wb");
if (file) {
    fwrite(&format, sizeof(format), 1, file);  // Save format first
    fwrite(binary.data(), 1, binary.size(), file);
    fclose(file);
}
```

```
FILE* file = fopen("shader.bin", "rb");
if (file) {
    GLenum format;
    fread(&format, sizeof(format), 1, file);
    fseek(file, 0, SEEK_END);
    long size = ftell(file) - sizeof(format);
    fseek(file, sizeof(format), SEEK_SET);

    std::vector<GLubyte> binary(size);
    fread(binary.data(), 1, size, file);
    fclose(file);

    GLuint program = glCreateProgram();
    glProgramBinary(program, format, binary.data(), size);

    GLint success = 0;
    glGetProgramiv(program, GL_LINK_STATUS, &success);
    if (!success) {
        // Handle failure
    }
}
```

[...]

PR #3897, opened 2025-10-17
gl_engine/shader: Introduced binary shader caching

wenjieshen:
We've implemented a comprehensive binary shader caching system to significantly improve OpenGL rendering performance on macOS by eliminating redundant shader compilation overhead on subsequent application runs.

This feature introduces the `GlShaderCache` class that provides persistent storage and retrieval of compiled OpenGL shader binaries. The system automatically caches compiled shader programs to the user's cache directory using a hash-based naming scheme and attempts to load cached binaries before falling back to source compilation.

Key implementation details:
- Utilizes OpenGL binary shader format support via `glGetProgramBinary`/`glProgramBinary` extensions
- Stores cached binaries in the System cache folder on macOS and Windows; other platforms are testing
- Integrates with the existing `GlProgram` compilation pipeline

The caching mechanism works transparently:
1. When a shader program is first compiled, the binary is automatically cached
2. On subsequent runs, the system attempts to load the cached binary
3. If successful, compilation time is virtually eliminated
4. If cache miss or load failure occurs, falls back to normal source compilation

Performance benefits:
- Eliminates shader compilation stutters during runtime
- No API changes required - caching works automatically

issue: https://github.com/thorvg/thorvg/issues/2884


wenjieshen:
@hermet @SergeyLebedkin
Thanks for reviewing. I’d like to confirm two points:

Binary shader support is an OpenGL 4.1+ feature, while our requirement is OpenGL 3. I added a helper function glIsProgramBinarySupportAvailable() in tvgGl.h to check runtime availability. This is the first runtime capability check in the project, so I’m not sure if this approach aligns with our design principles.

Although my MacBook Air supports OpenGL 4.1, it still returns GL_NUM_PROGRAM_BINARY_FORMATS = 0, meaning binary shader support isn’t available. I’m waiting for GPU access on a cloud platform to test the performance improvement in another environment.

While waiting, I’ll continue investigating the stroke rendering issue on the GL engine.

hermet:
> Binary shader support is an OpenGL 4.1+ feature, while our requirement is OpenGL 3. I added a helper function glIsProgramBinarySupportAvailable() in tvgGl.h to check runtime availability. This is the first runtime capability check in the project, so I’m not sure if this approach aligns with our design principles.

@wenjieshen Looks good, if the binary shaders doesn't working, we can fallback to a normal scenario.

hermet:
@wenjieshen Could you please double-check this as well? a few systems might use this `GL_ARB_get_program_binary` for gl3 case. But please hold this, Maybe we can bump up the opengl dependency to v4 as the minimum requirement... If so, we can drop the v3 use-case.

huming2207:
> @huming2207 Could you please double-check this as well? a few systems might use this GL_ARB_get_program_binary for gl3 case.

@hermet  Sorry I'm not quite familiar with OpenGL, I'm a firmware guy and I know more on low-level stuff only 😅

wenjieshen:
> @wenjieshen Could you please hold this? I think we can discuss the opengl dependency further. Maybe we can bump up the opengl dependency to v4 as the minimum requirement... If so, we can drop the v3 use-case.

@hermet Understood, I’ll pause the PR.
Agree that we should discuss the OpenGL dependency direction first. If we decide to move to 4.0+, we can remove the v3 fallback logic entirely.

hermet:
> Understood, I’ll pause the PR.

@wenjieshen I mean we can hold support v3 for binary shaders. I didn't mean hold this PR. Please go proceed. :)  Thanks.

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -465,6 +465,7 @@ PFNGLGENVERTEXARRAYSPROC                     glGenVertexArrays;
 //GL_VERSION_3_1
 PFNGLGETUNIFORMBLOCKINDEXPROC      glGetUniformBlockIndex;
 PFNGLUNIFORMBLOCKBINDINGPROC       glUniformBlockBinding;
+
```
hermet: please don't change code without any specific purpose.

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -476,6 +477,13 @@ PFNGLUNIFORMBLOCKBINDINGPROC       glUniformBlockBinding;
 //PFNGLGETACTIVEUNIFORMBLOCKIVPROC   glGetActiveUniformBlockiv;
 //PFNGLGETACTIVEUNIFORMBLOCKNAMEPROC glGetActiveUniformBlockName;
 
+// GL_VERSION_4_1
+PFNGLGETPROGRAMBINARYPROC glGetProgramBinary;
+PFNGLPROGRAMBINARYPROC glProgramBinary;
+
+// Capability flag for program binary support
+static bool glProgramBinarySupportAvailable = false;
```
hermet: This is good to add around 154 Line.

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -820,6 +832,23 @@ bool glInit()
 
     TVGLOG("GL_ENGINE", "OpenGL/ES version = v%d.%d", vMajor, vMinor);
 
+    glProgramBinarySupportAvailable = false;
```
hermet: This initialization seems not necessary.

src/renderer/gl_engine/tvgGlProgram.cpp
```diff
@@ -21,24 +21,36 @@
...
 #include "tvgGlProgram.h"
+#include "tvgGlShaderCache.h"
 
 /************************************************************************/
 /* Internal Class Implementation                                        */
 /************************************************************************/
 
 uint32_t GlProgram::mCurrentProgram = 0;
 
-
```
hermet: --

src/renderer/gl_engine/tvgGlProgram.cpp
```diff
@@ -21,24 +21,36 @@
...
 /************************************************************************/
 /* External Class Implementation                                        */
 /************************************************************************/
 
-GlProgram::GlProgram(const char* vertSrc, const char* fragSrc)
+GlProgram::GlProgram(const char *vertSrc, const char *fragSrc)
```
hermet: Please respect thorvg coding convention

src/renderer/gl_engine/tvgGlProgram.cpp
```diff
@@ -21,24 +21,36 @@
...
 {
+    uint32_t progObj = 0;
+
+#ifndef __EMSCRIPTEN__
+    progObj = GlShaderCache::read(vertSrc, fragSrc);
+
+    if (0 != progObj)
```
hermet: convention:
```
if () {
}
```

src/renderer/gl_engine/tvgGlProgram.cpp
```diff
@@ -51,61 +63,61 @@ GlProgram::GlProgram(const char* vertSrc, const char* fragSrc)
...
         GLint infoLen = 0;
         glGetProgramiv(progObj, GL_INFO_LOG_LENGTH, &infoLen);
         if (infoLen > 0)
         {
-            auto infoLog = tvg::malloc<char*>(sizeof(char) * infoLen);
+            auto infoLog = tvg::malloc<char *>(sizeof(char) * infoLen);
             glGetProgramInfoLog(progObj, infoLen, NULL, infoLog);
             TVGERR("GL_ENGINE", "Error linking shader: %s", infoLog);
             tvg::free(infoLog);
-
```
hermet: --

src/renderer/gl_engine/tvgGlProgram.cpp
```diff
@@ -51,61 +63,61 @@ GlProgram::GlProgram(const char* vertSrc, const char* fragSrc)
...
         }
         glDeleteProgram(progObj);
         progObj = 0;
         assert(0);
     }
+
```
hermet: --

src/renderer/gl_engine/tvgGlProgram.cpp
```diff
@@ -51,61 +63,61 @@ GlProgram::GlProgram(const char* vertSrc, const char* fragSrc)
...
     mProgramObj = progObj;
+#ifndef __EMSCRIPTEN__
+    GlShaderCache::write(progObj, vertSrc, fragSrc);
+#endif
 }
 
-
```
hermet: please don't change other code here(72 - 171) if you have any specific purpose.

src/renderer/gl_engine/tvgGlProgram.cpp
```diff
@@ -51,61 +63,61 @@ GlProgram::GlProgram(const char* vertSrc, const char* fragSrc)
     GLint linked;
     glGetProgramiv(progObj, GL_LINK_STATUS, &linked);
 
-    if (!linked) {
+    if (!linked)
```
hermet: --

src/renderer/gl_engine/tvgGlShaderCache.h
```diff
@@ -0,0 +1,44 @@
...
+public:
+    static bool write(uint32_t programId, const char* vertSrc, const char* fragSrc);
+    static uint32_t read(const char* vertSrc, const char* fragSrc);
+
+private:
+    /**
+     * @brief Get cache path for shader files
+     * @return Static buffer - DO NOT FREE
+     */
+    static char* getCachePath(const char* vertSrc, const char* fragSrc);
```
hermet: please remove useless comments, you can use const return type.

suggest name:
```
`static const char* path(...)`
```

1. GlShaderCache says "Cache"  itself
2. we can figure out the purpose of the method easily with return type and parameters.

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,283 @@
...
+    return progObj;
+#else
+    return 0;
+#endif
+}
+
+bool GlShaderCache::write(uint32_t programId, const char *vertSrc, const char *fragSrc)
+{
+#if defined(THORVG_FILE_IO_SUPPORT) && !defined(__EMSCRIPTEN__)
+    if (!programId || !vertSrc || !fragSrc)
```
hermet: you can use one line for simple logic 
```
    if (!programId || !vertSrc || !fragSrc) return false;
```

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,283 @@
...
+#else
+    #include <dirent.h>
+    #include <unistd.h>
+    #include <limits.h>
+    #include <sys/stat.h>
+#endif
+
+static char cachePath[PATH_MAX];
+
+uint32_t GlShaderCache::hashString(const char *str)
```
hermet: maybe you can use `unsigned long djb2Encode(const char* str); `  in `tvgCompressor.h`

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,283 @@
...
+    // Allocate buffer for binary data
+    auto binaryData = tvg::malloc<GLubyte *>(length);
+    if (!binaryData)
+    {
+        fclose(file);
+        return 0;
+    }
+
+    // Read binary data
+    size_t bytesRead = fread(binaryData, 1, length, file);
```
hermet: please use `auto` if possible.

hermet (commented):
@wenjieshen Please check comments and please respect thorvg coding convention.
Thanks.

src/renderer/gl_engine/tvgGlShaderCache.h
```diff
@@ -0,0 +1,40 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef _TVG_GL_SHADER_CACHE_H_
+#define _TVG_GL_SHADER_CACHE_H_
+
+#include "tvgGlCommon.h"
+
```
hermet, line 28: Suggestion: 
```
struct GlShaderCache
{
    static bool write(uint32_t programId, const char* vertSrc, const char* fragSrc);
    static uint32_t read(const char* vertSrc, const char* fragSrc);
    static const char* path(const char* vertSrc, const char* fragSrc);
    static uint32_t hashString(const char* str);
}
```

This is gl internal feature. Visibility isn't necessary.

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -152,6 +152,8 @@ static void* _getProcAddress(const char* procName) {
         return false;                                           \
     }
 
+// Capability flag for program binary support
+static bool glProgramBinarySupportAvailable = false;
```
hermet: Simple is better.
`_glProgramBinarySupport = false;`

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -820,6 +830,22 @@ bool glInit()
 
     TVGLOG("GL_ENGINE", "OpenGL/ES version = v%d.%d", vMajor, vMinor);
 
+#if defined(THORVG_GL_TARGET_GLES)
+    // OpenGL ES 3.0+ has these functions in core
```
hermet: please remove helpfuless comments

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -820,6 +830,22 @@ bool glInit()
...
+    if (vMajor >= 3) {
+        GL_FUNCTION_FETCH(glGetProgramBinary, PFNGLGETPROGRAMBINARYPROC);
+        GL_FUNCTION_FETCH(glProgramBinary, PFNGLPROGRAMBINARYPROC);
+        glProgramBinarySupportAvailable = true;
+    }
+#else
+    // OpenGL 4.1+ has these functions in core
```
hermet: please remove helpfuless comments

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -835,4 +861,9 @@ bool glTerm()
     return true;
 }
 
+bool glIsProgramBinarySupportAvailable()
```
hermet: `bool glProgramBinarySupport()`

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "tvgGlShaderCache.h"
+
+#ifdef _WIN32
```
hermet, line 28: Please indentation+
```
#ifdef _WIN32
    #define WIN32_LEAN_AND_MEAN
        #include <windows.h>
        #ifndef PATH_MAX
             #define PATH_MAX MAX_PATH
        #endif
#else
     #include <dirent.h>
     ...
#endif
```

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+#else
+#include <dirent.h>
+#include <limits.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#endif
+
+static char cachePath[PATH_MAX];
+
+uint32_t GlShaderCache::hashString(const char* str)
```
hermet: can't use `djb2Encode()` ?

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+#endif
+}
+
+/************************************************************************/
+/* External Class Implementation                                        */
+/************************************************************************/
+
+uint32_t GlShaderCache::read(const char* vertSrc, const char* fragSrc)
+{
+#if defined(THORVG_FILE_IO_SUPPORT) && !defined(__EMSCRIPTEN__)
```
hermet, line 169: --unused parameters warnings
Suggestion:
```
    if (!vertSrc || !fragSrc) return 0;

#if defined(THORVG_FILE_IO_SUPPORT) && !defined(__EMSCRIPTEN__)
```

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+    if (!vertSrc || !fragSrc) return 0;
+
+    // Check if program binary support is available at runtime
```
hermet: --unhelpful comments

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+    if (!glIsProgramBinarySupportAvailable()) {
+        TVGLOG("GL_ENGINE", "Program binary support not available, skipping cache read");
+        return 0;
+    }
+
+    // Get cache file path
+    auto cachePath = path(vertSrc, fragSrc);
+    if (!cachePath) return 0;
+
+    // Check if cache file exists
```
hermet: --unhelpful comments

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+    auto bytesRead = fread(binaryData, 1, length, file);
+    fclose(file);
+
+    if (bytesRead != static_cast<size_t>(length)) {
+        TVGLOG("GL_ENGINE", "Failed to read shader cache data: %s", cachePath);
+        tvg::free(binaryData);
+        return 0;
+    }
+
+    // Create program object
```
hermet: --unhelpful comments

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    // Allocate buffer for binary data
+    auto binaryData = tvg::malloc<GLubyte*>(length);
+    if (!binaryData) {
+        fclose(file);
+        return 0;
+    }
+
+    // Read binary data
```
hermet: --unhelpful comments

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+    GLsizei length = 0;
+
+    if (fread(&binaryFormat, sizeof(GLenum), 1, file) != 1 || fread(&length, sizeof(GLsizei), 1, file) != 1 ||
+        length < 1) {
+        TVGLOG("GL_ENGINE", "Failed to read shader cache header: %s", cachePath);
+        fclose(file);
+        return 0;
+    }
+
+    // Allocate buffer for binary data
```
hermet: --unhelpful comments

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+    // Write format and length first
+    fwrite(&binaryFormat, sizeof(GLenum), 1, file);
+    fwrite(&length, sizeof(GLsizei), 1, file);
+
+    // Write binary data
+    size_t written = fwrite(binaryData, 1, length, file);
+    fclose(file);
+
+    bool success = (written == static_cast<size_t>(length));
```
hermet: bool -> auto

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+    if (success) {
```
hermet:
```
    if (success) TVGLOG("GL_ENGINE", "Shader cache written: %s (%d bytes)", cachePath, length);
    else TVGLOG("GL_ENGINE", "Failed to write shader cache: %s", cachePath);
```

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+    GLint binaryLength = 0;
+    glGetProgramiv(programId, GL_PROGRAM_BINARY_LENGTH, &binaryLength);
+    if (binaryLength < 1) {
+        TVGLOG("GL_ENGINE", "Failed to get program binary length");
+        return false;
+    }
+
+    // Allocate buffer for binary data
+    auto binaryData = tvg::malloc<GLubyte*>(binaryLength);
+    if (!binaryData) {
```
hermet:
```
    if (!binaryData) return false;
```

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+        TVGLOG("GL_ENGINE", "Failed to open cache file for writing: %s", cachePath);
+        tvg::free(binaryData);
+        return false;
+    }
+
+    // Write format and length first
+    fwrite(&binaryFormat, sizeof(GLenum), 1, file);
+    fwrite(&length, sizeof(GLsizei), 1, file);
+
+    // Write binary data
```
hermet: -- unecessary comment

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+    // Write to file
+    FILE* file = fopen(cachePath, "wb");
+    if (!file) {
+        TVGLOG("GL_ENGINE", "Failed to open cache file for writing: %s", cachePath);
+        tvg::free(binaryData);
+        return false;
+    }
+
+    // Write format and length first
```
hermet: -- unnessary comment

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+    return 0;
+#endif
+}
+
+bool GlShaderCache::write(uint32_t programId, const char* vertSrc, const char* fragSrc)
+{
+#if defined(THORVG_FILE_IO_SUPPORT) && !defined(__EMSCRIPTEN__)
+    if (!programId || !vertSrc || !fragSrc) return false;
+
+    // Check if program binary support is available at runtime
```
hermet: -- unnessary comment

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+    if (!glIsProgramBinarySupportAvailable()) {
+        TVGLOG("GL_ENGINE", "Program binary support not available");
+        return false;
+    }
+
+    // Get cache file path
+    auto cachePath = path(vertSrc, fragSrc);
+    if (!cachePath) return false;
+
+    // Check if GL_ARB_get_program_binary is supported
```
hermet: //TODO: Check if GL_ARB_get_program_binary is supported

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,260 @@
...
+    GLint numFormats = 0;
+    glGetIntegerv(GL_NUM_PROGRAM_BINARY_FORMATS, &numFormats);
+    if (numFormats < 1) {
+        TVGLOG("GL_ENGINE", "Program binary not supported");
+        return false;
+    }
+
+    // Get program binary length
```
hermet: -- unnessary comment

hermet (commented):
Most your comments are unnecesary. Don't need to add comments for saying code itself.
We'd rather leave comments when it needs a unusual, special context and purposes.

src/renderer/gl_engine/tvgGlShaderCache.h
```diff
@@ -0,0 +1,38 @@
...
+struct GlShaderCache
+{
+    static bool write(uint32_t programId, const char* vertSrc, const char* fragSrc);
+    static uint32_t read(const char* vertSrc, const char* fragSrc);
+
+    static const char* path(const char* vertSrc, const char* fragSrc);
+    static uint32_t hashString(const char* str);
```
hermet: -> hash()

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,244 @@
...
+    GLsizei length = 0;
+    glGetProgramBinary(programId, binaryLength, &length, &binaryFormat, binaryData);
+
+    if (length < 1) {
+        TVGLOG("GL_ENGINE", "Failed to retrieve program binary");
+        tvg::free(binaryData);
+        return false;
+    }
+
+    FILE* file = fopen(cachePath, "wb");
```
hermet: auto

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -152,6 +152,8 @@ static void* _getProcAddress(const char* procName) {
         return false;                                           \
     }
 
+// Capability flag for program binary support
```
hermet: variable name itself already descriptive.
please remove this.

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -809,6 +815,10 @@ bool glInit()
     // GL_FUNCTION_FETCH(glGetActiveUniformBlockName, PFNGLGETACTIVEUNIFORMBLOCKNAMEPROC);
     GL_FUNCTION_FETCH(glUniformBlockBinding, PFNGLUNIFORMBLOCKBINDINGPROC);
 
+    // GL_ARB_get_program_binary (optional, may not be available on all platforms)
```
hermet: //TODO: GL_ARB_get_program_binary  ...

src/renderer/gl_engine/tvgGl.h
```diff
@@ -1490,6 +1500,13 @@
     //extern PFNGLGETACTIVEUNIFORMNAMEPROC      glGetActiveUniformName;
     //extern PFNGLGETACTIVEUNIFORMBLOCKIVPROC   glGetActiveUniformBlockiv;
     //extern PFNGLGETACTIVEUNIFORMBLOCKNAMEPROC glGetActiveUniformBlockName;
+
+    // GL_VERSION_4_1
+    extern PFNGLGETPROGRAMBINARYPROC          glGetProgramBinary;
+    extern PFNGLPROGRAMBINARYPROC             glProgramBinary;
+
+    // Check if program binary support is available at runtime
```
hermet: --

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,244 @@
...
+    TVGLOG("GL_ENGINE", "Shader cache loaded: %s (%d bytes)", cachePath, length);
+    return progObj;
+#else
+    return 0;
+#endif
+}
+
+bool GlShaderCache::write(uint32_t programId, const char* vertSrc, const char* fragSrc)
+{
+    if (!programId || !vertSrc || !fragSrc) return false;
```
hermet: --Unneccesary conditions
 (!vertSrc || !fragSrc)

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+#if defined(THORVG_FILE_IO_SUPPORT) && !defined(__EMSCRIPTEN__)
+
+    if (!glProgramBinarySupport()) {
+        TVGLOG("GL_ENGINE", "Program binary support not available");
```
hermet: `if (!glProgramBinarySupport()) return;`

you can rather add this log in `bool glInit()` (~847)

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,244 @@
...
+/* External Class Implementation                                        */
+/************************************************************************/
+
+uint32_t GlShaderCache::read(const char* vertSrc, const char* fragSrc)
+{
+    if (!vertSrc || !fragSrc) return 0;
+
+#if defined(THORVG_FILE_IO_SUPPORT) && !defined(__EMSCRIPTEN__)
+
+    if (!glProgramBinarySupport()) {
```
hermet: `if (!glProgramBinarySupport()) return;`

you can rather add this log in `bool glInit()` (~847)

src/renderer/gl_engine/tvgGlProgram.cpp
```diff
@@ -35,10 +36,21 @@ uint32_t GlProgram::mCurrentProgram = 0;
 
 GlProgram::GlProgram(const char* vertSrc, const char* fragSrc)
 {
+    uint32_t progObj = 0;
+
+#ifndef __EMSCRIPTEN__
+    progObj = GlShaderCache::read(vertSrc, fragSrc);
```
hermet: Simpler code
```
mProgObj = GlShaderCache::read(vertSrc, fragSrc);
if (mProgObj != 0) return;
```

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    TVGLOG("GL_ENGINE", "Shader cache loaded: %s (%d bytes)", cachePath, length);
+    return progObj;
+#else
+    return 0;
+#endif
+}
+
+bool GlShaderCache::write(uint32_t programId, const char* vertSrc, const char* fragSrc)
```
hermet: -> void. 
return value is not used anywhere.

src/renderer/gl_engine/tvgGlShaderCache.h
```diff
@@ -0,0 +1,38 @@
...
+#ifndef _TVG_GL_SHADER_CACHE_H_
+#define _TVG_GL_SHADER_CACHE_H_
+
+#include "tvgGlCommon.h"
+
+struct GlShaderCache
+{
+    static bool write(uint32_t programId, const char* vertSrc, const char* fragSrc);
+    static uint32_t read(const char* vertSrc, const char* fragSrc);
+
```
hermet: --

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,244 @@
...
+    if (!programId || !vertSrc || !fragSrc) return false;
+
+#if defined(THORVG_FILE_IO_SUPPORT) && !defined(__EMSCRIPTEN__)
+
+    if (!glProgramBinarySupport()) {
+        TVGLOG("GL_ENGINE", "Program binary support not available");
+        return false;
+    }
+
+    // Check if cache file exists and if the program binary is fully supported
```
hermet: good to relocate to ~ 220 line.

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,244 @@
...
+    // Compute hash from shader sources
+    uint32_t vertHash = hashString(vertSrc);
+    uint32_t fragHash = hashString(fragSrc);
+    uint32_t combinedHash = vertHash ^ (fragHash << 1);
+
+    char cacheDir[PATH_MAX];
+    const char* home = getenv("HOME");
+    if (!home) return nullptr;
+
+    #ifdef _WIN32
```
hermet: Please apply proper indentation. 
```
#ifdef
    ---
```

src/renderer/gl_engine/tvgGlShaderCache.cpp
wenjieshen: Thank you. I am setting up environment to ensure if it works in this case.

src/renderer/gl_engine/tvgGlShaderCache.cpp
wenjieshen: Thanks! I will put it into tvgGl.cpp.

src/renderer/gl_engine/tvgGlShaderCache.cpp
wenjieshen: Thanks, I’ll fix the indentation. 
One clarification: this block is already inside 
`#if defined(THORVG_FILE_IO_SUPPORT) && !defined(__EMSCRIPTEN__)`.
Should we indent mixed nested preprocessor directives too, or only the full macro code snippet per our style?

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,243 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "tvgGlShaderCache.h"
+
+#ifdef _WIN32
+    #define WIN32_LEAN_AND_MEAN
```
SergeyLebedkin: we should remove this line for windows or at least check for predefinisions
image

SergeyLebedkin (commented):
see comments

src/renderer/gl_engine/tvgGlShaderCache.cpp
hermet: @wenjieshen Please choose case 1 or 3 style. It's up to you.

Case 1 (O)
```
#if
    ....
#if   //nested
    ....
#endif
   ....
#endif
```

Case 2 (X)
```
#if
#if    //nested
    ....
#endif 
#endif
```


Case 3 (O)
```
#if
    #if    //nested
        ....
    #endif 
#endif
```

Case 4 (X)
```
#if
    #if    //nested
    ....
    #endif 
#endif
```

wenjieshen: Got it. This issue will be addressed when the Windows support is merged into the upcoming version.

hermet:
image

@wenjieshen Please check your commit message. I marked some inappropriate msg (recommend to remove), Thanks.
+ Add relative issue info in commit msg such as:
`issue: https://github.com/thorvg/thorvg/issues/2884`

src/renderer/gl_engine/tvgGlShaderCache.cpp
wenjieshen: After testing, djb won't generate a collision in our circumstances. I have applied djb2Encode in the latest version.

wenjieshen:
@hermet Thank you. I have updated the commit message.

wenjieshen:
# Shader Binary‐Cache Experiment Results

## Overview

This analysis examines shader creation and caching performance across 30 repeated executions, each compiling 16 shaders.  
Four distinct timing datasets were analyzed:

- **NVIDIA Binary Shader Cache (available):** Shader creation when the driver’s built-in binary shader cache is active.  
- **Without Binary Shader Cache:** Shader creation times when the driver cache is unavailable or disabled.  
- **Cache read:** Loading precompiled shaders from a custom in-house cache system.  
- **Cache write:** Writing shader binaries to that same custom cache.

## About the NVIDIA Auto Binary Shader Cache

The **NVIDIA Auto Binary Shader Cache** stores compiled shader binaries on disk so they can be reused across application runs instead of being recompiled from source each time.  
This cache is **automatically managed by NVIDIA drivers** and is **available by default** on most driver versions. However, the exact behavior and availability may differ depending on the **GPU vendor** and **driver version**.  

When present, the driver transparently saves and restores shader binaries from its cache directory, substantially reducing load times for subsequent runs.  

However, it's still recommended to maintain an **application-level shader cache**.  
Driver-managed caches can be unavailable, disabled, or behave inconsistently across different platforms, driver versions, or GPU vendors.  
Having your own cache ensures consistent performance and reliability regardless of driver behavior.

> **Reference:** [NVIDIA DRIVE OS 5.2 Linux SDK Development Guide – Binary Shaders](https://docs.nvidia.com/drive/drive-os-5.2.0.0L/drive-os/DRIVE_OS_Linux_SDK_Development_Guide/Graphics/graphics_binary_shaders.html#wwpID0EOHA)

## Summary Statistics

| Metric | Driver Cache | No Cache | Our Own Cache Read | Our Own Cache Write |
| --- | --- | --- | --- | --- |
| Mean (µs) | **585** | **10,952** | 198 | 102 |
| Median (µs) | 116 | 4,454 | 126 | 58 |
| 90th Percentile (µs) | 475 | 41,797 | 484 | 110 |
| 95th Percentile (µs) | 788 | 51,207 | 728 | 137 |
| 99th Percentile (µs) | 5,868 | 51,978 | 929 | 980 |
| Maximum (µs) | 50,400 | 57,287 | 1,102 | 1,383 |
| Minimum (µs) | 87 | 3,159 | 83 | 35 |

## Comparative Results

### 1. Shader Compilation Time
- The **Binary Shader Cache** reduces mean shader compilation time from roughly **10.95 ms to 0.59 ms**, achieving an average **54× speed-up** across runs.
- The median drop is even more dramatic, from 4.45 ms to 0.12 ms — roughly **37× faster**.

### 2. Cache Read vs Write Performance
- Reading shaders from the in-house cache averages **198 µs**, about **3× faster** than writing them (102 µs).
 **Note:** The write process is performed **only during the first run**, as it stores new shader binaries for reuse. Subsequent executions only perform read operations.

### 3. Distribution & Variability
- Without the Binary Shader Cache, shader creation exhibits long-tail latency, with outliers exceeding **50 ms**.
- With caching available, almost all shader compilations finish under **1 ms**, except during first-time runs.
- In-house cache read/write times remain consistently under **200 µs**.

---

## Visualizations

**Average per-run shader creation time comparison**  
per_run_average_comparison

**Distribution of all operation times**  
per_shader_average_comparison

**Speed-up ratio (no-cache ÷ cache)**  
distribution_boxplot

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include "tvgGl.h"
```
hermet, line 23:
```
#include <cstring>
#include "tvgGl.h"
#include "tvgCommon.h"
```

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -152,6 +153,22 @@ static void* _getProcAddress(const char* procName) {
...
 
+static bool _glProgramBinarySupport = false;
+
+static bool _isExtensionSupported(const char* extension)
+{
+    GLint count = 0;
+    glGetIntegerv(GL_NUM_EXTENSIONS, &count);
+
+    for (GLint i = 0; i < count; ++i) {
+        const char* ext = (const char*)glGetStringi(GL_EXTENSIONS, i);
```
hermet: `auto ext`

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -152,6 +153,22 @@ static void* _getProcAddress(const char* procName) {
...
+        if (ext && strcmp(ext, extension) == 0) {
```
hermet: one line:
`if (ext && !strcmp(ext, extension)) return true;`

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -820,6 +841,25 @@ bool glInit()
...
+    // For desktop OpenGL, check if extension GL_ARB_get_program_binary is available
+    // This extension GL_ARB_get_program_binary is available in OpenGL 3.0+ but we need to check for it explicitly
+    if ((vMajor > 4) || (vMajor == 4 && vMinor >= 1)) {
+        _glProgramBinarySupport = true;
+        TVGLOG("GL_ENGINE", "Program binary support enabled (OpenGL 4.1+ core)");
+    } else if (vMajor >= 3 && _isExtensionSupported("GL_ARB_get_program_binary")) {
+        _glProgramBinarySupport = true;
+        TVGLOG("GL_ENGINE", "Program binary support enabled (GL_ARB_get_program_binary extension)");
+    } else {
+        TVGLOG("GL_ENGINE", "Program binary support not available");
```
hermet, line 910: Not necessary, Let's remove this case.

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -820,6 +841,25 @@ bool glInit()
 
     TVGLOG("GL_ENGINE", "OpenGL/ES version = v%d.%d", vMajor, vMinor);
 
+#if !defined(THORVG_GL_TARGET_GLES)
+    // For desktop OpenGL, check if extension GL_ARB_get_program_binary is available
+    // This extension GL_ARB_get_program_binary is available in OpenGL 3.0+ but we need to check for it explicitly
+    if ((vMajor > 4) || (vMajor == 4 && vMinor >= 1)) {
+        _glProgramBinarySupport = true;
+        TVGLOG("GL_ENGINE", "Program binary support enabled (OpenGL 4.1+ core)");
+    } else if (vMajor >= 3 && _isExtensionSupported("GL_ARB_get_program_binary")) {
```
hermet: Please add comments to a relavent logic:
```
    // For desktop OpenGL, check if extension GL_ARB_get_program_binary is available
    // This extension GL_ARB_get_program_binary is available in OpenGL 3.0+ but we need to check for it explicitly
    } else if (.......) {
```

src/renderer/gl_engine/tvgGl.h
```diff
@@ -1490,6 +1499,13 @@
     //extern PFNGLGETACTIVEUNIFORMNAMEPROC      glGetActiveUniformName;
     //extern PFNGLGETACTIVEUNIFORMBLOCKIVPROC   glGetActiveUniformBlockiv;
     //extern PFNGLGETACTIVEUNIFORMBLOCKNAMEPROC glGetActiveUniformBlockName;
+
+    //GL_VERSION_4_1
+    //GL_VERSION_3_0 Extension: GL_ARB_get_program_binary
```
hermet: --1504

src/renderer/gl_engine/tvgGlProgram.cpp
```diff
@@ -35,10 +36,19 @@ uint32_t GlProgram::mCurrentProgram = 0;
 
 GlProgram::GlProgram(const char* vertSrc, const char* fragSrc)
 {
+
+#ifndef __EMSCRIPTEN__
+    mProgramObj = GlShaderCache::read(vertSrc, fragSrc);
+
+    if (0 != mProgramObj) {
```
hermet: `if (mProgramObj) return;`

src/renderer/gl_engine/tvgGlShaderCache.h
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+#ifndef _TVG_GL_SHADER_CACHE_H_
+#define _TVG_GL_SHADER_CACHE_H_
+
+#include "tvgGlCommon.h"
+
+struct GlShaderCache
+{
+    static void write(uint32_t programId, const char* vertSrc, const char* fragSrc);
```
hermet: Whatever name is, Please make a consistent naming. you used `progObj`, `programId` ... 

image

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+    // Build full cache file path
+#ifdef _WIN32
+    snprintf(cachePath, PATH_MAX, "%s\\shader_%08lx.bin", cacheDir, combinedHash);
+#else
+    snprintf(cachePath, PATH_MAX, "%s/shader_%08lx.bin", cacheDir, combinedHash);
+#endif
+
+    return cachePath;
+}
+
```
hermet, line 84: convention: 2 empty lines between functions

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+uint32_t GlShaderCache::read(const char* vertSrc, const char* fragSrc)
+{
+
+#if !defined(THORVG_FILE_IO_SUPPORT) || defined(__EMSCRIPTEN__)
+    return 0;
+#endif
+
+    if (!vertSrc || !fragSrc) return 0;
+
+    if (!glProgramBinarySupport()) {
```
hermet: `if (!glProgramBinarySupport()) return 0;`

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+    auto cachePath = path(vertSrc, fragSrc);
+    if (!cachePath) return 0;
+
+    auto file = fopen(cachePath, "rb");
+    if (!file) return 0;
+
+    GLenum binaryFormat = 0;
+    GLsizei length = 0;
+
+    if (fread(&binaryFormat, sizeof(GLenum), 1, file) != 1 || fread(&length, sizeof(GLsizei), 1, file) != 1 ||
```
hermet: one line 
`if (fread(&binaryFormat, sizeof(GLenum), 1, file) != 1 || fread(&length, sizeof(GLsizei), 1, file) != 1 || length < 1) {`

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+        length < 1) {
+        TVGLOG("GL_ENGINE", "Failed to read shader cache header: %s", cachePath);
+        fclose(file);
+        return 0;
+    }
+
+    auto binaryData = tvg::malloc<GLubyte*>(length);
+    if (!binaryData) {
```
hermet: Please remove memory exceptions.

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+    if (bytesRead != static_cast<size_t>(length)) {
+        TVGLOG("GL_ENGINE", "Failed to read shader cache data: %s", cachePath);
+        tvg::free(binaryData);
+        return 0;
+    }
+
+    // Load program binary to shader
+    uint32_t progObj = glCreateProgram();
+    if (!progObj) {
+        TVGLOG("GL_ENGINE", "Failed to load cached shader program object");
```
hermet: Not really concern case, we can't do anything. Please remove log.

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    glProgramBinary(progObj, binaryFormat, binaryData, length);
+    tvg::free(binaryData);
+
+    GLint linked = 0;
+    glGetProgramiv(progObj, GL_LINK_STATUS, &linked);
+
+    if (!linked) {
+        TVGLOG("GL_ENGINE", "Failed to link cached shader program: %s", cachePath);
```
hermet: Not really concern case, we can't do anything. Please remove log.

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+void GlShaderCache::write(uint32_t programId, const char* vertSrc, const char* fragSrc)
+{
+
+#if !defined(THORVG_FILE_IO_SUPPORT) || defined(__EMSCRIPTEN__)
+    return;
+#endif
+
+    if (!programId || !vertSrc || !fragSrc) return;
+
+    if (!glProgramBinarySupport()) {
```
hermet: one line please

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+    GLint binaryLength = 0;
+    glGetProgramiv(programId, GL_PROGRAM_BINARY_LENGTH, &binaryLength);
+    if (binaryLength < 1) {
+        TVGLOG("GL_ENGINE", "Failed to get program binary length");
+        return;
+    }
+
+    auto binaryData = tvg::malloc<GLubyte*>(binaryLength);
+    if (!binaryData) {
```
hermet: --remove exception

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+    glGetProgramBinary(programId, binaryLength, &length, &binaryFormat, binaryData);
+
+    if (length < 1) {
+        TVGLOG("GL_ENGINE", "Failed to retrieve program binary");
+        tvg::free(binaryData);
+        return;
+    }
+
+    // Check if cache file can be written
+    FILE* file = fopen(cachePath, "wb");
```
hermet: `auto`

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+        return;
+    }
+
+    fwrite(&binaryFormat, sizeof(GLenum), 1, file);
+    fwrite(&length, sizeof(GLsizei), 1, file);
+
+    auto written = fwrite(binaryData, 1, length, file);
+    fclose(file);
+
+    auto success = (written == static_cast<size_t>(length));
```
hermet: if (written == static_cast<size_t>(length)) LOG()

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+    if (success) TVGLOG("GL_ENGINE", "Shader cache written: %s (%d bytes)", cachePath, length);
+    else TVGLOG("GL_ENGINE", "Failed to write shader cache: %s", cachePath);
```
hermet: --

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,271 @@
...
+        if (!appData) return nullptr;
+    }
+    snprintf(cacheDir, PATH_MAX, "%s\\thorvg\\shaders", appData);
+
+    auto attribs = GetFileAttributesA(cacheDir);
+    if (attribs == INVALID_FILE_ATTRIBUTES) {
+        char thorvgDir[PATH_MAX];
+        snprintf(thorvgDir, PATH_MAX, "%s\\thorvg", appData);
+
+        if (!CreateDirectoryA(thorvgDir, NULL) && GetLastError() != ERROR_ALREADY_EXISTS) {
```
hermet: one line.

hermet (commented):
@wenjieshen Thanks, Please check more nit-picks.  
+ Please note that ThorVG don't handles memory alloc fails.

hermet:
@wenjieshen 
```
[35/376] Compiling C++ object src/libthorvg.so.1.0.0.p/renderer_gl_engine_tvgGlShaderCache.cpp.o
../src/renderer/gl_engine/tvgGlShaderCache.cpp: In static member function ‘static const char* GlShaderCache::path(const char*, const char*)’:
../src/renderer/gl_engine/tvgGlShaderCache.cpp:131:38: warning: ‘/shader_’ directive output may be truncated writing 8 bytes into a region of size between 1 and 4096 [-Wformat-truncation=]
  131 |     snprintf(cachePath, PATH_MAX, "%s/shader_%08lx.bin", cacheDir, combinedHash);
      |                                      ^~~~~~~~
In file included from /usr/include/stdio.h:980,
                 from /usr/include/c++/13/cstdio:42,
                 from /usr/include/c++/13/ext/string_conversions.h:45,
                 from /usr/include/c++/13/bits/basic_string.h:4109,
                 from /usr/include/c++/13/string:54,
                 from ../src/common/tvgCommon.h:33,
                 from ../src/renderer/tvgRender.h:28,
                 from ../src/renderer/gl_engine/tvgGlCommon.h:28,
                 from ../src/renderer/gl_engine/tvgGlShaderCache.h:26,
                 from ../src/renderer/gl_engine/tvgGlShaderCache.cpp:23:
In function ‘int snprintf(char*, size_t, const char*, ...)’,
    inlined from ‘static const char* GlShaderCache::path(const char*, const char*)’ at ../src/renderer/gl_engine/tvgGlShaderCache.cpp:131:13:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:54:35: note: ‘__builtin___snprintf_chk’ output between 21 and 4124 bytes into a destination of size 4096
   54 |   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   55 |                                    __glibc_objsize (__s), __fmt,
      |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   56 |                                    __va_arg_pack ());
      |                                    ~~~~~~~~~~~~~~~~~
 ```

**Comment by hermet** [comment]
@wenjieshen I think it's good to add `common/tvgEnv.h` (For now, code in the header could be better for size optimization), which provides the base thorvg config path based on a current system,  the gl engine calls the function for generating the shader cache path.

File: src/renderer/gl_engine/tvgGlShaderCache.cpp [hunk 70]
```diff
@@ -0,0 +1,271 @@
...
+
+    if (!glProgramBinarySupport()) {
+        return;
+    }
+
+    auto cachePath = path(vertSrc, fragSrc);
+    if (!cachePath) return;
+
+    GLint numFormats = 0;
+    glGetIntegerv(GL_NUM_PROGRAM_BINARY_FORMATS, &numFormats);
```
hermet: what do you think combine this with `glProgramBinarySupport()`?

wenjieshen: The idea is great. I am going to test if it can be combined with glProgramBinarySupport

wenjieshen: It works well. Thanks.

wenjieshen:
> @wenjieshen I think it's good to add `common/tvgEnv.h` (For now, code in the header could be better for size optimization), which provides the base thorvg config path based on a current system, the gl engine calls the function for generating the shader cache path.

Thank you for the suggestion. I've added `common/tvgEnv.h`, and extracted a method: `getCacheDir` in the header.

src/common/tvgEnv.h
```diff
@@ -0,0 +1,122 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef _TVG_ENV_H_
+#define _TVG_ENV_H_
+
```
hermet, line 25: Please  `#include "tvgCommon"`

probably you can combine 26 ~ 42 lines code with `tvgCommon`

src/common/tvgEnv.h
```diff
@@ -0,0 +1,122 @@
...
+        }
+#elif defined(__linux__)
+        // Linux: $XDG_CACHE_HOME/thorvg/shaders or ~/.cache/thorvg/shaders
+        auto xdgCache = getenv("XDG_CACHE_HOME");
+        if (xdgCache) {
+            snprintf(cacheDir, maxLen, "%s/thorvg/shaders", xdgCache);
+        } else {
+            auto home = getenv("HOME");
+            if (!home) return false;
+            snprintf(cacheDir, maxLen, "%s/.cache/thorvg/shaders", home);
```
hermet: This function needs to have `%s/.cache/thorvg/`

2 options, we can have:
1. you can just put the shader files in thorvg without `shaders`
2. you can append the `shaders` subfolder from gl engine

I bet on no.1

src/common/tvgEnv.h
```diff
@@ -0,0 +1,122 @@
...
+        }
+        snprintf(cacheDir, maxLen, "%s\\thorvg\\shaders", appData);
+
+        auto attribs = GetFileAttributesA(cacheDir);
+        if (attribs == INVALID_FILE_ATTRIBUTES) {
+            char thorvgDir[PATH_MAX];
+            snprintf(thorvgDir, PATH_MAX, "%s\\thorvg", appData);
+
+            if (!CreateDirectoryA(thorvgDir, NULL) && GetLastError() != ERROR_ALREADY_EXISTS) return false;
+
```
hermet, line 59: -- 
less LOC is preferred.

src/common/tvgEnv.h
```diff
@@ -0,0 +1,122 @@
...
+    #include <unistd.h>
+#endif
+
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+
+namespace tvg
+{
+    inline bool getCacheDir(char* cacheDir, size_t maxLen)
```
hermet: -> `cachedir()`

We don't essentially add set/get prefix.

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -20,6 +20,7 @@
  * SOFTWARE.
  */
 
+#include <cstring>
```
hermet: --
tvgCommon.h already includes cstring

src/common/tvgEnv.h
wenjieshen: Thanks. The headers related to string handling have been replaced with tvgCommon.

src/common/tvgEnv.h
wenjieshen: I agree, organizing the cache folder seems weird. 🙏

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,175 @@
...
+#ifdef _WIN32
+    #ifndef PATH_MAX
+        #define PATH_MAX MAX_PATH
+    #endif
+#else
+    #include <limits.h>
+#endif
+
+#ifndef __EMSCRIPTEN__
+static char cachePath[PATH_MAX];
```
hermet: Let's remove this static var.

src/renderer/gl_engine/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,175 @@
...
+#if !defined(THORVG_FILE_IO_SUPPORT) || defined(__EMSCRIPTEN__)
+    return 0;
+#endif
+
+    if (!vertSrc || !fragSrc) return 0;
+
+    if (!glProgramBinarySupport()) return 0;
+
+    // Read from cache if exists
+    auto cachePath = path(vertSrc, fragSrc);
```
hermet: char cachePath[PATH_MAX];
path(vertSrc, fragSrc, &cachePath);

src/renderer/gl_engine/tvgGlShaderCache.h
```diff
@@ -0,0 +1,36 @@
...
+#ifndef _TVG_GL_SHADER_CACHE_H_
+#define _TVG_GL_SHADER_CACHE_H_
+
+#include "tvgGlCommon.h"
+
+struct GlShaderCache
+{
+    static void write(uint32_t progObj, const char* vertSrc, const char* fragSrc);
+    static uint32_t read(const char* vertSrc, const char* fragSrc);
+    static const char* path(const char* vertSrc, const char* fragSrc);
```
hermet: something like: `static bool path(const char* vertSrc, const char* fragSrc, char** out);`

src/common/tvgEnv.h
```diff
@@ -0,0 +1,119 @@
...
+#else
+    #include <limits.h>
+    #include <sys/stat.h>
+    #include <unistd.h>
+#endif
+
+
+namespace tvg
+{
+    inline bool cacheDir(char* cacheDir, size_t maxLen)
```
hermet: cacheDir -> cachedir()

src/renderer/gl_engine/tvgGlShaderCache.h
```diff
@@ -0,0 +1,36 @@
...
+ * SOFTWARE.
+ */
+
+#ifndef _TVG_GL_SHADER_CACHE_H_
+#define _TVG_GL_SHADER_CACHE_H_
+
+#include "tvgGlCommon.h"
+
+struct GlShaderCache
+{
```
hermet, line 30: `static bool supported;`

src/renderer/gl_engine/tvgGl.cpp
```diff
@@ -152,6 +152,20 @@ static void* _getProcAddress(const char* procName) {
         return false;                                           \
     }
 
+static bool _glProgramBinarySupport = false;
```
hermet: let's add this in GlShaderCache class.

src/renderer/gl_engine/tvgGlProgram.cpp
```diff
@@ -35,10 +36,19 @@ uint32_t GlProgram::mCurrentProgram = 0;
 
 GlProgram::GlProgram(const char* vertSrc, const char* fragSrc)
 {
+
+#ifndef __EMSCRIPTEN__
+    mProgramObj = GlShaderCache::read(vertSrc, fragSrc);
+
+    if (mProgramObj) {
```
hermet: if (mProgramObj) return;

hermet:
@wenjieshen Maybe we can consider removing deprecated older shader files...

What do you think if we use like: `tvg_glshader_{TVG_VERSION}.bin`

The binary format could have three sections - [Tag(2byte) +  Hash(4 bytes) + Binary Shader]
The tag notify us that file is thorvg gl shader file. We can have a specific value for this.
The binary shader is stated with offset 6 bytes.

```
cache = load(tvg_glshader_{TVG_VERSION}  //see: THORVG_VERSION_NUMBER()
verify cache file is valid with the tag. 
use shader if the cache hash value equals current shader hash
```

If the current version shader binary is existed but the hash is different, we can remove the old and update with the latest.

wenjieshen:
Thank you for the suggestion. I’ve added a validation header to the binary file.
Screenshot 2025-10-29 at 9 27 25 PM
After corrupting a file (as shown in the screenshot above), the cache successfully detected the issue and removed the file.

However, identifying and removing redundant files is still challenging. At the moment, I don’t have a good way to determine which files are no longer needed. I’d prefer to address this together with the version that stores the mono cache file, so both features can be handled cleanly in the same update.

> @wenjieshen Maybe we can consider removing deprecated older shader files...
> 
> What do you think if we use like: `tvg_glshader_{TVG_VERSION}.bin`
> 
> The binary format could have three sections - [Tag(2byte) + Hash(4 bytes) + Binary Shader] The tag notify us that file is thorvg gl shader file. We can have a specific value for this. The binary shader is stated with offset 6 bytes.
> 
> ```
> cache = load(tvg_glshader_{TVG_VERSION}  //see: THORVG_VERSION_NUMBER()
> verify cache file is valid with the tag. 
> use shader if the cache hash value equals current shader hash
> ```
> 
> If the current version shader binary is existed but the hash is different, we can remove the old and update with the latest.

SergeyLebedkin (approved):
✅windows 11: OK
✅ubuntu (wsl2): OK
✅ubuntu (h/w): OK
✅macos (intel): OK (there are no supported binary formats)

hermet:
The file IO is expensive relatively. Current shaders are more than 10, it needs a consideration before pushing this. That would be the best if we can introduce the one single binary form....

hermet:
**1. Verifying Shaders**
- Current: Use a Hash & ThorVG version
- After: Use the ThorVG version only. ie) `CACHE_FOLDER/thorvg/gl_shaders_VERSION_INFO.bin`
- Later we can introduce a binary shader feature toggle option. 
- - On dev, we can disable the binary shader, but we can enable it manually when necessary.
- - Enable the binary shader feature with the official release binary.

**2. Lazy Loading or Immediate Loading**
 - Lazy Loading needs to hold the binary shader memory, We need to confirm the total size of the memory first.

wenjieshen:
> **1. Verifying Shaders**
> 
> * Current: Use a Hash & ThorVG version
> * After: Use the ThorVG version only. ie) `CACHE_FOLDER/thorvg/gl_shaders_VERSION_INFO.bin`
> * Later we can introduce a binary shader feature toggle option.
> * * On dev, we can disable the binary shader, but we can enable it manually when necessary.
> * * Enable the binary shader feature with the official release binary.
> 
> **2. Lazy Loading or Immediate Loading**
> 
> * Lazy Loading needs to hold the binary shader memory, We need to confirm the total size of the memory first.

@hermet 
I understand. Thanks for the suggestion.  
We can roll back this version if the loading efficiency gain doesn’t justify the additional memory overhead. On average, loading a single discrete shader takes 0.59 ms, while loading the combined binary file (containing all shaders) takes 0.87 ms. Since each example may require between 17 and 89 shaders, the unified loading approach can improve total loading performance by roughly **11× to 60×**, depending on how many shaders are needed.
### Memory Usage  
After combining the binary shader files, the **maximum additional memory usage** observed across all examples is **857.21 KB**.

### Shader Loading Time (per example)  
- **Mean (Average):** 0.867 ms  
- **Median (50th Percentile):** 0.877 ms  
- **Standard Deviation:** 0.175 ms  
- **Variance:** 0.031  
- **Minimum:** 0.183 ms (Fastest load time — *Animation.exe*)  
- **Maximum:** 1.322 ms (Slowest load time — *StrokeLine.exe*)

### Accumulated Write Time(per example)
- Blending.exe: $10.897$ ms
- Accessor.exe: $4.334$ ms
- SceneEffects.exe: $1.415$ ms
- Duplicate.exe: $0.455$ ms
- EffectDropShadow.exe: $0.343$ ms
### Shader Writing Time (per shader)  
- **Mean (Average):** 0.203 ms  
- **Median (50th Percentile):** 0.174 ms  
- **Standard Deviation:** 0.090 ms  
- **Variance:** 0.008  
- **Minimum:** 0.126 ms (Fastest write time)  
- **Maximum:** 0.670 ms (Slowest write time)

wenjieshen:
@hermet Hi Hermet, there is more practical testing data. The Binary shader includes the time spent on file I/O for loading, but does not include writing.
Exec | #Shaders | Original CSV total (ms) | Binary CSV total (ms) | Speedup (orig / binary)
-- | -- | -- | -- | --
0 | 80 | 588.421 | 13.490 | 43.62×
1 | 80 | 571.231 | 12.115 | 47.15×
2 | 80 | 571.627 | 10.817 | 52.85×
3 | 80 | 628.088 | 10.865 | 57.81×
4 | 80 | 579.044 | 11.483 | 50.43×
5 | 80 | 577.148 | 10.702 | 53.93×
6 | 80 | 582.733 | 10.572 | 55.12×
7 | 80 | 582.633 | 11.139 | 52.31×
8 | 80 | 578.692 | 11.152 | 51.89×
9 | 80 | 579.014 | 11.394 | 50.82×
10 | 80 | 550.348 | 9.966 | 55.22×

### Statistical comparison (using the 10 complete executions, 80 shaders each)

- Original (glprogram_timing.csv):
1. Mean total per execution: 583.9 ms
2. Std dev: 16.4 ms
3. Min / Max: 571.2 ms / 628.1 ms
- Binary (glprogram_timing_binary.csv):
1. Mean total per execution: 11.37 ms
2. Std dev: 0.87 ms
3. Min / Max: 10.57 ms / 13.49 ms

Relative:

Average speedup over the first 10 executions: ≈ 51.6× faster using the binary timings.

On average we save about ~572 ms per execution (583.9 ms − 11.37 ms).
[glprogram_timing.csv](https://github.com/user-attachments/files/23539988/glprogram_timing.csv)
[glprogram_timing_binary.csv](https://github.com/user-attachments/files/23539990/glprogram_timing_binary.csv)

wenjieshen:
On the Windows platform, the nVidia driver automatically caches binary shaders. Disabling this setting provides more accurate values.
image

src/common/tvgEnv.h
```diff
@@ -0,0 +1,119 @@
...
+ */
+
+#ifndef _TVG_ENV_H_
+#define _TVG_ENV_H_
+
+#include "tvgCommon.h"
+
+#ifdef _WIN32
+    #ifndef WIN32_LEAN_AND_MEAN
+        #define WIN32_LEAN_AND_MEAN
```
hermet: our config has this by default:

image

no need to add define.

Here portable if-def can be moved in `tvgCommon.h`

hermet (commented):
@wenjieshen please check a comment.

src/common/tvgEnv.h
wenjieshen: Thank you for your suggestion and improvement. I have rebased on the [PR](https://github.com/thorvg/thorvg/pull/3982) and made modifications based on it, resulting in extensive changes. However, it should normalize once the [PR](https://github.com/thorvg/thorvg/pull/3982) is merged.

hermet:
feature target: thorvg v2.0

hermet:
@wenjieshen could you please rebase this? Thanks!

src/common/tvgEnv.h
```diff
@@ -0,0 +1,115 @@
...
+            if (!appData) return false;
+        }
+        snprintf(output, maxLen, "%s\\thorvg", appData);
+
+        auto attribs = GetFileAttributesA(output);
+        if (attribs == INVALID_FILE_ATTRIBUTES) {
+            char thorvgDir[PATH_MAX];
+            snprintf(thorvgDir, PATH_MAX, "%s\\thorvg", appData);
+
+            if ((!CreateDirectoryA(thorvgDir, NULL) || !CreateDirectoryA(output, NULL)) && GetLastError() != ERROR_ALREADY_EXISTS) return false;
```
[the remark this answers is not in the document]
wenjieshen: Different access permissions.

hermet:
At boot time: 3%+ 
Only Shader loading part: 12%+

(Tested on Intel Arc Graphics)

@wenjieshen Unlike frame-time enhancements, a 3% improvement in boot time is somewhat underwhelming, especially when considering the maintenance overhead. Let's put this optimization on hold for now. We may see better performance improvements by considering a single, unified binary shader. However, since the overall burden is under 1ms, the improvement might not be very noticeable. Thanks for your trial & effort.

src/renderer/gpu_engine/gl/tvgGlShaderCache.cpp
```diff
@@ -0,0 +1,293 @@
...
+                tvg::free(entry.second.data);
+            cache.clear();
+            return;
+        }
+
+        CachedBinary cached;
+        cached.data = binaryData;
+        cached.length = entryHeader.length;
+        cached.format = entryHeader.binaryFormat;
+        cache[entryHeader.renderType] = cached;  // The last entry will overwrite the previous one with the same renderType
```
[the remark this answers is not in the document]
wenjieshen: It is fair. I updated.

hermet:
@wenjieshen Could you please rebase this one more time? Thanks.

Nor-s:
I tested this using **NVIDIA's cache-disable flag**. 

I ran it on the **Shapes**, **Blending**, and **SceneEffects** examples, 
and measured performance by inserting timing code at `GlRenderer::initShaders()`, `draw`, and `sync`.

I tested it like this:
```
__GL_SHADER_DISK_CACHE=0 ./builddir/src/SceneEffects gl
```

> [!IMPORTANT]
> If i enable the **NVIDIA cache** on **main**, the numbers come out **almost identical to the Binary Shader fetch**.


> [!NOTE]
> [test branch(cherry-pick binary shader feature)](https://github.com/thorvg/thorvg/tree/nor-s/binary-shader-test)
> **ENV: Arch linux, AMD 7600, Nvidia 3060ti**
> **Averaging method:**
> - **Main:** average of all 9 runs.
> - **Binary Shader:** average of the 6 **cache-hit** runs only. The first run of each iteration is excluded, since that run compiles the shaders and bakes the cache (including it would dilute the cache effect). 
> - And I cleared the cache for each individual test run.




### Shapes

| Stage | Change |
|---|---|
| init_shaders |  **−98.8%** |
| first canvas draw |  **−98.7%** |
| first canvas sync  | **+65.6%** |
| Boot (ms)  | **−53.2%** |

### Blending

| Stage | Change |
|---|---|
| init_shaders | **−98.2%** |
| first canvas draw |  **−85.6%** |
| first canvas sync | **+276.8%** |
| Boot (ms) |  **−44.6%** |


### Scene Effects

| Stage |  Change |
|---|---|
| init_shaders |  **−98.6%** |
| first canvas draw  | **−86.8%** |
| first canvas sync  | **+19.9%** |
| Boot (ms)| **−49.4%** |

hermet:
@wenjieshen please rebase it.

hermet:
leave to gpu driver.

Issue #3901, opened 2025-10-20
loader: file open refactoring


hermet:
Every loader has its own file open implementation. If the common loader handles this and request to loaders only opens from memory, we can achieve more consolidate logic and easier maintenance.

PR #3902, opened 2025-10-20
loaders: commonize file open logic

hermet:
consolidated file-opening into a general-purpose function for all loaders. Removed duplicate code to improve maintainability and readability.

issue: https://github.com/thorvg/thorvg/issues/3901


hermet:
size: -1kb (-s) / -2.6kb (O3),  when all loaders are enabled

Issue #3665, opened 2025-07-30
gl/wg: poor stroke quality


hermet:
GL has relatively lower stroke quality, and its tessellation can be improved as much as webgpu.

gl vs wg
<img width="742" height="370" alt="Image" src="https://github.com/user-attachments/assets/dd69315d-d485-4f67-bdf7-94819bf3bcad" />

hermet:
See this as well:: https://github.com/thorvg/thorvg/issues/3668

Tessellation in GL/WG should be considered along with the transformation (scaling).

Issue #3668, opened 2025-07-31
engines: excessive large stroke width has a different visual result.


hermet:
It seems that only the SW engine has even/odd visual holes. Need to figure out and align the result among engines.

sw:

<img width="1237" height="1256" alt="Image" src="https://github.com/user-attachments/assets/6919085f-4e91-48b5-817b-4370838ca8e2" />

gl:

<img width="1235" height="1250" alt="Image" src="https://github.com/user-attachments/assets/7f2ec42f-d242-4d60-a050-32797f8ae56c" />

wg:

<img width="1039" height="1053" alt="Image" src="https://github.com/user-attachments/assets/959b55a3-7b80-4cf6-a1df-66da9d72e2f5" />

```
    bool content(tvg::Canvas* canvas, uint32_t w, uint32_t h) override
    {
        //Prepare Round Rectangle
        auto shape1 = tvg::Shape::gen();
        shape1->moveTo(100, 100);
        shape1->lineTo(120, 100);
        shape1->lineTo(120, 120);
        shape1->close();
        shape1->fill(0, 255, 0);
        shape1->strokeWidth(70);
        shape1->strokeJoin(tvg::StrokeJoin::Round);
        shape1->strokeFill(255, 255, 255);

        tvg::Matrix m = {10, 0, -600, 0, 10, -500, 0, 0, 1};
        shape1->transform(m);

        canvas->push(shape1);

        return true;
    }
```

PR #3903, opened 2025-10-20
gl_engine/tessellator: Improve stroke tessellation quality

wenjieshen:
## Summary
This patch partially resolves issue [#3665](https://github.com/thorvg/thorvg/issues/3665)
 by improving stroke tessellation quality in the GL engine.
The updated algorithm produces smoother and more accurate stroke rendering, achieving visual results closer to the WebGPU engine.
Screenshot 2025-10-21 at 1 47 19 AM

### Changes

- Improved segment count calculation in the tessellation join process.

- Updated the member function `lineTo` and `close` to imitate `cubicTo` logic, passing the Transform matrix as a parameter to `round` to ensure proper segment computation.

### Current Status

Fixes the basic stroke quality problem described in #3665.

The odd–even rendering issue mentioned in [#3668](https://github.com/thorvg/thorvg/issues/3668)
 remains and will be addressed in a future patch.

### Related Issues

Partially resolves: #3665 (GL/WG: poor stroke quality)

Related: #3668 (Engines: excessive stroke width causes visual inconsistency)

### DIFF
--- a/src/renderer/gl_engine/tvgGlTessellator.cpp
+++ b/src/renderer/gl_engine/tvgGlTessellator.cpp
@@ -65,6 +65,7 @@ void Stroker::run(const RenderPath& path, const Matrix& m)
 {
     mBuffer->vertex.reserve(path.pts.count * 4 + 16);
     mBuffer->index.reserve(path.pts.count * 3);
+    mScale = tvg::scaling2D(m);
 
     auto validStrokeCap = false;
     auto pts = path.pts.data;
@@ -251,7 +252,7 @@ void Stroker::round(const Point &prev, const Point& curr, const Point& center)
     mRightBottom.y = std::max(mRightBottom.y, std::max(center.y, std::max(prev.y, curr.y)));
 
     // Fixme: just use bezier curve to calculate step count
-    auto count = Bezier(prev, curr, radius()).segments();
+    auto count = Bezier(prev * mScale, curr * mScale, radius() * length(mScale)).segments();
     auto c = _pushVertex(mBuffer->vertex, center.x, center.y);
     auto pi = _pushVertex(mBuffer->vertex, prev.x, prev.y);
     auto step = 1.f / (count - 1);
--- a/src/renderer/gl_engine/tvgGlTessellator.h
+++ b/src/renderer/gl_engine/tvgGlTessellator.h
@@ -71,6 +71,7 @@ private:
     State mState = {};
     Point mLeftTop = {0.0f, 0.0f};
     Point mRightBottom = {0.0f, 0.0f};
+    Point mScale;
 };
 
 class BWTessellator
--- a/src/common/tvgMath.h
+++ b/src/common/tvgMath.h
@@ -134,6 +134,14 @@ static inline float scaling(const Matrix& m)
 }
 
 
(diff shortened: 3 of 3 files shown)


SergeyLebedkin:
@wenjieshen 
I think using the entire matrix is ​​redundant for quality assessment. In this context, we're not interested in rotation and translation, only scale. There's a suggestion to initially calculate the scale
**float scaling(const Matrix& m)**
and then use it for tessellation.

wenjieshen:
@SergeyLebedkin 
Thanks for the suggestion. I actually tried using a single scale factor, but the Bézier curves looked distorted when only the radius was scaled in 1D. That’s why I just changed it to pass a 2D scale vector instead of a matrix to the `round` function and added the utilities handling Bézier scaling. If this feels like too much change on utilities, I can roll it back to the previous version( the matrix solution).

wenjieshen:
@hermet 
Thanks for the suggestion. I have changed to retain the scale as a member variable. It seems correct in several cases. I will finish the test in general later today.

wenjieshen:
@SergeyLebedkin @hermet Thank you for your patience. I checked there is no regression occurring. I think this PR is ready.
image

src/renderer/gl_engine/tvgGlTessellator.h
```diff
@@ -71,6 +71,7 @@ class Stroker
     State mState = {};
     Point mLeftTop = {0.0f, 0.0f};
     Point mRightBottom = {0.0f, 0.0f};
+    Point mScale = {1.0f, 1.0f};
```
hermet: Please remove unnecessary initialization, Thanks.
```
 Point mScale;
```

wenjieshen: Thank you. I've removed.

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -251,7 +252,7 @@ void Stroker::round(const Point &prev, const Point& curr, const Point& center)
     mRightBottom.y = std::max(mRightBottom.y, std::max(center.y, std::max(prev.y, curr.y)));
 
     // Fixme: just use bezier curve to calculate step count
-    auto count = Bezier(prev, curr, radius()).segments();
+    auto count = (Bezier(prev, curr, radius()) * mScale).segments();
```
hermet: "multiply" a point on a Bézier curve isn't in the standard sense; Please use this way:
`Bezier(prev * mScale, curr * mScale, radius() * mScale).segment()`

wenjieshen:
I ran the Lottie examples five times on each branch (tested on macOS):

- Main branch: 33.33 – 33.79 fps

- This branch: 35.13 – 35.70 fps

Interestingly, this branch shows slightly higher FPS, which isn’t expected.

SergeyLebedkin:
> I ran the Lottie examples five times on each branch (tested on macOS):
> 
> * Main branch: 33.33 – 33.79 fps
> * This branch: 35.13 – 35.70 fps
> 
> Interestingly, this branch shows slightly higher FPS, which isn’t expected.

This is explained by the fact that it is most likely that the reduction in the size of the Bézier curves is more common than the scale-up, which leads to a reduction in the number of points for tessellation

PR #3904, opened 2025-10-21
engines: add missing colorspace information

hermet:
Properly retain the requested colorspace information for each engine. Although these colorspaces are not currently utilized by gl, wg engines themselves, this information may be required by image loaders to target a specific colorspace.


SergeyLebedkin (approved):
Looks like its just setter and getter for cs setting. No logic changes

Issue #3826, opened 2025-09-22
lottie: enhance slot overriding for effect properties


hermet:
add support for the effect property overriding
```
            "ef": [
                {
                    "ty": 5,
                    "nm": "Stroke",
                    "np": 3,
                    "mn": "ADBE Slider Control",
                    "ix": 1,
                    "en": 1,
                    "ef": [
                        {
                            "ty": 0,
                            "nm": "Slider",
                            "mn": "ADBE Slider Control-0001",
                            "ix": 1,
                            "v": {
                                "a": 0,
                                "k": 20,
                                "ix": 1,
                                "sid": "stroke"
                            }
                        }
                    ]
                },
```

This might not be compatible with our current lottie slot-infra, Need to consider a design revision.

tinyjin:
A new slottable property schema might need to be defined to implement this functionality.


As shown, the Effects ("ef") field can contain multiple Effect objects, and each Effect ("ef") represents a specific type of effect that can be composited together.

<img width="674" height="615" alt="Image" src="https://github.com/user-attachments/assets/6fb3b319-bac0-48f9-a399-cb5a8b650c36" />


Since Effects are defined as part of the Layer structure, it is considered most reasonable to assign a unique sid to each individual Effect in order to enable effect slot overriding.

## Layer Effects with JSON Example

[slottable-layer-effects.json](https://github.com/user-attachments/files/22940069/slottable-layer-effects.json)

```json
{
    // …
    "slots": {
        "slider_control": {
            "p": {
                "ef": [
                        {
                            "ty": 0,
                            "nm": "Slider",
                            "mn": "ADBE Slider Control-0001",
                            "ix": 1,
                            "v": {
                                 // …
                            }
                        }
                    ]
            }
        }
    },
    "layers": [
        {
            // …
            "ef": [
                {
                    "ty": 5,
                    "nm": "Slider Control",
                    "np": 3,
                    "mn": "ADBE Slider Control",
                    "ix": 1,
                    "en": 1,
                    "sid": "slider_control", // `sid` for Slot Overriding
                    "ef": [
                        {
                            "ty": 0,
                            "nm": "Slider",
                            "sid": "slider_control",
                            "mn": "ADBE Slider Control-0001",
                            "ix": 1,
                            "v": {
                                // …
                            }
                        }
                    ]
                }
            ],
            "ip": 0,
            "op": 100,
            "st": 0,
            "bm": 0
        },
	// …
}
```

[...]

PR #3906, opened 2025-10-22
lottie/slot: support custom effect slot overriding

tinyjin:
This patch enables slot overriding for custom effects (LottieFxCustom).

Allowing Lottie animations to dynamically override effect properties for Custom effect
through the slot system, providing runtime customization of visual effects.

### Sample (default effect slot)

- [effect-default-slot.json](https://github.com/user-attachments/files/23117184/slot12.json)
- [multiple-effect-slot.json](https://github.com/user-attachments/files/23120664/slot12.json)

### Result (lottie-web vs ThorVG)

1. Single property test

CleanShot 2025-10-23 at 01 38 14

2. Multiple property test

CleanShot 2025-10-24 at 19 05 55

issue: https://github.com/thorvg/thorvg/issues/3826


src/loaders/lottie/tvgLottieModel.h
```diff
@@ -1017,7 +1022,7 @@ struct LottieLayer : LottieGroup
     LottieEffect* effectById(unsigned long id)
     {
         ARRAY_FOREACH(p, effects) {
-            if (id == (*p)->nm || id == (*p)->mn) return *p;
```
hermet, line 1173: Is it ok to change this? I remember `nm` is intentionally used.

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -118,7 +166,7 @@ struct LottieFxCustom : LottieEffect
         ARRAY_FOREACH(p, props) delete(p->property);
     }
 
-    Property* property(int type)
```
hermet, line 123: It would be nicer if you keep the original method name. Please check:
```
    using LottieObject::property;
    Property* property(int type);
```

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -142,9 +190,8 @@ struct LottieFxCustom : LottieEffect
         return nullptr;
     }
 
-    LottieProperty* property(const char* name)
+    LottieProperty* propertyById(unsigned long id)
```
hermet: Is it necessary to change the method name?

hermet (commented):
@tinyjin Thanks, Please check comments, only reviewed: "lottie: revise LottieEffect model to inherit from LottieObject"

src/loaders/lottie/tvgLottieModel.h
tinyjin: LottieObject has `id` property that is same value as `nm` in effect. So the same value is parsed where parseCommon is fired and stores value in `obj->id`. Also I changed effect object uses `obj->id` instead of `effect->nm`.

I guess the value can be unified, that is reason I changed on this as long as `effect->nm` and `obj->id` remain the same.

Image

Image

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -142,9 +191,8 @@ struct LottieFxCustom : LottieEffect
...
```
hermet: I don't think you need to change this as well.

hermet (commented):
@tinyjin Reviewed "lottie: revise LottieEffect model to inherit from LottieObject". 

Please check a comment. Thanks.

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -692,6 +704,18 @@ struct LottieTransform : LottieObject
         return nullptr;
     }
 
+    LottieProperty* backup(LottieProperty* prop) override
+    {
+        switch (prop->type) {
```
hermet: Some object could have duplicate type properties. how to figure it out?...

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -208,52 +208,8 @@ void LottieSlot::apply(LottieProperty* prop, bool byDefault)
     //apply slot object to all targets
     ARRAY_FOREACH(pair, pairs) {
         //backup the original properties before overwriting
```
hermet: please remove a comment this (210). it's no more useful.

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -208,52 +208,8 @@ void LottieSlot::apply(LottieProperty* prop, bool byDefault)
...
-                break;
-            }
-            case LottieProperty::Type::Image: {
-                if (copy) pair->prop = new LottieBitmap(static_cast<LottieImage*>(pair->obj)->data);
-                pair->obj->override(prop, !copy);
-                break;
-            }
-            default: break;
-        }
+        if (copy) pair->prop = pair->obj->backup(prop);
```
hermet: Rather than introducing "backup", Please consider to use `override()`, you could copy properties there.

hermet (commented):
Reviewed: "lottie/slot: support custom effect slot overriding"

src/loaders/lottie/tvgLottieModel.h
tinyjin: The initial PR used "ix" to match properties for overriding, but "ix" is not always guaranteed in JSON.

Added a `sid` field to `LottieProperty` and use it for custom effect cases where property types can be duplicated.

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -206,55 +206,7 @@ void LottieSlot::apply(LottieProperty* prop, bool byDefault)
     auto copy = !overridden && !byDefault;
```
hermet: Could you please change the variable name to `release` ?

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -206,55 +206,7 @@ void LottieSlot::apply(LottieProperty* prop, bool byDefault)
...
-            }
-            case LottieProperty::Type::Image: {
-                if (copy) pair->prop = new LottieBitmap(static_cast<LottieImage*>(pair->obj)->data);
-                pair->obj->override(prop, !copy);
-                break;
-            }
-            default: break;
-        }
-    }
+    ARRAY_FOREACH(pair, pairs) pair->prop = pair->obj->override(prop, !copy);
```
hermet:
```
ARRAY_FOREACH(pair, pairs) {
    pair->prop = pair->obj->override(prop, release);
}
```

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -778,15 +797,22 @@ struct LottieSolidFill : LottieSolid
...
+            else backup = new LottieColor(color);
+
             color.copy(*static_cast<LottieColor*>(prop), false);
         } else if (prop->type == LottieProperty::Type::Opacity) {
             if (release) opacity.release();
+            else backup = new LottieOpacity(opacity);
+
             opacity.copy(*static_cast<LottieOpacity*>(prop), false);
         }
+
```
hermet: --

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -167,6 +167,51 @@ struct LottieFxCustom : LottieEffect
...
+        LottieProperty* backup = nullptr;
+        ARRAY_FOREACH(p, props) {
+            if (p->property->type != prop->type || p->property->sid != prop->sid) continue;
+            auto target = p->property;
+            switch (prop->type) {
+                case LottieProperty::Type::Float: {
+                    auto cur = static_cast<LottieFloat*>(target);
+                    if (release) cur->release();
+                    else backup = new LottieFloat(*cur);
+
```
hermet: --

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -167,6 +167,51 @@ struct LottieFxCustom : LottieEffect
...
+                    cur->copy(*static_cast<LottieFloat*>(prop), false);
+                    break;
+                }
+                case LottieProperty::Type::Color: {
+                    auto cur = static_cast<LottieColor*>(target);
+                    if (release) cur->release();
+                    else backup = new LottieColor(*cur);
+
```
hermet: --

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -167,6 +167,51 @@ struct LottieFxCustom : LottieEffect
...
+                    cur->copy(*static_cast<LottieColor*>(prop), false);
+                    break;
+                }
+                case LottieProperty::Type::Vector: {
+                    auto cur = static_cast<LottieVector*>(target);
+                    if (release) cur->release();
+                    else backup = new LottieVector(*cur);
+
```
hermet: --

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -167,6 +167,51 @@ struct LottieFxCustom : LottieEffect
...
+                    cur->copy(*static_cast<LottieVector*>(prop), false);
+                    break;
+                }
+                case LottieProperty::Type::Integer: {
+                    auto cur = static_cast<LottieInteger*>(target);
+                    if (release) cur->release();
+                    else backup = new LottieInteger(*cur);
+
+                    cur->copy(*static_cast<LottieInteger*>(prop), false);
```
hermet, line 203: --

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -349,7 +349,10 @@ uint32_t LottieLoader::gen(const char* slots, bool byDefault)
         auto found = false;
         ARRAY_FOREACH(p, comp->slots) {
             if ((*p)->sid != sid) continue;  //find target
-            if (auto prop = parser.parse(*p)) custom->props.push({prop, *p});
+            if (auto prop = parser.parse(*p)) {
+              prop->sid = sid;
```
hermet: please check indentation.

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -349,7 +349,10 @@ uint32_t LottieLoader::gen(const char* slots, bool byDefault)
         auto found = false;
         ARRAY_FOREACH(p, comp->slots) {
             if ((*p)->sid != sid) continue;  //find target
-            if (auto prop = parser.parse(*p)) custom->props.push({prop, *p});
+            if (auto prop = parser.parse(*p)) {
+                prop->sid = sid;
```
hermet: Hmm. Shouldn't the property have a `sid` during the parsing stage?

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1326,6 +1331,11 @@ bool LottieParser::parseEffect(LottieEffect* effect, void(LottieParser::*func)(L
             else if (property && KEY_AS("mn")) property->mn = djb2Encode(getString());
             else skip();
         }
+
+        if (sid && property) {
```
hermet: Please consider this: 
`if (sid && property) registerSlot(effect, sid, property);`

hermet (commented):
@tinyjin please check comments. thanks.

src/loaders/lottie/tvgLottieLoader.cpp
tinyjin: @hermet 
The `sid` parsed from "slots" does not belong to the property itself. At parse time, the Property instance may not exist yet, so we can’t assign immediately.

The `sid` is returned at `auto sid = djb2Encode(parser.sid(idx == 0))` after all parsing.

Thus, I made line `prop->sid = sid;` to assign id after property is ready at slot generation step(parsing stage).

| CleanShot 2025-11-11 at 00 17 04@2x | CleanShot 2025-11-11 at 00 16 54@2x |
|:--:|:--:|
| From original property | From "slots" |

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -692,31 +697,37 @@ struct LottieTransform : LottieObject
         return nullptr;
     }
 
-    void override(LottieProperty* prop, bool release) override
+    LottieProperty* override(LottieProperty* prop, bool release) override
```
hermet, line 792: Can't use sid comparsion for overriding the specific target property?

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -349,7 +349,10 @@ uint32_t LottieLoader::gen(const char* slots, bool byDefault)
...
+                custom->props.push({prop, *p});
```
hermet: LottieSlot already has a `sid`, maybe you can assign `sid` inside parser.parse().

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -474,7 +474,10 @@ void LottieParser::registerSlot(LottieObject* obj, const char* sid, LottieProper
         (*p)->pairs.push({obj});
         return;
     }
-    comp->slots.push(new LottieSlot(context.layer, context.parent, val, obj, type));
+
+    auto slot = new LottieSlot(context.layer, context.parent, val, obj, prop->type);
```
hermet: `comp->slots.push(new LottieSlot(context.layer, context.parent, val, obj, prop->type));`

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1326,6 +1331,8 @@ bool LottieParser::parseEffect(LottieEffect* effect, void(LottieParser::*func)(L
             else if (property && KEY_AS("mn")) property->mn = djb2Encode(getString());
             else skip();
         }
+
```
hermet: --

hermet (commented):
@tinyjin ++nitpicks

hermet:
@tinyjin Could you please check memory usage diff with perf-test 100 anim? Thanks.

src/loaders/lottie/tvgLottieModel.h
tinyjin: @hermet Thanks for suggestion :)
I revised that logic with sid-based overriding for every slottable props execept Custom Effect:
[b3178b505676d9dd820cf9be37dbc4e6273ebe98](https://github.com/thorvg/thorvg/pull/3906/commits/b3178b505676d9dd820cf9be37dbc4e6273ebe98)

For custom effect, we cannot predict type just with sid, so leave original logic as is.

hermet:
binary size: +1.1kb

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -82,13 +82,61 @@ struct LottieStroke
...
+    virtual bool mergeable() { return false; }
+    virtual LottieProperty* property(uint16_t ix) { return nullptr; }
+
+    unsigned long id = 0;      //unique id by name generated by djb2 encoding
+    Type type;
+    bool hidden = false;       //remove?
+};
+
+
+struct LottieEffect : LottieObject
```
hermet, line 132: Maybe you can remove"enabled(144)" but use "hidden(126)" instead

tinyjin: [003e837362b4e7a9c2009f2c192abb47a9e0c80f](https://github.com/thorvg/thorvg/pull/3906/commits/003e837362b4e7a9c2009f2c192abb47a9e0c80f)

hermet:
@tinyjin please confirm the copilot comments. thanks.

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -685,31 +738,27 @@ struct LottieTransform : LottieObject
...
-                if (release) opacity.release();
-                opacity.copy(*static_cast<LottieOpacity*>(prop), false);
-                break;
-            }
-            default: break;
+    LottieProperty* override(LottieProperty* prop, bool release) override
+    {
+        LottieProperty* backup = nullptr;
+        if (rotation.sid == prop->sid) {
+            if (release) rotation.release();
```
hermet, line 796: noted: The logic can be streamlined like this:
```
if (rotation.sid == prop->sid) return new LottieScalar(prop, scale);
```

hermet (approved):

theashraf:
@tinyjin I'm testing these changes in dotLottie-rs and it doesn't seem that thorvg slot api is able to override the slotted effect value at runtime

tinyjin:
@theashraf Lemme chcek :)
May I have your test files? (OR did you test with same samples here?)

Issue #3889, opened 2025-10-15
gl/wgpu: blur effect is omitted


hermet:
When masking and effect are applied same time, effect is omitted. Need to enhance it.

Software:

<img width="1021" height="1058" alt="Image" src="https://github.com/user-attachments/assets/9a40cbf1-be99-4139-a61d-28ceb00a5762" />

GL/WGPU:

<img width="1021" height="1058" alt="Image" src="https://github.com/user-attachments/assets/f275c7b2-2a1e-4064-8c9b-30f08cbeef2a" />

[add_mask.json](https://github.com/user-attachments/files/22919503/add_mask.json)

PR #3911, opened 2025-10-27
wg/gl: fix blur direction param usage

SergeyLebedkin:
Currently, the direction parameter, which controls Gaussian blur, can take three values: 0: both, 1: horizontal, 2: vertical

However, in lottie animations, values other than those specified are encountered, for example 255.
We will assume that any value other than 1 or 2 means bidirectional blurring.

issue: https://github.com/thorvg/thorvg/issues/3889


hermet:
Thanks for reviewing this. This is a user problem. let them deal with it.

hermet:
https://github.com/thorvg/thorvg/pull/3913

Issue #3793, opened 2025-09-09
gl/wgpu: composition bugs


hermet:
cpu (correct)

<img width="1019" height="1014" alt="Image" src="https://github.com/user-attachments/assets/9401ef2b-1d9b-48d8-a460-b6d9c03d95de" />

gl (incorrect)

<img width="1009" height="1004" alt="Image" src="https://github.com/user-attachments/assets/ddc3f0db-5967-40db-a936-dcd12ccaba88" />

wgpu (incorrect)

<img width="1267" height="1266" alt="Image" src="https://github.com/user-attachments/assets/12573e1d-cb04-4b1b-97e5-28faeb46f835" />

[lf20_l3qxn9jy.json](https://github.com/user-attachments/files/22229203/lf20_l3qxn9jy.json)

SergeyLebedkin:
minimal broken example
keep investigating

```
        auto shape0 = tvg::Shape::gen();
        shape0->appendCircle(400, 400, 200, 200);
        shape0->fill(219, 91, 90, 255);

            auto shapeMsk = tvg::Shape::gen();
            shapeMsk->appendCircle(400, 400, 200, 200);
            shapeMsk->fill(219, 91, 90, 255);

            auto shapeTgt = tvg::Shape::gen();
            shapeTgt->appendRect(200, 200, 200, 400);
            shapeTgt->fill(170, 182, 207, 255);

            auto scene = tvg::Scene::gen();
            scene->mask(shapeMsk, tvg::MaskMethod::Alpha);
            scene->push(shapeTgt);
            scene->blend(tvg::BlendMethod::Multiply);
        
        canvas->push(shape0);
        canvas->push(scene);
```

PR #3912, opened 2025-10-27
wg_engine: introduce ability to combine masking and custom blend

SergeyLebedkin:
Until now, blending and masking settings couldn't be used simultaneously. 
If a mask was specified for an object, the blending settings were ignored, and the object was blended using the normal settings. 

Now it's possible to use compositions and blending in the same scene. 
This is achieved by reconfiguring the current render targets and render tree. 

This doesn't affect overall performance, but only adds functionality.

sw/ wg(before)/wg(now)
image

issue: https://github.com/thorvg/thorvg/issues/3793

### DIFF
--- a/src/renderer/wg_engine/tvgWgRenderer.cpp
+++ b/src/renderer/wg_engine/tvgWgRenderer.cpp
@@ -440,63 +440,77 @@ RenderCompositor* WgRenderer::target(const RenderRegion& region, TVG_UNUSED Colo
     // create and setup compose data
     WgCompose* compose = new WgCompose();
     compose->aabb = region;
+    compose->flags = flags;
     mCompositorList.push(compose);
     return compose;
 }
 
 
+// please, keep beginComposite/endComposite aligned by logic
 bool WgRenderer::beginComposite(RenderCompositor* cmp, MaskMethod method, uint8_t opacity)
 {
-    // save current composition settings
+    // current composition
     WgCompose* compose = (WgCompose*)cmp;
-    compose->method = method;
-    compose->opacity = opacity;
-    compose->blend = mBlendMethod;
-    WgSceneTask* sceneTaskCurrent = mSceneTaskStack.last();
     // allocate new render target and push to the render tree stack
-    WgRenderTarget* renderTarget = mRenderTargetPool.allocate(mContext);
-    mRenderTargetStack.push(renderTarget);
-    // create and setup new scene task
-    WgSceneTask* sceneTask = new WgSceneTask(renderTarget, compose, sceneTaskCurrent);
-    sceneTaskCurrent->children.push(sceneTask);
-    mRenderTaskList.push(sceneTask);
-    mSceneTaskStack.push(sceneTask);
+    WgRenderTarget* newRenderTarget = mRenderTargetPool.allocate(mContext);
+    mRenderTargetStack.push(newRenderTarget);
+    // current and new scenes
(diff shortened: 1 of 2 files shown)


src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -440,63 +440,77 @@ RenderCompositor* WgRenderer::target(const RenderRegion& region, TVG_UNUSED Colo
...
+        // new scene task must be masked
+        newSceneTask->compose->masked = true;
+        newSceneTask->compose->method = method;
+        newSceneTask->compose->opacity = opacity;
+        newSceneTask->compose->blend = BlendMethod::Normal;
+        newSceneTask->renderTargetMsk = curSceneTask->renderTarget;
+        newSceneTask->renderTargetDst = curSceneTask->renderTargetDst;
+        // cur scene task must not be blended
+        curSceneTask->renderTargetDst = nullptr;
+    } else // setup blending scenes configuration
```
hermet, line 484: Please use a convention
```
//....
} else if { 
```
or
```
} else if {   //....
```

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -440,63 +440,77 @@ RenderCompositor* WgRenderer::target(const RenderRegion& region, TVG_UNUSED Colo
...
-    mRenderTargetStack.push(renderTarget);
-    // create and setup new scene task
-    WgSceneTask* sceneTask = new WgSceneTask(renderTarget, compose, sceneTaskCurrent);
-    sceneTaskCurrent->children.push(sceneTask);
-    mRenderTaskList.push(sceneTask);
-    mSceneTaskStack.push(sceneTask);
+    WgRenderTarget* newRenderTarget = mRenderTargetPool.allocate(mContext);
+    mRenderTargetStack.push(newRenderTarget);
+    // current and new scenes
+    WgSceneTask* curSceneTask = mSceneTaskStack.last();
```
hermet: please use `auto`

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -440,63 +440,77 @@ RenderCompositor* WgRenderer::target(const RenderRegion& region, TVG_UNUSED Colo
...
+    WgSceneTask* newSceneTask = new WgSceneTask(newRenderTarget, compose, curSceneTask);
```
hermet: please use auto

hermet (approved):

hermet:
@SergeyLebedkin @wenjieshen please consider to fix gl engine as well. Thanks.

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -438,65 +438,79 @@ WgRenderer::~WgRenderer()
...
+        newSceneTask->compose->masked = true;
+        newSceneTask->compose->method = method;
+        newSceneTask->compose->opacity = opacity;
+        newSceneTask->compose->blend = BlendMethod::Normal;
+        newSceneTask->renderTargetMsk = curSceneTask->renderTarget;
+        newSceneTask->renderTargetDst = curSceneTask->renderTargetDst;
+        // cur scene task must not be blended
+        curSceneTask->renderTargetDst = nullptr;
+        // setup blending scenes configuration
+    } else if (method == MaskMethod::None) {
```
hermet: I think this is enough `else {`

PR #3917, opened 2025-10-28
examples: add font asset resolver

tinyjin:
CleanShot 2025-11-03 at 20 30 48@2x

## 1. Improve sample for Image Asset Resolver
The previous resolver example used a valid image path (logo.png) in the lottie file, which meant the image was already resolved during the parser/prepare stage. This caused the asset resolver callback to never be invoked, making the example ineffective for demonstrating the feature.

## 2. Add sample for Font Asset Resolver

Font data is loaded from external paths using the asset resolver.


examples/LottieExtension.cpp
```diff
@@ -269,9 +269,11 @@ struct UserExample : tvgexam::Example
             auto picture = resolver->picture();
 
             if (!tvgexam::verify(picture->resolver([](tvg::Paint* p, const char* src, void* data) {
-                if (p->type() != tvg::Type::Picture) return false;     //supposed to be a picture object
-                auto ret = static_cast<tvg::Picture*>(p)->load(src);   //load picture resources as demand
-                return (ret == (tvg::Result) 0) ? true : false;        //return true if resolving is successful
+                if (p->type() != tvg::Type::Picture) return false;                              //supposed to be a picture object
+                const char *ext = src + strlen(src) - 3;
```
hermet: 273, 274 looks not proper showcase here. does this really necessary?

hermet (commented):
@tinyjin please check a comment. thanks.

examples/LottieExtension.cpp
tinyjin: This handling is up to user side. In this demonstration I think not really necessary. Remove them out, thanks.

examples/LottieExtension.cpp
```diff
@@ -269,9 +269,9 @@ struct UserExample : tvgexam::Example
             auto picture = resolver->picture();
 
             if (!tvgexam::verify(picture->resolver([](tvg::Paint* p, const char* src, void* data) {
-                if (p->type() != tvg::Type::Picture) return false;     //supposed to be a picture object
-                auto ret = static_cast<tvg::Picture*>(p)->load(src);   //load picture resources as demand
```
hermet: This example needs to demonstrate users can load an image with a given source (src)

tinyjin: @hermet If the given source is valid and can be loaded through the asset resolver, it won’t fall back to the resolver since it’s already marked as resolved in LottieImage::prepare().

I assumed that marking was part of the intended logic. In that case, the example should be updated to properly demonstrate the use of the asset resolver.

Alternatively, we could consider deferring the execution of LottieImage::prepare() to within LottieBuilder, under the resolver check condition.

tinyjin:
@hermet I’m using the src value to remap the path. The updated example now demonstrates an unresolvable image issue caused by an incorrect location. Thanks for your feedback, please have a look again.

PR #3923, opened 2025-10-31
wasm: add asset resolver binding for image/font resolution

tinyjin:
## Overview
Expose the AssetResolver API to WebAssembly bindings, enabling JavaScript applications to provide custom asset resolution logic.

This allows dynamic loading of external resources (images, fonts) that are referenced in Lottie animations but not embedded, giving applications control over resource handling.

issue: https://github.com/thorvg/thorvg.web/issues/153

## Key Mechanism

CleanShot 2025-11-03 at 19 45 57@2x

Unlike native implementations where JS would need direct access to Picture and load functions (requiring extensive function mapping), this web impl simplifies the interface:

- WASM sends asset information (source path + JS user data) to JavaScript
- JavaScript handles fetching and converting assets to binary data
  - _WASM doesn't need 1:1 mapping for picture::load, text::load_
  - JavaScript can implement custom loading strategies (network, storage, etc.)
- JavaScript returns a standardized Blob object format: `{ filename: string, buffer: ArrayBuffer, mimetype: string }`
- WASM performs the actual loading into Picture or Text objects based on `mimetype`

This approach keeps the binding interface concise while maintaining full flexibility.

## Implementation

Since file buffers and user data are defined by JavaScript, a separate WasmAssetResolver implementation was introduced to handle JS value types efficiently.

- Converts JavaScript callbacks and user data members into C++ members and passes to animation during the `load()`.
- Uses `val` type instead of `void*` to simplify memory management and improve type safety

## Examples
| Font | Image |
|:--|:--|
|  |  |

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -32,8 +32,13 @@ using namespace tvg;
 
 EMSCRIPTEN_DECLARE_VAL_TYPE(ArrayBuffer);
 EMSCRIPTEN_DECLARE_VAL_TYPE(Float32Array);
+EMSCRIPTEN_DECLARE_VAL_TYPE(AssetResolverCallback);
 
 static const char* NoError = "None";
+
+static bool isValidProperty(const val& obj, const char* propName) {
+    return obj.hasOwnProperty(propName) && !obj[propName].isUndefined() && !obj[propName].isNull();
+}
 
 struct TvgEngineMethod
 {
@@ -322,6 +327,8 @@ public:
             filetype = "lottie+json";
         }
 
+        animation->picture()->resolver(resolver.func, &resolver.data);
+
         if (animation->picture()->load(data.c_str(), data.size(), filetype.c_str()) != Result::Success) {
             errorMsg = "load() fail";
             return false;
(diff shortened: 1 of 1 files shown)


src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -508,6 +517,45 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
...
+            auto mimetype = res["mimetype"].as<string>();
+
+            if (mimetype == "ttf") {
+                if (p->type() != Type::Text) return false;
+                if (Text::load(name.c_str(), buffer.data(), buffer.size(), mimetype.c_str(), false) != Result::Success) return false;
+                return static_cast<Text*>(p)->font(name.c_str()) == Result::Success;
+            }
+
+            if (p->type() != Type::Picture) return false;     
+            return static_cast<Picture*>(p)->load(buffer.data(), buffer.size(), mimetype.c_str(), nullptr, false) == Result::Success;
```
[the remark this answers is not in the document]
tinyjin: otf, woff are not supported atm

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -517,6 +565,7 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
     uint32_t               height = 0;
     float                  psize[2];         //picture size
     bool                   updated = false;
+    WasmAssetResolver*     resolver = nullptr;
```
hermet: There is no point to use heap memory. 
-> `WasmAssetResolver     resolver;`

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -508,6 +517,45 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
...
+        auto func = [callback](Paint* p, const char* src, void* data) -> bool {
+            auto& userData = *static_cast<val*>(data);
+            auto res = callback(val(src), userData);
+            if (res.isUndefined() || res.isNull()) return false;
+
+            auto name = res["name"].as<string>();
+            auto buffer = vecFromJSArray<char>(res["buffer"]);
+            auto mimetype = res["mimetype"].as<string>();
+
+            if (mimetype == "ttf") {
```
hermet: It's good to  remove the "ttf" format dependency.
```
if Type ==Picture   picture load
if Type == Text       text load
```

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -517,6 +565,7 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
     uint32_t               height = 0;
     float                  psize[2];         //picture size
     bool                   updated = false;
+    WasmAssetResolver      resolver;
```
hermet:
```
struct {
    std::function<bool(Paint* paint, const char* src, void* data)> func;
    val data; //user data for JS data type
} resolver; 
```

hermet (commented):
Please check for one more neat code

hermet (approved):

Issue #2920, opened 2024-10-31
gl/wg: quality enhancement?


hermet:
Bridge strings are not as clear compared to the software rasterizer.

wg:
![image](https://github.com/user-attachments/assets/23866a52-458c-45c7-a413-ba8ad4873357)

gl:
![image](https://github.com/user-attachments/assets/4733eed1-1a56-481c-92d0-0ec19ebacc64)

sw:
![image](https://github.com/user-attachments/assets/b7f0c554-d4c1-44f2-b243-dfaa4c941576)

hermet:
@RuiwenTang any opinion?

RuiwenTang:
Seems still has problem with the rendering of thin lines. Which svg file is this ?

hermet:
> Seems still has problem with the rendering of thin lines. Which svg file is this ?

@RuiwenTang `examples/resources/svg/svg2009.svg`

SpineyPete:
This is a semi-analytic technique for thin uniform stroke-like objects, you prevent the stroke from becoming thinner than a pixel and instead fade the alpha using a coverage metric.

I don't know how applicable it is to a 2d context. It also doesn't solve aliasing but it makes them considerably less objectionable for thin strokes like in the example. Of course it relies on the shape being a stroke in the first place. Conflation artefacts might become an issue, not sure there.

https://www.humus.name/index.php?page=3D&ID=89

wenjieshen:
I profiled this with Nsight in pixel space and confirmed two core issues:

1. The generated triangles are sometimes < 0.5px wide.
2. Some degenerate triangles are generated.

When the stroke width falls below a pixel, like this, rasterization becomes unstable and we get flicker/dropout instead of a consistent thin line.

The approach I'd like to try is similar to "phone-wire AA":

- Treat the stroke as a wire with a radius.
- If the transformed radius would make the wire < 1px wide in device space, clamp the geometry to 1px instead of letting it shrink further, and check the join and round to eliminate degenerate triangles.
- Scale alpha by radius reduction ratio so visually it still reads thinner than 1px instead of looking like a bold 1px bar.
- Very thin cases could go through a lightweight thin-line path that uses analytic AA along straight segments, rather than trying to analytically AA the full complex curve logic.

<img width="2612" height="1447" alt="Image" src="https://github.com/user-attachments/assets/16309e35-323b-4055-a965-fc1ec653c8d3" />

wenjieshen:
[...]

PR #3926, opened 2025-11-02
gl: handle degenerate paths as strokes

wenjieshen:
### Add support for rendering degenerate paths (2-point paths) as strokes in GL engine:
- Add RenderPath::optimize() to simplify paths by merging close points and flattening curves
- Add math utilities: shouldMergePts(), checkLine(), validateCubic(), checkLinePts()
- Add tesselateLineOnly() to handle 2-point paths with MIN_GL_STROKE_ALPHA opacity
- Update GlRenderer::prepare() to detect and handle degenerate paths

<figure>
Screenshot 2025-11-03 at 1 19 45 AM
<figcaption>Some pixels dropped</figcaption>
</figure>
<figure>
Screenshot 2025-11-03 at 1 14 09 AM
<figcaption>Resolved</figcaption>
</figure>

### Test
- Native System: The FPS of Feature branch is consistently **~+4.1%** above Main branch overall, peaking in the early section (**~+5.2%** around 60–200), tapering to **~+3.7%** by the 700–1000 range, while startup is roughly **+1.1%** slower on Feature branch
<figure>
output
<figcaption>Lottie Example 1000 Frames Comparison</figcaption>
</figure>

- WebGL: I haven't collected the exact performance data yet, but I haven't found any critical issues.

Related Issue: #2920

### DIFF
--- a/src/common/tvgMath.h
+++ b/src/common/tvgMath.h
@@ -393,6 +393,14 @@ static inline void log(const Point& pt)
 }
 
 
+static inline bool closed(const Point& lhs, const Point& rhs, float tolerance)
+{
+    float dx = lhs.x - rhs.x;
+    float dy = lhs.y - rhs.y;
+    return (dx * dx + dy * dy) < (tolerance * tolerance);
+}
+
+
 /************************************************************************/
 /* Line functions                                                       */
 /************************************************************************/
--- a/src/renderer/gl_engine/tvgGlCommon.h
+++ b/src/renderer/gl_engine/tvgGlCommon.h
@@ -29,6 +29,7 @@
 #include "tvgMath.h"
 
 #define MIN_GL_STROKE_WIDTH 1.0f
+#define MIN_GL_STROKE_ALPHA 0.25f
 
 #define MVP_MATRIX(w, h)                \
     float mvp[4*4] = {                  \
@@ -106,18 +107,23 @@ struct GlGeometryBuffer {
 
 struct GlGeometry
 {
-    bool tesselateShape(const RenderShape& rshape);
+    void prepare(const RenderShape& rshape);
+    bool tesselateShape(const RenderShape& rshape, float* opacityMultiplier = nullptr);
     bool tesselateStroke(const RenderShape& rshape);
+    bool tesselateLine(const RenderPath& path);
     void tesselateImage(const RenderSurface* image);
     bool draw(GlRenderTask* task, GlStageBuffer* gpuBuffer, RenderUpdateFlag flag);
     GlStencilMode getStencilMode(RenderUpdateFlag flag);
     RenderRegion getBounds() const;
+    void optimizePath(const RenderPath& path, const Matrix& transform);
 
     GlGeometryBuffer fill, stroke;
     Matrix matrix = {};
     RenderRegion viewport = {};
     RenderRegion bounds = {};
     FillRule fillRule = FillRule::NonZero;
+    RenderPath optimizedPath;
+    bool optimized = false;
 };
 
 
--- a/src/renderer/gl_engine/tvgGlGeometry.cpp
+++ b/src/renderer/gl_engine/tvgGlGeometry.cpp
@@ -25,6 +25,153 @@
 #include "tvgGlTessellator.h"
 #include "tvgGlRenderTask.h"
 
+// Optimize path in screen space with merging collinear lines, collapsing zero length lines, and removing unnecessary cubic beziers.
+static void _optimize(const RenderPath& in, RenderPath& out, const Matrix& matrix)
+{
+    static constexpr auto PX_TOLERANCE = 0.25f;
+
+    out.cmds.reserve(in.cmds.count);
+    out.pts.reserve(in.pts.count);
+
+    auto cmds = in.cmds.data;
+    auto cmdCnt = in.cmds.count;
+    auto pts = in.pts.data;
+
+    uint32_t prevIdx = 0;
+    uint32_t prevPrevIdx = 0;
+    auto hasPrevPrev = false;
+
+    // The suffix "T" indicates that the point is transformed.
+    Point lastOutT, prevOutT;
+
+    auto isIdentity = tvg::identity(&matrix);
+
+    auto point2Line = [](const Point& point, const Point& start, const Point& vec, const float vecLenSq, float& maxDist, float& minT, float& maxT) {
+        Point offset = point - start;
+        auto area = tvg::cross(vec, offset);
+        auto dist = fabsf(area) / sqrtf(vecLenSq); // if lineVecLenSq == 0, mean closed() hasn't been called
+        if (dist > maxDist) maxDist = dist;
+
+        auto t = tvg::dot(offset, vec) / vecLenSq;
+        if (t < minT) minT = t;
+        if (t > maxT) maxT = t;
+    };
+
+    auto validateCubic = [&point2Line](const Point& start, const Point& ctrl1, const Point& ctrl2, const Point& end, float& maxDist, float& minT, float& maxT) {
+        maxDist = 0.0f;
+        minT = FLT_MAX;
+        maxT = FLT_MIN;
+        auto vec = end - start;
+        auto vecLenSq = vec.x * vec.x + vec.y * vec.y;
+        point2Line(ctrl1, start, vec, vecLenSq, maxDist, minT, maxT);
+        point2Line(ctrl2, start, vec, vecLenSq, maxDist, minT, maxT);
+    };
+
+    auto point2LineSimple = [](const Point& point, const Point& start, const Point& end, float& dist, float& t) {
+        auto vec = end - start;
+        auto vecLenSq = vec.x * vec.x + vec.y * vec.y;
+        Point offset = point - start;
+        auto area = tvg::cross(vec, offset);
+        dist = fabsf(area) / sqrtf(vecLenSq); // if lineVecLenSq == 0, mean closed() hasn't been called
+        t = tvg::dot(offset, vec) / vecLenSq;
+    };
+
+    auto addLineCmd = [&](const Point& pt, const Point& ptT) {
+        out.cmds.push(PathCommand::LineTo);
+        out.pts.push(pt);
+        prevOutT = lastOutT;
+        lastOutT = ptT;
+        prevPrevIdx = prevIdx;
+        prevIdx = out.pts.count - 1;
+        hasPrevPrev = true;
+    };
+
+    auto processLineCollinear = [&](const Point& startT, const Point& pt, const Point& ptT) {
+        if (!hasPrevPrev || out.pts.count <= 1) {
+            addLineCmd(pt, ptT);
+            return;
+        }
+
+        float dist, t;
+        point2LineSimple(ptT, prevOutT, startT, dist, t);
+
+        if (dist > PX_TOLERANCE) {
+            addLineCmd(pt, ptT);
+            return;
+        }
+
+        if (t < -PX_TOLERANCE) {
+            out.pts[prevPrevIdx] = pt;
+            lastOutT = ptT;
+        } else if (t > 1.0f + PX_TOLERANCE) {
+            out.pts[prevIdx] = pt;
+            lastOutT = ptT;
+        }
+    };
+
+    auto processCubicTo = [&](const Point* cubicPts, const Point& startT) {
+        auto endT = isIdentity ? cubicPts[2] : cubicPts[2] * matrix;
+        if (tvg::closed(startT, endT, PX_TOLERANCE)) return;
+        auto ctrl1T = isIdentity ? cubicPts[0] : cubicPts[0] * matrix;
+        auto ctrl2T = isIdentity ? cubicPts[1] : cubicPts[1] * matrix;
+        float maxDist, minT, maxT;
+        validateCubic(startT, ctrl1T, ctrl2T, endT, maxDist, minT, maxT);
+
+        bool flatEnough  = (maxDist <= PX_TOLERANCE);
+        bool inSpan = (minT >= -PX_TOLERANCE) && (maxT <= 1.0f + PX_TOLERANCE);
+        if (flatEnough && inSpan) {
+            processLineCollinear(startT, cubicPts[2], endT);
+        } else {
+            out.cmds.push(PathCommand::CubicTo);
+            out.pts.push(cubicPts[0]);
+            out.pts.push(cubicPts[1]);
+            out.pts.push(cubicPts[2]);
+            prevOutT = lastOutT;
+            lastOutT = endT;
+            prevPrevIdx = prevIdx;
+            prevIdx = out.pts.count - 1;
+            hasPrevPrev = true;
+        }
+    };
+
+    for (uint32_t i = 0; i < cmdCnt; i++) {
+        switch (cmds[i]) {
+            case PathCommand::MoveTo: {
+                out.cmds.push(PathCommand::MoveTo);
+                out.pts.push(*pts);
+                lastOutT = isIdentity ? *pts : *pts * matrix;
+                prevIdx = out.pts.count - 1;
+                hasPrevPrev = false;
+                pts++;
+                break;
+            }
+            case PathCommand::LineTo: {
+                auto startT = lastOutT;
+                auto ptT = isIdentity ? *pts : (*pts) * matrix;
+                if (tvg::closed(startT, ptT, PX_TOLERANCE)) {
+                    pts++;
+                    break;
+                }
+                processLineCollinear(startT, *pts, ptT);
+                pts++;
+                break;
+            }
+            case PathCommand::CubicTo: {
+                processCubicTo(pts, lastOutT);
+                pts += 3;
+                break;
+            }
+            case PathCommand::Close: {
+                out.cmds.push(PathCommand::Close);
+                hasPrevPrev = false;
+                break;
+            }
+            default:
+                break;
+            }
+    }
+}
+
 bool GlIntersector::isPointInTriangle(const Point& p, const Point& a, const Point& b, const Point& c)
 {
     auto d1 = tvg::cross(p - a, p - b);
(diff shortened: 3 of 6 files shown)


meson.build
```diff
@@ -169,6 +169,7 @@ endif
 
 if get_option('tests')
    subdir('test')
+   subdir('test/internal')
```
hermet: Let's remove internal-methods test, we complete the test-suite only for stable APIs. Thanks.

wenjieshen: Got it! I have remove internal-methods test. Thank you.

src/common/tvgMath.h
```diff
@@ -386,6 +386,54 @@ static inline void log(const Point& pt)
...
+
+static inline void checkLinePts(const Point& p, const Point& p0, const Point& p1, float& dist, float& t)
+{
+    auto v = p1 - p0;
+    auto vv = v.x * v.x + v.y * v.y;
+    Point w = p - p0;
+    float area = cross(v, w);
+    dist = fabsf(area) / sqrtf(vv);
+    t = dot(w, v) / vv;
+}
```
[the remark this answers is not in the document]
wenjieshen: if vv == 0, mean shouldMergePts() hasn't been called.

src/renderer/gl_engine/tvgGlCommon.h
```diff
@@ -29,6 +29,7 @@
 #include "tvgMath.h"
 
 #define MIN_GL_STROKE_WIDTH 1.0f
+static constexpr float MIN_GL_STROKE_ALPHA = 0.5f;
```
hermet: Please ensure consistency. You can use #define like in `MIN_GL_STROKE_WIDTH` above.

src/renderer/tvgRender.cpp
```diff
@@ -102,6 +102,125 @@ bool RenderPath::bounds(const Matrix* m, BBox& box)
     return true;
 }
 
+
+void RenderPath::optimize(const RenderPath& in, RenderPath& out, const Matrix& matrix)
```
hermet: For now, you can place this in the GL engine only to avoid increasing the binary size in the software engine.

This kind of method definitely needs a description:
What is being optimized, and how?

```
// describe this method purpose and mechansim.
// if any reference you can add the link or information source.
void optimize(...)
```

src/common/tvgMath.h
```diff
@@ -386,6 +386,50 @@ static inline void log(const Point& pt)
 }
 
 
+static inline bool shouldMergePts(const Point& p1, const Point& p2, float tolerance)
```
hermet: 1. merge is too specific purpose. I think we can make the function more generic like:
-> `bool closed(lhs, rhs, tolerance)`

3. Please relocate this around 226 line for putting points methods together.

src/common/tvgMath.h
```diff
@@ -386,6 +386,50 @@ static inline void log(const Point& pt)
...
+{
+    float dx = p1.x - p2.x;
+    float dy = p1.y - p2.y;
+    return (dx * dx + dy * dy) < (tolerance * tolerance);
+}
+
+
+static inline void checkLine(const Point& p, const Point& p0, const Point& v, const float vv, float& maxDist, float& minT, float& maxT)
```
hermet: p, p0, v, vv? 
with the function prototype, I can't understand the function purpose... could you please suggest more a clear method name and parameters?

src/common/tvgMath.h
```diff
@@ -386,6 +386,50 @@ static inline void log(const Point& pt)
...
+    minT = FLT_MAX;
+    maxT = FLT_MIN;
+    auto v = p3 - p0;
+    auto vv = v.x * v.x + v.y * v.y;
+    checkLine(cp1, p0, v, vv, maxDist, minT, maxT);
+    checkLine(cp2, p0, v, vv, maxDist, minT, maxT);
+}
+
+
+static inline void checkLinePts(const Point& p, const Point& p0, const Point& p1, float& dist, float& t)
```
hermet: `void point2Line(p0, p1, p2,  dist, t);`

src/renderer/tvgRender.cpp
wenjieshen: Thank you, I have moved it to be a static internal function.

src/common/tvgMath.h
wenjieshen: Sorry for the inconvenience. I have updated the names to make them more meaningful.

wenjieshen:
The time measurement has been updated to allow stroke tessellation to take the optimized path. Performance has improved on the test example.

src/renderer/gl_engine/tvgGlCommon.h
```diff
@@ -106,8 +107,9 @@ struct GlGeometryBuffer {
 
 struct GlGeometry
 {
-    bool tesselateShape(const RenderShape& rshape);
-    bool tesselateStroke(const RenderShape& rshape);
+    bool tesselateShape(const RenderShape& rshape, const RenderPath& path);
+    bool tesselateStroke(const RenderShape& rshape, const RenderPath& path);
+    bool tesselateLineOnly(const RenderPath& path);
```
hermet: -> `tesselateLine()`

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1251,15 +1253,28 @@ RenderData GlRenderer::prepare(const RenderShape& rshape, RenderData data, const
     sdata->opacity = opacity;
     sdata->geometry.matrix = transform;
     sdata->geometry.viewport = vport;
-
+    RenderPath optPath;
+    if (rshape.path.pts.count > 2 &&(flags & (RenderUpdateFlag::Path | RenderUpdateFlag::Transform | RenderUpdateFlag::Color | RenderUpdateFlag::Gradient | RenderUpdateFlag::Stroke | RenderUpdateFlag::GradientStroke)) ){
```
hermet: this `if` condition looks not essential...

src/common/tvgMath.h
```diff
@@ -386,6 +386,50 @@ static inline void log(const Point& pt)
...
+    auto dist = fabsf(area) / sqrtf(vecLenSq); // if lineVecLenSq == 0, mean closed() hasn't been called
+    if (dist > maxDist) maxDist = dist;
+
+    auto t = dot(offset, vec) / vecLenSq;
+    if (t < minT) minT = t;
+    if (t > maxT) maxT = t;
+}
+
+
+static inline void validateCubic(const Point& start, const Point& ctrl1, const Point& ctrl2, const Point& end, float& maxDist, float& minT, float& maxT)
```
hermet: Can put this in `Bezier` class as a util method?

src/renderer/gl_engine/tvgGlTessellator.cpp
```diff
@@ -33,12 +33,17 @@ static uint32_t _pushVertex(Array<float>& array, float x, float y)
 }
 
 
+Stroker::Stroker(GlGeometryBuffer* buffer, float wdith, StrokeCap cap, StrokeJoin join) : mBuffer(buffer), mWidth(wdith), mCap(cap), mJoin(join)
```
hermet: i think we can keep only this constructor.
`wdith` -> `width`

hermet (commented):
Could you please review once more? It seems Geometry can handle the optimization not in GlRenderer, Those optimization could be nicely applied in tesselateShape() / tesselateStroke(). Thanks.

wenjieshen:
> Could you please review once more? It seems Geometry can handle the optimization not in GlRenderer, Those optimization could be nicely applied in tesselateShape() / tesselateStroke(). Thanks.

HI @hermet, I appreciate your suggestion. I created a new member variable in the GLGeometry class for the optimized path, which reduces interface changes and improves performance. However, the only downside is the expansion of memory life scope. I also update the time measure of the latest comparsion in the description of PR.

src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -134,17 +254,32 @@ bool GlIntersector::intersectImage(const RenderRegion region, const GlShape* ima
 }
 
 
+void GlGeometry::updateOptimizedPath(const RenderPath& path, const Matrix& transform)
+{
+    optimizedPath.cmds.clear();
+    optimizedPath.pts.clear();
+    if (path.pts.count > 2) {
```
hermet: Single point is not allowed in rendering, I think you can remove this condition.
The exception handling must be applied earlier, not here.

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1279,14 +1279,29 @@ RenderData GlRenderer::prepare(const RenderShape& rshape, RenderData data, const
     sdata->geometry.matrix = transform;
     sdata->geometry.viewport = vport;
 
+    if (flags & (RenderUpdateFlag::Path | RenderUpdateFlag::Transform)) {
```
hermet, line 1282: We can expect that this optimization is hidden by GlGoemetry.
It would be better if the optimization could be performed earlier, inside `tessellateShape()` or `tessellateStroke()`, tessllateLine() Or just some like a call `geometry.prepare()`

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1279,14 +1279,29 @@ RenderData GlRenderer::prepare(const RenderShape& rshape, RenderData data, const
...
+        sdata->geometry.updateOptimizedPath(rshape.path, transform);
+    }
+
     //TODO: Please precisely update tessellation not to update only if the color is changed.
     if (flags & (RenderUpdateFlag::Color | RenderUpdateFlag::Gradient | RenderUpdateFlag::Transform | RenderUpdateFlag::Path)) {
-        if (sdata->geometry.tesselateShape(rshape)) sdata->validFill = true;
+        const auto path2Use = sdata->geometry.hasOptimizedPath ? sdata->geometry.optimizedPath : rshape.path;
+        if (path2Use.pts.count > 2) {
+            if (sdata->geometry.tesselateShape(*(sdata->rshape))) sdata->validFill = true;
+        } else if (path2Use.pts.count == 2 && tvg::zero(sdata->rshape->strokeWidth())) {
```
hermet: Could you please explain more about this condition with a comment?  I can't hardly understand a user scenario which needs to 2 points shape handle here.

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1279,14 +1279,29 @@ RenderData GlRenderer::prepare(const RenderShape& rshape, RenderData data, const
...
     sdata->geometry.viewport = vport;
 
+    if (flags & (RenderUpdateFlag::Path | RenderUpdateFlag::Transform)) {
+        sdata->geometry.updateOptimizedPath(rshape.path, transform);
+    }
+
     //TODO: Please precisely update tessellation not to update only if the color is changed.
     if (flags & (RenderUpdateFlag::Color | RenderUpdateFlag::Gradient | RenderUpdateFlag::Transform | RenderUpdateFlag::Path)) {
-        if (sdata->geometry.tesselateShape(rshape)) sdata->validFill = true;
+        const auto path2Use = sdata->geometry.hasOptimizedPath ? sdata->geometry.optimizedPath : rshape.path;
```
hermet: It's good to hide this kind of complex class usage by `GlGeometry`.
The renderer should ideally just call `tessellateShape()` here as it does, and inside that function, the logic can determine the appropriate condition and perform either shape or line tessellation — whatever is needed.

hermet:
Only one concern point is that, this additional optimization process O(n) could affect to the normal rendering as a downside...

wenjieshen:
> Only one concern point is that, this additional optimization process O(n) could affect to the normal rendering as a downside...

I was quite surprised by the improvement in performance in this case. I initially thought we would need to make some trade-offs. However, it seems that this test case may have too many paths optmized, which resulted in a reduction in tesselation and the number of triangles. Once the version becomes stable, should I put in more effort into conducting various types of testing?

src/renderer/gl_engine/tvgGlRenderer.cpp
wenjieshen: The optimization is now fully integrated into the `GLGeometry::prepare` function, making it completely hidden from the renderer layer.

src/renderer/gl_engine/tvgGlRenderer.cpp
wenjieshen: When the current transform matrix scales a filled path down to device space, the path can degenerate on screen. In that case parts of it become thinner than one pixel. I added an inline comment and a small text visualization to make this clearer.

src/renderer/gl_engine/tvgGlRenderer.cpp
wenjieshen: Thank you. This code snippet has been integrated into the `GLGeometry::tesselateShape` function.

hermet:
> I was quite surprised by the improvement in performance in this case. I initially thought we would need to make some trade-offs. However, it seems that this test case may have too many paths optmized, which resulted in a reduction in tesselation and the number of triangles. Once the version becomes stable, should I put in more effort into conducting various types of testing?

@wenjieshen Luckily, the enhancement occurred in our more complex samples. However, logically, I assume it could lead to a performance downside when rendering clean and simple paths. It would be useful to observe the performance difference by generating thousands of random rectangle animations (i.e. https://tan-morgan-5.tiiny.site/) and see the results.

wenjieshen:
> 
> @wenjieshen Luckily, the enhancement occurred in our more complex samples. However, logically, I assume it could lead to a performance downside when rendering clean and simple paths. It would be useful to observe the performance difference by generating thousands of random rectangle animations (i.e. https://tan-morgan-5.tiiny.site/) and see the results.

@hermet I have completed the test. Thank you.
The feature branch shows a consistent, but small, performance regression compared to the combined main branch. The performance drop is evident at every workload, ranging from -0.42% to -1.27%.

Workload<br> Rectangles | Relative Change (AvgFPS) | Main Samples (Every 100 Frames) | Main MinFPS | Main AvgFPS | Main MaxFPS | Feature Samples (Every 100 Frames) | Feature MinFPS | Feature AvgFPS | Feature MaxFPS
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
6,000 | -0.44% | 128 | 50.97 | 53.55 | 54.02 | 127 | 50.4 | 53.32 | 53.76
9,000 | -0.82% | 85 | 33.82 | 35.8 | 36.24 | 84 | 33.64 | 35.51 | 35.82
12,000 | -1.18% | 64 | 25.51 | 26.88 | 27.18 | 63 | 25.09 | 26.56 | 26.93
15,000 | -0.42% | 50 | 20.04 | 21.23 | 21.68 | 50 | 20.04 | 21.14 | 21.46
18,000 | -1.27% | 42 | 17.06 | 17.72 | 17.83 | 41 | 16.35 | 17.49 | 17.74





Screenshot 2025-11-07 at 12 30 28 PM

wenjieshen:
@hermet here is the source code of the test case, thank you 🙏
[RectangleAnimations.cpp](https://github.com/user-attachments/files/23456158/RectangleAnimations.cpp)

src/common/tvgMath.h
```diff
@@ -393,6 +393,50 @@ static inline void log(const Point& pt)
...
```
hermet: Please inline `validateCubic()` and `point2Line()` into optimize() using closure style.

src/common/tvgMath.h
```diff
@@ -393,6 +393,50 @@ static inline void log(const Point& pt)
...
+    minT = FLT_MAX;
+    maxT = FLT_MIN;
+    auto vec = end - start;
+    auto vecLenSq = vec.x * vec.x + vec.y * vec.y;
+    point2Line(ctrl1, start, vec, vecLenSq, maxDist, minT, maxT);
+    point2Line(ctrl2, start, vec, vecLenSq, maxDist, minT, maxT);
+}
+
+
+static inline void point2Line(const Point& point, const Point& start, const Point& end, float& dist, float& t)
```
hermet: please remove inline and add implementation to cpp file.

src/renderer/gl_engine/tvgGlCommon.h
```diff
@@ -106,18 +107,23 @@ struct GlGeometryBuffer {
...
     GlStencilMode getStencilMode(RenderUpdateFlag flag);
     RenderRegion getBounds() const;
+    void updateOptimizedPath(const RenderPath& path, const Matrix& transform);
 
     GlGeometryBuffer fill, stroke;
     Matrix matrix = {};
     RenderRegion viewport = {};
     RenderRegion bounds = {};
     FillRule fillRule = FillRule::NonZero;
+    RenderPath optimizedPath = {};
```
hermet: `RenderPath optimizedPath;`

src/renderer/gl_engine/tvgGlCommon.h
```diff
@@ -106,18 +107,23 @@ struct GlGeometryBuffer {
...
-    bool tesselateShape(const RenderShape& rshape);
+    void prepare(const RenderShape& rshape);
+    bool tesselateShape(const RenderShape& rshape, float* opacityMultiplier = nullptr);
     bool tesselateStroke(const RenderShape& rshape);
+    bool tesselateLine(const RenderPath& path);
     void tesselateImage(const RenderSurface* image);
     bool draw(GlRenderTask* task, GlStageBuffer* gpuBuffer, RenderUpdateFlag flag);
     GlStencilMode getStencilMode(RenderUpdateFlag flag);
     RenderRegion getBounds() const;
+    void updateOptimizedPath(const RenderPath& path, const Matrix& transform);
```
hermet: -> `optimizePath()`

src/renderer/gl_engine/tvgGlCommon.h
```diff
@@ -106,18 +107,23 @@ struct GlGeometryBuffer {
...
 
     GlGeometryBuffer fill, stroke;
     Matrix matrix = {};
     RenderRegion viewport = {};
     RenderRegion bounds = {};
     FillRule fillRule = FillRule::NonZero;
+    RenderPath optimizedPath = {};
+    bool hasOptimizedPath = false;
```
hermet: -> `optimized = false;`

src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -25,6 +25,126 @@
...
+
+    uint32_t prevIdx = 0;
+    uint32_t prevPrevIdx = 0;
+    auto hasPrevPrev = false;
+
+    // The suffix "T" indicates that the point is transformed.
+    Point lastOutT;
+    Point prevOutT;
+
+    bool isIdentity = tvg::identity(&matrix);
```
hermet: bool -> auto

src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -25,6 +25,126 @@
...
+    auto cmds = in.cmds.data;
+    auto cmdCnt = in.cmds.count;
+    auto pts = in.pts.data;
+
+    uint32_t prevIdx = 0;
+    uint32_t prevPrevIdx = 0;
+    auto hasPrevPrev = false;
+
+    // The suffix "T" indicates that the point is transformed.
+    Point lastOutT;
```
hermet: `Point lastOutT, prevOutT`;

src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -25,6 +25,126 @@
...
+            lastOutT = ptT;
+        } else if (t > 1.0f + PX_TOLERANCE) {
+            out.pts[prevIdx] = pt;
+            lastOutT = ptT;
+        }
+    };
+
+    auto processCubicTo = [&](const Point* cubicPts, const Point& startT) {
+        auto endT = isIdentity ? cubicPts[2] : cubicPts[2] * matrix;
+        if (tvg::closed(startT, endT, PX_TOLERANCE)) {
```
hermet: `if (tvg::closed(startT, endT, PX_TOLERANCE))  return;`

src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -134,22 +254,68 @@ bool GlIntersector::intersectImage(const RenderRegion region, const GlShape* ima
...
+void GlGeometry::prepare(const RenderShape& rshape)
+{
+    updateOptimizedPath(rshape.path, matrix);
+}
+
+
+bool GlGeometry::tesselateShape(const RenderShape& rshape, float* opacityMultiplier)
 {
     fill.clear();
+    const RenderPath& path2Use = hasOptimizedPath ? optimizedPath : rshape.path;
```
hermet: `const RenderPath&` -> `const auto&`

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1279,14 +1279,22 @@ RenderData GlRenderer::prepare(const RenderShape& rshape, RenderData data, const
     sdata->geometry.matrix = transform;
     sdata->geometry.viewport = vport;
 
+    if (flags & (RenderUpdateFlag::Path | RenderUpdateFlag::Transform)) {
```
hermet, line 1282: i guess `RenderUpdateFlag::Stroke` also required?

wenjieshen: I have conducted several tests updating strokes as animations in the Stroke and GradientStroke examples. The `RenderUpdateFlag::Stroke` is unnecessary. Thank you. 
Although some gl engine glitches occur, they also happen in the main branch.

PR #3931, opened 2025-11-03
lottie: fixed text rendering when font is missing

### LINKED ISSUE #3930: lottie: TextDoc font does not fallback to default when unavailable
when a specified font is not available, the default font is not used as a fallback

tested using ThorVG viewer, the text does not render with a fallback font

```json
"d": {
  "sid": "my_text",
  "k": [
    {
      "s": {
        "f": "NotExisting", // font not available in font list
        "fc": [1, 0, 0, 1],
        "s": 200,
        "t": "Text",
        "j": 0
      },
      "t": 0
    }
  ]
}
```

[text.json](https://github.com/user-attachments/files/23302020/text.json)

Image

theashraf:
Removed early return when `text->font` is null, allowing `_fontText()` to apply default system font fallback for undefined fonts.

Fixes #3930

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -928,7 +928,7 @@ static void _fontText(LottieFont* font, TextDocument& doc, Scene* scene, const A
     while (token) {
         auto txt = Text::gen();
         if (txt->font(doc.name) == Result::Success) {}
-        else if (resolver && resolver->func(txt, font->path, resolver->data)) {}
+        else if (font && resolver && resolver->func(txt, font->path, resolver->data)) {}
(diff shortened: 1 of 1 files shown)


hermet:
@theashraf thanks for contribution, I think it needs one more guard in 927 line! 

`else if (resolver && resolver->func(txt, font->path, resolver->data)) {}`
 -> `else if (font && resolver && resolver->func(txt, font->path, resolver->data)) {}`

theashraf:
> @theashraf thanks for contribution, I think it needs one more guard in 927 line!
> 
> `else if (resolver && resolver->func(txt, font->path, resolver->data)) {}` -> `else if (font && resolver && resolver->func(txt, font->path, resolver->data)) {}`

Thanks @hermet for this catch, i've updated the PR

PR #3932, opened 2025-11-03
gl_engine: introduce ability to combine masking and custom blend

SergeyLebedkin:
Until now, blending and masking settings couldn't be used simultaneously. If a mask was specified for an object, the blending settings were ignored, and the object was blended using the normal settings.

Now it's possible to use compositions and blending in the same scene. This is achieved by reconfiguring the current render targets and render tree. 

This doesn't affect overall performance, but only adds functionality.

gl(old)/sw/gl(new)
image

issue: https://github.com/thorvg/thorvg/issues/3793

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -699,6 +699,33 @@ void GlRenderer::prepareCmpTask(GlRenderTask* task, const RenderRegion& vp, uint
 void GlRenderer::endRenderPass(RenderCompositor* cmp)
 {
     auto glCmp = static_cast<GlCompositor*>(cmp);
+    
+    // setup masking and blending render pass configurations
+    if ((glCmp->flags & (tvg::Blending | tvg::Masking)) == (tvg::Blending | tvg::Masking)) {
+        // rearrange render tree
+        auto selfPass = mRenderPassStack.last();
+        mRenderPassStack.pop();
+        auto prevPass = mRenderPassStack.last();
+        mRenderPassStack.pop();
+        auto maskPass = mRenderPassStack.last();
+        mRenderPassStack.pop();
+        mRenderPassStack.push(prevPass);
+        mRenderPassStack.push(maskPass);
+        mRenderPassStack.push(selfPass);
+        // setup composition properties
+        auto prevCompose = mComposeStack.last();
+        auto opacity = glCmp->opacity;
+        auto blendMethod = glCmp->blendMethod;
+        // self scene task must be masked but not blended
+        glCmp->method = prevCompose->method;
+        glCmp->opacity = 255;
+        glCmp->blendMethod = BlendMethod::Normal;
+        // prev scene task must be blended but not masked
+        prevCompose->method = MaskMethod::None;
+        prevCompose->opacity = opacity;
+        prevCompose->blendMethod = blendMethod;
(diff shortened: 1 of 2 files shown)


src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -699,6 +699,33 @@ void GlRenderer::prepareCmpTask(GlRenderTask* task, const RenderRegion& vp, uint
 void GlRenderer::endRenderPass(RenderCompositor* cmp)
 {
     auto glCmp = static_cast<GlCompositor*>(cmp);
+    
+    // setup masking and blending render pass configurations
+    if ((glCmp->flags & (tvg::Blending | tvg::Masking)) == (tvg::Blending | tvg::Masking)) {
```
wenjieshen, line 704:
```suggestion
    if (glCmp && (glCmp->flags & (tvg::Blending | tvg::Masking)) == (tvg::Blending | tvg::Masking)) {
```
It might be helpful to verify glCmp before we dereference it. This way the code stays robust even if the upstream call changes later.

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -699,6 +699,33 @@ void GlRenderer::prepareCmpTask(GlRenderTask* task, const RenderRegion& vp, uint
...
+        mRenderPassStack.pop();
+        auto maskPass = mRenderPassStack.last();
+        mRenderPassStack.pop();
+        mRenderPassStack.push(prevPass);
+        mRenderPassStack.push(maskPass);
+        mRenderPassStack.push(selfPass);
+        // setup composition properties
+        auto prevCompose = mComposeStack.last();
+        auto opacity = glCmp->opacity;
+        auto blendMethod = glCmp->blendMethod;
```
wenjieshen, line 718:
```suggestion
        const auto prevCompose = mComposeStack.last();
        const auto opacity = glCmp->opacity;
        const auto blendMethod = glCmp->blendMethod;
```
Since we don’t modify the stored opacity/blend values afterward, making them const could improve clarity and also allow the compiler to optimize better.

wenjieshen (commented):
Thank you for this contribution! The logic is solid, and the comments make the intent very clear. I only have a few minor suggestions that might help improve safety.

wenjieshen:
Executed all examples on the native system and WebGL; I didn't see any glitches so far.

src/renderer/gl_engine/tvgGlRenderer.cpp
hermet: @wenjieshen if the endRenderPass() is intended to have a valid cmp, null check doesn't helpful. Sometimes, causing it to crash intentionally can help identify the issue more obvious. (rather than broken visual)

src/renderer/gl_engine/tvgGlRenderer.cpp
SergeyLebedkin: @hermet what do you think? is it suitable? we did not use approach like this in our code

hermet: @SergeyLebedkin Trivial, I don't mind, since that optimization would be negligible. It's up to your decision.

wenjieshen (approved):

hermet (approved):

Issue #3653, opened 2025-07-28
wgpu/gl: trimpath optimization


hermet:
Current engines generate TrimPath data separately for each fill and stroke, resulting in exact duplicates.
A reusable TrimPath can be applied to both cases for efficiency.

SergeyLebedkin:
main vs PR
<img width="885" height="962" alt="Image" src="https://github.com/user-attachments/assets/51a2e346-957f-4b76-a291-2f734e5dd5ee" />

wenjieshen:
In the GL engine, tested with the same JSON file.
For the first 1000 frames, the feature branch is consistently but modestly faster:
- ~0.94% higher average FPS (per-interval and accumulated)
- Slightly more stable (lower standard deviation)

Metric | Original | Feature branch | Difference | Relative change
-- | -- | -- | -- | --
Mean per-10-frame FPS | 190.97 fps | 192.77 fps | +1.80 fps | +0.94%
Std dev (per-10-frame FPS) | 67.93 fps | 66.51 fps | -1.42 fps | slightly lower
Min per-10-frame FPS | 27.37 fps | 29.39 fps | +2.02 fps | —
Max per-10-frame FPS | 239.95 fps | 239.86 fps | -0.09 fps | ~same

PR #3935, opened 2025-11-05
wg_engine: trimpath optimization

SergeyLebedkin:
Current engine generate TrimPath data once for fill and stroke. 
The original path used if trimming is not necessary

issue: https://github.com/thorvg/thorvg/issues/3653

### DIFF
--- a/src/renderer/wg_engine/tvgWgRenderData.cpp
+++ b/src/renderer/wg_engine/tvgWgRenderData.cpp
@@ -169,18 +169,18 @@ void WgRenderDataShape::updateMeshes(const RenderShape &rshape, RenderUpdateFlag
     releaseMeshes();
     strokeFirst = rshape.strokeFirst();
 
+    // trim path if necessary
+    RenderPath trimmedPath;
+    auto& path = (rshape.trimpath() && rshape.stroke->trim.trim(rshape.path, trimmedPath)) ? trimmedPath : rshape.path;
+
     // update fill shapes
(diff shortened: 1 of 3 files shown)


src/renderer/wg_engine/tvgWgTessellator.cpp
```diff
@@ -59,15 +52,16 @@ BBox WgStroker::getBBox() const
     return {mLeftTop, mRightBottom};
 }
 
-void WgStroker::run(const RenderPath& path, const Matrix& m)
+void WgStroker::run(const RenderPath* path, const Matrix& m)
```
hermet: Is this change essential?

src/renderer/wg_engine/tvgWgRenderData.cpp
```diff
@@ -169,18 +169,20 @@ void WgRenderDataShape::updateMeshes(const RenderShape &rshape, RenderUpdateFlag
     releaseMeshes();
     strokeFirst = rshape.strokeFirst();
 
+    // trim path if necessary
+    const RenderPath* path = &rshape.path;
```
hermet: `auto& path = (rshape.trimpath() && rshape.stroke->trim.trim(rshape.path, trimmedPath)) ?  trimmedPath : rshape.path;`

hermet (commented):
@SergeyLebedkin please check a comment, I guess you don't need to change pointer/reference styles

hermet:
fps: ~+4%

PR #3938, opened 2025-11-06
wasm: negative dimension handling

BenjaminHalko:
Width and height are now normalized to positive values, and sign is preserved when resizing the animation picture. This ensures correct scaling and positioning for negative input dimensions.

### DIFF
--- a/src/bindings/wasm/tvgWasmLottieAnimation.cpp
+++ b/src/bindings/wasm/tvgWasmLottieAnimation.cpp
@@ -339,6 +339,8 @@ public:
         /* need to reset size to calculate scale in Picture.size internally before calling resize() */
         this->width = 0;
         this->height = 0;
+        this->canvasWidth = 0;
+        this->canvasHeight = 0;
 
         resize(width, height);
 
@@ -358,9 +360,7 @@ public:
 
         if (!canvas || !animation) return ArrayBuffer(val(typed_memory_view<uint8_t>(0, nullptr)));
 
-        int absWidth = std::abs(width);
-        int absHeight = std::abs(height);
-        if (!updated) return engine->output(absWidth, absHeight);
+        if (!updated) return engine->output(canvasWidth, canvasHeight);
 
         if (canvas->draw(true) != Result::Success) {
             errorMsg = "draw() fail";
@@ -371,7 +371,7 @@ public:
 
         updated = false;
 
-        return engine->output(absWidth, absHeight);
+        return engine->output(canvasWidth, canvasHeight);
     }
 
     bool update()
@@ -418,17 +418,17 @@ public:
 
         this->width = width;
         this->height = height;
+        canvasWidth = std::abs(width);
+        canvasHeight = std::abs(height);
 
-        auto absWidth = std::abs(width);
-        auto absHeight = std::abs(height);
         auto wsign = (width < 0) ? -1.0f : 1.0f;
         auto hsign = (height < 0) ? -1.0f : 1.0f;
 
-        engine->resize(canvas, absWidth, absHeight);
+        engine->resize(canvas, canvasWidth, canvasHeight);
 
-        auto scale = (psize[0] > psize[1]) ? absWidth / psize[0] : absHeight / psize[1];
+        auto scale = (psize[0] > psize[1]) ? canvasWidth / psize[0] : canvasHeight / psize[1];
         animation->picture()->size(psize[0] * scale * wsign, psize[1] * scale * hsign);
-        animation->picture()->translate(absWidth * 0.5f, absHeight * 0.5f);
+        animation->picture()->translate(canvasWidth * 0.5f, canvasHeight * 0.5f);
 
 
         updated = true;
@@ -566,6 +566,8 @@ private:
     TvgEngineMethod*       engine = nullptr;
     int32_t                width = 0;
     int32_t                height = 0;
+    uint32_t               canvasWidth = 0;
+    uint32_t               canvasHeight = 0;
     float                  psize[2];         //picture size
     bool                   updated = false;
     struct {


src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -413,13 +413,20 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
...
         this->width = width;
         this->height = height;
 
         engine->resize(canvas, width, height);
 
         auto scale = (psize[0] > psize[1]) ? width / psize[0] : height / psize[1];
-        animation->picture()->scale(scale);
+        animation->picture()->size(
+            psize[0] * scale * wsign,
+            psize[1] * scale * hsign);
```
tinyjin: This binding aims to render the Lottie component within the DOM, and I believe that considering negative values for an element's dimensions is not aligned with user expectations. While alternative fallback strategies using absolute values or sign processing can be considered, I think that incorporating additional calculations to handle invalid cases might inadvertently impact the proper functioning of valid cases.

Therefore, If we need handling for such a negative box size, I believe that an appropriate engine handling approach would be to perform value validation at the start of the function and exit early(_line 410-412_) for invalid inputs. (Ignore values or clamp to zero)

BenjaminHalko: @tinyjin There are cases where users might want to mirror the animation however. Which is why this is required. One of these cases being web support for Unity, where the y scale needs to be flipped to match Unity's texture format.

tinyjin: @BenjaminHalko If it’s required for Unity, I think we can proceed with implementing the mirroring logic on each player’s side as needed (for example, as you mentioned in Unity).

src/bindings/wasm/tvgWasmLottieAnimation.cpp
```diff
@@ -413,13 +413,20 @@ class __attribute__((visibility("default"))) TvgLottieAnimation
 
         canvas->sync();
 
+        auto wsign = (width < 0) ? -1.0f : 1.0f;
+        auto hsign = (height < 0) ? -1.0f : 1.0f;
+        width = std::abs(width);
+        height = std::abs(height);
```
tinyjin: Line 412 implements an early return to avoid reprocessing a size that has already been processed. However, if the parameter values are modified after that condition check, this optimization will be invalidated.

BenjaminHalko: @tinyjin Oh whoops, you are right about that optimization issue.

src/bindings/wasm/tvgWasmLottieAnimation.cpp
BenjaminHalko: @tinyjin 
thorvg.unity is using this wasm api, so are you saying that mirror functionality should be patched in and not added to base thorvg?

BenjaminHalko: Since the `.size()` function in picture already supports negative scaling, I assumed that not having that functionality exposed in the wasm API was simply an oversight. Not something that was supposed to be intentionally restricted.

tinyjin: Please check lines (566, 567). If resize is intended to support negative values, the unsigned variables should not be unsigned.

```
uint32_t               width = 0;
```

tinyjin: @BenjaminHalko Just out of curiosity-- compared to implementing this mirroring on the Unity side, would handling the resizing in the WASM binding be more optimal from Unity’s perspective? We could consider this change based on user-side needs.

BenjaminHalko: @tinyjin I fixed it by adding a separate value for the current picture size vs canvas size. So that we wouldn't also have to `abs` each frame in the render function.

BenjaminHalko: If having an extra set of variables is un-wise, then in the resize function instead of comparing against `this->width/height`, we could compare against the output of getting the picture size. We would still leave the width/height in the wasm object to represent the canvas size, which should always be positive.

```c++
 void resize(int width, int height)
    {
        if (!canvas || !animation) return;

        float curWidth, curHeight;
        animation->picture()->size(&curWidth, &curHeight);
        if (curWidth == width && curHeight == height) return;

        canvas->sync();

        this->width = std::abs(width);
        this->height = std::abs(height);

        auto wsign = (width < 0) ? -1.0f : 1.0f;
        auto hsign = (height < 0) ? -1.0f : 1.0f;

        engine->resize(canvas, this->width, this->height);

        auto scale = (psize[0] > psize[1]) ? this->width / psize[0] : this->height / psize[1];
        animation->picture()->size(psize[0] * scale * wsign, psize[1] * scale * hsign);
        animation->picture()->translate(this->width * 0.5f, this->height * 0.5f);


        updated = true;
    }
```

BenjaminHalko: @tinyjin (sorry, GitHub didn't show me your comment until now)

Implementing it on Unity's end is either, I patch the WASM implementation before distributing the prebuilt WASM file. Or in Unity itself, I manually reverse all the pixels (this is a bad approach).

BenjaminHalko: @tinyjin Never mind, I'm dumb. I forgot I remade the texture system, and there is now a simple way to just tell it to flip within Unity.

Feel free to close this issue if you don't think this should get merged

hermet: @BenjaminHalko let's handle this in unity side. thanks.

Issue #3821, opened 2025-09-18
Debug build for msvc fails with unresolved external symbol "bool __cdecl jmem_is_heap_pointer


Syndwar:
In conanfile from https://conan.io/center/recipes/thorvg?version=0.15.9 there is a code that states that Debug is not supported for msvc.

if is_msvc(self) and self.settings.build_type == "Debug":
    raise ConanInvalidConfiguration(
    f"{self.ref} doesn't support debug build on MSVC."
)

It is possible to build debug version of the thorvg library but the building fails on linkage with an error:

unresolved external symbol "bool __cdecl jmem_is_heap_pointer(void const *)" (?jmem_is_heap_pointer@@YA_NPEBX@Z) referenced in function "unsigned short __cdecl jmem_compress_pointer(void const *)" (?jmem_compress_pointer@@YAGPEBX@Z)

hermet:
This looks, Conan also needs to distribute thorvg debug package.

hermet:
@Syndwar The fix will be in v1.0-pre31 & v0.15.17. thanks.

PR #3959, opened 2025-11-12
lottie/jerryscript: fix debug build on MSVC

Nor-s:
## Issue: #3821

image

Linking error occurs when building with `extra=lottie_expressions` enabled(build examples):

```
2>libthorvg.a(loaders_lottie_jerryscript_jerry-core_jmem_jmem-allocator.cpp.obj) : error LNK2019: unresolved external symbol "bool __cdecl jmem_is_heap_pointer(void const *)" (?jmem_is_heap_pointer@@YA_NPEBX@Z) referenced in function "unsigned short __cdecl jmem_compress_pointer(void const *)" (?jmem_compress_pointer@@YAGPEBX@Z)
2>libthorvg.a(loaders_lottie_jerryscript_jerry-core_jmem_jmem-heap.cpp.obj) : error LNK2001: unresolved external symbol "bool __cdecl jmem_is_heap_pointer(void const *)" (?jmem_is_heap_pointer@@YA_NPEBX@Z)
2>.\Animation.exe : fatal error LNK1120: 1 unresolved externals
```

## History/ Problem

In commit [#2198](https://github.com/thorvg/thorvg/pull/2198/commits/9d6a12bf248457ceb96da18055df731d9ac53bb7#diff-d0d9eb0da9b37d46d57d81c1e4d130cc53471b8eed3a9251a5a624d248b4e262R26-R28)
, the build defines `JERRY_NDEBUG`, which disables the implementation of `jmem_is_heap_pointer()` because it is wrapped in:

```cpp
// jmem-heap.cpp
#ifndef JERRY_NDEBUG
/**
 * Check whether the pointer points to the heap
 *
 * Note:
 *      the routine should be used only for assertion checks
 *
 * @return true - if pointer points to the heap,
 *         false - otherwise
 */
bool
jmem_is_heap_pointer (const void *pointer) /**< pointer */
{
#if !JERRY_SYSTEM_ALLOCATOR
  return ((uint8_t *) pointer >= JERRY_HEAP_CONTEXT (area)
          && (uint8_t *) pointer <= (JERRY_HEAP_CONTEXT (area) + JMEM_HEAP_AREA_SIZE));
#else /* JERRY_SYSTEM_ALLOCATOR */
  JERRY_UNUSED (pointer);
  return true;
#endif /* !JERRY_SYSTEM_ALLOCATOR */
} /* jmem_is_heap_pointer */
#endif /* !JERRY_NDEBUG */

```

However, this function is still referenced inside `JERRY_ASSERT()` calls, such as:

image

`JERRY_ASSERT` does not completely remove the argument reference, causing the linker to look for the missing symbol and fail.

```cpp
#define JERRY_UNUSED(x) ((void) (x))
```

### DIFF
--- a/src/loaders/lottie/jerryscript/jerry-core/jrt/jrt.h
+++ b/src/loaders/lottie/jerryscript/jerry-core/jrt/jrt.h
@@ -93,10 +93,6 @@ void JERRY_ATTR_NORETURN jerry_unreachable (const char *file, const char *functi
 #define JERRY_ASSERT(x) \
   do                    \
   {                     \
-    if (false)          \
-    {                   \
-      JERRY_UNUSED (x); \
-    }                   \
   } while (0)
 
 #if defined(__GNUC__) || defined(__clang__)


src/loaders/lottie/jerryscript/jerry-core/jrt/jrt.h
```diff
@@ -95,7 +95,6 @@ void JERRY_ATTR_NORETURN jerry_unreachable (const char *file, const char *functi
   {                     \
     if (false)          \
```
hermet: @Nor-s thanks, how about removing whole code? (96 ~ 98)

PR #3965, opened 2025-11-13
examples: Add Windows support

wenjieshen:
- Add Windows batch script (all.bat) to run all examples automatically
- Configure SDL main handling for MSVC builds to prevent conflicts

This enables seamless example execution on Windows platforms

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -8,6 +8,8 @@ if gl_variant == 'OpenGL ES'
     examples_compiler_flags += '-DTHORVG_GL_TARGET_GLES=1'
 endif
 
+examples_compiler_flags += ['-DSDL_MAIN_HANDLED']
+
 examples_dep = [dependency('sdl2')]
 
 if all_engines or get_option('engines').contains('wg')
@@ -109,6 +111,8 @@ if get_option('bindings').contains('capi')
     endforeach
 endif
 
-execute_all_src = join_paths(meson.project_source_root(), 'examples/all.sh')
-execute_all_dst = join_paths(meson.project_build_root(), 'examples/all.sh')
-run_command('cp', execute_all_src, execute_all_dst, check: true)
+if cc.get_id() == 'msvc'
+    configure_file(input: 'all.bat', output: 'all.bat', copy: true)
+else
+    configure_file(input: 'all.sh', output: 'all.sh', copy: true)
+endif
--- a/examples/all.bat
+++ b/examples/all.bat
@@ -1,0 +1,28 @@
+@echo off
+setlocal
+
+rem Work in the folder where this .bat lives
+cd /d "%~dp0"
+
+rem Get arg; default to sw if none
+set "ARG=%~1"
+if not defined ARG set "ARG=sw"
+
+rem Validate
+if /I not "%ARG%"=="gl" if /I not "%ARG%"=="sw" if /I not "%ARG%"=="wg" (
+  echo Invalid argument: %ARG%
+  echo Usage: %~n0 [gl^|sw^|wg]
+  exit /b 1
+)
+
+for %%F in ("*.exe") do (
+  echo Launching %%~nxF %ARG%
+  start "" "%%~fF" %ARG%
+  timeout /t 2 /nobreak >nul
+  rem Try to close it, then force-close if needed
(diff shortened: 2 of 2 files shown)


hermet:
@wenjieshen, Could you please separate the commits into engine(lib) and examples? Thanks.

wenjieshen:
> @wenjieshen, Could you please separate the commits into engine(lib) and examples? Thanks.

Yes, thank you for your review.

examples/Capi.cpp
```diff
@@ -21,6 +21,9 @@
  */
 
 #include <thorvg_capi.h>
+#ifdef _MSC_VER
```
vtorri: maybe it should be `_WIN32`, not sure that it is specific to msvc

examples/meson.build
```diff
@@ -109,6 +109,10 @@ if get_option('bindings').contains('capi')
     endforeach
 endif
 
-execute_all_src = join_paths(meson.project_source_root(), 'examples/all.sh')
-execute_all_dst = join_paths(meson.project_build_root(), 'examples/all.sh')
-run_command('cp', execute_all_src, execute_all_dst, check: true)
+if host_machine.system() == 'windows'
```
vtorri: it should be `get_id()`, see comment

examples/Example.h
```diff
@@ -51,6 +54,12 @@
     #endif
 #endif
 
+#if defined _MSC_VER
```
hermet: This looks not a standard. Not recommended.

examples/Capi.cpp
hermet: It's good to add SDL_MAIN_HANDLED, Can we remove MSC_VER ifdef?

hermet (commented):
@wenjieshen please check comments. Thanks.

examples/meson.build
```diff
@@ -19,6 +19,11 @@ if all_engines or get_option('engines').contains('wg')
     endif
 endif
 
+desktop_oses = ['windows', 'cygwin', 'linux', 'darwin']
+if host_machine.system() in desktop_oses
+    examples_compiler_flags += ['-DSDL_MAIN_HANDLED']
```
hermet: If there is no any issues, Let's put this in `Example.h` in a simpler way.
`#define SDL_MAIN_HANDED`

examples/meson.build
```diff
@@ -109,6 +114,10 @@ if get_option('bindings').contains('capi')
     endforeach
 endif
 
-execute_all_src = join_paths(meson.project_source_root(), 'examples/all.sh')
-execute_all_dst = join_paths(meson.project_build_root(), 'examples/all.sh')
-run_command('cp', execute_all_src, execute_all_dst, check: true)
+if cc.get_id() == 'msvc'
```
hermet, line 114: Can confirm this? Recommend to use a consistent way if it works.

```
if cc.get_id() == 'msvc'
    configure_file(input: 'all.bat', output: 'all.bat', copy: true)
else
    configure_file(input: 'all.sh', output: 'all.sh', copy: true)
endif
```

examples/Example.h
```diff
@@ -51,6 +51,12 @@
...
+// Force to use High Performance GPU on laptops with NVIDIA Optimus or AMD Switchable Graphics
+extern "C" __declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
```
hermet: please remove these hacking style.

examples/Capi.cpp
wenjieshen: Thank you. I removed the define statements and added SDL_MAIN_HANDLED to the compiler flags once the platform is a desktop OS.

examples/meson.build
wenjieshen: Yes, I confirmed that.

examples/meson.build
wenjieshen: Are you suggesting removing this conditional statement in the`meson.build` and adding the following header in `Example.h`?
```cpp
#if defined(WIN32) || defined(__APPLE__) ||  defined(__linux__)
    #define SDL_MAIN_HANDED
#endif
```

hermet: @wenjieshen  yes, and no need a ifdef:
image

PR #3974, opened 2025-11-13
wasm: Migrate binding to thorvg.web

tinyjin:
see: https://github.com/thorvg/thorvg.web/pull/172


hermet:
Let's hold this before web side confirm.

tinyjin:
@hermet Removed and migrated to thorvg.web

Issue #3453, opened 2025-05-09
Add Basic GIF Animation Support in ThorVG


hermet:
ThorVG is open to adding basic GIF animation support to expand its usability for simple animated content. Supporting GIF would enable developers to import and render legacy or lightweight animated assets alongside vector-based graphics and Lottie animations.

The goal is not to fully replicate a GIF decoder pipeline but to provide a minimal and efficient way to parse and render frame-based GIF animations.

### Goals
-  Implement a lightweight GIF parser (or integrate an existing minimal library, e.g., giflib, stb_image, or custom decoder)
-  Extract individual frames, durations, and disposal methods
-  Convert GIF frames into ThorVG internal Surface.
-  Support basic playback control (looping, timing)
-  Integrate with the ThorVG animation timeline and rendering loop

### Considerations:
- Focus on read-only support
- Transparency and frame disposal handling must be respected
- Performance should be optimized for embedded and web targets

PR #3975, opened 2025-11-14
loader: Add support for gif

BenjaminHalko:
Adds support for loading the `.gif` format in `picture`. You can use `animation` to animate it as well.

pet

issue: https://github.com/thorvg/thorvg/issues/3453


hermet:
@BenjaminHalko Thanks, Could you please resolve the copilot comments? (you can close them if they are false-alarm)

BenjaminHalko:
@hermet OK, I think all the issues have been fixed 👍

hermet:
feature target: thorvg v2.0

hermet:
@BenjaminHalko Could you please rebase this? Thanks! 
+ commits must be squashed to 2 commits (feature, test)

BenjaminHalko:
Still fixing the formatting bugs...

BenjaminHalko:
@hermet Should be ready for review now

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,430 @@
...
+{
+    const uint8_t* data;
+    uint32_t size;
+    uint32_t pos;
+};
+
+// Custom InputFunc for reading from memory buffer
+static int memoryInputFunc(GifFileType* gif, GifByteType* buffer, int length)
+{
+    MemoryReader* reader = static_cast<MemoryReader*>(gif->UserData);
```
Nor-s:
```cpp
auto* reader = static_cast<MemoryReader*>
```

Since ThorVG prefers `auto`, could you please update it accordingly?


Image

src/common/tvgCommon.h
```diff
@@ -72,30 +72,32 @@ void operator delete(void* ptr) noexcept;
...
+    Png = 0,
+    Jpg,
+    Webp,
+    Svg,
+    Lot,
+    Sfnt,
+    Gif,
+    Raw,
+    Unknown
+};
```
Nor-s: These changes do not seem related to this feature. Do we need to adjust the position of the GIF enum?

Also, indentation and similar formatting changes seem unnecessary in this PR.

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,430 @@
...
+        int width = endX - startX;
+        if (width > 0) {
+            memset(&canvas32[canvasIdx], 0, width * sizeof(uint32_t));
+        }
+    }
+}
+
+void GifLoader::blitRow(const uint8_t* raster, size_t frameIdx, size_t canvasIdx, int count, const ColorMapObject* colorMap, int transparent)
+{
+    uint32_t* canvas32 = (uint32_t*)canvas;
```
Nor-s: Image

ThorVG prefers static_cast.

Nor-s (commented):
Hello, thank you for your contribution.

It looks like some parts do not align with ThorVG’s coding conventions. 
Could you please update those parts?

Thank you.

see: https://github.com/thorvg/thorvg/wiki/Coding-Convention

src/common/tvgCommon.h
BenjaminHalko: This enum change was required for type auto detection

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+}
+
+bool GifLoader::frame(float no)
+{
+    if (!gifFile || gifFile->ImageCount == 0) return false;
+
+    if (no < 0.0f) no = 0.0f;
+    if (no >= gifFile->ImageCount) no = gifFile->ImageCount - 1.0f;
+
+    uint32_t frameIndex = static_cast<uint32_t>(no);
```
Nor-s:
```cpp
    auto frameIndex = static_cast<uint32_t>(no);
```

src/loaders/external_gif/tvgGifLoader.h
```diff
@@ -0,0 +1,71 @@
...
+ * SOFTWARE.
+ */
+
+#ifndef _TVG_GIF_LOADER_H_
+#define _TVG_GIF_LOADER_H_
+
+#include <gif_lib.h>
+#include "tvgLoader.h"
+
+class GifLoader : public AnimLoader
```
Nor-s: Would it be possible to align this with the existing ThorVG loader style by using struct instead of class?

For example, `Loader` is declared as:

```cpp
struct LottieLoader : AnimLoader, Task
struct JpgLoader : ImageLoader
struct WebpLoader : ImageLoader, Task
struct PngLoader : ImageLoader
```


If we use struct, we can also remove the explicit public access specifier.

```cpp
struct GifLoader : AnimLoader
{
   GifLoader();
```

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+    deinterlace();
+
+    w = static_cast<float>(gifFile->SWidth);
+    h = static_cast<float>(gifFile->SHeight);
+    segmentEnd = static_cast<float>(gifFile->ImageCount);
+
+    calculateFrameRate();
+
+    // Allocate canvas - use size_t to prevent overflow
+    size_t canvasSize = static_cast<size_t>(gifFile->SWidth) * static_cast<size_t>(gifFile->SHeight) * 4;
```
Nor-s:
```cpp
auto canvasSize = static_cast<size_t>(w) * static_cast<size_t>(h) * sizeof(uint32_t);
``` 

size_t -> `auto` 
4 -> `sizeof(uint32_t)`

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+    }
+    if (totalDelay > 0) {
+        frameRate = (gifFile->ImageCount * 100.0f) / totalDelay;
+    } else {
+        frameRate = 10.0f;  // Default to 10 fps
+    }
+}
+
+void GifLoader::deinterlace()
+{
```
Nor-s: Could you double-check `deinterlace()` ? 
It seems that `DGifSlurp()` behavior differs by giflib version.

In my local test with giflib 6.x, 
applying this extra `deinterlace()` made the GIF render differently from Chrome. 

From the giflib history, newer versions appear to already place interlaced rows into their final positions during DGifSlurp().

For reference:
- [6eed8da: Experimental support for interlaced image reads in DGifSlurp()](https://sourceforge.net/p/giflib/code/ci/6eed8da894b10c43bbec51de87a31be977737c70/)
- [e740592: DGifSlurp() and EGifSpew() now read and write interlaced images
](https://sourceforge.net/p/giflib/code/ci/e74059267837c2831b2cbee9aaf488debdd3e06b/)

If this loader depends on that behavior, would it make sense to require an appropriate giflib version in `Meson` or `#ifdef`


### Test 

```cpp
void GifLoader::deinterlace()
{
    return; // off deinterlace
```


| off deinterlace | on deinterlace | chrome |
|-|-|-|
| Image|Image|Image|

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+}
+
+struct MemoryReader
+{
+    const uint8_t* data;
+    uint32_t size;
+    uint32_t pos;
+};
+
+static int _memoryInputFunc(GifFileType* gif, GifByteType* buffer, int length)
```
Nor-s: Could we move `MemoryReader` and `_memoryInputFunc()` to the Internal Class Implementation section?  And, add comment

```
```

or 

```cpp
static int _memoryInputFunc(GifFileType* gif, GifByteType* buffer, int length){
  struct MemoryReader { ... };

  auto memoryInput = [](GifFileType* gif, GifByteType* buffer, int length) -> int {
      ...
  };

  gifFile = DGifOpen(reader, memoryInput, &error);


}
```

src/loaders/external_gif/tvgGifLoader.h
```diff
@@ -0,0 +1,71 @@
...
+    void blitRow(const uint8_t* raster, size_t frameIdx, size_t canvasIdx, int count, const ColorMapObject* colorMap, int transparent);
+    void blitFrame(const SavedImage* frame, const GifImageDesc* imageDesc, const GraphicsControlBlock& gcb);
+    bool needsReset(uint32_t frameIndex);
+    void compositeFromStart(uint32_t frameIndex);
+    void calculateFrameRate();
+    void deinterlace();
+
+    GifFileType* gifFile = nullptr;
+    uint8_t* fileData = nullptr;
+    uint32_t fileSize = 0;
```
Nor-s: fileData -> conetent or data
fileSize -> size

Could we rename fileData/fileSize to data/size to align with the other loaders? 

```cpp
// LottieLoader
    const char* content = nullptr;      //lottie file data
    uint32_t size = 0;                  //lottie data size

// JpgLoader (external)
    unsigned char* data = nullptr;
    uint32_t size = 0;

// webpLoader(external)
    unsigned char* data = nullptr;
    uint32_t size = 0;
```

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+bool GifLoader::read()
+{
+    if (!Loader::read()) return true;
+    if (!gifFile || gifFile->ImageCount == 0) return false;
+
+    surface.cs = ImageLoader::cs;
+
+    // Decode and composite frame 0 - use size_t to prevent overflow
+    memset(canvas, 0, static_cast<size_t>(gifFile->SWidth) * static_cast<size_t>(gifFile->SHeight) * 4);
```
Nor-s: static_cast<size_t>(w) * static_cast<size_t>(h) * sizeof(uint32_t)

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+        if (!fileData) return false;
+        memcpy((unsigned char*)fileData, data, size);
+        freeData = true;
+    } else {
+        fileData = (unsigned char*)data;
+        freeData = false;
+    }
+    fileSize = size;
+
+    MemoryReader* reader = tvg::malloc<MemoryReader>(sizeof(MemoryReader));
```
Nor-s: MemoryReader* reader = tvg::malloc<MemoryReader>(sizeof(MemoryReader));

-> `MemoryReader reader;`   or  `MemoryReader reader = {fileData, fileSize, 0};`


Could MemoryReader be a local value instead of a heap allocation? 

It is small and only needs to live while `DGifSlurp()` synchronously reads 
the in-memory data inside `open(data, ...)`. 

We can pass &reader to `DGifOpen()`, then clear `gifFile->UserData` after `DGifSlurp()` returns.

  This would remove the extra dynamic allocation and the three `tvg::free(reader)` cleanup paths.

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+    deinterlace();
+
+    w = static_cast<float>(gifFile->SWidth);
+    h = static_cast<float>(gifFile->SHeight);
+    segmentEnd = static_cast<float>(gifFile->ImageCount);
+
+    calculateFrameRate();
+
+    // Allocate canvas - use size_t to prevent overflow
+    size_t canvasSize = static_cast<size_t>(gifFile->SWidth) * static_cast<size_t>(gifFile->SHeight) * 4;
```
Nor-s: `size_t canvasSize` ->     `auto canvasSize = static_cast<size_t>(w) * static_cast<size_t>(h) * sizeof(uint32_t)`

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+    if (!gifFile || gifFile->ImageCount == 0) return;
+
+    uint32_t totalDelay = 0;
+    for (int i = 0; i < gifFile->ImageCount; i++) {
+        GraphicsControlBlock gcb;
+        DGifSavedExtensionToGCB(gifFile, i, &gcb);
+        totalDelay += gcb.DelayTime;
+    }
+    if (totalDelay > 0) {
+        frameRate = (gifFile->ImageCount * 100.0f) / totalDelay;
```
Nor-s: Could we clarify the 100.0f factor here (Comment)? 

DelayTime is in 0.01-second units.

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+    } else {
+        frameRate = 10.0f;  // Default to 10 fps
```
Nor-s: Could we add a TVGERR() here when `totalDelay` is 0 and we fall back to 10 fps? 

Lottie logs when the framerate is invalid, and doing the same here would make it clearer that the GIF did not provide usable frame delay information.

```cpp
  src/loaders/lottie/tvgLottieLoader.cpp:198

  if (frameRate < FLOAT_EPSILON) {
      TVGLOG("LOTTIE", "Not a Lottie file? Frame rate is 0!");
      return false;
  }
```

also see: https://github.com/thorvg/thorvg/pull/4512#issuecomment-4815500996

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+    return true;
+}
+
+bool GifLoader::read()
+{
+    if (!Loader::read()) return true;
+    if (!gifFile || gifFile->ImageCount == 0) return false;
+
+    surface.cs = ImageLoader::cs;
```
Nor-s:
```cpp
    surface.cs = ImageLoader::cs;
//...
    surface.cs = abgr ? ColorSpace::ABGR8888S : ColorSpace::ARGB8888S;
```

Do we need this assignment? 

surface.cs is overwritten below with the actual GIF output color space, 
so this line seems redundant.

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+    // Decode and composite frame 0 - use size_t to prevent overflow
+    memset(canvas, 0, static_cast<size_t>(gifFile->SWidth) * static_cast<size_t>(gifFile->SHeight) * 4);
+    compositeFrame(0);
+    lastCompositedFrame = 0;
+
+    surface.buf8 = canvas;
+    surface.stride = gifFile->SWidth;
+    surface.w = gifFile->SWidth;
+    surface.h = gifFile->SHeight;
```
Nor-s: Since open() already stores the GIF canvas size in w/h, would it be better to use those values when setting up the surface? 
That would keep read() less dependent on gifFile internals.


```cpp
  surface.buf8 = canvas;
  surface.stride = static_cast<uint32_t>(w);
  surface.w = static_cast<uint32_t>(w);
  surface.h = static_cast<uint32_t>(h);
```

see jpgLoader

```cpp
void JpgLoader::run(unsigned tid)
{
    surface.cs = ColorSpace::ABGR8888;
    surface.buf8 = jpgdDecompress(decoder);
    surface.stride = static_cast<uint32_t>(w);
    surface.w = static_cast<uint32_t>(w);
    surface.h = static_cast<uint32_t>(h);
    surface.channelSize = sizeof(uint32_t);
    surface.premultiplied = true;
    surface.alphaIgnored = true;

    clear();
}
```

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+    int startX = (prevDesc->Left < 0) ? 0 : prevDesc->Left;
+    int endX = prevDesc->Left + prevDesc->Width;
+    if (endX > gifFile->SWidth) endX = gifFile->SWidth;
+
+    for (int y = startY; y < endY; y++) {
+        size_t canvasIdx = static_cast<size_t>(y) * gifFile->SWidth + startX;
+        int width = endX - startX;
+        if (width > 0) {
+            memset(&canvas32[canvasIdx], 0, width * sizeof(uint32_t));
+        }
```
Nor-s: prefer one-line form for simple conditions

```cpp
 if (width > 0) memset(&canvas32[canvasIdx], 0, width * sizeof(uint32_t));
```

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+    surface.h = gifFile->SHeight;
+    surface.cs = abgr ? ColorSpace::ABGR8888S : ColorSpace::ARGB8888S;
+    surface.channelSize = sizeof(uint32_t);
+
+    return true;
+}
+
+RenderSurface* GifLoader::bitmap()
+{
+    return &surface;
```
Nor-s: Do we need to override `bitmap()` here? 
AnimLoader already inherits `ImageLoader::bitmap()`, which returns

`&surface` only when surface.data is valid. 

Since this external GIF loader does not use Task, there is no `done()` call needed like the static GIF loaders.

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+Result GifLoader::segment(float begin, float end)
+{
+    if (!gifFile || gifFile->ImageCount == 0) return Result::InsufficientCondition;
+
+    if (begin < 0.0f) begin = 0.0f;
+    if (end > gifFile->ImageCount) end = static_cast<float>(gifFile->ImageCount);
+    if (begin > end) return Result::InvalidArguments;
+
+    segmentBegin = begin;
+    segmentEnd = end;
```
Nor-s: `segment()` currently only stores `segmentBegin/segmentEnd`, 

but the GIF loader does not seem to use those values when reporting totalFrame()/duration() or when selecting frames. Is this intentional for GIF, or should it follow the Lottie segment behavior?

see LottieLoader

```cpp
float LottieLoader::totalFrame()
{
    return segmentEnd - segmentBegin;
}

float LottieLoader::duration()
{
    return (segmentEnd - segmentBegin) / frameRate;
}

float LottieLoader::curFrame()
{
    return frameNo - startFrame();
}

bool LottieLoader::frame(float no)
```

Nor-s: If GIF segment playback is not intended to be supported, 
would it be clearer to return `Result::NonSupport `or log it explicitly instead of storing a range 
that has no effect?

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+        GraphicsControlBlock gcb;
+        DGifSavedExtensionToGCB(gifFile, lastCompositedFrame, &gcb);
+        if (gcb.DisposalMode == DISPOSE_PREVIOUS) return true;
+    }
+    return false;
+}
+
+void GifLoader::compositeFromStart(uint32_t frameIndex)
+{
+    memset(canvas, 0, static_cast<size_t>(gifFile->SWidth) * static_cast<size_t>(gifFile->SHeight) * 4);
```
Nor-s: `static_cast<size_t>(w) * static_cast<size_t>(h) * sizeof(uint32_t)`

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+    for (uint32_t i = 0; i <= frameIndex; i++) {
+        bool draw = true;
+        if (i < frameIndex) {
+            GraphicsControlBlock curGcb;
+            DGifSavedExtensionToGCB(gifFile, i, &curGcb);
+            if (curGcb.DisposalMode == DISPOSE_PREVIOUS) draw = false;
+        }
+        compositeFrame(i, draw);
+    }
+}
```
Nor-s: Could we split the final frame out of this loop? 

```cpp
      for (uint32_t i = 0; i < frameIndex; i++) {
          GraphicsControlBlock gcb;
          DGifSavedExtensionToGCB(gifFile, i, &gcb);
          compositeFrame(i, gcb.DisposalMode != DISPOSE_PREVIOUS);
      }

      compositeFrame(frameIndex);
```

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+        freeData = true;
+    } else {
+        fileData = (unsigned char*)data;
+        freeData = false;
+    }
+    fileSize = size;
+
+    MemoryReader* reader = tvg::malloc<MemoryReader>(sizeof(MemoryReader));
+    if (!reader) {
+        if (freeData) tvg::free(fileData);
```
Nor-s: This failure path seems to have the same ownership issue. If freeData is true, 

FileData is freed but fileData/freeData are not reset, so a later clear() could free the same pointer again. 

could we use clear() here or reset fileData/freeData after freeing it?

Nor-s:
```cpp
      if (freeData) tvg::free(fileData);
      fileData = nullptr;
```

or

```cpp
      clear();
      return false;
```

src/loaders/external_gif/tvgGifLoader.cpp
```diff
@@ -0,0 +1,422 @@
...
+    reader->data = fileData;
+    reader->size = fileSize;
+    reader->pos = 0;
+
+    int error = 0;
+    gifFile = DGifOpen(reader, _memoryInputFunc, &error);
+    if (!gifFile) {
+        TVGERR("GIF", "Failed to open GIF from memory: error code %d", error);
+        tvg::free(reader);
+        if (freeData) tvg::free(fileData);
```
Nor-s:
```cpp
      if (freeData) tvg::free(fileData);
      fileData = nullptr;
```

or

```cpp
      clear();
      return false;
```

Nor-s (commented):
@BenjaminHalko 

Hi, sorry for the delayed review. I have reviewed the external_gif part first.

Please feel free to ignore any comments that do not make sense or do not fit the direction of this PR.

I will review the static GIF loader within the next week.

Nor-s:
@BenjaminHalko 

Hi, would you be able to split the current PR into separate PRs for the static and the external?

Ideally, the current PR would only include the changes related to external GIFs,
and a new PR would contain only the static-related changes.

A stacked PR approach would also work if that makes things easier.

Also, it looks like there may be an issue with the current GIF unit tests. (It may be related to the segment changes.)
Would you be able to take a look at that as well?

```
-------------------------------------------------------------------------------
GIF Frames Counting
-------------------------------------------------------------------------------
../test/testAnimation.cpp:193
...............................................................................

../test/testAnimation.cpp:205: FAILED:
  REQUIRE( animation->totalFrame() == Approx(4).margin(0.001f) )
with expansion:
  121.0f == Approx( 4.0 )

Catch will terminate because it needed to throw an exception.
The message was: Test failure requires aborting test!
terminate called without an active exception
../test/testAnimation.cpp:205: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

===============================================================================
test cases:   8 |   7 passed | 1 failed
assertions: 102 | 100 passed | 2 failed


```



Thank you!

Nor-s:
Also, there are two more things I’d like to ask:

* Would you be able to add the issue number to the commit messages? and add a bit more detail (ex. It would be helpful to mention which versions of the GIF library are compatible.)
* I think the review feedback from the external changes could also be applied to the static changes. (ex `struct GifLoader`)

Sorry for the many requests, and thank you again.

BenjaminHalko:
@Nor-s Hello I've created a new PR separating the static and external deps. However GitHub seems to have forced the PR to be on my fork instead of the main repo. Did you want to have 2 PRs targeting main instead?
https://github.com/BenjaminHalko/thorvg/pull/1

hermet:
@BenjaminHalko Hi, CI test failed. I'd rather suggest PR order: 1. static, 2. external. Thanks.

BenjaminHalko:
@hermet Ready for review again. I will create a PR for the external loader once this is merged.

PR #3978, opened 2025-11-16
svg: Fixes MSVC compatibility issues

wenjieshen:
Addresses issues with the MSVC compiler by adding conditional compilation blocks to handle compiler-specific behavior.

Ensures minimal changes are applied specifically for the MSVC compiler.

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -3760,7 +3760,7 @@ static bool _svgLoaderParserForValidCheckXmlOpen(SvgLoaderData* loader, const ch
 
     if (attrs) {
         sz = attrs - content;
-        while ((sz > 0) && (isspace(content[sz - 1]))) sz--;
+        while ((sz > 0) && (isspace((unsigned char)content[sz - 1]))) sz--;
         if ((unsigned)sz >= sizeof(tagName)) return false;
         strncpy(tagName, content, sz);
         tagName[sz] = '\0';


src/loaders/lottie/jerryscript/jerry-core/jrt/jrt.h
```diff
@@ -23,8 +23,13 @@
 #include "jerryscript-port.h"
 #include "jrt-types.h"
 
-
-#define JERRY_NDEBUG
+#ifndef _MSC_VER
+  #define JERRY_NDEBUG
+#else
+  #ifdef NDEBUG
```
vtorri: instead of doing this, use the option `b_ndebug` of the meson project, see https://mesonbuild.com/Builtin-options.html#base-options

hermet:
@wenjieshen Thanks. 
1. It seems having two different fixes. It's better to separate to 2 commits 1. Lottie Fixes / 2. SVG Fixes
2. Could you please share error messages on your machine? I couldn't see any issues on them.

wenjieshen:
> @wenjieshen Thanks.
> 
> 1. It seems having two different fixes. It's better to separate to 2 commits 1. Lottie Fixes / 2. SVG Fixes
> 2. Could you please share error messages on your machine? I couldn't see any issues on them.

I got an assertion failure when I executed the example Lottie.
image

The JRT file patch could be ignored because it has been fixed since the commit cd74d8d657c1f71d1be6f547ea64510e65353c86.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3760,7 +3760,11 @@ static bool _svgLoaderParserForValidCheckXmlOpen(SvgLoaderData* loader, const ch
 
     if (attrs) {
         sz = attrs - content;
+#ifdef _MSC_VER
```
hermet: I guess we can rather retain only this code without if-def. 
`while ((sz > 0) && (isspace((unsigned char)content[sz - 1]))) sz--;`

hermet (commented):
@wenjieshen thanks, please check a comment.

src/loaders/svg/tvgSvgLoader.cpp
hermet: or maybe we can change the `sz` type to `unsigned char` from the beginning?

wenjieshen:
> @wenjieshen thanks, please check a comment.

Thank you.
I have tested this version on macOS and found it to work well.

hermet:
btw, Lottie examples should not to access svg loader... this is bad case. Could you please highlight `which sample` has that issue?

PR #3979, opened 2025-11-16
gl: Apply trim path during preparation instead of tessellation

wenjieshen:
Moves trim path operations from the tessellation phase to the preparation phase, where the trimmed path replaces the optimized path. This ensures the trimmed geometry is available throughout the rendering pipeline and prevents geometry loss during trim path operations.

The trim path is now applied once in GlGeometry::prepare() and reused by both fill and stroke tessellation, eliminating duplicate trim operations.

Related issues:
https://github.com/thorvg/thorvg/issues/3653


hermet (approved):
fps: +3%

hermet:
@wenjieshen just observed a regression in Lottie example, could you please double check this?

image

Issue #3896, opened 2025-10-17
lottie: inconsistent rendering


theashraf:
[circle-blob-icon-lottie-json-animation.json](https://github.com/user-attachments/files/22962929/circle-blob-icon-lottie-json-animation.json)

<img width="400" height="690" alt="Image" src="https://github.com/user-attachments/assets/7c8b2669-cf0b-4d55-b693-1af0ddd1e61f" />

theashraf:
seems related to expressions support in the animated TextDocument property

```json
"d": {
          "k": [
            {
              "s": {
                "sz": [516, 202],
                "ps": [-281, -84],
                "s": 205,
                "f": "Pona-Bold",
                "t": "77",
                "ca": 0,
                "j": 2,
                "tr": 0,
                "lh": 71,
                "ls": 0,
                "fc": [0.424, 0.675, 0.894]
              },
              "t": 0
            }
          ],
[...]

PR #3980, opened 2025-11-16
lottie/expressions: add `.value` accessor for numeric effect properties

Nor-s:
This patch fixes an issue where expressions such as:

```js
effect("Slider Control")("Slider").value.toFixed(0)
```

failed because numeric effect properties (Integer, Float, Opacity) did not
expose a `.value` accessor.  
These properties are now *boxed* into Number objects with a `.value` field.

- issue: #3896

#### Prior related work

- https://github.com/thorvg/thorvg/pull/3749

---

<details>
<summary> Effect → Property → Value lookup in After Effects </summary>

- AE treats the slider effect’s internal slider as a `Property` object.  
The `.value` accessor is required to retrieve the underlying primitive.
    - **AE documentation:** [effect](https://ae-expressions.docsforadobe.dev/objects/effect/#returns), [property-value](https://ae-expressions.docsforadobe.dev/objects/property/#value)

| Effect | Property| Value
| - | - | - |
|Image|Image| Image|

</details>

<details>
<summary>boxing: Number → Number Object </summary>

| [legacy-and-extend-script-engine](https://helpx.adobe.com/after-effects/using/legacy-and-extend-script-engine.html)|
|-|
|Image|

- AE’s Legacy engine requires .value, while the modern JavaScript engine automatically resolves the underlying numeric value. To support both behaviors, numeric effect properties must be boxed into a Number object and expose .value. This is done using `jerry_value_to_object()`.

```cpp
/**
 * Call ToObject operation on the api value.
 *
 * Note:
 *      returned value must be freed with jerry_value_free, when it is no longer needed.
 *
 * @return converted object value - if success
 *         thrown error - otherwise
 */
jerry_value_t
jerry_value_to_object (const jerry_value_t value) /**< input value */
{
  return jerry_return (ecma_op_to_object (value));
} /* jerry_value_to_object */

ecma_value_t
ecma_op_to_object (ecma_value_t value) /**< ecma value */
{
  ecma_check_value_type_is_spec_defined (value);
  ecma_builtin_id_t proto_id = ECMA_BUILTIN_ID_OBJECT_PROTOTYPE;
  uint8_t class_type;

  if (ecma_is_value_number (value))
  {
#if JERRY_BUILTIN_NUMBER
    proto_id = ECMA_BUILTIN_ID_NUMBER_PROTOTYPE;
#endif /* JERRY_BUILTIN_NUMBER */
    class_type = ECMA_OBJECT_CLASS_NUMBER;
  }
```

</details>

### Result

|main | fetch |
|-|-|
|issue|Fetch|

|**Note: Position and font-related issues are not yet resolved.**|
|-|
|image|

### DIFF
--- a/src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp
+++ b/src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp
@@ -267,6 +267,21 @@ jerry_value_to_object (const jerry_value_t value) /**< input value */
 } /* jerry_value_to_object */
 
 /**
+ * Call ToPrimitive operation on the api value.
+ *
+ * Note:
+ *      returned value must be freed with jerry_value_free, when it is no longer needed.
+ *
+ * @return converted primitive value - if success
+ *         thrown error - otherwise
+ */
+jerry_value_t
+jerry_value_to_primitive (const jerry_value_t value) /**< input value */
+{
+  return jerry_return (ecma_op_to_primitive (value, ECMA_PREFERRED_TYPE_NO));
+} /* jerry_value_to_primitive */
+
+/**
  * Call the ToString ecma builtin operation on the api value.
  *
  * Note:
--- a/src/loaders/lottie/jerryscript/jerry-core/include/jerryscript-core.h
+++ b/src/loaders/lottie/jerryscript/jerry-core/include/jerryscript-core.h
@@ -32,6 +32,7 @@ bool jerry_value_is_object (const jerry_value_t value);
 bool jerry_value_is_string (const jerry_value_t value);
 bool jerry_value_is_exception (const jerry_value_t value);
 jerry_value_t jerry_value_to_object (const jerry_value_t value);
+jerry_value_t jerry_value_to_primitive (const jerry_value_t value);
 jerry_value_t jerry_value_to_string (const jerry_value_t value);
 float jerry_value_as_number (const jerry_value_t value);
 int32_t jerry_value_as_int32 (const jerry_value_t value);
--- a/src/loaders/lottie/tvgLottieExpressions.cpp
+++ b/src/loaders/lottie/tvgLottieExpressions.cpp
@@ -64,6 +64,52 @@ static const char* EXP_EFFECT= "effect";
 static LottieExpressions* exps = nullptr;   //singleton instance engine
 
 
+static bool _isNumber(const jerry_value_t value)
+{
+    if (jerry_value_is_object(value)) {
+        auto primitive = jerry_value_to_primitive(value);
+        bool result = jerry_value_is_number(primitive);
+        jerry_value_free(primitive);
+        return result;
+    }
+    return jerry_value_is_number(value);
+}
+
+
+template<typename T, T (*Converter)(jerry_value_t)>
+inline T _asNumericType(const jerry_value_t value)
+{
+    if (jerry_value_is_object(value)) {
+        auto primitive = jerry_value_to_primitive(value);
+        if (jerry_value_is_number(primitive)) {
+            auto result = Converter(primitive);
+            jerry_value_free(primitive);
+            return result;
+        }
+        jerry_value_free(primitive);
+    }
+    return Converter(value);
+}
+
+
+static inline float _asNumber(const jerry_value_t value)
+{
+    return _asNumericType<float, jerry_value_as_number>(value);
+}
+
+
+static inline int32_t _asInt32(const jerry_value_t value)
+{
+    return _asNumericType<int32_t, jerry_value_as_int32>(value);
+}
+
+
+static inline uint32_t _asUint32(const jerry_value_t value)
+{
+    return _asNumericType<uint32_t, jerry_value_as_uint32>(value);
+}
+
+
 static ExpContent* _expcontent(LottieExpression* exp, float frameNo, void* data, size_t refCnt = 1)
 {
     auto ret = tvg::malloc<ExpContent>(sizeof(ExpContent));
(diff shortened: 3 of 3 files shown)


Nor-s:
> Thanks. The effect() value itself can be a numeric value; I think it can't be cast to an object always.
> 
> LottieExtensions, broken after: ...
> 
> Need a solution to support value as a primitives and object for "value" property...


sorry for not checking this more carefully.
The issue happens because the following JerryScript APIs don’t handle **Number object type**, only primitive numbers:

- `bool jerry_value_is_number (const jerry_value_t value);`
- `float jerry_value_as_number (const jerry_value_t value);`
- `int32_t jerry_value_as_int32 (const jerry_value_t value);`
- `uint32_t jerry_value_as_uint32 (const jerry_value_t value);`

I’m going to add wrapper helper functions around these APIs so they can correctly handle both primitive numbers and Number objects, and update the code to use those helpers instead.

Thank you.

Nor-s:
@hermet 

I've fixed the issue where Number objects (e.g., `new Number()`) were not handled correctly due to the underlying number APIs only supporting primitive values.


Added:
- Box Integer/Float/Opacity values as Number objects with `.value` field (in `_effectProperty`)
- Wrapper helpers: `_isNumber`, `_asNumericType`, `_asNumber`, `_asUint32`, `_asInt32`
- JerryScript APIs: `jerry_object_type`, `jerry_value_to_primitive`

Changed:
- `$bm_rt` unwrap Number object results in `LottieExpressions::evaluate()`
- all direct JerryScript number checks/conversions replaced with `_isNumber()`, `_asNumber()`, `_asUint32()`, and `_asInt32()`.


The following expression now works as intended:
```json
"x": "var $bm_rt; var slider = effect('Slider Control')('Slider'); var rotation = clamp(slider.value * 100, new Number(0), new Number(360)); $bm_rt = new Number(rotation);"
```


|expression example | fetch | lottie docs|
|-|-|-|
|issue (2)|issue (1)| issue |

[lottie (3).json](https://github.com/user-attachments/files/23615310/lottie.3.json)

Nor-s:
Note:
- `jerry_value_to_primitive` return undefined on failure (instead of throwing)
- Because the `jerry_throw_sz` API is not included on our side.

```cpp
/**
 * Call ToPrimitive operation on the api value.
 *
 * Note:
 *      returned value must be freed with jerry_value_free, when it is no longer needed.
 *
 * @return converted primitive value - if success
 *         thrown error - otherwise <--- origin
 *         undefined - otherwise <--- changed
 */
jerry_value_t
jerry_value_to_primitive (const jerry_value_t value) /**< input value */
{
  jerry_assert_api_enabled ();

  if (ecma_is_value_exception (value))
  {
       // return jerry_throw_sz (JERRY_ERROR_TYPE, ecma_get_error_msg (ECMA_ERR_VALUE_MSG)); <--- origin
       return jerry_undefined(); // <--- changed
  }

  return jerry_return (ecma_op_to_primitive (value, ECMA_PREFERRED_TYPE_NO));
} /* jerry_value_to_primitive */
```

hermet (commented):
Minimal changes are highly preferred.
This might just need a `jerry_value_is_object()` call for this update...

Nor-s:
> Minimal changes are highly preferred.
> 
> This might just need a `jerry_value_is_object()` call for this update...

jerry_value_is_object() only tells us whether the value is an object.

it does not provide a way to extract the underlying primitive from a Number object. 

As far as I know, way to retrieve the primitive value would be calling the JS valueOf() method(maybe overhead), 

but I’ll investigate whether JerryScript exposes a cleaner internal approach.

If there’s no way to unwrap a pure Number object (like 'new Number()'), then I may simply only handle the number object with '.value' this update

hermet:
> > Minimal changes are highly preferred.
> > This might just need a `jerry_value_is_object()` call for this update...
> 
> jerry_value_is_object() only tells us whether the value is an object.
> 
> it does not provide a way to extract the underlying primitive from a Number object.
> 
> As far as I know, way to retrieve the primitive value would be calling the JS valueOf() method(maybe overhead),
> 
> but I’ll investigate whether JerryScript exposes a cleaner internal approach.
> 
> If there’s no way to unwrap a pure Number object (like 'new Number()'), then I may simply only handle the number object with '.value' this update

@Nor-s Functions expect given jerry value with a certain value type. You can query the value is object or number.

- jerry_value_is_object()  -> handle as the object.
- jerry_value_is_number() -> handle as the number.

Please suggest a specific fix only for this issue scenario first, and then we can work on determining the best design.

If not certain, you can leave it to me. Thanks.

Nor-s:
> > > Minimal changes are highly preferred.
> > > This might just need a `jerry_value_is_object()` call for this update...
> > 
> > 
> > jerry_value_is_object() only tells us whether the value is an object.
> > it does not provide a way to extract the underlying primitive from a Number object.
> > As far as I know, way to retrieve the primitive value would be calling the JS valueOf() method(maybe overhead),
> > but I’ll investigate whether JerryScript exposes a cleaner internal approach.
> > If there’s no way to unwrap a pure Number object (like 'new Number()'), then I may simply only handle the number object with '.value' this update
> 
> @Nor-s Functions expect given jerry value with a certain value type. You can query the value is object or number.
> 
> * jerry_value_is_object()  -> handle as the object.
> * jerry_value_is_number() -> handle as the number.
> 
> Please suggest a specific fix only for this issue scenario first, and then we can work on determining the best design.
> 
> If not certain, you can leave it to me. Thanks.

Sorry for the misunderstanding. I’ve removed the object-type check and only added the `jerry_value_to_primitive` API. The implementation follows the same pattern as the existing `jerry_value_to_string` and `jerry_value_to_object` helpers.

Thank you for the review.

Nor-s:
I haven’t completed check Example yet, so I’m switching the PR back to draft for now.

Nor-s:
fix_issue

I’ve completed the remaining checks, so I’m reopening this PR from draft.

hermet:
closed by https://github.com/thorvg/thorvg/pull/4165

PR #3982, opened 2025-11-17
MSVC portability++

hermet:
move the windows compat to common side


hermet:
@wenjieshen please verify this if this works for you. Thanks.

Issue #3954, opened 2025-11-11
wg_engine: broken radial gradient rendering


hermet:
current:

<img width="990" height="989" alt="Image" src="https://github.com/user-attachments/assets/91fadf1b-5a87-435a-ba44-f7af09607330" />

to-be:

<img width="990" height="992" alt="Image" src="https://github.com/user-attachments/assets/508bed2e-2a64-4a37-bdef-cba3424e215d" />

```
<?xml version="1.0" encoding="utf-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="100px" height="100px" viewBox="0 0 100 100" xml:space="preserve">
<defs>
<radialGradient id="Gradient1" fx="0.0" fy="0.0" cx="1" cy="1" r="1">
<stop offset="0.0" stop-color="rgb(255,255,255)"/>
<stop offset="1.0" stop-color="rgb(0,0,0)"/>
</radialGradient>
</defs>
<rect id="Rect" fill="url(#Gradient1)" x="0" y="0" width="100.0" height="100.0"/>
</svg>
```

PR #3988, opened 2025-11-19
wg_engine: fix broken radial gradient rendering

SergeyLebedkin:
fixed gradient coefficient in case the general gradient equation has no solutions
works with all spread rules
image
https://github.com/thorvg/thorvg/issues/3954

### DIFF
--- a/src/renderer/wg_engine/tvgWgShaderSrc.cpp
+++ b/src/renderer/wg_engine/tvgWgShaderSrc.cpp
@@ -172,7 +172,12 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4f {
     let a = 1.0*dot(d1, d1) - 1.0*rd*rd;
     let b = 2.0*dot(d0, d1) - 2.0*r0*rd;
     let c = 1.0*dot(d0, d0) - 1.0*r0*r0;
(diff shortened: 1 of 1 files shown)


src/renderer/wg_engine/tvgWgShaderSrc.cpp
```diff
@@ -172,7 +172,10 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4f {
     let a = 1.0*dot(d1, d1) - 1.0*rd*rd;
     let b = 2.0*dot(d0, d1) - 2.0*r0*rd;
     let c = 1.0*dot(d0, d0) - 1.0*r0*r0;
-    let t = (-b + sqrt(b*b - 4*a*c))/(2*a);
+    let d = b*b - 4*a*c;
+    var t = 0.0;
+    if (d >= 0) { t = (-b + sqrt(d))/(2*a); }
```
hermet: Please consider using the fewest instructions for pixel-level operations as possible as you can:
```
    let d = b*b - 4*a*c 
    var t = 0.0;
    if (d >= 0) { 
        t = (-b + sqrt(d))/(2*a);
        if (t >= 1) t = 0.0;
    }
```

hermet (approved):

Issue #3999, opened 2025-11-23
gl_engine: regression bug


hermet:
trimpath is flickering during animation

curlprit:
```
commit 02ed0781db071c4942082ccd07eb37d18ea5c351
Author: wenjieshen <jay@lottiefiles.com>
Date:   Mon Nov 3 00:25:27 2025 +0800

    gl: handle degenerate paths as strokes

    Add support for rendering degenerate paths (2-point paths) as strokes in GL engine:
    - Add RenderPath::optimize() to simplify paths by merging close points and flattening curves
    - Add math utilities: shouldMergePts(), checkLine(), validateCubic(), checkLinePts()
    - Add tesselateLineOnly() to handle 2-point paths with MIN_GL_STROKE_ALPHA opacity
    - Update GlRenderer::prepare() to detect and handle degenerate paths
    - Add internal unit tests for path optimization

    Related Issue: #2920 
```

[sample.json](https://github.com/user-attachments/files/23693794/sample.json)

https://github.com/user-attachments/assets/96933b8f-257f-40d6-9930-cee31b49e8bf

wenjieshen:
Thank you for your patience. I have implemented a patch and conducted tests on all examples, including some animations with a trim path. While it may not be pixel-perfect, I did not detect any regressions in all tests for this version.

https://github.com/user-attachments/assets/e9040df3-5ef4-460a-9326-7d0d9ca44026

https://github.com/user-attachments/assets/8c70daaa-672a-46ea-ad77-2ba269173f0b

hermet:
@wenjieshen one another regression :( 

GL (wrong)

https://github.com/user-attachments/assets/ebf8653a-d7b9-4f1c-bb73-c95b626a7b70

SW (normal)

https://github.com/user-attachments/assets/4569a9ae-e310-4355-bf1e-4649a35b30d6

[offsetRectNoRounded.json](https://github.com/user-attachments/files/23785738/offsetRectNoRounded.json)

wenjieshen:
> [@wenjieshen](https://github.com/wenjieshen) one another regression :(
> 
> GL (wrong)
> 
>  gl.mp4 
> SW (normal)
> 
>  sw.mp4 
> [offsetRectNoRounded.json](https://github.com/user-attachments/files/23785738/offsetRectNoRounded.json)

Thanks for the patience, I have pushed the fixing patch for it.

PR #4000, opened 2025-11-23
gl: fix trimpath flickering regression

wenjieshen:
Fix trimpath flickering during animation by reordering path optimization.

The regression was caused by applying path optimization before trimming, which resulted in incorrect geometry for trimmed paths. This commit fixes the issue by:
- Moving optimizePath() call to after trimming operation
- Applying optimization only to the trimmed path when trimpath is active
- Ensuring optimizedPath is properly cleared when trimming fails

Fixes: #3999

### DIFF
--- a/src/renderer/gl_engine/tvgGlCommon.h
+++ b/src/renderer/gl_engine/tvgGlCommon.h
@@ -123,7 +123,6 @@ struct GlGeometry
     RenderRegion bounds = {};
     FillRule fillRule = FillRule::NonZero;
     RenderPath optimizedPath;
-    bool optimized = false;
 };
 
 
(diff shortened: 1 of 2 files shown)


hermet (commented):
@wenjieshen Now what scenario would use non-optimized path? It looks the optimized path is always be.

wenjieshen:
> @wenjieshen Now what scenario would use non-optimized path? It looks the optimized path is always be.

In the first version of tessellation, the optimized path is always utilized. This patch modifies the sequence of the trim path and optimization.

hermet:
> > @wenjieshen Now what scenario would use non-optimized path? It looks the optimized path is always be.
> 
> In the first version of tessellation, the optimized path is always utilized. This patch modifies the sequence of the trim path and optimization.

@wenjieshen hmm ok, `optimized` variable is necessary?

wenjieshen:
> > > @wenjieshen Now what scenario would use non-optimized path? It looks the optimized path is always be.
> > 
> > 
> > In the first version of tessellation, the optimized path is always utilized. This patch modifies the sequence of the trim path and optimization.
> 
> @wenjieshen hmm ok, `optimized` variable is necessary?

It is unnecessary, serving as a safe lock, to prevent an empty path once tessellation without calling GLGeometry::prepare(). I will remove it. Thank you.

hermet (approved):

Issue #3344, opened 2025-03-23
gl/wg: rendering quality (stroke join?)


hermet:
sw vs gl 
sample:  https://github.com/thorvg/thorvg/blob/main/examples/resources/lottie/cat_loader.json

![Image](https://github.com/user-attachments/assets/24f8bb1e-da2a-423e-a75f-19cfd2284714)

PR #4004, opened 2025-11-25
gl_engine: Fix stroke join rounding using angle-based interpolation

wenjieshen:
Replace linear interpolation with proper angle-based interpolation for stroke join rounding. This ensures correct arc generation by calculating start and end angles and interpolating between them based on orientation (clockwise/counter-clockwise), rather than linearly interpolating between points, which produces incorrect curves.

Related issues: #3344 
### The original interpolation
Screenshot 2025-11-24 201706

---

### The angle-based interpolation
Screenshot 2025-11-24 200702


hermet:
@shenwenjie I'm wondering if you can help this to wgpu as well. Guess they are almost same.

Issue #2095, opened 2024-03-26
svg: supports svg style class set


JSUYA:
```html
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="64" height="64">
<style>
<![CDATA[.B{clip-path:url(#B)}.C{stroke:#f54a42}.D{stroke-width:3.75}.E{stroke-miterlimit:10}.F{clip-path:url(#D)}.G{stroke-linecap:round}]]>
</style>

	<g class="B">
		<path clip-path="url(#F)" d="M10 10 L100 100 Z" fill="none" class="C D"/>
	</g>
</svg>
```

A class can be declared a space-separated set of classes.
https://www.w3.org/TR/SVG2/styling.html#ElementSpecificStyling
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#set-of-space-separated-tokens
In the above case, both C and D must be applied, so stoke color and width must be applied.

hermet:
@JSUYA Hello, can close this? or still need to enhance?

JSUYA:
> [@JSUYA](https://github.com/JSUYA) Hello, can close this? or still need to enhance?

Yes, you can close this with https://github.com/thorvg/thorvg/pull/4022.

PR #4006, opened 2025-11-25
svg_loader: Support style class set

JSUYA:
A class can be declared a space-separated set of classes. https://www.w3.org/TR/SVG2/styling.html#ElementSpecificStyling https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#set-of-space-separated-tokens

```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="64" height="64">
<style>
<![CDATA[.B{clip-path:url(#B)}.C{stroke:#f54a42}.D{stroke-width:10.75}.E{stroke-miterlimit:5}.F{stroke:#FF00FF}.G{stroke-opacity:0.5}]]>
</style>

	<g class="B">
		<path clip-path="url(#F)" d="M10 10 L100 100 Z" fill="none" class="C D E F G C"/>
	</g>
</svg>
```

path's property
.C{stroke:#f54a42}
.D{stroke-width:10.75}
.E{stroke-miterlimit:5}
~~.F{stroke:#FF00FF}~~ <- overwrite
.G{stroke-opacity:0.5}

related issue : https://github.com/thorvg/thorvg/issues/2095

### DIFF
--- a/src/loaders/svg/tvgSvgCssStyle.cpp
+++ b/src/loaders/svg/tvgSvgCssStyle.cpp
@@ -177,7 +177,12 @@ static void _copyStyle(SvgStyleProperty* to, const SvgStyleProperty* from, bool 
 }
 
 
-void _cssCopyStyleAttr(SvgNode* to, const SvgNode* from, bool overwrite = false)
+/************************************************************************/
+/* External Class Implementation                                        */
+/************************************************************************/
+
+
+void cssCopyStyleAttr(SvgNode* to, const SvgNode* from, bool overwrite)
 {
     //Copy matrix attribute
     if (from->transform && (overwrite || !(to->style->flags & SvgStyleFlags::Transform))) {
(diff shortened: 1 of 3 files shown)


src/loaders/svg/tvgSvgCssStyle.cpp
```diff
@@ -249,12 +250,74 @@ void cssApplyStyleToPostponeds(Array<SvgNodeIdPair>& postponeds, SvgNode* style)
...
+    bool allFound = true;
+
+    auto tempNode = tvg::calloc<SvgNode>(1, sizeof(SvgNode));
+    tempNode->style = tvg::calloc<SvgStyleProperty>(1, sizeof(SvgStyleProperty));
+    tempNode->type = node->type;
+    tempNode->style->opacity = 255;
+    tempNode->style->fill.opacity = 255;
+    tempNode->style->stroke.opacity = 255;
+
+    while (ptr < end) {
```
hermet: I’m concerned that the string parsing is being added in a rather ad-hoc way, and that a more consolidated approach might be preferable...

JSUYA: Do you want that work to proceed in this PR?

JSUYA: I've modified the code. Please review it again.

hermet:
@JSUYA Thanks, just for sure, somehow the color is not aligned properly, is it a remained job to do?

expected:

image

current:

image

Issue #4009, opened 2025-11-26
gl_engine: broken Intersect example


hermet:
Normal:

https://github.com/user-attachments/assets/1455a3ad-bd84-4feb-bf51-05c2c7cda457

GL:

https://github.com/user-attachments/assets/41236f31-1dd0-402d-8ed4-191ea0b95586

curlprit:
```
commit 6454b55d3b3ec740cd805cc33eda29de6aec2524
Author: Jay Shen <jay@lottiefiles.com>
Date:   Sun Nov 23 22:25:58 2025 +0800

    gl: fix trimpath flickering regression

    Fix trimpath flickering during animation by reordering path optimization.

    The regression was caused by applying path optimization before trimming,
    which resulted in incorrect geometry for trimmed paths. This commit fixes
    the issue by:
    - Moving optimizePath() call to after trimming operation
    - Applying optimization only to the trimmed path when trimpath is active
    - Ensuring optimizedPath is properly cleared when trimming fails

    Fixes: #3999 
```

PR #4011, opened 2025-11-26
gl: fix broken Intersect example

wenjieshen:
Fix rendering regression in Intersect example caused by improper flag management.

The regression was introduced in commit 6454b55d3 (gl: fix trimpath flickering regression) where the GLGeometry were unconditionally reset, causing incorrect rendering for shapes with composite methods like Intersect.

This commit fixes the issue by:
- Only resetting geometry when RenderUpdateFlag::Path is set
- Conditionally resetting validFill/validStroke flags based on update flags
- Properly initializing flags when creating new GlShape instances

Fixes: #4009

https://github.com/user-attachments/assets/f62cb518-ce16-491c-af71-428bda647006

*Simple performance check: Lottie example shows a frame rate at the 1000th frame of 36.90 fps compared to 36.59 fps, indicating no regression in performance.

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -1266,25 +1266,24 @@ RenderData GlRenderer::prepare(const RenderShape& rshape, RenderData data, const
     if (!sdata) {
         sdata = new GlShape;
         sdata->rshape = &rshape;
+        flags = RenderUpdateFlag::All;
     }
-    sdata->validStroke = sdata->validFill = false;
 
     if ((opacity == 0 && !clipper) || flags == RenderUpdateFlag::None) return sdata;
 
     sdata->viewWd = static_cast<float>(surface.w);
     sdata->viewHt = static_cast<float>(surface.h);
-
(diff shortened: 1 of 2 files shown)


src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1266,25 +1266,25 @@ RenderData GlRenderer::prepare(const RenderShape& rshape, RenderData data, const
     if (!sdata) {
         sdata = new GlShape;
         sdata->rshape = &rshape;
+        flags = RenderUpdateFlag::All;
+        sdata->validFill = sdata->validStroke = false;
```
hermet: Let's initialize them in the structure.

image

hermet (commented):
@wenjieshen nit-pick!

src/renderer/gl_engine/tvgGlRenderer.cpp
wenjieshen: Thank you for reviewing. I have resolved it.

hermet (approved):

Issue #3824, opened 2025-09-19
gl/wg: improve rendering pipeline with matrix revision


hermet:
Use a 3x3 matrix instead of a 4x4 matrix to create a streamlined, more compact data pipeline with improved memory efficiency.

hermet:
Point: **How many** transform(matri)x is mapped to the gpu memory in one single frame.

PR #4020, opened 2025-11-27
gl: refactor matrix handling from mat4 to mat3

wenjieshen:
This PR refactors the GL engine's matrix handling to use 3x3 matrices instead of 4x4 matrices for 2D transformations.

#### Changes

-   **Macro to modern C++**: Replaced preprocessor macros (`MVP_MATRIX`,  `MULTIPLY_MATRIX`,  `GET_MATRIX44`) with  `constexpr`  values and an inline  `getMatrix3Std140()`  function
-   **Memory optimization**: Reduced uniform buffer size from 64 bytes to 48 bytes per matrix (std140-packed mat3)
-   **Simplified shaders**: Updated vertex shaders to use `vec3`/`mat3`operations instead of  `vec4`/`mat4`
-   **Cleaner MVP calculation**: Leverages ThorVG's native `Matrix` struct and  `operator*`  instead of manual 4x4 matrix multiplication

**Impact Analysis**
In the heaviest GL example (`Blending`), the std140 `mat3` path allocates **21,757** matrices (**~1.04 MB** per frame). The previous `mat4` layout would upload **~1.39 MB** for the same frame, This **~0.35MB** reduction in uniform buffer traffic lowers the memory bus bandwidth requirement, which is critical for performance on bandwidth-constrained platforms.

#### ⚠️ Note

This implementation assumes **affine 2D transforms only** (where `e31=0`, `e32=0`, `e33=1`). The perspective divide in gradient shaders (`pos.xy / pos.w`) has been removed since it's unnecessary for affine transforms. **If ThorVG ever supports projective 2D transforms in the future, this would break and need to be revisited.**

**Verified GL rendering output across all examples**

Related issues: #3824

### DIFF
--- a/src/renderer/gl_engine/tvgGlCommon.h
+++ b/src/renderer/gl_engine/tvgGlCommon.h
@@ -28,57 +28,28 @@
 #include "tvgRender.h"
 #include "tvgMath.h"
 
-#define MIN_GL_STROKE_WIDTH 1.0f
-#define MIN_GL_STROKE_ALPHA 0.25f
+constexpr float MIN_GL_STROKE_WIDTH = 1.0f;
+constexpr float MIN_GL_STROKE_ALPHA = 0.25f;
 
-#define MVP_MATRIX(w, h)                \
-    float mvp[4*4] = {                  \
-        2.f / w,      0.0,  0.0f, 0.0f, \
-        0.0,     -2.f / h,  0.0f, 0.0f, \
-        0.0f,        0.0f, -1.0f, 0.0f, \
-        -1.f,         1.f,  0.0f, 1.0f  \
-    };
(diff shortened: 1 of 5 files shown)


hermet:
@wenjieshen could you please resolve all copilot comments? thanks.

wenjieshen:
> @wenjieshen could you please resolve all copilot comments? thanks.

Thank you for the reminder; I thought I had already resolved this. 🙏

hermet:
@tinyjin could you please confirm with web test? thanks.

hermet:
**fps:** ~0 (in the error of margin)
**memory:** ~25% reduction
**binary:** -184byte

hermet:
@wenjieshen great, thanks!

hermet:
@wenjieshen we might wanna 3d perspective support aiming to v2.0, but not v1.0.

Issue #4017, opened 2025-11-27
svg: style class color alignment


hermet:
expected:

<img width="483" height="486" alt="image" src="https://github.com/user-attachments/assets/014cb206-5bce-4a8d-ab6e-f67952919c5e" />

current:

<img width="594" height="430" alt="image" src="https://github.com/user-attachments/assets/a21e7dd6-13fc-4b46-b6c1-d45fc20c267f" />

_Originally posted by @hermet in https://github.com/thorvg/thorvg/issues/4006#issuecomment-3585385410_
            

```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="64" height="64">
<style>
<![CDATA[.B{clip-path:url(#B)}.C{stroke:#f54a42}.D{stroke-width:10.75}.E{stroke-miterlimit:5}.F{stroke:#FF00FF}.G{stroke-opacity:0.5}]]>
</style>

	<g class="B">
		<path clip-path="url(#F)" d="M10 10 L100 100 Z" fill="none" class="C D E F G C"/>
	</g>
</svg>
```

JSUYA:
Oh styles declared later in duplicate should not be applied.
I will fix it.

JSUYA:
https://github.com/thorvg/thorvg/pull/4022

PR #4022, opened 2025-11-28
svg_loader: Prevent duplicate declared classes from overwriting already applied classes

JSUYA:
Manages a list of applied class names. Once a class is applied,
it will not be applied again even if the class is parsed again.

related issue: https://github.com/thorvg/thorvg/issues/4017

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -3606,9 +3606,23 @@ static bool _cssApplyClass(SvgNode* node, const char* classString, SvgNode* styl
 
     char* tokPtr = nullptr;
     auto name = _parseName(classes, " ", &tokPtr);
+    tvg::Array<const char*> applyClasses;
 
     while (name) {
-        bool found = false;
+        auto isDuplicate = false;
+        ARRAY_FOREACH(p, applyClasses) {
+            if (STR_AS(*p, name)) {
+                isDuplicate = true;
+                break;
+            }
+        }
+        if (isDuplicate) {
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3627,6 +3641,7 @@ static bool _cssApplyClass(SvgNode* node, const char* classString, SvgNode* styl
     //Apply the merged style to the node (without overwriting existing styles)
     cssCopyStyleAttr(node, tempNode);
     _free(tempNode);
+    applyClasses.reset();
```
hermet: --

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3606,8 +3606,22 @@ static bool _cssApplyClass(SvgNode* node, const char* classString, SvgNode* styl
 
     char* tokPtr = nullptr;
     auto name = _parseName(classes, " ", &tokPtr);
+    tvg::Array<const char*> applyClasses;
 
     while (name) {
+        bool isDuplicate = false;
```
hermet: bool -> auto

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: Removed

src/loaders/svg/tvgSvgLoader.cpp
JSUYA: Changed

hermet (approved):

PR #4026, opened 2025-11-30
Fix division by zero in `_updateColorTable()`

nikitalita:
thorvg compiled with `ubsan` complains about float division by zero here. 

While this is technically undefined behavior, nearly all implementations will return `inf`, which will eventually result in an alpha value of `255` (since alpha is a uint8_t); Setting it to `0.0` in case of a divide by zero here will result in the same behavior.

### DIFF
--- a/src/renderer/sw_engine/tvgSwFill.cpp
+++ b/src/renderer/sw_engine/tvgSwFill.cpp
@@ -163,7 +163,8 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface* 
 
         auto curr = colors + j;
         auto next = curr + 1;
-        auto delta = 1.0f / (next->offset - curr->offset);
+        auto div = next->offset - curr->offset;
+        auto delta = div != 0.0f ? (1.0f / div) : 0.0f;
         auto a2 = MULTIPLY(next->a, opacity);
 
         if (!fill->translucent && a2 < 255) fill->translucent = true;


hermet:
> thorvg compiled with `ubsan` complains about float division by zero here.
> 
> While this is technically undefined behavior, nearly all implementations will return `inf`, which will eventually result in an alpha value of `255` (since alpha is a uint8_t); Setting it to `0.0` in case of a divide by zero here will result in the same behavior.

@nikitalita Hello, could you please suggest the example you made this scenario? Thanks.

nikitalita:
the specific instance where ubsan threw this error was when attempting to render this SVG to a software canvas:
CanvasItemMaterial

Using the default width and height for the image of 16x16 (stride: 16) with a color space of `tvg::SwCanvas::ABGR8888S`

Here's the specific code (ignore `p_image`, assume a scale of 1.0):
```c++
Error ImageLoaderSVG::create_image_from_utf8_buffer(Ref<Image> p_image, const uint8_t *p_buffer, int p_buffer_size, float p_scale, bool p_upsample) {
	ERR_FAIL_COND_V_MSG(Math::is_zero_approx(p_scale), ERR_INVALID_PARAMETER, "ImageLoaderSVG: Can't load SVG with a scale of 0.");

	std::unique_ptr<tvg::Picture> picture = tvg::Picture::gen();

	tvg::Result result = picture->load((const char *)p_buffer, p_buffer_size, "svg", true);
	if (result != tvg::Result::Success) {
		return ERR_INVALID_DATA;
	}
	float fw, fh;
	picture->size(&fw, &fh);

	uint32_t width = MAX(1, std::round(fw * p_scale));
	uint32_t height = MAX(1, std::round(fh * p_scale));

	const uint32_t max_dimension = 16384;
	if (width > max_dimension || height > max_dimension) {
		WARN_PRINT(vformat(
				String::utf8("ImageLoaderSVG: Target canvas dimensions %d×%d (with scale %.2f) exceed the max supported dimensions %d×%d. The target canvas will be scaled down."),
				width, height, p_scale, max_dimension, max_dimension));
		width = MIN(width, max_dimension);
		height = MIN(height, max_dimension);
	}

	picture->size(width, height);

	std::unique_ptr<tvg::SwCanvas> sw_canvas = tvg::SwCanvas::gen();
	Vector<uint8_t> buffer;
	buffer.resize(sizeof(uint32_t) * width * height);

	tvg::Result res = sw_canvas->target((uint32_t *)buffer.ptrw(), width, width, height, tvg::SwCanvas::ABGR8888S);
	if (res != tvg::Result::Success) {
		ERR_FAIL_V_MSG(FAILED, vformat("ImageLoaderSVG: Couldn't set target on ThorVG canvas, error code %d.", res));
	}

	res = sw_canvas->push(std::move(picture));
	if (res != tvg::Result::Success) {
		ERR_FAIL_V_MSG(FAILED, vformat("ImageLoaderSVG: Couldn't insert ThorVG picture on canvas, error code %d.", res));
	}

	res = sw_canvas->draw();
	if (res != tvg::Result::Success) {
		ERR_FAIL_V_MSG(FAILED, vformat("ImageLoaderSVG: Couldn't draw ThorVG pictures on canvas, error code %d.", res));
	}

	res = sw_canvas->sync();
	if (res != tvg::Result::Success) {
		ERR_FAIL_V_MSG(FAILED, vformat("ImageLoaderSVG: Couldn't sync ThorVG canvas, error code %d.", res));
	}

	p_image->set_data(width, height, false, Image::FORMAT_RGBA8, buffer);

	res = sw_canvas->clear(true);

	return OK;
}
```

hermet:
@nikitalita please let me know if you are expecting v0.15.x release with this fix. thanks.

nikitalita:
Yes, that would be nice. Godot uses the 0.xx branch downstream and we would like to update to fix ubsan tests.

Issue #2710, opened 2024-09-04
svg: handle spaces inside the text node


mgrudzinska:
seems that multiple spaces should be treated as single ones, at the beginning/end all should be ignored (before/after tspan are treated as in the middle of the text) 
[tt.svg.zip](https://github.com/user-attachments/files/16882017/tt.svg.zip)

capnm:
Here are some comments on this
https://razrfalcon.github.io/notes-on-svg-parsing/text/whitespaces.html

hermet:
thorvg:

![Image](https://github.com/user-attachments/assets/607f2fdb-7f19-4790-a674-43009a574622)

expected:
![Image](https://github.com/user-attachments/assets/ba61c625-c9dc-4675-9d5c-0bd0fd9ed63f)

PR #4031, opened 2025-12-02
svg_loader: Support for the xml:space attribute in <text>

JSUYA:
Supports `xml:space=“preserve”` and `"default"` attributes for `<text>`. The `“preserve”` attribute outputs all whitespace defined within `<text>` tag exactly as written(newline excluded). Since our XML parser removes all leading and trailing whitespace from CDATA(DATA) before passing it to the loader, it passes the text without removing whitespace when the `openedTag` is <text>.

example
```html
<svg viewBox="0 0 1000 1000" xmlns="http://www.w3.org/2000/svg">
    <!-- TEST 1 -->
    <text x="0" y="100" font-family="Arial" fill="#0000ff" font-size="50" xml:space="default">     SP     ACE    <tspan></tspan>   S
    test </text>
    <rect width="330" height="60" x="0" y="50" fill="none" stroke="#ff00ff" stroke-width="2" />

    <!-- TEST 2 -->
    <text x="0" y="200" font-family="Arial" fill="#0000ff" font-size="50" xml:space="preserve">     SP     ACE    <tspan></tspan>   S
    test </text>
    <rect width="330" height="60" x="0" y="150" fill="none" stroke="#ff00ff" stroke-width="2" />
</svg>
```

related issue : https://github.com/thorvg/thorvg/issues/2710

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -851,6 +851,47 @@ static void _applyTextFill(SvgStyleProperty* style, Text* text, const Box& vBox)
 }
 
 
+static char* _processText(const char* text, SvgXmlSpace space)
+{
+    if (!text) return nullptr;
+
+    auto len = strlen(text);
+    auto processed = (char*)tvg::malloc(len + 1);
+    auto dst = processed;
+    auto src = text;
+
+    if (space == SvgXmlSpace::Preserve) {
+        while (*src) {
+            if (*src == '\n' || *src == '\t' || *src == '\r') *dst++ = ' ';
+            else *dst++ = *src;
+            src++;
+        }
+        *dst = '\0';
+    } else {
+        auto spaceFound = false;
+        // Skip leading spaces
+        while (*src && isspace((unsigned char)*src)) src++;
+
+        while (*src) {
+            if (isspace((unsigned char)*src)) {
+                if (!spaceFound) {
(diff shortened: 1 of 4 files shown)


src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -851,6 +851,47 @@ static void _applyTextFill(SvgStyleProperty* style, Text* text, const Box& vBox)
 }
 
 
+static char* _processText(const char* text, SvgXmlSpace space)
```
hermet, line 854: Too much ad-hoc style addition, Parsing logics everywhere. See `tvgSvgLoader`.  We need to consider a well-organized and structured implementation. Consolidated parser in one place might be better to go.

JSUYA: This _processText() function isn't a parsing login.
SvgLoader basically stores the original parsed data. SceneBuilder uses the original data and node information to build the scene.  data is not transformed in SvgLoader.
The reason processText is needed is to create the appropriate text data according to SvgXmlSpace.

I agree with the idea that there's some redundant logic in svg loader, but i think.. this case is not.

hermet: Whether it's parsing or not, it's not important. We can consider an consolidate class to handle string iteration, searching and handling. Everywhere, svg has this kind of duplicate functions with ad-hoc style. :(

Issue #3561, opened 2025-06-19
wg_engine: clipping issue


mgrudzinska:
samples:

[clip.svg.zip](https://github.com/user-attachments/files/20817560/clip.svg.zip)
[clip2.svg.zip](https://github.com/user-attachments/files/20817559/clip2.svg.zip)

<img width="558" alt="Image" src="https://github.com/user-attachments/assets/788528d2-fc4e-46a8-bde1-e1359eca8550" />

SergeyLebedkin:
@hermet @BenjaminHalko 
Looks like wrong handling for cascade clipping handling for gl/wg

```cpp
        // clipper 0
        auto clip0 = tvg::Shape::gen();
        clip0->appendCircle(400, 500, 200, 200);
        clip0->fill(0, 0, 255);

        // clipper 1
        auto clip1 = tvg::Shape::gen();
        clip1->appendCircle(400, 300, 200, 200);
        clip1->fill(0, 255, 0);
        clip1->clip(clip0); // <===== clipped by clip0

        // shape
        auto circle = tvg::Shape::gen();
        circle->appendCircle(400, 400, 300, 300);
        circle->fill(255, 0, 0);
        circle->clip(clip1); // <===== clipped by clip1
        canvas->push(circle);
```
sw/(gl and wg)

<img width="854" height="442" alt="Image" src="https://github.com/user-attachments/assets/1377f31b-2e4c-49e8-8e35-e7f4f31f34bc" />

PR #4035, opened 2025-12-02
wg_engine: fix clipping issue

SergeyLebedkin:
before viewport of the clipper shape was ignored
fixed by adding clipper viewport appliance

https://github.com/thorvg/thorvg/issues/3561


wenjieshen (approved):
image

Thanks for the contribution. The issued case passed, and running all examples was overserved with significant regression.

PR #4036, opened 2025-12-03
svg_loader: Refactor the logic for skipping white space

JSUYA:
There is some duplicate logic to skip whitespace, so I'll clean it up.

https://github.com/thorvg/thorvg/pull/4031#discussion_r2580124909

### DIFF
--- a/src/loaders/svg/tvgSvgUtil.cpp
+++ b/src/loaders/svg/tvgSvgUtil.cpp
@@ -38,6 +38,34 @@ static uint8_t _hexCharToDec(const char c)
 /* External Class Implementation                                        */
 /************************************************************************/
 
+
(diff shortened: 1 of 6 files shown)


src/loaders/svg/tvgSvgUtil.cpp
```diff
@@ -38,6 +38,34 @@ static uint8_t _hexCharToDec(const char c)
 /* External Class Implementation                                        */
 /************************************************************************/
 
+
+const char* skipWhiteSpace(const char* itr, const char* itrEnd)
```
hermet: Suggest to use "svg" prefix in function name. thanks.

JSUYA: Updated function names

PR #4043, opened 2025-12-04
example: Resize example size to fit the window size

JSUYA:
The existing example fixed the canvas to the size specified in the window. 
So, even when the window size changed, the example did not resized, and noise appeared in the background.

Add a root scene(and background) at the top of paint tree added to the canvas. When the window size changes, adjust it to fit the new window size. Output it by fixing the aspect ratio based on the initial size.

+) Tested on sw, gl, not tested on wg.

### DIFF
--- a/examples/Example.h
+++ b/examples/Example.h
@@ -67,8 +67,8 @@ struct Example
 {
     uint32_t elapsed = 0;
 
-    virtual bool content(tvg::Canvas* canvas, uint32_t w, uint32_t h) = 0;
-    virtual bool update(tvg::Canvas* canvas, uint32_t elapsed) { return false; }
+    virtual bool content(tvg::Canvas* canvas, tvg::Scene* root, uint32_t w, uint32_t h) = 0;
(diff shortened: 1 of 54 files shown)


examples/Example.h
```diff
@@ -308,8 +321,49 @@ struct Window
...
+        }
+    }
+
     virtual void refresh() {}
+
+    void repack()
+    {
+        auto& paints = canvas->paints();
+
+        if (paints.size() == 2 && paints.front() == bg && paints.back() == root) return;
```
hermet: This repack step doesn't good as for the official user guidance...

JSUYA: So.. what do you think about passing root instead of canvas to example->content()?

JSUYA: I removed repack() and modified example->content() to pass root so that all content is pushed to the root scene.

JSUYA:
https://github.com/thorvg/thorvg/pull/4055

PR #4044, opened 2025-12-04
gl: Skip stencil usage for convex GL fills

wenjieshen:
To skip the stencil for convex GL fills, we track convexity during tessellation and expose this information through `GlGeometry::isConvex()`. Convex shapes can now be rendered with direct fills instead of stencil passes, which reduces overall workload.

Key changes include:

- Tracking winding and convexity in `tvgGlTessellator` while building paths and propagating this data to `GlGeometry`.
- Adding `isConvex()` to `GlGeometry` to allow us to bypass stencil mode for convex fills.
- Keeping non-convex paths unchanged to maintain correct winding behavior.

Performance improvements:

- Speedup of 31% FPS in one specific case:
  image

- Speedup of 6% FPS in the `Lottie` example.

- Reduced Graphic API calls by 32.1% in one specific case:
  image

No regressions were found in any examples or previously problematic test files.

### DIFF
--- a/src/renderer/gl_engine/tvgGlTessellator.h
+++ b/src/renderer/gl_engine/tvgGlTessellator.h
@@ -80,6 +80,7 @@ public:
     BWTessellator(GlGeometryBuffer* buffer);
     void tessellate(const RenderPath& path, const Matrix& matrix);
     RenderRegion bounds() const;
+    bool convex = true;
 
 private:
     uint32_t pushVertex(float x, float y);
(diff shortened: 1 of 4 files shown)


hermet:
Lottie: +10% fps 
Torus: +28% fps

hermet:
The grid is invisible on the surface on gl version:

gl(wg) vs software:

image

src/renderer/gl_engine/tvgGlTessellator.h
```diff
@@ -80,13 +80,19 @@ class BWTessellator
     BWTessellator(GlGeometryBuffer* buffer);
     void tessellate(const RenderPath& path, const Matrix& matrix);
     RenderRegion bounds() const;
+    bool isConvex() const { return convex; }
```
hermet: No need setter/getter, you can just use public convex

hermet (commented):
Please check a comment and commit msg:
starting title `gl_engine: xxxxxxx`

Issue #4041, opened 2025-12-04
gpu: alpha-premultiplied bug regressions


hermet:
GL/WGPU:

<img width="1140" height="1141" alt="Image" src="https://github.com/user-attachments/assets/3d56a0f9-e8e9-41a0-b155-979a9730a945" />

Expected:

<img width="1049" height="1035" alt="Image" src="https://github.com/user-attachments/assets/364bd7c1-4de9-433c-ad47-ce67d1c360bd" />

[test.zip](https://github.com/user-attachments/files/23922525/test.zip)

wenjieshen:
## Root Cause Analysis:

### Constraints
- **Hardware Filtering Occurs Before the Shader**: The GPU's texture sampler performs linear interpolation (blending pixels) prior to executing our Fragment Shader code.

- **Irreversible Data Corruption**: When the GPU's texture sampler interpolates straight alpha texels (e.g., the anti-aliasing feather in the color [0.996, 0.996, 0.996, 0.866] and transparent 0.0), it mathematically alters the RGB channels before we can access them.

    - By the time the data reaches the shader, the RGB values are already partially attenuated.

    - Applying `color.rgb * color.a` in the shader results in double attenuation, which causes the dark edge effect.

### Performance Cost of Alternatives
The only way to bypass this issue on the GPU is to disable hardware filtering, which leads to aliasing and jagged edges, or to manually fetch four neighboring pixels and interpolate them in code, which incurs a severe performance penalty.

### Conclusion
To achieve smooth resizing (GL_LINEAR) without artifacts, **the texture data must be premultiplied before it reaches the GPU's texture sampler.**

```mermaid
sequenceDiagram
    participant T as Texture Data
    participant S as GPU Sampler (Hardware)
    participant F as Fragment Shader

    Note over S: The Bottleneck

    rect rgb(40, 20, 20)
    Note right of T: Path A: Straight Alpha (Current)
    T->>S: RGBA (Straight)
    S->>S: ❌ Linear Filter blends RGB<br/>(Data is now mathematically incorrect)
    S->>F: Passes "Corrupted" Values
    F->>F: Shader Math Fails
    end

    rect rgb(20, 40, 20)
    Note right of T: Path B: CPU Premultiplication (Required)
    T->>T: ✅ CPU Premultiplies Data
    T->>S: RGBA (Premultiplied)
    S->>S: ✅ Linear Filter Blends Correctly
    S->>F: Passes Correct Values
    end
```

[...]

PR #4046, opened 2025-12-05
gl: convert straight alpha to premultiplied alpha on texture upload

wenjieshen:
- Move straight-alpha premultiplication from fragment shader to CPU during texture upload, using a buffered conversion path with OpenMP to keep upload O(n) but off the GPU hot path.
- Set texture color space to premultiplied variants after conversion so downstream pipeline treats uploads consistently.
- Update the GL image shader to indicate there is no long straight color space in the GL image texture.

---

The problematic file in https://github.com/thorvg/thorvg/issues/2863 performs correctly.
Screenshot 2025-12-05 161721

Screenshot 2025-12-05 162111

---

- The boot time recorded for the example `PicturePng` shows a difference of less than 1%.
- FPS in Lottie shows a difference of less than 0.05% after 2000 frames.
- Running all examples, particularly `Picture*` and `Image*` is not observed significant regression.

Related issues #4041

### DIFF
--- a/src/renderer/gpu_engine/gl/meson.build
+++ b/src/renderer/gpu_engine/gl/meson.build
@@ -42,7 +42,14 @@ else
     compiler_flags += ['-DTHORVG_GL_TARGET_GL=1']
 endif
 
+omp_dep = []
+
+if openmp
+    omp_dep = dependency('openmp', required: false)
+endif
+
 engine_dep += [declare_dependency(
     include_directories : include_directories('.'),
+    dependencies        : omp_dep,
     sources             : source_file,
 )]
--- a/src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
+++ b/src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
@@ -1261,6 +1261,16 @@ void GlRenderer::dispose(RenderData data)
     delete sdata;
 }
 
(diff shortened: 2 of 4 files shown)


SergeyLebedkin (approved):
will grad the solution for wg
approved

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1205,14 +1205,40 @@ void GlRenderer::dispose(RenderData data)
...
     GLuint tex = 0;
 
     GL_CHECK(glGenTextures(1, &tex));
 
     GL_CHECK(glBindTexture(GL_TEXTURE_2D, tex));
-    GL_CHECK(glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA8, image->w, image->h, 0, GL_RGBA, GL_UNSIGNED_BYTE, image->data));
+
+    if (!image->premultiplied && image->channelSize == sizeof(uint32_t)) {
+        auto buffer = (uint32_t*)malloc(image->w * image->h * sizeof(uint32_t));
+        auto h = static_cast<int32_t>(image->h);
```
hermet: @wenjieshen openmp is not the best for this. can't use the gpu power?

wenjieshen:
Convert to draft until completion of the GPGPU version by referring to WebGPU engines.

PR #4051, opened 2025-12-09
test: clean up the lottie unit-test

hermet:
- neat & clean test suites for minimize the workload

image


hermet:
@tinyjin Please check this, you can append more test-suites if it needs (ie. test4.lot, test5.lot ...). Any sample doesn't increase the coverage, please don't add or remove the older one. Thanks.

image

Issue #4033, opened 2025-12-02
gl_engine: blending anti-aliasing quality matter


hermet:
Some edge anti-aliasing blending appears incorrect in terms of alpha premultiplication.

Examples -> Blending

see: Luminosity, Darken, HardLight

<img width="356" height="209" alt="Image" src="https://github.com/user-attachments/assets/e642ba7d-8bc4-4586-bbad-5c396ed4ea39" />

wenjieshen:
### Root Cause Analysis

The SW and GL engines have fundamentally different rendering pipelines for blend modes:

**SW Engine Pipeline:**
- Anti-aliasing coverage is applied **after** blending via `INTERPOLATE(blendResult, dst, coverage)`

**GL Engine Pipeline:**
- Source shape is first rendered to a texture **WITH** anti-aliasing already baked in (premultiplied alpha)

### Comparison

| Aspect | SW Engine | GL Engine |
|--------|-----------|-----------|
| When AA is applied | After blending | Before blending |
Alpha handling | `BLEND_PRE()` | `mix(d.Sc, Rc, d.Da)`|
| Source unpremultiply | No | No |
|Edge case(Da = 0) | Safe div | fallback to d.Sc|

### Possible Solutions
1. ~~**Skip HSL blend for AA edges**: For pixels where source alpha < 1.0, fall back to normal alpha compositing instead of HSL blending~~ -> It breaks all transparent shapes, not just AA.
2. ~~**Unpremultiply behavior**: Unpremultiply source in GL shader to relief the dark outline~~ -> It just breaks the reuslt
3. **Derivative Reconstruction**: Use `fwidth` to get the gradient, then restore the color on the AA edges -> It may cause other artifact.
4. **Change GL pipeline**: Render source shapes without AA to the blend texture, apply AA after blending (significant pipeline change)

hermet:
@wenjieshen it's good to check wgpu implementation, rather than sw engine. cc:  @SergeyLebedkin

wenjieshen:
> [@wenjieshen](https://github.com/wenjieshen) it's good to check wgpu implementation, rather than sw engine. cc: [@SergeyLebedkin](https://github.com/SergeyLebedkin)

Thank you. I have checked the wgpu implemenation. It is also caused by AA timing.

  | WG | GL
-- | -- | --
Source | uPaintSettings.color (uniform) | texture(uSrcTexture, vUV)
Shape masking | Stencil | Alpha channel in FBO
[...]

Issue #3175, opened 2025-02-03
gl_engine: Incorrectly blending at the anti-aliased parts


hermet:
GL incorrectly blends the anti-aliased parts.

![Image](https://github.com/user-attachments/assets/2191d538-3f28-4222-b82d-917e89eac8d8)

_Originally posted by @hermet in https://github.com/thorvg/thorvg/issues/3163#issuecomment-2630046169_

wenjieshen:
This issue is similar to #4033. The input anti-aliased texture [^1] already exhibits a halo effect. When MSAA resolves the render target, it averages the samples again as follows. As a result, the halo effect is further emphasized.

[^1]: <img width="320" height="264" alt="Image" src="https://github.com/user-attachments/assets/35bdc6cc-5d88-4abc-8ba7-7a6115c2b444" />
<img width="158" height="131" alt="Image" src="https://github.com/user-attachments/assets/af42340f-1b9a-4c32-968f-f2a8d08bb7fa" />
<img width="161" height="133" alt="Image" src="https://github.com/user-attachments/assets/bf0f5be8-dd6f-475a-8e39-1a9f10abf251" />
<img width="162" height="132" alt="Image" src="https://github.com/user-attachments/assets/dc839aae-33cf-4992-b3aa-1681901e3ca6" />

PR #4053, opened 2025-12-09
gl: add direct shape blending path and shaders

wenjieshen:
Replace GlSimpleBlendTask with GlDirectBlendTask to perform blending directly in the shape shader without a separate compose pass. This copies only the required destination region and blends in a single draw.

- Introduced GlDirectBlendTask to blit the destination region and run shape blends directly instead of using the old compose path.
- Added shape-oriented blend programs (solid, linear, radial) selected via BlendSource and new shader headers with BlendRegion uniforms.
- Extracted reusable gradient helper functions to keep shader mains minimal while supplying the new blend shaders.
- Remove redundant blend shader headers

---

### Render to texture 
There are **two** textures. The source texture has been anti-aliased, which causes a halo effect due to its softened edges.
Screenshot 2025-12-08 214536

### Direct blending
There is only **one** textue. The MSAA is resolved after blending.
Screenshot 2025-12-08 203840

---

### Result
Screenshot 2025-12-08 203909
Screenshot 2025-12-08 181857

**Performance doesn't change, difference less than 0.1%** 
**No regressions found druing execution of all examples.**
issue: https://github.com/thorvg/thorvg/issues/4033, https://github.com/thorvg/thorvg/issues/3175

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderTask.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderTask.cpp
@@ -417,32 +417,35 @@ void GlClipTask::normalizeDrawDepth(int32_t maxDepth)
 }
 
 /************************************************************************/
-/* GlSimpleBlendTask Class Implementation                               */
+/* GlDirectBlendTask Class Implementation                               */
 /************************************************************************/
 
-GlSimpleBlendTask::GlSimpleBlendTask(BlendMethod method, GlProgram* program)
- : GlRenderTask(program), mBlendMethod(method)
- {
- }
-
-
-void GlSimpleBlendTask::run()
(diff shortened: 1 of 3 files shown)


hermet:
> * Refactor gradient functions for reuse in blend shaders
> * Add shape-specific blend shaders (solid, linear, radial)
> * Simplify blend program selection with BlendSource enum
> * Remove redundant blend shader headers

@wenjieshen Please note that, when you really want to get a code review but the changes are not easy to read, separating commits is mandatory rather than mixing everything into a single commit.

wenjieshen:
> > * Refactor gradient functions for reuse in blend shaders
> > * Add shape-specific blend shaders (solid, linear, radial)
> > * Simplify blend program selection with BlendSource enum
> > * Remove redundant blend shader headers
> 
> @wenjieshen Please note that, when you really want to get a code review but the changes are not easy to read, separating commits is mandatory rather than mixing everything into a single commit.

I apologize for the inconvenience. I have tried separating commits. The modified version has redundant code, and I will create a PR to remove it. Thank you.

PR #4055, opened 2025-12-09
examples: Enhance update() of examples

JSUYA:
Add for the window's width and height to the update(). Each example can receive the actual size of the canvas.
The SVG and PictureSVG examples now include an example that adjusts the scale of the content based on the actual size of the Canvas.

### DIFF
--- a/examples/Animation.cpp
+++ b/examples/Animation.cpp
@@ -61,7 +61,7 @@ struct UserExample : tvgexam::Example
         return true;
     }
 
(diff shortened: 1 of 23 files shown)


JSUYA:
https://github.com/thorvg/thorvg/pull/4043#pullrequestreview-3550015863
@hermet I've improved the way each example uses a canvas size. I think this is a better approach than fixing the window size.

JSUYA:
https://github.com/thorvg/thorvg/pull/4062

Issue #4024, opened 2025-11-29
ThorVG 1.0 Release


hermet:
### Candidates Tasks before v1.0
- [x] https://github.com/thorvg/thorvg/issues/4077
- [x] https://github.com/thorvg/thorvg/issues/4088
- [x] https://github.com/thorvg/thorvg/issues/4082
- [x] https://github.com/thorvg/thorvg/issues/4067
- [x] https://github.com/thorvg/thorvg/issues/3755
- [x] https://github.com/thorvg/thorvg/issues/3960
- [x] https://github.com/thorvg/thorvg/issues/3981
- [x] https://github.com/thorvg/thorvg/issues/3385
- [x] https://github.com/thorvg/thorvg/issues/3314
- [x] https://github.com/thorvg/thorvg/issues/3024
- [x] https://github.com/thorvg/thorvg/issues/4112
- [x] https://github.com/thorvg/thorvg/issues/4078
- [x] https://github.com/thorvg/thorvg/issues/3561
- [x] https://github.com/thorvg/thorvg/issues/4037
- [x] https://github.com/thorvg/thorvg/issues/2682
- [x] https://github.com/thorvg/thorvg/issues/1729

### API Clean UP
- [x] https://github.com/thorvg/thorvg/issues/3116
- [x] Bump up version from thorvg headers. https://github.com/thorvg/thorvg/commit/97fa741d720252482e136df3774dcbf47bd9142f

### Unit Test
-  https://github.com/thorvg/thorvg/wiki/Unit-Tests
- [x] Make it up 80% coverage

### Build/Regression Test
- https://github.com/thorvg/thorvg/wiki/Release-Management
- [x] No Regressions
- [x] Build errors / warnings (Web/Ubuntu/Windows/MacOS)

### Performance Comparison (v0.15 vs v1.0 / Intel Core 5, Intel Arc)
- [x] **Stress**
  - [x] cpu = **+1%** (60.4 -> 61.2)
  - [x] gl/es = **+58%** (39.1 -> 62.8)
  - [x] wgpu = **new** (60.2)
 
<img width="300" height="auto" alt="Image" src="https://github.com/user-attachments/assets/20450e8e-c890-4dd4-8553-7824b3dadbd7" />

- [x] **Lottie**
  - [x] cpu = **+4%** (28.2 -> 29.2)
  - [x] gl/es = **new** (19)
  - [x] wgpu  =  **new** (13.2)

[...]

PR #4075, opened 2025-12-27
infra: add versioned build support for thorvg-1

hermet:
rename output library to `libthorvg-1.so` with proper SONAME (`libthorvg-1.so.1`) and generate thorvg-1.pc via pkg-config for versioned dependency resolution.

This will avoid conflict with v0.x thorvg.

Please note that now you can use for v1.x thorvg
```
- #include <thorvg-1/thorvg.h>
- lthorvg-1
```
and use for v0.x thorvg:
```
- #include <thorvg.h>
- lthorvg
```
issue: https://github.com/thorvg/thorvg/issues/4024

### DIFF
--- a/.github/workflows/build_android.yml
+++ b/.github/workflows/build_android.yml
@@ -34,7 +34,7 @@ jobs:
         API: 21
       run: |
         sed -e "s|NDK|$NDK|g" -e "s|HOST_TAG|linux-x86_64|g" -e "s|API|$API|g" ./cross/android_x86_64.txt > /tmp/android_cross.txt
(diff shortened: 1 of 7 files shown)


vtorri:
I would suggest that thorvg-$version_maj to automatically use the majot version of thorvg. In case thorvg 2 is released some day...

hermet:
> I would suggest that thorvg-$version_maj to automatically use the majot version of thorvg. In case thorvg 2 is released some day...

@vtorri Please send a PR, if it really matters to you. Thanks!

Issue #4078, opened 2025-12-29
wg_engine: regression


hermet:
stroke is invisible..

<img width="1274" height="1282" alt="Image" src="https://github.com/user-attachments/assets/63fc2e5b-df1a-4d07-9868-cf9ab909e7b6" />

```
using namespace tvg;

struct UserExample : tvgexam::Example
{
    unique_ptr<tvg::Animation> animation;
    vector<Shape*> shapes;
    int cnt = 2000;

    bool content(tvg::Canvas* canvas, uint32_t w, uint32_t h) override
    {
        for (int i = 0; i < 2000; ++i) {
            auto s = Shape::gen();
            s->appendRect(rand() % 1600, rand() % 1600, rand() % 100, rand() % 100, rand() % 10, rand() % 10);
            s->fill(rand() % 255, rand() % 255, rand() % 255, 253 + rand() % 3);
            s->strokeFill(255, 255, 255);
            s->strokeWidth(3);
            canvas->push(s);
            shapes.push_back(s);
        }

        return true;
    }

    bool update(tvg::Canvas* canvas, uint32_t elapsed) override
    {
        for (auto s: shapes) {
            s->reset();
            s->appendRect(rand() % 1600, rand() % 1600, rand() % 100, rand() % 100, rand() % 10, rand() % 10);
        }

        canvas->update();

        return true;
    }
};
```

PR #4079, opened 2025-12-29
wg_engine:fixed a missing stroke mesh update

hermet:
trigger stroke mesh update if path is only updated.

issue: https://github.com/thorvg/thorvg/issues/4078


src/renderer/wg_engine/tvgWgRenderData.cpp
```diff
@@ -166,7 +166,8 @@ void WgRenderDataShape::updateVisibility(const RenderShape& rshape, uint8_t opac
 
 void WgRenderDataShape::updateMeshes(const RenderShape &rshape, RenderUpdateFlag flag, const Matrix& matrix)
 {
-    releaseMeshes();
+    releaseMeshes();  //OPTIMIZE: bad idea to reset meshes always. it could re-use the meshes if there hasn't no path changes.
```
hermet: @wenjieshen @SergeyLebedkin

Issue #4082, opened 2025-12-29
svg: gradient rendering is incorrectly affected by element transform (Paragraphic)


lostminds:
Rendering of gradient fill / stroke on elements appears to be transformed incorrectly by the transform set on the element.

For example 

```
<?xml version="1.0" encoding="utf-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="500px" height="500px" viewBox="0 0 500 500" xml:space="preserve">
<defs>
<linearGradient id="Gradient1">
<stop offset="0.0" stop-color="rgb(255,0,0)"/>
<stop offset="0.5" stop-color="rgb(0,0,0)"/>
<stop offset="1.0" stop-color="rgb(0,0,255)"/>
</linearGradient>
</defs>
<rect fill="url(#Gradient1)" x="-50" y="50" transform="translate(100,0)" width="300" height="100"/>
<rect fill="url(#Gradient1)" x="50" y="200" transform="translate(0,0)" width="300" height="100"/>
</svg>
```

Should render two rectangles with the same red-black-blue gradient (screenshot from Firefox):

![Image](https://github.com/user-attachments/assets/16f6268f-9e4c-4171-a32f-cd5b28d789ac)

However, in the ThorVG web viewer 1.0.0-pre34 the top rect gradient is offset by the transform on the element:

![Image](https://github.com/user-attachments/assets/5790d929-9bd3-4702-9202-88db257f70f9)

This seems to be a regression in 1.0-something as the same issue does not appear to be present in 0.15.13 (As embedded in Godot 4.5.1)

Svg test file:
![Image](https://github.com/user-attachments/assets/4488b028-c464-4991-a09e-5f9d3fbda508)

hermet:
+ @JSUYA

JSUYA:
I found that the patch below was not exist in 0.15.x, but only in the main branch. And after applying this patch, I found that the gradient position was incorrect. However, the codes applied to the svg loader did not seem to be particularly problematic.

606│ 333e65ef7a 2025-03-19 Hermet Park            o renderer: revise the Bounding Box Behavior

lostminds:
@hermet I see you added (Paragraphic) to the title, and while I found this issue working on Paragraphic it is not affected by the problem since it uses a 0.15.x build of ThorVG. 

[...]

PR #4083, opened 2025-12-30
svg_loader,renderer : fix incorrect gradient mapping by providing untransformed bounds

JSUYA:
This commit fixes a regression(333e65e) in SVG gradient rendering where gradients were being incorrectly mapped when the target element had a transformation. The root cause was that Paint::bounds() was always applying the local transform, even when untransformed bounds were requested (by passing nullptr as the parent matrix).

Changes:
- renderer: modify Paint::Impl::bounds to ignore local transform if pm is nullptr.
- svg_loader: update _bounds() helper to use the fixed bounds() API.

https://github.com/thorvg/thorvg/issues/4082

### DIFF
--- a/src/loaders/svg/tvgSvgSceneBuilder.cpp
+++ b/src/loaders/svg/tvgSvgSceneBuilder.cpp
@@ -50,9 +50,18 @@ static inline bool _isGroupType(SvgNodeType type)
 //a stroke width should be ignored for bounding box calculations
 static Box _bounds(Paint* paint)
 {
-    float x, y, w, h;
-    paint->bounds(&x, &y, &w, &h);
-    return {x, y, w, h};
+    Point pt4[4] = {};
+    if (PAINT(paint)->bounds(pt4, nullptr, true)) {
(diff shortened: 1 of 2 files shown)


hermet (commented):
Hackish is not really preferred. It could acquire the local space bbox if it requests bbox before attaching the paint to a parent.

JSUYA:
Currently, there are issues with the output of #4082, as well as 1372943623.svg and circles1.svg. Even after applying this PR, I found that 1372943623.svg still has issues. this PR is not appropriate and i will close it.

However, I dont know that this method is hackish.
I think the behavior of paint->bounds() has changed. and I'm not sure if this can be solved by simply changing the attach order in svg_loader. I haven't found a way yet.
Can this problem be solved in the renderer?

hermet:
@JSUYA I couldn't find any issues at `1372943623`, `circles1`. Please let me know if I'm missing there. Thanks.
`1372943623` 
tvg vs chrome
image

`circle1`
tvg vs chrome
image

hermet:
https://github.com/thorvg/thorvg/pull/4086

PR #4089, opened 2026-01-06
Fix DLL name

vtorri:


hermet:
replaced https://github.com/thorvg/thorvg/pull/4090

Issue #4088, opened 2026-01-06
Building on Windows creates thorvg-1-1.dll


net147:
Building latest commit of main branch on Windows creates thorvg-1-1.dll and thorvg-1-1.pdb.
Is the duplicated major version number expected?

vtorri:
on Windows, meson automatically adds the major version to the DLL name. I can provide a patch that fixes this, if you want

hermet:
@vtorri please, thanks!

hermet:
@net147 @vtorri The fix is in main branch. thanks!

net147:
> [@net147](https://github.com/net147) [@vtorri](https://github.com/vtorri) The fix is in main branch. thanks!

@vtorri I get "ld.exe: cannot find -lthorvg-1" when trying to link against ThorVG with MinGW on Windows.
It can't find thorvg-1 because the library file is named thorvg.lib. It needs to be renamed to thorvg-1.lib. After manually renaming thorvg.lib to thorvg-1.lib it works as expected.

PR #4091, opened 2026-01-07
infra: fix Windows import lib filename missing major version

net147:
issue: https://github.com/thorvg/thorvg/issues/4088


hermet (approved):
@net147 thanks!

PR #4103, opened 2026-01-09
lottie: round polystar point count to nearest integer

### LINKED ISSUE #3838: lottie: point counts on polystars are not rounded to the nearest integer
ThorVG 1.0.0-pre28 does not round the number of points in a polystar layer. This produces uneven rendering which is out of conformance with [the Lottie spec](https://lottie.github.io/lottie-spec/1.0.1/single-page/#specs-shapes-polystar).

<p>

To reproduce the bug, load this Lottie file into an online viewer which uses a recent version of ThorVG, such as [this viewer](https://lottiefiles.github.io/dotlottie-web/) or [this one](https://www.thorvg.org/viewer).

````json
{"ddd":0,"fr":60,"h":100,"ip":0,"layers":[{"ddd":0,"ind":3,"ty":4,"nm":"","ln":"","parent":2,"sr":1,"ks":{"a":{"a":0,"k":[0,0]},"o":{"a":0,"k":100},"p":{"a":0,"k":[0,0]},"r":{"a":0,"k":0},"s":{"a":0,"k":[100,100]},"sk":{"a":0,"k":0},"sa":{"a":0,"k":0}},"ao":0,"ip":0,"op":61,"st":0,"bm":0,"shapes":[{"ty":"sr","hd":false,"p":{"a":0,"k":[50,50]},"ir":{"a":0,"k":19.1},"is":{"a":0,"k":0},"or":{"a":0,"k":50},"os":{"a":0,"k":0},"r":{"a":0,"k":0},"pt":{"a":1,"k":[{"t":0,"s":[5],"i":{"x":1,"y":1},"o":{"x":0,"y":0}},{"t":60,"s":[9],"h":1}]},"sy":1},{"ty":"fl","hd":false,"bm":0,"c":{"a":0,"k":[0.847,0.714,0.376]},"r":1,"o":{"a":0,"k":100}}]},{"ddd":0,"ind":2,"ty":3,"nm":"","ln":"","parent":1,"sr":1,"ks":{"a":{"a":0,"k":[0,0]},"o":{"a":0,"k":100},"p":{"a":0,"k":[0,0]},"r":{"a":0,"k":0},"s":{"a":0,"k":[100,100]},"sk":{"a":0,"k":0},"sa":{"a":0,"k":0}},"ao":0,"ip":0,"op":61,"st":0,"bm":0},{"ddd":0,"ind":1,"ty":3,"nm":"","ln":"","parent":0,"sr":1,"ks":{"a":{"a":0,"k":[0,0]},"o":{"a":0,"k":100},"p":{"a":0,"k":[0,0]},"r":{"a":0,"k":0},"s":{"a":0,"k":[100,100]},"sk":{"a":0,"k":0},"sa":{"a":0,"k":0}},"ao":0,"ip":0,"op":61,"st":0,"bm":0},{"ddd":0,"ind":4,"ty":4,"nm":"","ln":"","parent":0,"sr":1,"ks":{"a":{"a":0,"k":[0,0]},"o":{"a":0,"k":100},"p":{"a":0,"k":[0,0]},"r":{"a":0,"k":0},"s":{"a":0,"k":[100,100]},"sk":{"a":0,"k":0},"sa":{"a":0,"k":0}},"ao":0,"ip":0,"op":61,"st":0,"bm":0,"shapes":[{"ty":"rc","hd":false,"p":{"a":0,"k":[50,50]},"r":{"a":0,"k":0},"s":{"a":0,"k":[100,100]}},{"ty":"fl","hd":false,"bm":0,"c":{"a":0,"k":[1,1,1]},"r":1,"o":{"a":0,"k":100}}]},{"ddd":0,"ind":0,"ty":3,"nm":"","ln":"","sr":1,"ks":{"a":{"a":0,"k":[0,0]},"o":{"a":0,"k":100},"p":{"a":0,"k":[0,0]},"r":{"a":0,"k":0},"s":{"a":0,"k":[100,100]},"sk":{"a":0,"k":0},"sa":{"a":0,"k":0}},"ao":0,"ip":0,"op":61,"st":0,"bm":0}],"meta":{},"nm":"","op":60,"v":"5.7.4","w":100}
````

https://github.com/thorvg/thorvg/issues/3732 is probably related.

### PR DESCRIPTION
**Pull request by reallynattu**
## Summary

Fixes polystar/polygon point count handling to comply with the Lottie specification.

## Problem

According to the [Lottie spec](https://lottie.github.io/lottie-spec/1.0.1/single-page/#specs-shapes-polystar), the point count (`pt`) property for polystar shapes **MUST be rounded to the nearest integer**.

Previously:
- `updateStar()` used the raw float value without any rounding
- `updatePolygon()` used `floor()` instead of `round()`

This caused uneven/incorrect rendering when point counts were animated with non-integer values.

## Solution

Changed both functions to use `roundf()` for the point count, ensuring spec compliance.

## Before/After

The test case from #3838 (animating point count from 5 to 9) should now render correctly with proper integer point counts at each frame.

Closes #3838
Related to #3732

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -520,7 +520,7 @@ void LottieBuilder::updateStar(LottiePolyStar* star, float frameNo, Matrix* tran
 {
     static constexpr auto POLYSTAR_MAGIC_NUMBER = 0.47829f / 0.28f;
 
-    auto ptsCnt = star->ptsCnt(frameNo, tween, exps);
+    auto ptsCnt = roundf(star->ptsCnt(frameNo, tween, exps));
     auto innerRadius = star->innerRadius(frameNo, tween, exps);
     auto outerRadius = star->outerRadius(frameNo, tween, exps);
     auto innerRoundness = star->innerRoundness(frameNo, tween, exps) * 0.01f;
@@ -635,7 +635,7 @@ void LottieBuilder::updatePolygon(LottieGroup* parent, LottiePolyStar* star, flo
 {
     static constexpr auto POLYGON_MAGIC_NUMBER = 0.25f;
 
-    auto ptsCnt = size_t(floor(star->ptsCnt(frameNo, tween, exps)));
+    auto ptsCnt = static_cast<size_t>(lroundf(star->ptsCnt(frameNo, tween, exps)));
     auto radius = star->outerRadius(frameNo, tween, exps);
     auto outerRoundness = star->outerRoundness(frameNo, tween, exps) * 0.01f;
 

### CODE REVIEW

**Comment by reallynattu** [comment]
## Visual Verification

Tested with the sample animation from the issue (point count animating from 5 → 9).

### Frame 10 Comparison (6-point star moment)

**Before (unfixed):** Points are uneven — partial point rendering causes one point to extend much further than others.

**After (fixed):** Clean 6-pointed star with all points symmetric and evenly distributed.

Build tested on macOS with `meson setup build -Dloaders="lottie,png" -Dsavers="gif" -Dtools="lottie2gif" -Dbuildtype=release`

The rounding fix produces the expected spec-compliant output.

File: src/loaders/lottie/tvgLottieBuilder.cpp [hunk 1]
```diff
@@ -510,7 +510,8 @@ void LottieBuilder::updateStar(LottiePolyStar* star, float frameNo, Matrix* tran
 {
     static constexpr auto POLYSTAR_MAGIC_NUMBER = 0.47829f / 0.28f;
 
-    auto ptsCnt = star->ptsCnt(frameNo, tween, exps);
+    //Lottie spec: point count MUST be rounded to the nearest integer
```
tinyjin: Please remove comment (Don't need to point corrected history here)

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -625,7 +626,8 @@ void LottieBuilder::updatePolygon(LottieGroup* parent, LottiePolyStar* star, flo
 {
     static constexpr auto POLYGON_MAGIC_NUMBER = 0.25f;
 
-    auto ptsCnt = size_t(floor(star->ptsCnt(frameNo, tween, exps)));
+    //Lottie spec: point count MUST be rounded to the nearest integer
```
tinyjin: --

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -625,7 +626,8 @@ void LottieBuilder::updatePolygon(LottieGroup* parent, LottiePolyStar* star, flo
...
+    auto ptsCnt = size_t(roundf(star->ptsCnt(frameNo, tween, exps)));
```
tinyjin: Please check this:
`size_t(roundf())` can be replaced with `lroundf`. `lroundf` with `static_cast` looks much optimal

tinyjin:
Regression check:
_(Including existing `"ty": "sr"` resources, looks no side effects)_

CleanShot 2026-01-10 at 11 31 00@2x

hermet (changes requested):
@tinyjin The logic needs to be carefully revised; the following code will become redundant after the fix. 
```
    auto anglePerPoint = (2.0f * MATH_PI / ptsCnt);
    auto halfAnglePerPoint = anglePerPoint * 0.5f;
    auto partialPointAmount = ptsCnt - floorf(ptsCnt);
    auto longSegment = false;
    auto numPoints = size_t(ceilf(ptsCnt) * 2);
```

Also, this will break the poly-animation interpolation. We need to double-confirm whether AE supports interpolation among the point count.

before:

https://github.com/user-attachments/assets/69108670-3994-4314-af0b-a9a431dbba1d

after:

https://github.com/user-attachments/assets/b743a488-615d-4e20-88e9-bd7b0d01618a

[polystar_anim.json](https://github.com/user-attachments/files/24558979/polystar_anim.json)

reallynattu:
Thanks for the detailed review @hermet!

Looking at the [Lottie spec polystar algorithm](https://lottie.github.io/lottie-spec/specs/shapes/#polystar), it explicitly converts point count to integer:

```python
points: int = int(round(pt))
```

This means the spec requires snapping to integer values rather than smooth interpolation. The current ThorVG partial-point morphing (while visually smooth) is non-compliant with the spec.

**Re: redundant code** - You're right that after this fix, the following becomes dead code:
```cpp
auto partialPointAmount = ptsCnt - floorf(ptsCnt);
```
And the `longSegment` logic tied to partial points. Should I remove that in this PR as well?

**Re: AE behavior** - Based on the spec, After Effects should also snap point counts to integers. The smooth morphing in the current implementation is a ThorVG-specific enhancement that doesn't match expected output.

reallynattu:
After reviewing the before/after comparison more carefully, I agree the smooth interpolation in ThorVG is actually the better behavior visually. The jerky snapping in my fix is a downgrade in user experience.

While the Lottie spec technically calls for integer rounding, ThorVG's smooth partial-point morphing is a valuable enhancement that produces superior visual results.

Withdrawing this PR. Thanks for the thorough review @hermet and @tinyjin!

hermet:
@reallynattu Thank you. I’ve reopened this PR because I believe it’s still valid at this point. Primarily, it’s appropriate to align with the Lottie spec (I’d like to double-confirm whether AE behaves the same way.) If both are aligned and the code in this PR is properly refined, we can push this.

Regarding interpolation, we can discuss it with LAC to decide whether to revise the feature or support it conditionally via an additional polystar option.

You can freely leave this to us. Thanks!

reallynattu:
Thanks for reopening @hermet!

**Additional finding:** I checked lottie-web (the reference Bodymovin player) and it also snaps to integers:

```javascript
// lottie-web/player/js/utils/shapes/ShapeProperty.js

// For stars:
var numPts = Math.floor(this.pt.v) * 2;

// For polygons:  
var numPts = Math.floor(this.pt.v);
```

So both the **Lottie spec** (`int(round(pt))`) and **lottie-web** (`Math.floor()`) use integer snapping — neither does smooth interpolation.

This suggests ThorVG's current smooth morphing is a unique enhancement beyond spec/reference behavior. Happy to refine the PR code if you'd like to proceed with spec alignment, or leave it to you as you mentioned. 👍

tinyjin:
@reallynattu @hermet 
Hello, yes this PR addresses similarity with lottie-web & lottie spec. I had also confirmed that lottie-web doesn't interpolate polystar point counts during animation. (So this PR, which removes polystar interpolation, produces rendering output closer to lottie-web)

The [Lottie spec](https://lottie.github.io/lottie-spec/1.0.1/single-page/#specs-shapes-polystar) only covers polystar rendering at a single point in time and doesn't address detailed animation interpolation logic.

However, Adobe After Effects actually does interpolate polystar points, as you can see below:

CleanShot 2026-01-12 at 17 38                                                                           
  54                      
                                                                                                            
  CleanShot 2026-01-13 at 12 37                                                                           
  47   

**This matches what ThorVG previously did.**

Thus, in my opinion, we have two valid approaches:
  1. Follow the Lottie spec approach (this PR) - Priority on official spec & compatibility with lottie-web(spec)
  2. Match AEP behavior (original ThorVG logic) - Priority on designer's expectations and smoother animations

hermet:
\+ @kudanai

kudanai:
Let's raise it on the spec repo for discussion..I'd suggest keeping it within spec as @tinyjin suggested for now.

hermet:
ThorVG complies with After Effects results, making it suitable for dynamic animations. No plan to fix it.

PR #4104, opened 2026-01-09
lottie: support direction for Path (d property)

### LINKED ISSUE #4072: lottie: support direction for Path
Lottie has a direction option(`d`, clockwise) for shape elements that enables sequential drawing using Trim Path.
ThorVG (many other runtimes as well) currently lacks this support only for the `Path` element but other normal shapes work well.

Image
https://lottie.github.io/lottie-spec/latest/specs/shapes/#path

reallynattu:
## Summary
Add path reversal for counter-clockwise direction (d=3) on Path shapes.

## Problem
The parser already correctly read the `d` property for Path shapes and set `clockwise = false` when `d == 3`, but the builder ignored this flag. This meant Path shapes didn't support direction changes for trim path animations.

## Solution
Added a `_reverse()` helper function that reverses the path commands and points, then called it in `updatePath()` when `!path->clockwise`.

The reversal algorithm:
- For each subpath (MoveTo to Close/next MoveTo)
- Start from the last endpoint as the new MoveTo point  
- Walk backwards through segments, adjusting control points for CubicTo

## Visual Verification

Test animation with two identical square paths and trim path effect:
- **Blue** (left): `d=1` (clockwise direction)
- **Red** (right): `d=3` (counter-clockwise direction)

| Trim @ 25% | Trim @ 50% |
|------------|------------|
| trim25 | trim75 |

**Notice how the paths trace in opposite directions:**
- Blue traces: top → right → bottom → left (clockwise)
- Red traces: top → left → bottom → right (counter-clockwise)

Full test animation GIF: [path_direction_test.gif](https://github.com/reallynattu/thorvg/blob/fix/path-direction/test/resources/path_direction_test.gif?raw=true)

This aligns Path behavior with other shapes (Rect, Ellipse, PolyStar) which already support the direction property.

Closes #4072


reallynattu:
## Visual Verification (Updated)

Removed test resource files from commit. Visual evidence for review:

**Test animation with two identical square paths and trim path effect:**
- **Blue** (left): `d=1` (clockwise direction)
- **Red** (right): `d=3` (counter-clockwise direction)

Notice how the paths trace in opposite directions:
- Blue traces: top → right → bottom → left (clockwise)
- Red traces: top → left → bottom → right (counter-clockwise)

This aligns Path behavior with other shapes (Rect, Ellipse, PolyStar) which already support the direction property.

reallynattu:
## Animation Demo

path_direction_test

hermet:
We'd rather apply the path reversal in parsing so we can avoid reversing the path every frame.

PR #4105, opened 2026-01-09
lottie: fix TextDocument expression support

### LINKED ISSUE #3896: lottie: inconsistent rendering

reallynattu:
This patch fixes two issues preventing TextDocument expressions from working:

1. **Add TextDoc case in `_buildValue()`** - handles text property value lookups during expression evaluation (was returning 'Non supported type' error)

2. **Wrap effect property values** - AE expressions expect `effect('name')('property').value` to access values. Without the `.value` wrapper, expressions like `effect('Slider Control')('Slider').value.toFixed()` return NaN.

## Testing

Tested with the Lottie from issue #3896 - the text now renders correctly with dynamic values from the slider expression.

closes #3896

### DIFF
--- a/src/loaders/lottie/tvgLottieExpressions.cpp
+++ b/src/loaders/lottie/tvgLottieExpressions.cpp
@@ -205,6 +205,11 @@ static jerry_value_t _buildValue(float frameNo, LottieProperty* property)
         }
         case LottieProperty::Type::Color: return _color((*static_cast<LottieColor*>(property))(frameNo));
         case LottieProperty::Type::Opacity: return jerry_number((*static_cast<LottieOpacity*>(property))(frameNo));
+        case LottieProperty::Type::TextDoc:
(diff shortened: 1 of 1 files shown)


reallynattu:
## Visual Comparison

**Before (main):** Expression fails with 'Non supported type for value? = 9'

**After (this PR):** Expression evaluates correctly - text shows '1x' → '2x'
pr4105_before
pr4105_after

hermet (changes requested):
At this point, reviewing immature code generated by AI is a waste of time. This PR breaks a lot.

hermet:
more considerable fix: https://github.com/thorvg/thorvg/pull/3980

PR #4108, opened 2026-01-11
wg_engine: skip stencil pass for simple shapes (triangles/quads)

reallynattu:
## Summary

Skip the expensive stencil rendering pass for simple shapes that are guaranteed or virtually guaranteed to be convex.

## Changes

For shapes with ≤4 tessellated vertices, force the convex rendering path:
- **Triangles (3 vertices)**: Mathematically always convex
- **Quads (4 vertices)**: Convex in virtually all practical cases (self-intersecting quads are extremely rare in Lottie content)

## Performance Impact

The convex rendering path skips the stencil buffer fill pass, reducing draw calls for each simple shape. This benefits Lottie animations that contain many rectangles, triangles, and simple polygons, which are very common.

## Risk Assessment

- **Triangles**: Zero risk - triangles are always convex by definition
- **Quads**: Minimal risk - self-intersecting (bowtie) quads would render incorrectly, but these essentially never occur in real Lottie content

## Before/After

Before: All shapes run convexity detection → some shapes may incorrectly use stencil path due to numerical precision
After: Simple shapes (≤4 vertices) → guaranteed convex path → no stencil overhead

### DIFF
--- a/src/renderer/wg_engine/tvgWgRenderData.cpp
+++ b/src/renderer/wg_engine/tvgWgRenderData.cpp
@@ -198,7 +198,10 @@ void WgRenderDataShape::updateMeshes(const RenderShape &rshape, RenderUpdateFlag
         } else {
             WgBWTessellator bwTess{&meshShape};
             bwTess.tessellate(optPath, matrix);
-            convex = bwTess.convex;
+            // Triangles (3 vertices) are mathematically always convex.
+            // Quads (4 vertices) are convex in virtually all practical cases.
+            // Force convex path to skip unnecessary stencil pass.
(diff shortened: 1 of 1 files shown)


wenjieshen:
Hi, thanks for the PR. I’m going to close it because we already compute convexity inside` tessellate()` (`updateConvexity()`), and overriding it with `vbuffer.count <= 4` can incorrectly mark concave shapes as convex and skip stencil, which risks incorrect rendering.

For example, Take this closed path (points in order):
```
(0, 0)
(2, 0)
(0.5, 0.5)
(0, 2)
Close to A
```
This quad is concave. Intuitively, the third point “dents inward”, and the tessellation still produce 4 vertices.

PR #4110, opened 2026-01-12
lottie/expressions: Fixed wiggle function frequency accuracy

tinyjin:
Fixed the wiggle() function to correctly apply frequency and added smooth interpolation for period transitions.

The previous implementation accumulated random values from all previous periods, causing excessive shaking.

This implementation:
- Generates deterministic per period using seed-based random values
- Interpolates between current and next period for smooth transitions
- Separates X/Y seed scales for independent axis values

This matches After Effects' wiggle behavior with accurate frequency control.

## Results

Test file:  [01. wiggle.json](https://github.com/user-attachments/files/24559695/01.wiggle.json)

| Current | Expectation(AE) | Patched |
|-----------|-----------|-----------|
| CleanShot 2026-01-12 at 17 09 52 | CleanShot 2026-01-12 at 17 10 19 | CleanShot 2026-01-12 at 17 12 02 |

### Compatibility with existing players

This patch aims to align with Adobe After Effects’ wiggle behavior. Since other players (e.g., lottie-web) show discrepancies in their wiggle expressions. and ThorVG has now been corrected, animations that rely on wiggle may look different between ThorVG and other players. (_ThorVG wiggle adjusted to closely match AE_)

Example file: [wiggle.json](https://github.com/user-attachments/files/24559753/wiggle.json)

| ThorVG (Current) | lottie-web | ThorVG (Patched) | Expectation(AE) |
|-------------------|----------|-------------------|----------|
| CleanShot 2026-01-12 at 17 20 07 | CleanShot 2026-01-12 at 17 19 44 | CleanShot 2026-01-12 at 17 19 15 | CleanShot 2026-01-12 at 17 20 59 |

### Added `_seededRand` function

Previously, wiggle() was over-shaking due to different random values being generated for each period.

I implemented a custom seed-based random generator (_seededRand) using [LCG algorithm](https://en.wikipedia.org/wiki/Linear_congruential_generator) (glibc style). This avoids global state mutation from srand()/rand() and minimizes overhead compared to the standard <random> library.

```cpp
// Deterministic random generator using glibc's LCG algorithm
static inline float _seededRand(int seed)
{
    seed = (seed * 1103515245 + 12345) & 0x7fffffff;
    return (float)seed / 2147483647.0f;
}
```

### DIFF
--- a/src/loaders/lottie/tvgLottieExpressions.cpp
+++ b/src/loaders/lottie/tvgLottieExpressions.cpp
@@ -899,7 +899,6 @@ static jerry_value_t _speedAtTime(const jerry_call_info_t* info, const jerry_val
 }
 
 
-
 static jerry_value_t _wiggle(const jerry_call_info_t* info, const jerry_value_t args[], const jerry_length_t argsCnt)
 {
     auto data = static_cast<ExpContent*>(jerry_object_get_native_ptr(info->function, &freeCb));
@@ -918,15 +917,40 @@ static jerry_value_t _wiggle(const jerry_call_info_t* info, const jerry_value_t 
         result = (*static_cast<LottieScalar*>(property))(data->frameNo);
     }
 
+    auto perlin1D = [](float x, int seed) {
+        auto x0 = (int)floorf(x);
+        auto x1 = x0 + 1;
+        auto fx = x - (float)x0;
+
+        // Apply quintic fade curve for smooth interpolation
+        // Quintic fade curve for smooth interpolation (6t^5 - 15t^4 + 10t^3)
+        // Used in Perlin noise for smoother transitions than linear interpolation
+        auto u = fx * fx * fx * (fx * (fx * 6.0f - 15.0f) + 10.0f);
+
+        // Simplified gradient function for 1D Perlin noise
+        // Deterministic random generator using glibc's LCG algorithm
+        auto gradient1D = [](int seed) {
+            seed = (seed * 1103515245 + 12345) & 0x7fffffff;
+            return ((float)seed / 2147483647.0f) < 0.5f ? -1.0f : 1.0f;
+        };
+
+        // Calculate dot products (in 1D, this is just multiplication with distance)
+        auto d0 = gradient1D(x0 * 100000 + seed) * fx;
+        auto d1 = gradient1D(x1 * 100000 + seed) * (fx - 1.0f);
+
+        // Interpolate between the two gradient influences(3)
+        return tvg::lerp(d0, d1, u) * 3.0f;
+    };
+
     for (int o = 0; o < octaves; ++o) {
-        auto repeat = (int)ceil(time * freq);
-        for (int i = 0; i < repeat; ++i) {
-            result.x += (_rand() * 2.0f - 1.0f) * amp;
-            result.y += (_rand() * 2.0f - 1.0f) * amp;
-        }
+        auto repeat = time * freq;
+        // Factors (1000000, 2000000) to separate X/Y axes to prevent seed collisions across octaves.
+        result.x += perlin1D(repeat, (1000000 + o)) * amp;
+        result.y += perlin1D(repeat, (2000000 + o)) * amp;
         freq *= 2.0f;
         amp *= ampm;
     }
+
     return _point2d(result);
 }
 


hermet:
@tinyjin please resolve copilot reviews. thanks.

tinyjin:
@hermet Thank you. I've resolved comments

src/loaders/lottie/tvgLottieExpressions.cpp
```diff
@@ -87,6 +87,12 @@ static float _rand()
     return (float)(rand() % 10000001) * 0.0000001f;
 }
 
+static inline float _seededRand(int seed)
```
hermet: `_rand()` or `_random()` is enough

src/loaders/lottie/tvgLottieExpressions.cpp
```diff
@@ -887,12 +893,25 @@ static jerry_value_t _wiggle(const jerry_call_info_t* info, const jerry_value_t
...
     for (int o = 0; o < octaves; ++o) {
-        auto repeat = (int)ceil(time * freq);
-        for (int i = 0; i < repeat; ++i) {
-            result.x += (_rand() * 2.0f - 1.0f) * amp;
-            result.y += (_rand() * 2.0f - 1.0f) * amp;
-        }
+        auto repeat = (int)floor(time * freq);
+        auto frac = time * freq - (float)repeat;
+
+        float startX = _seededRand(repeat * SEED_SCALE_X + o);
```
hermet: float -> auto

hermet:
@tinyjin Please check, Wiggle effect becomes barely visible...

With this PR:

https://github.com/user-attachments/assets/a14ab1ca-4114-446c-8635-b1835808e8f2

Before: 

https://github.com/user-attachments/assets/8e237ac4-6bcf-4f75-94ce-e35eafdc9e06

Lottie-Web:

https://github.com/user-attachments/assets/1bf33947-20bc-4135-8d13-9515995f0ece

tinyjin:
@hermet 
Previous linear interpolation based wiggle could not make closer smoothness to AE behavior.
That implementation was too smooth based on octaves and repeat values.

The previous one alternatively solved this problem by adding nested loop:
```cpp
for (int o = 0; o < octaves; ++o)
        for (int i = 0; i < repeat; ++i) // nested loop
```

However, I found this loop is incorrect, it made more vibration that the first loop could not make, but rather lead to different tremor sensitivity. (Was too shaken, and wasn't smooth like AE)

The value `auto repeat = time * freq` represents current period of the loop(_to have unique amplitude_), and should not push additional loop, I think.

---

I was trying to figure out exact logic in AE, however as we know, it's not public. As alternatively, I searched similar tries in the community. I could conclude that I can assume AE wiggle is based on [Perlin noise](https://en.wikipedia.org/wiki/Perlin_noise) algorithm.

When I tried simple PoC with that, the wiggle movement was quite enhanced, smoother/closer to AE one.

### refers:
- https://stackoverflow.com/questions/39216174/what-is-the-math-function-behind-the-wiggle-expression-in-after-effects
- https://www.reddit.com/r/AfterEffects/comments/1cb88m1/workflow_question_clamp_and_wiggle_expression_for/
- https://expressions200.rssing.com/chan-9202691/all_p672.html

---
This algorithm uses a quintic polynomial–based interpolation, producing smoother noise than linear interpolation or simple predictable jitter.

CleanShot 2026-02-26 at 18 44 59@2x


```cpp
static inline float _fade(float t)
{
    return t * t * t * (t * (t * 6.0f - 15.0f) + 10.0f);
}
```

So each cycle has random gradients, it creates smooth noise, resulting in effects closer to After Effects–style motion.

---

My implementation in the Perlin noise is quite adjusted for use of matching AE-like rendering. A few scale factors are adjusted, and take simplified gradient logic.

It looks more smoother and closer than original impl:

CleanShot 2026-02-26 at 18 25 20

---

For the bomb animation, I believe original ThorVG was too shaken(AE doesn't shake that much), it was actually comes from the additional loop I mentioned (`for (int i = 0; i < repeat; ++i)`).

In new one, it got closer. This is definitely more shaken than my previous fix:

| AE (Expectation) | ThorVG (Patched) |
|--------|-------|
| Before | After |


If it still looks weird, we can adjust by scaling up this factor:
```cpp
// Scale the gradient values to control the influence of the gradient
constexpr float SCALE = 3.0f;
```

In my experemential, the magic number `3.0f` is the most similar one.

---

As repeatedly said, the logic isn't public. I would not say it's 100% correct logic, but would say it can be considered. Please take a look, and verify whether the approach is valid. Thank you.

Issue #4112, opened 2026-01-14
gl: blending issue


hermet:
Test: https://thorvg-playground.surge.sh/showcase/blending

WebGL:
<img width="1035" height="auto" alt="Image" src="https://github.com/user-attachments/assets/8ee99d40-92d7-4892-9e5f-2e8f00b562eb" />

Expected:

<img width="1099" height="auto" alt="Image" src="https://github.com/user-attachments/assets/bd0f71a3-9d32-4c4d-9867-08029785e289" />

hermet:
@wenjieshen If this is a portability issue related to the OpenGL ES specification, it would be helpful to identify which specific features or usages in gl engine side are causing the problem.

wenjieshen:
### Current Progress

Since RenderDoc does not provide useful information, I have decided to leverage NSight to obtain raw debug information using the D3D11 layer. As a result, I can confirm that the direction is pointing to the render buffer because I haven't observed the use of [CopySubresourceRegion](https://learn.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-copysubresourceregion) in the event record.

After several attempts, I have determined that it is not possible to partially copy data from a multisample framebuffer in OpenGL ES. We need to copy the entire buffer instead.

I resolved this issue in [this pull request](https://github.com/thorvg/thorvg/pull/4118). However, we will be able to optimize this limitation once we no longer need the multisample framebuffer.

wenjieshen:
## NOTE
The blit burden remains unchanged. The attached picture shows that we only copy the partial region.
<img width="2048" height="1151" alt="Screenshot 2026-01-19 210616" src="https://github.com/user-attachments/assets/22008924-3962-43e4-ab79-4f7d8d640b52" />

PR #4118, opened 2026-01-18
gl_engine: fix partial blit fail on GLES MSAA buffer

wenjieshen:
## Description
This PR addresses Issue #4112 where OpenGL ES and WebGL rendering contexts fail to blit or resolve multisampled framebuffers when specifying partial regions.

## Key Changes
- **Renderer**: Modified [tvgGlRenderer.cpp](cci:7://file:///c:/Users/warch/OneDrive/repo/workspace/thorvg/src/renderer/gl_engine/tvgGlRenderer.cpp:0:0-0:0) to enforce full buffer blits (resolves) when compiling for non-Desktop GL targets (GLES/WebGL).[^note]
- **Coordinate Correction**: Introduced a `BlendRegion` uniform block in shaders (`BlendRegion`) to correctly map fragment coordinates to destination texture UVs. This is necessary because the full buffer blit changes the relationship between the viewport and the underlying texture coordinates compared to a partial blit.
- The blit burden remains unchanged. The attached picture shows that we only copy the partial region.
Screenshot 2026-01-19 210616

## Motivation
To ensure correct rendering and resolve crashes or artifacts on mobile (GLES) and web (WebGL) platforms caused by invalid partial MSAA blits.

[^note]: I intentionally don’t blit into a smaller FBO because it adds an extra full copy pass (read + write) that we can avoid. In the next pass, sampling from a full-size resolved texture is not inherently more expensive: the GPU only fetches the texels needed for the fragments we actually draw, not the whole texture. Since the blend is typically a single pass and the affected area isn’t known reliably ahead of time, a “small” FBO often ends up near full size anyway, making the extra blit pure overhead (and can introduce extra bandwidth/stalls on mobile).

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderTask.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderTask.cpp
@@ -331,30 +331,24 @@ void GlSceneBlendTask::run()
     GlComposeTask::run();
 
     const auto& vp = getViewport();
+    const auto width = mSrcFbo->getWidth();
+    const auto height = mSrcFbo->getHeight();
+    if (width <= 0 || height <= 0) return;
 
-    // For EMSCRIPTEN: perform an intermediate blit from multisampled FBO to resolve FBO
-    // Then perform a final blit to destination framebuffer.
-    // For other platforms: direct blit from target FBO to destination resolve FBO.
-
-    // Platform-specific source framebuffer binding
-#ifdef __EMSCRIPTEN__
     GL_CHECK(glBindFramebuffer(GL_READ_FRAMEBUFFER, mSrcFbo->getFboId()));
-    GL_CHECK(glBindFramebuffer(GL_DRAW_FRAMEBUFFER, mSrcFbo->getResolveFboId()));
-    GL_CHECK(glViewport(vp.x(), vp.y(), vp.w(), vp.h()));
-    GL_CHECK(glScissor(vp.x(), vp.y(), vp.w(), vp.h()));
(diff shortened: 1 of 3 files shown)


wenjieshen:
Hi @tinyjin, could you please help me confirm if this patch works correctly on WebGL? Thank you.🙏

tinyjin (approved):
Looks good to webgl

Image

src/renderer/gl_engine/tvgGlRenderTask.cpp
```diff
@@ -330,31 +330,26 @@ void GlSceneBlendTask::run()
...
-    GL_CHECK(glViewport(0, 0, mDstCopyFbo->getWidth(), mDstCopyFbo->getHeight()));
-    GL_CHECK(glScissor(0, 0, mDstCopyFbo->getWidth(), mDstCopyFbo->getHeight()));
+    GL_CHECK(glViewport(0, 0, srcVp.w(), srcVp.h()));
+    GL_CHECK(glScissor(0, 0, srcVp.w(), srcVp.h()));
     GL_CHECK(glBlitFramebuffer(vp.min.x, vp.min.y, vp.max.x, vp.max.y, 0, 0, vp.w(), vp.h(), GL_COLOR_BUFFER_BIT, GL_LINEAR));
+#else // TODO: create partial buffer when MSAA is disabled
+    const auto& vp = getViewport();
+    const auto width = mSrcFbo->getWidth();
+    const auto height = mSrcFbo->getHeight();
+    if (width <= 0 || height <= 0) return;
```
hermet: GL version doesn't have this exception. Does GLES only necessary?
Otherwise, Please don't add exception handling every time. It might just make the situation bad to figure out the problem. The problem must be handled earlier, not hiding.

wenjieshen: Thank you for the suggestion. It has been consistent across both GL and GLES.

PR #4121, opened 2026-01-20
gl_engine: Refactored some part of gl internal classes

hermet:
- removed unnecessary getter/setter methods
- removed Hungarian-style naming for consistency


hermet:
Fewer LOC are preferred to improve the quality, readability, and maintainability of the ThorVG codebase. This approach can also be applied to backend engines. Internal classes are not necessary for simple setters and getters.

PR #4122, opened 2026-01-20
refactor: make Bezier segments calculation iterative

wenjieshen:
Description: This PR refactors Bezier::segments() in tvgMath.cpp to address potential stack overflow issues and unbounded execution times.

Changes:

Replaced the recursive divide-and-conquer algorithm with an iterative while loop using Array<Bezier> as a stack.
Introduced a MaxSegments cap ($2^{10}$) to prevent excessive processing on degenerate or extremely complex curves.

### TEST
I did not find any significant regressions while running all the examples.

### DIFF
--- a/src/common/tvgMath.cpp
+++ b/src/common/tvgMath.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "tvgMath.h"
+#include "tvgArray.h"
 
 #define BEZIER_EPSILON 1e-2f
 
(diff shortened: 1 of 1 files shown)


src/common/tvgMath.cpp
```diff
@@ -477,13 +478,27 @@ bool Bezier::flatten() const
...
+    stack.push(*this);
+
+    while (!stack.empty()) {
+        auto current = stack.last();
+        stack.pop();
+        if (current.flatten()) {
+            ++count;
+            continue;
+        }
+        Bezier left, right;
```
hermet: how about move them out of while()?

src/common/tvgMath.cpp
```diff
@@ -477,13 +478,27 @@ bool Bezier::flatten() const
...
+        current.split(left, right);
+        stack.push(left);
+        stack.push(right);
+        if (count + stack.count >= MaxSegments) return MaxSegments;
```
hermet:
```
if (count + stack.count >= MaxSegments) {
    TVGERR("COMMON", "add error message");
    return MaxSegments;
}

src/common/tvgMath.cpp
wenjieshen: Thanks for the suggestion.

Issue #4125, opened 2026-01-21
lottie: Slot apply crashes for text animator color properties (incorrect LottieSolid cast)


hoyangtsai:
## Description

When applying a color slot (`sid` reference) to a text animator's fill color (`fc`) property, the `LottieSlot::apply()` method incorrectly casts the object to `LottieSolid*`, causing undefined behavior/crash. This is because text animator properties are stored in `LottieTextRange` objects, not `LottieSolid`.

## Steps to Reproduce

1. Create a Lottie animation with a text layer
2. Add a text animator with fill color (`fc`) property
3. Add an `sid` reference to the fill color pointing to a color slot
4. Load the animation in a ThorVG-based player
5. The animation fails to render (blank screen, 0 frames)

## Example Lottie JSON

```json
{
  "slots": {
    "textColor": {
      "p": {
        "a": 0,
        "k": [1, 0, 0, 1]
      }
    }
  },
  "layers": [{
    "ty": 5,
    "nm": "Text Layer",
    "t": {
      "d": { "k": [{ "s": { "t": "Hello", "f": "Arial", "s": 72 }, "t": 0 }] },
      "a": [{
        "nm": "Fill Animator",
        "s": { "t": 0, "b": 1, "s": { "a": 0, "k": 0 }, "e": { "a": 0, "k": 100 } },
        "a": {
          "fc": {
            "a": 0,
            "k": [0, 0, 0, 1],
            "sid": "textColor"
          }
        }
      }]
    }
  }]
}
```

## Root Cause Analysis

**Location:** `src/loaders/lottie/tvgLottieModel.cpp` line ~228

```cpp
void LottieSlot::apply(LottieProperty* prop, bool byDefault)
{
    // ...
    ARRAY_FOREACH(pair, pairs) {
        switch (type) {
            case LottieProperty::Type::Color: {
                if (copy) pair->prop = new LottieColor(
                    static_cast<LottieSolid*>(pair->obj)->color);  // <-- BUG
                pair->obj->override(prop, !copy);
                break;
            }
            // ...
        }
    }
}
```

**Problem:**
1. When parsing text animator `fc` with `sid`, `registerSlot()` is called with `obj = LottieTextRange`
2. In `LottieSlot::apply()`, the code assumes ALL color slots target `LottieSolid` objects
3. Casting `LottieTextRange*` to `LottieSolid*` is invalid
4. Accessing `->color` on the miscast pointer causes undefined behavior

**Registration Path:**
[...]

PR #4135, opened 2026-02-03
lottie/slot: Support Text Style slot overriding

tinyjin:
# Overview
Added support for Lottie Slot Overriding for [Text Style](https://lottiefiles.github.io/lottie-docs/text/#text-style) of Text Range, including textColor, strokeColor, strokeWidth and transform related properties(position, scale, rotation, opacity)

related issue : #4125, #4138

# Preview

## Default Slot (color, scale, strokeColor, strokeWidth)

file: [example.json](https://github.com/user-attachments/files/25042845/example.json)

| Original | Overriden |
|---------|-----------|
| Original | Overriden |

## Via API

original file : [original.json](https://github.com/user-attachments/files/25044200/original.json)

CleanShot 2026-02-03 at 21 40 09

### Text Color

```cpp
//slot (textColor override - red)
const char* slotJson = R"({"textColor":{"p":{"a":0,"k":[1,0,0,1]}}})";
auto slotId = slot->gen(slotJson);
if (!tvgexam::verify(slot->apply(slotId))) return false;
```

### Stroke Color

```cpp
//slot (strokeColor override - green)
const char* slotJson = R"({"strokeColor":{"p":{"a":0,"k":[0,1,0,1]}}})";
auto slotId = slot->gen(slotJson);
if (!tvgexam::verify(slot->apply(slotId))) return false;
```

### Stroke Width

```cpp
//slot (strokeWidth override - thicker)
const char* slotJson = R"({"strokeWidth":{"p":{"a":0,"k":8}}})";
auto slotId = slot->gen(slotJson);
if (!tvgexam::verify(slot->apply(slotId))) return false;
```

### Text Scale

```cpp
//slot (textScale override - scaled up)
const char* slotJson = R"({"textScale":{"a":0,"k":[150,150,100]}}})";
auto slotId = slot->gen(slotJson);
if (!tvgexam::verify(slot->apply(slotId))) return false;
```

### Text Position

```cpp
//slot (position override)
const char* slotJson = R"({"textPosition":{"p":{"a":0,"k":[30,30]}}})";
auto slotId = slot->gen(slotJson);
if (!tvgexam::verify(slot->apply(slotId))) return false;
```

### Letter Space

```cpp
//slot (letterSpace override)
const char* slotJson = R"({"letterSpace":{"p":{"a":0,"k":15}}})";
auto slotId = slot->gen(slotJson);
if (!tvgexam::verify(slot->apply(slotId))) return false;
```

### Line Space

```cpp
//slot (lineSpace override)
const char* slotJson = R"({"lineSpace":{"p":{"a":0,"k":10}}})";
auto slotId = slot->gen(slotJson);
if (!tvgexam::verify(slot->apply(slotId))) return false;
```

### Text Rotation

```cpp
//slot (rotation override)
const char* slotJson = R"({"textRotation":{"p":{"a":0,"k":30}}})";
auto slotId = slot->gen(slotJson);
if (!tvgexam::verify(slot->apply(slotId))) return false;
```

### Fill Opacity

```cpp
//slot (fillOpacity override)
const char* slotJson = R"({"fillOpacity":{"p":{"a":0,"k":50}}})";
auto slotId = slot->gen(slotJson);
if (!tvgexam::verify(slot->apply(slotId))) return false;
```

### Stroke Opacity

```cpp
//slot (strokeOpacity override)
const char* slotJson = R"({"strokeOpacity":{"p":{"a":0,"k":30}}})";
auto slotId = slot->gen(slotJson);
if (!tvgexam::verify(slot->apply(slotId))) return false;
```

### Text Opacity

```cpp
//slot (opacity override)
const char* slotJson = R"({"textOpacity":{"p":{"a":0,"k":50}}})";
auto slotId = slot->gen(slotJson);
if (!tvgexam::verify(slot->apply(slotId))) return false;
```

# Lottie Object Model Revision

CleanShot 2026-02-04 at 14 47 24@2x

- `LottieTextRange` inherits `LottieObject` to pair each overriding case per specific Text Range (LottieText could have multiple text range.
- Text Range slot overriding supports TextStyle only for now as needed.
- added `sid` in `LottieProperty` to explicitly target slot property (Since TextStyle has duplicated types, a single branching based on `LottieProperty::Type` is not feasible.)
- refactored another overriding cases, those have `LottieProperty::Type` based condition to make it verify `sid` instead of type check (Transform, LottieSolid)

> We may need to review additional memory usage by `sid` for all properties. (Alternatively, we can use `ix`, but this value isn't always guaranteed)

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -203,57 +203,11 @@ void LottieSlot::reset()
 
 void LottieSlot::apply(LottieProperty* prop, bool byDefault)
 {
-    auto copy = !overridden && !byDefault;
+    auto release = overridden || byDefault;
 
     //apply slot object to all targets
     ARRAY_FOREACH(pair, pairs) {
-        //backup the original properties before overwriting
-        switch (type) {
-            case LottieProperty::Type::Float: {
-                if (copy) pair->prop = new LottieFloat(static_cast<LottieTransform*>(pair->obj)->rotation);
-                pair->obj->override(prop, !copy);
-                break;
-            }
-            case LottieProperty::Type::Scalar: {
-                if (copy) pair->prop = new LottieScalar(static_cast<LottieTransform*>(pair->obj)->scale);
-                pair->obj->override(prop, !copy);
-                break;
-            }
-            case LottieProperty::Type::Vector: {
-                if (copy) pair->prop = new LottieVector(static_cast<LottieTransform*>(pair->obj)->position);
-                pair->obj->override(prop, !copy);
-                break;
-            }
-            case LottieProperty::Type::Color: {
-                if (copy) pair->prop = new LottieColor(static_cast<LottieSolid*>(pair->obj)->color);
-                pair->obj->override(prop, !copy);
-                break;
-            }
-            case LottieProperty::Type::Opacity: {
-                if (copy) {
-                    if (pair->obj->type == LottieObject::Type::Transform) pair->prop = new LottieOpacity(static_cast<LottieTransform*>(pair->obj)->opacity);
-                    else pair->prop = new LottieOpacity(static_cast<LottieSolid*>(pair->obj)->opacity);
-                }
-                pair->obj->override(prop, !copy);
-                break;
-            }
-            case LottieProperty::Type::ColorStop: {
-                if (copy) pair->prop = new LottieColorStop(static_cast<LottieGradient*>(pair->obj)->colorStops);
-                pair->obj->override(prop, !copy);
-                break;
-            }
-            case LottieProperty::Type::TextDoc: {
-                if (copy) pair->prop = new LottieTextDoc(static_cast<LottieText*>(pair->obj)->doc);
-                pair->obj->override(prop, !copy);
-                break;
-            }
-            case LottieProperty::Type::Image: {
-                if (copy) pair->prop = new LottieBitmap(static_cast<LottieImage*>(pair->obj)->bitmap);
-                pair->obj->override(prop, !copy);
-                break;
-            }
-            default: break;
-        }
+        pair->prop = pair->obj->override(prop, release);
     }
 
     if (!byDefault) overridden = true;
--- a/src/loaders/lottie/tvgLottieModel.h
+++ b/src/loaders/lottie/tvgLottieModel.h
@@ -271,16 +271,18 @@ struct LottieObject
         Text,
         Repeater,
         RoundedCorner,
-        OffsetPath
+        OffsetPath,
+        TextRange
     };
 
     virtual ~LottieObject()
     {
     }
 
-    virtual void override(LottieProperty* prop, bool release)
+    virtual LottieProperty* override(LottieProperty* prop, bool release)
     {
         TVGERR("LOTTIE", "Unsupported slot type");
+        return nullptr;
     }
 
     virtual bool mergeable() { return false; }
@@ -315,7 +317,7 @@ struct LottieGlyph
 };
 
 
-struct LottieTextRange
+struct LottieTextRange : LottieObject
 {
     enum Based : uint8_t { Chars = 1, CharsExcludingSpaces, Words, Lines };
     enum Shape : uint8_t { Square = 1, RampUp, RampDown, Triangle, Round, Smooth };
@@ -323,6 +325,7 @@ struct LottieTextRange
 
     LottieTextRange()
     {
+        LottieObject::type = LottieObject::TextRange;
         style.flags.fillColor = 0;
         style.flags.strokeColor = 0;
         style.flags.strokeWidth = 0;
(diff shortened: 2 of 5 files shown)


tinyjin:
@hermet
Hello, note that current implementation used the same approach as we did for Lottie Effects Slot (`sid` added). I will come up with performance comparison soon.

Plus, I’ve implemented full Text Style overriding.
Please let me know if any of these properties are unnecessary, as we can simplify the implementation for only needed.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -464,7 +464,7 @@ void LottieParser::parsePropertyInternal(T& prop)
 }
 
 
-void LottieParser::registerSlot(LottieObject* obj, const char* sid, LottieProperty::Type type)
+void LottieParser::registerSlot(LottieObject* obj, const char* sid, LottieProperty* prop)
```
hermet: LottieProperty& prop

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -271,16 +271,18 @@ struct LottieObject
...
+        OffsetPath,
+        TextRange
     };
 
     virtual ~LottieObject()
     {
     }
 
-    virtual void override(LottieProperty* prop, bool release)
+    virtual LottieProperty* override(LottieProperty* prop, bool release = false)
```
hermet: Does `false init` need? `bool release = false`

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -203,57 +203,11 @@ void LottieSlot::reset()
...
-                break;
-            }
-            case LottieProperty::Type::Image: {
-                if (copy) pair->prop = new LottieBitmap(static_cast<LottieImage*>(pair->obj)->bitmap);
-                pair->obj->override(prop, !copy);
-                break;
-            }
-            default: break;
-        }
+        pair->prop = pair->obj->override(prop, release);
```
hermet, line 210: Assuming the case `pair->prop` is `nullptr`. Is it safe scenario?

src/loaders/lottie/tvgLottieModel.h
```diff
@@ -383,6 +386,57 @@ struct LottieTextRange
             strokeColor.b = tvg::lerp<uint8_t>(strokeColor.b, color.b, factor);
         }
     }
+
+    LottieProperty* override(LottieProperty* prop, bool release = false) override
+    {
+        LottieProperty* backup = nullptr;
+        if (style.fillColor.sid == prop->sid) {
+            if (release) style.fillColor.release();
```
hermet, line 394: Suggest you to this order:

1. Change the value -> 2. Apply
```
style.fillColor.copy(*static_cast<LottieColor*>(prop), false);
if (release) style.fillColor.release();
else backup = new LottieColor(style.fillColor);
```

All other places as well.

hermet (commented):
+minor nit-picks

src/loaders/lottie/tvgLottieModel.h
tinyjin: @hermet
Doesn't the `copy` have to call after release/backup logic?
If we proceed release or backup after copy, then the current property could lose their value by `release`, or the `backup` value refers overidden value not  original one.

src/loaders/lottie/tvgLottieModel.cpp
tinyjin: returned `backup` data from `pair->obj->override(prop, release);` will be used in `reset()` function.

If they return `nullptr`, and `pair->prop` set to that value, then it could cause segmentation fault.
(`sid` mismatch)
```cpp
void LottieSlot::reset()
{
    if (!overridden) return;

    ARRAY_FOREACH(pair, pairs) {
        pair->obj->override(pair->prop, true); // >>> crash
        delete(pair->prop);
        pair->prop = nullptr;
    }
    overridden = false;
}
```

To address this correctly, `reset()` can verify this, I'd add `if (!pair->prop) continue;`

```cpp
ARRAY_FOREACH(pair, pairs) {
    if (!pair->prop) continue; // added
    pair->obj->override(pair->prop, true);
    delete(pair->prop);
    pair->prop = nullptr;
}
```

This prevents issues caused by a Slot overriding with an unregistered sid and resetting.
(I made change on `6593d85d85f2612b54397009b04b6ed6d1c1d458` commit, I will squash them in the end)

hermet: @tinyjin This patch should not introduce any new crashes, so please make sure to prevent crash-related issues within this PR. Thanks. (If not, please ignore this. But compared to the original logic, now pair->prop can be potentially nullptr in this function)

tinyjin: Ah, I was wrong. Even if `sid` mismatch case, we return slotId 0 to prevent this. Slot generation will proceed only when the `sid` is already registered. Thus this case would not happen. (Already prevented)

Sorry for confusion. (Removed that commit)

---
Tested with some edge cases.
1. Generating with JSON includes invalid `sid` -> Return slotId 0
2. Applying with invalid `slotId` -> Return `InvalidArguments`
3. Possible number of combinations such as (apply -> del -> apply again and ETC) -> working correctly

hermet (approved):

Issue #4134, opened 2026-02-02
gpu: Bake translation on CPU to reduce per-draw uniforms and enable more batching


wenjieshen:
## What / Why
Many 2D draws share the same pipeline state and differ only by translation.
Keeping translation in a per-draw transform matrix causes:
- frequent uniform updates
- reduced draw-call merging when only translation differs

## Proposal
For affine transforms, pre-apply translation on the CPU and draw geometry directly in device space.

- Decompose affine transform: `D = A * L + b`
- CPU: bake `b` into generated vertices / tessellated points
- GPU:
  - do not pass per-draw translation
  - optionally pass `invA` (mat2) **only if a shader actually needs it**
  - If a shader requires full local coordinates, skip this optimization for that path

## Pros
- Fewer per-draw uniform updates
- Better draw-call merging when translation is the only difference
- Lower CPU overhead per draw
- Simple, low-risk optimization (affine-only, per-shader opt-in)

## Implementation sketch
1. Detect the affine transform and split into `A` + `b.`
2. Apply `b` to geometry on CPU before upload
3. Use existing shaders; pass `invA` only when required
4. Fallback to the current path for shaders that need local coordinates

It is possible to resolve some device-dependent issues by mitigating the buffer stress.
#4132
#4129

Issue #4132, opened 2026-02-02
webgl: Artifacts are noticeable in blur effects on the playground.


hermet:
<img width="975" height="158" alt="Image" src="https://github.com/user-attachments/assets/dd0ed1ab-bfed-4f78-8307-57eafbf8346d" />

https://github.com/user-attachments/assets/dbf5310d-a0b4-4e38-910e-7a8039cc7a75

hermet:
cc : @wenjieshen

wenjieshen:
The PR #4425 has resolved this issue.
Local Test environment: Ubuntu 24.04.4 LTS / Google Chrome Version 148.0.7778.215

https://github.com/user-attachments/assets/328629a1-1c2e-4630-8a6d-04ccc0d1b476

hermet:
still observable (v1.0.7)

<img width="907" height="673" alt="Image" src="https://github.com/user-attachments/assets/34eb3ad6-94b8-4012-ad68-41119a22c82e" />

\+ @wenjieshen @tinyjin

Issue #4129, opened 2026-01-30
webgl: compatibility issue at dropshadow rendering


hermet:
Failed to draw shadows at our playground:
https://thorvg-playground.vercel.app/showcase/effect-drop-shadow

<img width="2662" height="1574" alt="Image" src="https://github.com/user-attachments/assets/51cc2181-8e55-4728-9475-4c8b8d42fb22" />

Test machine: Ubuntu / Intel Arc Graphics (thorvg.example + both GL / GLES working good)

[about-gpu-2026-01-30T16-09-13-869Z.txt](https://github.com/user-attachments/files/24969501/about-gpu-2026-01-30T16-09-13-869Z.txt)

hermet:
cc: @wenjieshen

Nor-s:
<img width="973" height="828" alt="Image" src="https://github.com/user-attachments/assets/4506ff99-3085-4daf-96a1-77031dfec257" />

```
WebGL warning: drawElementsInstanced: Buffer for uniform block is smaller than UNIFORM_BLOCK_DATA_SIZE. 12
WebGL warning: drawElementsInstanced: Buffer for uniform block is smaller than UNIFORM_BLOCK_DATA_SIZE.
WebGL warning: drawElementsInstanced: Buffer for uniform block is smaller than UNIFORM_BLOCK_DATA_SIZE.
WebGL warning: drawElementsInstanced: Buffer for uniform block is smaller than UNIFORM_BLOCK_DATA_SIZE.
WebGL warning: drawElementsInstanced: Buffer for uniform block is smaller than UNIFORM_BLOCK_DATA_SIZE. 17
After reporting 32, no further warnings will be reported for this WebGL context.
```

This can be reproduced on Mac and Firefox.

wenjieshen:
[...]

PR #4137, opened 2026-02-03
gl_engine: Apply model transform on CPU to optimize rendering pipeline

wenjieshen:
## Description
This PR moves the model matrix application from the Vertex Shader to the CPU geometry preparation stage. By pre-transforming vertices into World Space on the CPU, we unlock several critical performance improvements and simplify the rendering pipeline.

## Key Improvements
- **FPS Improvement**: The example `Lottie` gets **10.646%** fps improvement in the Apple M1 macOS 26, and 3.84% fps improvement in Windows 11 RTX 5080 Laptop.
- **UBO pressure mitigation**: Based on the logs from #4132 and #4129, the device supports the feature extensions, so the regression is likely due to UBO overflow.
- **Massive Reduction in GL State Changes**: By removing the need to update the `uModel` uniform for every single draw call, we significantly reduce driver overhead and API call frequency. This is crucial for scenes with many shapes.
- **Geometry Caching logic**: Introduces cached outWorld geometry. If a shape's transform hasn't changed, we reuse the pre-calculated world-space vertices, saving redundant computations on every frame.
- **Enables Future Batching**: This is a foundational step for draw call merging. Since geometries are now normalized to World Space on the CPU, multiple shapes can be easily combined into a single vertex buffer and drawn together, which was previously difficult with per-draw-call model uniforms.
- **Shader Simplification**: Removed the Matrix uniform block and complex `glUniformMatrix4fv` usage, resulting in leaner and faster shader execution.
## Technical Details
`GlRenderPass` now handles the View Matrix (Screen -> NDC) globally.
`GlGeometry` refactored to optimize and transform paths into `outWorld` buffers on the CPU.
Shaders updated to use a shared `uViewMatrix` instead of per-object model matrices.

Test: All examples

Related Issue:
#4134
#4132
#4129


src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -25,6 +25,44 @@
 #include "tvgGlRenderTask.h"
 #include "tvgGlTessellator.h"
 
+static RenderRegion _transformBounds(const RenderRegion& bounds, const Matrix& matrix)
+{
+    if (bounds.invalid() || tvg::identity(&matrix)) return bounds;
```
hermet: let's remove identity() check.  Let's priortize on the general case (tranfsormed).

src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -202,10 +255,21 @@ bool GlGeometry::tesselateStroke(const RenderShape& rshape)
...
-        stroker.run(rshape, optPath, matrix);
-        bounds = stroker.bounds();
+    auto strokeWidthWorld = strokeWidth * scaling(matrix);
+    if (!std::isfinite(strokeWidthWorld)) strokeWidthWorld = strokeWidth;
+
+    if (!tvg::zero(strokeWidthWorld)) {
+        Stroker stroker(&stroke, strokeWidthWorld, rshape.strokeCap(), rshape.strokeJoin(), rshape.strokeMiterlimit());
+        RenderPath dashedPath;
+        if (rshape.strokeDash(dashedPath)) {
+            RenderPath dashedPathWorld;
```
hermet: Does it need to copy to `dashedPathWorld`? Can't use directly `dashedPath`?

src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -25,6 +25,44 @@
...
+
+    auto transformed = RenderRegion{{int32_t(floor(left)), int32_t(floor(top))}, {int32_t(ceil(right)), int32_t(ceil(bottom))}};
+    if (transformed.valid()) return transformed;
+    return bounds;
+}
+
+
+static void _transformPath(const RenderPath& src, const Matrix& matrix, RenderPath& dst)
+{
+    dst.cmds.clear();
```
hermet: Please check 
```
dst.cmds.clear();
dst.pts.clear();
dts.cmds.push(src.cmds);
dts.pts.push(src.pts);
```

src/renderer/tvgRender.cpp
```diff
@@ -103,17 +103,34 @@ bool RenderPath::bounds(const Matrix* m, BBox& box)
...
 
 
-void RenderPath::optimize(RenderPath& out, const Matrix& matrix) const
+void RenderPath::optimize(RenderPath& out, const Matrix& matrix, RenderPath* outWorld) const
 {
 #if defined(THORVG_GL_RASTER_SUPPORT) || defined(THORVG_WG_RASTER_SUPPORT)
     static constexpr auto PX_TOLERANCE = 0.25f;
 
-    if (empty()) return;
+    if (outWorld == &out) outWorld = nullptr;
```
hermet: It seems unnecessary exceptions. (We should not expect those weird usage)

src/renderer/gl_engine/tvgGlCommon.h
```diff
@@ -34,25 +34,25 @@ constexpr float MIN_GL_STROKE_ALPHA = 0.25f;
...
-    matOut[3] = 0.0f;
-    matOut[4] = mat3.e12;
-    matOut[5] = mat3.e22;
-    matOut[6] = mat3.e32;
-    matOut[7] = 0.0f;
-    matOut[8] = mat3.e13;
-    matOut[9] = mat3.e23;
-    matOut[10] = mat3.e33;
-    matOut[11] = 0.0f;
+    matOut[0] = mat3.e11; matOut[4] = mat3.e12; matOut[8] = mat3.e13;
```
hermet, line 49: Seems you can use getMatrix3() call here instead of duplicate code.

src/renderer/gl_engine/tvgGlCommon.h
```diff
@@ -90,9 +90,13 @@ struct GlGeometry
     GlGeometryBuffer fill, stroke;
     Matrix matrix = {};
     RenderRegion viewport = {};
-    RenderRegion bounds = {};
+    RenderRegion fillBounds = {};
+    RenderRegion strokeBounds = {};
     FillRule fillRule = FillRule::NonZero;
     RenderPath optPath;  //optimal path
+    RenderPath optPathWorld;  //optimized path transformed by matrix
```
hermet: matrix -> world matrix

hermet: "transformed by matrix" sounds very ambiguous.

Question: `optPath` doesn't apply the world transform and `optPathWorld` applies the world transform? The only difference is that?

hermet:
Lottie: 4.23% fps+
Stress: 5.27% fps+

src/renderer/gl_engine/tvgGlCommon.h
wenjieshen: Thank you. My first implementation is calling `getMatrix3()` here, but it should be `matOut[3] = 0.0f` in this funciton instead of `matOut[3] = mat3.e12;`. As a result, the code cannot be reused.

src/renderer/tvgRender.cpp
wenjieshen: Thank you. I have removed.

src/renderer/gl_engine/tvgGlGeometry.cpp
wenjieshen: Thank you. I have removed it.

src/renderer/gl_engine/tvgGlGeometry.cpp
wenjieshen: Thank you for the comment. I have changed the API signature to use it directly.

src/renderer/gl_engine/tvgGlCommon.h
wenjieshen: Thanks. I have changed the API signature to maintain only one optPath.

src/renderer/gl_engine/tvgGlGeometry.cpp
wenjieshen: I have removed this function. Thank you.

src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -256,7 +291,8 @@ void GlGeometry::tesselateImage(const RenderSurface* image)
     fill.index.push(1);
     fill.index.push(3);
 
-    bounds = {{0, 0}, {int32_t(image->w), int32_t(image->h)}};
+    auto localBounds = RenderRegion{{0, 0}, {int32_t(image->w), int32_t(image->h)}};
+    fillBounds = _transformBounds(localBounds, matrix);
```
hermet: one liner.
`fillBounds = _transformBounds(_transformBounds(localBounds, matrix), matrix); `

src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -202,10 +238,19 @@ bool GlGeometry::tesselateStroke(const RenderShape& rshape)
...
-        Stroker stroker(&stroke, strokeWidth, rshape.strokeCap(), rshape.strokeJoin());
-        stroker.run(rshape, optPath, matrix);
-        bounds = stroker.bounds();
+    auto strokeWidthWorld = strokeWidth * scaling(matrix);
+    if (!std::isfinite(strokeWidthWorld)) strokeWidthWorld = strokeWidth;
+
+    if (!tvg::zero(strokeWidthWorld)) {
+        Stroker stroker(&stroke, strokeWidthWorld, rshape.strokeCap(), rshape.strokeJoin(), rshape.strokeMiterlimit());
+        RenderPath dashedPathWorld;
+        if (rshape.strokeDash(dashedPathWorld, &matrix)) {
```
hermet: two liner.
```
if ( ) ..
else ...
```

src/renderer/tvgRender.cpp
```diff
@@ -103,12 +103,16 @@ bool RenderPath::bounds(const Matrix* m, BBox& box)
...
 
-void RenderPath::optimize(RenderPath& out, const Matrix& matrix) const
+void RenderPath::optimize(RenderPath& out, const Matrix& matrix, bool outputTransformed) const
 {
 #if defined(THORVG_GL_RASTER_SUPPORT) || defined(THORVG_WG_RASTER_SUPPORT)
     static constexpr auto PX_TOLERANCE = 0.25f;
 
-    if (empty()) return;
+    if (empty()) {
+        out.cmds.clear();
```
hermet: -> `out.clear();`

src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -25,6 +25,26 @@
...
+    auto rt = Point{float(bounds.max.x), float(bounds.min.y)} * matrix;
+    auto rb = Point{float(bounds.max.x), float(bounds.max.y)} * matrix;
+
+    auto left = min(min(lt.x, lb.x), min(rt.x, rb.x));
+    auto top = min(min(lt.y, lb.y), min(rt.y, rb.y));
+    auto right = max(max(lt.x, lb.x), max(rt.x, rb.x));
+    auto bottom = max(max(lt.y, lb.y), max(rt.y, rb.y));
+
+    auto transformed = RenderRegion{{int32_t(floor(left)), int32_t(floor(top))}, {int32_t(ceil(right)), int32_t(ceil(bottom))}};
+    if (transformed.valid()) return transformed;
```
hermet: Hmm... What cases could be failed at validation?... bounds.invalid() is already in 30.

src/renderer/gl_engine/tvgGlCommon.h
```diff
@@ -90,9 +90,12 @@ struct GlGeometry
     GlGeometryBuffer fill, stroke;
     Matrix matrix = {};
     RenderRegion viewport = {};
-    RenderRegion bounds = {};
+    RenderRegion fillBounds = {};
+    RenderRegion strokeBounds = {};
     FillRule fillRule = FillRule::NonZero;
     RenderPath optPath;  //optimal path
+    bool fillWorld = false;
```
hermet: Please switch order for memory packing.
```
float
bool
```

src/renderer/tvgRender.cpp
```diff
@@ -103,12 +103,16 @@ bool RenderPath::bounds(const Matrix* m, BBox& box)
 }
 
 
-void RenderPath::optimize(RenderPath& out, const Matrix& matrix) const
+void RenderPath::optimize(RenderPath& out, const Matrix& matrix, bool outputTransformed) const
```
hermet: Briefly checked this function. It seems like you are killing the wgpu case. It would be better to separate the function for the GL version, and later—when wgpu and GL can be unified—we can refactor it.

src/renderer/gl_engine/tvgGlGeometry.cpp
wenjieshen: Thank you. I squashed them into one line.

src/renderer/gl_engine/tvgGlGeometry.cpp
wenjieshen: Thank you. I have eliminated the braces.

src/renderer/tvgRender.cpp
wenjieshen: I found that this is unnecessary in the current code base. Thank you.

src/renderer/gl_engine/tvgGlGeometry.cpp
wenjieshen: Thank you. I have removed the legacy code.

src/renderer/gl_engine/tvgGlCommon.h
wenjieshen: I completed it, thanks. 👍

src/renderer/tvgRender.cpp
wenjieshen: I tested the WG engine using `all.sh`, and it functions well in this version. Regarding code readability, I plan to unify wgpu and GL later. The WG can get more advantage of this improvement because it sends 16 float numbers `WgShaderTypeMat4x4f` to the uniform buffer. Thank you.

wenjieshen:
> @wenjieshen Please check `BoundingBox`
> 
Thank you. I have resolved it. 

image

src/renderer/tvgRender.cpp
wenjieshen: Hi, I separated the implementation by a boolean and a macro, so that the WG engine won't be affected. and we can quickly refactor later.

wenjieshen:
Hi @tinyjin, could you please test it on Webcanvas when you are available? Thank you.

src/renderer/gl_engine/tvgGlRenderPass.cpp
```diff
@@ -25,9 +25,28 @@
 #include "tvgGlRenderPass.h"
 #include "tvgGlRenderTask.h"
 
-GlRenderPass::GlRenderPass(GlRenderTarget* fbo): mFbo(fbo), mTasks(), mDrawDepth(0) {}
+static Matrix _viewMatrix(const RenderRegion& vp)
+{
+    if (vp.w() == 0 || vp.h() == 0) return tvg::identity();
```
hermet: @wenjieshen 0-size viewport sounds like a wrong case. Please make it sure this exception is an acceptable case. Otherwise, please remove it.

src/renderer/gl_engine/tvgGlRenderPass.cpp
```diff
@@ -25,9 +25,28 @@
...
+
+    Matrix postMatrix{};
+    tvg::identity(&postMatrix);
+    translate(&postMatrix, {(float)-vp.sx(), (float)-vp.sy()});
+
+    Matrix mvp = tvg::identity();
```
hermet: please do manual initialization `Matrix mvp = { .......... };`

src/renderer/gl_engine/tvgGlRenderPass.cpp
```diff
@@ -25,9 +25,28 @@
 #include "tvgGlRenderPass.h"
 #include "tvgGlRenderTask.h"
 
-GlRenderPass::GlRenderPass(GlRenderTarget* fbo): mFbo(fbo), mTasks(), mDrawDepth(0) {}
+static Matrix _viewMatrix(const RenderRegion& vp)
+{
+    if (vp.w() == 0 || vp.h() == 0) return tvg::identity();
+
+    Matrix postMatrix{};
```
hermet: We can make it more compact. Please do manual initializaion.
`Matrix postMatrix = { 1.0f, ........... x, y ... };`

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1503,4 +1428,23 @@ GlRenderer* GlRenderer::gen(TVG_UNUSED uint32_t threads)
...
\ No newline at end of file
+}
+
+
+static void _solidUniforms(GlShape& sdata, const RenderColor& c, RenderUpdateFlag flag, float* solidInfo)
+{
+    auto a = MULTIPLY(c.a, sdata.opacity);
+
+    if (flag & RenderUpdateFlag::Stroke) {
+        float strokeWidth = sdata.geometry.strokeRenderWidth;
```
hermet: float -> auto

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1503,4 +1428,23 @@ GlRenderer* GlRenderer::gen(TVG_UNUSED uint32_t threads)
...
+        if (strokeWidth < MIN_GL_STROKE_WIDTH) {
+            float alpha = strokeWidth / MIN_GL_STROKE_WIDTH;
```
hermet: float -> auto

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1047,10 +967,26 @@ bool GlRenderer::bounds(RenderData data, Point* pt4, const Matrix& m)
...
+                }
+            } else {
+                // GL stroke vertices are generated in world space.
+                // Normalize to local space first, then remap to the caller-requested space.
+                // - OBB path passes m = identity -> result becomes local (caller applies model later).
+                const auto inverseModel = sdata->geometry.inverseMatrix();
+
+                for (uint32_t i = 0; i < vertexes.count / 2; i++) {
+                    Point vert = {vertexes[i * 2 + 0], vertexes[i * 2 + 1]};
+                    vert *= *inverseModel;
```
hermet: `vert *= (inverseModel * m);`
You can make the code more compact with less LOC.

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1047,10 +967,26 @@ bool GlRenderer::bounds(RenderData data, Point* pt4, const Matrix& m)
...
+                    vert *= m;
+                    bbox.min = min(bbox.min, vert);
```
hermet: This as well:
`box = {min(bbox.min, vert), max( ...) };`

src/renderer/gl_engine/tvgGlCommon.h
```diff
@@ -78,6 +78,21 @@ struct GlGeometryBuffer {
...
+    {
+        if (!inverseMatrixDirty) return &cachedInverseMatrix;
+        inverse(&matrix, &cachedInverseMatrix);
+        inverseMatrixDirty = false;
+        return &cachedInverseMatrix;
+    }
+
+    void setMatrix(const Matrix& tr)
+    {
+        if (matrix == tr) return;
```
hermet: matrix comparsion will have 9 elements values comparsion but they are float values.
I'm wondering it is really helpful or not.
I'm arguing this, but please consider the general scenario. We also have the Transform dirty flag.

wenjieshen:
Hi @hermet, I have addressed the comments. Thank you for your review.

src/renderer/gl_engine/tvgGlGeometry.cpp
```diff
@@ -170,9 +193,10 @@ bool GlGeometry::tesselateShape(const RenderShape& rshape, float* opacityMultipl
 
     // Handle normal shapes with more than 2 points
     BWTessellator bwTess{&fill};
-    bwTess.tessellate(optPath, matrix);
+    auto identityMatrix = tvg::identity();
+    bwTess.tessellate(optPath, identityMatrix);
```
hermet: identity matrix is useless multplication... 
in case of Lottie, `2048 matrix multiply per a frame ` could be saved.

Maybe you can consider this:
```
void BWTessellator::tessellate(const RenderPath& path, const Matrix* matrix = nullptr)
{
  result = transform ?  transform * val : val;
}
```

hermet (commented):
@wenjieshen Please check one more comment. Thanks.

src/renderer/gl_engine/tvgGlGeometry.cpp
wenjieshen: Thank you very much! It is indeed a bug. I provided the identity for convenience, and I told myself I should remove it once the concept proves feasible. 😅

src/renderer/gl_engine/tvgGlShaderSrc.cpp
```diff
@@ -593,9 +588,9 @@ const char* BLIT_FRAG_SHADER = TVG_COMPOSE_SHADER(
 );
 
 const char* BLEND_SHAPE_SOLID_FRAG_HEADER = R"(
-layout(std140) uniform ColorInfo {
+layout(std140) uniform SolidInfo {
     vec4 solidColor;
-} uColorInfo;
+} uSolidInfo;
```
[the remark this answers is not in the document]
hermet: @wenjieshen Please close this if it's invalid.

hermet:
> ## Pull request overview
> Copilot reviewed 18 out of 18 changed files in this pull request and generated 2 comments.
> 
> 💡 [Add Copilot custom instructions](/thorvg/thorvg/new/main/.github/instructions?filename=*.instructions.md) for smarter, more guided reviews. [Learn how to get started](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot).

@wenjieshen Please double-check those copilot reviews, otherwise submit yourself. thanks!

src/renderer/gl_engine/tvgGlShaderSrc.cpp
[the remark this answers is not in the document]
wenjieshen: Thank you. I won't take the risk of disrupting the buffer layout.

PR #4147, opened 2026-02-09
wg_engine: bake model transform into geometry vertices

wenjieshen:
# PR Description

## Summary
This PR combines two renderer refactors and related plumbing updates:
1. Bake model transforms into geometry (instead of per-object transform uniforms) in WG paths.
2. Decouple WG solid fill/stroke from per-draw paint UBO slices using an aux color vertex stream.

It also unifies shared path optimization APIs and updates GL call sites accordingly.

## Motivation
- Reduce per-draw uniform/bind-group overhead in WG.
- Simplify stencil/depth shader interfaces after transform baking.
- Remove duplicated path optimization logic (`optimizeGL` / `optimizeWG`).

## Key Changes
- Shared render core:
  - Replace `RenderPath::optimizeGL()` and `RenderPath::optimizeWG()` with `RenderPath::optimize()`.
- WG transform model:
  - Pre-transform geometry during path processing/tessellation.
  - Remove paint-transform dependency from WG paint uniform usage.
  - Update related bounds/intersection flow and compositor/pipeline binding expectations.
- WG solid path:
  - Bake opacity on CPU (`alpha *= So`) and use shader-side `So = 1.0`.
  - Add staged solid-color stream (`vec4` per draw) + GPU buffer.
  - Bind solid color as an instance-rate aux vertex slot for solid pipelines.
  - Update solid normal/blend shaders and solid pipeline layouts/bind flow.
- GL alignment:
  - Migrate GL geometry prep to shared `optimize()`.
  - Minor transform/bounds and stroke-width handling cleanups for consistency.

## Limitations
- Transform changes can increase CPU work when geometry must be regenerated.
- Solid color stream is still one `vec4` per draw item (no dedup/packing optimization yet).
- The commit groups multiple refactors, reducing bisect granularity.

## FPS gain in the Example Lottie:
- ≈ 10% on Windows11, RTX 5080 Laptop
- ≈ 2% on macOS, M1

### DIFF
--- a/src/renderer/wg_engine/tvgWgCompositor.cpp
+++ b/src/renderer/wg_engine/tvgWgCompositor.cpp
@@ -24,6 +24,23 @@
 #include "tvgWgShaderTypes.h"
 #include <iostream>
 
+void WgCompositor::updateViewMat(WgContext& context, uint32_t width, uint32_t height)
+{
+    if (bindGroupViewMat && viewMatWidth == width && viewMatHeight == height) return;
+
+    WgShaderTypeMat4x4f viewMat(width, height);
+    bool bufferChanged = context.allocateBufferUniform(bufferViewMat, &viewMat, sizeof(viewMat));
+
+    if (bufferChanged || !bindGroupViewMat) {
+        context.layouts.releaseBindGroup(bindGroupViewMat);
+        bindGroupViewMat = context.layouts.createBindGroupBuffer1Un(bufferViewMat);
+    }
+
+    viewMatWidth = width;
+    viewMatHeight = height;
+}
+
+
 void WgCompositor::initialize(WgContext& context, uint32_t width, uint32_t height)
 {
     // pipelines (external handle, do not release)
@@ -32,9 +49,7 @@ void WgCompositor::initialize(WgContext& context, uint32_t width, uint32_t heigh
     // initialize opacity pool
     initPools(context);
     // allocate global view matrix handles
-    WgShaderTypeMat4x4f viewMat(width, height);
-    context.allocateBufferUniform(bufferViewMat, &viewMat, sizeof(viewMat));
-    bindGroupViewMat = context.layouts.createBindGroupBuffer1Un(bufferViewMat);
+    updateViewMat(context, width, height);
     // create render targets handles
     resize(context, width, height);
     // composition and blend geometries
@@ -63,7 +78,10 @@ void WgCompositor::release(WgContext& context)
     // release global view matrix handles
     context.layouts.releaseBindGroup(bindGroupViewMat);
     context.releaseBuffer(bufferViewMat);
+    viewMatWidth = 0;
+    viewMatHeight = 0;
     // release stage buffer
+    stageBufferSolidColor.release(context);
     stageBufferPaint.release(context);
     stageBufferGeometry.release(context);
     // release pipelines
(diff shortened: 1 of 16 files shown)


hermet:
lottie: fps 4%+
stress: fps 3%+

(Tested on AMD Ryzen 75800X / RTX 4700)

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1443,3 +1444,11 @@ static void _solidUniforms(GlShape& sdata, const RenderColor& c, RenderUpdateFla
     solidInfo[2] = c.b / 255.f;
     solidInfo[3] = a / 255.f;
 }
+
+
+static bool _fastTransformMatch(const Matrix& lhs, const Matrix& rhs)
```
hermet: you can just use matrix operator== 
`m ==  m2;`

src/renderer/wg_engine/tvgWgCompositor.cpp
```diff
@@ -24,6 +24,24 @@
 #include "tvgWgShaderTypes.h"
 #include <iostream>
 
+void WgCompositor::updateViewMat(WgContext& context, uint32_t width, uint32_t height)
+{
+    if (width == 0 || height == 0) return;
```
hermet: please remove them. this should not allow the zero size viewport.

src/renderer/wg_engine/tvgWgRenderer.cpp
```diff
@@ -283,11 +294,25 @@ bool WgRenderer::bounds(RenderData data, Point* pt4, const Matrix& m)
...
-                    bbox.max = max(bbox.max, vert);
+
+                if (_fastTransformMatch(m, renderData->transform)) {
+                    for (uint32_t i = 0; i < vertexes.count; i++) {
+                        Point vert = vertexes[i];
+                        bbox.min = min(bbox.min, vert);
+                        bbox.max = max(bbox.max, vert);
+                    }
+                } else {
+                    Matrix inverseModel{};
```
hermet: Please don't initialize it unnecessarily.
`Matrix inverseModel;`

src/renderer/wg_engine/tvgWgCompositor.cpp
```diff
@@ -372,6 +400,23 @@ void WgCompositor::drawMesh(WgContext& context, WgMeshData* meshData)
 };
 
 
+void WgCompositor::drawMeshSolid(WgContext& context, WgMeshData* meshData, uint32_t solidColorInd)
+{
+    assert(meshData);
```
hermet: please remove asserts.

hermet (changes requested):
You shouldn't touch the gl renderer in this PR.

src/renderer/gl_engine/tvgGlRenderer.cpp
wenjieshen: Thank you. I've removed it.

src/renderer/wg_engine/tvgWgCompositor.cpp
wenjieshen: Thank you, I removed it.

src/renderer/wg_engine/tvgWgCompositor.cpp
wenjieshen: Thank you. I removed asserts.

wenjieshen:
> You shouldn't touch the gl renderer in this PR.
I reverted all the GL renderer changes, except for the parts needed to unify the shared optimize-path functions. Thank you.

src/renderer/wg_engine/tvgWgRenderer.cpp
wenjieshen: Thank you. I removed it.

hermet (approved):

PR #4149, opened 2026-02-10
test(lottie): slot regression test

theashraf:


test/testLottie.cpp
```diff
@@ -262,6 +262,79 @@ TEST_CASE("Lottie Quality", "[tvgLottie]")
...
+        REQUIRE(canvas->draw() == Result::Success);
+        REQUIRE(canvas->sync() == Result::Success);
+
+        REQUIRE(animation->apply(0) == Result::Success);
+
+        REQUIRE(canvas->update() == Result::Success);
+        REQUIRE(canvas->draw() == Result::Success);
+        REQUIRE(canvas->sync() == Result::Success);
+
+        REQUIRE(picture->load(TEST_DIR"/slot.json") == Result::Success);
```
tinyjin: Unit test is failing becuase `load` is called after rendering fired. Once the `draw()`/`update()` is called, it would raise `InsufficientCondition`.

Could you separate test cases, or define different animations for this?

tinyjin (changes requested):
Thanks. please check comments

test/testLottie.cpp
tinyjin: Slot API usage looks correct, just usage of `load` is invalid.

theashraf: @tinyjin I've removed the invalid usage of `load`

hermet:
> Image
> I am wondering if it's the case reproduces edge case we've faced. When I check out to v1.0.0 (which had issue), and run this unit test, it passes.

@tinyjin Did you try with the sanitizer? `-Db_sanitize="address,undefined"`

tinyjin:
@hermet I tested both, same passes with `Lottie Slot Regression Test` TC

hermet:
@tinyjin If you've confirmed, we don't necessarily need to apply. 
@theashraf Thanks for your contribution!

PR #4150, opened 2026-02-10
lottie: support 3D orientation(or) property

tinyjin:
Support the "or" (orientation) property in 3D layer transforms, which was the only unsupported transform key from the Lottie spec.

Orientation provides a base 3D rotation [x, y, z] that is applied before individual rx/ry/rz rotations. The implementation references lottie-web's rotation order:
> Rz(rz) * Ry(ry) * Rx(rx) * Rz(or.z) * Ry(or.y) * Rx(or.x)

### DIFF
--- a/src/loaders/lottie/tvgLottieData.h
+++ b/src/loaders/lottie/tvgLottieData.h
@@ -130,6 +130,23 @@ static inline RGB32 lerp(const RGB32& s, const RGB32& e, float t)
     };
 }
 
+struct Point3
+{
+    float x = 0.0f, y = 0.0f, z = 0.0f;
+};
+
+static inline Point3 operator+(const Point3& a, const Point3& b)
+{
+    return {a.x + b.x, a.y + b.y, a.z + b.z};
+}
+static inline Point3 operator-(const Point3& a, const Point3& b)
+{
+    return {a.x - b.x, a.y - b.y, a.z - b.z};
+}
+static inline Point3 operator*(const Point3& a, float t)
+{
+    return {a.x * t, a.y * t, a.z * t};
+}
 }
 
 #endif //_TVG_LOTTIE_COMMON_
--- a/src/loaders/lottie/tvgLottieExpressions.h
+++ b/src/loaders/lottie/tvgLottieExpressions.h
@@ -69,6 +69,13 @@ struct LottieExpressions
     }
 
     template<typename Property>
+    bool result(float frameNo, Point3& out, LottieExpression* exp)
+    {
+        // TODO:
+        return false;
+    }
+
+    template<typename Property>
     bool result(float frameNo, RGB32& out, LottieExpression* exp)
     {
         auto bm_rt = evaluate(frameNo, exp);
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -313,6 +313,23 @@ bool LottieParser::getValue(Point& pt)
     return true;
 }
 
+bool LottieParser::getValue(Point3& pt)
+{
+    if (peekType() == kNullType) return false;
+    if (peekType() == kArrayType) {
+        enterArray();
+        if (!nextArrayValue()) return false;
+    }
+
+    pt.x = getFloat();
+    pt.y = getFloat();
+    pt.z = getFloat();
+
+    while (nextArrayValue())
+        getFloat();  // drop
+
+    return true;
+}
 
 bool LottieParser::getValue(RGB32& color)
 {
@@ -565,7 +582,6 @@ LottieEllipse* LottieParser::parseEllipse()
     return ellipse;
 }
 
-
 LottieTransform* LottieParser::parseTransform(bool ddd)
 {
     auto transform = new LottieTransform;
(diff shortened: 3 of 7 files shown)


tinyjin:
Orientation value has JSON structure like `[0,0,180.758]` (Point3D style), and we don't have 3D property within LottieGenericProperty. So I made own keyframe & static value parsing logic in `parseOrientation()`, that is duplicated logic with existing `parseProperty`.

If we introduce 3D point type to generic property, it might be more tidy. However I didn't make that big change yet since only usage of that is just to orientation. (I'd love to discuss better decision)

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -566,6 +566,87 @@ LottieEllipse* LottieParser::parseEllipse()
...
+                    }
+                }
+                orient->x.prepare();
+                orient->y.prepare();
+                orient->z.prepare();
+            } else skip();
+        } else skip();
+    }
+    return orient;
+}
```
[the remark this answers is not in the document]
hermet: @tinyjin Adding a new parsing logic looks wrong approach. Please check the copilot review.

tinyjin:
> @tinyjin Adding a new parsing logic looks wrong approach. Please check the copilot review.

@hermet The way to simplify parsing logic is to introduce `Point3D`(LottieVec3) type, I think.
The previous `parseOrientation()` was 80 lines duplicating the existing parseKeyFrame<T> boilerplate, because orientation's x/y/z are packed into a single JSON array and the parser can't iterate the same object three times.

```json
{"t": 0, "s": [10, 20, 30]}
```

With a single `Point3D` value, the existing parseKeyFrame<T> infrastructure works as-is, and the entire custom function is replaced with one `parseProperty(*transform->orient)` call.

Since the usage of Point3D is very limited to orientation. I am not certain yet whether it's good to introduce for only this one. I think we have two directions:
1. Keep existing logic and introduce Point3D once another use-case happenes
2. Apply Point3D (naming can be changed)

So I kept Point3D patch in separated commit(8ebede), if it doesn't look good, we can drop and go with way no. 1.
What do you think?

> Also please note, I implemented `Point3D` and related inline functions in `tvgLottieData`, not `thorvg.h` or `tvgMath`, since it's only used with Lottie loader. (_We can also change this for code consistency_)

tinyjin:
@hermet 
I have tested 2D orientation, and found this example only has Z-axis orientation.

Thus we can make 2D only orientation patch not to break exsiting parsing infra, but it won't resolve the issue on our current test file.

CleanShot 2026-02-27 at 15 33 22@2x


Should we just ignore and still need to make this PR with 2D only patch?
Or shall we hold this and investigate more on 3D vector support?

tinyjin:
// hold for investigating unified 3D layer support

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -178,7 +178,7 @@ struct LottieExpression
 //Property would have an either keyframes or single value.
 struct LottieProperty
 {
-    enum class Type : uint8_t {Invalid = 0, Integer, Float, Scalar, Vector, PathSet, Color, Opacity, ColorStop, TextDoc, Image};
+    enum class Type : uint8_t {Invalid = 0, Integer, Float, Scalar, Vector, PathSet, Color, Opacity, ColorStop, TextDoc, Image, Vec3};
```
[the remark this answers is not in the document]
tinyjin: Currently only used in orientation. Can do this after Vec3 applied for all types? (Position, Scale, Rotation...)

tinyjin:
@hermet Orientation is mostly used with z-axis value. Either vec3 or the first patch is ready. Re-requesting a review, Thanks!

src/loaders/lottie/tvgLottieExpressions.h
```diff
@@ -68,6 +68,19 @@ struct LottieExpressions
         return true;
     }
 
+    template<typename Property>
```
hermet, line 71: Let’s remove expressions in this PR(or separate PR) . We don’t have any practical expression support for 3d values.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -314,6 +314,24 @@ bool LottieParser::getValue(Point& pt)
...
+    if (peekType() == kArrayType) {
+        enterArray();
+        if (!nextArrayValue()) return false;
+    }
+
+    pt.x = getFloat();
+    pt.y = getFloat();
+    pt.z = getFloat();
+
+    while (nextArrayValue()) getFloat();  //drop
```
hermet: Let's remove this exception.

src/loaders/lottie/tvgLottieData.h
```diff
@@ -130,6 +130,17 @@ static inline RGB32 lerp(const RGB32& s, const RGB32& e, float t)
     };
 }
 
+
+struct Point3D
```
hermet: `Point3` is enough

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -1038,5 +1038,6 @@ using LottieScalar = LottieGenericProperty<LottieScalarFrame<Point>, Point, Lott
 using LottieVector = LottieGenericProperty<LottieVectorFrame<Point>, Point, LottieProperty::Type::Vector, 0>;
 using LottieColor = LottieGenericProperty<LottieScalarFrame<RGB32>, RGB32, LottieProperty::Type::Color>;
 using LottieOpacity = LottieGenericProperty<LottieScalarFrame<uint8_t>, uint8_t, LottieProperty::Type::Opacity>;
+using LottieVec3 = LottieGenericProperty<LottieScalarFrame<Point3D>, Point3D, LottieProperty::Type::Vec3>;
```
hermet: `LottieVector3`

hermet:
size: +3kb

PR #4168, opened 2026-02-22
lottie: add marker segment query API

### LINKED ISSUE #4167: lottie: add begin and end frames to tvg_lottie_animation_get_marker
Currently, there is no read-only method to retrieve the begin and end frames of a Lottie marker; the workaround of calling `tvg_lottie_animation_set_marker` followed by `tvg_animation_get_segment` is undesirable because it mutates the `Tvg_Animation` internal state (`totalframes` and `duration`), risking unintended playback side effects. To allow dotLottie to inspect marker boundaries without altering state, would be great if we update `tvg_lottie_animation_get_marker` to output the `begin` and `end` frames alongside the marker name, ideally accepting `NULL` pointers for these new parameters to preserve flexibility when only the name is needed.

C API **Current signature:**

```c
TVG_API Tvg_Result tvg_lottie_animation_get_marker(Tvg_Animation animation, uint32_t idx, const char** name);
```

**Suggested signature:**

```c
TVG_API Tvg_Result tvg_lottie_animation_get_marker(Tvg_Animation animation, uint32_t idx, const char** name, float* begin, float* end);
```

theashraf:
## Summary
- **C++**: Unified `const char* marker(uint32_t idx, float* begin = nullptr, float* end = nullptr)` — replaces old `const char* marker(uint32_t idx)` overload, returns marker name and optionally outputs begin/end frame values
- **CAPI**: Add `tvg_lottie_animation_get_marker_info()` with `name`, `begin`, and `end` out-params; deprecate old `tvg_lottie_animation_get_marker()` (kept for ABI compatibility)
- Tests updated to cover begin/end frame queries (26 assertions)

Resolves #4167

### DIFF
--- a/src/bindings/capi/thorvg_capi.h
+++ b/src/bindings/capi/thorvg_capi.h
@@ -3202,6 +3202,22 @@ TVG_API Tvg_Result tvg_lottie_animation_get_marker(Tvg_Animation animation, uint
 
 
 /**
+ * @brief Gets the marker information by a given index.
+ *
+ * @param[in] animation The Lottie animation object.
+ * @param[in] idx The index of the animation marker, starts from 0.
+ * @param[out] name The name of the marker. Pass @c NULL to not receive this value.
+ * @param[out] begin The start frame of the marker. Pass @c NULL to not receive this value.
+ * @param[out] end The end frame of the marker. Pass @c NULL to not receive this value.
+ *
+ * @retval TVG_RESULT_INVALID_ARGUMENT In case @p animation is invalid, @c idx is out of range, or all output parameters are @c NULL.
+ *
+ * @note Experimental API
+ */
+TVG_API Tvg_Result tvg_lottie_animation_get_marker_info(Tvg_Animation animation, uint32_t idx, const char** name, float* begin, float* end);
+
+
+/**
  * @brief Interpolates between two frames over a specified duration.
  *
  * This method performs tweening, a process of generating intermediate frame
--- a/src/bindings/capi/tvgCapi.cpp
+++ b/src/bindings/capi/tvgCapi.cpp
@@ -1235,6 +1235,19 @@ TVG_API Tvg_Result tvg_lottie_animation_get_marker(Tvg_Animation animation, uint
 }
 
 
+TVG_API Tvg_Result tvg_lottie_animation_get_marker_info(Tvg_Animation animation, uint32_t idx, const char** name, float* begin, float* end)
+{
+#ifdef THORVG_LOTTIE_LOADER_SUPPORT
+    if (!animation || (!name && !begin && !end)) return TVG_RESULT_INVALID_ARGUMENT;
+    auto n = reinterpret_cast<LottieAnimation*>(animation)->marker(idx, begin, end);
+    if (!n) return TVG_RESULT_INVALID_ARGUMENT;
+    if (name) *name = n;
+    return TVG_RESULT_SUCCESS;
+#endif
+    return TVG_RESULT_NOT_SUPPORTED;
+}
+
+
 TVG_API Tvg_Result tvg_lottie_animation_tween(Tvg_Animation animation, float from, float to, float progress)
 {
 #ifdef THORVG_LOTTIE_LOADER_SUPPORT
--- a/src/loaders/lottie/thorvg_lottie.h
+++ b/src/loaders/lottie/thorvg_lottie.h
@@ -74,11 +74,25 @@ public:
      * @param[in] idx The index of the animation marker, starts from 0.
      *
      * @retval The name of marker when succeed, @c nullptr otherwise.
-     * 
+     *
      * @see LottieAnimation::markersCnt()
      * @since 1.0
      */
     const char* marker(uint32_t idx) noexcept;
+
+    /**
+     * @brief Gets the marker name and segment range by a given index.
+     *
+     * @param[in] idx The index of the animation marker, starts from 0.
+     * @param[out] begin The start frame of the marker. Pass @c nullptr to not receive this value.
+     * @param[out] end The end frame of the marker. Pass @c nullptr to not receive this value.
+     *
+     * @retval The name of marker when succeed, @c nullptr otherwise.
+     *
+     * @see LottieAnimation::markersCnt()
+     * @note Experimental API
+     */
+    const char* marker(uint32_t idx, float* begin, float* end) noexcept;
 
     /**
      * @brief Updates the value of an expression variable for a specific layer.
--- a/src/loaders/lottie/tvgLottieLoader.cpp
+++ b/src/loaders/lottie/tvgLottieLoader.cpp
@@ -440,11 +440,13 @@ uint32_t LottieLoader::markersCnt()
 }
 
 
-const char* LottieLoader::markers(uint32_t index)
+const char* LottieLoader::markers(uint32_t index, float* begin, float* end)
 {
     if (!ready() || index >= comp->markers.count) return nullptr;
-    auto marker = comp->markers.begin() + index;
-    return (*marker)->name;
+    auto marker = comp->markers[index];
+    if (begin) *begin = marker->time;
+    if (end) *end = marker->time + marker->duration;
+    return marker->name;
 }
 
 
--- a/test/testLottie.cpp
+++ b/test/testLottie.cpp
@@ -176,6 +176,10 @@ TEST_CASE("Lottie Marker", "[tvgLottie]")
         //Set marker name before loaded
         REQUIRE(animation->segment("sectionC") == Result::InsufficientCondition);
 
+        //Get marker info before loaded
+        float markerBegin, markerEnd;
+        REQUIRE(animation->marker(0, &markerBegin, &markerEnd) == nullptr);
+
         //Animation load
         REQUIRE(picture->load(TEST_DIR"/segment.lot") == Result::Success);
 
@@ -191,8 +195,30 @@ TEST_CASE("Lottie Marker", "[tvgLottie]")
         //Get marker name by index
         REQUIRE(!strcmp(animation->marker(1), "sectionB"));
 
-        //Get marker name by invalid index
+        //Get marker name and segment by index
+        REQUIRE(!strcmp(animation->marker(0, &markerBegin, &markerEnd), "sectionA"));
+        REQUIRE(markerBegin == 0.0f);
+        REQUIRE(markerEnd == 22.0f);
+
+        REQUIRE(!strcmp(animation->marker(1, &markerBegin, &markerEnd), "sectionB"));
+        REQUIRE(markerBegin == 22.0f);
+        REQUIRE(markerEnd == 33.0f);
+
+        REQUIRE(!strcmp(animation->marker(2, &markerBegin, &markerEnd), "sectionC"));
+        REQUIRE(markerBegin == 33.0f);
+        REQUIRE(markerEnd == 63.0f);
+
+        //Get marker with only begin
+        REQUIRE(!strcmp(animation->marker(0, &markerBegin, nullptr), "sectionA"));
+        REQUIRE(markerBegin == 0.0f);
+
+        //Get marker with only end
+        REQUIRE(!strcmp(animation->marker(0, nullptr, &markerEnd), "sectionA"));
+        REQUIRE(markerEnd == 22.0f);
+
+        //Get marker by invalid index
         REQUIRE(animation->marker(-1) == nullptr);
+        REQUIRE(animation->marker(-1, &markerBegin, &markerEnd) == nullptr);
 
         REQUIRE(animation->segment(nullptr) == Result::Success);
     }
--- a/src/loaders/lottie/tvgLottieAnimation.cpp
+++ b/src/loaders/lottie/tvgLottieAnimation.cpp
@@ -103,9 +103,15 @@ uint32_t LottieAnimation::markersCnt() noexcept
 
 const char* LottieAnimation::marker(uint32_t idx) noexcept
 {
+    return marker(idx, nullptr, nullptr);
+}
+
+
+const char* LottieAnimation::marker(uint32_t idx, float* begin, float* end) noexcept
+{
     auto loader = to<PictureImpl>(pImpl->picture)->loader;
     if (!loader) return nullptr;
-    return static_cast<LottieLoader*>(loader)->markers(idx);
+    return static_cast<LottieLoader*>(loader)->markers(idx, begin, end);
 }
 
 
--- a/src/loaders/lottie/tvgLottieLoader.h
+++ b/src/loaders/lottie/tvgLottieLoader.h
@@ -94,7 +94,7 @@ public:
 
     //Marker Supports
     uint32_t markersCnt();
-    const char* markers(uint32_t index);
+    const char* markers(uint32_t index, float* begin = nullptr, float* end = nullptr);
     bool segment(const char* marker, float& begin, float& end);
     Result segment(float begin, float end) override;
 


src/bindings/capi/thorvg_capi.h
```diff
@@ -3146,17 +3146,19 @@ TVG_API Tvg_Result tvg_lottie_animation_get_markers_cnt(Tvg_Animation animation,
...
+ * @param[out] begin The start frame of the marker. Pass @c NULL to not receive this value.
+ * @param[out] end The end frame of the marker. Pass @c NULL to not receive this value.
  *
- * @retval TVG_RESULT_INVALID_ARGUMENT In case @c nullptr is passed as the argument or @c idx is out of range.
+ * @retval TVG_RESULT_INVALID_ARGUMENT In case @c idx is out of range or @p animation is invalid.
  *
  * @since 1.0
  */
-TVG_API Tvg_Result tvg_lottie_animation_get_marker(Tvg_Animation animation, uint32_t idx, const char** name);
+TVG_API Tvg_Result tvg_lottie_animation_get_marker(Tvg_Animation animation, uint32_t idx, const char** name, float* begin, float* end);
```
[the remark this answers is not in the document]
theashraf: This is not a breaking change — the original `tvg_lottie_animation_get_marker()` signature is preserved as-is. A new separate function `tvg_lottie_animation_get_marker_segment()` was added instead.

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -443,8 +443,17 @@ uint32_t LottieLoader::markersCnt()
...
     if (!ready() || index >= comp->markers.count) return nullptr;
-    auto marker = comp->markers.begin() + index;
-    return (*marker)->name;
+    return (*(comp->markers.begin() + index))->name;
+}
+
+
+bool LottieLoader::markers(uint32_t index, float* begin, float* end)
+{
+    if (!ready() || index >= comp->markers.count) return false;
```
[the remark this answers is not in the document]
theashraf: The null check for `begin` is already enforced at the public API level in `LottieAnimation::marker()` before this internal function is called. Adding a redundant check here would be inconsistent with the rest of the codebase's internal functions.

src/loaders/lottie/thorvg_lottie.h
```diff
@@ -69,16 +69,18 @@ class TVG_API LottieAnimation final : public Animation
...
+     * @param[out] end The end frame of the marker. Pass @c nullptr to not receive this value.
      *
      * @retval The name of marker when succeed, @c nullptr otherwise.
-     * 
+     *
      * @see LottieAnimation::markersCnt()
      * @since 1.0
      */
-    const char* marker(uint32_t idx) noexcept;
+    const char* marker(uint32_t idx, float* begin = nullptr, float* end = nullptr) noexcept;
```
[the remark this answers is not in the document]
theashraf: This is intentional. The maintainer explicitly requested a single unified function with default parameters in the issue discussion (#4167). Since `begin` and `end` default to `nullptr`, existing C++ source code calling `marker(idx)` continues to compile without changes — source compatibility is preserved. The ABI change is expected and acceptable for this project's release cycle.

src/bindings/capi/tvgCapi.cpp
```diff
@@ -1208,6 +1208,19 @@ TVG_API Tvg_Result tvg_lottie_animation_get_marker(Tvg_Animation animation, uint
...
 
+TVG_API Tvg_Result tvg_lottie_animation_get_marker_info(Tvg_Animation animation, uint32_t idx, const char** name, float* begin, float* end)
+{
+#ifdef THORVG_LOTTIE_LOADER_SUPPORT
+    if (!animation) return TVG_RESULT_INVALID_ARGUMENT;
+    auto n = reinterpret_cast<LottieAnimation*>(animation)->marker(idx, begin, end);
+    if (!n) return TVG_RESULT_INVALID_ARGUMENT;
+    if (name) *name = n;
+    return TVG_RESULT_SUCCESS;
+#endif
```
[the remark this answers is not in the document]
theashraf: The PR description has been updated to reflect the final API naming (`_info` suffix and deprecation of old `get_marker`). The naming was chosen through maintainer feedback in the issue discussion (#4167). The code and description are now aligned.

theashraf:
> @theashraf We can't change the official APIs, this will break ABI, Please discuss those additions with new APIs. Thanks!

@hermet The PR is ready for review. I've updated the APIs according to our discussion in the referenced issue.

hermet:
> > @theashraf We can't change the official APIs, this will break ABI, Please discuss those additions with new APIs. Thanks!
> 
> @hermet The PR is ready for review. I've updated the APIs according to our discussion in the referenced issue.

@tinyjin let's go review.

src/bindings/capi/thorvg_capi.h
```diff
@@ -3154,9 +3154,26 @@ TVG_API Tvg_Result tvg_lottie_animation_get_markers_cnt(Tvg_Animation animation,
  *
  * @retval TVG_RESULT_INVALID_ARGUMENT In case @c nullptr is passed as the argument or @c idx is out of range.
  *
+ * @deprecated Use tvg_lottie_animation_get_marker_info() instead.
```
tinyjin: I'd like to check @hermet 
Can we deprecate this now? I think we need to set up deprecation plan.

src/bindings/capi/tvgCapi.cpp
```diff
@@ -1208,6 +1208,19 @@ TVG_API Tvg_Result tvg_lottie_animation_get_marker(Tvg_Animation animation, uint
 }
 
 
+TVG_API Tvg_Result tvg_lottie_animation_get_marker_info(Tvg_Animation animation, uint32_t idx, const char** name, float* begin, float* end)
+{
+#ifdef THORVG_LOTTIE_LOADER_SUPPORT
+    if (!animation) return TVG_RESULT_INVALID_ARGUMENT;
+    auto n = reinterpret_cast<LottieAnimation*>(animation)->marker(idx, begin, end);
+    if (!n) return TVG_RESULT_INVALID_ARGUMENT;
+    if (name) *name = n;
```
tinyjin, line 1244: assign first and decide for simpler code.
```cpp
*name = reinterpret_cast<LottieAnimation*>(animation)->marker(idx, begin, end);
if (!(*name)) return TVG_RESULT_INVALID_ARGUMENT;
```

src/loaders/lottie/thorvg_lottie.h
```diff
@@ -69,16 +69,18 @@ class TVG_API LottieAnimation final : public Animation
     uint32_t markersCnt() noexcept;
     
     /**
-     * @brief Gets the marker name by a given index.
+     * @brief Gets the marker name and segment boundaries by a given index.
```
tinyjin: nitpick for wording
segment boundaries -> `segment range`

src/loaders/lottie/thorvg_lottie.h
```diff
@@ -69,16 +69,18 @@ class TVG_API LottieAnimation final : public Animation
...
+     * @param[out] end The end frame of the marker. Pass @c nullptr to not receive this value.
      *
      * @retval The name of marker when succeed, @c nullptr otherwise.
-     * 
+     *
      * @see LottieAnimation::markersCnt()
      * @since 1.0
      */
-    const char* marker(uint32_t idx) noexcept;
+    const char* marker(uint32_t idx, float* begin = nullptr, float* end = nullptr) noexcept;
```
tinyjin: @theashraf Wondering, is there much case to pick singular value?
If so good to go, no need default value otherwise.

test/testLottie.cpp
```diff
@@ -176,6 +176,10 @@ TEST_CASE("Lottie Marker", "[tvgLottie]")
         //Set marker name before loaded
         REQUIRE(animation->segment("sectionC") == Result::InsufficientCondition);
 
+        //Get marker info before loaded
+        float b, e;
+        REQUIRE(animation->marker(0, &b, &e) == nullptr);
```
tinyjin: can descriptive naming?
begin, end or markerBegin/markerEnd

test/testLottie.cpp
```diff
@@ -191,8 +195,31 @@ TEST_CASE("Lottie Marker", "[tvgLottie]")
         //Get marker name by index
         REQUIRE(!strcmp(animation->marker(1), "sectionB"));
 
-        //Get marker name by invalid index
+        //Get marker name and segment by index
+        float begin, end;
```
tinyjin: no need to redefine float values again.
> 180:  float b, e;

src/loaders/lottie/thorvg_lottie.h
theashraf: @tinyjin I don't think picking a singular value is common, i'm interested in both, but to make the API flexible for the future use cases it may rise i thought it might be good to keep both optional

src/bindings/capi/tvgCapi.cpp
theashraf: @tinyjin, this assumes name is always a valid pointer. However, a user might pass nullptr if they don't need the marker name. Dereferencing it without a check will cause a crash; we should verify it's not null first

so this check if necessary, but i might be missing something 👀 

```
if (name) *name = n;
```

test/testLottie.cpp
theashraf: resolved in this commit https://github.com/thorvg/thorvg/pull/4168/commits/3bf6d7a4e3c3cb1d7718dd93710961b44f6ff4a7

src/loaders/lottie/thorvg_lottie.h
theashraf: resolved in this commit https://github.com/thorvg/thorvg/pull/4168/commits/ef22be1ec100fc10921d22c333dacc05282c46e5

src/bindings/capi/tvgCapi.cpp
tinyjin: notice the condition `if (!animation) return TVG_RESULT_INVALID_ARGUMENT;` changed.
The guard is necessary so far.

However, in that condition, if we accept singular picking, I think it should be also Invalid Argument when every input is `nullptr` (name / begin / end)

src/loaders/lottie/thorvg_lottie.h
tinyjin: @theashraf This change could break ABI. Rather than changing existing API, you may want to consider overloading:
```cpp
const char* marker(uint32_t idx) noexcept; // keep
+ const char* marker(uint32_t idx, float* begin, float* end) noexcept;
```

We can deprecate original one later.

src/bindings/capi/tvgCapi.cpp
theashraf: 1. i didn't remove that check, it's still there
```cpp
    if (!animation) return TVG_RESULT_INVALID_ARGUMENT;
```

2. if every input can't be nullptr, then we're not assuming singular picking, which is absolutely fine in my case, i'll add that guard as well

theashraf: @tinyjin resolved in this commit https://github.com/thorvg/thorvg/pull/4168/changes/1cab27b425e0aa29fc1b7e3fc92f36f5c231c6bd

src/loaders/lottie/thorvg_lottie.h
theashraf: @tinyjin resolved in https://github.com/thorvg/thorvg/pull/4168/changes/1cab27b425e0aa29fc1b7e3fc92f36f5c231c6bd

src/bindings/capi/thorvg_capi.h
hermet: @tinyjin @theashraf No. Let's deprecate APIs later.

theashraf: done!

src/bindings/capi/thorvg_capi.h
```diff
@@ -3201,6 +3201,22 @@ TVG_API Tvg_Result tvg_lottie_animation_get_markers_cnt(Tvg_Animation animation,
...
+ *
+ * @param[in] animation The Tvg_Animation pointer to the Lottie animation object.
+ * @param[in] idx The index of the animation marker, starts from 0.
+ * @param[out] name The name of the marker. Pass @c NULL to not receive this value.
+ * @param[out] begin The start frame of the marker. Pass @c NULL to not receive this value.
+ * @param[out] end The end frame of the marker. Pass @c NULL to not receive this value.
+ *
+ * @retval TVG_RESULT_INVALID_ARGUMENT In case @p animation is invalid, @c idx is out of range, or all output parameters are @c NULL.
+ *
+ * @since 1.0
```
tinyjin: `v1.0` is already introduced. Should mark as `Experimental API`

src/bindings/capi/thorvg_capi.h
```diff
@@ -3201,6 +3201,22 @@ TVG_API Tvg_Result tvg_lottie_animation_get_markers_cnt(Tvg_Animation animation,
 TVG_API Tvg_Result tvg_lottie_animation_get_marker(Tvg_Animation animation, uint32_t idx, const char** name);
 
 
+/**
+ * @brief Gets the marker information by a given index.
+ *
+ * @param[in] animation The Tvg_Animation pointer to the Lottie animation object.
```
tinyjin: ` * @param[in] animation The Lottie animation object.`

src/loaders/lottie/thorvg_lottie.h
```diff
@@ -74,12 +74,26 @@ class TVG_API LottieAnimation final : public Animation
...
+     * @brief Gets the marker name and segment range by a given index.
+     *
+     * @param[in] idx The index of the animation marker, starts from 0.
+     * @param[out] begin The start frame of the marker. Pass @c nullptr to not receive this value.
+     * @param[out] end The end frame of the marker. Pass @c nullptr to not receive this value.
+     *
+     * @retval The name of marker when succeed, @c nullptr otherwise.
+     *
+     * @see LottieAnimation::markersCnt()
+     * @since 1.0
```
tinyjin: Experimental API here too

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -440,11 +440,13 @@ uint32_t LottieLoader::markersCnt()
 }
 
 
-const char* LottieLoader::markers(uint32_t index)
+const char* LottieLoader::markers(uint32_t index, float* begin, float* end)
 {
     if (!ready() || index >= comp->markers.count) return nullptr;
-    auto marker = comp->markers.begin() + index;
-    return (*marker)->name;
+    auto marker = *(comp->markers.begin() + index);
```
tinyjin:
```
auto marker = comp->markers[index];
```

.clang-format
tinyjin, line 88: accidental addition? If so, please remove change

tinyjin (commented):
Almost good. Please check comments :)

src/bindings/capi/thorvg_capi.h
theashraf: Fixed — replaced `@since 1.0` with `@note Experimental API`.

src/bindings/capi/thorvg_capi.h
theashraf: Fixed — updated to `@param[in] animation The Lottie animation object.`

src/loaders/lottie/thorvg_lottie.h
theashraf: Fixed — marked as `@note Experimental API` here as well.

src/loaders/lottie/tvgLottieLoader.cpp
theashraf: Fixed — simplified to `auto marker = comp->markers[index];`

.clang-format
theashraf: Yes, accidental — removed in the latest push.

test/testLottie.cpp
```diff
@@ -191,8 +195,30 @@ TEST_CASE("Lottie Marker", "[tvgLottie]")
...
+        REQUIRE(!strcmp(animation->marker(0, &markerBegin, nullptr), "sectionA"));
+        REQUIRE(markerBegin == 0.0f);
+
+        //Get marker with only end
+        REQUIRE(!strcmp(animation->marker(0, nullptr, &markerEnd), "sectionA"));
+        REQUIRE(markerEnd == 22.0f);
+
+        //Get marker by invalid index
         REQUIRE(animation->marker(-1) == nullptr);
+        REQUIRE(animation->marker(-1, &markerBegin, &markerEnd) == nullptr);
```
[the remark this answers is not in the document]
theashraf: This is a pre-existing pattern in the codebase (e.g. `testPaint.cpp: shape->opacity(-1)`). The implicit conversion to `UINT32_MAX` is intentional for testing out-of-range behavior.

tinyjin:
@theashraf squashed and rebased commits for convenience (code didn't change)

tinyjin (approved):
LGTM

hermet:
https://github.com/thorvg/thorvg/pull/4183

Issue #4169, opened 2026-02-23
lottie: slot apply does not update rendered output


theashraf:
[slots.json](https://github.com/user-attachments/files/25476309/slots.json)

```cpp
TEST_CASE("Lottie Slot Render Update", "[tvgLottie]")
{
    REQUIRE(Initializer::init() == Result::Success);
    {
        const uint32_t W = 800;
        const uint32_t H = 800;
        auto buffer = unique_ptr<uint32_t[]>(new uint32_t[W * H]());

        auto canvas = SwCanvas::gen();
        REQUIRE(canvas);
        REQUIRE(canvas->target(buffer.get(), W, W, H, ColorSpace::ABGR8888) == Result::Success);

        auto animation = unique_ptr<LottieAnimation>(LottieAnimation::gen());
        REQUIRE(animation);

        auto picture = animation->picture();
        REQUIRE(picture->load(TEST_DIR"/slot.lot") == Result::Success);
        REQUIRE(canvas->add(picture) == Result::Success);

        //Render initial frame
        REQUIRE(canvas->update() == Result::Success);
        REQUIRE(canvas->draw() == Result::Success);
        REQUIRE(canvas->sync() == Result::Success);

        //Sample pixel at rectangle center (layer 8, ExpressionLayer, ~256,256)
        auto pixelBefore = buffer[256 * W + 256];
        REQUIRE(pixelBefore != 0);

        //Apply a 45° rotation via rect_rotation slot
        auto slotId = animation->gen(R"({"rect_rotation":{"p":{"a":0,"k":45}}})");
        REQUIRE(slotId > 0);
        REQUIRE(animation->apply(slotId) == Result::Success);

        //Render again
        REQUIRE(canvas->update() == Result::Success);
        REQUIRE(canvas->draw() == Result::Success);
        REQUIRE(canvas->sync() == Result::Success);

        auto pixelAfter = buffer[256 * W + 256];

        //Pixel should change after 45° rotation
        REQUIRE(pixelBefore != pixelAfter);  // FAILS
    }
    REQUIRE(Initializer::term() == Result::Success);
}
```

```
REQUIRE( pixelBefore != pixelAfter )
with expansion:
  4294672484 (0xfffb8064) != 4294672484 (0xfffb8064)
```

`apply()` succeeds but doesn't invalidate the cached frame for the next render.

theashraf:
[...]

PR #4170, opened 2026-02-23
lottie/slot: fix stale scene data on slot override/reset while paused

tinyjin:
When the animation is paused (no active `frame()` calls), triggering a scene rebuild via `apply()` or `del()` left stale layer data visible after re-rendering.

`comp->clear()` was only called inside `frame()` and `tween()`, so rebuild paths never cleared the previous scene.

Move `comp->clear()` into `run()`, directly before `builder->update()`, and remove it from `frame()` and `tween()`. Since `run()` is the single execution point for all rebuilds.

issue: https://github.com/thorvg/thorvg/issues/4169

| Before | After |
|------|-------|
| CleanShot 2026-02-23 at 17 12 28 | CleanShot 2026-02-23 at 17 12 02 |

### DIFF
--- a/src/loaders/lottie/tvgLottieLoader.cpp
+++ b/src/loaders/lottie/tvgLottieLoader.cpp
@@ -427,7 +427,10 @@ void LottieLoader::sync()
 {
     done();
 
-    if (build) run(0);
+    if (build) {
+        if (comp) comp->clear();
+        run(0);
+    }
 }
 
 


src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -305,7 +305,10 @@ bool LottieLoader::apply(uint32_t slotcode, bool byDefault)
         }
     }
     curSlot = slotcode;
-    if (applied) build = true;
+    if (applied) {
+        if (comp) comp->clear();
```
hermet: hmm clear here...  and when rebuild the scene? Please double-check it. Thanks.

hermet (changes requested):
@tinyjin please confirm the comment. thanks.

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -62,8 +62,10 @@ bool LottieLoader::prepare()
 
 void LottieLoader::run(unsigned tid)
 {
-    if (comp) builder->update(comp, frameNo);      //update frame
-    else if (prepare()) builder->update(comp, 0);  //initial loading
+    if (comp) {
+        comp->clear();
```
hermet: You can not clear the scene on a thread. it's not safe.

please check this:

```
void LottieLoader::sync()
{
    done();

    if (build) {
        if (comp) comp->clear();
        run(0);
    }
}
```

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -498,8 +499,6 @@ bool LottieLoader::tween(float from, float to, float progress)
 
     builder->onTween(shorten(to), progress);
 
-    if (comp) comp->clear();     //clear synchronously
```
hermet, line 501: No, you should not remove this

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -391,8 +391,6 @@ bool LottieLoader::frame(float no)
 
     builder->offTween();
 
-    if (comp) comp->clear();     //clear synchronously
```
hermet, line 394: No, you should not remove this

src/loaders/lottie/tvgLottieLoader.cpp
```diff
@@ -427,7 +425,10 @@ void LottieLoader::sync()
 {
     done();
 
-    if (build) run(0);
+    if (build) {
+      if (comp) comp->clear();
```
hermet: + indentation check

src/loaders/lottie/tvgLottieLoader.cpp
tinyjin: ah yes, I was on checking. Forgot to remove

Issue #3801, opened 2025-09-12
lottie: rendering discrepancies (dotlottie)


theashraf:
<img width="400" height="663" alt="Image" src="https://github.com/user-attachments/assets/aea1e444-3219-4b45-aed3-707eacf516f2" />

[data.json](https://github.com/user-attachments/files/22289678/data.json)

<img width="400" height="667" alt="Image" src="https://github.com/user-attachments/assets/383a4fd6-2ee7-4252-b809-cac81e38e35f" />

[test.json](https://github.com/user-attachments/files/22289696/test.json)

rgprog:
Reduced test case for the first file:

````json
[...]

PR #4172, opened 2026-02-23
lottie: fix solid fill color distortion with bezier easing

tinyjin:
When a fill color keyframe has Bezier easing with a negative Y tangent, the interpolator's progress() can return values outside [0, 1], causing the lerp result to exceed the valid RGB range [0, 255]. This overflow wraps on cast to uint8_t, producing incorrect colors. Add a `tvg::lerp` overload for RGB32 that clamps each channel to [0, 255].

issue(partly): https://github.com/thorvg/thorvg/issues/3801

---

test file:  [solid-fill-ezier easing overshoot.json](https://github.com/user-attachments/files/25479712/solid-fill-ezier.easing.overshoot.json)

> This file has out tangent with y=-1.471 (negative)

When I test lottie-web ignores this issue, since the Web HTML renderer/Canvas automatically clamps [RGB](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/rgb) values to 0-255.

Ghostty 2026-02-23 18 17 12

> When the value exceeds over 255, discrepancy.

| Issue | Patched |
| --- | --- |
 | CleanShot 2026-02-23 at 18 31 32 | CleanShot 2026-02-23 at 18 30 56 |

### DIFF
--- a/src/loaders/lottie/tvgLottieData.h
+++ b/src/loaders/lottie/tvgLottieData.h
@@ -23,9 +23,11 @@
 #ifndef _TVG_LOTTIE_COMMON_
 #define _TVG_LOTTIE_COMMON_
 
-#include <cmath>
-#include "tvgCommon.h"
 #include "tvgArray.h"
+#include "tvgMath.h"
+
+namespace tvg
+{
 
 struct PathSet
 {
@@ -118,5 +120,16 @@ static inline RGB32 operator*(const RGB32& lhs, float rhs)
     return {(int32_t)nearbyintf(lhs.r * rhs), (int32_t)nearbyintf(lhs.g * rhs), (int32_t)nearbyintf(lhs.b * rhs)};
 }
 
+
+static inline RGB32 lerp(const RGB32& s, const RGB32& e, float t)
+{
+    return {
+        tvg::clamp((int32_t)(s.r + (e.r - s.r) * t), 0, 255),
+        tvg::clamp((int32_t)(s.g + (e.g - s.g) * t), 0, 255),
+        tvg::clamp((int32_t)(s.b + (e.b - s.b) * t), 0, 255)
+    };
+}
+
+}
 
 #endif //_TVG_LOTTIE_COMMON_


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -308,7 +308,12 @@ bool LottieBuilder::updateSolidFill(LottieGroup* parent, LottieObject** child, f
 
     ctx->merging = nullptr;
     auto color = fill->color(frameNo, tween, exps);
-    ctx->propagator->fill(color.r, color.g, color.b, opacity);
+    ctx->propagator->fill(
```
hermet: @tinyjin This looks ad-hoc fix, there are a bunch of LottieColor usages.

hermet (changes requested):
@tinyjin Please check a comment.

src/loaders/lottie/tvgLottieBuilder.cpp
tinyjin: Yes, the original fixes only apply to `fill` function.
I implemented lerp method for RGB32 under tvg namespace in lottie loader, so that RGB32 used in lottie will be clamp 0-255 when interpolation.

please have a look. Thanks

src/loaders/lottie/tvgLottieData.h
```diff
@@ -119,4 +120,16 @@ static inline RGB32 operator*(const RGB32& lhs, float rhs)
 }
 
 
+namespace tvg {
+    static inline RGB32 lerp(const RGB32& s, const RGB32& e, float t)
+    {
+        return {
+            tvg::clamp((int32_t)nearbyintf(s.r + (e.r - s.r) * t), 0, 255),
```
hermet: Let's truncate the floating values.
 
```
        return {
            tvg::clamp((int32_t)(s.r + (e.r - s.r) * t), 0, 255),
            tvg::clamp((int32_t)(s.g + (e.g - s.g) * t), 0, 255),
            tvg::clamp((int32_t)(s.b + (e.b - s.b) * t), 0, 255)
        };
```

src/loaders/lottie/tvgLottieData.h
```diff
@@ -119,4 +120,16 @@ static inline RGB32 operator*(const RGB32& lhs, float rhs)
 }
 
 
+namespace tvg {
```
hermet: please remove the namespace.

tinyjin: @hermet  I think namespace is necessary? The `lerp` method is replacing under namespace:
```
// tvgLottieProperty.h
return tvg::lerp(value, next->value, t);
```

hermet: @tinyjin Please add namespace to the full file scope.

tinyjin: @hermet done!

Issue #4154, opened 2026-02-12
gl: regression bug


hermet:
Visual art is broken at SVG example

<img width="306" height="244" alt="Image" src="https://github.com/user-attachments/assets/17eac4a8-1f45-457e-84eb-eb820718904c" />

Normal:

<img width="248" height="240" alt="Image" src="https://github.com/user-attachments/assets/49e0c8fa-e2b8-41d0-a7c6-1a6046693ae7" />

Curlprit:
```
commit cdc523212e9a0442607f37aa4708dd9ea4aa89da (HEAD)
Author: Jay Shen <jay@lottiefiles.com>
Date:   Thu Nov 27 15:54:25 2025 +0800

    renderer: improved path optimization tolerance calculation

    Replace PX_TOLERANCE with length-based tEps for accurate collinear checks.
    Tested across examples and prior problematic cases with no Lottie.exe FPS
    regressions. Related: #3999
```

wenjieshen:
I found the issue came from how backward collinear points were handled after the transformed-path optimization moved into a shared CPU path.
I tried two fix directions. First, a focused change that keeps backward-collinear points instead of rewriting earlier vertices. Second, a broader conservative-optimization approach that resolved the issue more directly and was simpler.

<img width="2784" height="2160" alt="Image" src="https://github.com/user-attachments/assets/3d324ffa-dd00-4993-97f9-9ec60b31b6c8" />

PR #4176, opened 2026-02-24
renderer: remove colinear merge and use thin-path flag

wenjieshen:
## Root cause

A regression in transformed-path collinear reduction was introduced when model transforms moved to CPU geometry optimization.

In `processLineCollinear()`, we rewrote an earlier vertex (`out.pts[prevPrevIdx] = ptT`) instead of preserving that backward turn as an explicit segment.  
For thin/near-collinear contours, this could alter contour progression and winding, causing shape flip/backtracking artifacts (issue #4154).

This behavior was introduced in `7be22d20` and later affected both GL and WG after `7f758a4f` unified usage of the transformed optimizer.

## Resolution

- Remove only the colinear merge process in `RenderPath::optimize(...)
- Since colinear points are no longer merged, collect thinness during optimization and return it by call-by-reference.
Use that thinness result in GL/WG to choose stroke tessellation fallback, instead of relying on optPath.pts.count == 2.

(For cubic paths, thinness detection does not require a precise area once the cubic is confirmed non-degenerate and non-line-like.)

This keeps simplification benefits for forward collinear extension while preventing winding/shape flips in backward cases, resolving #4154.
image

The SVG that accompanies the commit `7be22d20` also displays the correct rendering result.
image

https://github.com/user-attachments/assets/2d4b1ef0-524e-4ae0-a493-b477fc2c8f3f

- The FPS of example Lottie is the same.
- No obvious regression during running all examples.

### DIFF
--- a/src/renderer/gl_engine/tvgGlCommon.h
+++ b/src/renderer/gl_engine/tvgGlCommon.h
@@ -91,7 +91,7 @@ struct GlGeometry
     void prepare(const RenderShape& rshape);
     bool tesselateShape(const RenderShape& rshape, float* opacityMultiplier = nullptr);
     bool tesselateStroke(const RenderShape& rshape);
-    bool tesselateLine(const RenderPath& path);
+    bool tesselateThinPath(const RenderPath& path);
     void tesselateImage(const RenderSurface* image);
     bool draw(GlRenderTask* task, GlStageBuffer* gpuBuffer, RenderUpdateFlag flag);
     GlStencilMode getStencilMode(RenderUpdateFlag flag);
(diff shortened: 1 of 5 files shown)


hermet:
@wenjieshen I can't see the fix, Could you please double-check it? 

image

wenjieshen:
> @wenjieshen I can't see the fix. Could you please double-check it?
> 
I am sorry for the inconvenience. I have improved the approach. Could you please test it again when you are available? Thank you.

wenjieshen:
> please check this:
> 
Thank you. I realized that merging collinear lines can be quite risky, as different tolerances or conditions may cause conflicts between two test cases. I have removed the merging process and replaced it with an accumulative approach to determine if the entire path forms a thin line.

hermet (approved):

PR #4179, opened 2026-02-26
lottie/expressions: support $bm_mod operator

tinyjin:
see : https://lottiefiles.github.io/lottie-docs/expressions/#mod

### DIFF
--- a/src/loaders/lottie/tvgLottieExpressions.cpp
+++ b/src/loaders/lottie/tvgLottieExpressions.cpp
@@ -539,6 +539,12 @@ static jerry_value_t _div(const jerry_call_info_t* info, const jerry_value_t arg
 }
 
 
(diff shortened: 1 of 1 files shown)


src/loaders/lottie/tvgLottieExpressions.cpp
```diff
@@ -539,6 +539,14 @@ static jerry_value_t _div(const jerry_call_info_t* info, const jerry_value_t arg
 }
 
 
+static jerry_value_t _mod(const jerry_call_info_t* info, const jerry_value_t args[], const jerry_length_t argsCnt)
+{
+    auto a = _number(args[0]);
```
hermet: return jerry_number(fmod(_number(args[0]), _number(args[1])));

Issue #4187, opened 2026-03-04
gl_engine: regression bug


hermet:
Some random glitches are observable

<img width="166" height="193" alt="Image" src="https://github.com/user-attachments/assets/a39a54f8-627b-4337-a653-a92141636392" />

curlprit:
```
commit 914138ded6b07e0ed2c90e40f97f3fae9358b8fd (HEAD)
Author: Jay Shen <jay@lottiefiles.com>
Date:   Sun Jan 18 19:56:27 2026 +0800

    gl_engine: fix partial blit fail on GLES MSAA buffer

    OpenGLES and WebGL do not allow blitting/resolving multisampled
    framebuffers using partial regions. It requires full buffer blits.
    This commit forces full buffer blits for GLES/GL targets and
    introduces a BlendRegion uniform to correctly recalculate texture
    coordinates when the render target size differs from the viewport.
    Fixes #4112.
```

[intelia_logo_animation.json](https://github.com/user-attachments/files/25746324/intelia_logo_animation.json)

PR #4189, opened 2026-03-05
gl_engine: fix GL scene blend blit source/extent mismatch

wenjieshen:
In GlSceneBlendTask::run() (GL path), the blit target is mDstCopyFbo->resolvedFbo but viewport/scissor were configured from mSrcFbo->viewport.

This can mismatch destination buffer space and clip the copy, leaving stale pixels that show up as blend artifacts.

Set GL viewport/scissor from mDstCopyFbo->width/height before blitting.

Related issue: https://github.com/thorvg/thorvg/issues/4187

https://github.com/user-attachments/assets/bbc2a647-3812-4cd4-a73e-bd884146adac

### DIFF
--- a/src/renderer/gl_engine/tvgGlShaderSrc.cpp
+++ b/src/renderer/gl_engine/tvgGlShaderSrc.cpp
@@ -689,6 +689,71 @@ void getFragData() {
 vec4 postProcess(vec4 R) { return R; }
 )";
 
+// GL keeps a viewport-sized dst copy, so src/dst can share vUV.
+// GLES/WebGL must keep a full resolved dst copy because MSAA resolve/blit is only valid for the
+// full buffer there; down-blitting into a smaller FBO would add another full copy pass. Rebuild
+// dst UV from gl_FragCoord + BlendRegion instead of reusing vUV.
+#if defined(THORVG_GL_TARGET_GL)
+const char* BLEND_IMAGE_FRAG_HEADER = R"(
+uniform sampler2D uSrcTexture;
+uniform sampler2D uDstTexture;
+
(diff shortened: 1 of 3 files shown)


hermet:
@wenjieshen what do you think?
image

wenjieshen:
Hi @hermet ,
I apologize for not noticing this earlier. We are reusing a larger copy buffer but only partially overwriting it, which can leave stale pixels. I added the GL specialized shader back because I would prefer to maintain two versions of the shader code instead of clearing the buffers every frame. However, the GLES version must clear the buffers because of its restriction, but it is still faster then bliting twice or bliting whole screen.

src/renderer/gl_engine/tvgGlShaderSrc.cpp
```diff
@@ -689,6 +689,67 @@ void getFragData() {
 vec4 postProcess(vec4 R) { return R; }
 )";
 
+#if defined(THORVG_GL_TARGET_GL)
```
hermet, line 696: @wenjieshen This needs a comment in code. Please describe why this shader needs to be separated for GL and GLES and what the differences between them are. Thanks.

hermet (commented):
@wenjieshen Please check a comment.

src/renderer/gl_engine/tvgGlShaderSrc.cpp
wenjieshen: Thank you for the review. I have added a comment.

tinyjin (approved):
Confirmed no regression in all Web examples and Lottie test assets I have.

Image

Issue #4120, opened 2026-01-19
Compromised thread safety due to incorrect use of std::atomic? (tvgSwRenderer rendererCnt, maybe other locations)


lobotony:
Hello,

first of all: great library, thank you for your hard work!

I've been able to use it in a project, but we've recently stumbled over crashes. I can't post code due to the complexity of the scenario with somewhat concurrent loads of SVGs and PNGs, but while investigating the issue, we found the following indicators prior to a crash: 

`atomic<int32_t> rendererCnt` in tvgSwRenderer.cpp would dip under the threshold of -1, leading to the occasional crash. 

Looking at the code, it seems to me the `rendererCnt` variable is not guarded properly against concurrent modification. 

There are a few places with code similar to` if(rendererCnt == X) {...}`, but this will not prevent multiple threads from crossing this valid threshold in short succession, and then all of them performing the same change on the variable. 

I'm not particularly familiar with std::atomic, but looking at the documentation and a few examples, it seems to me that you'd have to use a combination of `compare_exchange_weak`, `compare_exchange_strong`, `fetch_sub` and `fetch_add` to correctly guard against concurrent modification without a lock. Neither of these calls can be found in the project. 

Maybe this would also explain the other concurrency issues that were reported on github?

hermet:
@lobotony Hi, Please let me clarify your scenario first.

Did you use a ThorVG canvas with multiple threads? Basically, ThorVG Canvas APIs are not thread-safe. You must ensure that each canvas instance and all paint operations on it are used from the same thread.

Canvas instance A accessing on Single Thread 1 → OK

Canvas instance B accessing on on Single Thread 2 → OK

Canvas instance A accesing on multiple Threads, 1 and 2 → Not OK

lobotony:
@hermet thanks for the quick answer!

Each UI element currently has its own canvas. 
Each control loads a single SVG file, renders it into a bitmap and caches the result. 
The UI is bound to the main thread as usual, but the resource loading is asynchronous and multi-threaded.
So it should be strictly one render thread per canvas. 

I have the following hypothesis:

* in theory, all UI elements and canvases are safely separated, and their memory and threading should not interfere
* lets say we have three elements, they loaded their resources, rendered and cached, and now their canvases go out of scope, because the result is cached. 
* one of the SwCanvas destructors calls SwRenderer::term(), which decides that rendererCnt can be set to -1
[...]

PR #4206, opened 2026-03-12
engines: improve thread safety

hermet:
Allow users to create engines on worker threads.
Enhance thread safety in multithreaded environments by using a scoped locking mechanism for global
engine resources.

Test Code
```
#include <iostream>
#include <thread>
#include <vector>
#include <thorvg-1/thorvg.h>

using namespace std;
using namespace tvg;


void worker(int id) {
    std::cout << "Worker " << id << " is running\n";
    delete(canvas);
}

{
    tvg::Initializer::init(4);

    const int numThreads = 100;
    std::vector<std::thread> workers;

    // thread ceate
    for (int i = 0; i < numThreads; ++i) {
        workers.emplace_back(worker, i);
    }

    // thread join
    for (auto& t : workers) {
        t.join();
    }

    tvg::Initializer::term();

    return 0;
}
```

issue: https://github.com/thorvg/thorvg/issues/4120

### DIFF
--- a/test/testSwCanvas.cpp
+++ b/test/testSwCanvas.cpp
@@ -20,6 +20,8 @@
  * SOFTWARE.
  */
 
+#include <thread>
+#include <vector>
 #include <thorvg.h>
 #include "config.h"
 #include "catch.hpp"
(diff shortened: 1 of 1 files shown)


src/meson.build
```diff
@@ -57,7 +57,7 @@ subdir('savers')
 
 thorvg_lib_dep = [common_dep, utils_dep, loader_dep, saver_dep]
 
-if get_option('threads') and host_machine.system() != 'windows' and host_machine.system() != 'android'
+if host_machine.system() != 'windows' and host_machine.system() != 'android'
```
hermet, line 60: Need to double-confirm any side effect.

PR #4220, opened 2026-03-16
lottie: skip text layout when bbox size is not specified

tinyjin:
Restore baseline offset for text layers without a bounding box. Use the layout only when a bounding box is specified.

Fix regression for 
[lottie.json](https://github.com/user-attachments/files/26039029/lottie.-.2026-03-17T082549.132.json)

This lottie doesn't have "sz"(bbx size), then previous layout and align logic made text shifed in negative direction.

culprit: https://github.com/thorvg/thorvg/commit/4d2830ca48bdd45a451a68f3509639f97cb77211

Ghostty 2026-03-17 08 22 45

Also verified test file from original issue.

Ghostty 2026-03-17 08 20 06

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -974,8 +974,10 @@ void LottieBuilder::updateURLFont(LottieLayer* layer, float frameNo, LottieText*
     paint->size(doc.size * 75.0f); //1 pt = 1/72; 1 in = 96 px; -> 72/96 = 0.75
     if (text->font && text->font->style && strstr(text->font->style, "Italic")) paint->italic();
     paint->text(buf);
+
+    float lottieAscent = text->font->ascent * doc.size / 100.0f;
     paint->layout(doc.bbox.size.x, doc.bbox.size.y);
-    paint->translate(doc.bbox.pos.x, doc.bbox.pos.y);
+    paint->translate(doc.bbox.pos.x, doc.bbox.pos.y + (doc.bbox.size.y > 0.0f ? lottieAscent : 0.0f) - doc.shift);
 
     //align the text to the base line
     TextMetrics metrics;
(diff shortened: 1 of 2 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -974,13 +974,16 @@ void LottieBuilder::updateURLFont(LottieLayer* layer, float frameNo, LottieText*
...
     if (text->font && text->font->style && strstr(text->font->style, "Italic")) paint->italic();
     paint->text(buf);
-    paint->layout(doc.bbox.size.x, doc.bbox.size.y);
-    paint->translate(doc.bbox.pos.x, doc.bbox.pos.y);
-
-    //align the text to the base line
-    TextMetrics metrics;
-    paint->metrics(metrics);
-    paint->align(doc.justify, metrics.ascent / (metrics.ascent - metrics.descent));
+    if (doc.bbox.size.x > 0.0f || doc.bbox.size.y > 0.0f) {
```
hermet: I guess analytically, we can try derive a stable y-align value, rather than relying on this different conditional logic.

tinyjin: @hermet The bbox check logic looks required, as lottie-web also add extra offset when the size is available.
(_Shift, lottieAcsent are added as well, they need to be added if available_)

But I simplified logic by ensuring size is not negative in parsing time. (The spec accepts negative values, it's just float, so preventing logic is on the engine side, I think)

And then made simple line y-align value, still condition needed, just simplified in ternary style.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -148,7 +148,12 @@ bool LottieParser::getValue(TextDocument& doc)
         else if (KEY_AS("ls")) doc.shift = getFloat();
         else if (KEY_AS("fc")) getValue(doc.color);
         else if (KEY_AS("ps")) getValue(doc.bbox.pos);
-        else if (KEY_AS("sz")) getValue(doc.bbox.size);
+        else if (KEY_AS("sz")) {
+            getValue(doc.bbox.size);
+            //bbox size is positive
+            doc.bbox.size.x = tvg::clamp(doc.bbox.size.x, 0.0f, FLT_MAX);
```
hermet, line 154: The bbox could have signed positioning. This looks totally a wrong turn.

hermet:
https://github.com/thorvg/thorvg/pull/4235

PR #4221, opened 2026-03-17
fix(sw_engine): prevent blend mode leak into compositor scope

BenjaminHalko:
## Summary
Fix blend mode state leaking from a blended Scene's compositor back to the parent surface in the SW renderer, causing color corruption when a blend group is nested inside a masked/matted composition (e.g. a Lottie track matte).
## Problem
When `SceneImpl::render()` sets a blend mode via `renderer->blend()` before creating a compositor via `target()`, the parent surface retains the stale blender after the compositor switches the render target. If the parent surface has an active compositor (e.g. a mask/matte), the blender leaks to sibling shapes rendered afterward, corrupting their output with wrong colors.
This produces visible artifacts (bright cyan/magenta) in Lottie animations that combine track mattes (`td`/`tt`/`tp`) with blend mode groups (`bm`). The GL and WebGPU renderers are unaffected because they apply blend modes at compositing time rather than during shape rasterization.
## Fix
In `SwRenderer::target()`, when creating a blending compositor inside an existing compositor, move the blend state from the parent surface to the new compositor surface. This keeps the blend active for children rendering on the compositor while preventing it from leaking back to the parent.
The condition `(flags & CompositionFlag::Blending) && surface->compositor` ensures this only triggers for the nested case - standalone blend mode layers (no parent compositor) are unaffected.
## Test
- [blend_matte.json](https://github.com/user-attachments/files/26099981/blend_matte.json) - Lottie with blend groups inside track mattes: was corrupted, now renders correctly
- [blend.json](https://github.com/user-attachments/files/26098141/blend.json) - Lottie with layer-level blend modes (no mattes): blend effects preserved, no regression

| Without fix | With fix |
|:-:|:-:|
| corrupted | correct |

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -426,7 +426,7 @@ void LottieBuilder::appendRect(Shape* shape, Point& pos, Point& size, float r, b
     }
 
     if (ctx->offset) {
-        ctx->offset->modifyRect(to<ShapeImpl>(temp)->rs.path, to<ShapeImpl>(shape)->rs.path);
+        ctx->offset->rect(to<ShapeImpl>(temp)->rs.path, to<ShapeImpl>(shape)->rs.path);
         Paint::rel(temp);
     }
 }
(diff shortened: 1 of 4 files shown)


BenjaminHalko:
OK, this should be ready to merge now! 👍

hermet:
@BenjaminHalko Thanks for your PR. Could you please rebase the commits? I think you only need a core fix with 1 commit. and don't need to add a test case in this case. Thanks.

BenjaminHalko:
@hermet Done! 👍

hermet (changes requested):
Thanks for the contribution, but a regression has been observed. This needs to be fixed more carefully.
Feel free to leave this to the maintainers to fix. Thanks.

before -> after
Image

[1f409.json](https://github.com/user-attachments/files/26134734/1f409.json)

hermet:
related: https://github.com/thorvg/thorvg/pull/4240

hermet:
can be replaced with https://github.com/thorvg/thorvg/pull/4242

Issue #4141, opened 2026-02-05
gl, wgpu: Optimize the rendering performance by reducing the intersection overhead


wenjieshen:
## Background

We currently choose between the optimized draw method and the safe draw method based on whether a path self-intersects. Exact self-intersection checks are expensive for complex paths and can become a performance bottleneck.

## Proposal

Implement an approximate self-intersection detector that is significantly faster than exact geometry tests and is “good enough” for choosing the draw path. The detector may occasionally misclassify, as minor rendering differences are acceptable in exchange for speed.

wenjieshen:
### Progress

I investigated this from the BW tessellation fast-path side.

`main` already has a cheap streamed convexity check: it watches the turn sign of consecutive edges and uses that to decide whether the path can stay on the convex triangle-fan path.

I first tried a bbox-based approximation for likely self-intersection. In plain terms, the idea was:
- flatten the path into emitted edges
- build a small AABB for each edge
- look for overlap between non-adjacent edge boxes, or bucket them into the same spatial region
- if unrelated edges overlap in the same area, treat the path as likely unsafe and fall back

That can catch bow-ties because the two crossing diagonals both pass through the center region, so their edge boxes overlap around the crossing. Even without exact line-line intersection, that is enough to conservatively say “this is probably not safe for one triangle fan”.

I dropped that approach because it regressed FPS in practice.

So the current change does not add a broader bbox/self-intersection pass. Instead, it stays in the same cheap single-pass family as `main`, but makes the existing streamed classifier more conservative by also rejecting:
- extra contours
- excessive x/y direction flips
- repeated collinear backtracking
- obvious cubic loop cases
- cubic paths based on emitted edges more conservatively

So the difference from `main` is:
- `main`: local turn-sign check only
- current patch: same streaming idea, but with several extra low-cost fan-safety checks

PR #4230, opened 2026-03-19
gl/wg: reject unsafe convex tessellation paths

wenjieshen:
## Summary

`main` already uses a turn-sign check to decide whether the BW tessellator can stay on the convex triangle-fan fast path. This patch keeps that approach and tightens it with a few extra cheap checks, instead of adding a broader bbox-based approximate self-intersection pass.

The bbox approach was tested first, but it regressed FPS, so it was dropped.

This change only extends the existing streaming classifier also to reject:
- extra contours
- excessive x/y direction flips
- repeated collinear backtracking
- obvious cubic loop cases

As a result, contours that are unsafe for the fast fan path now fall back more conservatively, while simple convex paths keep the existing fast path.

Increase the FPS by 1-2% in the Lottie Example.  
Increase the FPS by 8-9% in the Stress Example.

Binary size: 
In clang compiler
Main
__TEXT	__DATA	__OBJC	others	dec	hex
825,792 	540672	0	81920	1523712	174000
This branch
__TEXT	__DATA	__OBJC	others	dec	hex
824,524 	540672	0	81920	1523712	174000
-1268 Bytes
Related issue: https://github.com/thorvg/thorvg/issues/4141

### DIFF
--- a/src/common/tvgMath.h
+++ b/src/common/tvgMath.h
@@ -468,6 +468,117 @@ struct Bezier
     static void bounds(BBox& box, const Point& start, const Point& ctrl1, const Point& ctrl2, const Point& end);
 };
 
+static inline bool edgesCross(const Point& p0, const Point& p1, const Point& p2, const Point& p3)
+{
+    auto orientSign = [](const Point& a, const Point& b, const Point& c) {
+        auto value = cross(b - a, c - a);
+        if (zero(value)) return int8_t{0};
+        return (value > 0.0f) ? int8_t{1} : int8_t{-1};
+    };
+
+    auto straddlesLine = [&](const Point& a, const Point& b, const Point& c, const Point& d) {
+        return orientSign(a, b, c) * orientSign(a, b, d) < 0;
+    };
+
+    return straddlesLine(p0, p1, p2, p3) && straddlesLine(p2, p3, p0, p1);
+}
+
+// Conservative triangle-fan safety check
+// The fill is emitted as (v0,v1,v2), (v0,v2,v3), (v0,v3,v4), ...
+// Usage:
+// 1. Setup: default-construct `ConvexProbe probe` for one tessellation pass.
+// 2. Feed one contour at a time: call `nextContour()` on MoveTo, then stream
+//    each contour edge with `addEdge(curr - prev)`, and finish with
+//    `addContourClose(first - prev)` on Close. For cubic segments, callers may
+//    pre-reject looping control polygons with `edgesCross(start, ctrl1, ctrl2, end)`.
+// 3. Read `probe.convex` as the result. `true` means the contour stayed within
+//    the cheap triangle-fan fast-path assumptions; `false` means fall back.
+// Cost: O(1) work per edge and O(1) memory.
+struct ConvexProbe
+{
+    bool convex = true;
+    int8_t winding = 0;
+    Point firstEdge = {};
+    Point prevEdge = {};
+    int8_t prevXDir = 0;
+    int8_t prevYDir = 0;
+    uint8_t xDirChanges = 0;
+    uint8_t yDirChanges = 0;
+    uint8_t reversals = 0;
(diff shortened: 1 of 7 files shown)


wenjieshen:
> @wenjieshen We’d prefer to place the additional math functions in the header so they are only embedded when gpu engines used.

Thanks. I have moved helpers out of the common object file.

src/common/tvgMath.h
```diff
@@ -468,6 +468,122 @@ struct Bezier
...
+    };
+
+    auto straddlesLine = [&](const Point& a, const Point& b, const Point& c, const Point& d) {
+        return orientSign(a, b, c) * orientSign(a, b, d) < 0;
+    };
+
+    return straddlesLine(p0, p1, p2, p3) && straddlesLine(p2, p3, p0, p1);
+}
+
+// Conservative triangle-fan safety check
```
hermet, line 486: @wenjieshen  Could you please elaborate the document with an example of its usage?

hermet:
@wenjieshen 
```
../src/common/tvgMath.h:495:1: warning: multi-line comment [-Wcomment]
  495 | //    |        \
      | ^
[52/53] Installing files.
```

hermet:
Lottie: +6.10%
LottieBigSize: +100%
ManyRects: +2.53%

src/common/tvgMath.h
```diff
@@ -468,6 +468,120 @@ struct Bezier
...
+    };
+
+    void resetContour()
+    {
+        firstEdge = {};
+        prevEdge = {};
+        prevXDir = prevYDir = 0;
+        xDirChanges = yDirChanges = 0;
+        reversals = 0;
+        contourHasEdges = false;
```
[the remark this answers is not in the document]
wenjieshen: You are creating a bug.

src/common/tvgMath.h
hermet: @wenjieshen  example usage sample:
/*!
* \brief Creates a Canvas object.
*
* \code
* Tvg_Canvas *canvas = NULL;
*
* tvg_engine_init(TVG_ENGINE_SW, 4);
* canvas = tvg_swcanvas_create();
*
* //set up the canvas buffer
* uint32_t *buffer = NULL;
* buffer = (uint32_t*) malloc(sizeof(uint32_t) * 100 * 100);
* if (!buffer) return;
*
* tvg_swcanvas_set_target(canvas, buffer, 100, 100, 100, TVG_COLORSPACE_ARGB8888);
*
* //set up paints and add them into the canvas before drawing it
*
* tvg_canvas_destroy(canvas);
* tvg_engine_term(TVG_ENGINE_SW);
* \endcode

wenjieshen: I have updated, thank you.

hermet:
@wenjieshen What we need to do is, see the true binary size. These objects are not stripped and does not show the overall size info. Pointless. Please use the size command. Even SwRaster is not related to your change at all. 
image

Usage Commands
```
$ size <binary>  //recommend
text    data     bss     dec     hex
12345   678      90      13113   0x3339

other options:
$ ls -lh <binary>
$ du -h <binary>
```

src/common/tvgMath.h
```diff
@@ -468,6 +468,121 @@ struct Bezier
     static void bounds(BBox& box, const Point& start, const Point& ctrl1, const Point& ctrl2, const Point& end);
 };
 
+#if defined(THORVG_GL_RASTER_SUPPORT) || defined(THORVG_WG_RASTER_SUPPORT)
```
hermet: We can remove #if condition.

wenjieshen: Thank you. I removed it.

wenjieshen:
> @wenjieshen What we need to do is, see the true binary size. These objects are not stripped and does not show the overall size info. Pointless. Please use the size command. Even SwRaster is not related to your change at all. 

Thank you. I have updated.

PR #4235, opened 2026-03-20
lottie: improve text height alignment correctness

hermet:
Added edge-case handling when a carriage return ('\r') appears at the end of the string. Currently, ThorVG preprocesses input text by remapping carriage return ('\r') and end-of-text ('\3') to line feed ('\n') for modern compatibility.


hermet:
thorvg:
image

expected:
image

PR #4237, opened 2026-03-20
gl_engine: optimize GlCanvas, cache GlProgram globally, doubling the FPS

liungkejin:
Optimizes the GL engine by caching GLProgram instances globally (per-thread) and updates solid batching code to reuse the cached program, aiming to reduce per-renderer shader setup overhead and increase FPS.

Changes:

Introduces a thread_local global mPrograms cache and a GlRenderer::program(RenderTypes) accessor.
Makes shader initialization (initShaders) static and idempotent based on the global cache.
Updates GlSolidBatch to use the global program accessor instead of per-renderer program storage.

### DIFF
--- a/src/renderer/gl_engine/tvgGlRenderer.cpp
+++ b/src/renderer/gl_engine/tvgGlRenderer.cpp
@@ -97,8 +97,6 @@ GlRenderer::~GlRenderer()
 {
     flush();
 
-    ARRAY_FOREACH(p, _programs) delete(*p);
-
     _rendererMtx.lock();
     --_rendererCnt;
     _rendererMtx.unlock();


wenjieshen:
@liungkejin, could you please address Copilot's review? You can resolve the conversation directly if you believe the AI did not provide a valuable opinion. Thank you. 🙏

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -106,6 +107,9 @@ GlRenderer::~GlRenderer()
 
 void GlRenderer::initShaders()
 {
+    if (!mPrograms.empty()) {
```
hermet:
```
    if (!mPrograms.empty()) eturn; // already inited
```

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -38,6 +38,7 @@
 
 static int32_t _rendererCnt = -1;
 static mutex _rendererMtx;
+static thread_local Array<GlProgram*> mPrograms;
```
hermet: `mPrograms` -> `_programs`

src/renderer/gl_engine/tvgGlRenderer.cpp
```diff
@@ -1418,3 +1425,7 @@ GlRenderer* GlRenderer::gen(TVG_UNUSED uint32_t threads, TVG_UNUSED EngineOption
 
     return new GlRenderer;
 }
+
+GlProgram *GlRenderer::program(RenderTypes type) {
```
hermet: convention fix++
```
function()
{
}
```

hermet:
@liungkejin Thanks for your contribution. It would be better to include the PR description in the commit message.

hermet:
@wenjieshen What did you test with them in thorvg.example? I think it needs a multi-canvas usage scenario.

wenjieshen:
> @wenjieshen What did you test with them in thorvg.example? I think it needs a multi-canvas usage scenario.

The crash I mentioned happened on the multi-canvas example.

UNSUPPORTED (log once): POSSIBLE ISSUE: unit 0 GLD_TEXTURE_INDEX_2D is unloadable and bound to sampler type (Float) - using zero texture because texture unloadable
MultiCanvas(4468,0x2062c6c40) malloc: *** error for object 0x852d4d550: pointer being freed was not allocated
MultiCanvas(4468,0x2062c6c40) malloc: *** set a breakpoint in malloc_error_break to debug
zsh: abort      MultiCanvas gl

wenjieshen:
Hi @liungkejin, could you please share your use case and environment so I can begin more precise testing? Thank you.

liungkejin:
> Thanks for the contribution. I reviewed the change and, based on my testing including not limited to the repo: [thorvg.example](https://github.com/thorvg/thorvg.example), I can’t approve it as is. I’m seeing about a 0 - 4% FPS regression on macOS, Windows 11, and Ubuntu, and also a crash during deallocation on macOS.
> 
> ### [P0] Shared thread-local program cache is freed per renderer
> * `_programs` is shared by every GlRenderer on a thread, but `~GlRenderer()`still deletes it unconditionally. If two GL canvases exist on the same thread, destroying either one invalidates the surviving renderer's program pointers; because the array is not cleared, later `preRender()` calls skip `initShaders()` and reuse freed `GlPrograms`.
> 
> ### [P1] Program cache is not keyed by GL context
> * The cache now lives in `thread_local` storage, but `GlProgram` objects are created with `glCreateProgram() `against the current GL context. `target() `can retarget a renderer to a different context without clearing `_programs`, so later draws may reuse program IDs from another context, causing invalid draws, driver fallback, or FPS regressions.

I will add a new test project later and then supplement some benchmarks.
P0 does have an issue, and I will fix it as soon as possible.
Regarding the P1 issue, since the GL context itself is bound to a thread, if multi-threaded usage is required, it will definitely go through the glInit() and glTerm() lifecycle methods. Therefore, I don't think there will be a problem of reusing GlProgram.

wenjieshen:
Hi @liungkejin Thank you for the update and for the work on this. The P1 issue is also my main suspicion for the FPS regression, especially in the Lottie example, but that’s only my current read. 

I’m open to tradeoffs if they bring clear benefits in other use cases, but based on the current results, I don’t think that has been shown yet. 

Still, I appreciate the contribution and the effort to improve this area.🙏

PR #4238, opened 2026-03-20
gl_engine: GlCanvas support setup MSAA samples, increases the FPS (disable MSAA)

liungkejin:


hermet:
This needs to align the API spec first.

PR #4246, opened 2026-03-24
ttf: fix heap-buffer-overflow on truncated UTF-8 sequence

tinyjin:
_codepoints() read continuation bytes unconditionally, causing a 1-byte out-of-bounds read when a multi-byte lead byte appeared at the end of the string.

```
text->text("hello\xF0\x9F");
```

CleanShot 2026-03-24 at 16 07 42@2x

### DIFF
--- a/src/loaders/ttf/tvgTtfLoader.cpp
+++ b/src/loaders/ttf/tvgTtfLoader.cpp
@@ -171,13 +171,13 @@ static size_t _codepoints(const char** utf8)
     if (!(*p & 0x80U)) {
         (*utf8) += 1;
         return *p;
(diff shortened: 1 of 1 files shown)


hermet (changes requested):
1. It needs to properly check the memory boundary, not just whether the value exists.
2. After this PR, users may interpret broken text rendering as a ThorVG issue. Fundamentally, applications need to verify that their input text is valid to ensure proper handling.
3. Ultimately, ThorVG may need to notify users to correct their input data in this edge case.

hermet:
https://github.com/thorvg/thorvg/pull/4249

PR #4248, opened 2026-03-24
lottie: fix offsetting partially degenerated cubics

tinyjin:
Until now, cases where a Bezier curve had `start == ctrl1` or `ctrl2 == end` were incorrectly treated as linear segments. This led to incorrect rendering, for example when offsetting a polystar with outer roundness > 0 and inner roundness == 0. Now such cases handled as proper curves with full cubic behavior.

original PR: https://github.com/thorvg/thorvg/pull/3523
test file : 
[sample1 (2).json](https://github.com/user-attachments/files/26225006/sample1.2.json)

CleanShot 2026-03-25 at 04 58 11@2x

CleanShot 2026-03-25 at 04 57 49@2x

### DIFF
--- a/src/loaders/lottie/tvgLottieModifier.cpp
+++ b/src/loaders/lottie/tvgLottieModifier.cpp
@@ -26,6 +26,12 @@
 /* LottieModifier                                                       */
 /************************************************************************/
 
+static bool _colinear(const Point* p)
+{
+    return tvg::zero(*p - *(p + 1)) && tvg::zero(*(p + 2) - *(p + 3));
+}
+
+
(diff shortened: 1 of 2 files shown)


tinyjin:
@hermet The patch is still valid, so I rebased at the commit (b98363).

The issue comes from when polistar has `outer roundness > 0 && inner roundness == 0`, the engine renders the curve into just straight lines, which is rendering discrepancy.

Also I refactored at commit (49c64c), since the original if-else condition got longer.
(_I will squash the commits, but leave as another commit for reviewing convenience_)

hermet:
@tinyjin If you think if some new addition needs a refactoring, please conduct the refactoring first. Thanks.

Issue #4244, opened 2026-03-24
svg : `<use>` element references fall into an infinite loop


JSUYA:
While testing the W3C SVG test set, I found that a specific SVG fell into an infinite loop.

```html
<svg id="svg-root" width="100%" height="100%"
  viewBox="0 0 480 360" xmlns="http://www.w3.org/2000/svg"
  xmlns:xlink="http://www.w3.org/1999/xlink">
  <!--======================================================================-->
  <!--=  Copyright 2008 World Wide Web Consortium, (Massachusetts          =-->
  <!--=  Institute of Technology, European Research Consortium for         =-->
  <!--=  Informatics and Mathematics (ERCIM), Keio University).            =-->
  <!--=  All Rights Reserved.                                              =-->
  <!--=  See http://www.w3.org/Consortium/Legal/.                          =-->
  <!--======================================================================-->
  <d:SVGTestCase xmlns:d="http://www.w3.org/2000/svg" xmlnsd="http://www.w3.org/2000/02/svg/testsuite/description/"
    template-version="1.4" reviewer="CL" author="Microsoft" status="accepted"
    version="$Revision: 1.3 $" testname="$RCSfile: struct-use-12-f.svg,v $">
    <d:testDescription xmlnsDeactivated="http://www.w3.org/1999/xhtml" hrefDeactivated="http://www.w3.org/TR/SVG11/struct.html#UseElement">
      <p>
        Tests that recursive 'use' instances do not block rendering.
      </p>
      <p>
        Various scenarios that directly and indirectly create circular references via the 'use' tag. A 'g' element is used when 
        structural elements are necessary. None of the 'use' scenarios render anything. 'useLongCycle' tests a chain of recursive 
        'use' instances that eventually cycles back to the first element. In 'useNested' 'use' elements are nested, with the child 
        referring to the parent. In 'useNestedGroup' a 'use' instance references a parent 'g' element. In 'useIndirectNestedGroup' 
        a 'use' instance indirectly references its own parent 'g'. In 'useMultipleIndirectNestedGroup', two 'use' instances 
        reference their parent 'g' elements, and additional 'use' instances refer to these self-referencing 'use' elements. A green 
        'rect' is used to verify that rendering was processed up to that point.
      </p>
    </d:testDescription>
    <d:operatorScript xmlnsDeactivated="http://www.w3.org/1999/xhtml">
      <p>
        Run the test. No interaction required.
      </p>
    </d:operatorScript>
[...]

PR #4250, opened 2026-03-24
svg_loader: fix infinite loop caused by circular use element references

JSUYA:
<use> copies and draws another element by ID
If A references B and B references A, the loader keeps re-queuing both forever. so if the pending queue stops shrinking (everything just cycles), treat them as circular and discard them

related issue : https://github.com/thorvg/thorvg/issues/4244

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -3003,8 +3003,18 @@ static void _cloneNode(SvgNode* from, SvgNode* parent, int depth)
 
 static void _clonePostponedNodes(Inlist<SvgNodeIdPair>* cloneNodes, SvgNode* doc)
 {
+    uint32_t cloneNodesCount = cloneNodes->count;
+    uint32_t postponeCount = 0;
     auto nodeIdPair = cloneNodes->front();
     while (nodeIdPair) {
+        if (postponeCount >= cloneNodesCount) {
+            do {
+                TVGERR("SVG", "Circular use reference detected, discarding '%s'.", nodeIdPair->id ? nodeIdPair->id : "");
+                tvg::free(nodeIdPair->id);
+                tvg::free(nodeIdPair);
+            } while ((nodeIdPair = cloneNodes->front()));
+            break;
+        }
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3292,8 +3292,20 @@ static void _cloneNode(SvgNode* from, SvgNode* parent, int depth)
...
 {
+    uint32_t cloneNodesCount = 0;
+    INLIST_FOREACH((*cloneNodes), tmp) cloneNodesCount++;
+
+    uint32_t postponeCount = 0;
     auto nodeIdPair = cloneNodes->front();
     while (nodeIdPair) {
+        if (postponeCount >= cloneNodesCount) {
+            do {
+                TVGLOG("SVG", "Circular use reference detected, discarding '%s'.", nodeIdPair->id);
```
hermet: I think this is an error case(TVGERR), users must to fix.

JSUYA: I updated code

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3003,8 +3003,20 @@ static void _cloneNode(SvgNode* from, SvgNode* parent, int depth)
 
 static void _clonePostponedNodes(Inlist<SvgNodeIdPair>* cloneNodes, SvgNode* doc)
 {
+    uint32_t cloneNodesCount = 0;
+    INLIST_FOREACH((*cloneNodes), tmp) cloneNodesCount++;
```
hermet: https://github.com/thorvg/thorvg/pull/4263

JSUYA: I updated it to use count of inlist.

hermet:
@JSUYA Please add relevant issue number in the commit msg, in the next time. thanks!

```
    svg_loader: fix infinite loop caused by circular use element references
    
    <use> copies and draws another element by ID
    If A references B and B references A, the loader keeps re-queuing both forever.
    so if the pending queue stops shrinking (everything just cycles), treat them as circular and discard them
    
    issue: #4244
```

Issue #4243, opened 2026-03-23
svg: support blending styles


hermet:
ThorVG already supports these blending modes, so we can properly map the styles to API calls.

**Blending styles:**
```
style="mix-blend-mode:normal"
style="mix-blend-mode:multiply"
style="mix-blend-mode:screen"
style="mix-blend-mode:overlay"
style="mix-blend-mode:darken"
style="mix-blend-mode:lighten"
style="mix-blend-mode:color-dodge"
style="mix-blend-mode:color-burn"
style="mix-blend-mode:hard-light"
style="mix-blend-mode:soft-light"
style="mix-blend-mode:difference"
style="mix-blend-mode:exclusion"
style="mix-blend-mode:hue"
style="mix-blend-mode:saturation"
style="mix-blend-mode:color"
style="mix-blend-mode:luminosity"
```

**Examples:**
```
<circle cx="238.5" cy="227.5" r="227.5" fill="url(#paint30_radial_57_195)" style="mix-blend-mode:soft-light"/>
```

ThorVG APIs:
```
/**
 * @brief Enumeration indicates the method used for blending paint. Please refer to the respective formulas for each method.
 *
 * Notation: S(source paint as the top layer), D(destination as the bottom layer), Sa(source paint alpha), Da(destination alpha)
 *
 * @see Paint::blend()
 *
 * @since 0.15
 */
enum struct BlendMethod : uint8_t
{
    Normal = 0,        ///< Perform the alpha blending(default). S if (Sa == 255), otherwise (Sa * S) + (255 - Sa) * D
    Multiply,          ///< Takes the RGB channel values from 0 to 255 of each pixel in the top layer and multiples them with the values for the corresponding pixel from the bottom layer. (S * D)
    Screen,            ///< The values of the pixels in the two layers are inverted, multiplied, and then inverted again. (S + D) - (S * D)
    Overlay,           ///< Combines Multiply and Screen blend modes. (2 * S * D) if (D < 128), otherwise 255 - 2 * (255 - S) * (255 - D)
    Darken,            ///< Creates a pixel that retains the smallest components of the top and bottom layer pixels. min(S, D)
    Lighten,           ///< Only has the opposite action of Darken Only. max(S, D)
    ColorDodge,        ///< Divides the bottom layer by the inverted top layer. D / (255 - S)
    ColorBurn,         ///< Divides the inverted bottom layer by the top layer, and then inverts the result. 255 - (255 - D) / S
[...]

Issue #4190, opened 2026-03-06
Blend mode inconsistencies


mbasaglia:
Some blend modes seem to have a different behaviour between ThorVG and browsers.

I rendered the same image with lottie-web SVG and ThorVG.
Lottie-web just sets the `mix-blend-mode` CSS attribute to the corresponding value so the result is how the browser itself renders that blend mode.

For thorvg I was using the software renderer and a custom loader (not the thorvg lottie parser), ARGB8888 color space.

All the other blend modes seem to work correctly.

For reference here are the w3c definitions of the affected blend modes:

https://www.w3.org/TR/compositing-1/#blendingsoftlight
https://www.w3.org/TR/compositing-1/#blendingnonseparable

# Examples

This is the JSON I was testing with: [blend-mode-example.json](https://github.com/user-attachments/files/25797366/blend-mode-example.json)

Here are the images I've rendered (first thorvg, second lottie-web). For each blend mode I provide lottie number, CSS value name, and Thorvg enum name.

## `9` / `soft-light` /  `tvg::BlendMode::SoftLight`

<img width="256" height="256" alt="Image" src="https://github.com/user-attachments/assets/b6f92101-68ed-4fa8-88e4-19c0a437d39b" />
<img width="256" height="256" alt="Image" src="https://github.com/user-attachments/assets/36165726-cd49-42c5-858e-e37296635f57" />

## `13` / `saturation` / `tvg::BlendMode::Saturation`

<img width="256" height="256" alt="Image" src="https://github.com/user-attachments/assets/0698360a-4d05-4bef-93cf-986c95f690d3" />
<img width="256" height="256" alt="Image" src="https://github.com/user-attachments/assets/ce8fffdc-32e8-49f3-a9fd-48da82894472" />

## `14` / `color` / `tvg::BlendMode::Color`

<img width="256" height="256" alt="Image" src="https://github.com/user-attachments/assets/239dee76-16e5-40a0-abe8-c43368642f98" />
<img width="256" height="256" alt="Image" src="https://github.com/user-attachments/assets/c03051df-4aa0-4123-af31-a0a8abf8ae19" />

## `15` / `luminosity` / `tvg::BlendMode::Luminosity`

<img width="256" height="256" alt="Image" src="https://github.com/user-attachments/assets/a9d925ab-cd3e-4d9c-b736-47f140671e56" />
[...]

PR #4255, opened 2026-03-25
svg_loader: support blend mode(`mix-blend-mode` in style)

JSUYA:
It supports SVG blend modes. First, a base code(funtions, enums) is added based on the multiply mode. and It will be added later after testing for each mode.

issue : https://github.com/thorvg/thorvg/issues/4243

Test Svg
blend_modes

thorvg
image

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -305,9 +305,6 @@ _PARSE_TAG(FillRule, fillRule, FillRule, fillRuleTags, FillRule::NonZero)
 
 
 /* parse the blend mode applied to an element.
- * Value:   normal | multiply | screen | overlay | darken | lighten | color-dodge | color-burn |
- *          hard-light | soft-light | difference | exclusion | hue | saturation | color | luminosity
- * Initial:    normal
  * https://www.w3.org/TR/compositing-1/#mix-blend-mode
  */
 static constexpr struct


src/loaders/svg/tvgSvgLoaderCommon.h
```diff
@@ -153,6 +153,26 @@ constexpr void operator|=(SvgStrokeFlags& a, const SvgStrokeFlags b)
     a = SvgStrokeFlags(int(a) | int(b));
 }
 
+enum class SvgBlendMode
```
hermet: I believe we don’t need to add these enum classes. You can directly use the ThorVG blend methods.

JSUYA: I thought so at first as well, but I added SvgBlendMode to maintain the consistency of the current SVG loader. I do not believe that Tvg blendmodes will match SVG blendmodes 100% in the future.

I think it is better to store SVG-specific information in all properties of the SVG loader's node tree.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -409,6 +432,7 @@ static Paint* _applyProperty(SvgLoaderData& loaderData, SvgNode* node, Shape* vg
     vg->fillRule(style->fill.fillRule);
     vg->order(!style->paintOrder);
     vg->opacity(style->opacity);
+    if (style->flags & SvgStyleFlags::BlendMode) vg->blend(_toTvgBlendMethod(style->blendMode));
```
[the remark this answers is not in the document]
JSUYA: We can add the svg sample file attached to PR's message to thorvg.example.

JSUYA:
https://github.com/thorvg/thorvg/pull/4255/checks?check_run_id=68540144173
It seems to perceive complexity simply as increasing as the number of `case` statements increases. How can this be ignored? I think this a false positive.

src/loaders/svg/tvgSvgLoaderCommon.h
hermet: >  I do not believe that Tvg blendmodes will match SVG blendmodes 100% in the future.

@JSUYA very negative at your opinion.
1. Both should respect w3c spec. 
2. Without blending option alignment, the SVG loader itself cannot handle blending at all. Please don't make svg loader unnecessarily complicated.

hermet:
> https://github.com/thorvg/thorvg/pull/4255/checks?check_run_id=68540144173 It seems to perceive complexity simply as increasing as the number of `case` statements increases. How can this be ignored? I think this a false positive.

@JSUYA Please use ThorVG blending modes directly.

src/loaders/svg/tvgSvgLoaderCommon.h
JSUYA: ok. I updated code

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -304,6 +304,23 @@ static constexpr struct
...
+ * Value:   normal | multiply
+ * Initial:    normal
+ * https://www.w3.org/TR/compositing-1/#mix-blend-mode
+ */
+static constexpr struct
+{
+    BlendMethod blendMode;
+    const char* tag;
+} blendModeTags[] = {
+    { BlendMethod::Multiply, "multiply" }
```
hermet: @JSUYA let's add all available options together.

JSUYA: I added more options.
image

Some modes produce different results compared to the Chrome. So I wanted to add the structure first.
anyway, blend(hue, saturation, color) option seems to produce different results. The same applies to scenes. Please refer to the tvg code.

```
  // Background rects (amber top, teal bottom)
  auto bg1 = tvg::Shape::gen();
  bg1->appendRect(3, 18, 74, 36);
  bg1->fill(0xFF, 0xB3, 0x00);
  canvas->add(bg1);

  auto bg2 = tvg::Shape::gen();
  bg2->appendRect(3, 54, 74, 36);
  bg2->fill(0x00, 0xAC, 0xC1);
  canvas->add(bg2);

  // Group with hue blend mode
  scene->blend(tvg::BlendMethod::Hue);  // <-- here

  auto c1 = tvg::Shape::gen();
  c1->appendCircle(30, 55, 26, 26);
  c1->fill(0xE9, 0x1E, 0x63);
  scene->add(c1);

  auto c2 = tvg::Shape::gen();
  c2->appendCircle(60, 55, 26, 26);
  c2->fill(0x7C, 0x4D, 0xFF);
  scene->add(c2);

  canvas->add(scene);
```

hermet: @JSUYA Please verify with GL, there is a known HSL blending issue with software blending, you can ignore the blending accuracy now.
image

https://github.com/thorvg/thorvg/issues/4190

hermet: @wenjieshen  Hue result looks different.

THorVG GL  vs Chrome
image

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -304,6 +304,38 @@ static constexpr struct
 _PARSE_TAG(FillRule, fillRule, FillRule, fillRuleTags, FillRule::NonZero)
 
 
+/* parse the blend mode applied to an element.
+ * Value:   normal | multiply | screen | overlay | darken | lighten | color-dodge | color-burn |
```
hermet: "Value" can be skipped. The code already provides sufficient mode information, and the added link can be referenced if needed.

JSUYA: Not all tags declared in SvgLoader have a description, but this should generally be maintained for management purposes regarding the values ​​supported by the tag.
(A tag definition without a description is simply missing.)

- "normal" is not actually a parsed value, but it is a value supported by the current specification. It does not match the code.
- For a simpler, please refer to fillRuleTags.

hermet: @JSUYA We can check. 
-> `blendModeTags[]`  and `https://www.w3.org/TR/compositing-1/#mix-blend-mode`

enough.

src/loaders/svg/tvgSvgSceneBuilder.cpp
```diff
@@ -967,6 +972,7 @@ static Scene* _sceneBuildHelper(SvgParserContext& ctx, const SvgNode* node, cons
         }
     }
     scene->opacity(node->style->opacity);
+    if (node->style->flags & SvgStyleFlags::BlendMode) scene->blend(node->style->blendMode);
 
     auto p = _applyFilter(ctx, scene, node, vBox, svgPath);
     return static_cast<Scene*>(_applyComposition(ctx, p, node, vBox, svgPath));
```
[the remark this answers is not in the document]
JSUYA: I added the commit because it is a valid review.

Issue #2256, opened 2024-05-09
lottie: support multi-threading for the expressions


hermet:
Currently, Expressions are designed not to operate in a Task Scheduler-based Loader. The JerryScript engine that executes Expressions appears to be designed based on global resources, and it seems possible to consider and apply synchronization techniques to attempt improvements.

Nor-s:
I found an example of using JerryScript with pthread-based multithreading.
Applying this approach to ThorVG may allow expressions to be supported in a multithreaded environment.

- relative issue: https://github.com/jerryscript-project/jerryscript/issues/4864
- example: https://github.com/jerryscript-project/jerryscript/blob/9860d66a56ed44f62e1dafb9900e6f2c886a56d3/docs/02.API-REFERENCE.md#external-context-functions

PR #4256, opened 2026-03-25
lottie/expressions: support multi-threading for Lottie expressions

Nor-s:
## Support Lottie Expressions in Multi-Thread Environment

issue: #2256

### Summary

Previously, Lottie Expressions were not supported in a multi-threaded environment.
This patch adds multi-thread support, enabling Lottie Expressions to work
consistently regardless of the threading model in use.

### How

Leveraged JerryScript's Port API to achieve multi-thread support.

- JerryScript provides a **Port API** for extending engine behavior.
- In particular, the **Context-related Port API** is designed for running multiple
  independent JS engine instances — **Allows the user to provide external buffer for isolated engine contexts, so that user can configure the heap size at runtime and run multiple JS applications simultaneously.**
    - `jerry_port_context_alloc`: Allocate a new context for the engine.
    - `jerry_port_context_get`: Get the currently active context of the engine.
    - `jerry_port_context_free`: Free the currently used context.
- For more details, please refer to the links in the **Reference** section below.

#### Overview

```mermaid
sequenceDiagram
    participant Main as Main Thread
    participant Mgr as LottieExpressionsMgr
    participant W1 as Worker Thread 1
    participant W2 as Worker Thread 2

    Main->>Mgr: LottieBuilder() → ref()

    W1->>Mgr: expressions()
    Mgr->>W1: new LottieExpressions (jerry_init → per-thread ctx)
    W2->>Mgr: expressions()
    Mgr->>W2: new LottieExpressions (jerry_init → per-thread ctx)

    W1->>W1: evaluate() (TLS context already set by jerry_init)
    W2->>W2: evaluate() (TLS context already set by jerry_init)

    Main->>Mgr: ~LottieBuilder() → unref()

    loop Cross-thread cleanup
        Mgr->>Mgr: jerry_port_context_set(W1.ctx) → delete
        Mgr->>Mgr: jerry_port_context_set(W2.ctx) → delete
    end
    Mgr->>Mgr: _instances.clear()
```

#### Key Changes

- **`LottieExpressionsMgr`**: New Manager Class (following `tvgSwMemPool` pattern)  that provides per-thread LottieExpressions instances. Internally maintains an `Array<LottieExpressionsSlot>`, where each slot maps a `thread::id` to its own `LottieExpressions` and `jerry_context_t`. `LottieBuilder` call `ref()` on construction and `unref()` on destruction; when the last `LottieBuilder` is destroyed, all instances are cleaned up.
  
- **`jerry_port_context_set()`**: New ThorVG-specific Port API added to JerryScript. This is NOT part of the original JerryScript API. It allows the main thread to activate a worker thread's JS context before releasing its resources, enabling safe cross-thread destruction of `LottieExpressions`.

### Result (debugoptimized, optimization=3)

| Single Thread  | 4 Thread  |
|-|-|
|image| image|

> **MacOS, M5 Pro, 48GB**

**Performance (4 Threads): ~fps: ~2.3x+~**
**Binary Size: +1,310**

> The memory previously in the bss section is now dynamically allocated Per Thread: 
> image
> image

### Reference

- [JerryScript issue](https://github.com/jerryscript-project/jerryscript/issues/4864)
- [JerryScript V2 Example](https://github.com/jerryscript-project/jerryscript/blob/v2.4.0/docs/02.API-REFERENCE.md#external-context-functions)
- [JerryScript V3 ExternalContext API](https://github.com/jerryscript-project/jerryscript/blob/v3.0.0/docs/05.PORT-API.md#external-context)
- [JerryScript V3 jerry-port-context.c](https://github.com/jerryscript-project/jerryscript/blob/v3.0.0/jerry-port/common/jerry-port-context.c)


Nor-s:
### Issue

| main | multithread|
|-|-|
|Screenshot 2026-03-26 at 4 20 54 PM |  Screenshot 2026-03-26 at 8 46 01 PM|




Main Branch, certain expressions may take unexpectedly long to evaluate, and dispatch failures can occur. These issues appear more frequently in this patch.

- fixed: https://github.com/thorvg/thorvg/pull/4262

hermet:
@tinyjin This needs the web-test verification. Thanks.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1616,6 +1616,7 @@ bool LottieBuilder::update(LottieComposition* comp, float frameNo)
         //tweening is not necessary.
         if (equal(frameNo, tween.frameNo)) offTween();
     }
+    exps = LottieExpressionsMgr::expressions();
```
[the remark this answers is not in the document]
Nor-s: `exps` must be refreshed every frame because we can't predict which worker thread will evaluate the expressions — the thread may change between frames, so we need to look it up each time.

src/loaders/lottie/tvgLottieExpressionsMgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+void LottieExpressionsMgr::unref()
+{
+    if (--_refCnt == 0) {
+        for (auto& slot : _slots) {
+            jerry_port_context_set(slot.ctx);
+            delete(slot.exps);
+        }
+        _slots.clear();
+    }
```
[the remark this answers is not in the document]
Nor-s: `refCnt` is only modified on the main thread, and `unref()` is called from `LottieLoader::done()` after synchronization, so a lock shouldn't be necessary here.

src/loaders/lottie/tvgLottieExpressionsMgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+    slot.exps = new LottieExpressions();
+    slot.ctx = jerry_port_context_get();
+    return slot.exps;
+}
+
+
+void LottieExpressionsMgr::ref()
+{
+    ++_refCnt;
+}
```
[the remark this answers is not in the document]
Nor-s: `LottieExpressionsMgr::ref()` runs on the main thread (`Picture::load` → `LoadMgr::loader` → `_findByPath` → `LottieLoader::LottieLoader()`), and `unref()` also runs on the main thread (`~LottieLoader` → `LottieLoader::done` → `~LottieBuilder` → `LottieExpressionsMgr::unref()`). 
Since both are confined to the main thread, synchronization on `_refCnt` should be unnecessary.

hermet:
@Nor-s This optimization could be only valid when the ThorVG thread option (-Dthreads=true / THORVG_THREAD_SUPPORT) is enabled. Could you please tweak it to apply only when this option is turned on? And then let's see the binary size.

Nor-s:
> @Nor-s This optimization is only valid when the ThorVG thread option (-Dthreads=true / THORVG_THREAD_SUPPORT) is enabled. Could you please try to tweak it to apply only when this option is turned on? And then let's see the binary size.

Sure, will do. Thank you.

tinyjin:
@hermet @Nor-s 
~WASM size : 1kb-- (Comparing to main)~

The thread option is disabled in the web, no regression(FPS drop/memory usage) found yet.

src/loaders/lottie/tvgLottieExpressionsMgr.cpp
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (c) 2024 - 2026 ThorVG project. All rights reserved.
```
tinyjin: Copyright (c) 2026 ThorVG project. All rights reserved.

src/loaders/lottie/tvgLottieExpressionsMgr.h
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) 2024 - 2026 ThorVG project. All rights reserved.
```
tinyjin: Copyright (c) 2026 ThorVG project. All rights reserved.

src/loaders/lottie/tvgLottieExpressionsMgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "tvgLottieModel.h"
+#include "tvgLottieExpressionsMgr.h"
+#include "tvgLottieExpressions.h"
+
+#ifdef THORVG_LOTTIE_EXPRESSIONS_SUPPORT
+
+#include "jerryscript-port.h"
+
+#include <thread>
```
tinyjin: Expressions can run in both dual mode (threaded, non-threaded)?
Doesn't this need?

#ifdef THORVG_THREAD_SUPPORT
#include <thread>
#endif

src/loaders/lottie/tvgLottieExpressionsMgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+/************************************************************************/
+
+struct LottieExpressionsSlot
+{
+    LottieExpressions* exps;
+    jerry_context_t* ctx;
+    thread::id tid;
+};
+
+static Array<LottieExpressionsSlot> _slots;
```
tinyjin: Expressions domain has own custom slot impl. It looks little disheveled.

src/loaders/lottie/jerryscript/jerry-core/include/jerryscript-port.h
```diff
@@ -117,6 +117,17 @@ struct jerry_context_t *jerry_port_context_get (void);
...
+/**
+ * Set the active context for the current thread.
+ *
+ * Note: This is NOT part of the original JerryScript API.
+ * Added by ThorVG so that the main thread can activate a worker thread's
+ * context before releasing its TLS resources (jerry_cleanup).
+ *
+ * @param context_p: pointer to the context to activate
+ */
+void jerry_port_context_set (struct jerry_context_t *context_p);
```
tinyjin: I am not sure adding a new ThorVG-specific Port API to jerryscript source code is fine. Modifying vendor sourcecode could be sensitive in both area (maintainability & license). I'd hear @hermet 's voice too.

src/loaders/lottie/tvgLottieExpressionsMgr.cpp
tinyjin: Can't put `jerry_context_t* ctx;` in LottieExpressions, and manage there?

src/loaders/lottie/tvgLottieExpressionsMgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+    slot.tid = tid;
+    slot.exps = new LottieExpressions();
+    slot.ctx = jerry_port_context_get();
+    return slot.exps;
+}
+
+
+void LottieExpressionsMgr::ref()
+{
+    ++_refCnt;
```
tinyjin: doesn't need `ScopedLock(_key)`?

tinyjin (commented):
Thanks. Could you check comments?

src/loaders/lottie/tvgLottieExpressionsMgr.cpp
Nor-s: Thanks. I revised it to work in dual mode for both threaded and non-threaded builds.

In the non-threaded case, it keeps the previous logic, and the binary size is smaller than before.

Binary size: -26 bytes

src/loaders/lottie/tvgLottieExpressionsMgr.cpp
Nor-s: added `ScopedLock` for `unref`, `ref`.  Thank you.

src/loaders/lottie/tvgLottieExpressionsMgr.h
Nor-s: updated copyright, Thank you.

src/loaders/lottie/tvgLottieExpressionsMgr.cpp
Nor-s: > Expressions domain has own custom slot impl. It looks little disheveled.

I'll update the naming. Thank you.

Nor-s: > Can't put `jerry_context_t* ctx;` in LottieExpressions, and manage there?

`jerry_context_t` is only used in `thread-enabled` builds. If we manage it in `LottieExpressions`, we'd need to add `THORVG_THREAD_SUPPORT` guards there, which would clutter the code unnecessarily.

Since this is thread-related code, I think it's better to keep it managed in `tvgLottieExpressionsMgr.cpp`.

src/loaders/lottie/jerryscript/jerry-core/include/jerryscript-port.h
Nor-s: > I am not sure adding a new ThorVG-specific Port API to jerryscript source code is fine. Modifying vendor sourcecode could be sensitive in both area (maintainability & license). I'd heard @hermet 's voice too.

I agree **`jerryscript-port.h`** is where JerryScript's official API lives, so adding ThorVG-specific functions there seems unnecessary.

However, the context `set/get` mechanism itself is essential for running ThorVG's Expressions in a multi-threaded environment. Here's the flow:

1. When `LottieExpressions` is first created, a context is stored in `tls_context_p` (in `jerry-port-context.cpp`, Worker Thread).
2. While `LottieExpressions` evaluate on a worker thread, JerryScript operates through this `tls_context_p`.
3. When all `LottieBuilder` instances are destroyed, `LottieExpressionsMgr::unref` can then set the context back and properly release it (in Main Thread).
    -  But from the main thread, `tls_context_p` is not accessible. So `tvgLottieExpressionsMgr.cpp` needs to get and set the context information.


As a compromise, I think we can place `tvg_jerry_context_set()` in `jerry-port-context.cpp` — which is the user-customizable port layer, not the official JerryScript API. 

Then `tvgLottieExpressionsMgr.cpp` can simply `extern` this function and use it.

Nor-s:
### BinarySize

**threads=false**: diff: -26

**threads=true** : diff: +1372

hermet:
> @hermet @Nor-s ~WASM size : 1kb-- (Comparing to main)~

@tinyjin you might wanna compae the binary size with `wc -h` or `wasm-objdump`

ex)
**$ wc thorvg.wasm**
  7228  95928 **986803** thorvg.wasm

**$ wasm-objdump thorvg.wasm  -h**
thorvg.wasm:    file format wasm 0x1
Sections:
     Type start=0x0000000b end=0x000004b2 (size=0x000004a7) count: 154
   Import start=0x000004b5 end=0x000014e0 (size=0x0000102b) count: 155
 Function start=0x000014e3 end=0x00002058 (size=0x00000b75) count: 2884
    Table start=0x0000205a end=0x00002061 (size=0x00000007) count: 1
   Memory start=0x00002063 end=0x0000206a (size=0x00000007) count: 1
   Global start=0x0000206c end=0x00002079 (size=0x0000000d) count: 2
   Export start=0x0000207c end=0x000024ef (size=0x00000473) count: 43
     Elem start=0x000024f2 end=0x00002d4f (size=0x0000085d) count: 1
     Code start=0x00002d53 end=0x000d3707 (size=0x000d09b4) count: 2884
     Data start=0x000d370b end=0x000f0e1c (size=0x0001d711) count: 2
   Custom start=0x000f0e1f end=**0x000f0eb3** (size=0x00000094) "target_features"

hermet:
build breaks, tested on the windows (msys2)
```
herme@neoni-pc UCRT64 /d/Projects/thorvg
$ ninja -C build install
ninja: Entering directory `build'
[243/244] Linking target src/libthorvg-1.dll
FAILED: [code=1] src/libthorvg-1.dll
"c++" @src/libthorvg-1.dll.rsp
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: src/libthorvg-1.dll.p/loaders_lottie_jerryscript_jerry-core_api_jerryscript.cpp.obj: in function `jerry_init':
D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp:82:(.text+0xf): undefined reference to `jerry_port_context_alloc'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp:86:(.text+0x1a): undefined reference to `jerry_port_context_get'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp:100:(.text+0x5c): undefined reference to `jerry_port_context_get'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: src/libthorvg-1.dll.p/loaders_lottie_jerryscript_jerry-core_api_jerryscript.cpp.obj: in function `jerry_cleanup':
D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp:114:(.text+0x8a): undefined reference to `jerry_port_context_get'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp:127:(.text+0xd1): undefined reference to `jerry_port_context_get'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp:143:(.text+0x128): undefined reference to `jerry_port_context_free'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: src/libthorvg-1.dll.p/loaders_lottie_jerryscript_jerry-core_ecma_base_ecma-gc.cpp.obj: in function `ecma_gc_set_object_visited':
D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/ecma/base/ecma-gc.cpp:92:(.text+0xd0): undefined reference to `jerry_port_context_get'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/ecma/base/ecma-gc.cpp:92:(.text+0x19b): undefined reference to `jerry_port_context_get'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/ecma/base/ecma-gc.cpp:92:(.text+0x2a5): undefined reference to `jerry_port_context_get'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/ecma/base/ecma-gc.cpp:92:(.text+0x2ef): undefined reference to `jerry_port_context_get'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/ecma/base/ecma-gc.cpp:94:(.text+0x311): undefined reference to `jerry_port_context_get'
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: src/libthorvg-1.dll.p/loaders_lottie_jerryscript_jerry-core_ecma_base_ecma-gc.cpp.obj:D:\Projects\thorvg\build/../src/loaders/lottie/jerryscript/jerry-core/ecma/base/ecma-gc.cpp:98: more undefined references to `jerry_port_context_get' follow
collect2.exe: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```

hermet:
**fps: +7%**

@Nor-s Could you please double-check the fps (diff rates) with the next? Thanks.
1. thorvg + main + LottieExpressions(0 thread)  
2. thorvg + this PR + LottieExpressions(4 threads)

src/loaders/lottie/jerryscript/jerry-core/include/jerryscript-port.h
hermet: breaks on the windows (MSYS2)

image

That header looks necessary. extern is not really good idea.

tinyjin:
@hermet @Nor-s Here is detailed WASM size comparison:

- `.wasm` is **-781 B (-0.09%)** little smaller than `main`.
- Largest contributors: `Data` (-292 B) section.
- JS glue file is same.

### Overall File Size
  | File | `multithreading` | `main` | Diff |
  |---|---|---|---|
  | `thorvg.wasm` | 863,188 B (843 KB) | 863,969 B (844 KB) | **-781 B (-0.09%)** |
  | `thorvg.js` | 189,569 B | 189,569 B | **0 B** |

### `wasm-objdump -h`
| Section | `multithreading` | `main` | Diff |
|---|---|---|---|
| Type | 1,244 B | 1,244 B | 0 B |
| Import | 1,040 B | 1,040 B | 0 B |
| Function | 2,598 B | 2,602 B | **-4 B** |
| Table | 5 B | 5 B | 0 B |
| Memory | 7 B | 7 B | 0 B |
| Global | 13 B | 13 B | 0 B |
| Export | 996 B | 996 B | 0 B |
| Elem | 2,169 B | 2,171 B | **-2 B** |
| DataCount | 2 B | 2 B | 0 B |
| Code | 734,127 B | 734,162 B | **-35 B** |
| Data | 120,948 B | 121,240 B | **-292 B** |
| **Total** | **863,188 B** | **863,969 B** | **-781 B** |

Nor-s:
@hermet 

I’ve identified some overhead while testing across different environments. 

When `-Dthreads=false`, there is no noticeable difference compared to main.

However, with `-Dthreads=true`, there is a clear overhead when using a Thread Local Storage. 

 **On macOS, Windows, and WSL,** (main vs this pr)
- With single thread, this results in roughly a 20% drop. 
- With 4 threads, performance improves to about 1.8x. 


**In Windows/MSYS2 environments,** (main vs this pr)
-  the overhead from expression thread execution is more pronounced
-  On my machine, using 4 threads actually led to around a 20% performance decrease. 


Through profiling, I identified performance issues in the following functions, and I believe the degradation is caused by repeated access to Thread Local Storage (TLS).
(In particular, the performance of thread_local seems to be worse in MSYS2 compared to other environments.)

- `jerry_port_context_get`
- `jmem_decompress_pointer`
- `jmem_heap_find_prev`

Currently, I have implemented an experimental setup using a single global JerryScript engine with mutex-based locking across multiple threads. In this configuration, I observed performance similar to single-threaded execution even in a multi-threaded environment, and consistent performance on MSYS2 as well. (The performance gain is nothing, but it is expected to provide greater consistency in a multi-threaded environment.)

If I cannot find a solution related to TLS, I plan to replace the code with this approach.

Thank you.

hermet:
@Nor-s Eventually, we need this information. Could you please address the ?% for each cases?

**1. thread option enabled:**
- thorvg(main) + LottieExpressions(0 thread) vs  thorvg(PR) + LottieExpressions(4 threads) -> %fps diff?
- thorvg(main) + LottieExpressions(0 thread) vs  thorvg(PR) + LottieExpressions(0 threads) -> %fps diff?

**2. thread option disabled:**
- thorvg(main) + LottieExpressions(0 thread) vs  thorvg(PR) + LottieExpressions(0 threads) -> %fps diff?

Nor-s:
> @Nor-s Eventually, we need this information. Could you please address the ?% for each cases?
> 
> **1. thread option enabled:** thorvg + main + LottieExpressions(0 thread) -> ?% fps diff thorvg + PR + LottieExpressions(4 threads) -> ?% fps diff
> 
> **2. thread option disabled:** thorvg + main + LottieExpressions(0 thread) -> ?% fps diff thorvg + PR + LottieExpressions(4 threads) -> ?% fps diff


---

#### Mac

**1. thread option enabled** 
- thorvg(main) + LottieExpressions(0 thread) vs thorvg(PR) + LottieExpressions(4 threads) -> +91.36% 
- thorvg(main) + LottieExpressions(0 thread) vs thorvg(PR) + LottieExpressions(0 threads) -> **-26.79%** 

**2. thread option disabled**
- thorvg(main) + LottieExpressions(0 thread) vs thorvg(PR) + LottieExpressions(0 threads) -> ignore


---


#### Windows(MSVC)

**1. thread option enabled**
- thorvg(main) + LottieExpressions(0 thread) vs thorvg(PR) + LottieExpressions(4 threads) -> +143.89%
- thorvg(main) + LottieExpressions(0 thread) vs thorvg(PR) + LottieExpressions(0 threads) -> **-6.11%**

**2. thread option disabled**
- thorvg(main) + LottieExpressions(0 thread) vs thorvg(PR) + LottieExpressions(0 threads) -> ignore

---

#### Windows(MSYS2)

**1. thread option enabled**
- thorvg(main) + LottieExpressions(0 thread) vs thorvg(PR) + LottieExpressions(4 threads) -> +27.95%
- thorvg(main) + LottieExpressions(0 thread) vs thorvg(PR) + LottieExpressions(0 threads) -> **-51.55%**

**2. thread option disabled**
- thorvg(main) + LottieExpressions(0 thread) vs thorvg(PR) + LottieExpressions(0 threads) -> ignore

hermet:
> * thorvg(main) + LottieExpressions(0 thread) vs thorvg(PR) + LottieExpressions(0 threads) -> -26.79%

@Nor-s Let's try tweaking this regression!

Nor-s:
### Manual measurement

- Mac M5pro, 48GB
- Ryzen 7600, 64GB

#### Mac

**1. thread option enabled** 
- main vs PR LottieExpressions(4 threads) -> +115% 
- main vs PR LottieExpressions(0 threads) -> **-7.23%** 

**2. thread option disabled**
- main vs PR LottieExpressions(0 threads) -> ignore


---


#### Windows(MSVC, DLL)

**1. thread option enabled** 
- main vs PR LottieExpressions(4 threads) -> +136.62% 
- main vs PR LottieExpressions(0 threads) -> **-3.94%** 

**2. thread option disabled**
- main vs PR LottieExpressions(0 threads) -> ignore


---

#### Windows(MSYS2)

**1. thread option enabled** 
- main vs PR LottieExpressions(4 threads) -> +113.23% 
- main vs PR LottieExpressions(0 threads) -> **-10.9%** 

**2. thread option disabled**
- main vs PR LottieExpressions(0 threads) -> ignore

---

#### WSL2, ubuntu

**1. thread option enabled** 
- main vs PR LottieExpressions(4 threads) -> +39.72% 
- main vs PR LottieExpressions(0 threads) -> ignore

**2. thread option disabled**
- main vs PR LottieExpressions(0 threads) -> ignore


### Auto Test

| | [threads=false](https://github.com/Nor-s/perfaction/actions/runs/24377242952) | [threads=true,threadCnt=0](https://github.com/Nor-s/perfaction/actions/runs/24355844932) | [threads=true, threadCnt=4](https://github.com/Nor-s/perfaction/actions/runs/24355776052) |
| - | - | - | -|
| ubuntu |+0.88% | -1.07% | +73.71%|
| mac | +0.75% | -7.17% | +76.95%|
| msvc |+0.06% |-6.21% |+86.74% |
| msys2 | +0.57%| -12.87%| +54.93%|

Nor-s:
## Changed

### 1. Replace `thread_local` with platform-native TLS

image


MinGW does not support native `thread_local` — it falls back to emulated TLS (`__emutls`), which internally uses `pthread_getspecific` and `pthread_spin_lock` as shown above. This adds significant overhead on every TLS access. (thread=0: -50% fps)

Changed all `thread_local` to `__thread` (GCC/Clang) and `__declspec(thread)` (MSVC),  to ensure native TLS on all platforms.


### 2. Remove `jerry_port_context_get()` from `JERRY_CONTEXT` macro
    
image

`jerry_port_context_get()` was defined in a separate translation unit, so the compiler could never inline it. Every `JERRY_CONTEXT` access resulted in a full function call + TLS lookup, , which became the single largest bottleneck in profiling.
    
Replaced with a direct extern `__thread`, `__declspec(thread)` variable reference so the compiler can access TLS without a function call.

### 3. Add `JERRY_DEFINE_CURRENT_CONTEXT()` for tls caching

 Even with native TLS, repeated `__thread` variable access has overhead since each access goes through `_tlv_get_addr` (macOS) or similar platform mechanisms. JerryScript's `JERRY_CONTEXT` macro is called hundreds of times per frame across many functions.

Adopted the approach from [jerryscript-project/jerryscript#2470](https://github.com/jerryscript-project/jerryscript/pull/2470): cache the context pointer into a local variable at the entry of hot functions, so that TLS is read once per function instead of once per `JERRY_CONTEXT` access. This yielded an additional ~5% improvement in local benchmarks.

tinyjin:
@Nor-s 

## WASM Size Report (Since latest update)

  - `.wasm` is **-37 B (-0.00%)** slightly smaller than `main`.
  - Largest contributor: `Code` (-37 B) section.
  - JS glue file is same.

  ### Overall File Size

  | File | `multithreading` | `main` | Diff |
  |------|-------------------|--------|------|
  | `thorvg.wasm` | 861,039 B (841 KB) | 861,076 B (841 KB) | **-37 B (-0.00%)** |
  | `thorvg.js` | 189,569 B | 189,569 B | **0 B** |

  ### wasm-objdump -h

  | Section | `multithreading` | `main` | Diff |
  |---------|-------------------|--------|------|
  | Type | 1,228 B | 1,228 B | 0 B |
  | Import | 1,040 B | 1,040 B | 0 B |
  | Function | 2,617 B | 2,617 B | 0 B |
  | Table | 5 B | 5 B | 0 B |
  | Memory | 7 B | 7 B | 0 B |
  | Global | 13 B | 13 B | 0 B |
  | Export | 996 B | 996 B | 0 B |
  | Elem | 2,173 B | 2,173 B | 0 B |
  | DataCount | 2 B | 2 B | 0 B |
  | Code | 734,236 B | 734,273 B | **-37 B** |
  | Data | 118,683 B | 118,683 B | 0 B |
  | **Total** | **861,039 B** | **861,076 B** | **-37 B** |

tinyjin (approved):

src/loaders/lottie/tvgLottieExpressions.cpp
```diff
@@ -1524,10 +1521,12 @@ LottieExpressions::LottieExpressions()
     jerry_init(JERRY_INIT_EMPTY);
 
     //build magic string arrays from the magic pool
-    auto p = _magicPool;
-    for (uint32_t i = 0; i < MAGIC_STRING_COUNT; ++i) {
-        _magicStrings[i] = (const jerry_char_t*)p;
-        p += _magicLengths[i];
+    if (_magicStrings[0] == nullptr) {
```
hermet, line 1524: `if (!_magicStrings[0])`

hermet (commented):
@Nor-s  Could you please consider unifying Expressions and ExpressionsMgr?

ref)
```
    // global objects, attributes, and methods per local thread instance
    struct LocalInstance
    {
        jerry_value_t global;
        jerry_value_t comp;
        jerry_value_t thisComp;
        jerry_value_t thisLayer;
        jerry_value_t thisProperty;
    };
    Array<LocalInstance> instances;
```

Nor-s:
@hermet 

Thank you for the review.

The ExpressionsMgr logic has been moved into LottieExpressions.

Lifecycle management is still handled in the same way as before. (through `instance()` and `retrieve()`, LottieExpressions is singleton)

This change also adds a `LocalInstance` parameter to the build helper functions.

src/loaders/lottie/tvgLottieExpressions.h
```diff
@@ -147,47 +155,57 @@ struct LottieExpressions
...
 
-    //singleton (no thread safety)
-    static LottieExpressions* instance();
-    static void retrieve(LottieExpressions* instance);
-
 private:
     LottieExpressions();
     ~LottieExpressions();
 
+    struct LocalInstance
```
hermet: -> struct Instance;

src/loaders/lottie/tvgLottieExpressions.h
```diff
@@ -147,47 +155,57 @@ struct LottieExpressions
...
+        jerry_value_t thisComp;
+        jerry_value_t thisLayer;
+        jerry_value_t thisProperty;
+#ifdef THORVG_THREAD_SUPPORT
+        jerry_context_t* ctx;
+        thread::id tid;
+#endif
+    };
+
+    LocalInstance& local();
```
hermet: local() -> instance()

src/loaders/lottie/tvgLottieExpressions.cpp
```diff
@@ -1501,76 +1509,119 @@ jerry_value_t LottieExpressions::evaluate(float frameNo, LottieExpression* exp)
...
+}
+
+
+LottieExpressions::LocalInstance& LottieExpressions::local()
+{
+#ifdef THORVG_THREAD_SUPPORT
+    auto tid = this_thread::get_id();
+    ScopedLock lock(_lockKey);
+
+    for (auto inst : instances) {
```
hermet: Have you tried ?

-> `thread_local LocalInstance instance;`

I assume what only we need the code look like this:
```
LottieExpressions::Instance& LottieExpressions::instance()
{
    return instance;
}
```

hermet (commented):
@Nor-s + one more feedback. Thanks.

hermet:
@Nor-s Tested with LottieExpressions. The result seems wrong to me. Any opinion?
main + threads(0) = 61fps
PR + threads(0)  = 56fps
PR + threads(4) = 64fps

Nor-s:
> @Nor-s Tested with LottieExpressions. The result seems wrong to me. Any opinion? main + threads(0) = 61fps PR + threads(0) = 56fps PR + threads(4) = 64fps

@hermet  Could you share what environment you tested this in?

hermet:
> > @Nor-s Tested with LottieExpressions. The result seems wrong to me. Any opinion? main + threads(0) = 61fps PR + threads(0) = 56fps PR + threads(4) = 64fps
> 
> Could you share what environment you tested this in?

@Nor-s wsl!

> > @Nor-s Tested with LottieExpressions. The result seems wrong to me. Any opinion? main + threads(0) = 61fps PR + threads(0) = 56fps PR + threads(4) = 64fps
> 
> @hermet Could you share what environment you tested this in?

@Nor-s my fault, Tested again here:
```
main (thread=on / number = 0) = 62fps
main (thread=off) = 65 fps

PR (thread=on / number = 0) = 55 fps
PR (thread=on / number = 4) = 121 fps
PR (thread=off) = 65fps
```

src/loaders/lottie/tvgLottieExpressions.h
Nor-s: @hermet 

How about 
- `LocalInstance` → `Context`
- `local()` → `context()`

`instance()` is already used by `LottieExpressions::instance`.

src/loaders/lottie/tvgLottieExpressions.cpp
Nor-s: Initially, due to lifecycle issues, I used the `tid` to identify which thread it came from. 
(At destructor time, thread-local storage isn't accessible, so memory can't be freed.)

But if we go with an approach that doesn't use pointers, it should be feasible.

I'll try. Thanks for the review.

Nor-s:
@hermet 

I did try `thread_local LocalInstance instance;`
but ran into a problem:
* LocalInstance (now Context) must be cleared from the main thread, and thread_local storage can't be cleared across threads.
* Working around this needs additional code, which grows the binary (2374 → 2406 bytes) and adds complexity.

So I just applied the naming changes for now. Thanks!

hermet (approved):

PR #4259, opened 2026-03-27
build: skip pthread lookup when targeting emscripten

tinyjin:
Preventing WASM build error from pthread link error.

```
src/meson.build:61:43: ERROR: C++ shared or static library 'pthread' not found
```


tinyjin:
@hermet WASM already disables threads option, maybe `get_option('threads')` condition might be better than emscripten check. I added condition similarly just like original ones for consistency.

PR #4262, opened 2026-03-27
lottie/expressions: fix memory leak

Nor-s:
## Fix Memory Leak in `jerry_object_get_index`

The return value of `jerry_object_get_index` must be freed with `jerry_value_free` 
when it is no longer needed.

image

Without proper cleanup, memory leaks may accumulate over time, causing exceptions 
in some expressions during long-running Expressions animations — eventually leading 
them to be disabled.

### Reference

- [JerryScript API – jerry_object_get_index](https://github.com/jerryscript-project/jerryscript/blob/master/docs/02.API-REFERENCE.md#jerry_object_get_index)

### DIFF
--- a/src/loaders/lottie/tvgLottieExpressions.cpp
+++ b/src/loaders/lottie/tvgLottieExpressions.cpp
@@ -1076,8 +1076,12 @@ static jerry_value_t _key(const jerry_call_info_t* info, const jerry_value_t arg
     if (exp->property->type == LottieProperty::Type::Float) {
         jerry_object_set_index(obj, 0, value);
     } else if (exp->property->type == LottieProperty::Type::Scalar || exp->property->type == LottieProperty::Type::Vector) {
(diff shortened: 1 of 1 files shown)


Nor-s:
Maybe, We need also free 

- jerry_object_set_index: https://github.com/jerryscript-project/jerryscript/blob/master/docs/02.API-REFERENCE.md#jerry_object_set_index
- jerry_object_set_sz : https://github.com/jerryscript-project/jerryscript/blob/master/docs/02.API-REFERENCE.md#jerry_object_set_sz

However, if we always assume a valid case, `jerry_value_free` may not be necessary.



### JerryScript's API Example

```
{
  jerry_value_t object;
  jerry_value_t value_to_set;

  ... // create or copy object and value to set

  jerry_value_t ret_val = jerry_object_set_index (object, 5, value_to_set);

  ...

  jerry_value_free (value_to_set);
  jerry_value_free (ret_val);
  jerry_value_free (object);
}
```

PR #4277, opened 2026-03-31
gl_engine: cache gradient ramps in a reusable atlas

wenjieshen:
## Summary
- cache GL gradient ramps in a reusable atlas so gradient draws can sample uploaded ramps instead of rebuilding stop uploads on each draw
- track cached image texture ownership with texture stamps and invalidate both texture and gradient caches on GL context changes
- keep the GL renderer prepare cleanup limited to state updates shared by both image and shape paths

## Why
- reduce repeated GPU uploads for recurring gradients
- keep the renderer-side cache bookkeeping aligned with the existing shared image texture path without restating that earlier change as the headline


wenjieshen:
Unexpectedly, baking gradient ramp textures causes a noticeable FPS drop in scenes with a certain number of gradient fills. I tried three approaches to manage the textures:

Lazy texture creation
Deferred texture upload
Texture ramp atlases

So far, I haven’t found a better optimization, so I’d like to close this PR.

PR #4282, opened 2026-04-01
infra: fix cross build with opengl/es, link libGLESv3.so / OpenGL/ES framework directly

liungkejin:
Currently, the gl_engine of thorvg relies on dlopen/dlsym, which cannot run on the iOS system. Even on Android and macOS, runtime crashes occur after I compile the static library and link it into my application.

Therefore, for better compatibility, I believe it is necessary to link the system's GL/ES dynamic libraries. I have modified the cross file and gl_engine/meson.build. Meanwhile, I have added ohos_aarch64.txt to support cross-platform compilation for OpenHarmonyOS.

* Android

```
[built-in options]
c_args = ['-DANDROID=1', '-D__ANDROID__=1']
cpp_args = ['-DANDROID=1', '-D__ANDROID__=1']
cpp_link_args = ['-lGLESv3', '-Wl,-z,max-page-size=16384', '-Wl,-z,common-page-size=16384']
```

Added the `ANDROID` and `__ANDROID__` macro definitions
`-lGLESv3` links the system GLESv3 dynamic library directly
`'-Wl,-z,max-page-size=16384', '-Wl,-z,common-page-size=16384'` is adapted for Android 16KB page size

* iOS

Directly link OpenGLES.framework in `src/renderer/gl_engine/meson.build`

```
if target_machine.system() == 'ios'
   message('Detected iOS platform, linking OpenGLES.framework')
   gles_dep = dependency(
      'OpenGLES',
      required: true
   )
   engine_dep += gles_dep
```

* macOS

Directly link OpenGL.framework in `src/renderer/gl_engine/meson.build`

```
elif target_machine.system() == 'darwin'
   message('Detected macOS platform, linking OpenGL.framework')
   gles_dep = dependency(
      'OpenGL',
      required: true
   )
   engine_dep += gles_dep
endif
```

In addition, I have modified the following code:
```
bool GlRenderer::target(void* display, void* surface, void* context, int32_t id, uint32_t w, uint32_t h, ColorSpace cs)
{
    if (w == 0 || h == 0) return false;
    ....
}
```
I removed the `if (!context || ...)` judgment for the context. I hold the view that we do not need to check the validity of the context, which should be managed by the caller. Especially on the iOS system, the context type is EAGLContext. At present, the only purpose of this context is to call glMakeCurrent, and I think glMakeCurrent should be implemented by the caller instead of us managing or using this context.

### DIFF
--- a/cross/android_aarch64.txt
+++ b/cross/android_aarch64.txt
@@ -2,6 +2,7 @@
 # https://developer.android.com/ndk/guides/other_build_systems
 
 [binaries]
+c       = 'NDK/toolchains/llvm/prebuilt/HOST_TAG/bin/aarch64-linux-androidAPI-clang'
 cpp     = 'NDK/toolchains/llvm/prebuilt/HOST_TAG/bin/aarch64-linux-androidAPI-clang++'
 ar      = 'NDK/toolchains/llvm/prebuilt/HOST_TAG/bin/llvm-ar'
 as      = 'NDK/toolchains/llvm/prebuilt/HOST_TAG/bin/llvm-as'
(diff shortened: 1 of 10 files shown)


hermet (changes requested):
1. Please do not mix GL portability changes with build script(cross) patches.
2. iOS/macOS OpenGL/ES are deprecated, they are not targets officially supported by ThorVG for productions.

PR #4293, opened 2026-04-04
lottie/expressions: replace jerry_eval with jerry_parse/jerry_run and use compiled code

Nor-s:
#### Summary

- Replace `jerry_eval` (parse + execute combined) with separate `jerry_parse` and `jerry_run` APIs
- Cache compiled code per expression to avoid redundant parsing on every frame

### result

| Mac m5pro | github action runner, thread=4 | github action runner, thread=0 |
|-|-|-|
| + 10% | [6% ~12%](https://github.com/Nor-s/perfaction/actions/runs/27230513274) | [11%~14%](https://github.com/Nor-s/perfaction/actions/runs/27230393391)|

### DIFF
--- a/src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp
+++ b/src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp
@@ -17,13 +17,15 @@
 
 #include <math.h>
 
-#include "ecma-eval.h"
+#include "ecma-builtins.h"
+#include "ecma-exceptions.h"
 #include "ecma-function-object.h"
 #include "ecma-gc.h"
 #include "ecma-init-finalize.h"
 #include "ecma-objects-general.h"
 #include "ecma-objects.h"
 #include "jcontext.h"
+#include "js-parser.h"
 
 /** \addtogroup jerry Jerry engine interface
  * @{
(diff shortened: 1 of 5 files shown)


src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp
```diff
@@ -145,24 +146,60 @@ jerry_cleanup (void)
...
+ * @return function object value - if script was parsed successfully,
+ *         thrown error - otherwise
  */
 jerry_value_t
-jerry_eval (const jerry_char_t *source_p, /**< source code */
-            size_t source_size, /**< length of source code */
-            uint32_t flags) /**< jerry_parse_opts_t flags */
+jerry_parse (const jerry_char_t *source_p, /**< source code */
+             size_t source_size, /**< length of source code */
+             uint32_t flags) /**< jerry_parse_opts_t flags */
```
Nor-s, line 162: Original API `jerry_parse_common` to handle [jerry_parse_options_t](https://github.com/jerryscript-project/jerryscript/blob/master/docs/02.API-REFERENCE.md#jerry_parse_option_enable_feature_t). ThorVG doesn't use parse options, so the essential logic is inlined directly.

- [api docs: jerry_parse](https://github.com/jerryscript-project/jerryscript/blob/master/docs/02.API-REFERENCE.md#jerry_parse)

- [code: jerry_parse_common](https://github.com/jerryscript-project/jerryscript/blob/50200152feb724a74a5f64e44d7885151537cfad/jerry-core/api/jerryscript.c#L354-L461)

- [code: jerry_parse](https://github.com/jerryscript-project/jerryscript/blob/50200152feb724a74a5f64e44d7885151537cfad/jerry-core/api/jerryscript.c#L465-L489)

src/loaders/lottie/jerryscript/jerry-core/api/jerryscript.cpp
```diff
@@ -145,24 +146,60 @@ jerry_cleanup (void)
...
+  ecma_object_t *object_p = ecma_get_object_from_value (script);
+  ecma_extended_object_t *ext_object_p = (ecma_extended_object_t *) object_p;
+
+  const ecma_compiled_code_t *bytecode_data_p;
+  bytecode_data_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_compiled_code_t, ext_object_p->u.cls.u3.value);
+
+  JERRY_ASSERT (CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags) == CBC_FUNCTION_SCRIPT);
+
+  return jerry_return (vm_run_global (bytecode_data_p, object_p));
+} /* jerry_run */
```
Nor-s, line 204: [api: jerry_run](https://github.com/jerryscript-project/jerryscript/blob/v3.0.0/docs/02.API-REFERENCE.md#jerry_run)

[code: jerry_run](https://github.com/jerryscript-project/jerryscript/blob/50200152feb724a74a5f64e44d7885151537cfad/jerry-core/api/jerryscript.c#L516-L550)

Nor-s:
@hermet 

Closing this PR for now.

JerryScript uses one fixed-size buffer per thread, and the buffer size is 512 KB. 
In the current `thorvg.example` LottieExpressions case, caching 100 Lottie files already requires about 110 KB of bytecode.

This means that with around 500 or more Lottie files, the cached bytecode can exceed the available buffer, so this approach not work reliably at that scale.

Also, this change would effectively limit the supported use case/specification, 
while the expected performance gain is relatively small. 

For that reason, I think it is better to close this PR.

PR #4294, opened 2026-04-05
Add support for retrieving accessor names

hermet:
- Introduce Accessor APIs to retrieve original ID names
  of corresponding SVG nodes.
- Improve access performance by maintaining a accessible targets.
- APIs are Experimental (beta) version.

Later we can only assign the svg paint id onlt if the accessible is enabled
for now we leave it as it does for backward compatibility.

```
C++ APIs:
 + const char* Accessor::name(uint32_t id)
  + bool Picture::accessible

C APIs:
 + const char* tvg_accessor_name(Tvg_Accessor, uint32_t id)
 + tvg_picture_set_accessible(Tvg_Picture picture, bool accessible)
```

issue: #421

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1743,15 +1743,18 @@ struct TVG_API Picture : Paint
     Result filter(FilterMethod method) noexcept;
 
     /**
-     * @brief Retrieve a paint object from the Picture scene by its Unique ID.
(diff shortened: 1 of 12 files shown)


hermet:
binary diff: +0.8kb without svg

Issue #4280, opened 2026-04-01
test: Floating-point failure on x86 32-bit CPUs


jmkim:
Dear team o/

On x86 32-bit CPUs (e.g. i386, i486, i586, i686), the following test failure appears:
* [**i686 GNU/Linux** build failure full log](https://buildd.debian.org/status/fetch.php?pkg=thorvg&arch=i386&ver=1.0.3%2Bdfsg2-1&stamp=1774809850&raw=0)
* [**i686-at386 GNU Hurd** build failure full log](https://buildd.debian.org/status/fetch.php?pkg=thorvg&arch=hurd-i386&ver=1.0.3%2Bdfsg2-3&stamp=1775007581&raw=0)
    ```
    ../test/testPaint.cpp:170: FAILED:
      REQUIRE( h == 100.0f )
    with expansion:
      100.00002f == 100.0f
    ```

Enabling SSE2 makes the test pass. In Debian, the successful build uses `-msse2 -mfpmath=sse`: [debian/rules](https://salsa.debian.org/debian/thorvg/-/blob/debian/unstable/debian/rules#L5-L7)  
* [**i686 GNU/Linux** build success full log](https://buildd.debian.org/status/fetch.php?pkg=thorvg&arch=i386&ver=1.0.3%2Bdfsg2-4&stamp=1775022187&raw=0)
* [**i686-at386 GNU Hurd** build success full log](https://buildd.debian.org/status/fetch.php?pkg=thorvg&arch=hurd-i386&ver=1.0.3%2Bdfsg2-4&stamp=1775023981&raw=0)
    ```
    ../test/testPaint.cpp:170: PASSED:
      REQUIRE( h == 100.0f )
    with expansion:
      100.0f == 100.0f
    ```

I am not sure whether this is expected behaviour, so I am reporting it here. Many thanks!

hermet:
If this is an issue that is unavoidable due to the architecture, relaxing the test pass criteria could be an option.

hermet:
@jmkim @Nor-s  The test fix will be in v1.0.4, Thanks!

PR #4299, opened 2026-04-06
test: relax Bounding Box checks for precision variance

Nor-s:
- issue: #4280

On 32-bit x86 with -mfpmath=387 -O2, Debian reproduces a small x87 floating-point variance (100.00002f vs 100.0f) in the Bounding Box test.

I tested an x86 32-bit x87 build on Ubuntu with -m32 -mfpmath=387, but could not reproduce the failure. This suggests the issue may be environment-specific floating-point variance rather than a deterministic ThorVG logic bug.

Given the small magnitude of the difference, relaxing exact float comparisons in the test may be worth considering.

| before | after |
| - | - |
| image| image|

```
  export CFLAGS='-O2 -g3 -fno-omit-frame-pointer -m32 -march=i686 -mfpmath=387'
  export CXXFLAGS='-O2 -g3 -fno-omit-frame-pointer -m32 -march=i686 -mfpmath=387'
  export LDFLAGS='-m32'
  meson setup build-x87-dbgopt -Dtests=true -Dengines=sw -Dloaders=all -Dsavers=gif -Dsimd=false -Dstatic=true --buildtype=debugoptimized --wipe
  meson compile -C build-x87-dbgopt
  ./build-x87-dbgopt/test/tvgUnitTests "Bounding Box" --success
```

### DIFF
--- a/test/testPaint.cpp
+++ b/test/testPaint.cpp
@@ -164,10 +164,10 @@ TEST_CASE("Bounding Box", "[tvgPaint]")
 
         //Positive
         REQUIRE(shape->bounds(&x, &y, &w, &h) == Result::Success);
-        REQUIRE(x == 100.0f);
-        REQUIRE(y == 121.0f);
-        REQUIRE(w == 20.0f);
-        REQUIRE(h == 100.0f);
+        REQUIRE(x == Approx(100.0f).margin(0.001f));
+        REQUIRE(y == Approx(121.0f).margin(0.001f));
+        REQUIRE(w == Approx(20.0f).margin(0.001f));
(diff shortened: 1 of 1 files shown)


test/testPaint.cpp
```diff
@@ -167,7 +167,7 @@ TEST_CASE("Bounding Box", "[tvgPaint]")
         REQUIRE(x == 100.0f);
         REQUIRE(y == 121.0f);
         REQUIRE(w == 20.0f);
-        REQUIRE(h == 100.0f);
+        REQUIRE(h == Approx(100.0f).margin(0.001f));
```
hermet, line 170: @Nor-s  Please add margin to all xywh

PR #4303, opened 2026-04-07
lottie: resolve package-root relative image asset paths

tanjingshining:
## Summary
Fixes external image asset resolution when the Lottie JSON is under an `animations/` directory and assets use a **`u` value with a leading `/`** (package-root relative), as in many **dotLottie**-style layouts after extraction.
## Changes
- **LottieLoader**: If the JSON file’s parent directory is named `animations`, set `dirName` to that directory’s parent so sibling folders such as `images/` resolve correctly.
- **parseImage**: Strip leading `/` and `\` from `u`, then join `dirName`, path prefix, and filename without missing or duplicating separators.
## Tests
- Fixture: `test/resources/lottie_pkg/` (minimal JSON + 1×1 PNG).
- Test: `Lottie package-root asset path` in `test/testLottie.cpp` when `THORVG_PNG_LOADER_SUPPORT` and `THORVG_FILE_IO_SUPPORT` are defined.
```bash
meson setup build -Dtests=true
ninja -C build test/tvgUnitTests
./build/test/tvgUnitTests "Lottie package-root asset path"

### CODE REVIEW

**Comment by tinyjin** [comment]
@tanjingshining Could you check if the assetResolver could cover your use-case?

https://github.com/thorvg/thorvg.example/blob/main/src/LottieExtension.cpp#L283

**Comment by tanjingshining** [comment]
@tinyjin Thanks for the pointer — I did check AssetResolver.

It is only invoked from LottieBuilder::updateImage when a resolver is set on the loader; the callback receives the Picture, the already composed string in image->bitmap.path, and the user data. If no resolver is registered, external assets are loaded with picture->load(bitmap.path) in LottieImage::prepare, so the path ThorVG builds must already be correct for the default file-IO case.

So in principle an embedder could work around a wrong path by interpreting src in the resolver and loading from a remapped location, but that would (1) not fix the out-of-the-box load("…/animations/foo.json") path, (2) push dotLottie/package-root semantics into every app that does not use a resolver, and (3) duplicate logic that is really part of Lottie’s dir + u + p resolution.

The change in the PR keeps default behavior aligned with common package layouts (e.g. JSON under animations/ and u with a leading /) without requiring a custom resolver; the resolver remains useful for opaque assets or non-file sources.

**Comment by tinyjin** [comment]
@tanjingshining The AssetResolver is made for that use.
Since the resolver gives application a control, in the callback, you can put your own file IO logic.

I don't see limitation of resolver in case mentioned and your json example.

We may need to define what actual problem is:
> not fix the out-of-the-box load("…/animations/foo.json") path

If relative path isn't working in some reason, I think you can load with absolute path in resolver:
```cpp
picture->load("path/to/your-dir/dot.png"));
```

If this is the case where resolver has limitation other than usability in specific case, please feel free to let me know.
Also check below, I could render your json+image with assetResolver:
```cpp
//asset resolver (image)
{
    auto resolver = std::unique_ptr<tvg::LottieAnimation>(tvg::LottieAnimation::gen());
    auto picture = resolver->picture();

    auto func = [](tvg::Paint* p, const char* src, void* data) {
        if (p->type() != tvg::Type::Picture) return false;

        // Can resolve in abs path to correctly grab an image anywhere
        auto assetPath = string(src).replace(0, sizeof(EXAMPLE_DIR"/lottie/extensions/") - 1, EXAMPLE_DIR"/");
        auto ret = static_cast<tvg::Picture*>(p)->load(EXAMPLE_DIR"/image/dot.png");
        return (ret == (tvg::Result) 0) ? true : false;
    };

    if (!tvgexam::verify(picture->resolver(func, nullptr))) return false;
    if (!tvgexam::verify(picture->load(EXAMPLE_DIR"/pkg_root_u.json"))) return false;

    canvas->add(picture);
}
```

### thorvg_pkg_path_test + dot.png

CleanShot 2026-04-07 at 17 46 02@2x

tanjingshining:
Hi @tinyjin — thanks again for pointing to AssetResolver and the example in LottieExtension.cpp. We agree that the resolver is the right mechanism when the app needs custom IO (non-file sources, sandboxes, remapping, etc.), and we’re happy to keep using it in examples for those cases.

The PR is about a narrower, orthogonal issue: default file resolution for typical dotLottie packages on disk.

The dotLottie specs describe a zip layout where the animation JSON lives under animations/ (v1) or a/ (v2), while raster assets live under the archive root in images/ (v1) or i/ (v2). The format is intended so that paths are resolved relative to the package root, not necessarily relative to the JSON file’s directory.

Today, if the loader uses dirName = dirname(path_to_json) and the asset block uses something like u: "images/" (as many exports do), the composed path becomes:

…/<animations|a>/images/…

but after unzip the files are actually at:

…/images/… (v1) or …/i/… (v2).

So picture->load(path_to_json) can fail without any app-specific policy — not because the resolver is missing, but because the default base directory does not match the layout described in the dotLottie spec for that case. The resolver then forces every embedder to re-implement the same “package root vs JSON dir” logic, which is easy to get wrong and duplicates what is really format semantics, not application policy.

We’re not trying to replace AssetResolver. We’d like the default path composition to align with common dotLottie-on-disk layouts so plain load(json) works out of the box; the resolver would remain for everything that shouldn’t be hardcoded (HTTP, encrypted assets, virtual FS, etc.).

If the patch feels too broad, we’re happy to narrow the scope or add focused tests (e.g. unzip tree with manifest.json + animations/ + images/) — whatever helps you merge with confidence.

Thanks for considering it.

hermet:
@tanjingshining What dotlottie player do you have? Maybe this can be your solution? 
```
* @param[in] rpath A resource directory path, if the @p data needs to access any external resources.
Result load(const char* data, uint32_t size, const char* mimeType, const char* rpath = nullptr, bool copy = false)
```

test/resources/lottie_pkg/README
```diff
@@ -0,0 +1,9 @@
+Test fixture for Lottie file layout similar to dotLottie extract layout:
+  animations/<file>.json   — main animation
+  images/<file>            — bitmap assets
+
+Asset entries use "u": "/images/..." (leading slash = package root), which must
```
hermet, line 5: Need to confirm:
ThorVG is oriented toward the standard Lottie spec; The policy/behavior shouldn't be changed for dotLottie.

tanjingshining:
Thanks @hermet @tinyjin for the clarification and the rpath pointer.

We’ve updated our demo to load the Lottie JSON from memory with Picture::load(data, size, "lottie+json", packageRoot, copy=true) so external assets resolve against the dotLottie extract root. That covers our dotLottie-on-disk case without changing ThorVG’s default file-path behavior.

Given that, I’m closing this PR to avoid pushing a policy change you’d prefer not to take for standard Lottie vs dotLottie. If you ever want the same ergonomics for plain load(jsonPath) in the future, we can revisit with a narrower patch or tests.

Thanks again for the review.

tanjingshining:
@hermet @tinyjin  Hi ThorVG maintainers,

I’m working on packaging ThorVG for Debian (or preparing metadata for a downstream distro). Debian packages are normally uploaded by distro maintainers, not necessarily by upstream, so I wanted to reach out briefly.

Upstream plans
Do you have any plans to maintain an official Debian package yourselves? If not, that’s totally fine — I’m mainly trying to avoid duplicating effort or conflicting with your intentions.

Permission / expectations
Are you comfortable with a third-party maintainer shipping ThorVG in Debian (or similar), as long as we follow the root LICENSE (MIT), reproduce the required notices, and document the bundled components (e.g. WebP, RapidJSON, JerryScript) in debian/copyright?

Release source
For reproducible packaging we’d rely on tagged releases / GitHub release tarballs (or the same source you consider authoritative). If you have a preferred distribution format (signed tarballs, etc.), a short note in the README or a reply here would help.

Security contact (optional)
If you’re open to it, a SECURITY.md or a documented security contact would help distros track CVEs — not a blocker for packaging, but appreciated.

Thanks for ThorVG and for any guidance you can share.

Best regards.

hermet:
> @hermet @tinyjin Hi ThorVG maintainers,
> 
> I’m working on packaging ThorVG for Debian (or preparing metadata for a downstream distro). Debian packages are normally uploaded by distro maintainers, not necessarily by upstream, so I wanted to reach out briefly.
> 
> Upstream plans Do you have any plans to maintain an official Debian package yourselves? If not, that’s totally fine — I’m mainly trying to avoid duplicating effort or conflicting with your intentions.
> 
> Permission / expectations Are you comfortable with a third-party maintainer shipping ThorVG in Debian (or similar), as long as we follow the root LICENSE (MIT), reproduce the required notices, and document the bundled components (e.g. WebP, RapidJSON, JerryScript) in debian/copyright?
> 
> Release source For reproducible packaging we’d rely on tagged releases / GitHub release tarballs (or the same source you consider authoritative). If you have a preferred distribution format (signed tarballs, etc.), a short note in the README or a reply here would help.
> 
> Security contact (optional) If you’re open to it, a SECURITY.md or a documented security contact would help distros track CVEs — not a blocker for packaging, but appreciated.
> 
> Thanks for ThorVG and for any guidance you can share.
> 
> Best regards.

@tanjingshining Hi, thorvg debian is already in forky (testing) https://packages.debian.org/forky/libthorvg1

\+ cc: @jmkim

tanjingshining:
@hermet  Thanks for letting me know.

Issue #2239, opened 2024-05-02
Support nested `<svg>` elements


JSUYA:
[Example]
```html
<svg width="64" height="64" viewBox="0 0 64 64" xmlns="http://www.w3.org/2000/svg" fill="none">
<rect rx="8" fill="#80ff45" width="64" height="64"/>
    <svg x="12.5%" y="37.5%" width="50%" height="32" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg">
        <rect rx="4" fill="#ff4596" width="32" height="50%"/>
    </svg>
</svg>
```

Our SVG loader parses the `<svg>` element at the check of validation(`header()`) and skips any `<svg>` elements it encounters thereafter.
We can make `loader->doc` an array type and make the first doc node the main doc node.
And every node can store and reference its parent `<svg>` element node.
This work should be process in conjunction with valid check function. it may require more time or may not proceed according to our policy.

JSUYA:
#4306

PR #4306, opened 2026-04-07
svg_loader: Add a log message to unsupported nested <svg> elements

JSUYA:
Nested <svg> elements are an SVG Full feature and are not part of the SVG Tiny spec that thorvg follows.
Subsequently, to prevent unexpected crashes that may occur in stack.pop(), `<svg>` is treated as `<g>`

issues: #2239

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -3229,7 +3229,10 @@ static void _svgLoaderParserXmlOpen(SvgParserContext* ctx, const char* content, 
             node = method(ctx, nullptr, attrs, attrsLength, xmlParseAttributes);
             ctx->doc = node;
         } else {
-            if (STR_AS(tagName, "svg")) return; //Already loaded <svg>(SvgNodeType::Doc) tag
+            if (STR_AS(tagName, "svg")) {
+                TVGLOG("SVG", "Nested <svg> element is not supported.");
+                method = _createGNode;
+            }
             if (ctx->stack.count > 0) parent = ctx->stack.last();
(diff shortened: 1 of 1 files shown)


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3229,7 +3229,10 @@ static void _svgLoaderParserXmlOpen(SvgParserContext* ctx, const char* content,
             node = method(ctx, nullptr, attrs, attrsLength, xmlParseAttributes);
             ctx->doc = node;
         } else {
-            if (STR_AS(tagName, "svg")) return; //Already loaded <svg>(SvgNodeType::Doc) tag
+            if (STR_AS(tagName, "svg")) {
+                TVGERR("SVG", "Nested <svg> element is not supported.");
```
hermet: An error is an unacceptable scenario and must be fixed. If this is a compromised solution by a thorvg spec, we would rather log the situation to inform the user.

JSUYA: I changed the log type.

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -3229,7 +3229,10 @@ static void _svgLoaderParserXmlOpen(SvgParserContext* ctx, const char* content,
...
+                return;
+            }
```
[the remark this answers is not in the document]
JSUYA: I have confirmed that unexpected behavior can occur when stack.pop() is called on `</svg>`.
`<svg>` is ignored, but to prevent crashes, it is treated as a `<g>` node.

PR #4308, opened 2026-04-08
lottie: fix gradient alpha interpolation

Neizety:
When color and alpha stop positions didn't align, alpha was interpolated using the last output stop as a reference instead of the true neighboring alpha stops, producing incorrect opacity values. Rewrite the merge to walk both arrays in tandem, interpolating alpha between the correct bounding opacity stops

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -351,80 +351,85 @@ uint32_t LottieGradient::populate(ColorStop& color, size_t count)
 {
     if (!color.input) return 0;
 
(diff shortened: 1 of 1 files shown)


Neizety:
left before, right after
[sticker.json](https://github.com/user-attachments/files/26556257/sticker.json)


image

Neizety:
(j += 2; break) causes a regression (arithmetic overflow when calculating opacity) that triggers animation normalization. Most likely, the designer used an old version of rlottie with this regression and didn’t notice it. Most likely, an older version of rlottie with this regression was used during the design, so it went unnoticed

PR #4309, opened 2026-04-08
lottie/expressions: refactor _buildProperty

Nor-s:
image

`_buildProperty` recreated all function objects every frame
(takes 26% of total samples, 17,200 / 66,008). 

Now it creates them once per context and reuses them on subsequent calls by swapping native pointers in-place.

### Note

For functions sharing an `ExpContent`, `updateContent` retrieves the pointer via `jerry_object_get_native_ptr` and modifies the struct fields directly in C++. Since multiple function objects reference the same `ExpContent*`, updating one applies to all.

### Result

image

Own CPU samples for _buildProperty function, over 4000 ticks: 26% -> 12%

| | fps % | BinarySize |
|-|-|-|
| local machine (Mac M5 pro) | +13% | no check |
| github action runner | [+13%~18%](https://github.com/Nor-s/perfaction/actions/runs/24090103952) |[-190~+192](https://github.com/Nor-s/perfaction/actions/runs/24114433290) |

### DIFF
--- a/src/loaders/lottie/tvgLottieExpressions.cpp
+++ b/src/loaders/lottie/tvgLottieExpressions.cpp
@@ -1101,106 +1101,95 @@ static void _buildProperty(float frameNo, jerry_value_t context, LottieExpressio
     jerry_object_set_sz(context, EXP_VALUE, value);
     jerry_value_free(value);
 
-    auto valueAtTime = jerry_function_external(_valueAtTime);
-    jerry_object_set_sz(context, "valueAtTime", valueAtTime);
-    jerry_object_set_native_ptr(valueAtTime, nullptr, exp);
-    jerry_value_free(valueAtTime);
+    //check if functions are already registered. If not (first build), set all functions.
+    //otherwise, update native_ptr and update shared data in ExpContent
+    auto check = jerry_object_get_sz(context, "valueAtTime");
+    auto init = jerry_value_is_undefined(check);
+    jerry_value_free(check);
 
-    auto velocity = jerry_number(0.0f);
-    jerry_object_set_sz(context, "velocity", velocity);
-    jerry_value_free(velocity);
+    //set or update a native function bound to exp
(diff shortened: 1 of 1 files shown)


wenjieshen:
### Manual validation
Screenshot 2026-04-08 at 11 59 32 AM

- No observable issues.
- FPS: 23% Improvement on Ubuntu 24.04.4

hermet:
@Nor-s please deal with the copilot comments. thanks!

Nor-s:
> @Nor-s please deal with the copilot comments. thanks!

@hermet 

Hello, I expect the Copilot review to take 1 day.

image

The content of the Copilot is about considering the above case in the Expression engine.
Because of caching, problems occur when users overwrite functions.
(However, this may also be the case for JerryScript's built-in functions.)

I will review this.

If, after review, we need to prevent this, JerryScript provides an option to prevent overwriting,
so we will use that. If we add this, the binary size is expected to increase.
Thank you.

hermet:
> > @Nor-s please deal with the copilot comments. thanks!
> 
> @hermet
> 
> Hello, I expect the Copilot review to take 1 day.
> 
> image
> The content of the Copilot is about considering the above case in the Expression engine. Because of caching, problems occur when users overwrite functions. (However, this may also be the case for JerryScript's built-in functions.)
> 
> I will review this.
> 
> If, after review, we need to prevent this, JerryScript provides an option to prevent overwriting, so we will use that. If we add this, the binary size is expected to increase. Thank you.

@Nor-s I think it's not really relate to this PR, We can separate the issue.

Nor-s:
@hermet 

image


I added log messages for overwrite cases for debugging,
The review comments have been resolved.

hermet:
fps: +11%
size: -

@Nor-s Could you please elaborate the commit msg in the next? Thanks.

hermet:
Need to fix this regression:
**LottieInteraction**
`[E] LOTTIE (../src/loaders/lottie/tvgLottieExpressions.cpp 1183): ExpContent lost due to function overwrite`

PR #4322, opened 2026-04-10
lottie/expressions: register magic strings for faster property lookup

Nor-s:
- Register 44 frequently used Lottie expression property names (e.g. `value`, `transform`, `position`, `$bm_rt`) as JerryScript external magic strings.      

- Magic strings are interned by the engine at init, so repeated lookups on the per-frame hot path (`buildProperty`, `buildLayer`, `buildTransform`, `buildGlobal`, `buildComp`), reducing **allocation/deallocation overhead** for expressions.

### **Change**

- Expose `jerry_register_magic_strings` API in the embedded JerryScript build
- Define a static string pool sorted by length, then lexicographically                                                    
- Build pointer/length arrays at init and register them once via `jerry_register_magic_strings`      

### **Result**

remove hot path: `jerry_string`

| before | after |
|-|-|
|image| image|

| | fps | binarySize|
|-|-|-|
|local machine(Mac M5 pro) | +21%  | |
|github action runner | [+15%~19%](https://github.com/Nor-s/perfaction/actions/runs/24214719236)| [+412~474](https://github.com/Nor-s/perfaction/actions/runs/24227349397)|

#### Reference

- [jerry_register_matic_strings](https://github.com/jerryscript-project/jerryscript/blob/master/docs/02.API-REFERENCE.md#jerry_register_magic_strings)

<details>                                                                                                                                
<summary>jerry_object_set_sz -> jerry_string -> ecma_find_special_string </summary>    

```c++
/**
 * Set a property to the specified object with the given name.
 *
 * Note:
 *      returned value must be freed with jerry_value_free, when it is no longer needed.
 *
 * @return true value - if the operation was successful
 *         value marked with error flag - otherwise
 */
jerry_value_t
jerry_object_set_sz (jerry_value_t object, /**< object value */
                     const char *key_p, /**< property key */
                     const jerry_value_t value) /**< value to set */
{
  jerry_value_t key_str = jerry_string_sz (key_p);
  jerry_value_t result = jerry_object_set (object, key_str, value);
  ecma_free_value (key_str);
  return result;
} /* jerry_object_set */


/**
 * Create string value from the input zero-terminated ASCII string.
 *
 * @return created string
 */
jerry_value_t
jerry_string_sz (const char *str_p) /**< pointer to string */
{
  const jerry_char_t *data_p = (const jerry_char_t *) str_p;
  return jerry_string (data_p, lit_zt_utf8_string_size (data_p), JERRY_ENCODING_CESU8);
} /* jerry_string_sz */



/**
 * Create a string value from the input buffer using the specified encoding.
 * The content of the buffer is assumed to be valid in the specified encoding, it's the callers responsibility to
 * validate the input.
 *
 * See also: jerry_validate_string
 *
 * @return created string
 */
jerry_value_t
jerry_string (const jerry_char_t *buffer_p, /**< pointer to buffer */
              jerry_size_t buffer_size, /**< buffer size */
              jerry_encoding_t encoding) /**< buffer encoding */
{
  ecma_string_t *ecma_str_p = NULL;
  //JERRY_ASSERT (jerry_validate_string (buffer_p, buffer_size, encoding));

  switch (encoding)
  {
    case JERRY_ENCODING_CESU8:
    {
      ecma_str_p = ecma_new_ecma_string_from_utf8 (buffer_p, buffer_size);
      
      

/**
 * Allocate new ecma-string and fill it with characters from the utf8 string
 *
 * @return pointer to ecma-string descriptor
 */
ecma_string_t *
ecma_new_ecma_string_from_utf8 (const lit_utf8_byte_t *string_p, /**< utf-8 string */
                                lit_utf8_size_t string_size) /**< string size */
{
  JERRY_ASSERT (string_p != NULL || string_size == 0);
  JERRY_ASSERT (lit_is_valid_cesu8_string (string_p, string_size));

  ecma_string_t *string_desc_p = ecma_find_special_string (string_p, string_size);
  
  

/**
 * Checks whether a string has a special representation, that is, the string is either a magic string,
 * an external magic string, or an uint32 number, and creates an ecma string using the special representation,
 * if available.
 *
 * @return pointer to ecma string with the special representation
 *         NULL, if there is no special representation for the string
 */
static ecma_string_t *
ecma_find_special_string (const lit_utf8_byte_t *string_p, /**< utf8 string */
                          lit_utf8_size_t string_size) /**< string size */
{


```

</details>

### DIFF
--- a/src/loaders/lottie/tvgLottieExpressions.cpp
+++ b/src/loaders/lottie/tvgLottieExpressions.cpp
@@ -63,29 +63,39 @@ static const char* EXP_EFFECT= "effect";
 
 /**
  * external magic strings for the per-frame hot path (buildProperty, buildLayer, buildTransform, bm_rt).
- * magic pool: sorted by length, then lexicographically. null-separated,
- *             must be static, because 'jerry_register_magic_strings' does not copy
- *             update MAGIC_POOL_COUNT if this pool changes
- * pointers/lengths are computed once at init and stored in .bss (zero binary cost).
+ * magic pool: sorted by length, then lexicographically. packed in a single static blob,
+ *             because 'jerry_register_magic_strings' does not copy the string data.
+ * lengths/count are generated from a single list at compile time; pointers are built once at init.
  */
+#define TVG_LOTTIE_MAGIC_STRING_LIST(X) \
+    X("key") \
+    X("comp") X("name") X("time") \
+    X("index") X("layer") X("scale") X("speed") X("value") X("width") \
+    X("$bm_rt") X("effect") X("height") X("parent") X("toComp") X("wiggle") \
+    X("content") X("enabled") X("inPoint") X("numKeys") X("opacity") \
+    X("duration") X("hasAudio") X("hasVideo") X("outPoint") X("position") X("rotation") X("thisComp") X("velocity") \
+    X("hasParent") X("numLayers") X("startTime") X("thisLayer") X("timeRemap") X("transform") \
+    X("nearestKey") \
+    X("anchorPoint") X("audioActive") X("speedAtTime") X("valueAtTime") \
+    X("thisProperty") \
+    X("frameDuration") X("propertyIndex") \
+    X("velocityAtTime")
+
+#define TVG_LOTTIE_MAGIC_POOL_ENTRY(str) str
 static constexpr const char _magicPool[] =
-    "key\0" /* length: 3 */
-    "comp\0" "name\0" "time\0" /* length: 4 */
-    "index\0" "layer\0" "scale\0" "speed\0" "value\0" "width\0" /* length: 5 */
-    "$bm_rt\0" "effect\0" "height\0" "parent\0" "toComp\0" "wiggle\0" /* length: 6 */
-    "content\0" "enabled\0" "inPoint\0" "numKeys\0" "opacity\0" /* length: 7 */
-    "duration\0" "hasAudio\0" "hasVideo\0" "outPoint\0" "position\0" "rotation\0" "thisComp\0" "velocity\0" /* length: 8 */
-    "hasParent\0" "numLayers\0" "startTime\0" "thisLayer\0" "timeRemap\0" "transform\0" /* length: 9 */
-    "nearestKey\0" /* length: 10 */
-    "anchorPoint\0" "audioActive\0" "speedAtTime\0" "valueAtTime\0" /* length: 11 */
-    "thisProperty\0" /* length: 12 */
-    "frameDuration\0" "propertyIndex\0" /* length: 13 */
-    "velocityAtTime"; /* length: 14 */
+    TVG_LOTTIE_MAGIC_STRING_LIST(TVG_LOTTIE_MAGIC_POOL_ENTRY);
+#undef TVG_LOTTIE_MAGIC_POOL_ENTRY
 
-#define MAGIC_POOL_COUNT 44
+#define TVG_LOTTIE_MAGIC_LENGTH_ENTRY(str) (jerry_length_t)(sizeof(str) - 1),
+static constexpr jerry_length_t _magicLengths[] = {
+    TVG_LOTTIE_MAGIC_STRING_LIST(TVG_LOTTIE_MAGIC_LENGTH_ENTRY)
+};
+#undef TVG_LOTTIE_MAGIC_LENGTH_ENTRY
 
-static const jerry_char_t* _magicStrings[MAGIC_POOL_COUNT];
-static jerry_length_t _magicLengths[MAGIC_POOL_COUNT];
+#define MAGIC_STRING_COUNT (sizeof(_magicLengths) / sizeof(_magicLengths[0]))
+static const jerry_char_t* _magicStrings[MAGIC_STRING_COUNT];
+
+#undef TVG_LOTTIE_MAGIC_STRING_LIST
 
 static LottieExpressions* exps = nullptr;   //singleton instance engine
 
@@ -1513,15 +1523,13 @@ LottieExpressions::LottieExpressions()
 {
     jerry_init(JERRY_INIT_EMPTY);
 
-    //build pointer/length arrays from the magic pool
+    //build magic string arrays from the magic pool
     auto p = _magicPool;
-    for (uint32_t i = 0; i < MAGIC_POOL_COUNT; ++i) {
+    for (uint32_t i = 0; i < MAGIC_STRING_COUNT; ++i) {
         _magicStrings[i] = (const jerry_char_t*)p;
-        auto len = strlen(p);
-        _magicLengths[i] = (jerry_length_t)len;
-        p += len + 1;
+        p += _magicLengths[i];
     }
-    jerry_register_magic_strings(_magicStrings, MAGIC_POOL_COUNT, _magicLengths);
+    jerry_register_magic_strings(_magicStrings, MAGIC_STRING_COUNT, _magicLengths);
 
     _buildMath(buildGlobal());
 }
--- a/src/loaders/lottie/tvgLottieExpressions.cpp
+++ b/src/loaders/lottie/tvgLottieExpressions.cpp
@@ -61,6 +61,32 @@ static const char* EXP_VALUE = "value";
 static const char* EXP_INDEX = "index";
 static const char* EXP_EFFECT= "effect";
 
+/**
+ * external magic strings for the per-frame hot path (buildProperty, buildLayer, buildTransform, bm_rt).
+ * magic pool: sorted by length, then lexicographically. null-separated,
+ *             must be static, because 'jerry_register_magic_strings' does not copy
+ *             update MAGIC_POOL_COUNT if this pool changes
+ * pointers/lengths are computed once at init and stored in .bss (zero binary cost).
+ */
+static constexpr const char _magicPool[] =
+    "key\0" /* length: 3 */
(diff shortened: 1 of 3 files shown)


src/loaders/lottie/tvgLottieExpressions.cpp
```diff
@@ -61,6 +61,28 @@ static const char* EXP_VALUE = "value";
...
+    "index\0" "layer\0" "scale\0" "speed\0" "value\0" "width\0" /* length: 5 */
+    "$bm_rt\0" "effect\0" "height\0" "parent\0" "toComp\0" "wiggle\0" /* length: 6 */
+    "content\0" "enabled\0" "inPoint\0" "numKeys\0" "opacity\0" /* length: 7 */
+    "hasAudio\0" "hasVideo\0" "outPoint\0" "position\0" "rotation\0" "velocity\0" /* length: 8 */
+    "hasParent\0" "numLayers\0" "startTime\0" "timeRemap\0" "transform\0" /* length: 9 */
+    "nearestKey\0" /* length: 10 */
+    "anchorPoint\0" "audioActive\0" "speedAtTime\0" "valueAtTime\0" /* length: 11 */
+    "propertyIndex\0" /* length: 13 */
+    "velocityAtTime"; /* length: 14 */
+static constexpr uint32_t _magicCount = 37;
```
tinyjin: maintainability idea:
The _magicCount should be updated alongside with _magicPool, which could lead mistake easily.

Maybe we can make the _magicCount calculated from _magicPool array via constexpr function?
Could you check if something like this could be ideal?

```cpp
static constexpr uint32_t _countNulls(const char* p, size_t n) {
      auto c = 0;
      auto n = sizeof(_magicPool);
      for (size_t i = 0; i < n; i++) if (!_magicPool[i]) c++;
      return c;
  }

static constexpr uint32_t _magicCount = _countNulls();
```

src/loaders/lottie/tvgLottieExpressions.cpp
```diff
@@ -61,6 +61,28 @@ static const char* EXP_VALUE = "value";
...
+    "time\0" /* length: 4 */
+    "index\0" "layer\0" "scale\0" "speed\0" "value\0" "width\0" /* length: 5 */
+    "$bm_rt\0" "effect\0" "height\0" "parent\0" "toComp\0" "wiggle\0" /* length: 6 */
+    "content\0" "enabled\0" "inPoint\0" "numKeys\0" "opacity\0" /* length: 7 */
+    "hasAudio\0" "hasVideo\0" "outPoint\0" "position\0" "rotation\0" "velocity\0" /* length: 8 */
+    "hasParent\0" "numLayers\0" "startTime\0" "timeRemap\0" "transform\0" /* length: 9 */
+    "nearestKey\0" /* length: 10 */
+    "anchorPoint\0" "audioActive\0" "speedAtTime\0" "valueAtTime\0" /* length: 11 */
+    "propertyIndex\0" /* length: 13 */
+    "velocityAtTime"; /* length: 14 */
```
tinyjin: Looks like some strings which are also frequently used are missing:

- thisComp, thisLayer, thisProperty... and so on

Just curious, if it's intentional

tinyjin (commented):
Thanks, the performance gain is observed from my end too.
Left for minor points, please check.

src/loaders/lottie/tvgLottieExpressions.cpp
Nor-s: Sorry, I missed the strings in `LottieExpressions::buildGlobal`, `LottieExpressions::buildComp` "comp, thisComp, thisLayer, thisProperty".

 I'll add them. Thank you for catching that!

src/loaders/lottie/tvgLottieExpressions.cpp
Nor-s: This sounds like a great idea! I'll make the fix. Thanks for the review!

Nor-s:
> Thanks, the performance gain is observed from my end too.
> Left for minor points, please check.

I have reviewed the points. Thank you.

src/loaders/lottie/tvgLottieExpressions.cpp
```diff
@@ -61,6 +61,37 @@ static const char* EXP_VALUE = "value";
...
+    "content\0" "enabled\0" "inPoint\0" "numKeys\0" "opacity\0" /* length: 7 */
+    "duration\0" "hasAudio\0" "hasVideo\0" "outPoint\0" "position\0" "rotation\0" "thisComp\0" "velocity\0" /* length: 8 */
+    "hasParent\0" "numLayers\0" "startTime\0" "thisLayer\0" "timeRemap\0" "transform\0" /* length: 9 */
+    "nearestKey\0" /* length: 10 */
+    "anchorPoint\0" "audioActive\0" "speedAtTime\0" "valueAtTime\0" /* length: 11 */
+    "thisProperty\0" /* length: 12 */
+    "frameDuration\0" "propertyIndex\0" /* length: 13 */
+    "velocityAtTime"; /* length: 14 */
+
+static constexpr uint32_t _countNulls() {
```
hermet: @Nor-s Let's using a magic number.
ie) `#define MAGIC_POOL_COUNT`

src/loaders/lottie/tvgLottieExpressions.cpp
```diff
@@ -61,6 +61,37 @@ static const char* EXP_VALUE = "value";
...
 static const char* EXP_EFFECT= "effect";
 
+/**
+ * external magic strings for the per-frame hot path (buildProperty, buildLayer, buildTransform, bm_rt).
+ * magic pool: sorted by length, then lexicographically. null-separated,
+ *             must be static, because 'jerry_register_magic_strings' does not copy
+ * pointers/lengths are computed once at init and stored in .bss (zero binary cost).
+ */
+static constexpr const char _magicPool[] =
+    "key\0" /* length: 3 */
```
hermet: This looks a bit like an unusual style. How about using a character array instead? (`const char` ->  `const char*`)  It would be more efficient since strlen() could be computed at compile time, and you could remove the explicit `\0` separators.

hermet (changes requested):
@Nor-s Please check comments. Thanks.

src/loaders/lottie/tvgLottieExpressions.cpp
Nor-s: Thanks for the review. After seeing your suggestion, I noticed the binary size had increased, so I set up a benchmark to compare the two approaches. The results were:

### BinarySize result:


#### const char [] 

-  [+472~ 498](https://github.com/Nor-s/sizecheck-action-test/pull/2#issuecomment-4232421804)
- local: + 484

#### const char* []

```cpp
static const char* _magicString[] = {
    "key",
    "comp", "name", "time",
    "index", "layer", "scale", "speed", "value", "width",
    "$bm_rt", "effect", "height", "parent", "toComp", "wiggle",
    "content", "enabled", "inPoint", "numKeys", "opacity",
    "duration", "hasAudio", "hasVideo", "outPoint", "position", "rotation", "thisComp", "velocity",
    "hasParent", "numLayers", "startTime", "thisLayer", "timeRemap", "transform",
    "nearestKey",
    "anchorPoint", "audioActive", "speedAtTime", "valueAtTime",
    "thisProperty",
    "frameDuration", "propertyIndex",
    "velocityAtTime",
};
```
- [+620~5584](https://github.com/Nor-s/sizecheck-action-test/pull/1#issuecomment-4232396429)

- local: +1514



### References
- https://github.com/chromium/chromium/blob/main/docs/native_relocations.md
- https://glandium.org/blog/?p=2361




The variance in the pointer array results (+620 ~ 5,584) seems quite large depending on the platform. I'd appreciate any thoughts on how best to approach this. 

Please let me know if I've misunderstood anything. Thanks!

hermet: Ok, good. If so, ca't we predefine the strlen in advance?
ie) `_magicString[] / _magicStringLen[]`

src/loaders/lottie/tvgLottieExpressions.cpp
Nor-s: Updated to use a fixed magic number (`MAGIC_POOL_COUNT`) instead of `_countNulls()`.

src/loaders/lottie/tvgLottieExpressions.cpp
Nor-s: @hermet 

Updated to predefine the magic string lengths. 

To avoid increasing the binary size, the string array is still initialized at runtime as pointers into the packed pool.

BinarySize: [+584~630](https://github.com/Nor-s/sizecheck-action-test/pull/11)

hermet (changes requested):
@Nor-s I think there are duplicate string usages, such as in `EXP_CONTENT` or `setFunction(..., "valueAtTime")`. I guess we can reduce the ROM size by eliminating these duplications. Could you please double-check this?

Nor-s:
> @Nor-s I think there are duplicate string usages, such as in `EXP_CONTENT` or `setFunction(..., "valueAtTime")`. I guess we can reduce the ROM size by eliminating these duplications. Could you please double-check this?

@hermet 



I double-checked this. Some duplication does remain between the magic-string registration data and ordinary literals such as `EXP_CONTENT` or `"valueAtTime"`.

I tried reducing that as well, but it ended up increasing either the metadata/pointer cost or the runtime-init code size, so I kept the current layout as the better overall ROM trade-off.

hermet:
fps: 19%+
size: 100

hermet (approved):

PR #4335, opened 2026-04-13
infra: add binary size diff reporter workflow

Nor-s:
## Summary

Adds GitHub Actions workflows to report binary size changes for `libthorvg-1.so.1.0.0` on every pull request.

### `binary_size.yml`

- Check out the pull request as merged into `main`.
- Resolve the current `main` baseline from `origin/main`.
- Build `libthorvg-1.so.1.0.0` from both revisions.
- Strip both binaries.
- Collect the raw `size` output as artifacts.

### `binary_size_comment.yml`

- Download the artifacts from the build workflow.
- Parse the raw `size` outputs and compute the delta.
- Render the markdown summary.
- Post the result as a pull request comment.
  - For pull requests from the same repository, resolve the PR number from workflow context metadata.
  - For pull requests from forks, fall back to resolving the PR by head repository owner and branch.

## Matrix

- x86_64 gcc / clang
- x86 gcc (32-bit)
- arm64 gcc / clang

## Notes

- the comparison reflects the final size impact if the PR were merged into the latest `main`
- the `workflow_run` pattern is used so pull request comments can still be posted for forked PRs
- a new comment is posted on every re-run
- example: 
    - https://github.com/Nor-s/sizecheck-action-test/pull/7
    - https://github.com/Nor-s/sizecheck-action-test/pull/8
    - https://github.com/Nor-s/sizecheck-action-test/pull/5
    - https://github.com/Nor-s/sizecheck-action-test/pull/16

### DIFF
--- a/.github/workflows/binary_size.yml
+++ b/.github/workflows/binary_size.yml
@@ -1,0 +1,185 @@
+name: Binary Size
+
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: ${{ matrix.label }}
+    runs-on: ${{ matrix.runs_on }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # x86_64
+          - label: x86_64 gcc
+            key: x86_64-gcc
+            runs_on: ubuntu-24.04
+            extra_packages: 'libturbojpeg0-dev libpng-dev libwebp-dev libgles-dev'
+            cc: gcc
+            cxx: g++
+            c_args: ''
+            cpp_args: ''
+            c_link_args: ''
+            cpp_link_args: ''
+
+          - label: x86_64 clang
+            key: x86_64-clang
+            runs_on: ubuntu-24.04
+            extra_packages: 'clang lld libomp-dev libturbojpeg0-dev libpng-dev libwebp-dev libgles-dev'
+            cc: clang
+            cxx: clang++
+            c_args: ''
+            cpp_args: ''
+            c_link_args: ''
+            cpp_link_args: ''
+
+          # x86 32-bit 
+          - label: x86 gcc (32-bit)
+            key: x86-gcc
+            runs_on: ubuntu-24.04
+            extra_packages: 'gcc-multilib g++-multilib'
+            cc: gcc
+            cxx: g++
+            c_args: '-m32'
+            cpp_args: '-m32'
+            c_link_args: '-m32'
+            cpp_link_args: '-m32'
+
+          # arm64 (GitHub ARM runner)
+          - label: arm64 gcc
+            key: arm64-gcc
+            runs_on: ubuntu-24.04-arm
+            extra_packages: 'libturbojpeg0-dev libpng-dev libwebp-dev libgles-dev'
+            cc: gcc
+            cxx: g++
+            c_args: ''
+            cpp_args: ''
+            c_link_args: ''
+            cpp_link_args: ''
+
+          - label: arm64 clang
+            key: arm64-clang
+            runs_on: ubuntu-24.04-arm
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_request_size_diff.yml
```diff
@@ -0,0 +1,189 @@
...
+        LIB_NAME="libthorvg-1.so.1.0.0"
+
+        build_thorvg() {
+          local src_dir="$1" build_dir="$2"
+          local meson_args=(
+            -Dloaders="all"
+            -Dengines="cpu, gl"
+            -Dsavers="all"
+            -Dbindings="capi"
+            --optimization=s
```
hermet: Please add: `-Dstatic=true`
`--optimization=s -> -Doptimization=s`

.github/workflows/pull_request_size_diff_comment.yml
```diff
@@ -0,0 +1,181 @@
...
+            pathlib.Path("report.md").write_text("## Binary Size\n\nNo data collected.\n")
+            raise SystemExit(0)
+
+        def delta(base, pr):
+            diff = pr - base
+            pct = (diff / base * 100) if base else 0
+            sign = "+" if diff > 0 else ""
+            return f"{sign}{diff} ({sign}{pct:.2f}%)"
+
+        lines = ["## Binary Size (`libthorvg-1.so.1.0.0`, stripped, `-Os`)", ""]
```
hermet: Let's go simple:

`"## Binary Size (`libthorvg-1.so.1.0.0`, stripped, `-Os`)"`
-> `## Binary Size Report`

.github/workflows/pull_request_size_diff.yml
```diff
@@ -0,0 +1,189 @@
...
+            --errorlogs
```
hermet: Can replace `--errorlogs` with `-Derrorlogs=true`?

hermet (commented):
@Nor-s Please remove these 2 reports:
- Build options
- Compilers

What we normally need is the Delta information. The others are not essential. We can check the details via the GitHub Actions link. Thanks!

Nor-s:
> @Nor-s Please remove these 2 reports:
> 
> * Build options
> * Compilers
> 
> What we normally need is the Delta information. The others are not essential. We can check the details via the GitHub Actions link. Thanks!

Updated the size diff workflow and report as discussed. Thanks for the review.


image


image

image

.github/workflows/pull_request_size_diff_comment.yml
```diff
@@ -0,0 +1,158 @@
...
+        ]
+        for key, sizes in sorted(results.items()):
+            m, p = sizes.get("main"), sizes.get("pr")
+            if not m or not p:
+                lines.append(f"| {key} | n/a | n/a | n/a |")
+                continue
+            lines.append(f"| {key} | {m['total']:,} | {p['total']:,} | {delta(m['total'], p['total'])} |")
+
+        lines += [
+            "", "<details><summary>Detail (text / data)</summary>", "",
```
hermet: `Detail (text / data)` -> `Details`

hermet (commented):
@Nor-s More feedbacks. Thanks.
1. Does it need to be split into two yml files?
2. We'd rather neat name `pull_request size diff` -> `binary size`
3. `Detail (text / data)` -> `Details`

Nor-s:
> * Does it need to be split into two yml files?

@hermet 

Yes, splitting into two YML files is recommended to safely comment on PRs from forks using the workflow_run pattern.

If we don’t need fork PR comments, we can keep it as a single workflow.

reference: https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

image

.github/workflows/pull_request_size_diff_comment.yml
Nor-s: image

Updated. Thanks for feedbacks.

PR #4338, opened 2026-04-15
lottie: support rounded corners for sharp corners

tinyjin:
CubicTo segments with sharp corners (zero in/out tangent) were not being detected for rounding. Added sharpCorner detection and roundCurveCorner() with linear interpolation.

Test file : [Curve with Roundness.json](https://github.com/user-attachments/files/26760670/Curve.with.Roundness.json)

| Before | Patched |
|----------|----------|
|  |  |

---

### Note for bezier rounding implementation.
Initially I implemented corner rounding with [De Casteljau](https://en.wikipedia.org/wiki/De_Casteljau%27s_algorithm) style Curve splitting, which is the usual choice for bezier subdivision, and can simulate accurate curves. However this approach has subtle differences from other players(e.g lottie-web), skottie also has this issue.

| Skottie | [ThorVG](https://github.com/thorvg/thorvg/commit/eee14fd0dc538890e50ff997485108d06ace7b80) |
|--------|-------|
|  |  |

I could conclude that Lottie's round corners uses linear interpolation for vertex placement and circular arc approximation without curve splitting.

This may produce less accurate Bezier curve, but stays consistent with existing Lottie players (lottie-web, lottie-ios, etc.).

### DIFF
--- a/src/loaders/lottie/tvgLottieModifier.cpp
+++ b/src/loaders/lottie/tvgLottieModifier.cpp
@@ -31,6 +31,10 @@ static bool _colinear(const Point* p)
     return tvg::zero(*p - *(p + 1)) && tvg::zero(*(p + 2) - *(p + 3));
 }
 
+static bool _sharpCorner(const Point* p)
+{
+    return tvg::zero(*p - *(p + 1)) && tvg::zero(*(p + 1) - *(p + 2));
+}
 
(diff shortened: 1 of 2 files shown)


hermet (approved):

hermet:
@tinyjin nice job!

PR #4340, opened 2026-04-16
sw_engine: align special blend helpers with raster-local formulas

wenjieshen:
### Summary
This PR aligns the software renderer’s special blend modes with browser-style compositing behavior.

It updates SoftLight and the non-separable blend modes, Hue, Saturation, Color, and Luminosity, to use local raster blend helpers instead of the previous HSL conversion path. It also adds ABGR-specific non-separable blend helpers so ABGR and ARGB targets produce matching logical RGBA results.

### Changes
- Pass the active `SwSurface` into software blend callbacks so blend helpers can use the target color-space channel layout.
- Add surface-specific channel split helpers alongside the existing join helpers.
- Rework Hue, Saturation, Color, and Luminosity blend modes to use luminance/saturation operations instead of RGB/HSL conversion.
- Update SoftLight blending to match the compositing formula.
- Add SW engine coverage for non-separable blend parity between ABGR and ARGB targets, plus a SoftLight reference pixel test.
- The FPS in the Example Lottie maintains a difference of less than 0.2% after 3,000 frames.
### Related Issue
Related to https://github.com/thorvg/thorvg/issues/4190

### Verification
- Added `Non-separable blend colorspace parity` coverage in `test/testSwEngine.cpp`.

image

### DIFF
--- a/src/renderer/cpu_engine/tvgSwCommon.h
+++ b/src/renderer/cpu_engine/tvgSwCommon.h
@@ -26,7 +26,6 @@
 #include <algorithm>
 #include "tvgCommon.h"
 #include "tvgMath.h"
-#include "tvgColor.h"
 #include "tvgRender.h"
 
 #define SW_CURVE_TYPE_POINT 0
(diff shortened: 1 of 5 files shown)


src/renderer/cpu_engine/tvgSwCommon.h
```diff
@@ -599,73 +598,175 @@ static inline uint32_t opBlendHardLight(uint32_t s, uint32_t d)
...
+    b += d;
+    CLIP_COLOR(r, g, b);
+}
+
+static inline float SAT(float r, float g, float b)
+{
+    return (std::max(r, std::max(g, b)) - std::min(r, std::min(g, b)));
+}
+
+static inline float NORM8(uint8_t c)
```
hermet: cpu engine is considered to use integer-based calculations.

src/renderer/cpu_engine/tvgSwCommon.h
```diff
@@ -599,73 +598,175 @@ static inline uint32_t opBlendHardLight(uint32_t s, uint32_t d)
     return BLEND_PRE(JOIN(255, f(C1(s), o.r), f(C2(s), o.g), f(C3(s), o.b)), s, o.a);
 }
 
+static inline float LUM(float r, float g, float b)
```
hermet: tvgSwRaster.cpp

Must be considered with this:

Image

src/renderer/cpu_engine/tvgSwRenderer.cpp
```diff
@@ -303,6 +303,8 @@ bool SwRenderer::target(pixel_t* data, uint32_t stride, uint32_t w, uint32_t h,
     surface->cs = cs;
     surface->channelSize = CHANNEL_SIZE(cs);
     surface->premultiplied = true;
+    surface->blender = nullptr;
```
hermet: not necessary. they are initialized already.

src/renderer/cpu_engine/tvgSwRenderer.cpp
```diff
@@ -538,16 +542,16 @@ bool SwRenderer::blend(BlendMethod method)
             surface->blender = opBlendExclusion;
             break;
         case BlendMethod::Hue:
-            surface->blender = opBlendHue;
+            surface->blender = abgr ? opBlendHueABGR : opBlendHue;
```
hermet: We can’t introduce blending modes for each color format. I don’t really see the need for that.

hermet (changes requested):

hermet:
@wenjieshen We need to consider the lightweight and compact nature of the SW engine. Adding floating conversing, duplicate & conditional logic is not acceptable.

wenjieshen:
Hi @hermet,

I believe I’ve done my best to make this patch more compact and to replace the previous float/HSL path with integer calculations.

There are two points where I would like your opinion:

1. For the non-separable blend modes, the helpers need the surface-specific `split` / `join` functions to handle the target color layout correctly. Because of that, I changed `SwBlender` to receive `const SwSurface*` and passed the surface through the blend-related raster/fill paths. It works, but it touches many signatures, so I’m not sure whether this is the right design or if you would prefer another way to expose `split` / `join`.

2. For SoftLight, I tried both a smaller runtime integer-calculation version and a LUT version. In a rough helper-level benchmark, the LUT version is about 6x faster. I also tested with the Lottie example on the CPU backend, and the FPS increased by about 4%. The binary size increases by about 0.04%. I’m not sure whether this tradeoff is appropriate or whether the smaller non-LUT version would be preferable.

Could you share your opinion on these two points? If you think a different direction would be better, or if you would prefer to take over this part from here, I’m also okay with that. Thank you 🙏 

Appendix: SoftLight tradeoff

The smaller non-LUT version calculates `dc(d)` at runtime:

```text
if d <= 63:
    dc = polynomial(d) / 255^2
else:
    dc = sqrt(d * 255) * 256
```

So it keeps the binary smaller, but the hot path still needs division and square-root calculation.

The LUT version stores the 256 precomputed `dc(d)` values:

```text
dc = dc_table[d]
```

So it avoids the runtime `dc(d)` calculation, but adds a small static table.

hermet:
https://github.com/thorvg/thorvg/pull/4427

PR #4352, opened 2026-04-21
lottie/expressions: scale uint8_t type correctly

tinyjin:
The Lottie parser converts uint8_t property values (e.g. opacity) from 0-100 to 0-255 range via *2.55f scaling, but the expression engine returned results in the original 0-100 range without this conversion.

This caused visual discrepancies such as opacity capping differently.

Apply the same `2.55f` scaling to uint8_t expression results, matching the parser's behavior.

Test file : [auto_fade.json](https://github.com/user-attachments/files/26918133/auto_fade.json)

| Expectation | ThorVG(Main) | ThorVG(Patched) |
|--------|--------|--------|
|  |  |  |

### DIFF
--- a/src/loaders/lottie/tvgLottieExpressions.h
+++ b/src/loaders/lottie/tvgLottieExpressions.h
@@ -54,6 +54,20 @@ struct LottieExpressions
     }
 
     template<typename Property>
+    bool result(float frameNo, uint8_t& out, LottieExpression* exp) {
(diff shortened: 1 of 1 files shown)


Nor-s (approved):
I've confirmed the flickering issue is resolved, and no regressions were found in the LottieExpressions and LottieInteraction examples. Thank you!

hermet (approved):

PR #4353, opened 2026-04-21
lottie/expressions: free jerry_value_t on eval exception path

tinyjin:
jerry_eval() returns a jerry_value_t that must be freed by the caller regardless of whether it is an exception. The exception path was missing jerry_value_free(eval), leaking the value on every failed expression evaluation.

### DIFF
--- a/src/loaders/lottie/tvgLottieExpressions.cpp
+++ b/src/loaders/lottie/tvgLottieExpressions.cpp
@@ -1494,6 +1494,7 @@ jerry_value_t LottieExpressions::evaluate(float frameNo, LottieExpression* exp)
 
     if (jerry_value_is_exception(eval)) {
         TVGERR("LOTTIE", "Failed to dispatch the expressions!");
+        jerry_value_free(eval);
         exp->disabled = true;
         return jerry_undefined();
     }


Nor-s (approved):
lgtm

hermet (approved):

hermet:
> arm64-gcc	874,074	23,088	878,170	23,088	+4096 (+0.46%

@Nor-s  +4kb looks strange...

Nor-s:
> > arm64-gcc	874,074	23,088	878,170	23,088	+4096 (+0.46%
> 
> @Nor-s  +4kb looks strange... 


I'll check it on Ubuntu ARM64 VM.

hermet:
@tinyjin Did you have any specific issues with this?

Nor-s:
> > arm64-gcc	874,074	23,088	878,170	23,088	+4096 (+0.46%
> 
> @Nor-s +4kb looks strange...

@hermet 

image

The PR itself only increases LottieExpressions::evaluate() by about 8 bytes.

On Ubuntu arm64 with GCC 13, that small change probably changes the linker layout and causes .text grow by 4096 bytes.

This does not reproduce with GCC 15(diff 0byte), 
so it looks like a GCC 13 / AArch64 linker layout side effect.

Nor-s:
image

Also, it seems like GCC 13 on ARM64 tends to be unstable BinarySize side.

I've noticed similar issues in other PRs as well. Would it be worth excluding it?


- https://github.com/thorvg/thorvg/pull/4340

hermet:
> so, it seems like GCC 13 on ARM64 tends to be unstable BinarySize side.
> 
> I've noticed similar issues in other PRs as well. Would it be worth excluding it?

@Nor-s How about upgrade GCC15 in CI?

PR #4355, opened 2026-04-21
lottie: Parse "shapes" property only for Shape layers

LizzaM1net:
LottieFiles spec says only Shape layer has "shape" property.
https://lottiefiles.github.io/lottie-docs/layers/#shape-layer

It could lead to crashes because e.g. Image layer could get children other than LottieImage, which will lead to crash because of static_cast
https://github.com/thorvg/thorvg/blob/9ab446ccbeb2066fc51eed24c82957270f39e1c6/src/loaders/lottie/tvgLottieBuilder.cpp#L951

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -1558,7 +1558,10 @@ LottieLayer* LottieParser::parseLayer(LottieLayer* precomp)
             layer->transform = parseTransform(ddd);
         }
         else if (KEY_AS("ao")) layer->autoOrient = getInt();
-        else if (KEY_AS("shapes")) parseShapes(layer->children);
+        else if (KEY_AS("shapes")) {
+            if (layer->type == LottieLayer::Shape)
+                parseShapes(layer->children);
+        }
         else if (KEY_AS("ip")) layer->inFrame = getFloat();
         else if (KEY_AS("op")) layer->outFrame = getFloat();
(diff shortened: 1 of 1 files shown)


hermet (changes requested):
Image should not have children. User needs to fix those lottie asset. This just hide an issue.

LizzaM1net:
@hermet, so ThorVG should crash when processing malformed asset by design?

hermet:
@LizzaM1net Where did you get that resource? Malformed Lottie content is not acceptable. Please fix it or report the issue to the source. Hiding the issue in ThorVG could lead to more difficult situations, especially if it results in broken or unexpected visuals. Crashes are often useful for identifying the root cause very quickly.

> @hermet, so ThorVG should crash when processing malformed asset by design?

No but not with this kind of case-by-case handling. We will see. btw, thanks for your contribution.

LizzaM1net:
We're here trying to integrate ThorVG to our Qml toolkit to display animations. Applications we are developing needs to be failsafe, but whole animation engine in our apps is good attack vector. We need engine to not crash or hang even if malformed asset was loaded by attacker.

We have amount of manually distorted files to test lottie engines with. ThorVG successfully loaded much more malformed assets then rlottie, so I think ThorVG already silently processes some malformed assets. I can attach all our files if you find this helpful

I also agree that we should not hide information about asset errors from developer. Maybe we could write any errors that happened to log together with loading safe part of asset?

hermet:
https://github.com/thorvg/thorvg/pull/4377

Issue #4107, opened 2026-01-09
svg: improve the text support


hermet:
The following elements/tags related to Text are supported for addition in SVG.

- [x] `tspan`
- [x] `text-anchor`
- [x] `dx`, `dy`
- [x] `alignment-baseline`
- [x] `dominant-baseline`
- [x] `font-weight`
- [x] `letter-spacing`
- [x] `word-spacing`

_Originally posted by @hermet in https://github.com/thorvg/thorvg/discussions/4101#discussioncomment-15451092_

JSUYA:
A quick review...
- `letter-spacing` can be implemented using the `spacing()` API.
- `text-anchor` can be implemented using the `align(x)` API.
- For `dominant-baseline`, I expect that we can support `auto`, middle (or central), and 'hanging'. However, because it relies on the font's baseline-table, implementing it with the align(y) API may be difficult.

Other than that, the Text class should support the functionality.
In particular, `dx`( and `dy`?) can define properties for each glyph.
 (ex `<text dx="5 5 5 10%" x="10%" y="50%">SVGLOADER</text>)`

PR #4357, opened 2026-04-22
svg_loader: add text-anchor attribute support

JSUYA:
Parse the text-anchor attribute and apply it using the Text::align()
The anchor value is stored on SvgStyleProperty as a float (0=start, 0.5=middle, 1=end).

https://github.com/thorvg/thorvg/issues/4107


JSUYA:
test SVG
text_anchor_spec


thorvg
image

JSUYA:
Once tspan support is merged, it should also be supported to tspan. I plan to add it in a new PR at that time.

src/loaders/svg/tvgSvgBuilder.cpp
```diff
@@ -922,6 +922,8 @@ static Paint* _textBuildHelper(SvgParserContext& ctx, const SvgNode* node, const
     text->text(processedText);
     tvg::free(processedText);
 
+    if (node->style->textAnchor > 0.0f) text->align(node->style->textAnchor, 0.0f);
+
```
hermet: just api call is enough without conditional check.

hermet (approved):

PR #4358, opened 2026-04-22
loader: update loader functions to accept `LoaderOps` parameter

Chubercik:
When [updating ThorVG to version 1.0.4](https://github.com/godotengine/godot/pull/118784), I came across a compilation error - it seemed like there was an API mismatch between `.cpp` and `.h` files; indeed, adding the new `LoaderOps` parameter solves the issue.

I'm not sure what the reason for this issue is. Was it simply overlooked in the commit https://github.com/thorvg/thorvg/commit/06b0c19eead79c3e1c66b1738ac5bf6bb10f60cc?

### DIFF
--- a/src/renderer/tvgLoader.cpp
+++ b/src/renderer/tvgLoader.cpp
@@ -266,7 +266,7 @@ bool LoaderMgr::retrieve(Loader* loader)
     return true;
 }
 
-tvg::Loader* LoaderMgr::loader(const char* filename, bool* invalid)
+tvg::Loader* LoaderMgr::loader(const char* filename, const LoaderOps* ops, bool* invalid)
 {
 #ifdef THORVG_FILE_IO_SUPPORT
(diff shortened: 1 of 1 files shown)


hermet:
@Chubercik Hi, please check [this](https://github.com/thorvg/thorvg/pull/4361). `tvgLoader.cpp` is actually redundant and is not used in the build. If you are using a custom build configuration, you should use `tvgLoaderMgr.cpp` instead. Thanks.

Chubercik:
> @Chubercik Hi, please check [this](https://github.com/thorvg/thorvg/pull/4361). `tvgLoader.cpp` is actually redundant and is not used in the build. If you are using a custom build configuration, you should use `tvgLoaderMgr.cpp` instead. Thanks.

I see, that makes sense.. Thanks for the quick reply and sorry for the false alarm 😅

Issue #3258, opened 2025-02-20
text: otf format cannot be loaded


tinyjin:
Somewhat, otf file cannot be loaded and used. Seems like the file data is wrongly loaded by JPG Loader?

![Image](https://github.com/user-attachments/assets/8b47c717-fa51-415a-84a8-0f11eb4173ed)

![Image](https://github.com/user-attachments/assets/6f1ec456-2fdb-4e29-9e6a-cae0db0f903f)

hermet:
Note: thorvg doesn't support the otf officially.

tinyjin:
[DMSans Regular.otf.zip](https://github.com/user-attachments/files/18885907/DMSans.Regular.otf.zip)

Sharing font data i've tested with, just in case :)

PR #4370, opened 2026-04-29
otf: support opentype font

hermet:
- support parsing CFF table for Name, Top DICT, CharStrings
- parse CharStrings bytecode, acquire the outline info
- transform outlines to thorvg render paints
- TODO: composite glyph / kerning implementation
- otf is disabled by default.

otf font support must be enabled through the build configuration by
"loaders=otf", the subsequent define is "THORVG_OTF_LOADER_SUPPORT"

issue: #3397 #3258

image

image


hermet:
+700 bytes when otf is disabled.

Issue #4369, opened 2026-04-27
svg: Image not being shown inside pattern


thatoneham:
the png in the original svg doesnt show at all.

<img width="831" height="824" alt="Image" src="https://github.com/user-attachments/assets/21be2867-329d-4392-bfa0-20cc18c8874e" />

whilst one without the pattern tag (i removed it and set matrix directly) does

<img width="814" height="821" alt="Image" src="https://github.com/user-attachments/assets/bc4d33e8-b036-4e4e-a1dc-d6f895aea4a3" />

here are the 2 svgs

<img width="300" height="216" alt="Image" src="https://github.com/user-attachments/assets/a3ed5413-3efb-40c8-a211-e52cd70c9b67" />
<img width="300" height="216" alt="Image" src="https://github.com/user-attachments/assets/a4aaa230-8d55-4506-b0e9-43954dcfcd90" />

hermet:
\+ @JSUYA

JSUYA:
```
[L] SVG (../src/loaders/svg/tvgSvgBuilder.cpp 438): The fill's url not supported.
```
https://github.com/thorvg/thorvg/blob/main/src/loaders/svg/tvgSvgBuilder.cpp#L396


This image uses unsupported fill' URL feature and pattern elements.
also, there is an issue where the image's MIME TYPE is not recognized because it is written in uppercase.
(This can be fixed immediately with a simple patch.)

However, the pattern is a different matter. It may take a little more time.

hermet:
relate: https://github.com/thorvg/thorvg/issues/1319 https://github.com/thorvg/thorvg/issues/1816

PR #4371, opened 2026-04-30
svg_loader: accept image data mime case variants

JSUYA:
Also, allow to the encoding type and "image" string uppercase.

+https://github.com/thorvg/thorvg/issues/4369


hermet:
@tinyjin @Nor-s It seems lottie also needs this change.

Issue #4381, opened 2026-05-06
cpu: AntiAliasing option support


huming2207:
Hi @hermet and @tinyjin 

I've got a new e-ink Spectra E6 display with my own ESP32 driver board and got it working with ThorVG last night. It's a new one with higher resolution, not the 7.3 inch 800x480 panels. But it's the same as the old one that as 1-bit monochrome black + white + green + red + yellow + blue colour:

<img width="768" height="1024" alt="Image" src="https://github.com/user-attachments/assets/ff92d3c6-111f-4184-9243-f4a2b226a63b" />

It works mostly fine with my own vibe-coded dithering algorithm with ThorVG, see: https://github.com/huming2207/inkframe-fw/blob/26b6f18bcdcf9a5d99f6d8774486bfe68c7698ab/main/display_manager.cpp#L148

But there's a issue if I look closer - the edges are not "smooth" because dithering algorithm and the 1-bit monochrome will not handle the antialiasing properly:

<img width="762" height="779" alt="Image" src="https://github.com/user-attachments/assets/43faf2ab-ad3e-4a59-a7ec-f8c481fddf62" />

I went to ask the DeepWiki chatbot and I think it gives me the wrong answer. It recommends me to use `Picture::filter()` with `FilterMethod::Nearest` but that won't disable the antialiasing completely.

If this anti-aliasing toggle API doesn't really exist, maybe I will give it a try and implement my own. What do you think?

Regards,
Jackson

huming2207:
Seems like I might be able to modify my dithering algorithm by making use the alpha channel - at least DeepSeek told me so 😅:

<img width="1023" height="749" alt="Image" src="https://github.com/user-attachments/assets/d84780e3-42e2-49c4-8958-c14ec56c48c6" />

I will have a try tonight.

I also think it might be not common for most of the use cases of this library with AA disabled, except someone with a monochrome screen like me. I will try poking around my dithering pipeline first. If it still looks bad then I will upstream a patch to add AA toggle.

hermet:
@huming2207 Hi, for quick test, you can forcely turn off the aa here:

`tvgSwRenderer.cpp`
<img width="1731" height="262" alt="Image" src="https://github.com/user-attachments/assets/6ce5c781-d9d7-422b-8190-dc770141296e" />


@wenjieshen Regardless of this issue, What do you think about supporting enable/disable AA on the GPU engine side? any cons or concern points? Assuming the API look like this:
```
[...]

PR #4382, opened 2026-05-07
engine: add Aliased option for SW (Experimental APIs)

Nor-s:
issue: #4381

### Summary

```cpp
enum struct EngineOption : uint8_t
{
    None = 0,                    /**< No engine options are enabled. This may be used to explicitly disable all optional behaviors. */
    Default = 1 << 0,            /**< Uses the default rendering mode. */
    SmartRender = 1 << 1,        /**< Enables automatic partial (smart) rendering optimizations. */
    Aliased = 1 << 2     /**< Disables anti-aliased rendering @note Experimental API */
};
```
- Introduce a Aliased option for the software renderer

```
C++ API
 + enum struct EngineOption:: Aliased

C API
 + TVG_ENGINE_OPTION_ALIASED
```

| AA Off | AA On |
|-|-|
|image|image|

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -133,9 +133,10 @@ enum struct ColorSpace : uint8_t
  */
 enum struct EngineOption : uint8_t
 {
-    None = 0,                   /**< No engine options are enabled. This may be used to explicitly disable all optional behaviors. */
-    Default = 1 << 0,           /**< Uses the default rendering mode. */
-    SmartRender = 1 << 1        /**< Enables automatic partial (smart) rendering optimizations. */
+    None = 0,                    /**< No engine options are enabled. This may be used to explicitly disable all optional behaviors. */
+    Default = 1 << 0,            /**< Uses the default rendering mode. */
+    SmartRender = 1 << 1,        /**< Enables automatic partial (smart) rendering optimizations. */
+    Aliased = 1 << 2             /**< Disables anti-aliased rendering. @note Experimental API */
 };
 
 
--- a/src/renderer/cpu_engine/tvgSwRenderer.cpp
+++ b/src/renderer/cpu_engine/tvgSwRenderer.cpp
@@ -101,7 +101,7 @@ struct SwShapeTask : SwTask
        Additionally, the stroke style should not be dashed. */
     bool antialiasing(float strokeWidth)
     {
-        return strokeWidth < 2.0f || rshape->stroke->dash.count > 0 || rshape->stroke->first || rshape->trimpath() || rshape->stroke->color.a < 255;
+        return renderer->antiAlias && (strokeWidth < 2.0f || rshape->stroke->dash.count > 0 || rshape->stroke->first || rshape->trimpath() || rshape->stroke->color.a < 255);
     }
 
     float validStrokeWidth(bool clipper)
@@ -150,7 +150,7 @@ struct SwShapeTask : SwTask
         if (updateShape || flags[0] & RenderUpdateFlag::Stroke) {
             if (strokeWidth > 0.0f) {
                 shapeResetStroke(shape, rshape, transform, renderer->mpool, tid);
-                if (!shapeGenStrokeRle(shape, rshape, transform, clipBox, curBox, renderer->mpool, tid)) goto err;
+                if (!shapeGenStrokeRle(shape, rshape, transform, clipBox, curBox, renderer->mpool, tid, renderer->antiAlias)) goto err;
                 if (auto fill = rshape->strokeFill()) {
                     auto ctable = (flags[0] & RenderUpdateFlag::GradientStroke) ? true : false;
                     if (ctable) shapeResetStrokeFill(shape);
@@ -925,5 +925,7 @@ SwRenderer::SwRenderer(uint32_t threads, EngineOption op)
 
     mpool = mpoolReq();
 
-    if (op == EngineOption::None) dirtyRegion.support = false;
+    auto byDefault = (op == EngineOption::Default);
+    dirtyRegion.support = (byDefault || (op & EngineOption::SmartRender));
+    antiAlias = (byDefault || !(op & EngineOption::Aliased));
 }
--- a/src/renderer/cpu_engine/tvgSwRenderer.h
+++ b/src/renderer/cpu_engine/tvgSwRenderer.h
@@ -77,13 +77,14 @@ struct SwRenderer : RenderMethod
     static bool term();
 
     SwSurface*           surface = nullptr;           // active surface
-    SwMpool* mpool;                                   // designated memory pool
+    SwMpool*             mpool;                       // designated memory pool
+    bool                 antiAlias;                   // anti-aliasing support
 
 private:
+    bool                 fulldraw = true;             //buffer is cleared (need to redraw full screen)
     Array<SwTask*>       tasks;                       //async task list
     Array<SwSurface*>    compositors;                 //render targets cache list
     RenderDirtyRegion    dirtyRegion;                 //partial rendering support
-    bool                 fulldraw = true;             //buffer is cleared (need to redraw full screen)
 
     ~SwRenderer();
 
--- a/src/renderer/tvgCanvas.cpp
+++ b/src/renderer/tvgCanvas.cpp
@@ -207,7 +207,8 @@ GlCanvas* GlCanvas::gen(EngineOption op) noexcept
 {
 #ifdef THORVG_GL_ENGINE_SUPPORT
     if (engineInit > 0) {
-        if (op == EngineOption::SmartRender) TVGLOG("RENDERER", "GlCanvas doesn't support Smart Rendering");
+        if (op & EngineOption::SmartRender) TVGLOG("RENDERER", "GlCanvas doesn't support Smart Rendering");
+        if (op & EngineOption::Aliased) TVGLOG("RENDERER", "GlCanvas doesn't support Aliased");
         auto renderer = GlRenderer::gen(TaskScheduler::threads(), op);
         if (!renderer) return nullptr;
         renderer->ref();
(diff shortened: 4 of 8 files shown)


Nor-s:
```cpp
SwRle* rleRender(SwRle* rle, const SwOutline* outline, const RenderRegion& bbox, SwMpool* mpool, unsigned tid, bool antiAlias);


bool shapeGenStrokeRle(SwShape& shape, const RenderShape* rshape, const Matrix& transform, const RenderRegion& clipBox, RenderRegion& renderBox, SwMpool* mpool, unsigned tid)
{
    SwOutline* shapeOutline = nullptr;

    //Dash style with/without trimming
    if (rshape->stroke->dash.length > DASH_PATTERN_THRESHOLD) {
        shapeOutline = _genDashOutline(rshape, transform, mpool, tid, rshape->trimpath());
    //Trimming & Normal style
    } else {
        shapeOutline = shape.outline ? shape.outline : _genOutline(shape, rshape, transform, mpool, tid, false, rshape->trimpath());
    }

    if (!shapeOutline) return false;

    if (!strokeParseOutline(shape.stroke, *shapeOutline, mpool, tid)) return false;

    auto strokeOutline = strokeExportOutline(shape.stroke, mpool, tid);
    auto ret = mathUpdateOutlineBBox(strokeOutline, clipBox, renderBox, false);
    if (ret) shape.strokeRle = rleRender(shape.strokeRle, strokeOutline, renderBox, mpool, tid, true);

    return ret;
}

// static void _horizLine(RleWorker& rw, int32_t x, int32_t y, int32_t area, int32_t aCount)
    if (!rw.antiAlias) coverage = 255;
```

Nor-s:
AntiAliasing
Default Option -> 0xff  

or

DisableAntiAliasing
Default Option -> 0xff & (~DisableAntiAliasing)

Nor-s:
Users use `EngineOption::SmartRender` as a single option. 

---

thorvg.web: https://github.com/thorvg/thorvg.web/blob/9db1aa21605ab3d0259edde39b3b422aa42fc9c3/wasm/webcanvas/tvgWasmWebCanvas.cpp#L70

rive-thorvg: https://github.com/xseignard/rive-thorvg/blob/208c68c9b638949c49ce3308a03632105ba3ba43/src/rive_thorvg.cpp#L237

TinyPixOS: https://github.com/TinyPiXOS/TinyPiXOS/blob/c89d097e5c3517da2750a012c06fc252fc295a34/src/src/TpGUI/GUI/TpScreen.cpp#L70

Lottie-godot: https://github.com/byourself22/Lottie-godot/blob/2a6989b20dffd503f93084c99661c75dc9ded005/src/lottie_animation.cpp#L1719

---

Changing the options to a bitfield has the potential to break many things, 
and I don't think this is something to consider in the current.

Therefore, for API compatibility, at this stage we should add a separate `DisableAntiAliasing` option instead.

This option behaves as if AntiAliasing is turned off from the Default option.

Nor-s:
@hermet 

While adding an `AntiAliasing` option, I initially considered making `EngineOption` work as a bit flag API. 

ex) `EngineOption::SmartRender | EngineOption::AntiAliasing`, `EngineOption::AntiAliasing`, `EngineOption::SmartRender`

However, I changed the approach to adding a `DisableAntiAliasing` option instead.
  
  
- Some projects may already use `EngineOption::SmartRender` as a standalone option.
    - Treating `EngineOption` as a full bit flag API at this stage may affect existing users unexpectedly.
    - Keeping anti-aliasing enabled for `SmartRender` could work, but the behavior would be less explicit.
- A full bit flag design should probably be reviewed in a future version(Major or Minor version update), including:
    - scoped enum(enum struct) vs plain enum
    - whether to provide public bitwise operators: `SmartRender`,  `AntiAliasing`, ETC... should compose 
 
For now, `DisableAntiAliasing` is more explicit and keeps the existing option behavior compatible.

src/renderer/cpu_engine/tvgSwRenderer.cpp
```diff
@@ -90,6 +90,7 @@ struct SwShapeTask : SwTask
     SwShape shape;
     const RenderShape* rshape = nullptr;
     bool clipper = false;
+    bool antiAliasing = true;
```
hermet: This could be a Surface level variable. You can store the value in `SwRenderer` & `SwSurface`, not shape level, to save memory.

hermet (changes requested):
@Nor-s Please check comments. Thanks!

inc/thorvg.h
```diff
@@ -133,9 +133,10 @@ enum struct ColorSpace : uint8_t
  */
 enum struct EngineOption : uint8_t
 {
-    None = 0,                   /**< No engine options are enabled. This may be used to explicitly disable all optional behaviors. */
-    Default = 1 << 0,           /**< Uses the default rendering mode. */
-    SmartRender = 1 << 1        /**< Enables automatic partial (smart) rendering optimizations. */
+    None = 0,                    /**< No engine options are enabled. This may be used to explicitly disable all optional behaviors. */
+    Default = 1 << 0,            /**< Uses the default rendering mode. */
+    SmartRender = 1 << 1,        /**< Enables automatic partial (smart) rendering optimizations. */
+    DisableAntiAliasing = 1 << 2 /**< Disables anti-aliased rendering from the default rendering mode. @note Experimental API */
```
hermet: Let's go shorter name ->`Aliased`

src/renderer/cpu_engine/tvgSwRenderer.h
```diff
@@ -77,13 +77,14 @@ struct SwRenderer : RenderMethod
     static bool term();
 
     SwSurface*           surface = nullptr;           // active surface
-    SwMpool* mpool;                                   // designated memory pool
+    SwMpool*             mpool;                       // designated memory pool
+    bool                 antiAlias = true;            // anti-aliasing support
 
 private:
+    bool                 fulldraw = true;             //buffer is cleared (need to redraw full screen)
```
Nor-s, line 84: for memory alignment

inc/thorvg.h
```diff
@@ -133,9 +133,10 @@ enum struct ColorSpace : uint8_t
  */
 enum struct EngineOption : uint8_t
 {
-    None = 0,                   /**< No engine options are enabled. This may be used to explicitly disable all optional behaviors. */
-    Default = 1 << 0,           /**< Uses the default rendering mode. */
-    SmartRender = 1 << 1        /**< Enables automatic partial (smart) rendering optimizations. */
+    None = 0,                    /**< No engine options are enabled. This may be used to explicitly disable all optional behaviors. */
+    Default = 1 << 0,            /**< Uses the default rendering mode. */
+    SmartRender = 1 << 1,        /**< Enables automatic partial (smart) rendering optimizations. */
+    Aliased = 1 << 2             /**< Disables anti-aliased rendering from the default rendering mode. @note Experimental API */
```
hermet: ->  `Disables anti-aliased rendering. @note Experimental API`

src/renderer/cpu_engine/tvgSwRenderer.cpp
```diff
@@ -101,6 +101,7 @@ struct SwShapeTask : SwTask
        Additionally, the stroke style should not be dashed. */
     bool antialiasing(float strokeWidth)
     {
+        if (!renderer->antiAlias) return false;
```
hermet: `return renderer->antiAlias && (strokeWidth < 2.0f || rshape->stroke->dash.count > 0 || rshape->stroke->first || rshape->trimpath() || rshape->stroke->color.a < 255);`

src/renderer/tvgCanvas.cpp
```diff
@@ -208,6 +208,7 @@ GlCanvas* GlCanvas::gen(EngineOption op) noexcept
 #ifdef THORVG_GL_ENGINE_SUPPORT
     if (engineInit > 0) {
         if (op == EngineOption::SmartRender) TVGLOG("RENDERER", "GlCanvas doesn't support Smart Rendering");
+        if (op == EngineOption::Aliased) TVGLOG("RENDERER", "GlCanvas doesn't support disabling Anti-Aliasing");
```
hermet: `GlCanvas doesn't support Aliased`

src/renderer/cpu_engine/tvgSwRenderer.h
```diff
@@ -77,13 +77,14 @@ struct SwRenderer : RenderMethod
     static bool term();
 
     SwSurface*           surface = nullptr;           // active surface
-    SwMpool* mpool;                                   // designated memory pool
+    SwMpool*             mpool;                       // designated memory pool
+    bool                 antiAlias = true;            // anti-aliasing support
```
hermet: no need to initialize this here, constructor does.

src/renderer/cpu_engine/tvgSwRenderer.cpp
```diff
@@ -926,4 +927,5 @@ SwRenderer::SwRenderer(uint32_t threads, EngineOption op)
     mpool = mpoolReq();
 
     if (op == EngineOption::None) dirtyRegion.support = false;
```
hermet: It must be look like:
```
if (op == EngineOption::Default || (op & EngineOption::SmartRender)) dirtyregion.support = true;  //you can ma ke the value false by default.
if (op == EngineOption::Default || !(op & EngineOption::Aliased)) antiAliased = true
```

You might wanna add necessary bitoperation in `tvgRender.h` ie)

Image

hermet (changes requested):
@Nor-s Please check comments. Thanks!

src/renderer/cpu_engine/tvgSwRenderer.cpp
```diff
@@ -925,5 +925,8 @@ SwRenderer::SwRenderer(uint32_t threads, EngineOption op)
 
     mpool = mpoolReq();
 
-    if (op == EngineOption::None) dirtyRegion.support = false;
+    auto hasOption = [&](EngineOption option) { return (static_cast<uint8_t>(op) & static_cast<uint8_t>(option)) != 0; };
```
hermet: Please push the operator overriding in the `tvgRender.h`.  for other engines use it. 

```
  auto byDefault = (op == EngineOption::Default)
  dirtyRegion.support = (byDefault || op & EngineOption::SmartRender);
```

see also:

Image

hermet (commented):
@Nor-s

hermet (approved):

PR #4385, opened 2026-05-11
lottie: fix stale curSlot after slot del()

theashraf:
Commit 8af22a30 added an early-return in `apply()` guarded by `curSlot == slotcode`, but `del()` does not invalidate `curSlot`. If `gen()` is later called with byte-identical JSON, djb2 returns the same code and the next `apply()` short-circuits without applying the slot.

Reset `curSlot` in `del()` when it matches the slot being removed.


theashraf:
Reproducer (self-contained C++, uses ThorVG's existing `test/resources/slot.lot`): https://gist.github.com/theashraf/526492493c9dbf272fd4fea9029fadf2

Build & run instructions are in the file's top comment. Exit code is 1 on a buggy build and 0 with this patch.

hermet (approved):

PR #4389, opened 2026-05-13
infra: create or update binary size report comment

Nor-s:
### Summary

- Run build, test, regression, and binary size workflows only when relevant paths change.

- Binary size reporting now uses create-or-update comments instead of creating a new PR comment on each run.

### ex) Binary Size Report

The binary size report is updated in place:

https://github.com/Nor-s/sizecheck-action-test/pull/23#issuecomment-4439449963

image

### ex) When Updating Only `.gitignore`

The workflows are skipped when a PR changes only unrelated files.

image

### DIFF
--- a/.github/workflows/binary_size_report.yml
+++ b/.github/workflows/binary_size_report.yml
@@ -82,11 +82,12 @@ jobs:
         pathlib.Path("report.md").write_text("\n".join(lines) + "\n")
         PY
 
-    - name: Post comment
+    - name: Create or update comment
       uses: actions/github-script@v7
       with:
         script: |
           const fs = require('fs');
+          const reportHeader = '## Binary Size Report';
           const body = fs.readFileSync('report.md', 'utf8');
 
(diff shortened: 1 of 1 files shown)


.github/workflows/binary_size.yml
```diff
@@ -4,6 +4,14 @@ on:
   pull_request:
     branches:
       - main
+    paths:
```
hermet: @Nor-s this path introduction is essential?

Nor-s: @hermet 
This isn't essential, but I thought it would be preferable for CI not to run 
when the changes aren't in those paths

hermet: Let's remove them, that’s more burdensome for us, since we need to maintain the proper dependencies whenever files or folders change.

Nor-s: @hermet 
I've updated it. Thanks for the feedback!

PR #4399, opened 2026-05-19
gl_engine: reset GL library handles after unload

Nor-s:
Without clearing these handles, subsequent glInit() calls cannot reload the GL/EGL libraries after glTerm() unloads them.

This can happen when tests repeatedly call init and term.

### DIFF
--- a/src/renderer/gpu_engine/gl/tvgGl.cpp
+++ b/src/renderer/gpu_engine/gl/tvgGl.cpp
@@ -884,15 +884,31 @@ bool glInit()
 bool glTerm()
 {
 #if defined(_WIN32) && !defined(__CYGWIN__)
-    if (_libGL) FreeLibrary(_libGL);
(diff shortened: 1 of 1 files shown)


src/renderer/gpu_engine/gl/tvgGl.cpp
```diff
@@ -885,16 +885,24 @@ bool glTerm()
 {
 #if defined(_WIN32) && !defined(__CYGWIN__)
     if (_libGL) FreeLibrary(_libGL);
+    _libGL = nullptr;
```
hermet:
```
 if (_libGL) {
    FreeLibrary(_libGL);
    libGL = nullptr;
}
```

wenjieshen (approved):
Thank you for the contribution.
I didn't encounter any crashes during execution of the examples on Win32 and Ubuntu.

Issue #4354, opened 2026-04-21
engines: stroke rendering


hermet:
Stroke transform looks incorrect at the end of the rendering.

GL/CPU
<img width="2097" height="1074" alt="Image" src="https://github.com/user-attachments/assets/f7a9f2a5-58d8-47ae-a0d6-df100467fc93" />

Expected:

<img width="1042" height="814" alt="Image" src="https://github.com/user-attachments/assets/480444fd-6c9a-4c1d-9463-611391fa7f62" />

[Untitled file (28).json](https://github.com/user-attachments/files/26920432/Untitled.file.28.json)

Issue #4287, opened 2026-04-02
cpu: replace integer-based float simulation with native floating-point operations


hermet:
The current software (SW) engine uses an integer-based approach to simulate floating-point operations. While this method can provide deterministic behavior and potential performance benefits in certain environments, it introduces **additional complexity and limits numerical precision**.

This issue proposes evaluating and potentially replacing the integer-based floating-point simulation with native floating-point operations.

### Consideration
Native floating-point operations may not always outperform integer-based implementations, especially in embedded environments. However, in practice, the most performance-critical workload is likely in rasterization and buffer writes rather than arithmetic operations. Additionally, modern systems generally include hardware FPU support, which reduces the performance gap between integer and floating-point computations.

### Expected Benefits
- Reduced binary size, significant code maintenance benefit.
- scalable arithmetic data range.

hermet:
related issue: https://github.com/thorvg/thorvg/issues/4284#issuecomment-4180725261 https://github.com/thorvg/thorvg/issues/3842

Issue #3593, opened 2025-07-05
sw_engine: stroke transforming issue


hermet:
The original aspect ratio of the stroke is not preserved when non-uniform (different X and Y) scaling is applied.

[For Hermet.json](https://github.com/user-attachments/files/21070327/For.Hermet.json)

https://github.com/user-attachments/assets/b28688c9-8cb2-4069-b47b-a0fb8b1faf66

Issue #2758, opened 2024-09-23
sw_engine: wrong stroke roundness


mgrudzinska:
sample:
```
<svg width="1600" height="600" xmlns="http://www.w3.org/2000/svg">
   <rect x="100" y="100" width="600" height="400" fill="none" stroke="#00f" stroke-width="100"/>
   <rect x="100" y="100" width="600" height="400" fill="none" stroke="#F00" stroke-width="100" rx="150" ry="3"/>
</svg>
```

target 800x800, no scaling applied

result:
<img width="500" alt="Zrzut ekranu 2024-09-23 o 14 17 09" src="https://github.com/user-attachments/assets/60284b5f-c5e3-4a6b-a15a-74a7919296ef">

expected:
<img width="250" alt="Zrzut ekranu 2024-09-23 o 14 18 39" src="https://github.com/user-attachments/assets/e060cf94-7c7f-43ee-8f76-5b7acd60c171">

PR #4405, opened 2026-05-20
cpu_engine: path processing revision

hermet:
- replace the integer-based 16:16 fixed-point computation with floating-point, for more scalable data size

- revised the path processing pipeline for generate the stroke geometry before applying the transform for consistent rendering output

issue: #4354, #3593, #2758, #1925, #954, #4287


hermet:
**FPS Compare**
- `Lottie`: **+2.23%**
- `LottieBig`: **+2.16%**
- `SmartRender`: **+1.57%**
- `ManyRects`: **+9.53%**

**Lottie Peak Memory** 
- in the error margin (142mb -> 141mb)

PR #4406, opened 2026-05-21
lottie: copy embedded font data to avoid use-after-free

theashraf:
When a Lottie embeds a font, `LottieFont::prepare()` loads it with `copy=false`, so the font loader only keeps a pointer to the `LottieFont`'s `b64src` buffer instead of owning a copy.

That loader is cached globally and shared by font name across animations, so it can outlive the `LottieFont` that owns the data. Once that animation is destroyed, `b64src` is freed and the cached loader is left pointing at freed memory — any other animation still using the same font then renders missing or garbled glyphs.

Passing `copy=true` lets the loader own its own copy of the font data, so it no longer depends on the animation's lifetime.

The in-memory `Text::load(name, data, ...)` overload also leaked the cached loader's `sharing` count on repeated loads of the same font name — the file-path overload already compensates for this, but this one did not. With `copy=true` that would keep the copied font data pinned and stop `term()`/unload from ever freeing the loader, so the second commit drops the extra reference when an existing cached loader is reused.

Reproducible with multiple players sharing an embedded font: LottieFiles/dotlottie-web#877

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -301,7 +301,7 @@ void LottieFont::prepare()
 {
     if (!b64src) return;
 
-    Text::load(name, b64src, size, mime, false);
+    Text::load(name, b64src, size, mime, true);
 }
 
 
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -301,7 +301,10 @@ void LottieFont::prepare()
 {
     if (!b64src) return;
 
-    Text::load(name, b64src, size, mime, true);
+    if (Text::load(name, b64src, size, mime, true) == Result::Success) {
+        tvg::free(b64src);
+        b64src = nullptr;
+    }
 }
 
 
--- a/src/renderer/tvgText.cpp
+++ b/src/renderer/tvgText.cpp
@@ -75,7 +75,9 @@ Result Text::load(const char* name, const char* data, uint32_t size, const char*
     }
 
     LoaderOps ops = {Type::Text};
-    if (!LoaderMgr::loader(name, data, size, mimeType, &ops, copy)) return Result::NonSupport;
+    auto loader = LoaderMgr::loader(name, data, size, mimeType, &ops, copy);
+    if (!loader) return Result::NonSupport;
+    if (loader->sharing > 0) --loader->sharing;
     return Result::Success;
 }
 


src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -301,7 +301,7 @@ void LottieFont::prepare()
 {
     if (!b64src) return;
 
-    Text::load(name, b64src, size, mime, false);
+    Text::load(name, b64src, size, mime, true);
 }
```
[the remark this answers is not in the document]
theashraf: resolved in https://github.com/thorvg/thorvg/pull/4406/commits/82c7821f5d9d4ad6e2a3e628b0c1a6d5559b0172

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -301,7 +301,7 @@ void LottieFont::prepare()
 {
     if (!b64src) return;
 
-    Text::load(name, b64src, size, mime, false);
+    Text::load(name, b64src, size, mime, true);
```
tinyjin: I think with the `copy`, loader will have replica, it'd be redundant during lifecycle without this.

```cpp
 if (Text::load(name, b64src, size, mime, true) == Result::Success) {
        tvg::free(b64src);
        b64src = nullptr;
}
```

tinyjin (commented):
Left comment in lottie front.

Overall, It seems to need refactoring along with renderer(tvgText) side together.
(Rather than, post decrement(--), we may increase refCount from where it is actually occupied like `++loader->sharing` in `Result font(const char* name)`)

src/loaders/lottie/tvgLottieModel.cpp
theashraf: Done in a5852a2 — with copy=true the loader keeps its own copy, so b64src is redundant after a successful load. Confirmed prepare() is its only reader, so freeing it there is safe.

theashraf:
On the sharing refcount: the `--loader->sharing` in the in-memory `Text::load(name, data, ...)` mirrors what the file-path overload already does (`if (loader->sharing > 0) --loader->sharing;` with the "font loading doesn't mean sharing" note), so it makes the two overloads consistent rather than adding a new pattern.

I agree the cleaner long-term fix is to `++sharing` only where the loader is actually occupied (e.g. `TextImpl`) instead of inside the `font()`/`anyfont()` lookups, which would let both overloads drop the compensating decrement. That also touches the file-path overload and the unload path, so it's a broader refactor than this use-after-free fix.

I'd suggest keeping this PR scoped to the UAF fix + the b64src cleanup, and handling the refcount refactor as a follow-up. Happy to take a pass at moving the increment to the occupancy sites if you'd prefer it here.

hermet:
@theashraf `LottieFont` may need to call `Result Text::unload(const char* filename)` when it is destroyed, rather than requiring other changes. Could you please double-check this? Thanks.

image

hermet:
see also: https://github.com/thorvg/thorvg/pull/4413

hermet:
@theashraf @tinyjin please reopen this if the issue is still validate on the latest thorvg. thanks.

Issue #4412, opened 2026-05-26
Bounding box behavior


JSUYA:
It looks that the calculation of the Bezier curve bounds is inaccurate.

<img width="810" height="679" alt="Image" src="https://github.com/user-attachments/assets/33f9c448-1149-4097-adb2-5c6ae0527b9b" />

```cpp
#include "Example.h"

struct UserExample : tvgexam::Example
{
    static void quadTo(tvg::Shape* shape, float curX, float curY, float ctrlX, float ctrlY, float endX, float endY)
    {
        float c1x = (curX + 2 * ctrlX) / 3.0f;
        float c1y = (curY + 2 * ctrlY) / 3.0f;
        float c2x = (endX + 2 * ctrlX) / 3.0f;
        float c2y = (endY + 2 * ctrlY) / 3.0f;
        shape->cubicTo(c1x, c1y, c2x, c2y, endX, endY);
    }

    //call bounds() BEFORE canvas->add(), same as SVG loader's _applyPatternProperty
    void drawBBox(tvg::Canvas* canvas, tvg::Paint* paint)
    {
        float x, y, w, h;
        paint->bounds(&x, &y, &w, &h);

        canvas->add(paint);

        auto bound = tvg::Shape::gen();
        bound->appendRect(x, y, w, h);
        bound->strokeWidth(2.0f);
        bound->strokeFill(255, 0, 0, 200);
        float dash[] = {6.0f, 4.0f};
        bound->strokeDash(dash, 2);
        canvas->add(bound);
    }

    bool content(tvg::Canvas* canvas, uint32_t w, uint32_t h) override
    {
        //Case 1: curve dips ABOVE start/end y (y-min extremum missed)
        //Q: start=(100,250) ctrl=(250,50) end=(400,250)
        //curve peak at y~183, but bounds() reports y=250
        {
            auto shape = tvg::Shape::gen();
            shape->moveTo(100, 250);
            quadTo(shape, 100, 250, 250, 50, 400, 250);
            shape->strokeWidth(4.0f);
            shape->strokeFill(30, 100, 255);
            drawBBox(canvas, shape);
        }

        //Case 2: closed shape from pattern.svg (the actual failing case)
        //M 320 220 Q 460 180 580 260 Q 540 380 380 380 Q 300 320 320 220 Z
        //shifted to (50, 300) for layout
        {
            float ox = -270, oy = 80;
            auto shape = tvg::Shape::gen();
[...]

PR #4417, opened 2026-05-28
common: fix Bezier bounds for degenerate cubics

JSUYA:
When a cubic is elevated from a quadratic (e.g. SVG Q command), the leading coefficient becomes zero, causing division by zero in the quadratic solver. The extremum is silently missed and the bounding box collapses to the endpoints.

Handle the zero case by solving the linear equation directly.

https://github.com/thorvg/thorvg/issues/4412

### DIFF
--- a/src/common/tvgMath.cpp
+++ b/src/common/tvgMath.cpp
@@ -440,6 +440,18 @@ void Bezier::bounds(BBox& box, const Point& start, const Point& ctrl1, const Poi
         auto a = -1.0f * start + 3.0f * ctrl1 - 3.0f * ctrl2 + end;
         auto b = start - 2.0f * ctrl1 + ctrl2;
         auto c = -1.0f * start + ctrl1;
+        if (tvg::zero(a)) {
+            if (tvg::zero(b)) return;
+            auto t = -c / (2.0f * b);
+            if (t > 0.0f && t < 1.0f) {
+                auto s = 1.0f - t;
+                auto q = s * s * s * start + 3.0f * s * s * t * ctrl1 + 3.0f * s * t * t * ctrl2 + t * t * t * end;
+                if (q < min) min = q;
+                if (q > max) max = q;
+            }
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


JSUYA:
I will not refactoring Bezier::bounds() in this PR.

hermet:
@JSUYA jfyi, the bounding box logic exists only for legacy purposes, and it does not provide accurate information when a stroke is involved in the shape.

JSUYA:
> @JSUYA jfyi, the bounding box logic exists only for legacy purposes, and it does not provide accurate information when a stroke is involved in the shape.

Yes, I understand that when a stroke is included, bounds are calculated through separate logic, and that this Bezier bounds() logic operates only on path information.
The problem with this function is that it attempts a zero division on the values ​​that make up the expression, which leads to inaccurate results.
[(Line 446 in src/common/tvgMath.cpp)](https://github.com/thorvg/thorvg/pull/4417/changes#diff-1249028d0c63729eb5cf84a257e798c94b3cd2d636a62c41888367585d1ee7c5L446)
In my opinion, this is a separate problem from the fact that bounds() does not guarantee accurate results.

I can find cases similar to this.
https://github.com/google/skia/blob/33a1c4f3e06782f4c14330d22d05dbb931f5a454/src/core/SkGeometry.cpp#L95-L100

src/common/tvgMath.cpp
```diff
@@ -423,6 +423,22 @@ float Bezier::angle(float t) const
 }
 
 
+static bool _linearExtrema(float a, float b, float c, float start, float ctrl1, float ctrl2, float end, float& min, float& max)
```
hermet: @JSUYA could you please inline this rather than separate func? thanks.

JSUYA: I moved the logic inside the bounds() function.

JSUYA:
rebase branch

hermet:
unitified here: https://github.com/thorvg/thorvg/pull/4410

Issue #4407, opened 2026-05-21
engines: inconsistency of Gaussian Blur effect


tinyjin:
GaussianBlur effect produces visibly different output across SW / GL / WG renderers, and lottie-web.

test file: [Main Scene (8).json](https://github.com/user-attachments/files/28089119/Main.Scene.8.json)

| WebGPU | WebGL | Software | Expected(lottie-web) |
|---|---|---|---|
| <img width="100%" src="https://github.com/user-attachments/assets/d33d1a0c-87cd-4637-9e01-5c627bd7a0a9" /> | <img width="100%" src="https://github.com/user-attachments/assets/29cbbdf6-3b9f-415d-95f6-1bfdaa3a9389" /> | <img width="100%" src="https://github.com/user-attachments/assets/160f2ac5-6c8d-4a47-993a-5f28c8821c24" /> | <img width="100%" src="https://github.com/user-attachments/assets/d425c6be-6eb9-4ae1-ba26-2fadaa5b5b97" /> |

wenjieshen:
### Current progress
---
There are actually 4 issues with the GL engine revealed by this issue.

1. The invisible gradient. (Resolved in https://github.com/thorvg/thorvg/pull/4418/changes)
2. Inconsistent Gaussian Blur Algorithm (GL and WG are the same)
<img width="1914" height="971" alt="Image" src="https://github.com/user-attachments/assets/7891068f-ac0f-4bad-83eb-796df820b033" />

3. The artifact on the edge of the viewport  (Resolved in https://github.com/thorvg/thorvg/pull/4425)
<img width="370" height="323" alt="Image" src="https://github.com/user-attachments/assets/3ce51a5f-2e17-46dd-b321-b2f6c92067fb" />

4. Inconsistent accumulated color burn and color dodge blending. ThorVG already matches the W3C compositing formula for this sample. The mismatch is against Lottie-web’s SVG renderer as composited by Chrome, not against the W3C blend rules. Since Lottie-web delegates blend behavior to browser CSS/SVG mix-blend-mode, its output is not a renderer-independent reference. Therefore, no further ThorVG GL blend modification is justified unless ThorVG explicitly decides to emulate lottie-web/Chrome SVG behavior as a compatibility target.

<img width="1395" height="457" alt="Image" src="https://github.com/user-attachments/assets/c26a6151-4efd-47a9-b491-3ef948e5af25" />

[...]

PR #4425, opened 2026-05-31
gl_engine: clamp Gaussian blur sampling to valid region

wenjieshen:
## Summary

Fix GL Gaussian blur sampling at region edges when using pooled intermediate render targets.

The blur source texture can be larger than the valid rendered region because pooled FBOs are power-of-two aligned. This updates Gaussian blur and drop shadow blur to pass texture-local viewport bounds, then clamps the shader kernel range so samples outside the valid region are skipped and the remaining weights are renormalized.
image
image

## Details

- Use `{0, 0, vp.sw(), vp.sh()}` as the blur viewport bounds for pooled intermediate textures.
- Clamp vertical and horizontal Gaussian kernel ranges against `gl_FragCoord`.
- Fall back to the current texel if no valid blur samples are available.
- No render-task or FBO flow changes.

## Testing

- Rendering verification is left for manual visual testing. No observable regression during all.sh execution.
- Maintain FPS consistency.

Related issue: https://github.com/thorvg/thorvg/issues/4407

### DIFF
--- a/src/renderer/gpu_engine/gl/tvgGlShaderSrc.cpp
+++ b/src/renderer/gpu_engine/gl/tvgGlShaderSrc.cpp
@@ -1113,17 +1113,19 @@ void main()
     float sigma = uGaussian.sigma * uGaussian.scale;
     float weightSum = 0.0;
     int radius = int(uGaussian.extend);
-    
-    for (int y = -radius; y <= radius; ++y) {
-        vec2 offset = vec2(0.0, float(y) * texelSize.y);
-        vec2 coord = vUV + offset;
-        float pixCoord = uViewport.vp.y - coord.y / texelSize.y;
-        float weight = pixCoord < uViewport.vp.w ? gaussian(float(y), sigma) : 0.0;
+    vec2 texelStep = vec2(0.0, texelSize.y);
+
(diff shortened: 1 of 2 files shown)


hermet:
@wenjieshen could you please double-check if this is related? https://github.com/thorvg/thorvg/issues/4132

src/renderer/gpu_engine/gl/tvgGlShaderSrc.cpp
hermet, line 1104: @shenwenjie As a separate note, gaussian looks can be computed in CPU level in advance rather than computing every pixel. 

At very least, computing the following values on every pixel seems inefficient. It may be worth exploring possible optimizations.

`(2.0 * sigma * sigma)`
`(sqrt(2.0 * 3.141592) * sigma)`

hermet (approved):
applied neat shaper code.

src/renderer/gpu_engine/gl/tvgGlShaderSrc.cpp
wenjieshen: I will conduct a new PR for it. Thank you.

wenjieshen:
> @wenjieshen could you please double-check if this is related? #4132

Thank you. I have conducted the test environment and confirmed this PR resolves #4132. :pray:

PR #4427, opened 2026-06-01
cpu_engine: replace the blending algorithm from ADOBE to W3C spec

hermet:
- softlight, hue, saturate, color, luminosity
- moved the hsl2rgb function to svg loader which only uses
- migrated the blender oeprations to the designated source file
- added extra refactoring

issue: https://github.com/thorvg/thorvg/issues/4190

Co-Authored-By: Hermet Park <hermetpark@gmail.com>


src/loaders/svg/tvgSvgUtil.cpp
```diff
@@ -103,3 +104,41 @@ size_t svgUtilURLDecode(const char *src, char** dst)
     return idx + 1;
 }
 
+void svgHSL2RGB(float h, float s, float l, uint8_t& r, uint8_t& g, uint8_t& b)
```
JSUYA: Since you're refactoring anyway, how about using 'To' instead of '2'? Like _hexCharToDec().

wenjieshen (approved):
Thank you for the contribution. There is no observable during the execution of all examples. The comparison also shows that the blending has been switched to the W3C spec correctly.
Image

Issue #4232, opened 2026-03-20
infra: pixel perfection test tool with CI


hermet:
Since ThorVG has released v1.0, we can introduce the pixel-perfect tests integrated into CI to improve the robustness of ThorVG’s rendering & test automation.

PR #4431, opened 2026-06-04
infra: add pixel perfection CI

Nor-s:
Add a Pixel Test CI that renders against a reference commit using thorvg.pixel-inspector and compares it with the PR merge result.

- pixel_test.yml: build reference & PR, run pixel comparison per backend, export per-backend PDF reports, upload report/summary artifacts
- pixel_test_report.yml: post or update a PR comment with the per-backend compared/diff/failed table and a link to the PDF report

issue: #4232

## Screenshot

- https://github.com/Nor-s/thorvg/pull/1#issuecomment-4621323701

The screenshot below shows an example of applying the recent commit 8af2027491253b14e9a303aa1d464a13ce98ea50 — (a regression fix → confirm that 2 alpha-value changes were corrected).

| comment | report |
|-|-|
|image|image|

The reporting mechanism is similar to binary_size. It updates the (existing) comment.

A link is provided where the PDF report can be downloaded.

### DIFF
--- a/.github/workflows/pixel_test.yml
+++ b/.github/workflows/pixel_test.yml
@@ -1,0 +1,172 @@
+name: Pixel Test
+
+on: pull_request
+
+permissions:
+  contents: read
+
+env:
+  THORVG_PIXEL_REFERENCE_REF: v1.0.6
+  WGPU_NATIVE_VERSION: v27.0.4.0
+
+jobs:
+  pixel_test:
+    name: reference vs PR merge
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    env:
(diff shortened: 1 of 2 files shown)


.github/workflows/pixel_test.yml
```diff
@@ -0,0 +1,167 @@
+name: Pixel Test
+
+on: pull_request
+
+permissions:
+  contents: read
+
+env:
+  THORVG_PIXEL_REFERENCE_REF: 8af2027491253b14e9a303aa1d464a13ce98ea50
```
Nor-s:
```
THORVG_PIXEL_REFERENCE_REF: 8af2027491253b14e9a303aa1d464a13ce98ea50
```

We can swap the reference revision by editing this value in pixel_test.yml.

hermet:
@Nor-s Could you please set up the base snapshot for v1.0.6? Let's begin from there.

Issue #4394, opened 2026-05-14
lottie: a error log in test


hermet:
Tested with lottie set: https://github.com/thorvg/thorvg.test-suite/tree/main/lottie/d

[E] LOTTIE (../src/loaders/lottie/tvgLottieBuilder.cpp 957): Expected image data.
[E] LOTTIE (../src/loaders/lottie/tvgLottieBuilder.cpp 957): Expected image data.

hermet:
Tested with lottie set:  https://github.com/thorvg/thorvg.test-suite/tree/main/lottie/c

run Lottie with GL or WGPU
`[E] LOTTIE (../src/loaders/lottie/tvgLottieExpressions.cpp 1512): Failed to dispatch the expressions!`

Nor-s:
> Tested with lottie set: https://github.com/thorvg/thorvg.test-suite/tree/main/lottie/c
> 
> run Lottie with GL or WGPU `[E] LOTTIE (../src/loaders/lottie/tvgLottieExpressions.cpp 1512): Failed to dispatch the expressions!`

This error occurred in 24588.json.
It was also reproduced in sw, and it also occurs at tag v1.0.0.
I'll look further into the cause of the error, which reproduces immediately even when run standalone.

hermet:
> > Tested with lottie set: https://github.com/thorvg/thorvg.test-suite/tree/main/lottie/c
> > run Lottie with GL or WGPU `[E] LOTTIE (../src/loaders/lottie/tvgLottieExpressions.cpp 1512): Failed to dispatch the expressions!`
> 
> This error occurred in 24588.json. It was also reproduced in sw, and it also occurs at tag v1.0.0. I'll look further into the cause of the error, which reproduces immediately even when run standalone.

\+ @tinyjin

tinyjin:
The log also shows in the SW case
It comes from two unsupported expression APIs and two edge case:
- **Layer.sourceRectAtTime()**
- **Layer.marker**
- **Layer.content(N)** // access by number
- **Basic shape properties accessor** // Rect and Ellipse // var rectW = content("rect").size[0]

PR #4436, opened 2026-06-06
lottie/expressions: support content() access by numeric index

tinyjin:
The index uses the item's authored `"ix"`,
or its order in the shape list when absent.

related: https://github.com/thorvg/thorvg/issues/4394

test file : 
[content_index.json](https://github.com/user-attachments/files/28661799/content_index.json)

| Before | After | Expectation |
|---------|---------|--------|
| CleanShot 2026-06-06 at 17 10 29@2x | CleanShot 2026-06-06 at 17 16 59 |  CleanShot 2026-06-06 at 17 10 49 |


hermet:
> The index uses the item's authored `"ix"`, or its order in the shape list when absent.

Mixing these two concepts seems very dangerous...

@tinyjin Could you please suggest a practical (issue)  resource for testing? (not sample generated by you)

hermet:
minimum fix: https://github.com/thorvg/thorvg/pull/4482

tinyjin:
> > The index uses the item's authored `"ix"`, or its order in the shape list when absent.
> 
> Mixing these two concepts seems very dangerous...
> 
> @tinyjin Could you please suggest a practical (issue) resource for testing? (not sample generated by you)

@hermet 
[24588.json](https://github.com/user-attachments/files/29239801/24588.json) asset has numeric access with `content()`. But visually no diff with this patch because that asset misses another expressions indistinguishable to the eye. The patch would fix compliance issue to avoid error log reported in related issue.

For another practical assets, I found two more (which use `content(N)`).
- [15541.json](https://github.com/user-attachments/files/29243870/15541.json)
- [1202.json](https://github.com/user-attachments/files/29243874/1202.json)

But still, they miss another spec, so the visual enhancement doesn't happen yet.
The error logs due to content numeric access are resolved.

hermet:
@tinyjin An implementation based solely on self-created samples is not acceptable in this expressions, as it could validates assumptions made by your implementer. Please make it sure that fix points from the the real-world 2 samples. At least these mixing concepts looks not reasonable.

hermet:
@tinyjin Please make it sure and reopen it. Thanks.

Issue #4393, opened 2026-05-14
gl: error log on Lottie Test (regression)


hermet:
`[E] Common (../src/common/tvgMath.cpp 495): Bezier segments count exceeded the maximum limit.`

tested with Lottie example: https://github.com/thorvg/thorvg.test-suite/tree/main/lottie/a

hermet:
Recently, expressions have become workable even on multiple threads, and now they are executable and can produce different rendering results. So this is not a newly introduced bug, but a hidden one that was only observed by chance.

curlprit: 
```
commit e266efc36ef7d770bdbcb2ad01969ac60fb4f23f (HEAD)
Author: NSG <soongeon@lottiefiles.com>
Date:   Wed May 6 11:13:30 2026 +0900

    lottie/expressions: support multi-threading for Lottie expressions

    - reduce TLS access in JERRY_CONTEXT
    - Introduce JERRY_DEFINE_CURRENT_CONTEXT macro to minimize TLS access from JERRY_CONTEXT
    - ref: https://github.com/jerryscript-project/jerryscript/pull/2470

    issue: #2256
```

wenjieshen:
Hi @Nor-s , 

I discovered that the issue arises from loading the specific layer called `kon` in the file `16729.json`. This layer generates a matrix with NaN values after the execution of the function `static bool _update(LottieTransform* transform, float frameNo, Matrix& matrix, uint8_t& opacity, bool autoOrient, Tween& tween, LottieExpressions* exps)`, which is called by `void LottieBuilder::updateTransform(LottieLayer* layer, float frameNo)`.

You can find the file here: [16729.json]
[16729.json](https://github.com/user-attachments/files/27786767/16729.json)

c.c. @hermet

hermet:
\+ @tinyjin

PR #4440, opened 2026-06-07
lottie/expressions: handle 1d wrapper objects as scalars correctly

tinyjin:
Scalar values exposed to JS are wrapped in an object to support the .value accessor.
In `jerry_value_t _number(float value)`

But `_addsub()`, `_muldiv()` classified only a primitive numbe as 1d, so a wrapped scalar fell into the 2d-point path and read a non-existent index as NaN.

issue: https://github.com/thorvg/thorvg/issues/4393

### Regression Overview
- since: `v1.0.2`
- culprit: https://github.com/thorvg/thorvg/commit/e458d6823
- test file:  [16729.json](https://github.com/user-attachments/files/28677248/16729.json)

| 1.0.1 | Current | Patched |
| CleanShot 2026-06-07 at 16 03 52 | CleanShot 2026-06-07 at 16 02 15 | CleanShot 2026-06-07 at 16 01 53 |

In the GL/WG Engine, they show `Bezier segments count exceeded the maximum limit` error in tessellation.
This log is also revolved by this patch.

CleanShot 2026-06-07 at 16 05 20@2x

### Regression Verification

Original commit adjusted numeric access for expressions using the .value accessor, and that behavior is still retained.

test file: [circle-blob-icon-lottie-json-animation (5).json](https://github.com/user-attachments/files/28677343/circle-blob-icon-lottie-json-animation.5.json)

| Expressions | Single |
|---------|---------|
| CleanShot 2026-06-07 at 16 15 13 | CleanShot 2026-06-07 at 16 14 07 |

### DIFF
--- a/src/loaders/lottie/tvgLottieExpressions.cpp
+++ b/src/loaders/lottie/tvgLottieExpressions.cpp
@@ -176,7 +176,29 @@ static jerry_value_t _color(RGB32 rgb)
     return value;
 }
 
+// return true if the number is 1d otherwise 2d, return false
+static bool _number(jerry_value_t obj, Point& out)
+{
+    if (jerry_value_is_number(obj)) {
+        out.x = jerry_value_as_number(obj);
+        return true;
+    }
 
+    // 1d or 2d object
+    auto v1 = jerry_object_get_index(obj, 0);
+    auto v2 = jerry_object_get_index(obj, 1);
+
+    out.x = jerry_value_as_number(v1);
+    jerry_value_free(v1);
+
+    if (jerry_value_is_undefined(v2)) return true;
+    out.y = jerry_value_as_number(v2);
+    jerry_value_free(v2);
+
+    return false;
+}
+
+// TODO: may need to replace with _number(jerry_value_t obj, Point& out)
 static float _number(jerry_value_t obj)
 {
     if (jerry_value_is_number(obj)) return jerry_value_as_number(obj);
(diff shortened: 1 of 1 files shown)


Nor-s (approved):
image


LGTM. Handling it as 1D seems reasonable to me.
Also, for the pixel-perfect test, no diffs were found except for the asset that already had the issue.
And I confirmed that the error log has been removed.

image


+ and fixed some lottie files
    - thorvg.test-suite/lottie/a/17068.json
    - thorvg.test-suite/lottie/expressions/insta_camera.json
    - thorvg.test-suite/lottie/expressions/worm.json

Thank you.

src/loaders/lottie/tvgLottieExpressions.cpp
```diff
@@ -513,6 +513,16 @@ static void _buildLayer(jerry_value_t context, float frameNo, LottieLayer* layer
 }
 
 
+static bool _is1d(const jerry_value_t v)
+{
+    if (!jerry_value_is_object(v)) return true;
+    auto v1 = jerry_object_get_index(v, 1);
```
hermet: Somehow, we can introduce a more optimal approach for argument 2, which is ultimately used in the 2D calculation logic. Simply checking it and dropping it here does not seem to be the best...

The sequence must look like this:
try string -> if not string, try 2d -> try 1d if not 2d.

Issue #4071, opened 2025-12-24
lottie: unexpected rendering behaviour


theashraf:
Lottie: 

[INTRO_LETTER_SOUP_DSKTP.json](https://github.com/user-attachments/files/24323242/INTRO_LETTER_SOUP_DSKTP.json)

https://github.com/user-attachments/assets/f7821ff1-df9f-4fae-80ad-3bb185c6265f

related https://github.com/LottieFiles/dotlottie-web/issues/651

Issue #3839, opened 2025-09-25
lottie: support multi-dimensional easings


rgprog:
ThorVG 1.0.0-pre28 does not support multi-dimensional easings (described in [the spec](https://lottie.github.io/lottie-spec/1.0.1/single-page/#specs-properties-easing-handle): "one element per dimension so you can have different easing curves per dimension"). When a keyframe for a vector property has non-scalar easing handles, ThorVG seems to ignore all but one component of each handle.

To reproduce the bug, load this Lottie file into an online viewer which uses a recent version of ThorVG, such as [this viewer](https://lottiefiles.github.io/dotlottie-web/) or [this one](https://www.thorvg.org/viewer).

````json
[...]

PR #4441, opened 2026-06-07
lottie: support per-dimension easing

tinyjin:
When keyframe easing (i/o) is split into per-dimension arrays, parse a separate bezier interpolator per dimension.

Previously, applied a single curve to all dimensions.

```js
// supported
 {
    "t": 0,
    "s": [100, 100],
    "i": { "x": 0.10, "y": 1.00 },
    "o": { "x": 0.90, "y": 0.00 }
  }

// not supported ([1] value ignored)
{
    "t": 0,
    "s": [100, 100],
    "i": { "x": [0.10, 0.90], "y": [1.00, 0.05] },
    "o": { "x": [0.90, 0.10], "y": [0.00, 0.95] }
},
```

issue: https://github.com/thorvg/thorvg/issues/4071, https://github.com/thorvg/thorvg/issues/3839

test file:  [INTRO_LETTER_SOUP_DSKTP (4).json](https://github.com/user-attachments/files/28678978/INTRO_LETTER_SOUP_DSKTP.4.json)

| Before | After |
|---------|---------|
| CleanShot 2026-06-07 at 18 47 52 | CleanShot 2026-06-07 at 18 47 16 |

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -349,12 +349,26 @@ bool LottieParser::getValue(RGB32& color)
 }
 
 
-void LottieParser::getInterpolatorPoint(Point& pt)
+void LottieParser::getInterpolatorPoint(Point pt[DimCnt])
 {
+    auto getField = [&](float& x, float& y) {
+        if (peekType() != kArrayType) {
+            x = y = getFloat();
+            return;
+        }
+        enterArray();
+        if (!nextArrayValue()) return;
+        x = getFloat();
+        if (nextArrayValue()) {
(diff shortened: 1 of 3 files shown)


src/common/tvgMath.h
```diff
@@ -508,6 +508,11 @@ static inline T lerp(const T &start, const T &end, float t)
     return static_cast<T>(start + (end - start) * t);
 }
 
+static inline Point lerp(const Point& start, const Point& end, float tx, float ty)
```
hermet: the above lerp doesn't work?

tinyjin: Updated without adding this. (removed unnecessary lerp)

PR #4448, opened 2026-06-09
loaders/webp: fix heap OOB write in VP8L Huffman table build (CVE-2023-4863)

dkgkdfg65:
## Summary

The vendored WebP decoder in `src/loaders/webp/` is a pre-1.3.2 copy of libwebp and is affected by **CVE-2023-4863** (the in-the-wild "libwebp" heap overflow). When ThorVG is built with the `webp` loader and uses this bundled copy (static builds always use it; shared builds fall back to it when system libwebp is absent), a crafted lossless (VP8L) WebP triggers a heap-buffer-overflow **write**.

This can be reached through the public API (`Picture::load("x.webp")` / `Picture::load(data, size, "webp")`) and, more interestingly, through untrusted SVG `<image>` data URIs: `tvgSvgBuilder.cpp` registers `webp` as a base64 image mime and calls `picture->load(decoded, size, "webp")`.

## Root cause

`ReadHuffmanCodes()` in `dec/vp8l.cpp` allocates a single fixed-size buffer for all Huffman tables:

```cpp
huffman_tables = tvg::malloc<HuffmanCode>(num_htree_groups * table_size * sizeof(*huffman_tables));
```

where `table_size = kTableSize[color_cache_bits]`. `VP8LBuildHuffmanTable()` (`utils/huffman.cpp`) then fills second-level tables into that buffer, advanced only by `next += size`, **without ever checking that the actually built table size stays within the allocation**. A malicious bitstream makes the real table grow past the bound, so `ReplicateValue()` writes out of bounds.

ASan stack (vendored sources, before fix):

```
WRITE of size 4 ... 0 bytes to the right of an 11816-byte region
  #0 ReplicateValue           src/loaders/webp/utils/huffman.cpp:57
  #1 VP8LBuildHuffmanTable     src/loaders/webp/utils/huffman.cpp:189
  #2 ReadHuffmanCode           src/loaders/webp/dec/vp8l.cpp:297
  #3 ReadHuffmanCodes          src/loaders/webp/dec/vp8l.cpp:376
  #4 DecodeImageStream         src/loaders/webp/dec/vp8l.cpp:1302
  #5 VP8LDecodeHeader          src/loaders/webp/dec/vp8l.cpp:1509
  #6 WebPDecodeRGBA            src/loaders/webp/dec/webp.cpp:644
allocated by:
  tvg::malloc<HuffmanCode> ... ReadHuffmanCodes  src/loaders/webp/dec/vp8l.cpp:355
```

## Fix

This ports the **official upstream libwebp fix, commit `902bc91`** (released in libwebp 1.3.2), to the vendored copy. The single fixed-width `HuffmanCode*` table is replaced by the segmented `HuffmanTables` API:

- `utils/huffman.h` — add `HuffmanTablesSegment` / `HuffmanTables`, declare `VP8LHuffmanTablesAllocate` / `VP8LHuffmanTablesDeallocate`, and change `VP8LBuildHuffmanTable()` to take `HuffmanTables*`.
- `utils/huffman.cpp` — split `BuildHuffmanTable()` (which now supports a `NULL` dry-run that only computes the required size) from the `VP8LBuildHuffmanTable()` wrapper. The wrapper first sizes the table, then allocates a fresh segment whenever the current one cannot hold it, and only then fills it — so the buffer can never be overrun. Adds allocate/deallocate.
- `dec/vp8li.h` — `huffman_tables_` becomes an in-place `HuffmanTables`.
- `dec/vp8l.cpp` — allocate via `VP8LHuffmanTablesAllocate()`, advance `curr_segment->curr_table`, and free via `VP8LHuffmanTablesDeallocate()`.

The algorithm matches upstream 1.3.2; the only adaptation is using ThorVG's `tvg::malloc` / `tvg::free` wrappers in place of `WebPSafeMalloc` / `WebPSafeFree`. To keep the change minimal and reviewable, only the CVE-2023-4863 fix (`902bc91`) is ported — the unrelated packed-table / htree-index-mapping changes that also landed in 1.3.2 are intentionally **not** included.

## Verification (ASan)

Built the vendored `dec/*.cpp dsp/*.cpp utils/*.cpp` with `clang++ -fsanitize=address`, decoding via the same `WebPDecodeRGBA` entry that `WebpLoader::run()` uses.

| input | before fix | after fix |
|---|---|---|
| public CVE-2023-4863 reproducer (`bad.webp`) | heap-buffer-overflow WRITE (ASan abort) | returns `NULL`, no crash, exit 0 |
| in-the-wild-style reproducer (`nso.webp`) | heap-buffer-overflow WRITE (ASan abort) | returns `NULL`, no crash, exit 0 |
| valid **lossless** WebP (VP8L) | decodes 2048x396 | decodes 2048x396, **pixel-identical** to upstream libwebp 1.3.2 (same SHA-256) |
| valid **lossy** WebP (VP8) | decodes 128x128 | decodes 128x128 |

So the malformed inputs are now rejected cleanly while normal decoding (including the lossless Huffman path that was changed) is unaffected.

## Disclosure note

CVE-2023-4863 is a long-public 2023 CVE, so this is submitted directly as a fix PR. (Private vulnerability reporting is enabled on the repo, but its REST report endpoint currently returns HTTP 500 for external reporters, so a coordinated private report was not possible.)

### DIFF
--- a/src/loaders/webp/dec/vp8l.cpp
+++ b/src/loaders/webp/dec/vp8l.cpp
@@ -202,11 +202,11 @@ static int ReadHuffmanCodeLengths(
   int symbol;
   int max_symbol;
   int prev_code_len = DEFAULT_CODE_LENGTH;
-  HuffmanCode table[1 << LENGTHS_TABLE_BITS];
+  HuffmanTables tables;
 
-  if (!VP8LBuildHuffmanTable(table, LENGTHS_TABLE_BITS,
-                             code_length_code_lengths,
-                             NUM_CODE_LENGTH_CODES)) {
+  if (!VP8LHuffmanTablesAllocate(1 << LENGTHS_TABLE_BITS, &tables) ||
+      !VP8LBuildHuffmanTable(&tables, LENGTHS_TABLE_BITS,
+                             code_length_code_lengths, NUM_CODE_LENGTH_CODES)) {
     goto End;
   }
 
@@ -226,7 +226,7 @@ static int ReadHuffmanCodeLengths(
     int code_len;
     if (max_symbol-- == 0) break;
     VP8LFillBitWindow(br);
-    p = &table[VP8LPrefetchBits(br) & LENGTHS_TABLE_MASK];
+    p = &tables.curr_segment->start[VP8LPrefetchBits(br) & LENGTHS_TABLE_MASK];
     VP8LSetBitPos(br, br->bit_pos_ + p->bits);
     code_len = p->value;
     if (code_len < kCodeLengthLiterals) {
@@ -249,6 +249,7 @@ static int ReadHuffmanCodeLengths(
   ok = 1;
 
  End:
+  VP8LHuffmanTablesDeallocate(&tables);
   if (!ok) dec->status_ = VP8_STATUS_BITSTREAM_ERROR;
   return ok;
 }
@@ -256,7 +257,8 @@ static int ReadHuffmanCodeLengths(
 // 'code_lengths' is pre-allocated temporary buffer, used for creating Huffman
 // tree.
 static int ReadHuffmanCode(int alphabet_size, VP8LDecoder* const dec,
-                           int* const code_lengths, HuffmanCode* const table) {
+                           int* const code_lengths,
+                           HuffmanTables* const table) {
   int ok = 0;
   int size = 0;
   VP8LBitReader* const br = &dec->br_;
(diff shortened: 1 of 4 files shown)


Nor-s:
image

With this patch, the PictureWebp example works correctly.

The changes in this issue are mostly identical to: [Fix OOB write in BuildHuffmanTable.](https://github.com/webmproject/libwebp/commit/2af26267cdfcb63a88e5c74a85927a12d6ca1d76)


However, there are some parts where our code does not match the code in that version.
The slight differences appear to be due to these commits. (vp8l_dec.cpp#L226-L249)
- https://github.com/webmproject/libwebp/commit/067031eaed11e91d9914e1e872738c7bdf075e0b


Considering maintainability, I think it would be better to update libwebp itself.


Updating to v1.6.0 would bring the following benefits:

libwebp v1.6.0 (06/30/2025)
- Further security-related hardening and increased fuzzing coverage
- Crash/leak fixes on error/OOM and other bug fixes
- Lossless encode/decode performance improvements
- Alpha decode fix

dkgkdfg65:
On the CodeFactor check: the two remaining flags are `Complex Method` notices on `BuildHuffmanTable` and `ReadHuffmanCodeLengths`, which are the verbatim upstream libwebp functions -- the fix only swapped the table allocation/free, not their control flow. I left them unchanged so the port stays a 1:1 match with upstream 902bc91 rather than refactoring the security code just to satisfy the linter. Happy to mark them third-party/won't-fix or adjust if you'd prefer. (I also reverted a stray `.codefactor.yml` I'd pushed -- it can't take effect from a fork branch anyway, so the PR is back to just the fix you tested.)

Issue #4442, opened 2026-06-08
gl: pixel test result mismatches(v1.0.5 - main)


Nor-s:
<img width="1977" height="352" alt="Image" src="https://github.com/user-attachments/assets/9c419954-4738-4123-9694-fcc7d6676179" />

- [/thorvg.test-suite/blob/main/lottie/d/iaf_okay.json](https://github.com/thorvg/thorvg.test-suite/blob/main/lottie/d/iaf_okay.json)

rendering issue on the GL side with a specific asset.

It works fine on v1.0.5, but the problem occurs on main and v1.0.6.

PR #4453, opened 2026-06-11
gl_engine: batch stencil-cover geometry

wenjieshen:
## Summary

Batch compatible GL stencil-cover work so disjoint complex shapes can share render tasks instead of emitting one stencil/cover task per shape.

## Changes

- Add `GlStencilCoverBatch` to merge compatible stencil tasks and cover rectangles within the same render pass.
- Extend `GlStencilCoverTask` to run multiple stencil and cover subtasks under one stencil setup.
- Reuse shared rectangle index data and geometry-buffer helpers across solid and stencil-cover batching.
- Clamp primitive, cover, clip, image, and solid-batch viewports to the active paint bounds instead of the broader framebuffer/pass bounds.
- Flush pending stencil-cover batches at clip, pass transition, and final sync boundaries to preserve render ordering.

## Why

This reduces GL render-task overhead for disjoint complex shapes while keeping overlapping stencil regions isolated. The viewport changes also prevent fast-tracked clipped Lottie/precomp regions from affecting neighboring content when the pass viewport is larger than the active paint bounds.

## Testing

- 25% FPS increase on the Lottie example on the Mac mini M1, Mac Air M3.
- 3% FPS on the Lottie example on the Windows laptop with RTX 5080
- I will postpone addressing the differences identified by pixel-perfect inspection and combine it with issue #4442.
- ~100%in the situation that a large number of shapes are complex shapes, with as little overlapping as possible
## Validation
The 6 complex shapes only use 2 draw calls.
Screenshot 2026-06-14 045306


src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
```diff
@@ -188,44 +224,72 @@ RenderRegion GlRenderer::viewportRegion(const RenderRegion& vp, const RenderRegi
...
+    auto yGl = passViewport.sh() - y - h;
+    auto stencilViewport = RenderRegion{{x, yGl}, {x + w, yGl + h}};
+    coverTask->setViewport(stencilViewport);
+
+    if (color) addStencilCoverSolidLayout(coverTask, gpuBuffer, bbox, *color);
+    else addStencilCoverPositionLayout(coverTask, gpuBuffer, bbox);
+    coverTask->mUseDrawArrays = true;
+    coverTask->mArrayMode = GL_TRIANGLES;
+    coverTask->mArrayOffset = 0;
+    coverTask->mIndexCount = COVER_VERTEX_COUNT;
```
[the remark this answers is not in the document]
wenjieshen: We avoid adding getter/setter wrappers for this internal hot path.

src/renderer/gpu_engine/gl/tvgGlRenderTask.h
```diff
@@ -98,32 +100,36 @@ class GlRenderTask
     uint32_t getIndexOffset() const { return mIndexOffset; }
     uint32_t getIndexCount() const { return mIndexCount; }
 
-private:
     GlProgram* mProgram;
     RenderRegion mViewport = {};
```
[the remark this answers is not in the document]
wenjieshen: We avoid adding getter/setter wrappers for this internal hot path.

tinyjin (approved):
I've also seen FPS improvement in my laptop & web tests. The visual looks also fine with my test sets.

hermet:
@wenjieshen 1 diff has been picked by this change. Please confirm it. Thank.
image

wenjieshen:
> @wenjieshen 1 diff has been picked by this change. Please confirm it. Thank. image

Hi @hermet, I checked the pixel-inspector result. Most of the diff is on anti-aliased edges, not on solid filled areas.

The highest DIFF RATIO is 0.00977408, so the weighted visual difference is small. 
This looks like a normal GL rasterization / subpixel / rounding difference, not a clear rendering regression. I think this should not block the PR for now. We can still keep it visible in the report and tune the tolerance later if needed.

src/renderer/gpu_engine/gl/tvgGlGpuBuffer.cpp
```diff
@@ -25,6 +25,12 @@
 #include <math.h>
 #include <string.h>
 
+#ifdef __EMSCRIPTEN__
+static constexpr uint32_t WASM_STAGE_BUFFER_CACHE = 4 * 1024 * 1024;
```
hermet: It would be better to avoid this condition whenever possible. If we apply this as the default behavior, what problems could arise? Also, it would be helpful to know where this value exactly comes from. Please add appropriate comments explaining it  in the code. Thanks.

hermet (commented):
@wenjieshen Please check a comment. Thanks.

src/renderer/gpu_engine/gl/tvgGlGpuBuffer.cpp
wenjieshen: Thank you for the feedback.

I kept the current behavior and added comments.

`reset()` is safe here because it runs after upload; GPU buffers already own the draw data. The WebGL-only threshold is only for dropping large cached CPU staging buffers to reduce WASM peak memory. Native GL keeps clear() to reuse large buffers and avoid realloc churn.

I also renamed the constants to `RESET_THRESHOLD` to make clear they are only cache-reset thresholds, not rendering limits.

Thank you.

src/renderer/gpu_engine/gl/tvgGlGpuBuffer.cpp
```diff
@@ -25,6 +25,16 @@
 #include <math.h>
 #include <string.h>
 
+// Stage buffers are temporary CPU upload data. reset() is safe after upload:
+// the GPU buffers already own the draw data. WebGL drops large cached buffers;
+// native GL keeps them for reuse.
```
hermet: @wenjieshen just to be sure, are there any measurable advantages or concrete test/sample that clearly show meaningful beneifts? please confirm  a clear, data-driven benefit. Otherwise, please remove such if-def, Thanks.

src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
```diff
@@ -41,6 +41,41 @@ static mutex _rendererMtx;
 static constexpr float IDENTITY_VERTEX[] = {-1.f, 1.f, -1.f, -1.f, 1.f, 1.f, 1.f, -1.f};
 static constexpr uint32_t RECT_INDEX[] = {0, 1, 2, 2, 1, 3};
 static constexpr uint32_t RECT_INDEX_COUNT = sizeof(RECT_INDEX) / sizeof(RECT_INDEX[0]);
+static constexpr uint32_t COVER_VERTEX_COUNT = 6;
+
+struct StencilCoverVertex
```
hermet: Defining these structs and functions here looks like a bad design smell. The GL renderer was designed to serve as an integration layer between the core engine and the GL backend for main rendering pipeline and context, not as a place for rasterization-specific mechanisms such as stencil-related functionality.

src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
```diff
@@ -765,7 +860,7 @@ void GlRenderer::prepareCmpTask(GlRenderTask* task, const RenderRegion& vp, uint
 void GlRenderer::endRenderPass(RenderCompositor* cmp)
 {
     auto glCmp = static_cast<GlCompositor*>(cmp);
-    
+
```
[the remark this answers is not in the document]
wenjieshen: False alarm. We will initialize the batch state when the pass starts.

src/renderer/gpu_engine/gl/tvgGlGpuBuffer.cpp
wenjieshen: Thank you. I have aligned them

src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
wenjieshen: Thank you. I have moved them to the stencil batch file.

src/renderer/gpu_engine/gl/tvgGlRenderTask.h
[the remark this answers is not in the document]
wenjieshen: I added it back

src/renderer/gpu_engine/gl/tvgGlGpuBuffer.cpp
```diff
@@ -162,34 +167,33 @@ uint32_t GlStageBuffer::reserveIndex(uint32_t size, void** dst)
...
     }
 
-    mGpuIndexBuffer.bind(GlGpuBuffer::Target::ELEMENT_ARRAY_BUFFER);
-    mGpuIndexBuffer.updateBufferData(GlGpuBuffer::Target::ELEMENT_ARRAY_BUFFER, mIndexBuffer.count, mIndexBuffer.data);
-    mGpuIndexBuffer.unbind(GlGpuBuffer::Target::ELEMENT_ARRAY_BUFFER);
-
-    mStageBuffer.clear();
-    mAuxBuffer.clear();
-    mIndexBuffer.clear();
+    if (mStageBuffer.reserved > STAGE_BUFFER_RESET_THRESHOLD) mStageBuffer.reset();
```
hermet: @wenjieshen This looks like a fundamentally flawed implementation. What kind of resource would actually cause memory usage to hit the limit? Shouldn't memory be reused proactively so that usage remains bounded, rather than continually growing until it reaches the maximum available limit? If the system ends up consuming all available memory, it seems to indicate a very inefficient memory management strategy. For a rendering engine, that would be a fairly serious architectural concern.

hermet:
backup the cap-buffer-staging: https://github.com/thorvg/thorvg/tree/jay/gl/cap-staging-buffer

hermet:
Test Machine: Ryzen 7 5800X / RTX4070 (Windows)
- Lottie: **Margin of Error**
- ManyRects: **-6.2%**
- LottieBigSize: **Margin of Error**

Test Machine: Intel Core Ultra 5 / Intel Arc/ (Linux)
- Lottie: **+93%**
- ManyRects: --
- LottieBigSize: **-11%**

hermet (changes requested):
regression found:

https://github.com/user-attachments/assets/d05d54c0-924c-4a82-b257-37ea52639e14

Nor-s:
- **Lottie**: +8%
- Star Test Asset: 207.97%

MacBook M5 pro

hermet:
This result looks remarkably different. We need to clarify this:

image

Nor-s:
> This result looks remarkably different. We need to clarify this:
> 
> image

@hermet 

This is not an issue with this PR. I’ll check it.

Nor-s:
@hermet 

This resource changes color when loaded. 

https://github.com/user-attachments/assets/b7a62e47-6cac-4fbe-b2cb-d322d27b2f01

hermet:
> @hermet 
> 
> This resource changes color when loaded. 
> 
> https://github.com/user-attachments/assets/b7a62e47-6cac-4fbe-b2cb-d322d27b2f01
> 
> 

@Nor-s if this is not by this PR, please confirm what brought the diff.

src/renderer/gpu_engine/gl/tvgGlStencilCoverBatch.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+void GlStencilCoverBatch::append(GlRenderTask* stencil, GlRenderTask* cover, const RenderRegion& bounds, const RenderRegion& viewBounds, const GlGeometryBuffer* stencilBuffer, uint32_t* stencilIndices)
+{
+    if (!mergeCover(cover, viewBounds)) {
+        task->mCoverTasks.push(cover);
+        coverViewBounds = viewBounds;
+    }
+    if (!merge(stencil, viewBounds, stencilBuffer, stencilIndices)) task->mStencilTasks.push(stencil);
+    addBounds(bounds);
+}
```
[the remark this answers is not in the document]
wenjieshen: Updated.

wenjieshen:
> regression found:
> 
>  2026-06-19-12-37-41.mp4

Hi @hermet, I have updated. Could you please check it again? Thank you. 🙏

hermet (approved):

PR #4459, opened 2026-06-13
lottie/text: implement text follow path in url font 

tinyjin:
Extend Lottie Text Follow Path to URL font rendering, previously embedded-only.

Refactor the path sampling into `RenderPath::pointAt`, simplifying LottieTextFollowPath and removing duplicated logic.

test file :  [font-follow-path.json](https://github.com/user-attachments/files/28910155/font-follow-path.json)

| ThorVG(Now) | lottie-web | Skottie |
| CleanShot 2026-06-13 at 18 26 59 | CleanShot 2026-06-13 at 18 25 35 | CleanShot 2026-06-13 at 18 26 08 |

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -994,6 +994,14 @@ void LottieBuilder::updateURLFont(LottieLayer* layer, float frameNo, LottieText*
         if (free) tvg::free(src);
     }
 
+    //text on the path
+    if (text->follow && ((uint32_t)text->follow->maskIdx < layer->masks.count)) {
+        auto path = text->pooling();
+        path->reset();
+        layer->masks[text->follow->maskIdx]->pathset(frameNo, to<ShapeImpl>(path)->rs.path, nullptr, tween, exps);
+        to<TextImpl>(paint)->layout(to<ShapeImpl>(path)->rs.path, text->follow->firstMargin(frameNo, tween, exps));
+    }
+
     //text build
     auto len = strlen(doc.text);
     auto buf = tvg::malloc<char>(len + 1);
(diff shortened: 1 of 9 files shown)


src/renderer/tvgRender.cpp
```diff
@@ -647,6 +647,66 @@ static void _get(float& begin, float& end)
 }
 
 
+Point RenderPath::pointAt(float at, float& angle) const
```
hermet, line 650: This kind of Lottie-only feature will not be included in the default binary, or at least it must be in the header for compile-time inclusion.
Also, RenderPath already provides a general-purpose function: `Point point(float progress)`. Probably we can revise it to use.

src/loaders/sfnt/tvgSfntLoader.cpp
```diff
@@ -484,6 +484,50 @@ void SfntLoader::wrapEllipsis(FontMetrics& fm, const Point& box, const char* utf
     _align(fm.align, box, {cursor.x, fm.size.y}, line, out.pts.count, out);  //last line
 }
 
+void SfntLoader::wrapOnPath(FontMetrics& fm, const char* utf8, const char* end, const RenderPath& path, float offset, RenderPath& out)
```
hermet, line 487: This is well outside the general Font/Text specification and requires careful consideration.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -994,6 +994,14 @@ void LottieBuilder::updateURLFont(LottieLayer* layer, float frameNo, LottieText*
         if (free) tvg::free(src);
     }
 
+    //text on the path
+    if (text->follow && ((uint32_t)text->follow->maskIdx < layer->masks.count)) {
+        auto path = text->pooling();
+        path->reset();
+        layer->masks[text->follow->maskIdx]->pathset(frameNo, to<ShapeImpl>(path)->rs.path, nullptr, tween, exps);
+        to<TextImpl>(paint)->layout(to<ShapeImpl>(path)->rs.path, text->follow->firstMargin(frameNo, tween, exps));
```
hermet, line 1002: A proper design seems necessary.

hermet (changes requested):
The approach and solution are overly tailored to fit the feature. We should discuss a proper design for the functionality.

PR #4435, opened 2026-06-06
lottie/expressions: support sourceRectAtTime()

tinyjin:
Layer.sourceRectAtTime() returns the source bounding box {top, left, width, height}.

see: https://lottiefiles.github.io/lottie-docs/expressions/#layersourcerectattime

related: https://github.com/thorvg/thorvg/issues/4394

test file: 
[21. Source Rectangle.json](https://github.com/user-attachments/files/28659452/21.Source.Rectangle.json)

| Current | Patched| Expectation |
|---------|---------|---------|
| texidddddisdidd | • • • Thervo Example (Botware) | CleanShot 2026-06-06 at 13 35 55@2x |

PR #4467, opened 2026-06-15
cpu_engine: enhance thread-safety of bounds() API

tinyjin:
bounds() internally used a fixed slot (tid 0) of the shared pool. Calling it from multiple threads thus raced on that slot and crashed at runtime. (_working well in GPU engines_)

Isolate the pool used by bounds() per call to make it thread-safe and fix the crash.

### Reproduce

Currently, `bounds` isn't internally referenced, test in caller-side multi-threaded env to reproduce this:
```cpp
void query(int callerIdx)
{
    for (size_t i = callerIdx; i < shapes.size(); i += NUM_CALLERS) {
        float x, y, w, h;
        if (shapes[i]->bounds(&x, &y, &w, &h) == tvg::Result::Success) {
            aabb[i * 2 + 0] = {x, y};
            aabb[i * 2 + 1] = {w, h};
        } else {
            aabb[i * 2 + 1] = {0.0f, 0.0f};
        }
    }
}

// query `bounds()` concurrently (cpu -> crash, gpu -> no issue)
std::vector<std::thread> pool;
for (int t = 0; t < NUM_CALLERS; ++t) {
    pool.emplace_back(&UserExample::query, this, t);
}
for (auto& th : pool) th.join();
```

full source : [BoundingBoxThreaded.cpp](https://github.com/user-attachments/files/28947726/BoundingBoxThreaded.cpp)

### DIFF
--- a/src/renderer/cpu_engine/tvgSwCommon.h
+++ b/src/renderer/cpu_engine/tvgSwCommon.h
@@ -468,7 +468,7 @@ bool shapeGenFillColors(SwShape& shape, const Fill* fill, const Matrix& transfor
 bool shapeGenStrokeFillColors(SwShape& shape, const Fill* fill, const Matrix& transform, SwSurface* surface, uint8_t opacity, bool ctable);
 void shapeResetFill(SwShape& shape);
 void shapeResetStrokeFill(SwShape& shape);
-bool shapeStrokeBBox(SwShape& shape, const RenderShape* rshape, Point* pt4, const Matrix& m, SwMpool* mpool);
+bool shapeStrokeBBox(const RenderShape* rshape, Point* pt4, const Matrix& m);
 void shapeDelFill(SwShape& shape);
 
 void strokeReset(SwStroke* stroke, const RenderShape* shape, const Matrix& transform, SwMpool* mpool, unsigned tid);
--- a/src/renderer/cpu_engine/tvgSwRenderer.cpp
+++ b/src/renderer/cpu_engine/tvgSwRenderer.cpp
@@ -694,7 +694,7 @@ bool SwRenderer::bounds(RenderData data, Point* pt4, const Matrix& m)
     auto task = static_cast<SwShapeTask*>(data);
     task->done();
 
-    return shapeStrokeBBox(task->shape, task->rshape, pt4, m, task->renderer->mpool);
+    return shapeStrokeBBox(task->rshape, pt4, m);
 }
 
 
--- a/src/renderer/cpu_engine/tvgSwShape.cpp
+++ b/src/renderer/cpu_engine/tvgSwShape.cpp
@@ -554,14 +554,17 @@ void shapeDelFill(SwShape& shape)
 }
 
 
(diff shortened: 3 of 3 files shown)


hermet:
@tinyjin Please submit an issue with a practical lottie example. Thanks.

tinyjin:
@hermet Its not related to Lottie

hermet:
@tinyjin What led you to write that code? Basically, ThorVG APIs are not thread-safe. I couldn’t find the point you're trying to make in your example.

tinyjin:
@hermet 
### **1. What led you to write that code?**
-> I found calling `bounds()` from multi-threaded env causes crash
-> It was discovered when implementing #4435 The patch uses bounds() API to get source rect that Expressions requires.
-> But it's just how it went. I thought it can be treated as an independent issue (Even if it is not used in lottie as now, it can cause a crash in general use case)

### **2. ThorVG APIs are not thread-safe.**
-> Are we? Then I'd understand users cannot use ThorVG in own thread pools (it's not guaranteed, and may crash expectidly)
-> I was understanding we don't limit whether users use our task scheduler or use their own

Regardless of No.2, I don’t also see a strong reason to enforce this restriction, and since it works fine on other GPU engines, I think it would be reasonable to allow it through a patch. However, if you still believe that this API should not guarantee thread safety, please let me know. I won’t consider it an issue in that case. Otherwise, yes I can open a new issue so we can review it as you requested.

hermet:
@tinyjin 

> 
> ### **1. What led you to write that code?**
> -> I found calling `bounds()` from multi-threaded env causes crash -> It was discovered when implementing #4435 The patch uses bounds() API to get source rect that Expressions requires. -> But it's just how it went. I thought it can be treated as an independent issue (Even if it is not used in lottie as now, it can cause a crash in general use case)

 Your example shows API calls made on the application side. It's shame that it violates the rule you're advocating. 

> ### **2. ThorVG APIs are not thread-safe.**
> -> Are we? Then I'd understand users cannot use ThorVG in own thread pools (it's not guaranteed, and may crash expectidly) -> I was understanding we don't limit whether users use our task scheduler or use their own
> 
> Regardless of No.2, I don’t also see a strong reason to enforce this restriction, and since it works fine on other GPU engines, I think it would be reasonable to allow it through a patch. However, if you still believe that this API should not guarantee thread safety, please let me know. I won’t consider it an issue in that case. Otherwise, yes I can open a new issue so we can review it as you requested.

First, your fix does not actually fix the issue. A crash can still occur whenever multiple threads access resource pool 0 concurrently. You have only made it work for your specific use case by masking the issue, and this change could potentially introduce even more critical problems elsewhere.

Second, the GPU engine does not currently support multithreading properly, so the behavior you are seeing only works by coincidence. Once we introduce further parallelization, this issue may surface at any time.  Currently, none of the ThorVG APIs are guaranteed to be thread-safe. Only for threading-safe bounds() doesn't make sense. Thread safety is the caller's responsibility.

As I mentioned before, please open an issue if you are experiencing a problem. We can then review the root cause and discuss an appropriate solution. Thanks.

PR #4469, opened 2026-06-16
lottie/text: fix box word-wrap overhang in url font

tinyjin:
Previously, text box wrap of url font constrained to be word-wrapped, causing long word to be overflown.

In the box text, long word should also be wrapped.

| Before | After | Expectation |
| CleanShot 2026-06-16 at 16 55 28@2x | CleanShot 2026-06-16 at 16 55 11@2x | image |

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -1035,7 +1035,7 @@ void LottieBuilder::updateURLFont(LottieLayer* layer, float frameNo, LottieText*
     paint->text(buf);
     paint->layout(doc.bbox.size.x, doc.bbox.size.y);
     paint->translate(doc.bbox.pos.x, doc.bbox.pos.y);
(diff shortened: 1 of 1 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1035,7 +1035,7 @@ void LottieBuilder::updateURLFont(LottieLayer* layer, float frameNo, LottieText*
     paint->text(buf);
     paint->layout(doc.bbox.size.x, doc.bbox.size.y);
     paint->translate(doc.bbox.pos.x, doc.bbox.pos.y);
-    if (doc.bbox.size.x > 0.0f) paint->wrap(TextWrap::Word);
+    if (doc.bbox.size.x > 0.0f) paint->wrap(TextWrap::Smart);
```
hermet, line 1038: btw, I don't see any necessity of the box size check here...

hermet (approved):

Issue #4473, opened 2026-06-18
renderer: allow Picture::load() to reload data from memory


hermet:
## Summary

Currently, `Picture::load()` for raw memory data does not allow the picture content to be reloaded once it has been loaded.

It would be useful to support reloading data from memory so that applications can update picture content dynamically without recreating the `Picture` object.

## Motivation

Some applications generate or modify image data at runtime. In such cases, users may want to replace the underlying memory buffer and reload the content through the existing `Picture` instance.

The current behavior requires creating a new `Picture` object, which may be inconvenient for dynamic content workflows.

## Proposal

Allow `Picture::load()` to be called again on an existing `Picture` instance when the source is provided as raw memory data. Or alternative api could be considered.

Example:

```cpp
auto picture = tvg::Picture::gen();

// Result load(const uint32_t* data, uint32_t w, uint32_t h, ColorSpace cs, bool copy = false) noexcept;
picture->load(data1, w, h, cs);
/* render */

picture->load(data1, w, h, cs);
/* render updated content */
```

## Expected Benefits

* Supports dynamic content updates.
* Simplifies user workflows.
* Avoids unnecessary object recreation when only the source data changes.

PR #4479, opened 2026-06-20
renderer: support raw bitmap reloading in Picture::load()

hermet:
- Allow raw bitmap data to be reloaded unless the picture contains a vector asset.
- Mark bitmap reloads as image updates so renderer caches are refreshed.

issue: #4473

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1612,6 +1612,7 @@ struct TVG_API Picture : Paint
      *
      * @retval Result::InvalidArguments In case no data are provided or the @p size is zero or less.
      * @retval Result::NonSupport When trying to load a file with an unknown extension.
+     * @retval Result::InsufficientCondition If a vector asset has already been loaded into the picture.
      *
      * @warning It's the user responsibility to release the @p data memory.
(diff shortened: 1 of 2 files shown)


wenjieshen (commented):
Basic raw bitmap reload works in my test.

I am sorry that I didn't expect this scenario, so there is a GL shared-texture case: if two Pictures share the same raw buffer, `release()` only drops `refCnt`, then `retain()` reuses the old `texId` without re-uploading pixels. So the shared texture can stay stale.

I can patch it here if you prefer, or push a PR after this PR is merged. Thank you.

hermet:
> Basic raw bitmap reload works in my test.
> 
> I am sorry that I didn't expect this scenario, so there is a GL shared-texture case: if two Pictures share the same raw buffer, `release()` only drops `refCnt`, then `retain()` reuses the old `texId` without re-uploading pixels. So the shared texture can stay stale.
> 
> I can patch it here if you prefer, or push a PR after this PR is merged. Thank you.

@wenjieshen I will double-check it, Thanks for pointing out.

PR #4483, opened 2026-06-22
media: add experimental Apple video playback support

Nor-s:
> [!NOTE]
> This is a PR targeting the hermet/video branch.

image

Implement the AVFoundation media loader for Apple video playback.
Replace the stub Apple media loader with playback, seek, loop,
volume, and mute controls.

Run AVFoundation work on a ref-counted serial dispatch queue.
Poll decoded frames at the media frame rate and copy them into
an owned ring buffer for bitmap() consumption.

Handle preferred video transforms, natural playback finish, still
frames for paused seek, and timer cleanup while paused or closed.

Wire the Objective-C++ source and required Apple frameworks into Meson.

PR #4484, opened 2026-06-22
media: refresh media frames as image updates

Nor-s:
> [!IMPORTANT]
> This PR is intended to be applied on top of the `hermet/video` branch.

Keep media pictures on the update path even when no render flag is set, so bitmap() can fetch newly decoded frames during playback.

Mark fetched media frames with the existing Image update flag, and refresh GL textures when surface pixels change in-place.

### macOS Video Frame Flow

image

The diagram shows how video playback works on macOS: a serial dispatch
queue fills a 3-frame ring buffer, while `Picture::update()` fetches the
latest frame through `bitmap()` and refreshes renderer image data with
`RenderUpdateFlag::Image`.

### DIFF
--- a/src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
+++ b/src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
@@ -1262,6 +1262,7 @@ RenderData GlRenderer::prepare(RenderSurface* image, RenderData data, const Matr
     sdata->viewWd = static_cast<float>(surface.w);
     sdata->viewHt = static_cast<float>(surface.h);
 
+    auto refreshTexture = ((flags & RenderUpdateFlag::Image) != RenderUpdateFlag::None);
     auto sourceChanged = (sdata->texSource != image) || (sdata->texFilter != filter);
     if (sdata->texId == 0 || sourceChanged || cacheStale) {
         auto ownsTexture = sdata->texId && (sdata->texStamp == mTextures.stamp);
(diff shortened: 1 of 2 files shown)


src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
```diff
@@ -1271,7 +1272,7 @@ RenderData GlRenderer::prepare(RenderSurface* image, RenderData data, const Matr
         sdata->texFilter = filter;
         sdata->texStamp = mTextures.stamp;
         sdata->geometry = GlGeometry();
-    }
+    } else if (refreshTexture) TextureMgr::upload(sdata->texId, image, filter);
```
tinyjin, line 1275: I'd set this condition
```cpp
else if (flags & RenderUpdateFlag::Image) {
    TextureMgr::upload(sdata->texId, image, filter);
}
```

src/renderer/tvgPicture.h
```diff
@@ -56,8 +56,8 @@ struct PictureImpl : Picture
 
     bool skip(RenderUpdateFlag flag)
     {
-        if (flag == RenderUpdateFlag::None) return true;
-        return false;
+        if (flag != RenderUpdateFlag::None) return false;
+        return !loader || loader->type != FileType::Media;
```
tinyjin, line 60: There is just a minor concern that the semantics of render flags might be compromised because only media ignores render flags.

I'd lean toward setting render flags at the source rather than special-casing media in common skip logic to make cleaner for render-flag semantics.

The animation loader already follows this pattern, however that has an explicit frame API, while media uses play() with engine(based on platform impl) timing, so the update point is less obvious, which makes this patch understandable. 

On web, media decoding and frame push happen outside native code, so a webframe-update API was structurally necessary, and I could naturally place to handle render flag update. (I can remove that logic when this merged)

At the moment I respect this change and will keep exploring clearner options.

hermet:
Coud you please add this in your PR with a separate commit? Thanks.
https://github.com/thorvg/thorvg/pull/4483

PR #4494, opened 2026-06-23
gl_engine: fix missing texture data upload on bitmap updates

hermet:
When bitmap data is updated, the corresponding OpenGL texture must reload the pixel data to properly refresh the rendered content.

This fixes a missing texture upload that could cause stale bitmap content to be displayed.

### DIFF
--- a/src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
+++ b/src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
@@ -1253,9 +1253,9 @@ void GlRenderer::dispose(RenderData data)
 
 RenderData GlRenderer::prepare(RenderSurface* image, RenderData data, const Matrix& transform, const Array<RenderData>& clips, uint8_t opacity, FilterMethod filter, RenderUpdateFlag flags)
 {
-    //TODO: redefine GlImage.
     if (opacity == 0) return data;
 
+    //TODO: redefine GlImage?
     auto sdata = static_cast<GlShape*>(data);
     if (!sdata) sdata = new GlShape;
 
@@ -1263,12 +1263,10 @@ RenderData GlRenderer::prepare(RenderSurface* image, RenderData data, const Matr
     if (flags == RenderUpdateFlag::None && !cacheStale) return data;
 
     sdata->validFill = false;
-
     sdata->viewWd = static_cast<float>(surface.w);
     sdata->viewHt = static_cast<float>(surface.h);
 
-    auto sourceChanged = (sdata->texSource != image) || (sdata->texFilter != filter);
-    if (sdata->texId == 0 || sourceChanged || cacheStale) {
+    if (cacheStale || sdata->texId == 0 || sdata->texSource != image || sdata->texFilter != filter) {
         auto ownsTexture = sdata->texId && (sdata->texStamp == mTextures.stamp);
         if (ownsTexture) disposeTexture(mTextures.release(sdata->texSource, sdata->texFilter, sdata->texId));
         sdata->texId = mTextures.retain(image, filter);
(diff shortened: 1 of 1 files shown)


wenjieshen (commented):
Hi @Nor-s,

Thank you. 
It works well.
However, `TextureMgr::upload` uses `glTexImage2D` only. 
It could be improved as follows:
- Use `glTexSubImage2D()` when the existing texture size/format matches the new bitmap;
- Use `glTexImage2D()` only when the texture needs to be recreated/redefined

For frequent uploads, it would be best to make this mandatory.

### Test
https://github.com/user-attachments/assets/161d28f6-4136-4f41-8629-9b088e4d657c

Nor-s:
> Hi @Nor-s,
> 
> Thank you. It works well. However, `TextureMgr::upload` uses `glTexImage2D` only. It could be improved as follows:
> 
> * Use `glTexSubImage2D()` when the existing texture size/format matches the new bitmap;
> * Use `glTexImage2D()` only when the texture needs to be recreated/redefined
> 
> For frequent uploads, it would be best to make this mandatory.
> 
> ### Test
>  Screen.Recording.2026-06-23.at.10.14.07.PM.mov

@wenjieshen 
Hello, thank you for the review.

I updated the code to use `glTexSubImage2D()`.

 In the prepare() path, retain() is already performed when the texture is invalid, stale, or when properties such as the filter have changed. 
 
 Therefore, when the same surface is used and the Image render update flag is set, 
 we can refresh the existing texture by passing only the texture ID and surface to glTexSubImage2D().

hermet:
> I updated the code to use `glTexSubImage2D()`.

@Nor-s  @wenjieshen  I dont' see any benefits to use it here until we have a partial upload of the bitmap. For now, this just makes bigger change. If there is no clear benefit, let's not introduce unnecessary burden.

image

Nor-s:
image

@hermet  @wenjieshen
 I measured performance by playing dozens of videos, but I didn’t see any significant difference, so I reverted the change.

wenjieshen:
> > I updated the code to use `glTexSubImage2D()`.
> 
> @Nor-s @wenjieshen I dont' see any benefits to use it here until we have a partial upload of the bitmap. For now, this just makes bigger change. If there is no clear benefit, let's not introduce unnecessary burden.
> 
> image

Just a small clarification from my side: I may not have explained the intention clearly enough. I was not suggesting glTexSubImage2D only for partial bitmap uploads.

NVIDIA's GLES tips say that when updating an existing texture, glTexSubImage* should be used instead of redefining the entire contents with TexImage when possible. The same document also describes glTexImage* functions as graphics memory allocations.

So my suggestion was only to avoid using the redefine/allocation-related path for frequent existing-texture updates. Since there is no practical difference here, I understand that this may not be worth the extra change.

hermet:
> NVIDIA's GLES tips say that when updating an existing texture, glTexSubImage* should be used instead of redefining the entire contents with TexImage when possible. The same document also describes glTexImage* functions as graphics memory allocations.

@wenjieshen any visible link? If glTexSubImage2D() performs better, please submit a patch. Thanks!

wenjieshen:
> > NVIDIA's GLES tips say that when updating an existing texture, glTexSubImage* should be used instead of redefining the entire contents with TexImage when possible. The same document also describes glTexImage* functions as graphics memory allocations.
> 
> @wenjieshen any visible link? If glTexSubImage2D() performs better, please submit a patch. Thanks!

Hi @hermet, 

Here is the document.
https://docs.nvidia.com/jetson/archives/r34.1/DeveloperGuide/text/SD/GraphicsProgramming/OpenglEsProgrammingTips.html

I only mentioned it because it could resolve performance issues on older devices.
However, I collected the information again after I saw the benchmark from @Nor-s .
Many discussions said most modern drivers optimize glTextImage*, so this habit isn't that important now. 

Thank you.

PR #4503, opened 2026-06-25
media: add web video playback gluelayer

tinyjin:
Implemented tvgWebMediaLoader
> paired with https://github.com/thorvg/thorvg.web/pull/292

## ThorVG Web Media Loader Interface

Media playback is delegated to JavaScript.

```js
Module.mediaPlayer = (loader, bytes) => new MyPlayer(bytes);

class MyPlayer {
  sync()      // called once per render (state sync)
  play()
  pause()
  stop()
  seek(sec)
  loop(on)
  volume(v)   // 0 - 1
  mute(on)
  dispose()
}

sync() {
  if (!ready) return null;                                                    // not ready
  if (!newFrame) return { width, height, duration };    // sync metadata
  return { width, height, duration, data, time };          // frame delivered
}
```

### DIFF
--- a/src/loaders/media/web/tvgWebMediaLoader.cpp
+++ b/src/loaders/media/web/tvgWebMediaLoader.cpp
@@ -1,0 +1,142 @@
+/*
+ * Copyright (c) 2026 ThorVG project. All rights reserved.
+
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include <emscripten/bind.h>
+#include "tvgWebMediaLoader.h"
+
+using emscripten::val;
+using emscripten::typed_memory_view;
+
+/************************************************************************/
+/* Internal Class Implementation                                        */
+/************************************************************************/
+
+EMSCRIPTEN_BINDINGS(tvg_web_media)
+{
+    emscripten::register_type<WebMediaPlayer>("WebMediaPlayer");
+    emscripten::register_type<WebMediaState>("WebMediaState | null");
+}
+
+/************************************************************************/
+/* External Class Implementation                                        */
+/************************************************************************/
+
+WebMediaLoader::~WebMediaLoader()
+{
+    if (!js.isUndefined() && !js.isNull()) js.call<void>("dispose");
+    tvg::free(surface.buf32);
+}
+
+bool WebMediaLoader::open(const char* data, uint32_t size, const LoaderOps* ops, bool copy)
+{
+    auto generator = val::module_property("createMediaPlayer");
+    if (generator.isUndefined()) return false;
+
+    auto bytes = val::global("Uint8Array").new_(val(typed_memory_view(size, reinterpret_cast<const uint8_t*>(data))));
+    js = WebMediaPlayer(generator(reinterpret_cast<uintptr_t>(this), bytes));
+    return !js.isNull();
+}
+
+bool WebMediaLoader::sync()
+{
+    auto state = js.call<WebMediaState>("sync");
+    if (state.isNull()) return false;
+
+    totalTime = state["duration"].as<float>();
+
+    auto width = state["width"].as<uint32_t>();
+    auto height = state["height"].as<uint32_t>();
+    if (width != surface.w || height != surface.h) {
+        surface.setup(tvg::realloc<uint32_t>(surface.buf32, width * height * sizeof(uint32_t)), width, width, height, sizeof(uint32_t), ColorSpace::ABGR8888S);
+        w = static_cast<float>(width);
+        h = static_cast<float>(height);
+    }
+
+    auto data = state["data"];
+    if (!surface.buf32 || data.isUndefined()) return false;
+
+    val(typed_memory_view(surface.stride * surface.h * surface.channelSize, reinterpret_cast<uint8_t*>(surface.buf32))).call<void>("set", data);
+    surface.premultiplied = false;
+    curTime = state["time"].as<float>();
+    return true;
+}
+
+RenderSurface* WebMediaLoader::bitmap()
+{
+    sync();
+    return BitmapLoader::bitmap();
+}
+
+Result WebMediaLoader::play()
+{
+    paused = false;
+    js.call<void>("play");
+    return Result::Success;
+}
+
+Result WebMediaLoader::pause()
+{
+    paused = true;
+    js.call<void>("pause");
+    return Result::Success;
+}
+
+Result WebMediaLoader::stop()
+{
+    paused = true;
+    curTime = 0.0f;
+    js.call<void>("stop");
+    return Result::Success;
+}
+
+Result WebMediaLoader::seek(float seconds)
+{
+    curTime = seconds;
+    js.call<void>("seek", seconds);
+    return Result::Success;
+}
+
+Result WebMediaLoader::loop(bool on)
+{
+    looping = on;
+    js.call<void>("loop", on);
+    return Result::Success;
+}
+
+Result WebMediaLoader::volume(float volume)
+{
+    audioVolume = volume;
+    js.call<void>("volume", volume);
+    return Result::Success;
+}
+
+Result WebMediaLoader::mute(bool on)
+{
+    muted = on;
+    js.call<void>("mute", on);
+    return Result::Success;
+}
+
+MediaLoader* MediaLoader::gen()
+{
+    return new WebMediaLoader;
+}
--- a/src/loaders/media/web/tvgWebMediaLoader.h
+++ b/src/loaders/media/web/tvgWebMediaLoader.h
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2026 ThorVG project. All rights reserved.
+
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef _TVG_WEB_LOADER_H_
+#define _TVG_WEB_LOADER_H_
+
+#include <emscripten/val.h>
+#include "tvgMediaLoader.h"
+
+EMSCRIPTEN_DECLARE_VAL_TYPE(WebMediaPlayer)
+EMSCRIPTEN_DECLARE_VAL_TYPE(WebMediaState)
+
+struct WebMediaLoader : MediaLoader
+{
+    WebMediaPlayer js = WebMediaPlayer(emscripten::val::undefined());
+
+    ~WebMediaLoader() override;
+
+    // Loader interface
+    bool open(const char* data, uint32_t size, const LoaderOps* ops, bool copy) override;
+    RenderSurface* bitmap() override;
+    bool sync() override;
+
+    // Playback controls
+    Result play() override;
+    Result pause() override;
+    Result stop() override;
+    Result seek(float seconds) override;
+    Result loop(bool on) override;
+    Result volume(float volume) override;
+    Result mute(bool on) override;
+};
+
+#endif  //_TVG_WEB_LOADER_H_
--- a/src/loaders/media/meson.build
+++ b/src/loaders/media/meson.build
@@ -9,6 +9,8 @@ source_file = [
 media_dep = []
 if system == 'darwin' or system == 'ios'
     subdir('apple')
+elif system == 'emscripten'
+    subdir('web')
 else
     # Use the fallback factory when no platform-specific media backend is available.
     source_file += ['tvgMediaLoader.cpp']
--- a/src/loaders/media/web/meson.build
+++ b/src/loaders/media/web/meson.build
@@ -1,0 +1,9 @@
+web_media_src = files(
+   'tvgWebMediaLoader.h',
+   'tvgWebMediaLoader.cpp'
+)
+
+media_dep += [declare_dependency(
+    include_directories : include_directories('.'),
+    sources : web_media_src
+)]


hermet (changes requested):
@tinyjin I only check the CAPI bindings commit.  Please polish up the API docs.
- Please properly add return and parameter info.
- Keep the doc style consistency. "\\" -> "@"
- Please use the tvg-format.

src/bindings/capi/thorvg_capi.h
```diff
@@ -3229,6 +3236,188 @@ TVG_API Tvg_Result tvg_lottie_animation_set_audio_resolver(Tvg_Animation animati
...
+ *
+ * The module enables control of host-decoded video playback.
+ *
+ * @{
+ */
+
+/**
+ * @brief Creates a new video playback controller.
+ *
+ * @return A new Tvg_Video object, or @c nullptr on failure.
```
hermet: ~or @c nullptr on failure.~

src/bindings/capi/thorvg_capi.h
```diff
@@ -3229,6 +3236,188 @@ TVG_API Tvg_Result tvg_lottie_animation_set_audio_resolver(Tvg_Animation animati
...
+ * @note Experimental API
+ */
+TVG_API Tvg_Video tvg_video_new(void);
+
+/**
+ * @brief Releases the given video object and its resources.
+ *
+ * @param[in] video The Tvg_Video object to release.
+ *
+ * @retval TVG_RESULT_INVALID_ARGUMENT An invalid @p video pointer.
```
hermet: Please remove this for all apis.

~* @retval TVG_RESULT_INVALID_ARGUMENT An invalid @p video pointer.~

src/bindings/capi/thorvg_capi.h
```diff
@@ -3229,6 +3236,188 @@ TVG_API Tvg_Result tvg_lottie_animation_set_audio_resolver(Tvg_Animation animati
...
+ * @note Experimental API
+ */
+TVG_API Tvg_Result tvg_video_del(Tvg_Video video);
+
+/**
+ * @brief Retrieves the picture that displays the video frames.
+ *
+ * The picture is owned by the video; add it to a canvas to render the frames.
+ *
+ * @param[in] video A Tvg_Video object.
```
hermet: A Tvg_Video object -> video objec.t

hermet (changes requested):
@tinyjin review capi bindings.

Please review your documentation again and keep it consistent style with the others. I added a few comments for your reference.

src/bindings/capi/tvgCapi.cpp
```diff
@@ -1307,6 +1310,139 @@ TVG_API Tvg_Result tvg_lottie_animation_set_audio_resolver(Tvg_Animation animati
     return TVG_RESULT_NOT_SUPPORTED;
 }
 
+/************************************************************************/
+/* Video API (Experimental)                                            */
```
hermet: ~(Experimental)~

tinyjin:
@hermet I've touched overall API description to keep it consistent with others. Could you review once more?

hermet:
@tinyjin please properly resolve the copilot review. thanks.

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+Result WebMediaLoader::open(uint32_t w, uint32_t h, float duration, ColorSpace cs)
+{
+    if (w == 0 || h == 0) return Result::InvalidArguments;
+
+    auto bytes = sizeof(uint32_t) * w * h;
+    tvg::free(surface.buf32);
+    surface.buf32 = tvg::malloc<uint32_t>(bytes);
+    if (!surface.buf32) return Result::FailedAllocation;
+    memset(surface.buf32, 0, bytes);
```
Nor-s:
```cpp
    surface.buf32 = tvg::malloc<uint32_t>(bytes);
    if (!surface.buf32) return Result::FailedAllocation;
    memset(surface.buf32, 0, bytes);
```

Could we use tvg::calloc() instead of tvg::malloc() followed by memset()?

Nor-s (commented):
@tinyjin  Could you please rebase this branch? 

Thank you.

src/loaders/media/web/tvgWebMediaLoader.cpp
tinyjin: @Nor-s I think video update rewrite the whole buffer. Memset is just useless at this context. Removed the line, and keep first line as is. Thank you!

tinyjin:
History - they are merged in base PR
- Updated API needed for Web Video support
- Added CAPI for ThorVG Video

| 42b7c302837874f3716de23cd36161d2db2e78d5 & 69fbead3d68297664beeddaab84b9f1bfdb094a3 are general updates that's also needed for Web Support (WebCanvas)

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,126 @@
...
+Result WebMediaLoader::volume(float volume)
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+    audioVolume = volume;
+    return Result::Success;
+}
+
+Result WebMediaLoader::mute(bool on)
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
```
hermet: This is a bad restriction. Player options can be configured at any time.

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,126 @@
...
+Result WebMediaLoader::loop(bool on)
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+    looping = on;
+    return Result::Success;
+}
+
+Result WebMediaLoader::volume(float volume)
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
```
hermet: This is an unnecessary restriction. Player options can be configured at any time.

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,126 @@
...
+Result WebMediaLoader::seek(float seconds)
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+    curTime = seconds;
+    return Result::Success;
+}
+
+Result WebMediaLoader::loop(bool on)
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
```
hermet: This is an unnecessary restriction. Player options can be configured at any time.

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,126 @@
...
+Result WebMediaLoader::pause()
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+    paused = true;
+    return Result::Success;
+}
+
+Result WebMediaLoader::stop()
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
```
hermet: Please align with the spec:
```
     * @retval Result::InsufficientCondition If the video is not loaded.
```

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,126 @@
...
+    paused = true;
+    curTime = 0.0f;
+    return Result::Success;
+}
+
+Result WebMediaLoader::seek(float seconds)
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
```
hermet: Please align with the spec:
```
     * @retval Result::InsufficientCondition If the video is not loaded.
     * @retval Result::InvalidArguments If @p seconds is out of range.
```

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,126 @@
...
+Result WebMediaLoader::play()
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+    paused = false;
+    return Result::Success;
+}
+
+Result WebMediaLoader::pause()
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
```
hermet: Please align with the spec:
```
     * @retval Result::InsufficientCondition If the video is not started playing.
```

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,126 @@
...
+    surface.cs = ColorSpace::ABGR8888S;
+    surface.premultiplied = false;
+    curTime = time;
+
+    return Result::Success;
+}
+
+Result WebMediaLoader::play()
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
```
hermet: Please align with the spec:
```
     * @retval Result::InsufficientCondition If the video is not loaded.
```

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,126 @@
...
+    this->totalTime = duration;
+
+    return Result::Success;
+}
+
+Result WebMediaLoader::update(float time)
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+
+    ScopedLock lock(surface.key);
```
hermet: 1. Surface configuration (colorspace and premultiplied alpha) must be specified during initialization (it seems already done in open() call). Shouldn't every update call.
2. Does `ScopedLock` essential here?

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    auto bytes = sizeof(uint32_t) * w * h;
+    tvg::free(surface.buf32);
+    surface.buf32 = tvg::malloc<uint32_t>(bytes);
+    if (!surface.buf32) return Result::FailedAllocation;
+
+    surface.stride = w;
+    surface.w = w;
+    surface.h = h;
+    surface.cs = ColorSpace::ABGR8888S;
```
hermet: 1. cs must be suggested by user like this:

```
Result WebMediaLoader::open(uint32_t w, uint32_t h, ColorSpace cs, float duration)
```

2. the open() method doesn't open anything, naming needs to be changed with something like `setup()`

hermet (changes requested):
@tinyjin Please check comments. thanks.

hermet:
@Nor-s, please check my comments above. It seems that AVF needs to align with the specification as well.

src/loaders/media/web/tvgWebMediaLoader.cpp
tinyjin: `surface.buf32` is available after `setup` (loaded), it's currently aligned with api spec.

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    surface.stride = w;
+    surface.w = w;
+    surface.h = h;
+    surface.cs = format = cs;
+    surface.channelSize = sizeof(uint32_t);
+    surface.premultiplied = _premultiplied(cs);
+
+    this->w = (float)w;
+    this->h = (float)h;
```
Nor-s: `(float)` -> `(static_cast)`

src/loaders/media/web/tvgWebMediaLoader.cpp
tinyjin: I've moved this to `sync`.
The surface config line is necessary since `tvgSwRaster` updates it in frame.

`ScopedLock` also, web thread preset runs under workers.

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,153 @@
...
+Result WebMediaLoader::play()
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+    paused = false;
+    return Result::Success;
+}
+
+Result WebMediaLoader::pause()
+{
+    if (paused) return Result::InsufficientCondition;
```
Nor-s: the paused is not a case. paused -> paused (success)
not started playing -> paused (insufficient condition)

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,153 @@
...
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+    paused = true;
+    curTime = 0.0f;
+    return Result::Success;
+}
+
+Result WebMediaLoader::seek(float seconds)
+{
+    if (seconds < 0.0f || (totalTime > 0.0f && seconds > totalTime)) return Result::InvalidArguments;
```
Nor-s: see: 

```
Result Video::seek(float seconds) noexcept
{
    FETCH_LOADER(Result::InsufficientCondition);
    if (seconds < 0.0f || seconds > loader->totalTime) return Result::InvalidArguments;
    return loader->seek(seconds);
}
```

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,153 @@
...
+    if (!surface.buf32) return Result::InsufficientCondition;
+    paused = true;
+    curTime = 0.0f;
+    return Result::Success;
+}
+
+Result WebMediaLoader::seek(float seconds)
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+    if (seconds < 0.0f || seconds > totalTime) return Result::InvalidArguments;
```
Nor-s: this must be api level handling. not here.

```
// tvg_video_seek
TVG_API Tvg_Result tvg_video_seek(Tvg_Video video, float seconds)
{
#ifdef THORVG_MEDIA_LOADER_SUPPORT
    if (video) return (Tvg_Result) reinterpret_cast<Video*>(video)->seek(seconds);
    return TVG_RESULT_INVALID_ARGUMENT;
#endif
    return TVG_RESULT_NOT_SUPPORTED;
}

// Vide::seek already check seconds
Result Video::seek(float seconds) noexcept
{
    FETCH_LOADER(Result::InsufficientCondition);
    if (seconds < 0.0f || seconds > loader->totalTime) return Result::InvalidArguments;
    return loader->seek(seconds);
}
```

Nor-s (approved):

src/loaders/media/web/tvgWebMediaLoader.h
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+#ifndef _TVG_WEB_LOADER_H_
+#define _TVG_WEB_LOADER_H_
+
+#include "tvgMediaLoader.h"
+
+struct WebMediaLoader : MediaLoader
+{
+    ColorSpace format = ColorSpace::Unknown;
```
hermet: MediaLoader already contains the colorspace value  -> `ImageLoader::RenderSurface::cs`.

You can use it.

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,152 @@
...
+ * SOFTWARE.
+ */
+
+#include "tvgWebMediaLoader.h"
+
+/************************************************************************/
+/* Internal Class Implementation                                        */
+/************************************************************************/
+
+static bool _validate(ColorSpace cs)
```
hermet: Please remove this. `WebMediaLoader` itself doesn't necessarily force a custom spec. If needed, you can enforce it in WebCanvas for your use case, or simply leave it to the canvas engine.

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,152 @@
...
+{
+    tvg::free(surface.buf32);
+}
+
+Result WebMediaLoader::setup(uint32_t w, uint32_t h, ColorSpace cs, float duration)
+{
+    if (w == 0 || h == 0 || duration <= 0.0f || !_validate(cs)) return Result::InvalidArguments;
+
+    auto buffer = tvg::calloc<uint32_t>(w * h, sizeof(uint32_t));
+    if (!buffer) return Result::FailedAllocation;
```
hermet: remove this

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,152 @@
...
+WebMediaLoader::~WebMediaLoader()
+{
+    tvg::free(surface.buf32);
+}
+
+Result WebMediaLoader::setup(uint32_t w, uint32_t h, ColorSpace cs, float duration)
+{
+    if (w == 0 || h == 0 || duration <= 0.0f || !_validate(cs)) return Result::InvalidArguments;
+
+    auto buffer = tvg::calloc<uint32_t>(w * h, sizeof(uint32_t));
```
hermet: you can use the `realloc()`

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,152 @@
...
+    frameUpdated = true;
+
+    return Result::Success;
+}
+
+bool WebMediaLoader::sync()
+{
+    if (!frameUpdated) return false;
+
+    ScopedLock lock(surface.key);
```
hermet: Any reason to use lock here? Otherwise please remove it.

hermet (changes requested):
@tinyjin please check more comments. thanks

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,152 @@
...
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+    paused = true;
+    curTime = 0.0f;
+    return Result::Success;
+}
+
+Result WebMediaLoader::seek(float seconds)
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
```
hermet: API already handles this. Not necessary

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,152 @@
...
+Result WebMediaLoader::pause()
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+    paused = true;
+    return Result::Success;
+}
+
+Result WebMediaLoader::stop()
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
```
hermet: API already handles this. Not necessary

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,152 @@
...
+    surface.cs = format;
+    surface.premultiplied = _premultiplied(format);
+    frameUpdated = false;
+
+    return true;
+}
+
+Result WebMediaLoader::play()
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
```
hermet: API already handles this. Not necessary

src/loaders/media/web/tvgWebMediaLoader.cpp
tinyjin: In the API, only the loader state is checked, but on the web the glue layer does not handle loading directly, so the order of video loading and setup() calls cannot be guaranteed.

surface.buf32 is same to checking whether the video load is ready. The glue layer does not perform the actual load, but it  indicating condition that loading has happened.

Without this, we'd miss the exception case where the user loads the video but skips setup(). If we treat that as an API constraint, it's removable.

src/loaders/media/web/tvgWebMediaLoader.h
tinyjin: That value can be overwritten with targeted colorspace in SW engine. Since video play rewrites buffer every frame, we need to preserve the original cs from `setup`(load) in other place.

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,141 @@
...
+    return Result::Success;
+}
+
+bool WebMediaLoader::sync()
+{
+    if (!frameUpdated) return false;
+
+    surface.cs = format;
+    surface.premultiplied = _premultiplied(format);
+    frameUpdated = false;
```
hermet: sync() needs to return the updated condition.
let's return the `frameUpdated` value.
```
auto ret = frameUpdated;
frameUpdated = false;
return ret;
```

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,141 @@
...
+Result WebMediaLoader::play()
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
+    paused = false;
+    return Result::Success;
+}
+
+Result WebMediaLoader::pause()
+{
+    if (!surface.buf32) return Result::InsufficientCondition;
```
hermet: this exception is not necessary here.

src/loaders/media/web/tvgWebMediaLoader.cpp
hermet: seek, stop other playback bindings needs to get through APIs. exceptions would be happends in ThorVG APIs.
[WebCanvas API - ThorVG APIs - glue layer methods - WebCanvas Integration]

src/loaders/media/web/tvgWebMediaLoader.h
hermet: It sounds like the integration is not tuned because the engine’s color space and the video’s color space are not properly aligned. Please double-check whether the media codec configures its output color space to match the current color-space setting of the WebCanvas backend.

hermet (changes requested):
@tinyjin Please double-check comments. Thanks.

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,140 @@
...
+    this->h = static_cast<float>(h);
+    totalTime = duration;
+    curTime = 0.0f;
+    frameUpdated = false;
+    paused = true;
+
+    return Result::Success;
+}
+
+Result WebMediaLoader::update(float time)
```
hermet: Result WebMediaLoader::update(uint32_t* src, float time)

hermet (commented):
@tinyjin WebMediaLoader needs to override:
```
virtual bool Loader::open(const char* data, uint32_t size, const LoaderOps* ops, bool copy);
virtual bool read()
virtual bool close()
```

src/loaders/media/web/tvgWebMediaLoader.cpp
tinyjin: Then looks like something like this needed:
```cpp
struct MediaOps : LoaderOps
{
    uint32_t w, h;
    float duration;
    ColorSpace cs;
};
```
web doesn't read those data directly, should have interface to push data from JS to WASM

hermet: Please check. video player interface  must be consistent for the integration.
web >picture::load()->media gluelayer-> wasm (callback) -> web -> media gluelayer::setup()

tinyjin: @hermet 
lets say we send 4K clip (2~4GB) in `open(const char* data, uint32_t size, const LoaderOps* ops, bool copy);` as `data`.

In the build target wasm32, It will exceed the heap limit (4GB). Despite of data is just for passing instantly.
i see it's a current limitations of WebAssembly. (If we want native pointer, anyhow it should be uploaded in WASM heap)

I initially reviewed that approach, I came up with current approach:
WASM doesn't dominate video binary data, but receives frame buffer.

Or If we just aim video under minimum size (e.g <= 1GB), its possible.

tinyjin: > Result WebMediaLoader::update(uint32_t* src, float time)

Plus, the JS allocated buffer memory can be reused, so only added pointer in `setup`

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+EM_JS(void, _web_player_mute, (int handle, bool on), {
+    Module["webPlayers"]["table"].get(handle)["mute"](!!on);
+});
+
+/************************************************************************/
+/* External Class Implementation                                        */
+/************************************************************************/
+
+WebPlayer* WebPlayer::gen(WebMediaLoader* loader, const char* data, uint32_t size)
```
hermet: let's unity this method with `open()`

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,210 @@
...
+void WebPlayer::mute(bool on)
+{
+    _web_player_mute(handle, on);
+}
+
+
+WebMediaLoader::~WebMediaLoader()
+{
+    delete(player);
+    tvg::free(surface.buf32);
```
hermet, line 46: destroy this memory by the owner.

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,210 @@
...
+{
+    if (player) return false;
+    player = WebPlayer::gen(this, data, size);
+    return player != nullptr;
+}
+
+bool WebMediaLoader::sync()
+{
+    auto ret = frameUpdated;
+    if (ret) surface.premultiplied = _premultiplied(surface.cs);
```
hermet: --

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,210 @@
...
+WebMediaLoader::~WebMediaLoader()
+{
+    delete(player);
+    tvg::free(surface.buf32);
+}
+
+bool WebMediaLoader::open(const char* data, uint32_t size, const LoaderOps* ops, bool copy)
+{
+    if (player) return false;
+    player = WebPlayer::gen(this, data, size);
```
hermet: //  where is the surface setting?
open() / bitmap() / sync()

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,210 @@
...
+    Module["webPlayers"]["table"].get(handle)["loop"](!!on);
+});
+
+EM_JS(void, _web_player_volume, (int handle, float volume), {
+    Module["webPlayers"]["table"].get(handle)["volume"](volume);
+});
+
+EM_JS(void, _web_player_mute, (int handle, bool on), {
+    Module["webPlayers"]["table"].get(handle)["mute"](!!on);
+});
```
hermet:
```
EM_JS(void, _web_player_configure_surface, (.......), {
   ....
});
```

src/loaders/media/web/tvgWebMediaLoader.cpp
tinyjin: Loader owns `surface.buf32`

src/loaders/media/web/tvgWebMediaLoader.cpp
tinyjin: Now, loader directly write buffer. such an interface looks no longer needed
```
see: WebPlayer::sync
```

tinyjin:
Enhanced glue interface, without `EM_JS`. Brought more type-safety and structural contract defined with `emscripten::val` style glue.

Users can implement web media playback in any backend, by just completing this bridge:
```js
Module.createMediaPlayer = (loader, bytes) => new MyPlayer(bytes);

class MyPlayer {
  sync()      // called once per render (state sync)
  play()
  pause()
  stop()
  seek(sec)
  loop(on)
  volume(v)   // 0 - 1
  mute(on)
  dispose()
}
```

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,211 @@
...
+}
+
+RenderSurface* WebMediaLoader::bitmap()
+{
+    if (surface.buf32) return BitmapLoader::bitmap();
+
+    uint32_t size[2];
+    float duration, time;
+    if (player->sync(nullptr, size, &time, &duration)) {
+        surface.buf32 = tvg::calloc<uint32_t>(size[0] * size[1], sizeof(uint32_t));
```
hermet: Please check 
`void RenderSurface::setup(pixel_t* data, uint32_t stride, uint32_t w, uint32_t h, uint8_t channelSize, ColorSpace cs, bool alphaIgnored = false)`

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if (player->sync(surface.buf32, nullptr, &time, nullptr) < 2) return false;
+
+    curTime = time;
+    surface.premultiplied = false;
+    return true;
+}
+
+RenderSurface* WebMediaLoader::bitmap()
+{
+    if (surface.buf32) return BitmapLoader::bitmap();
```
hermet: please make this style code:
```
if (!surface.buf32) {
   ...
}
Bitmaploader::bitmap()
```

src/loaders/media/web/tvgWebMediaLoader.cpp
hermet: how to change the buffer if the video size is changed?

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    return new WebPlayer{WebMediaPlayer(js)};
+}
+
+WebPlayer::~WebPlayer()
+{
+    js.call<void>("dispose");
+}
+
+int WebPlayer::sync(uint32_t* buf, uint32_t* size, float* time, float* duration)
```
hermet: This is a bad symptom. Are pointers essential here for `size`, `time`, `duration`?

hermet (commented):
@tinyjin could you please check this one (https://github.com/thorvg/thorvg/pull/4656)?
https://github.com/thorvg/thorvg/pull/4656/changes/2c87ab70ac279659f00390bdd4162477db8a9ed1

src/loaders/media/web/tvgWebMediaLoader.cpp
tinyjin: Updated to use this

tinyjin: valid, added logic in `sync()`

src/loaders/media/web/tvgWebMediaLoader.cpp
```diff
@@ -0,0 +1,143 @@
...
+{
+    auto state = js.call<WebMediaState>("sync");
+    if (state.isNull()) return false;
+
+    totalTime = state["duration"].as<float>();
+
+    auto width = state["width"].as<uint32_t>();
+    auto height = state["height"].as<uint32_t>();
+    if (width != surface.w || height != surface.h) {
+        tvg::free(surface.buf32);
```
hermet: please use realloc()

hermet (approved):
@tinyjin please check a comment. thanks.

PR #4505, opened 2026-06-26
gl_engine: align drop shadow uniform layout

Nor-s:
issue: #4129

Firefox/WebGL reported Buffer for uniform block is smaller than `UNIFORM_BLOCK_DATA_SIZE` when rendering the drop shadow effect. 

`GlDropShadow` uniform data was missing std140 tail padding, so its bound range was `40` bytes while the shader block size required `48` bytes.

|before | after |
|-|-|
|image|image|

### DIFF
--- a/src/renderer/gpu_engine/gl/tvgGlEffect.cpp
+++ b/src/renderer/gpu_engine/gl/tvgGlEffect.cpp
@@ -109,6 +109,7 @@ GlRenderTask* GlEffect::render(RenderEffectGaussianBlur* effect, GlRenderTarget*
 struct GlDropShadow: GlGaussianBlur {
     float color[4];
     float offset[2];
+    float dummy1[2];  // padding for std140 block alignment
 };
 
 


Nor-s:
| before| after|
|-|-|
|image|image|

@wenjieshen 

I was able to reproduce the same issue in Chrome on a Linux machine(NixOS, intel integrated GPU), 

and confirmed that the drop shadow has been fixed by the changes in this patch.

wenjieshen:
> 
> 
> | before| after|
> 
> |-|-|
> 
> |image|image|
> 
> 
> 
> @wenjieshen 
> 
> 
> 
> I was able to reproduce the same issue in Chrome on a Linux machine(NixOS, intel integrated GPU), 
> 
> 
> 
> and confirmed that the drop shadow has been fixed by the changes in this patch.

Thank you. It is a very good news. I will test it after I can use Linux machine. 🙏

src/renderer/gpu_engine/gl/tvgGlEffect.cpp
```diff
@@ -109,7 +109,9 @@ GlRenderTask* GlEffect::render(RenderEffectGaussianBlur* effect, GlRenderTarget*
 struct GlDropShadow: GlGaussianBlur {
     float color[4];
     float offset[2];
+    float dummy1[2];
```
wenjieshen: Could it be `offset[4]` to align with std140? Or could you please name it a meaningful name, like `alignment`, `padding`, etc. 

Thank you.

wenjieshen (approved):
Thank you for the contribution.
I didn't observe any regression or performance drop.
There is a readiablity suggestion, could you please consider it? Thank you.

src/renderer/gpu_engine/gl/tvgGlEffect.cpp
Nor-s: Because it aligns with the dummy0 naming in `GlGaussianBlur`, I think this is better.

```cpp
struct GlGaussianBlur {
    float sigma{};
    float scale{};
    float extend{};
    float dummy0{};
};
```

Thank you for your review.

wenjieshen: Hi @Nor-s, 
Good catch, thanks for fixing this.

Instead of adding `dummy1`, could we try using `alignas(16)` for the CPU-side std140 block? We already use `alignas` in a few GL uniform block structs, so this would keep the layout handling consistent with the existing code.

For example:

```cpp
struct alignas(16) GlGaussianBlur {
    float sigma{};
    float scale{};
    float extend{};
};

struct alignas(16) GlDropShadow : GlGaussianBlur {
    alignas(16) float color[4]{};
    alignas(8) float offset[2]{};
};

static_assert(sizeof(GlGaussianBlur) == 16);
static_assert(sizeof(GlDropShadow) == 48);
```

`alignas(16)` tells the C++ compiler to align the struct to a 16-byte boundary. Since std140 uniform blocks are also rounded to 16-byte alignment, this lets the compiler provide the trailing padding naturally, instead of adding manual dummy fields.

I think this makes the layout intention clearer, and also helps us avoid growing a `dummy0`, `dummy1`, `dummy2` family every time a std140 block needs alignment.

hermet: @Nor-s Please add appropriate comments to any unusual code. Thanks.

Nor-s: Hello @wenjieshen , I reviewed the use of `alignas`, but there are already many places in the GL shader code that use `dummy` fields.

Since this patch is a bug fix, changing that now would likely expand the scope of the patch. 
I will separate that cleanup and submit it in a separate PR.

Thank you for the review.

hermet (approved):

hermet:
@Nor-s good job!

PR #4508, opened 2026-06-26
gl_engine: reject empty clips before drawing

wenjieshen:
Reject shape and image paints early when any clip has no drawable geometry, matching the renderer behavior for empty clips.

With that guarantee in place, simplify drawClip() by removing the defensive non-drawable skip and lazy mask quad setup.

https://github.com/user-attachments/assets/e1d60f7e-7497-4924-8046-564d7a624a8b

Related issue: https://github.com/thorvg/thorvg/issues/4442

### DIFF
--- a/src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
+++ b/src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
@@ -42,6 +42,21 @@ static constexpr float IDENTITY_VERTEX[] = {-1.f, 1.f, -1.f, -1.f, 1.f, 1.f, 1.f
 static constexpr uint32_t RECT_INDEX[] = {0, 1, 2, 2, 1, 3};
 static constexpr uint32_t RECT_INDEX_COUNT = sizeof(RECT_INDEX) / sizeof(RECT_INDEX[0]);
 
+static bool _skipRender(const Array<RenderData>& clips)
+{
+    if (clips.empty()) return false;
+
+    // Clip geometries are prepared during update(), before this paint renders.
+    // Clip masks are intersected, so one empty clip makes the paint invisible.
+    ARRAY_FOREACH(p, clips) {
+        auto clip = static_cast<GlShape*>(*p);
+        auto flag = (clip->geometry.stroke.vertex.count > 0) ? RenderUpdateFlag::Stroke : RenderUpdateFlag::Path;
+        if (!clip->geometry.drawable(flag)) return true;
+    }
+
+    return false;
+}
+
 void GlRenderer::disposeTexture(GLuint texId)
 {
     if (!texId) return;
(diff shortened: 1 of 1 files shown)


src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
```diff
@@ -42,6 +42,17 @@ static constexpr float IDENTITY_VERTEX[] = {-1.f, 1.f, -1.f, -1.f, 1.f, 1.f, 1.f
 static constexpr uint32_t RECT_INDEX[] = {0, 1, 2, 2, 1, 3};
 static constexpr uint32_t RECT_INDEX_COUNT = sizeof(RECT_INDEX) / sizeof(RECT_INDEX[0]);
 
+static bool _clipsRejectAll(const Array<RenderData>& clips)
```
hermet: 1. This function name is too ambiguous. Please add a descriptive comment explaining what it does, or rename it for clear purpose (e.g., rejectRender / skipRender).
2. You might want to move the empty() check into this function.

src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
```diff
@@ -42,6 +42,17 @@ static constexpr float IDENTITY_VERTEX[] = {-1.f, 1.f, -1.f, -1.f, 1.f, 1.f, 1.f
...
+{
+    ARRAY_FOREACH(p, clips) {
+        auto clip = static_cast<GlShape*>(*p);
+        auto flag = (clip->geometry.stroke.vertex.count > 0) ? RenderUpdateFlag::Stroke : RenderUpdateFlag::Path;
```
hermet, line 53: Just to confirm, are the clips info are valid at this stage? (looks drawClip happens later)

hermet (commented):
@wenjieshen Please check comments.

src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
wenjieshen: Thank you for the suggestion. I have updated as suggested.

src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
wenjieshen: Yes, the clip info is valid there.

src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
```diff
@@ -42,6 +42,21 @@ static constexpr float IDENTITY_VERTEX[] = {-1.f, 1.f, -1.f, -1.f, 1.f, 1.f, 1.f
 static constexpr uint32_t RECT_INDEX[] = {0, 1, 2, 2, 1, 3};
 static constexpr uint32_t RECT_INDEX_COUNT = sizeof(RECT_INDEX) / sizeof(RECT_INDEX[0]);
 
+static bool _skipRender(const Array<RenderData>& clips)
```
hermet, line 45: Probably, we can move this check to the prepare() stage (not in render) so that we can return early and avoid unnecessary overhead...

hermet (approved):

Issue #2928, opened 2024-11-04
svg: wrong clippath test result


hermet:
ThorVG:
![image](https://github.com/user-attachments/assets/ffad9227-03fe-4545-9bc1-9ca66980776d)

Expected:
![clippath](https://github.com/user-attachments/assets/f4451300-1165-4c3c-81a8-9606d26fe314)

JSUYA:
Could you also upload svg file?

hermet:
@JSUYA the second image is svg. you can "save as"

JSUYA:
There are two issues related to clippath in the above sample. One of them is that we do not yet support clipPath element having clip-path attribute. We need additional implementation for this part.
```
...
<clipPath id="clipIntersection" clip-path="url(#clip1)">
    <use x="0" y="0" width="200" height="200" xlink:href="#clip2Shape" />
</clipPath>
...
```

JSUYA:
The second one seems to require the union of path feature.
When `<clipPath>` has multiple children, we add the shape path(point) info of each child to the clipper.
In this case, only the intersection of the shape paths is output.
```
<clipPath id="clipUnion">
    <polygon id="clip1Shape" points="100,10 40,180 190,60 10,60 160,180 100,10" stroke="blue" />
    <circle id="clip2Shape" cx="100" cy="100" r="65" />
</clipPath>

<rect x="10" y="10" width="180" height="180" fill="red" opacity="0.5"
    clip-path="url(#clipUnion)" transform="translate(400)" />

```

I tested duplicating the target, clipping each one, and then putting(stacking) it in one scene, but the result is different when opacity is applied, and it seems that the memory usage will increase when clipping the image.

I am researching whether the logic for calculating the outline of the sw renderer can be reused. Is it looks possible?

hermet:
@JSUYA Did you check these options?

`MaskingMethods.cpp`
![image](https://github.com/user-attachments/assets/3c3d8464-40e2-4eaa-af4f-ba22684cf37a)

hermet:
@JSUYA any opinion?

PR #4509, opened 2026-06-26
svg: fix clipPath union and nested clip-path

JSUYA:
Apply clip regions as alpha masks instead of a single clipper shape:

- Union: each child becomes a separate opaque shape, so multiple children form their union instead of cancelling via a shared fill-rule.
- Intersection: a clipPath with its own clip-path attribute is applied by wrapping the target and masking again.
- Composition: nested clipPaths and SVG masks are preserved together by wrapping before the second mask.

Validation:

- Rendered 17 focused SVG cases with ThorVG, Chrome, and librsvg/Cairo.
- Covered union, nested and child clipping, clipPathUnits, transforms, use, mask, filter, pattern, opacity, basic shapes, and cyclic references.
- Ran unit tests and ASan/UBSan builds locally.

issue: #2928

### DIFF
--- a/src/loaders/svg/tvgSvgBuilder.cpp
+++ b/src/loaders/svg/tvgSvgBuilder.cpp
@@ -36,6 +36,7 @@
 /************************************************************************/
 
 static bool _appendClipShape(SvgParserContext& ctx, SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath, const Matrix* transform);
+static bool _applyClip(SvgParserContext& ctx, Paint* paint, Paint* content, const SvgNode* node, const SvgNode* clipNode, const Box& vBox, const string& svgPath, Paint** result);
 static Scene* _sceneBuildHelper(SvgParserContext& ctx, const SvgNode* node, const Box& vBox, const string& svgPath, bool mask, int depth);
 static Paint* _applyPatternProperty(SvgParserContext& ctx, Shape* vg, SvgNode* node, SvgNode* patternNode, const Box& vBox, const string& svgPath);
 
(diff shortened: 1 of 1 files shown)


hermet:
@JSUYA Could you please rebase this? Thanks.

JSUYA:
> @JSUYA Could you please rebase this? Thanks.

Hi, i had rebased branch.

src/loaders/svg/tvgSvgBuilder.cpp
```diff
@@ -35,7 +35,7 @@
 /* Internal Class Implementation                                        */
 /************************************************************************/
 
-static bool _appendClipShape(SvgParserContext& ctx, SvgNode* node, Shape* shape, const Box& vBox, const string& svgPath, const Matrix* transform);
```
hermet, line 38: This method doesn't seem necessary to change. Otherwise please explain why this change is benefit. Thanks.

hermet (commented):
@JSUYA Please check the comments.

src/loaders/svg/tvgSvgBuilder.cpp
JSUYA: I will proceed with refactoring.
For these functions, the return value will be whether the function succeeded or failed, and the necessary paint objects will be passed as parameters.

hermet: I don't see any functional difference either. The previous version is more intuitive to me. If there's no compelling reason for this change, please keep the original implementation. Thanks.

JSUYA: I have modified code to keep the original implementation. Please review it again. Thank you.

hermet (changes requested):
@JSUYA could you please double-check this?

before:

image


after: 

Image

svg:

test_embedded_image_with_clipper

hermet:
> @JSUYA could you please double-check this?
> 
> before:
> 
> image
> after:
> 
> Image
> svg:
> 
> test_embedded_image_with_clipper

@Nor-s could you please add this sample in our pixel inspector? thanks.

JSUYA:
There was a problem with the transform calculation step in complex situations involving clippath, mask, use, and symbol.
I have updated the PR. and I checked output even on the problematic SVG.

image


This is a SVG image created by selecting a few cases that can be simply verified among the possible combination scenarios.

(screenshot)
image

(svg)

clip-path-regression

hermet (approved):

Issue #3342, opened 2025-03-21
svg: spec coverage++


hermet:
thorvg:

![Image](https://github.com/user-attachments/assets/3f69f0d5-48e7-4ca4-945e-1fd1a7314de0)

expected:

![Image](https://github.com/user-attachments/assets/38a09581-87e7-428f-90a4-35b56f579059)

sample: https://github.com/thorvg/thorvg/blob/main/examples/resources/svg/juanmontoya_lingerie.svg

hermet:
+ @JSUYA

PR #4510, opened 2026-06-26
svg_loader: support the maskUnits attribute

JSUYA:
The mask userspace was  unsupported, so the mask covered the whole element instead of being clipped to its region as the SVG spec and browsers do.

Parse the region and clip the mask to it: scaled by the object bounding box for objectBoundingBox, or in user space for userSpaceOnUse.

issue : #3342


hermet (approved):
@JSUYA nice!

Issue #4478, opened 2026-06-20
gpu: tritone color effect bug


hermet:
This result looks remarkably different. We need to clarify this:

<img width="2042" height="545" alt="image" src="https://github.com/user-attachments/assets/d34707bc-7e29-447c-8898-5eabac4433df" />

_Originally posted by @hermet in https://github.com/thorvg/thorvg/issues/4453#issuecomment-4748531282_

hermet:
Both GL, WGPU engines often visualizes the same resource differently, resulting in different outputs.

correct:
<img width="202" height="178" alt="Image" src="https://github.com/user-attachments/assets/6a68f334-df4a-4be6-bf3e-42147377c5b2" />

wrong:
<img width="202" height="178" alt="Image" src="https://github.com/user-attachments/assets/dc4789d8-237b-461a-ab84-5034f5ee0847" />

hermet:
This seems observable after this:

```
commit b9c84b2b499c8b276489c0e101a9ab1e213c6456 (HEAD)
Author: Jinny You <jinny@lottiefiles.com>
Date:   Tue Feb 3 16:59:39 2026 +0900

    lottie/slot: Support Text Style slot overriding
```

cc: @tinyjin

wenjieshen:
Hi @hermet ,

I cannot reproduce it on Mac and Windows.
It might be resolved on the main branch.

https://github.com/user-attachments/assets/72b78a9e-6c48-4575-ad9c-977a825d3324

hermet:
@Nor-s I captured the diff report from the latest PR. Could you please double-check whether this is a false alarm?

<img width="2549" height="752" alt="Image" src="https://github.com/user-attachments/assets/22c92112-f948-4f08-9463-18a0bcd7e219" />

hermet:
https://github.com/thorvg/thorvg.pixel-inspector/pull/10

PR #4512, opened 2026-06-26
lottie: tritone color effect bug

Nor-s:
#4478

### Summary

This was an issue that occurred very randomly.

The asset did not contain an element corresponding to blend, which caused a garbage value to be used. 
As a result, a bug occurred where the color changed randomly.

This issue was reproduced only in the Ubuntu environment and the WASM web environment. (CPU, GL, WG)

Setting blend = 0 should be sufficient, but I think we need to initialize default values for structures where certain elements may be missing.

### Screenshot

|before | after| 
|-|-|
|Screenshot 2026-06-27 at 04 24 32|Screenshot 2026-06-27 at 04 25 58|

### Details

The asset contains only three tritone effect properties:

```json
// triton.json
"ef": [
                {
                    "ty": 23,
                    "en": 1,
                    "ef": [
                        {
                            "ty": 2,
                            "v": {
                                "k": [
                                    1,
                                    0.8,
                                    1,
                                    1
                                ],
                                "a": 0
                            }
                        },
                        {
                            "ty": 2,
                            "v": {
                                "k": [
                                    0.1,
                                    0.8,
                                    1,
                                    1
                                ],
                                "a": 0
                            }
                        },
                        {
                            "ty": 2,
                            "v": {
                                "k": [
                                    0.4,
                                    0.8,
                                    0,
                                    1
                                ],
                                "a": 0
                            }
                        }
                    ]
                }
            ],
```

The parser maps tritone properties by array index:

```cpp
  //src/loaders/lottie/tvgLottieParser.cpp:1491
  
  if (idx == 0) parsePropertyInternal(tritone->bright);
  else if (idx == 1) parsePropertyInternal(tritone->midtone);
  else if (idx == 2) parsePropertyInternal(tritone->dark);
  else if (idx == 3) parsePropertyInternal(tritone->blend);
```

Since idx == 3 is missing in this asset, blend keeps its default value. The builder later uses it when adding the render effect:

```cpp
  // src/loaders/lottie/tvgLottieBuilder.cpp
            case LottieEffect::Tritone: {
                auto effect = static_cast<LottieFxTritone*>(*p);
                auto dark = effect->dark(frameNo);
                auto midtone = effect->midtone(frameNo);
                auto bright = effect->bright(frameNo);
                layer->scene->add(SceneEffect::Tritone, dark.r, dark.g, dark.b, midtone.r, midtone.g, midtone.b, bright.r, bright.g, bright.b, (int)effect->blend(frameNo));
                break;
            }
```

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.h
+++ b/src/loaders/lottie/tvgLottieModel.h
@@ -205,7 +205,7 @@ struct LottieFxTritone : LottieEffect
     LottieColor bright;
     LottieColor midtone;
     LottieColor dark;
-    LottieOpacity blend;
+    LottieOpacity blend = 0;
 
     LottieFxTritone()
     {


hermet (commented):
@Nor-s Please fix the asset to have a blend value.

Nor-s:
> @Nor-s Please fix the asset to have a blend value.

https://github.com/thorvg/thorvg.pixel-inspector/pull/10

I’ve opened a PR with the changes in pixel-tool.
Thank you.

hermet:
@Nor-s Let's discard this change today. We would not judge the default value by the engine.

PR #4527, opened 2026-07-04
gl_engine: Fix GL_CHECK usage in conditional statements

xiaozhuai:
## Summary

This fixes unsafe uses of `GL_CHECK` inside conditional statements in the GL renderer.

In debug builds, `GL_CHECK` expands to more than one statement:

```cpp
stmt;
assert(glGetError() == GL_NO_ERROR);
```

When it is used directly as a single-line `if` or `if/else` body, the macro expansion can either break the `if/else` control flow or cause the error check to run outside the intended condition.

This PR wraps the affected `GL_CHECK` calls in braces so the expanded statements remain inside the correct conditional blocks.

## Trigger

https://github.com/thorvg/thorvg/blob/e753da99ba8ea19905a69f9eb76a51b1470e8e74/src/renderer/gpu_engine/gl/tvgGlRenderTask.cpp#L97-L98

### DIFF
--- a/src/renderer/gpu_engine/gl/tvgGlRenderTask.cpp
+++ b/src/renderer/gpu_engine/gl/tvgGlRenderTask.cpp
@@ -94,8 +94,11 @@ void GlRenderTask::run()
         }
     }
 
-    if (mUseDrawArrays) GL_CHECK(glDrawArrays(mArrayMode, mArrayOffset, mIndexCount));
-    else GL_CHECK(glDrawElements(GL_TRIANGLES, mIndexCount, GL_UNSIGNED_INT, reinterpret_cast<void*>(mIndexOffset)));
+    if (mUseDrawArrays) {
+        GL_CHECK(glDrawArrays(mArrayMode, mArrayOffset, mIndexCount));
+    } else {
+        GL_CHECK(glDrawElements(GL_TRIANGLES, mIndexCount, GL_UNSIGNED_INT, reinterpret_cast<void*>(mIndexOffset)));
+    }
 
     // setup attribute layout
     for (uint32_t i = 0; i < mVertexLayout.count; i++) {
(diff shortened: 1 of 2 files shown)


wenjieshen (changes requested):
Thanks for updating the patch. 🙏 

Could we keep this change limited to adding braces around the affected if / else call sites instead of changing the `GL_CHECK` macro like [0279e92](https://github.com/thorvg/thorvg/commit/0279e92f5de31bcc6ad25b6c038351fed2bb93fa) ?

`GL_CHECK` is also used for declarations in places like `tvgGlProgram.cpp`, so changing the macro forces unrelated code changes there. 
Since the issue is only with conditional call sites, adding braces seems like the smaller and safer fix for this PR.

xiaozhuai:
> Could we keep this change limited to adding braces around the affected if / else call sites instead of changing the GL_CHECK macro like [0279e92](https://github.com/thorvg/thorvg/commit/0279e92f5de31bcc6ad25b6c038351fed2bb93fa) ?

It’s very risky, and the same problem might come up again later. That's why I update this pr and change the `GL_CHECK` macro instead.

That's fine with me. If you're sure you want to do it this way, I'll revert it to https://github.com/thorvg/thorvg/commit/0279e92f5de31bcc6ad25b6c038351fed2bb93fa

wenjieshen:
> > Could we keep this change limited to adding braces around the affected if / else call sites instead of changing the GL_CHECK macro like [0279e92](https://github.com/thorvg/thorvg/commit/0279e92f5de31bcc6ad25b6c038351fed2bb93fa) ?
> 
> It’s very risky, and the same problem might come up again later. That's why I update this pr and change the `GL_CHECK` macro instead.
> 
> That's fine with me. If you're sure you want to do it this way, I'll revert it to [0279e92](https://github.com/thorvg/thorvg/commit/0279e92f5de31bcc6ad25b6c038351fed2bb93fa)

Thanks, I understand the concern.

Yes, please revert this PR to the localized bracing fix in 0279e92.

I think adding braces is enough for this issue because it fixes the actual problematic `if` / `else` call sites without changing the existing semantics of `GL_CHECK`. Changing the macro is a broader preventive change, and it also forces unrelated updates such as the `tvgGlProgram.cpp` changes.

For this PR, I’d prefer the smaller-scoped fix. Thank you for the contribution. 🙏

xiaozhuai:
> Yes, please revert this PR to the localized bracing fix in 0279e92.

Done~

wenjieshen (approved):
Thank you for the contribution. 🙏 
I also ran examples on macOS manually, and they remain correct.

PR #4546, opened 2026-07-10
test: fix the GL/WG unit tests on macOS

Nor-s:
### [test: support the GL unit tests on macOS](https://github.com/thorvg/thorvg/commit/1a483b7edeef144100414eb5050be4f27cdfc044)

The EGL cannot create a desktop GL context on macOS
due to the EGL compatibility (no system EGL, ANGLE is GLES-only).
Introduce a CGL-based harness to make the tests work with
desktop GL on macOS, rendering into an offscreen FBO.
The EGL path remains for Linux and GLES.

### [test: fix wgpu crash on macOS](https://github.com/thorvg/thorvg/commit/fb14dccf0f5690ecf4936c5bef849f25181e7225)

WG Basic Draw and WG Image Draw processed every blend and mask
combination in a single draw call. On macOS, this exceeded wgpu's
internal Metal command buffer limit and caused:

    Error in wgpuQueueSubmit: Validation Error
    Caused by: Parent device is lost

Counting pass creation in beginRenderPassMS() and beginRenderPass(),
together with copy encoding regions between passes, confirmed:

    Metal backend limit: 4,096
    WG Basic Draw:       >4,096
    WG Image Draw:       >4,096

WebGPU does not define this command buffer limit. It is internal to
wgpu's Metal backend.

Run draw, sync, and remove after each blend method to bound the number
of pending command buffers.

References:
- https://github.com/gfx-rs/wgpu/pull/8574

### DIFF
--- a/test/meson.build
+++ b/test/meson.build
@@ -27,21 +27,26 @@ test_file = [
 ]
 
 if gl_engine
-    test_egl_dep = dependency('egl', required: false)
-
-    if gl_variant == 'OpenGL ES'
-        test_gl_dep = dependency('glesv2', required: false)
-        if not test_gl_dep.found()
-            test_gl_dep = cc.find_library('GLESv2', required: false)
-        endif
-    else
-        test_gl_dep = dependency('gl', required: false)
-    endif
-
-    if test_egl_dep.found() and test_gl_dep.found()
-        test_dep += [test_egl_dep, test_gl_dep]
+    if host_machine.system() == 'darwin'
+        test_dep += dependency('appleframeworks', modules: ['OpenGL'])
         test_compiler_flags += ['-DTHORVG_GL_TEST_SUPPORT']
         test_file += ['testGlCanvas.cpp', 'testGlEngine.cpp']
+    else
+        test_egl_dep = dependency('egl', required: false)
+        if gl_variant == 'OpenGL ES'
+            test_gl_dep = dependency('glesv2', required: false)
+            if not test_gl_dep.found()
+                test_gl_dep = cc.find_library('GLESv2', required: false)
+            endif
+        else
+            test_gl_dep = dependency('gl', required: false)
+        endif
+
+        if test_egl_dep.found() and test_gl_dep.found()
+            test_dep += [test_egl_dep, test_gl_dep]
+            test_compiler_flags += ['-DTHORVG_GL_TEST_SUPPORT']
+            test_file += ['testGlCanvas.cpp', 'testGlEngine.cpp']
+        endif
     endif
 endif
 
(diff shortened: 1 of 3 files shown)


test/meson.build
```diff
@@ -27,21 +27,31 @@ test_file = [
 ]
 
 if gl_engine
-    test_egl_dep = dependency('egl', required: false)
-
     if gl_variant == 'OpenGL ES'
```
hermet: Please divide them into equivalent comparison groups
```
if host machine == darwin  //machine compare
   ........
else  // other machines
  if gl_variant == 'OpenGL ES'
  else // gl
endif
```

test/testGlEngine.h
```diff
@@ -29,6 +29,78 @@
 
 #include <thorvg.h>
 
+#if defined(THORVG_GL_TEST_CGL_SUPPORT)
```
hermet: `__APPLE__`  this is clean and simple.

test/testWgEngine.cpp
```diff
@@ -130,9 +130,10 @@ TEST_CASE("WG Basic draw", "[tvgWgEngine]")
                 if (maskOp != MaskMethod::None) REQUIRE(shape4->mask(mask(), maskOp) == Result::Success);
                 REQUIRE(canvas->add(shape4) == Result::Success);
             }
+            REQUIRE(canvas->draw(true) == Result::Success);
```
hermet, line 133: @wenjieshen Please review whether this is the right fix. Somehow it just hide the problem.

hermet (commented):
@Nor-s please check comments. thanks.

Nor-s:
@hermet  
Thank you for the review. I've updated the code.

test/testWgEngine.cpp
wenjieshen: I believe this is the correct solution for me. While it may conceal the issue, it also makes it easier to identify the problem when the test fails.

hermet: @wenjieshen what you do mean by "it also makes it easier to identify the problem when the test fails."?
 how this fix help to identify the problem?

hermet: @Nor-s @wenjieshen This is not about the test only problem. The root problem is that users have no way of knowing about this limitation when using ThorVG. We might wanna probably find a way to prevent the crash by querying the GPU capabilities and limiting GPU resource usage accordingly.

hermet (approved):

test/testWgEngine.cpp
wenjieshen: Yes,  the out-of-limitation is an issue and should be resolved in the engine. I thought this change was to identify the problematic mask operation when the test failed because I only considered the test itself.

PR #4550, opened 2026-07-14
lottie: invalidate the layer transform cache when a tween is turned off

theashraf:
While a dynamic tween is active the builder writes the interpolated matrix into the layer transform cache, but stamps it with the un-tweened frame number. Turning the tween off leaves that matrix in place: `updateTransform()` sees the cache is already on the requested frame and returns early, so the scene stays frozen on the tweened pose instead of falling back to the real frame.

It shows up when a tween is cancelled mid-flight. Calling `frame()` with the frame the animation is already on turns the tween off, but nothing rebuilds, so the last interpolated pose sticks.

Mark the tween dirty in `off()` and let one update pass through, so the transform is rebuilt from the actual frame.


hermet:
@theashraf thanks for your pr, could you please confirm the fix with this? https://github.com/thorvg/thorvg/pull/4559

PR #4561, opened 2026-07-15
gl: skip transparent shapes in hit tests

wenjieshen:
GL keeps the prepared geometry when opacity drops to zero. 
Rejecting transparent render data prevents invisible shapes from reporting hits and aligns the result with that of a shape starting with zero opacity.

### DIFF
--- a/src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
+++ b/src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
@@ -1265,11 +1265,18 @@ void GlRenderer::dispose(RenderData data)
 
 RenderData GlRenderer::prepare(RenderSurface* image, RenderData data, const Matrix& transform, const Array<RenderData>& clips, uint8_t opacity, FilterMethod filter, RenderUpdateFlag flags)
 {
-    if (opacity == 0) return data;
-
     //TODO: redefine GlImage?
     auto sdata = static_cast<GlShape*>(data);
     if (!sdata) sdata = new GlShape;
+
+    if (opacity == 0) {
+        sdata->opacity = 0;
+        sdata->deferredFlags |= flags;
+        return sdata;
+    }
+
+    flags |= sdata->deferredFlags;
+    sdata->deferredFlags = RenderUpdateFlag::None;
 
     auto cacheStale = sdata->texId && (sdata->texStamp != mTextures.stamp);
     if (flags == RenderUpdateFlag::None && !cacheStale) return data;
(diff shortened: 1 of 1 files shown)


hermet:
@wenjieshen please properly resolve the copilot comment. thanks.

src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
```diff
@@ -1394,6 +1394,7 @@ bool GlRenderer::intersectsShape(RenderData data, TVG_UNUSED const RenderRegion&
 {
     if (!data) return false;
     auto shape = (GlShape*)data;
+    if (shape->opacity == 0) return false;
     const auto& bbox = shape->geometry.getBounds();
```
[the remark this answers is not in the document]
wenjieshen: The image has the same patch now.

wenjieshen:
> @wenjieshen please properly resolve the copilot comment. thanks.

Hi @hermet, I added the same fix for images with zero opacity. Thank you.

src/renderer/gpu_engine/gl/tvgGlRenderer.cpp
```diff
@@ -1394,6 +1401,7 @@ bool GlRenderer::intersectsShape(RenderData data, TVG_UNUSED const RenderRegion&
 {
     if (!data) return false;
     auto shape = (GlShape*)data;
+    if (shape->opacity == 0) return false;
```
hermet, line 1404: @wenjieshen Could you please verify why wgpu works without this opacity check?

```

bool WgRenderer::intersectsShape(RenderData data, TVG_UNUSED const RenderRegion& region)
{
    if (!data) return false;
    auto shape = (WgRenderDataShape*)data;
    RenderRegion bbox = {
        {(int32_t)shape->aabb.min.x, (int32_t)shape->aabb.min.y},
        {(int32_t)shape->aabb.max.x, (int32_t)shape->aabb.max.y}
    };
    if (region.intersected(bbox)) {
        if (region.contained(bbox)) return true;
        WgIntersector intersector;
        return intersector.intersectShape(RenderRegion::intersect(region, bbox), shape);
    }
    return false;
}

```

wenjieshen: Hi @hermet,

WGPU avoids the original bug because it still prepares and refreshes visibility at opacity 0. GL defers preparation to save memory for permanently transparent paints and retains existing cached geometry, so hit testing requires an explicit opacity check.

Thank you.

PR #4562, opened 2026-07-15
lottie/slot: fix image slot type confusion in parseAsset

theashraf:
An image slot whose asset json has no width/height and e:0 was classified as audio by parseAsset()'s data heuristic, then blindly cast to LottieImage, reading garbage past the LottieAudio allocation (intermittent SIGSEGV in Paint::ref during slot json parsing).

Since the slot property type is already known to be Image, pass that to parseAsset() to force the image branch, and reject non-image results (e.g. a precomp asset with "layers") instead of casting them — the same guard LottieBuilder::updateImage already applies on the builder side.

Reproduce with an animation whose slotted image asset carries a path and no dimensions:

```json
{"v":"5.7.4","fr":30,"ip":0,"op":30,"w":100,"h":100,
 "assets":[{"id":"img_0","sid":"image_slot","p":"cat.png","u":"images/","e":0}],
 "layers":[{"ddd":0,"ind":1,"ty":2,"refId":"img_0","ks":{},"ip":0,"op":30,"st":0}]}
```

and override the slot:

```json
{"image_slot":{"p":{"p":"dog.png","u":"images/","e":0}}}
```

Unit tests pass (meson -Dtests=true -Dloaders=all -Dsavers=all -Dbindings=capi -Dtools=all).

### DIFF
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -1060,7 +1060,7 @@ void LottieParser::parseImage(LottieImage* image, const char* data, const char* 
 }
 
 
-LottieObject* LottieParser::parseAsset()
+LottieObject* LottieParser::parseAsset(bool image)
 {
     enterObject();
 
@@ -1094,7 +1094,7 @@ LottieObject* LottieParser::parseAsset()
         else skip();
     }
     if (data) {
-        if (!strncmp(data, "data:image/", 11) || width != 0.0f || height != 0.0f) {
+        if (image || !strncmp(data, "data:image/", 11) || width != 0.0f || height != 0.0f) {
             auto asset = new LottieImage;
             parseImage(asset, data, subPath, embedded, width, height);
             if (sid) registerSlot(asset, sid, asset->bitmap);
@@ -1757,10 +1757,13 @@ LottieProperty* LottieParser::parse(LottieSlot* slot)
         case LottieProperty::Type::Image: {
             LottieObject* obj = nullptr;
             while (auto key = nextObjectKey()) {
-                if (KEY_AS("p")) obj = parseAsset();
+                if (KEY_AS("p")) obj = parseAsset(true);
                 else skip();
             }
-            if (!obj) return nullptr;
+            if (!obj || obj->type != LottieObject::Type::Image) {
+                delete(obj);
+                return nullptr;
+            }
             prop = new LottieBitmap(static_cast<LottieImage*>(obj)->bitmap);
             delete(obj);
             break;
--- a/src/loaders/lottie/tvgLottieModel.h
+++ b/src/loaders/lottie/tvgLottieModel.h
@@ -949,6 +949,8 @@ struct LottieImage : LottieObject
     LottieBitmap bitmap;
     bool resolved = false;
 
+    LottieImage() { LottieObject::type = LottieObject::Image; }
+
     LottieProperty* override(LottieProperty* prop, bool release) override
     {
         LottieProperty* backup = nullptr;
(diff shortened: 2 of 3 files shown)


src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1094,7 +1094,7 @@ LottieObject* LottieParser::parseAsset()
         else skip();
     }
     if (data) {
-        if (!strncmp(data, "data:image/", 11) || width != 0.0f || height != 0.0f) {
+        if (image || !strncmp(data, "data:image/", 11) || width != 0.0f || height != 0.0f) {
```
tinyjin, line 1097: In the LottieSpec, Image should have size (w/h) while audio doesn't. Having the `image` parameter looks to define duplicated condition.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1757,10 +1757,13 @@ LottieProperty* LottieParser::parse(LottieSlot* slot)
...
             while (auto key = nextObjectKey()) {
-                if (KEY_AS("p")) obj = parseAsset();
+                if (KEY_AS("p")) obj = parseAsset(true);
                 else skip();
             }
-            if (!obj) return nullptr;
+            if (!obj || obj->type != LottieObject::Type::Image) {
+                delete(obj);
+                return nullptr;
+            }
```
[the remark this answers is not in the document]
tinyjin: Fundamentally, the Audio should not attempt overriding to Image

tinyjin (commented):
Thank you. Before we move onto patching. We'd need to verify if animation file doesn't violate spec. (_I think this applies to images without w/h information_)

Since ThorVG adheres to the Lottie specification, I believe it is not possible to write defensive code for every possible scenario.

Please let me know if there are something unavoidable in production side!

src/loaders/lottie/tvgLottieParser.cpp
theashraf: it's only duplicated when the file is spec-clean. the crash here is an image asset with no w/h getting classified as audio, then blind-cast to LottieImage. since we already know the slot is an image at this point, forcing it and rejecting a non-image result is more robust than trusting the w/h heuristic.

theashraf:
the w/h part is arguable but it's not the whole story. even a spec-valid file can drop an audio asset or a precomp (layers) into an image slot, and the old path cast that straight to LottieImage and crashed. also the slots spec never defines the override shape, so "no w/h in the override" isn't clearly out of spec. so it's not really avoidable on our side, the type check is there to not crash on a mismatched asset.

theashraf:
@tinyjin unrelated to this fix but worth flagging, i ran into another crash in the slot path while testing this. if a slot override's `p` object carries its own `sid`, parseAsset() registers that asset into comp->slots, but parse(LottieSlot*) deletes it right after, so the slot keeps a dangling obj and the next apply() hits a use-after-free. it's pre-existing on main, not from this change.

i'll send a separate follow-up PR for it (along with the null-deref copilot flagged for unhandled slot types, same area) so this one stays focused on the image type confusion. shout if you'd rather bundle them here.

tinyjin:
> @tinyjin unrelated to this fix but worth flagging, i ran into another crash in the slot path while testing this. if a slot override's `p` object carries its own `sid`, parseAsset() registers that asset into comp->slots, but parse(LottieSlot*) deletes it right after, so the slot keeps a dangling obj and the next apply() hits a use-after-free. it's pre-existing on main, not from this change.
> 
> i'll send a separate follow-up PR for it (along with the null-deref copilot flagged for unhandled slot types, same area) so this one stays focused on the image type confusion. shout if you'd rather bundle them here.

Thanks for finding. Yes you can do that, or file an issue for new one!

PR #4563, opened 2026-07-15
capi: support unsetting the asset resolver with a null callback

theashraf:
tvg_picture_set_asset_resolver() wrapped a null callback in a valid lambda, so passing NULL installed a resolver that dereferences NULL once an asset is resolved, instead of clearing it as the API doc states ("To unset the resolver, pass nullptr").

Forward the null callback to Picture::resolver(), which already supports clearing, matching what tvg_lottie_animation_set_audio_resolver() does. As with setting, clearing after the picture is loaded returns InsufficientCondition.


hermet (approved):

hermet:
@theashraf thanks!

PR #4564, opened 2026-07-15
loader: fix a font sharing count leak on cached data loading

theashraf:
LoaderMgr::loader(name, data, ...) returns early when the font is already cached, but font() increments the sharing count on that hit and the caller (Text::load) discards the result, so the count was never balanced and the loader leaked at unload. Loading the same font data twice (e.g. a lottie with an embedded font, loaded twice) strands the loader past Initializer::term().

Apply the same compensation Text::load() already performs for the file-path route (extends d50632d0 to the data path).


hermet:
replaced with: https://github.com/thorvg/thorvg/pull/4597

PR #4572, opened 2026-07-16
renderer: guard mutex behind flag

BenjaminHalko:
ThorVG guards its task scheduler's locking behind `THORVG_THREAD_SUPPORT` (`tvgTaskScheduler.h`), but `tvgSwRenderer.cpp`'s `_rendererMtx` misses that guard, so the file does not compile against a single-threaded libc++ (`wasm32-wasip2`, where `std::mutex` does not exist).

### DIFF
--- a/src/renderer/cpu_engine/tvgSwRenderer.cpp
+++ b/src/renderer/cpu_engine/tvgSwRenderer.cpp
@@ -34,7 +34,11 @@
 /************************************************************************/
 
 static int32_t _rendererCnt = -1;
+#ifdef __STDCPP_THREADS__
 static mutex _rendererMtx;
+#else
+static struct { void lock() {} void unlock() {} } _rendererMtx;
+#endif
 
 struct SwTask : Task
 {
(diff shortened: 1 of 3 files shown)


hermet (changes requested):
@BenjaminHalko Hello, firstly, I'm rejecting this,  `rendererMtx` is intended for the multi-threaded environment where users run ThorVG on worker threads. It isn't quite related to ThorVG's internal nested threading support (THORVG_THREAD_SUPPORT). Maybe this might be a solution instead: 
1. `__STDCPP_THREADS__`  // preffered
2. `#if defined(__wasm__) && defined(__wasi__)`

\+ It would be nice if you could share the error message and confirm the fix with above approaches. Thanks.

BenjaminHalko:
This was the error message:
```
cargo:warning=deps/thorvg\src/renderer/sw_engine\tvgSwRenderer.cpp:37:8: error: unknown type name 'mutex'
cargo:warning=   37 | static mutex _rendererMtx;
```

I'll see about changing the if condition

hermet (approved):

hermet:
@BenjaminHalko thanks!

PR #4575, opened 2026-07-18
common/lock: mutex portability++

hermet:
Still there were mutex usages in sw engine (mempool) this revised the tvg Lock for mutex portability

- StrictLock enables the mutex only if __STDCPP_THREADS__
- Replaced invidual mutex usage with StrictLock in engines


hermet:
\+ @BenjaminHalko

BenjaminHalko:
Thanks @hermet ! 🙏

PR #4571, opened 2026-07-16
lottie/text: fix point text vertical alignment

hoyangtsai:
Follow-up to #4462 (which fixed **box** text) — this fixes the **point** text case of #4334.

## Problem

For point text (no `sz`), `updateURLFont` places the first baseline ~one ascent plus a valign-slack shift too high:

1. the sfnt loader bakes `m[5] = hhea.ascent * scale` into every glyph (`tvgSfntLoader.cpp`), so the first baseline sits **one ascent below the drawable top** — correct for box text (its cap-top meets the box top), wrong for point text;
2. point text also still gets `valign = ascent / (ascent − descent)`.

Net first baseline = `origin + ascent − blockHeight·ascentRatio` instead of `origin`.

## Fix

For point text (`bbox.size.y <= 0`), cancel the loader's ascent offset (translate up by `ascent`) and use `valign 0`, so the first baseline lands on the layer origin — matching lottie-web / After Effects. **Box text is unchanged.**

## Verification

Repro: point text, 4 lines, left-justified, URL/data-URI font (`chars` empty); `fontSize 86`, `lh 75.2`; rendered at 512×512. First-baseline Y (scene coords), with **lottie-web 5.12.2** as the reference renderer:

| line | lottie-web (reference) | ThorVG v1.0.7 (before) | ThorVG patched (after) |
|------|------------------------|------------------------|------------------------|
| 1 | 141 | ~52 *(clipped above top)* | **141.1** |
| 2 | 218 | ~127 | **216.8** |
| 3 | 293 | ~202 | **291.7** |
| 4 | 368 | *(hidden)* | **367.4** |

Patched ThorVG lands within ~1px of lottie-web on every line; the first line is no longer clipped. Screenshot below — left is the editor canvas, right is the dotlottie-web preview built with this patch; the two are now in parity:

point-text fix parity

A prebuilt dotlottie-web wasm carrying this patch, for quick in-browser testing: [dotlottie-player-thorvg-pointtext-patched.wasm](https://raw.githubusercontent.com/hoyangtsai/thorvg/pr-assets-4334/assets/dotlottie-player-thorvg-pointtext-patched.wasm).

## Notes

- Verified on one scene; more justify / anchor / line-count cases are worth a check.
- This is the `updateURLFont` (URL / data-URI font, `chars` empty) path — the embedded-`chars` glyph path likely needs the same treatment.

PR #4462, opened 2026-06-14
lottie/text: fix box text vertical alignment

tinyjin:
In the URL font, baseline-align is always applied to both
- point text (without "sz")
- box text (when "sz" sets)

With the box text, vertical alignment should anchor from the top, not baseline.

issue: https://github.com/thorvg/thorvg/issues/4334

test file : [font_box_auto_resize_sample.json](https://github.com/user-attachments/files/28924992/font_box_auto_resize_sample.json)

| Before | After | Expectation |
|---------|---------|---------|
| CleanShot 2026-06-14 at 17 52 15@2x | CleanShot 2026-06-14 at 17 51 05@2x | image |

Issue #4334, opened 2026-04-13
lottie: url font text dismatch in y-alignment


tinyjin:
[right_center.json](https://github.com/user-attachments/files/26673538/right_center.json)

> Note: this isn't regression. Found new case while I test various cases with url font rendering

| Expectation | ThorVG |
|--------|--------|
| <img width="1022" height="1016" alt="Image" src="https://github.com/user-attachments/assets/5005948a-9f68-452d-a273-06c74b982f39" />  | <img width="1596" height="1490" alt="Image" src="https://github.com/user-attachments/assets/de9f5897-6132-49f6-b635-885679e1aec6" /> |

tinyjin:
+ Additional Case

<img width="1666" height="838" alt="Image" src="https://github.com/user-attachments/assets/b5ec9cb0-d051-4a1b-8426-baffbc8424db" />

[font_box_auto_resize_sample.json](https://github.com/user-attachments/files/28924553/font_box_auto_resize_sample.json)

tinyjin:
This asset has three issues
1. incorrect box text vertical alignment
2. incorrect text layout caused by ignored `ascent` (_when lottie ascent !== thorvg font engine ascent_)
3. should ignore ascent accumulation when box position is absent ("ps")

<img width="1708" height="1048" alt="Image" src="https://github.com/user-attachments/assets/caf82675-b4a1-4a20-9514-4335647f9b40" />

hoyangtsai:
Hey @tinyjin, just a quick check. Is this issue related to #4462? I saw the PR was merged to main. When would it be released?

tinyjin:
@hoyangtsai Hello, the PR 4462 is already released in v1.0.7, thanks.

tinyjin:
+ point text case in url text path:

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/1577a6cf-a898-484f-88f5-2781b00af4ce" />

sample by @hoyangtsai 
[point text.json](https://github.com/user-attachments/files/30163240/point.text.json)

PR #4579, opened 2026-07-20
lottie/text: fix point text vertical alignment (backport #4571 to v1.0.x)

hoyangtsai:
Backport of #4571 to the `v1.0.x` release branch.

#4571 fixed the **point text** case of #4334 on `main`. This cherry-picks it onto `v1.0.x` so it can be consumed without waiting for the next release tag.

## Why the release branch

Downstream runtimes pin `v1.0.x` lineage rather than `main`. [`dotlottie-rs`](https://github.com/LottieFiles/dotlottie-rs) currently pins v1.0.6, and every one of its historical bumps has tracked a `v1.0.x` tag. Picking up the fix from `main` would mean pulling 314 commits into the runtime behind every dotLottie player; this backport is 3 lines on top of a branch that already carries v1.0.7.

## The change

Identical in substance to #4571. For point text (`bbox.size.y <= 0`), cancel the sfnt loader's baked-in `hhea.ascent` offset by translating up one ascent, and use `valign 0` — so the first baseline lands on the layer origin, matching lottie-web and After Effects.

Note this supersedes the conditional `valign` introduced for box text in #4462 (already on this branch via v1.0.7): the `size.y > 0` box path still resolves to `valign 0` with an unchanged y translate, so **box text behaviour is untouched**.

## Verification

- Cherry-picks cleanly onto `v1.0.x` (auto-merge, 3 insertions / 3 deletions, one file). The pre-image here is already the #4462 form, so the two fixes compose as intended.
- `src/loaders/lottie/tvgLottieBuilder.cpp` compiles clean on this branch.
- Rendering behaviour was verified live on `main` in #4571 against lottie-web 5.12.2 as reference: first-baseline Y within ~1px on every line (141/218/293/368), top line no longer clipped.

Scope is unchanged from #4571 — the `updateURLFont` (URL / data-URI font, empty `chars`) path. The embedded-`chars` glyph path likely needs the same treatment, and #4334 stays open for its remaining sub-cases.

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -1034,14 +1034,14 @@ void LottieBuilder::updateURLFont(LottieLayer* layer, float frameNo, LottieText*
     if (text->font && text->font->style && strstr(text->font->style, "Italic")) paint->italic();
     paint->text(buf);
     paint->layout(doc.bbox.size.x, doc.bbox.size.y);
-    paint->translate(doc.bbox.pos.x, doc.bbox.pos.y);
     if (doc.bbox.size.x > 0.0f) paint->wrap(TextWrap::Smart);
 
     //align the text to the base line, or top within the box
     TextMetrics metrics;
     paint->metrics(metrics);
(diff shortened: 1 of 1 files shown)


hermet:
v1.0.x is not contribution branch target.

PR #4595, opened 2026-07-23
media: add experimental Windows video playback support

Nor-s:
Implement the Windows media loader using the Media Foundation
IMFMediaEngine for video open, playback, seek, loop, volume,
and mute controls.

The engine drives decoding, A/V clocking, and audio output
internally, so no separate audio pipeline is required. sync()
polls decoded frames in frame-server mode and copies them into
the render surface through a D3D11 staging texture, with a WARP
fallback for machines without a usable GPU.

Wire the Windows sources and required system libraries into Meson.


hermet:
@Nor-s please rebase and iron out. thanks.

Issue #4551, opened 2026-07-14
lottie: compliance problem


hermet:
Either one is wrong. Need to confirm if thorvg is problematic.

thorvg:

<img width="830" height="818" alt="Image" src="https://github.com/user-attachments/assets/0f05a9f3-8b51-4f38-9dfb-edb6530a3e9e" />

skottie:

<img width="1303" height="1272" alt="Image" src="https://github.com/user-attachments/assets/7264974b-5bb9-44ca-9099-3cb5482777dc" />

[8b0807a480d0.json](https://github.com/user-attachments/files/29995482/8b0807a480d0.json)

hermet:
invalid asset

PR #4596, opened 2026-07-24
lottie: clamp opacity/color to the valid range on parsing

tinyjin:
Opacity (0 ~ 100) and color channels (0 ~ 1)
are stored as uint8, but the Lottie spec defines
them as unbounded numbers.

Clamp inside REMAP255 so every value narrowed
to an 8-bit channel.

issue: https://github.com/thorvg/thorvg/issues/4551

test file: [opacity.json](https://github.com/user-attachments/files/30333776/opacity.json)

| Before | After |
|---------|---------|
| CleanShot 2026-07-24 at 14 43 03@2x | CleanShot 2026-07-24 at 14 42 46@2x |

### DIFF
--- a/src/loaders/lottie/tvgLottieCommon.h
+++ b/src/loaders/lottie/tvgLottieCommon.h
@@ -98,7 +98,7 @@ struct TextDocument
 
 static inline int32_t REMAP255(float val)
 {
-    return (int32_t)nearbyintf(val * 255.0f);
+    return (int32_t)nearbyintf(tvg::clamp(val, 0.0f, 1.0f) * 255.0f);
 }
 
 
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -272,11 +272,11 @@ bool LottieParser::getValue(uint8_t& val)
 {
     if (peekType() == kArrayType) {
         enterArray();
-        if (nextArrayValue()) val = (uint8_t)(getFloat() * 2.55f);
+        if (nextArrayValue()) val = (uint8_t)REMAP255(getFloat() * 0.01f);
         //discard rest
         while (nextArrayValue()) getFloat();
     } else {
-        val = (uint8_t)(getFloat() * 2.55f);
+        val = (uint8_t)REMAP255(getFloat() * 0.01f);
     }
     return false;
 }


tinyjin:
@Nor-s This may require updating much of golden test assets.
The fix includes fine tones correctness. (Previously softer or more intense)

Nor-s:
> @Nor-s This may require updating much of golden test assets. The fix includes fine tones correctness. (Previously softer or more intense)

@tinyjin 
Thank you for sharing.
Currently, the goldens are generated as snapshot versions in the CI environment. 
So, we just need to update the snapshots.

Nor-s:
@hermet 
Once this PR is merged, would it be okay to update the snapshots based on that commit?
Thank you.

hermet:
> Lottie spec defines

@tinyjin could you please share it?

tinyjin:
@hermet You can refer to these:
- https://lottie.github.io/lottie-spec/latest/specs/schema/#/$defs/shapes/shape-style
- https://lottiefiles.github.io/lottie-docs/schema/#/$defs/shapes/shape-style

CleanShot 2026-07-24 at 17 31 54@2x

For clearity, the issue sample has problem with opacity value. (Not color) And the opacity doesn't have scheme level number boundary.

The color value, the scheme limits it into [0-1], however without clamp it makes rendering differences from other players(lottie-web, skia), because they practically clamp it.

You can check out this:
- [color_test_1.json](https://github.com/user-attachments/files/30339514/color_test_1.json)
- [color_test_2.json](https://github.com/user-attachments/files/30339516/color_test_2.json)

hermet:
@tinyjin Thanks. I would consider this is not a ThorVG issue today, but the result itself is undefined. Here are a few considerations in my mind.

- We have no evidence that users intentionally set values greater than 100 (or 1) in their assets when 100 is the defined maximum. What authoring tools allow this and why?
- Some implementations, such as Skia and web browsers, may handle these out-of-range values gracefully, but that behavior is not clearly defined or required by the specification.
- If we decide to support out-of-range values, we might wanna reconsider the entire value pipeline for example, using a wider value type than 8 bits and removing clamping throughout the lottie process including interpolation.

Let's carefully see this.

tinyjin:
@hermet I understand, yes let see!

tinyjin:
Maybe we can suggest this to add spec to LAC?

hermet:
> Maybe we can suggest this to add spec to LAC?

Exception handling for unexpected usage may be better left to individual implementations, since it does not address the root cause.

PR #4597, opened 2026-07-24
loader: fix a font sharing count leak on cached data loading

hermet:
LoaderMgr::loader(name, data, ...) returns early
when the font is already cached, but font() increments the sharing count on that hit and the caller (Text::load) discards the result, so the count was never balanced and the loader leaked at unload. Loading the same font data twice (e.g. a lottie with an embedded font, loaded twice) strands the loader past Initializer::term().

Now, properly account for shared fonts by forcefully clearing the font cache during termination.


hermet:
@theashraf hi, could you please confirm your fix with this?

Issue #4520, opened 2026-07-02
lottie: rendering discrepancies


theashraf:
<img width="600" height="400" alt="Image" src="https://github.com/user-attachments/assets/1a8ee860-d9a7-4de5-97b9-32d8f7e414fd" />

[template 2.json](https://github.com/user-attachments/files/29583635/template.2.json)

hermet:
```
auto inner = tvg::Scene::gen();
auto shape = tvg::Shape::gen();
shape->appendRect(20, 50, 100, 100);
shape->fill(255, 0, 0);
inner->add(shape);

auto scene = tvg::Scene::gen();
scene->add(inner);
scene->opacity(128);

auto parent = tvg::Scene::gen();
parent->add(scene);
parent->add(tvg::SceneEffect::GaussianBlur, 3.0, 0, 0, 100);
canvas->add(parent);
```

PR #4598, opened 2026-07-24
renderer: fix scene opacity loss in post-processing composition

tinyjin:
In a direct composition path, it doesn't
examine the child's own composition purpose.

Therefore, children requiring Opacity
composition also skipped.

related: https://github.com/thorvg/thorvg/issues/4520

---

```cpp
auto inner = tvg::Scene::gen();
shape->appendRect(20, 50, 100, 100);
shape->fill(255, 0, 0);
inner->add(shape);

scene->add(inner);
scene->opacity(128);

auto parent = tvg::Scene::gen();
parent->add(scene);
parent->add(tvg::SceneEffect::GaussianBlur, 3.0, 0, 0, 100);
canvas->add(parent);
```

- Expected : 50% red
- Actual : 100% red (opacity ignored)

| Before | After |
|---------|---------|
| CleanShot 2026-07-24 at 20 23 50@2x | CleanShot 2026-07-24 at 20 24 10@2x |

### DIFF
--- a/src/renderer/tvgScene.h
+++ b/src/renderer/tvgScene.h
@@ -133,7 +133,7 @@ struct SceneImpl : Scene
 
         RenderCompositor* cmp = nullptr;
         // its parent is already in composition mode, maybe parasitize its surface
-        auto incomposite = (uint8_t(CompositionFlag::PostProcessing) & uint8_t(flag)) && !effects;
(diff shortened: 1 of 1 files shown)


hermet (dismissed):
Here Incomposition is not for blending mode.

hermet:
@tinyjin Please leave this. This needs a considerable solution. Thanks for trying to fix.

PR #4602, opened 2026-07-26
gl: reduce redundant per-draw setup

### LINKED ISSUE #4555: gl: reduce redundant resource binding and state setup per render task
## What / Why

The GL backend performs repeated resource lookup, state query, binding, and vertex-layout setup for every `GlRenderTask`.

The current `GlRenderTask::run()` path includes:

* Looking up `uDepth` and `uViewMatrix`
* Calling `glGetIntegerv(GL_ARRAY_BUFFER_BINDING, ...)`
* Binding array buffers
* Enabling vertex attributes
* Calling `glVertexAttribPointer()`
* Binding textures and updating sampler uniforms
* Calling `glUniformBlockBinding()` for each UBO resource
* Calling `glBindBufferRange()` for each UBO resource
* Disabling vertex attributes after the draw
* Restoring the previous array buffer

`GlProgram::load()` already avoids redundant `glUseProgram()` calls, but uniform locations and uniform-block indices are still queried through GL instead of being cached by the program.

This issue does not assume that these calls are the current performance bottleneck. The goal is to remove clearly redundant hot-path work and measure whether it improves CPU submission time, particularly for image and gradient workloads that bind more resources per task.

## Goals

* Reduce redundant GL API calls in `GlRenderTask::run()`
* Avoid driver queries in the per-draw hot path
* Cache immutable program resource locations
* Skip state changes when the requested state is already active
* Preserve rendering correctness across GL, GLES, and WebGL2 targets
* Define a clear GL-state ownership boundary for ThorVG rendering

### 1. Cache program resource locations

Cache uniform locations and uniform-block indices per `GlProgram`.

Candidates include:

* `uDepth`
* `uViewMatrix`
* Texture sampler uniforms
* `TransformInfo`
* `GradientInfo`
* `ColorInfo`
* `BlendRegion`
* Other program-specific uniform blocks

The cache may be populated during program initialization or lazily on first lookup.

### 2. Assign uniform-block bindings once

Uniform-block-to-binding-point mappings are properties of a linked program and generally do not need to be assigned for every draw.

Move `glUniformBlockBinding()` out of `GlRenderTask::run()` and assign fixed binding points when the program is initialized.

The draw path should only update the buffer range when necessary.

### 3. Introduce a renderer-owned GL state cache

Track at least:

* Current program
* Array buffer
* Element array buffer
* Enabled vertex attributes
* Vertex attribute layout
* Active texture unit
* Texture bound to each unit
* UBO buffer, offset, and range for each binding point
* Scissor rectangle
* Framebuffer where appropriate
* Blend, depth, color-mask, and stencil state where appropriate

Skip GL calls when the cached state already matches the requested state.

### 4. Remove per-task state restoration

Avoid querying and restoring `GL_ARRAY_BUFFER_BINDING` for every render task.

If ThorVG must preserve the caller's GL state, capture the required external state once before executing the render-task tree and restore it once after rendering. Do not perform this preservation separately for every task.

### 5. Handle direct state mutations

Composition, stencil, clipping, blending, and effect tasks currently issue direct GL state changes.

The state cache must either:

* Route these changes through the cache
* Explicitly invalidate affected cache entries

Special handling is also required for constant vertex attributes such as `mUseVertexColor`.

## Acceptance Criteria

* No `glGetIntegerv()` call remains in the per-task draw path
* Uniform and uniform-block locations are queried at most once per program
* `glUniformBlockBinding()` is not called for every draw
* Redundant buffer, texture, and vertex-attribute operations are skipped
* Rendering output remains unchanged
* GL, GLES, and WebGL2 builds remain functional
* CPU render-submission time is measured before and after
* No significant regression is introduced in GPU time or total frame time

wenjieshen:
## Goal

GL tasks repeat resource queries and state changes on each draw, even
when the requested state is already active. That adds avoidable
CPU-side submission work, especially for images and gradients
that bind more resources.

The goal is for repeated draws to pay only for state that actually
changes, without altering rendered output.

## Result

Image benchmark throughput nearly doubles. Linear-gradient throughput
improves by about 2.1x, and radial-gradient throughput by about 1.7x.
Circle and rectangle throughput remain effectively neutral.

Pixel comparison reports no rendering differences.

### Benchmarks

| Benchmark | Default | Rotation |
|---|---:|---:|
| Circle | 1.0073× | 1.0074× |
| Image | 1.9879× | 1.9923× |
| LinearGradient | 2.1686× | 2.1035× |
| RadialGradient | 1.6914× | 1.6885× |
| Rect | 0.9983× | 0.9987× |
| Stroke | 1.1635× | 1.2044× |
| StrokeRect | 1.1664× | 1.1810× |

### ThorVG.examples
All confidence intervals include zero, so none is a statistically reliable change.
### Janitor
The experience feels more consistent when the explode effect happens.
## Approach

- Reuse one renderer-owned GL state cache across the complete
  sync task tree.
- Cache immutable program lookups, fixed block bindings, and sampler
  assignments.
- Make vertex layouts identify their source VBO.
- Skip redundant buffer and attribute setup when the layout is already
  active.
- Invalidate cached assumptions at renderer ownership boundaries while
  preserving the existing cleanup behavior.

## GL state ownership

ThorVG does not preserve caller GL state across `sync()`. Applications
sharing the context must rebind their own state afterward.

issue: #4555


src/renderer/gpu_engine/gl/tvgGlStateCache.h
```diff
@@ -0,0 +1,78 @@
...
+    void stencilFuncSeparate(GLenum face, GLenum function, GLint reference, GLuint mask);
+    void stencilOp(GLenum stencilFail, GLenum depthFail, GLenum depthPass);
+    void stencilOpSeparate(GLenum face, GLenum stencilFail, GLenum depthFail, GLenum depthPass);
+
+    void clearColor(GLfloat red, GLfloat green, GLfloat blue, GLfloat alpha);
+    void clearStencil(GLint value);
+    void clearDepth(double value);
+
+private:
+    struct Impl;
```
hermet: please dont' use pimpl pattern for intenal dev. it's over engineering.

hermet (changes requested):

src/renderer/gpu_engine/gl/tvgGlStateCache.h
wenjieshen: Hi @hermet, sorry for the late notification. I have improved the code and reduced the binary size. Thank you. 🙏

hermet:
IntelCore Ultra5 / Intel Arc Graphics:

| Test | Rate |
|---|---:|
| Lottie | +1.24% |
| BigDonut | -6.19% |
| Flying | -0.25% |
| ManyStar | -2.87% |
| Samurai | +3.66% |

- None of them show any meaningful improvement.
- It's hard to say that Janitor is faster than before.

hermet:
M2 Pro:

Confirmed the gradient enhancements.
| Test | Rate |
|---|---:|
| Linear | +105.3% |
| Radial | +57.3% |
| Image | +77.9% |
| Rect | +1.5% |
| Circle | 0.0% |
| Stroke | +17.4% |
| StrokeRect | +7.9% |

The following tests are within the margin of error.
| Test | Rate |
|---|---:|
| Lottie | -0.5% |
| BigDonut | +7.4% |
| Flying | -4.0% |
| ManyStar | +1.4% |
| Samurai | -16.4% |

hermet:
@wenjieshen LGTM but this needs a practical example to demonstrate the performance benefit. Please come up with one. Thanks.

wenjieshen:
[linear_gradient_wave_2500_aurora.json](https://github.com/user-attachments/files/31176736/linear_gradient_wave_2500_aurora.json)
This PR improves the FPS of the animation example that plays the above file by approximately 2.05× (+104.7%) over parent e7f39407.

The Lottie Creator gains ~50% improvement while playing this animation.

Setup:
Existing Animation example, unchanged
linear_gradient_wave_2500_aurora.json
OpenGL, release builds, 1024×1024, Apple M1
Four alternating 60-second runs per revision
VSync disabled via SDL_GL_SetSwapInterval(0); observed FPS exceeded 1000, confirming it was uncapped
Median steady-tail result was similar: 2.06×

hermet:
> [linear_gradient_wave_2500_aurora.json](https://github.com/user-attachments/files/31176736/linear_gradient_wave_2500_aurora.json) This PR improves the FPS of the animation example that plays the above file by approximately 2.05× (+104.7%) over parent [e7f3940](https://github.com/thorvg/thorvg/commit/e7f39407e3db4608914c43b397e7c8bc867b12dc).
> 
> The Lottie Creator gains ~50% improvement while playing this animation.
> 
> Setup: Existing Animation example, unchanged linear_gradient_wave_2500_aurora.json OpenGL, release builds, 1024×1024, Apple M1 Four alternating 60-second runs per revision VSync disabled via SDL_GL_SetSwapInterval(0); observed FPS exceeded 1000, confirming it was uncapped Median steady-tail result was similar: 2.06×

@tinyjin please confirm on the web end. thanks.

hermet (approved):
+67.27% (on windows ryzen 7 5800x / rtx 4070)

tinyjin (approved):

PR #4607, opened 2026-07-27
cpu_engine: fix a crash in stroke bounds of an unprepared shape

hermet:
shapeStrokeBBox() enters the stroking path on the authored stroke width alone and dereferences shape.stroke, but the raster pass may have left it null: a zero-opacity paint's task is marked ready without running (prepareCommon), and a zero-alpha solid stroke is discarded by shapeDelStroke (validStrokeWidth). Querying such a shape's bounds (e.g. Paint::bounds / tvg_paint_get_aabb) then crashed on the null stroke.

Allocate the stroke lazily before resetting it, matching shapeResetStroke().


BenjaminHalko:
@hermet Thanks! LGTM!

PR #4609, opened 2026-07-27
svg/parser: Prevent a crash in font-face declarations

### LINKED ISSUE #4605: SVG: nullptr crash parsing a css @font-face with no trailing ;
Hey,

I was going through the [W3 SVG 1.1 test suite](https://www.w3.org/Graphics/SVG/Test/20110816/archives/W3C_SVG_11_TestSuite.tar.gz) and tried out some of the svg > fonts-desc-* examples. That lead to a crash (ThorVG 1.0.7)

This is a minimal reproducible example

```xml
<svg xmlns="http://www.w3.org/2000/svg" width="160" height="90">
  <style>@font-face{font-family:Ext;src:url('missing.svg#Font')}</style>
  <text x="80" y="45" font-family="Ext" font-size="20">Hello</text>
</svg>
```

In `xmlParseW3CAttribute()`, when the src declaration has no trailing `;` and none follows later in the document, `strchr(buf, ';')` returns null and the font-face special case dereferences it:

```cpp
if (auto src = strstr(buf, "src")) {
    if (src < sep) {
        if (next + 1 < end) next = (char*)strchr(next + 1, ';');  // next is null -> strchr(0x1)
```

`if (next && src < sep)` fixes it.

jeongsunyong:
## Problem

Loading an SVG containing an `font-face` `src` descriptor without a trailing semicolon caused an access violation.

When no semicolon followed the declaration, `strchr()` returned `nullptr`. The font-face-specific path incremented that pointer before searching for another semicolon.

Fixes #4605

## Solution

Check that the initial semicolon exists before searching for the next one.

## Test Resource

The manual verification uses the minimal SVG reported in #4605.

For visual before-and-after verification, a white background, text alignment, and text color were added for visibility. The `font-face` declaration that triggers the issue was unchanged.

```xml
<svg xmlns="http://www.w3.org/2000/svg" width="160" height="90">
  <style>
    @font-face {
      font-family: Ext;
      src: url('missing.svg#Font')
    }
  </style>
  <rect width="160" height="90" fill="white"/>
  <text x="80" y="45"
        text-anchor="middle"
        font-family="Ext"
        font-size="20"
        fill="black">Hello</text>
</svg>
```

The missing external font is intentional. The SVG should load without crashing and may use the fallback font.

### Manual Reproduction

- Before the fix: access violation (`0xC0000005`)
- After the fix: loaded without crashing

### Picture Tests

- 8 test cases passed
- 57 assertions passed

### Full Unit Tests

Tested on Windows with MSVC.

- Ok: 1
- Fail: 0

### GitHub CI

- 23 checks passed

### thorvg.example

Manually verified the same SVG using the `PictureSvg` example.

- Before the fix: access violation (`0xC0000005`)
- After the fix: loaded and rendered using the fallback font

### Before Fix

The `PictureSvg` example terminates with an access violation.

image

### After Fix

The same SVG is loaded and rendered without crashing.

image

### Unit Tests

The full Windows unit test suite passes without failures.

image

### DIFF
--- a/src/loaders/svg/tvgXmlParser.cpp
+++ b/src/loaders/svg/tvgXmlParser.cpp
@@ -438,7 +438,7 @@ bool xmlParseW3CAttribute(const char* buf, unsigned bufLength, xmlAttributeCb fu
         auto next = (char*)strchr(buf, ';');
 
         if (auto src = strstr(buf, "src")) {//src tag from css font-face contains extra semicolon
-            if (src < sep) {
+            if (next && src < sep) {
                 if (next + 1 < end) next = (char*)strchr(next + 1, ';');
                 else break;
             }


Nor-s:
image

@jeongsunyong 

another github user is being mentioned in the commit message.
I think need to fix this.

test/testPicture.cpp
```diff
@@ -249,6 +249,17 @@ TEST_CASE("Load SVG Data", "[tvgPicture]")
     Paint::rel(picture);
 }
 
+TEST_CASE("Load SVG font-face without trailing semicolon", "[tvgPicture]")
```
hermet: Please remove a test case.

jeongsunyong: Thanks. I’ll remove the test case as requested.

jeongsunyong:
> image
> @jeongsunyong
> 
> another github user is being mentioned in the commit message. I think need to fix this.

Thanks. GitHub interpreted `@font-face` as a user mention.
I’ll remove the @ from the commit message.

src/loaders/svg/tvgXmlParser.cpp
```diff
@@ -438,7 +438,7 @@ bool xmlParseW3CAttribute(const char* buf, unsigned bufLength, xmlAttributeCb fu
         auto next = (char*)strchr(buf, ';');
 
         if (auto src = strstr(buf, "src")) {//src tag from css font-face contains extra semicolon
-            if (src < sep) {
+            if (next && src < sep) {
```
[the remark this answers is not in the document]
jeongsunyong: @Nor-s I understand that Copilot’s comment is about a different CSS
case where `:` is missing and `sep` becomes null.

This issue already existed before my change and is different from
#4605. Should I include the additional null check in this PR, or
should I test the `sep` case separately and open a new issue if I
confirm that it causes an actual problem?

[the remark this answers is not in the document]
Nor-s: @jeongsunyong 

If it is separate from this issue, you can resolve it(close copilot's comment).
If the Copilot’s comment can be reproduced, please open a new issue or new PR.

Thank you.

JSUYA, line 441: Thanks for the fix. LGTM.
This modification is the same as what was mentioned in the issue and resolves the problem.

The current svg loader specification does not cover all the various use cases related to font-face. Therefore, the current modification is appropriate.

JSUYA (approved):

src/loaders/svg/tvgXmlParser.cpp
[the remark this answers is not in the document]
jeongsunyong: Thank you. I’ll keep this PR focused on #4605 and verify the `sep`
case separately.

jeongsunyong: Thank you for the review and clarification.

hermet:
@jeongsunyong thanks, could you please add yourself in [CONTRIBUTORS](https://github.com/thorvg/thorvg/blob/main/CONTRIBUTORS.md) ?

PR #4612, opened 2026-07-28
engines: draw a cap dot for degenerate closed subpaths (M x y Z)

### LINKED ISSUE #3251: gpu: handling very short closed paths
right now Thorvg handles `M 100 100 L 100 100` but not `M 100 100 Z` which should be handled in the same way:

```
<svg viewBox="0 0 400 400" xmlns="http://www.w3.org/2000/svg">
    <path d="M 100 100 z " fill="none" stroke="red" stroke-width="40" stroke-linecap="round"/>
</svg>
```

nothing is drawn, expected a red dot

mnchnk:
### Summary

A subpath that consists only of moveTo() followed by close() (e.g. M 100 100 Z) has zero length, but was not rendering a round/square stroke cap as a dot on any engine — unlike an explicit zero-length lineTo (e.g. M 100 100 L 100 100), which cpu_engine already handles correctly (#3066).

This PR makes all three engines (cpu / gl / wg) consistently draw a cap dot for this degenerate case, matching the reference behavior seen in Adobe After Effects and lottie-web.

Fixes #3251

### Root cause

**cpu_engine** (tvgSwStroke.cpp)
_endSubPath() only draws a cap dot (_addCap()) in its "open subpath" branch. A moveTo + close() subpath is technically closed (closedSubPath == true), so it always took the "closed subpath" branch instead, which only handles corner joins — and a degenerate zero-length subpath has no real corner to join, so nothing was drawn.

**gl_engine / wg_engine** (tvgGlTessellator.cpp / tvgWgTessellator.cpp)
Stroker::close() / WgStroker::close() only ever called join(), never cap(). A degenerate subpath (nothing drawn between moveTo and close) needs a cap dot instead of a join, but the code path to draw one (cap()) was never reached from close().

### Changes
cpu_engine: _endSubPath() now treats a subpath as "open" (taking the cap-drawing branch) whenever it's degenerate (subPathLength == 0 and center == subPathStart), regardless of whether close() was actually called — since a zero-length path has no meaningful direction to join.
gl_engine / wg_engine: close() now calls cap() instead of join() when nothing was actually drawn in the subpath (firstPtDir is still (0,0)).
**Verification**
cpu: confirmed locally — the repro from #3251 now renders the dot with the default (software) backend.
gl: confirmed locally — the repro now renders the dot with the gl backend.
wg: WgStroker is structurally identical to gl_engine's Stroker (same fix applied 1:1), but I wasn't able to verify this locally due to the WebGPU (wgpu-native) setup on Windows/MSYS2. Would appreciate a check from someone who can run the wg backend, or I'm happy to investigate further if needed.

Repro used
`cpp
shape->moveTo(200, 200);
shape->close();
shape->strokeFill(255, 0, 0);
shape->strokeCap(tvg::StrokeCap::Round);
shape->strokeWidth(10);
canvas->add(shape);`

### DIFF
--- a/src/renderer/cpu_engine/tvgSwStroke.cpp
+++ b/src/renderer/cpu_engine/tvgSwStroke.cpp
@@ -666,7 +666,12 @@ static void _beginSubPath(SwStroke& stroke, const Point& to, bool closed)
 
 static void _endSubPath(SwStroke& stroke)
 {
-    if (stroke.closedSubPath) {
+    //a degenerate (zero-length) subpath has no real direction to close a loop with.
+    //treat it the same as an open path so round/square caps still render as a dot,
+    //even though it was technically closed (e.g. moveTo(x,y); close();).
+    auto degenerate = tvg::zero(stroke.subPathLength) && stroke.center == stroke.subPathStart;
+
+    if (stroke.closedSubPath && !degenerate) {
         //close the path if needed
         if (stroke.center != stroke.subPathStart) _lineTo(stroke, stroke.subPathStart);
 
--- a/src/renderer/gpu_engine/gl/tvgGlTessellator.cpp
+++ b/src/renderer/gpu_engine/gl/tvgGlTessellator.cpp
@@ -180,10 +180,15 @@ void Stroker::close()
 {
     if (length(mState.prevPt - mState.firstPt) > 0.015625f) {
         lineTo(mState.firstPt);
+        join(mState.firstPtDir);
+    } else if (mState.firstPtDir.x == 0.f && mState.firstPtDir.y == 0.f) {
+        // moveTo + close with nothing actually drawn in between (degenerate subpath):
+        // draw a cap dot instead of joining, matching the sw_engine's zero-length behavior.
+        cap();
+    } else {
+        // join firstPt with prevPt
+        join(mState.firstPtDir);
     }
-
-    // join firstPt with prevPt
-    join(mState.firstPtDir);
 }
 
 
--- a/src/renderer/gpu_engine/wg/tvgWgTessellator.cpp
+++ b/src/renderer/gpu_engine/wg/tvgWgTessellator.cpp
@@ -178,10 +178,15 @@ void WgStroker::close()
 {
     if (length(mState.prevPt - mState.firstPt) > 0.015625f) {
         lineTo(mState.firstPt);
+        join(mState.firstPtDir);
+    } else if (mState.firstPtDir.x == 0.f && mState.firstPtDir.y == 0.f) {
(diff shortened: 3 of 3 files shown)


Nor-s:
image

image

@mnchn-prog 

Please address the GitHub Copilot review comments as appropriate. 
If any of them are not valid or applicable, feel free to close them.

Could you also check the `bounding box` and `gradient stroke` examples using the SW engine on `thorvg.example`?

The results seem significantly different from before, 
so we need to determine whether this is an issue with the Pixel Inspector or with the code itself.

hermet:
@mnchn-prog We haven't decided on the direction of the specification yet.

mnchnk:
Thanks for the feedback. Regarding the cpu_engine change — the Pixel Inspector report showed 106/492 differences on the sw backend (vs. 0 for gl/wg), so the fix as written is clearly touching more cases than the intended moveTo + close() scenario. I'd like to dig into which of those 106 cases are unintended and narrow the condition down, if that's useful.

On the spec direction — I understand the backward compatibility concern you raised in the issue. I lean toward following the AE/lottie-web reference behavior for consistency, but I'd like to understand more about what the backward compatibility risk looks like in practice (e.g. is there existing content that relies on the current "no dot" behavior?), so I can factor that into a narrower fix if the team decides to move forward with this.

hermet:
@mnchnk There is no global standard that defines the expected rendering result for the rendering features. Different rendering engines, services, and applications may reasonably interpret them in different ways.

ThorVG is an independent rendering engine with no direct relationship to AE or lottie-web. Neither of them uses ThorVG, and ThorVG is not being developed specifically for their needs. In addition, AE is not aligned with W3C specifications, and both rendering behavior often differs from each other. Therefore, AE or lottie-web behavior alone should not be treated as the basis for defining ThorVG's rendering semantics in this case.

ThorVG's responsibility is to provide a clear, consistent, and stable rendering specification. Services and applications built on top of ThorVG can then implement their desired behavior based on that specification. As long as the API behavior is well defined, they can explicitly decide whether a particular case should be rendered as a point, a line, or nothing at all.

For this reason, I do not think ThorVG should change its existing behavior without a strong and broadly applicable user requirement. Also, before introducing such a change, we should carefully consider who would actually benefit from it, what practical problem it solves, and how many behavioral regressions it could introduce for existing users and services.

Unless there is a compelling real-world use case that requires this behavior now, it would be safer to preserve the current behavior. If this semantic change is still considered necessary, it would be more appropriate to defer it to a major release, where compatibility-breaking changes can be evaluated and introduced more carefully. That's why it's still leaves as a `question` for ticket.

PR #4613, opened 2026-07-28
lottie/slot: allow image overriding without dimensions

tinyjin:
A slot may omit any key, so an image override can arrive with no "w"/"h". Two things went wrong with such a payload.

An asset object carries no type tag, so parseAsset() tells an image from a sound by its dimensions. A path-based image slot therefore fell through to the audio branch and was built as a LottieAudio, which the image slot case then read back as a LottieImage.

The dimensions were also passed to picture->size() as they were, collapsing the picture to 0x0 so the image disappeared from the scene.

Both are addressed here:
- LottieBitmap defaults its dimensions to -1 now, and picture->size() is skipped while they stay below zero.
- The asset kind of a slot is settled by the property type its paired target was registered with, instead of being guessed from the payload.

> Originally commited by @theashraf in 61687a62fb022c2afcbf063244ae2134069c1f33, 73d8ef79fc3507fb914fbe60a725f376f61d0c93

### Reproduction

- full source : [LottieSlotImageSize.cpp](https://github.com/user-attachments/files/30454717/LottieSlotImageSize.cpp)

| Before | After |
|---------|---------|
| CleanShot 2026-07-28 at 17 55 01@2x | CleanShot 2026-07-28 at 17 54 00@2x |

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -976,8 +976,9 @@ void LottieBuilder::updateImage(LottieGroup* layer)
 
     //resolve an image asset if need
     if (resolver && !image->resolved) {
-        resolver->func(picture, image->bitmap.path, resolver->data);
-        picture->size(image->bitmap.width, image->bitmap.height);
+        auto& bitmap = image->bitmap;
+        if (bitmap.width != 0.0f && bitmap.height != 0.0f) resolver->func(picture, bitmap.path, resolver->data);
+        if (bitmap.width >= 0.0f && bitmap.height >= 0.0f) picture->size(bitmap.width, bitmap.height);
         image->resolved = true;
     }
 
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -300,10 +300,12 @@ void LottieImage::prepare(bool external)
     //Prepare the Picture image
     auto result = Result::Unknown;
     auto picture = Picture::gen();
-    if (bitmap.size > 0) result = picture->load(bitmap.data, bitmap.size, bitmap.mimeType);
-    else if (external) result = picture->load(bitmap.path);
+    if (bitmap.width != 0.0f && bitmap.height != 0.0f) {
+        if (bitmap.size > 0) result = picture->load(bitmap.data, bitmap.size, bitmap.mimeType);
+        else if (external) result = picture->load(bitmap.path);
+    }
     if (result == Result::Success) resolved = true;
-    picture->size(bitmap.width, bitmap.height);
+    if (bitmap.width >= 0.0f && bitmap.height >= 0.0f) picture->size(bitmap.width, bitmap.height);
     bitmap.picture = picture;
     picture->ref();
 }
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -1038,7 +1038,7 @@ void LottieParser::parseImage(LottieImage* image, const char* data, const char* 
     image->prepare(external);
 }
 
-LottieObject* LottieParser::parseAsset()
+LottieObject* LottieParser::parseAsset(LottieProperty::Type type)
 {
     enterObject();
 
@@ -1049,8 +1049,8 @@ LottieObject* LottieParser::parseAsset()
     const char* sid = nullptr;
     const char* data = nullptr;
     const char* subPath = nullptr;
-    float width = 0.0f;
-    float height = 0.0f;
+    float width = -1.0f;
+    float height = -1.0f;
     auto embedded = false;
 
     while (auto key = nextObjectKey()) {
@@ -1072,7 +1072,7 @@ LottieObject* LottieParser::parseAsset()
         else skip();
     }
     if (data) {
-        if (!strncmp(data, "data:image/", 11) || width != 0.0f || height != 0.0f) {
+        if (type == LottieProperty::Type::Image || !strncmp(data, "data:image/", 11) || width >= 0.0f || height >= 0.0f) {
             auto asset = new LottieImage;
             parseImage(asset, data, subPath, embedded, width, height);
             if (sid) registerSlot(asset, sid, asset->bitmap);
@@ -1734,7 +1734,7 @@ LottieProperty* LottieParser::parse(LottieSlot* slot)
         case LottieProperty::Type::Image: {
             LottieObject* obj = nullptr;
             while (auto key = nextObjectKey()) {
-                if (KEY_AS("p")) obj = parseAsset();
+                if (KEY_AS("p")) obj = parseAsset(slot->type);
                 else skip();
             }
             if (!obj) return nullptr;
--- a/src/loaders/lottie/tvgLottieParser.cpp
+++ b/src/loaders/lottie/tvgLottieParser.cpp
@@ -1028,14 +1028,15 @@ void LottieParser::parseAudio(LottieAudio* audio, const char* data, const char* 
     parseAssetSource(audio->src, data, subPath, "data:audio/", embedded, audio->src.external);
 }
 
-void LottieParser::parseImage(LottieImage* image, const char* data, const char* subPath, float width, float height, bool embedded)
+void LottieParser::parseImage(LottieImage* image, const char* data, const char* subPath, float width, float height, bool sized, bool embedded)
 {
     if (!parseAssetSource(image->asset, data, subPath, "data:image/", embedded, image->asset.external)) return;
+    image->asset.sized = sized;
     image->asset.width = width;
     image->asset.height = height;
 }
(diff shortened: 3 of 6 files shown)


tinyjin:
@theashraf Could you confirm if specified issue resolved?

theashraf (approved):
Thanks @tinyjin, that resolves my issue

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1094,7 +1094,7 @@ LottieObject* LottieParser::parseAsset()
         else skip();
     }
     if (data) {
-        if (!strncmp(data, "data:image/", 11) || width != 0.0f || height != 0.0f) {
+        if (type == LottieProperty::Type::Image || !strncmp(data, "data:image/", 11) || width >= 0.0f || height >= 0.0f) {
```
hermet: 1. We might need to confirm these conditoon is essential, otherwise remove them.
`width >= 0.0f || height >= 0.0f`

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -977,7 +977,7 @@ void LottieBuilder::updateImage(LottieGroup* layer)
     //resolve an image asset if need
     if (resolver && !image->resolved) {
         resolver->func(picture, image->bitmap.path, resolver->data);
```
hermet: zero size bitmap is meaningless in the rendering?
```
if (image->bitmap.width > 0.0f && image->bitmap.height > 0.0f) {
    resolver->func(picture, image->bitmap.path, resolver->data);
    picture->size(image->bitmap.width, image->bitmap.height);
    image->resolved = true;
}
```

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -305,7 +305,7 @@ void LottieImage::prepare(bool external)
     if (bitmap.size > 0) result = picture->load((const char*)bitmap.data, bitmap.size, bitmap.mimeType);
```
hermet: same here:
```
if (bitmap.width > 0.0f && bitmap.height > 0.0f) {
    if (bitmap.size > 0) result = picture->load((const char*)bitmap.data, bitmap.size, bitmap.mimeType);
    ...
    picture->size(bitmap.width, bitmap.height);
}
bitmap.picture = picture;
picture.ref();
```

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1071,8 +1071,8 @@ LottieObject* LottieParser::parseAsset()
     const char* sid = nullptr;
     const char* data = nullptr;
     const char* subPath = nullptr;
-    float width = 0.0f;
```
hermet, line 1049: This might not need a change.

hermet (commented):
@tinyjin Please check comments. thanks.

hermet:
@tinyjin please resolve the conflict. thanks.

src/loaders/lottie/tvgLottieParser.cpp
tinyjin: This looks essential, for path image in lottie parsing

src/loaders/lottie/tvgLottieBuilder.cpp
tinyjin: when size is `-1`, that means slot image set in undefined size. By this patch we try to load that image with default bitmap size.

However, `resolver->func(picture, image->bitmap.path, resolver->data)` is good point.
I agree zero size bitmap doesn't have to call resolver to render (anyway it wouldn't use)

src/loaders/lottie/tvgLottieParser.cpp
tinyjin: it depends on policy. In the Lottie spec, image can be overwritten with zero size (_users might want to use this scenario to hide an image_ regardless it's efficient way or not)

If we set `0` to omitted case, there is no way to handle zero size overriding intended.

This may be something we choose not to handle, since doing so could add unnecessary complexity to the code without offering much practical value, even if the intention is to comply with the specification.

src/loaders/lottie/tvgLottieBuilder.cpp
tinyjin: updated so it doesn't call resolver callback unnecessarily

src/loaders/lottie/tvgLottieParser.cpp
tinyjin: fyi. this zero size overriding has been supported, thus I chose to still handle this (_width/height sentinel is set to -1_)

hermet:
resolved conflicts.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1047,8 +1047,8 @@ LottieObject* LottieParser::parseAsset()
...
-    float height = 0.0f;
+    float width = -1.0f;
```
hermet: @tinyjin Rather than value based, Let's introduce a boolean for clearer condition context and better maintainability.

src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -1028,14 +1028,15 @@ void LottieParser::parseAudio(LottieAudio* audio, const char* data, const char*
...
+void LottieParser::parseImage(LottieImage* image, const char* data, const char* subPath, float width, float height, bool sized, bool embedded)
 {
     if (!parseAssetSource(image->asset, data, subPath, "data:image/", embedded, image->asset.external)) return;
+    image->asset.sized = sized;
     image->asset.width = width;
     image->asset.height = height;
 }
 
-LottieObject* LottieParser::parseAsset()
+LottieObject* LottieParser::parseAsset(LottieProperty::Type type)
```
hermet, line 1039: does this type introduction is required for the dimension overriding?

tinyjin: @hermet Yes this is required not to mislead LottieImage to LottieAudio in path-based asset. (They don't have header string, they aren't encoded + the slot can omit size data)

I thought using the existing `type` would be clear, we can also just change this to `bool image` style otherwise.

hermet (approved):

src/loaders/lottie/tvgLottieProperty.h
```diff
@@ -935,6 +935,7 @@ struct LottieTextDoc : LottieProperty
 
 struct LottieAsset : LottieProperty, AssetSrc
 {
+    bool sized = false;
```
hermet, line 938: please add the sized at the end of properties.

hermet (commented):
@tinyjin minor nitpick

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -976,7 +976,7 @@ void LottieBuilder::updateImage(LottieLayer* layer, float frameNo)
     //resolve an image asset if need
     if (resolver && !image->valid) {
         resolver->func(picture, image->asset.path, resolver->data);
-        picture->size(image->asset.width, image->asset.height);
+        if (image->asset.sized) picture->size(image->asset.width, image->asset.height);
```
hermet: I guess this condition isn't necessary. The picture size should be required by default regarless of slot overriding.

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -308,7 +308,7 @@ Picture* LottieImage::get()
     auto result = Result::Unknown;
     if (asset.size > 0) result = picture->load(asset.data, asset.size, asset.mimeType);
     else if (asset.external) result = picture->load(asset.path);
-    picture->size(asset.width, asset.height);
+    if (asset.sized) picture->size(asset.width, asset.height);
```
hermet: same

src/loaders/lottie/tvgLottieProperty.h
tinyjin: @hermet I put `sized` there intentionally, for padding. The incoming bytes starts from `53`IINW. 1 byte(bool) can come at the top.

src/loaders/lottie/tvgLottieBuilder.cpp
tinyjin: without this, the slot image (omits w/h) will set the picture to zero size

src/loaders/lottie/tvgLottieModel.cpp
tinyjin: this also looks needed (same reason)

hermet: @tinyjin  Image should have size. overriden image also must have it. Please double-check it.

hermet: Or maybe I think we need to override the w/h values when the size is explicitly enabled. (Shouldn't override the values with zero)

tinyjin: @hermet updated

Issue #4604, opened 2026-07-27
Lottie asset resolver: font requests don't carry the font name (dotlottie)


theashraf:
**Problem**

`Tvg_Picture_Asset_Resolver` receives only `src`. For text layers, `updateURLFont()` in `src/loaders/lottie/tvgLottieBuilder.cpp` invokes the resolver with `text->font->path` (the Lottie `fPath`) whenever `paint->font(doc.name)` fails — the referenced font name (`fName`) is passed only in the no-path case (`"name:<fName>"`).

Since the font lookup is by name and re-runs on every frame update, a resolver must register the loaded data under the exact `fName` for resolution to stick. But `fName` is an arbitrary label that can't be derived from `fPath`, so resolvers are forced to pre-parse the animation's `fonts.list` to build a `fPath` → `fName` map. Without it, the callback re-fires for every text layer on every frame.

(Images are fine: they latch via `image->resolved` and the bytes load directly into the paint.)

**Proposal**

Pass the font name to the callback, e.g.:

```c
typedef bool (*Tvg_Picture_Asset_Resolver)(Tvg_Paint paint, const char* src, const char* name, void* data);
```

with `name` null for images. Any equivalent shape (always prefixing `name:<fName>|<fPath>`, a dedicated font resolver, etc.) works too — the point is that the resolver shouldn't need to parse the Lottie JSON itself to know which name to register the font under. The API is marked experimental, so a signature change should still be possible.

tinyjin:
So far, my thinking is as follows:

Asset Resolver should be a common interface for assets in general, not something limited to specific asset types like images or fonts.

It would be better to have a design where users can simply switch based on the one source path.
Also, instead of passing a font name as a parameter, I believe users should be able to register fonts under any name they choose in the asset resolver.

This also helps ensure that the resolver can be used not only for path correction, but also for asset substitution. In that sense, the API can avoid accidentally matching a new font file to completely different name just because of a constrained design.

I’ll patch it so that callbacks are not re-fired after a successful resolve, unless the text doc's font name changes dynamically.

The details may still be adjusted a bit, this is my current point of view.

PR #4624, opened 2026-07-29
lottie/text: stop refiring the asset resolver for URL fonts

tinyjin:
The font asset resolver lets users resolve an arbitrary font under an arbitrary name based on src. However, when the resolved font is registered under a name that is not defined in the lottie asset (fName), updateURLFont() always treats it as a font load failure, so the already resolved and rendered font is re-resolved every frame, and the callback is invoked every frame as well.

Until now this forced the user side to parse the fName registered in the JSON separately and match it. With this patch, resolving with a new font name no longer causes that problem.

issue: https://github.com/thorvg/thorvg/issues/4604

# Note

Implements an alias on the font loader so a resolver can supply the new font it matches by src, correcting font stays loaded.

CleanShot 2026-07-30 at 03 12 16@2x

When the asset resolver maps a font onto the text paint, the loader is aliased to the font name the lottie document currently sets. A font that has already been resolved is therefore found on the following frames, and the resolver is no longer fired every frame for it.

The lottie-declared name is not overwritten by the name the user registered, so when the font name changes dynamically(_such as keyframes, expressions or slots_), the lookup misses again and the callback is fired for the new request.

> FYI that the alias is currently used only for the resolver case.
Aliasing is common in font management, though it is usually designed as a 1:N relation. this could be revised if the extensibility is needed.

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -994,7 +994,7 @@ void LottieBuilder::updateURLFont(LottieLayer* layer, float frameNo, LottieText*
     if (paint->font(doc.name) != Result::Success) {
         char* src;
         bool free = false;
-        if (text->font && text->font->path) src = text->font->path;
+        if (text->font && text->font->path && tvg::equal(doc.name, text->font->name)) src = text->font->path;
         else {
             auto len = (strlen(doc.name) + 6);
             src = tvg::malloc<char>(sizeof(char) * len);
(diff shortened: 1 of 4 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1003,7 +1003,7 @@ void LottieBuilder::updateURLFont(LottieLayer* layer, float frameNo, LottieText*
         }
         if (!resolver || !resolver->func(paint, src, resolver->data)) {
             paint->font(nullptr);  //fallback to any available font
-        }
+        } else LoaderMgr::aliasing(to<TextImpl>(paint)->loader, doc.name);
```
hermet, line 1006: This hackish code is not allowed.

src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1003,7 +1003,7 @@ void LottieBuilder::updateURLFont(LottieLayer* layer, float frameNo, LottieText*
         }
         if (!resolver || !resolver->func(paint, src, resolver->data)) {
```
hermet, line 1004: Lets pass a reserved name.
`func(paint, doc.name (name) , src, resolver->data)`

hermet (changes requested):

Issue #4047, opened 2025-12-05
Canvas origin


angelofb:
Is it possible to set the origin of the canvas? For example, at the bottom-left instead of the top-left.

hermet:
@angelofb Hi! Currently, it's not supported. You can probably append a root scene and control the coordinates using a transform.

PR #4627, opened 2026-07-29
api: add origin reference point API for Paint

mnchnk:
### Summary
Answers #4047 — is it possible to set the canvas/paint origin (e.g. bottom-left instead of top-left)?

There was no way to do this directly. Paint::translate()'s coordinate system is hardcoded to top-left with the vertical axis pointing down. This PR adds Paint::origin(Origin position), letting a paint's local coordinate system reference the bottom-left corner instead, with the vertical axis pointing up.

```
enum struct Origin : uint8_t
{
    TopLeft = 0,  //default
    BottomLeft
};

Result origin(Origin position) noexcept;
```

### Design notes
Scoped to TopLeft/BottomLeft only, matching what the issue actually asked for. Since enum values are purely additive, TopRight/BottomRight/Center can be added later without breaking compatibility.

Placed on Paint (not Scene/Canvas), since it's conceptually the same category of feature as translate()/rotate()/scale(), which already live there and apply uniformly to any paint type (Shape/Scene/Picture/Text).

Like other Paint transform setters, returns Result::InsufficientCondition if a custom transform() matrix has already been applied.

### Implementation

My first attempt implemented origin() on top of the existing transform(Matrix) override mechanism, but that turned out to conflict with translate(): setting a custom matrix flips on overriding, which silently blocks any translate() call made afterward, and rebuilding the matrix from scratch also discarded any translate() set beforehand.

Instead, origin() now stores its state (Origin, offset) independently from the existing translate()/rotate()/scale() values in Paint::Impl::tr, and both are only combined into the final matrix inside tr.update() (called lazily whenever the transform is queried). This mirrors how translate/rotate/scale already coexist via separate stored primitives (tx, ty, degree, scale) rather than each mutating a shared matrix directly. As a result, origin() and translate() can be called in either order.

The bottom-left offset is computed once (at origin() call time) from the paint's current local bounds (Paint::Impl::bounds() with an identity matrix), so it reflects the paint's own geometry rather than a hardcoded value.

### Testing
```
struct UserExample : tvgexam::Example
{
    {
        shape->moveTo(50, 0);
        shape->lineTo(0, 100);
        shape->lineTo(100, 100);
        shape->close();
        shape->fill(255, 0, 0);         //빨강: 원래(TopLeft)
        shape->translate(50, 50);
        canvas->add(shape);

        auto flipped = tvg::Shape::gen();
        flipped->moveTo(50, 0);
        flipped->lineTo(0, 100);
        flipped->lineTo(100, 100);
        flipped->close();
        flipped->fill(0, 0, 255);       //파랑: origin(BottomLeft)
        flipped->origin(tvg::Origin::BottomLeft);
        flipped->translate(200, 50);
        canvas->add(flipped);

        return true;
    }
};
```
image

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -322,6 +322,18 @@ enum struct Type : uint8_t
     RadialGradient         ///< RadialGradient class
 };
 
+/**
+ * @brief Enumeration specifying the origin reference point of a paint's coordinate system.
+ * 
+ * @see Paint::origin()
+ * 
+ * @since 1.2
+ */
+enum struct Origin : uint8_t
+{
+    TopLeft = 0, ///< Default. Origin at the top-left corner, with the horizontal and vertical axes pointing right and down.
+    BottomLeft  ///< Origin at the bottom-left corner, with the vertical axis pointing up.
+};
 
 /**
  * @brief A data structure representing a point in two-dimensional space.
@@ -777,6 +789,21 @@ struct TVG_API Paint
      * @since 1.0
      */
     static void rel(Paint* paint) noexcept;
+
+    /**
+     * @brief Sets the origin reference point for the paint's coordinate system.
+     *
+     * By default, the origin is at the top-left corner of the paint,
+     * with the horizontal and vertical axes pointing right and down.
+     *
+     * @param[in] position The origin position to apply. The default is @c Origin::TopLeft.
+     *
+     * @retval Result::InsufficientCondition in case a custom transform is applied.
+     * @see Paint::transform()
+     *
+     * @since 1.2
+     */
+    Result origin(Origin position) noexcept;
 
 protected:
     virtual ~Paint();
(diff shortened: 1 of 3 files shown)


hermet:
@mnchnk Sorry. We will not accept arbitrary feature additions.

For issues labeled as questions, we first need to reach an alignment on the direction. The pros and cons of adding a feature should be carefully evaluated. In particular, when introducing a new API, the specification should be thoroughly aligned in advance.

mnchnk:
Sorry for jumping straight to an implementation without aligning on the direction first — that was my mistake.

For context, the reasoning behind the design: I scoped it to just Origin::TopLeft/BottomLeft, matching what the issue actually asked for (enum values are additive, so more can be added later without breaking compatibility). I put it on Paint rather than Scene/Canvas since it's the same category of feature as translate()/rotate()/scale(), which already live there and apply uniformly to any paint type.

hermet:
@mnchnk The issue is still in the question state, and we have not yet had a concrete discussion & direction. We cannot add every requested change simply because someone asks for it.

PR #4629, opened 2026-07-29
svg_loader: fix memory errors on malformed elements

JSUYA:
mask-type and character data are written into the node union without checking which element they land on, and a postponed CSS class keeps a pointer that a repeated class attribute frees. Check the node type before writing to the union, and read the class back from the node.

Cases fixed:
 - `<image href="data:image/svg+xml;base64,..." mask-type="Alpha"/>`
mask-type is handled through the generic style attribute table, so it wrote node->node.mask.type on any element - on an <image> or a <use> that corrupts a pointer the scene builder later dereferences
 - `<svg><text><rect x="1"/>A</text></svg>`
character data is appended to node->node.text of whatever element is open, and on a <rect> that pointer aliases the geometry floats
 - `<svg ""><g "=""class="l"e=""class="""><style>`
the second class attribute frees the class pointer that the postponed pass still holds

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -1061,6 +1061,7 @@ static void _handleFilterAttr(TVG_UNUSED SvgParserContext* ctx, SvgNode* node, c
 
 static void _handleMaskTypeAttr(TVG_UNUSED SvgParserContext* ctx, SvgNode* node, const char* value)
 {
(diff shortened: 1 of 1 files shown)


JSUYA:
This PR is a defensive code against malform tags formatting and has the same goal as the work at https://github.com/thorvg/thorvg/pull/4628.
Although not directly related to the reported issue, it was processed as a separate PR for a common purpose.

Issue #4630, opened 2026-07-30
trimpath() only normalizes offsets by one period, contradicting the documented circular range


mungjin01:
`inc/thorvg.h:1411` documents that `begin`/`end` outside 0–1 "are wrapped around in a
manner similar to angle wrapping, effectively treating the range as circular".
In practice `_get()` (`tvgRender.cpp:635`) applies `±1.0f` at most once, so only
`[-1, 2]` is handled; anything beyond renders nothing, with no error or warning.

This breaks the most natural way to write a rotating trim animation:

```cpp
// after ~3 seconds begin exceeds 2.0 and the shape silently disappears
shape->trimpath(0.34f * elapsed, 0.34f * elapsed + 0.2f);
```

**Why it renders nothing rather than erroring.** `_trim()` converts the normalized pair
into absolute arc lengths (`trimStart = begin * totalLength`). With `begin = 2.06` on a
path of length 1382, `trimStart` becomes ~2847 — past the end of the path. `_trimPath()`
walks `len` from 0 to `totalLength`, so `len <= trimStart` stays true for the whole walk
while both inner conditions (`len + dLen > trimEnd`, `len + dLen > trimStart + EPSILON`)
stay false. Neither `_trimAt()` nor `_add()` ever runs and `out` is left empty.

`RenderTrimPath::trim()` still returns `true` (it only returns `false` for
`pts.count < 2` or `begin == end`), so callers such as
`_genOutline()` proceed with an empty path. There is no exception, no log, and no
`Result` error — the shape just vanishes.

**Suggested fix:** fold both values by one common integer offset before the existing
seam logic — this preserves the length and orientation of the segment, so all currently
working inputs reduce to identical values:

```cpp
auto shift = floorf(begin);
begin -= shift;
end -= shift;
if (fabsf(end - begin) >= 1.0f) { begin = 0.0f; end = 1.0f; return; }
```

Alternatively, if the current behaviour is intended, the documented contract should be
narrowed to state the supported range.

### Example
```
#include "Example.h"

struct UserExample : tvgexam::Example
{
    static constexpr float SPEED = 0.34f;  //revolutions per second
    static constexpr float SPAN = 0.2f;    //visible fraction of the ring

    void spinner(tvg::Canvas* canvas, float cx, float cy, float begin, uint8_t r, uint8_t g, uint8_t b)
    {
        auto ring = tvg::Shape::gen();
        ring->appendCircle(cx, cy, 170, 170);
[...]

PR #4631, opened 2026-07-30
renderer/trimpath: normalize offsets beyond one period

mungjin01:
`trimpath()` documents that `begin`/`end` outside the 0-1 range "are wrapped around in a manner similar to angle wrapping, effectively treating the range as circular", but `_get()` in `tvgRender.cpp` applied a `±1.0f` correction at most once. Only inputs within `[-1, 2]` were normalized; anything beyond rendered nothing at all.

This breaks the most natural way to drive a rotating trim animation, since the offset grows without bound:

```cpp
// once elapsed exceeds ~5.9s, begin > 2.0 and the shape silently disappears
shape->trimpath(0.34f * elapsed, 0.34f * elapsed + 0.2f);
```

The failure is silent. `_trim()` converts the pair into absolute arc lengths, so `begin = 2.06` on a path of length 1382 puts `trimStart` at ~2847, past the end of the path. The walk in `_trimPath()` then satisfies no branch and emits nothing, while `RenderTrimPath::trim()` still returns `true` (it only fails on `pts.count < 2` or `begin == end`). Callers proceed with an empty path, and there is no exception, log, or `Result` error.

The first commit folds both values by one common integer offset before the existing seam logic. Subtracting the same offset from both preserves the span's length and orientation, so every input in the previously working `[-1, 2]` domain reduces to exactly the same pair as before:

| input | before | after |
|---|---|---|
| `(0.3, 0.7)` | `(0.3, 0.7)` | same |
| `(1.3, 1.7)` | `(0.3, 0.7)` | same |
| `(-0.7, -0.3)` | `(0.3, 0.7)` | same |
| `(0.9, 1.1)` | `(0.9, 0.1)` wrapped | same |
| `(3.06, 3.26)` | nothing rendered | `(0.06, 0.26)` |
| `(-5.3, -5.1)` | nothing rendered | `(0.7, 0.9)` |

The second commit is separated for review because it defines behavior rather than preserving it. A span of a full period or longer (e.g. `(0.0, 2.0)`) previously reduced to an arbitrary sub-segment or an empty result depending on the values. Such a span geometrically covers the entire path, so it is now clamped to `(0, 1)`, and the header documents this.

The third commit adds coverage in `test/testSwEngine.cpp`, rendering trimmed strokes into offscreen buffers and requiring that out-of-range offsets produce pixel-identical output to their in-range counterparts. All offsets are binary-exact fractions so the folded and reference values rasterize identically without a tolerance. The full-period case is compared against `(1, 2)` rather than `(0, 1)`, since `RenderTrimPath::valid()` treats `(0, 1)` as "no trim" and draws a closed shape whose seam rasterizes differently from an open trimmed one.

Verified with `ninja -C build test` (all unit tests pass). 

Issues: #4630

### DIFF
--- a/inc/thorvg.h
+++ b/inc/thorvg.h
@@ -1409,6 +1409,7 @@ struct TVG_API Shape : Paint
      * @brief Sets the trim of the shape along the defined path segment, allowing control over which part of the shape is visible.
      *
      * If the values of the arguments @p begin and @p end exceed the 0-1 range, they are wrapped around in a manner similar to angle wrapping, effectively treating the range as circular.
+     * If the span between @p begin and @p end is a full period or longer, the entire path is displayed.
      *
      * @param[in] begin Specifies the start of the segment to display along the path.
      * @param[in] end Specifies the end of the segment to display along the path.
--- a/src/renderer/tvgRender.cpp
+++ b/src/renderer/tvgRender.cpp
@@ -639,6 +639,13 @@ static void _get(float& begin, float& end)
     begin -= shift;
     end -= shift;
 
+    //a span of a full period or longer covers the entire path
+    if (fabsf(end - begin) >= 1.0f) {
+        begin = 0.0f;
+        end = 1.0f;
+        return;
+    }
+
     auto loop = true;
 
     if (begin > 1.0f && end > 1.0f) loop = false;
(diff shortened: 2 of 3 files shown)


hermet:
@mungjin01 please send a PR when the issue is confirmed. Thanks.

PR #4632, opened 2026-07-30
lottie: fix use-after-free on image slot overrides

mshll:
## Problem

Applying image slot overrides (`LottieAnimation::override`) intermittently renders the wrong image in a slot, or nothing at all. With multiple image slots in one override batch, updating one slot can corrupt others. The corruption is nondeterministic and much easier to reproduce with `threads > 0`.

## Root cause

`LottieImage::prepare()` loads the encoded image with the default `copy = false`, so the picture loader borrows `bitmap.data` (e.g. `PngLoader::open` stores the pointer with `freeData = false`) and, with threading enabled, queues the actual decode on a worker via `TaskScheduler::request`.

Two callers free that buffer while the loader still holds it:

1. **Slot override parsing** (`tvgLottieParser.cpp`, `LottieProperty::Type::Image` case): after `parseAsset()` → `prepare()`, it copies the bitmap into a `LottieBitmap` prop - whose copy-ctor keeps only `picture`/`width`/`height` - and immediately `delete(obj)`s the parsed `LottieImage`. `~LottieBitmap` → `release()` → `tvg::free(data)`.
2. **`LottieImage::override()`**: releases the previously-owned buffer while the Picture (and its cached loader) live on.

Consequences:

- The queued decode reads freed memory → the slot renders garbage or fails to decode and draws nothing (permanently, since the dead Picture survives).
- The loader stays registered in `LoaderMgr`'s cache under `HASH_KEY(data)` - the raw freed address. When the allocator recycles that address for the next image's base64 buffer (common for consecutive slots in one batch), `_findFromCache` matches type + address and silently binds the new image to the previous image's decoded surface.

The cache code's own comment states the invariant being violated: *"caching is only valid for shareable."* A buffer that gets freed out from under the loader is not shareable.

## Fix

Pass `copy = true` in `LottieImage::prepare()`. The loader then owns its copy of the encoded data (freeing the caller's buffer is safe at any point), and `copy = true` also skips pointer-keyed cache registration, eliminating the aliasing path. Cost is one memcpy of the encoded data per image load.

An alternative would be transferring ownership of `data`/`size`/`mimeType` into `LottieBitmap`'s copy in the parser, which avoids the memcpy but does not address `override()` or the cache key. Happy to rework in that direction if preferred.

## Testing

- Full unit test suite passes (`-Dloaders="lottie, png, jpg, svg" -Dthreads=true`), including the `Lottie Slot` cases.
- Reproduced and verified the fix downstream in a production iOS app using thorvg through dotlottie-rs (software renderer, `threads = 2`, animations with 4 image slots re-overridden interactively): before the patch, slot updates intermittently blanked or cross-bound other slots; after it, behavior is correct across repeated override batches, pauses, and restarts.

### DIFF
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -302,9 +302,12 @@ void LottieImage::prepare(bool external)
     //Prepare the Picture image
     auto result = Result::Unknown;
     auto picture = Picture::gen();
-    if (bitmap.size > 0) result = picture->load(bitmap.data, bitmap.size, bitmap.mimeType);
+    if (bitmap.size > 0) result = picture->load(bitmap.data, bitmap.size, bitmap.mimeType, nullptr, true);
     else if (external) result = picture->load(bitmap.path);
-    if (result == Result::Success) resolved = true;
+    if (result == Result::Success) {
+        resolved = true;
+        bitmap.AssetSrc::release();
+    }
     picture->size(bitmap.width, bitmap.height);
     bitmap.picture = picture;
     picture->ref();


tinyjin:
@mshll Could you provide sample resource and slot JSON you've used? Thanks

mshll:
@tinyjin Sure, here is a minimal synthetic sample (the original asset I hit this with is proprietary, but this reproduces it).

The animation has two image layers whose assets carry `sid` (`slot_a`, `slot_b`), each an embedded 8x8 PNG (red / blue). The override replaces them with a green and a yellow PNG. Re-generating and re-applying the same batch a few times - as an interactive caller updating slots would - is enough.

<details>
<summary><b>slot-repro.json</b> (animation)</summary>

```json
{"v": "5.7.0", "fr": 60, "ip": 0, "op": 60, "w": 100, "h": 100, "nm": "slot-repro", "ddd": 0, "assets": [{"id": "img_a", "w": 8, "h": 8, "u": "", "p": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAgAAAAICAYAAADED76LAAAAFklEQVR4nGP8z8DwnwEPYMInOXwUAAASWwIOH0pJXQAAAABJRU5ErkJggg==", "e": 1, "sid": "slot_a"}, {"id": "img_b", "w": 8, "h": 8, "u": "", "p": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAgAAAAICAYAAADED76LAAAAFklEQVR4nGNkYPj/nwEPYMInOXwUAAAQXQIOZWZ6QQAAAABJRU5ErkJggg==", "e": 1, "sid": "slot_b"}], "layers": [{"ddd": 0, "ind": 1, "ty": 2, "nm": "A", "refId": "img_a", "ks": {"o": {"a": 0, "k": 100}, "r": {"a": 0, "k": 0}, "p": {"a": 0, "k": [30, 50, 0]}, "a": {"a": 0, "k": [4, 4, 0]}, "s": {"a": 0, "k": [400, 400, 100]}}, "ip": 0, "op": 60, "st": 0}, {"ddd": 0, "ind": 2, "ty": 2, "nm": "B", "refId": "img_b", "ks": {"o": {"a": 0, "k": 100}, "r": {"a": 0, "k": 0}, "p": {"a": 0, "k": [70, 50, 0]}, "a": {"a": 0, "k": [4, 4, 0]}, "s": {"a": 0, "k": [400, 400, 100]}}, "ip": 0, "op": 60, "st": 0}]}
```
</details>

<details>
<summary><b>slot-override.json</b> (slot overrides)</summary>

```json
{"slot_a": {"p": {"w": 8, "h": 8, "p": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAgAAAAICAYAAADED76LAAAAFklEQVR4nGNk+M/wnwEPYMInOXwUAAARXAIO3AvTcAAAAABJRU5ErkJggg==", "e": 1}}, "slot_b": {"p": {"w": 8, "h": 8, "p": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAgAAAAICAYAAADED76LAAAAFklEQVR4nGP8/5/hPwMewIRPcvgoAAAXZAMNHn1weQAAAABJRU5ErkJggg==", "e": 1}}}
```
</details>

<details>
<summary><b>slot-repro.cpp</b> (renders and samples one pixel per slot)</summary>

```cpp
// Minimal repro for image slot override use-after-free.
#include <thorvg.h>
#include <thorvg_lottie.h>
#include <cstdio>
#include <cstdlib>
#include <string>
#include <fstream>
#include <sstream>

using namespace tvg;

static std::string readFile(const char* path)
{
    std::ifstream f(path);
    std::stringstream ss;
    ss << f.rdbuf();
    return ss.str();
}

{
    if (argc < 3) {
        fprintf(stderr, "usage: %s <animation.json> <slots.json>\n", argv[0]);
        return 1;
    }

    if (Initializer::init(2) != Result::Success) return 1;

    auto animation = LottieAnimation::gen();

    auto data = readFile(argv[1]);
    if (picture->load(data.data(), (uint32_t)data.size(), "lottie", nullptr, false) != Result::Success) {
        fprintf(stderr, "load failed\n");
        return 1;
    }
    picture->size(100, 100);

    static uint32_t buffer[100 * 100];
    canvas->target(buffer, 100, 100, 100, ColorSpace::ARGB8888);
    canvas->add(picture);

    auto slots = readFile(argv[2]);

    // Re-generating and re-applying the same override batch mirrors an
    // interactive caller updating slots repeatedly.
    for (int i = 0; i < 5; ++i) {
        auto id = animation->gen(slots.c_str());
        if (id == 0) {
            fprintf(stderr, "gen failed\n");
            return 1;
        }
        if (animation->apply(id) != Result::Success) {
            fprintf(stderr, "apply failed\n");
            return 1;
        }
        animation->frame(0);
        canvas->update();
        canvas->draw(true);
        canvas->sync();

        // slot_a is overridden to solid green at (30,50); slot_b to solid yellow at (70,50)
        auto a = buffer[50 * 100 + 30];
        auto b = buffer[50 * 100 + 70];
        printf("iter %d: slot_a=%08X (expect FF00FF00) slot_b=%08X (expect FFFFFF00)\n", i, a, b);

        animation->del(id);
    }

    printf("done\n");
    Initializer::term();
    return 0;
}

```
</details>

Output on current `main` (`Initializer::init(2)`, SwCanvas):

```
iter 0: slot_a=FF00FF00 (expect FF00FF00) slot_b=FFFFFF00 (expect FFFFFF00)
iter 1: slot_a=FF00FF00 (expect FF00FF00) slot_b=FF00FF00 (expect FFFFFF00)
iter 2: slot_a=FF00FF00 (expect FF00FF00) slot_b=FFFFFF00 (expect FFFFFF00)
iter 3: slot_a=FF00FF00 (expect FF00FF00) slot_b=FF00FF00 (expect FFFFFF00)
iter 4: slot_a=FF00FF00 (expect FF00FF00) slot_b=FF00FF00 (expect FFFFFF00)
```

From iter 1 on, `slot_b` intermittently renders `slot_a`'s green image: `slot_b`'s freshly decoded buffer gets the heap address of a previously freed one, and the loader cache (keyed on that raw pointer) binds it to the stale decoded surface.

With this PR applied, all iterations render the expected colors, plain and under `MallocScribble=1`.

src/loaders/lottie/tvgLottieModel.cpp
```diff
@@ -302,7 +302,7 @@ void LottieImage::prepare(bool external)
     //Prepare the Picture image
     auto result = Result::Unknown;
     auto picture = Picture::gen();
-    if (bitmap.size > 0) result = picture->load((const char*)bitmap.data, bitmap.size, bitmap.mimeType);
+    if (bitmap.size > 0) result = picture->load((const char*)bitmap.data, bitmap.size, bitmap.mimeType, nullptr, true);
```
tinyjin: it'd relieve situation. But the point is that
Setting `copy` to `true`, then memory usage is still occupied even after `prepare()`.

If the copy is necessary, original buffer should be thrown away after loaded successfully.

tinyjin (changes requested):
Thank you for contribution. Please check comment :)

mshll:
@tinyjin Thanks for the  comment, good point. Now it frees the original buffer once the load succeeds

tinyjin (commented):
@mshll Could you squash the commits? thanks
Image

mshll:
@tinyjin done

tinyjin:
@mshll Could you please confirm whether current commit resolves your issue?

mshll:
@tinyjin yes still works with `AssetSrc::release()`

tinyjin (approved):
@mshll Thanks for your contribution!

hermet:
by somehow, we need to find a skip memory copy...

hermet:
@tinyjin @mshll  Sorry for the conflict, but coincidentally, I made this commit. We can close this if it resolves your issue. Could you please confirm it? Thanks.

https://github.com/thorvg/thorvg/commit/7531a212be57a461ca986023b8ca08376e7935f0

tinyjin:
@hermet @mshll 
CleanShot 2026-08-14 at 15 02 12@2x

It looks resolved with current main in my test example.
Previously, the overwritten box stayed green-green boxes since the new image coincidentally referenced previous slot case image.

[LottieSlotImageUpdate.cpp](https://github.com/user-attachments/files/31056739/LottieSlotImageUpdate.cpp)

hermet:
@mshll @tinyjin I close this, thanks for your contribution!

PR #4633, opened 2026-07-30
gif: introduce GIF image loader

Nor-s:
#### Features:
- Decode LZW image data with global and local color tables,
  transparency, and interlacing.
- Handle Do Not Dispose, Restore to Background, and Restore to
  Previous frame disposal.
- Calculate animation duration from per-frame delays.
- Support frame seeking and segment playback.

#### Limitations:
- Skip Plain Text rendering, Comment Extension metadata, and
  Application Extension metadata, including loop counts.
- Clear Restore to Background regions to transparent instead of
  restoring the logical-screen background color.

> [!NOTE]
> ```
> # experimental in the development
> media_loader = get_option('loaders').contains('media')
> gif_loader = get_option('loaders').contains('gif')
> ```


Nor-s:
#3975

PR #4634, opened 2026-07-30
lottie/loader: support time remapping slots

theashraf:
`tm` is an animatable float, so it should be slottable like any other float property. It never was — a `sid` on `tm` parsed fine and did nothing.

Slot overriding needs both sides: the parser must pass the owning object to `parseProperty()` so `parseCommon()` reaches the `sid` branch, and that object needs `override()` to match the sid. `LottieLayer` had neither.

```json
{"time_remap":{"p":{"a":0,"k":1.5}}}
```

Values are in seconds, matching raw `tm` semantics, since `remap()` resolves them through `frameAtTime()`. Only precomp and audio layers consult `remap()`.


tinyjin (approved):
LGTM, thanks @theashraf !

Issue #4621, opened 2026-07-28
svg: Speed up color lookup with binary search


J-unStiN:
It is assumed that the color array is of static constexpr type and is already sorted alphabetically.

Therefore, it appears that using binary search can improve speed.

```cpp
// Current implementation
for (unsigned int i = 0; i < (sizeof(colors) / sizeof(colors[0])); i++) {
    if (!strcasecmp(colors[i].name, str)) { ... }
}
```

Considering that ThorVG strictly avoids STL dependencies, it seems possible to implement a simple and lightweight custom binary search function suitable for this purpose. (or a simple hash lookup)

I would like to ask for your opinion on whether it is acceptable to improve this.

hermet:
Hi, I can only speculate it have a negligible impact. feel free to test and PR if it's practically effective. Thanks.

PR #4640, opened 2026-08-01
svg: optimize named color lookup with binary search

J-unStiN:
- Replace linear search O(N) with binary search O(log N) in _toColor()
- Update the sorting of the colors[] array to alphabetical order to support binary search

issue: #4621 

Benchmark results (1M iterations):
- Best case (first entries): linear ~17ms, binary ~45ms
- Average case (mid entries): linear ~310ms, binary ~45ms (~7x faster)
- Worst case (last entries): linear ~536ms, binary ~44ms (~12x faster)

---

| Scenario | Color Name (Index) | Linear Search Time | Binary Search Time | Performance Comparison |
|---|---|---:|---:|---|
| **Best Case** (Front of array) | `aliceblue` (1st) | 16.03 ms | 51.99 ms | Linear is ~**3.2x** faster |
| | `antiquewhite` (2nd) | 20.71 ms | 38.13 ms | Linear is ~**1.8x** faster |
| | `aqua` (3rd) | 17.92 ms | 38.48 ms | Linear is ~**2.1x** faster |
| **Average Case** (Middle of array) | `lightgray` (Around middle) | 297.83 ms | 55.01 ms | Binary is ~**5.4x** faster |
| | `lightgreen` (Around middle) | 310.45 ms | 68.69 ms | Binary is ~**4.5x** faster |
| | `lightgrey` (Middle) | 319.62 ms | 12.02 ms | Binary is ~**26.6x** faster |
| **Worst Case** (End of array) | `yellow` (Near end) | 529.80 ms | 41.43 ms | Binary is ~**12.8x** faster |
| | `yellowgreen` (Near end) | 547.99 ms | 54.50 ms | Binary is ~**10.0x** faster |
| | `whitesmoke` (Last) | 531.03 ms | 36.91 ms | Binary is ~**14.4x** faster |

### DIFF
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -461,6 +461,7 @@ static unsigned char _parseColor(const char* value, char** end)
 }
 
 
+// sorted by name
 static constexpr struct
 {
     const char* name;
@@ -694,7 +695,7 @@ static bool _toColor(const char* str, uint8_t& r, uint8_t&g, uint8_t& b, char** 
             }
         }
     } else {
-        //Handle named color (binary search)
+        // Handle named color (binary search) - colors[] must remain sorted
         int low = 0, high = static_cast<int>(sizeof(colors) / sizeof(colors[0])) - 1;
         while (low <= high) {
             auto mid = (low + high) / 2;
--- a/src/loaders/svg/tvgSvgLoader.cpp
+++ b/src/loaders/svg/tvgSvgLoader.cpp
@@ -491,8 +491,8 @@ static constexpr struct
     { "darkcyan", 0xff008b8b },
     { "darkgoldenrod", 0xffb8860b },
     { "darkgray", 0xffa9a9a9 },
-    { "darkgrey", 0xffa9a9a9 },
     { "darkgreen", 0xff006400 },
+    { "darkgrey", 0xffa9a9a9 },
     { "darkkhaki", 0xffbdb76b },
     { "darkmagenta", 0xff8b008b },
     { "darkolivegreen", 0xff556b2f },
@@ -520,9 +520,9 @@ static constexpr struct
     { "gold", 0xffffd700 },
     { "goldenrod", 0xffdaa520 },
     { "gray", 0xff808080 },
-    { "grey", 0xff808080 },
     { "green", 0xff008000 },
     { "greenyellow", 0xffadff2f },
+    { "grey", 0xff808080 },
     { "honeydew", 0xfff0fff0 },
     { "hotpink", 0xffff69b4 },
     { "indianred", 0xffcd5c5c },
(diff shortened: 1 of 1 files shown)


Nor-s:
Hi, I’d like to reproduce the benchmark on my machine as well.

Could you share the exact build and benchmark setup, including the benchmark code and commands, 
compiler and version, optimization flags, OS/CPU, and the input data used?

Thanks!

src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -694,14 +694,19 @@ static bool _toColor(const char* str, uint8_t& r, uint8_t&g, uint8_t& b, char**
             }
         }
     } else {
-        //Handle named color
-        for (unsigned int i = 0; i < (sizeof(colors) / sizeof(colors[0])); i++) {
-            if (!strcasecmp(colors[i].name, str)) {
-                r = ((uint8_t*)(&(colors[i].value)))[2];
-                g = ((uint8_t*)(&(colors[i].value)))[1];
-                b = ((uint8_t*)(&(colors[i].value)))[0];
+        //Handle named color (binary search)
```
JSUYA: // Handle named color (binary search) - colors[] must remain sorted

colors[] must always be kept sorted.
Since there is no code to enforce this, you must at least indicate it with a comment.

J-unStiN:
thank you, @Nor-s 
rather than committing the code used for the benchmark,
I will provide you with the branch containing that code.
https://github.com/J-unStiN/thorvg/commit/f794c45c745d2a01438268cbca087376f6c5223c

Apple clang version 17.0.0 (clang-1700.4.4.1)
Target: arm64-apple-darwin25.5.0
Thread model: posix
CPU: Apple M1 Pro
OS: MacOS Tahoe 26.5.2

src/loaders/svg/tvgSvgLoader.cpp
J-unStiN: It would be good to add comments as well.

If we switch to binary search and need to forcibly prevent sorting of array items for 'colors',
how about creating an order for 'name' and 'value' through test code and blocking it at the CI level?

J-unStiN: How about adding a comment to the current PR and registering the order sorting as an issue?

JSUYA: > How about adding a comment to the current PR and registering the order sorting as an issue?

There is no need to do that. colors[] will almost no change. It is sufficient to simply leave a hint so that the problem can be caught if an issue arises later.

J-unStiN: thank you

I applied the comments as you instructed, and added a sorting comment by name to colors
line:464 // sorted by name

JSUYA (approved):

Nor-s (approved):
Hello, I measured the performance impact of this code change using an SVG example in `thorvg.example`.

* macOS, M5 Pro, `optimization=3`, `threads=false`

#### Performance

**How**
* Used Instruments to measure the portion of the `UserExample::content` function’s duration attributable to `_toColor`
* Measured using **Processor Trace**

**Result**
- `_toColor` **duration**: **reduced by approximately** **16.6%**
- `_toColor` accounted for approximately **0.22%** of the total `content()` duration before the change and **0.17%** after.

> [!NOTE]
> Assets that use `_toColor` extensively may see a larger performance improvement, but the impact appears to be negligible in typical cases.

hermet:
@J-unStiN Good job. You might be interested in further tuning. The named color table isn't really optimized—we don't use the alpha value, and you can replace the integer with tvg::RGB so assigning the value becomes a simple copy.

see:
image

image

PR #4648, opened 2026-08-05
media: centralize video playback state management

Nor-s:
Move playback state transitions to the Video API layer so media backends expose consistent state after successful control requests.

Replace separate playback flags with a compact state enum and leave backend-specific loaders responsible only for platform operations and asynchronous transitions such as end-of-stream.

### DIFF
--- a/src/loaders/media/tvgVideo.cpp
+++ b/src/loaders/media/tvgVideo.cpp
@@ -64,32 +64,46 @@ Video::~Video()
 Result Video::play() noexcept
 {
     FETCH_LOADER(Result::InsufficientCondition);
-    return loader->play();
+    auto ret = loader->play();
+    if (ret == Result::Success) loader->state = MediaLoader::State::Playing;
+    return ret;
 }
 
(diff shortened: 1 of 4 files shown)


src/loaders/media/tvgVideo.cpp
```diff
@@ -64,32 +64,46 @@ Video::~Video()
 Result Video::play() noexcept
 {
     FETCH_LOADER(Result::InsufficientCondition);
-    return loader->play();
+    auto ret = loader->play();
+    if (ret == Result::Success) loader->state = MediaLoader::State::Playing;
```
hermet, line 68: Please don't expose the loader state handling to the video. Other APis as well.

hermet (changes requested):

src/loaders/media/tvgVideo.cpp
Nor-s: I’ll keep things as they are. Thank you for your review.

PR #4652, opened 2026-08-05
lottie: simplify property overrides

hermet:
- Add a variadic helper to match and replace slot properties.
- Remove duplicated override logic from Lottie model objects.
- Use the common color interpolation helper for text ranges.


hermet:
Hold due to the binary size increase.

Issue #1758, opened 2023-11-02
a broken gif conversion 


hermet:
Somewhat this gif code is buggy:

generated frame image no 37 by gif:
![image](https://github.com/thorvg/thorvg/assets/3711518/14fbeaf1-34f5-4a6e-a31c-e718943d3fba)

frame capture from the canvas at no.37:
![image](https://github.com/thorvg/thorvg/assets/3711518/bfedee98-f58e-444f-935c-8047edd9e871)

hermet:
![image](https://github.com/thorvg/thorvg/assets/3711518/d8f08e56-6f24-4511-b112-90533d1bb392)

JSUYA:
The first frame of some animation image palettes are being created incorrectly. I tried several things (creating and using a palette in advance) based on the gif.h code base, but I could not solve the problem. I think you can also add external_gif using giflib(?). What do you think?

hermet:
I think we can, but I feel like it's becoming burdensome for us to maintain all that stuff when we still need to pay attention to the existing core (buggy) features more than others, for instance, 'svg'.

alpakaDurumi:
Palettes can be built incorrectly on frames other than the first one as well.

Looking at `tvgGifEncoder.cpp`, `_splitPalette()` always halves both the palette range and the pixel set at every level, so one palette entry ends up covering roughly `numPixels / 256` pixels. Any color using fewer pixels than that cannot fill a leaf node on its own — it shares a leaf with other colors and is lost to the averaging done there.

Not reproducing every color when the source has more than 256 is an inherent GIF limitation, but this issue also occurs on content that uses far fewer colors.

As a possible direction, two things come to mind:

1. In `_splitPalette()`, stop recursing when every pixel in the range has the same color.
2. When the range holds multiple colors, split on a value boundary rather than always at the midpoint, so that pixels of the same color are never separated into different nodes.

Dumping the color table data out of the exported GIF, I found that a single palette contains a many entries holding **duplicate colors**. Collapsing those, together with the change in (2), seems like it could leave room for colors that only cover a few pixels to survive.

I'd like to work on this if you could assign it to me. (I'm also looking into whether the first frame's palette generation has a separate problem)

PR #4655, opened 2026-08-05
savers/gif: fix broken gif palette creation

alpakaDurumi:
## Summary

GIF conversion produced colors that do not exist in the source. Two independent defects in the palette code were responsible, and this PR fixes both.

- `_splitPalette()` handed out palette entries in proportion to pixel count, so colors covering a small area could not hold an entry of their own.
- `_makePalette()` skipped the alpha filter on the first frame, letting invisible pixels take entries that nothing would look up.

## 1. Split the pixel set by color value, not in half

`_splitPalette()` halved both the palette range and the pixel set at every level, so one palette entry ended up covering roughly `numPixels / 256` pixels. Any color using fewer pixels than that could not fill a leaf on its own, so it shared a leaf with unrelated colors and was lost to the averaging done there.

The boundary is now placed on a color value rather than on a fixed pixel count, so pixels sharing a value are never separated into different subtrees. A subset whose pixels are all one color has nothing left to split, so it stops there and is marked as a leaf, and the lookup no longer walks a subtree whose entries all hold the same color.

The median is now read from a histogram of the split component, which needs no reordering, so `_partitionByMedian()` and `_partition()` are gone. Instead, `_partitionByValue()` was added.

## 2. Build the first frame's palette correctly

`_makePalette()` computes the set of pixels to feed the palette by calling `_pickChangedPixels()` conditionally. Since that call only happens when a `lastFrame` exists, the first frame cannot use the

```cpp
    if (frame[3] >= TRANSPARENT_THRESHOLD)
```

filter inside `_pickChangedPixels()`. Pixels whose alpha is below `TRANSPARENT_THRESHOLD` therefore take part in building the first frame's palette, and the palette comes out broken.

`_makePalette()` now calls `_pickChangedPixels()` unconditionally so that alpha thresholding always runs, and `_pickChangedPixels()` gained a `!lastFrame` test so that on the first frame every pixel takes part in building the palette.

Also, when no pixel passes the alpha thresholding, `numPixels` becomes 0 and `_splitPalette()` turns into a no-op, leaving unexpected values in the palette. A `memset` was added to `gifBegin()` to clear the palette explicitly in that case.

## Results

| file | rendering (before) | rendering (after) | palette (before) | palette (after) |
| --- | --- | --- | --- | --- |
| Walk.json (frame 0) | test_f0 | test_f0 | out1 | out3 |
| swinging.json (frame 14) | test_f14 | test_f14 | out2 | out4 |

The palette dumps were produced with [this script](https://gist.github.com/alpakaDurumi/6ead468e1b2fefe55abd70185578826b). Each row is a frame and each column is one of the 256 entries in that frame's local color table. The narrower panel on the right shows only the unique colors of the same frame.

Issues: https://github.com/thorvg/thorvg/issues/1758

### DIFF
--- a/src/savers/gif/tvgGifEncoder.cpp
+++ b/src/savers/gif/tvgGifEncoder.cpp
@@ -265,11 +265,11 @@ static void _splitPalette(uint8_t* image, int numPixels, int firstElt, int lastE
     _splitPalette(image+subPixelsA*4, subPixelsB, splitElt, lastElt,  splitElt+splitDist, splitDist/2, treeNode*2+1, pal);
 }
 
-
 // Finds all pixels that have changed from the previous image and
 // moves them to the from of the buffer.
 // This allows us to build a palette optimized for the colors of the
 // changed pixels only.
+// With no previous frame, every visible pixel counts as changed.
 static int _pickChangedPixels(const uint8_t* lastFrame, uint8_t* frame, int numPixels, bool transparent)
 {
     int numChanged = 0;
@@ -277,7 +277,7 @@ static int _pickChangedPixels(const uint8_t* lastFrame, uint8_t* frame, int numP
 
     for (int ii=0; ii < numPixels; ++ii) {
         if (frame[3] >= TRANSPARENT_THRESHOLD) {
-            if (transparent || (lastFrame[0] != frame[0] || lastFrame[1] != frame[1] || lastFrame[2] != frame[2])) {
+            if (!lastFrame || transparent || (lastFrame[0] != frame[0] || lastFrame[1] != frame[1] || lastFrame[2] != frame[2])) {
                 writeIter[0] = frame[0];
                 writeIter[1] = frame[1];
                 writeIter[2] = frame[2];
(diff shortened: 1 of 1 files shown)


Nor-s:
@alpakaDurumi 

Since this commit addresses a separate issue, could you please open a separate PR for it?


https://github.com/thorvg/thorvg/pull/4655/changes/12c21bd206da8c65b2c8a97dafd963811fd2fd53

Nor-s (approved):
I found no issues with the logic or the output files. 
Below are the quantitative metrics affected by this code.


### Performance check

I benchmarked all 151 Lottie assets in `thorvg.example` with transparent and white backgrounds, for a total of 302 cases. Every GIF was generated at **400×400 resolution and 20fps**. Each case was run three times using a release build with `loaders=all`, and the median wall time was used.

> hyperfine --warmup 1 --runs 10


The average wall-time improvement across all cases was **20.853%**.

Top 3 improvements:

- `textrange.json` (transparent): **71.345%**
- `extensions/slot0.json` (transparent): **65.041%**
- `water_filling.json` (transparent): **62.364%**

Lowest 3 results:

- `rufo.json` (white): **−2.403%**
- `expressions/loopOut.json` (white): **+1.552%**
- `dash-offset.json` (white): **+1.798%**

hermet (approved):
Great, you can see this as well:
https://github.com/thorvg/thorvg/issues/4669

PR #4671, opened 2026-08-11
lottie: remove duplicate remapping

hermet:
The given layer frameNo is already remapped during update.

### DIFF
--- a/src/loaders/lottie/tvgLottieBuilder.cpp
+++ b/src/loaders/lottie/tvgLottieBuilder.cpp
@@ -1653,8 +1653,9 @@ static void _attachFont(LottieComposition* comp, LottieLayer* parent)
 {
     if (!parent->children.empty()) {
         // TODO: Consider to migrate this attachment to the frame update time.
-        auto text = static_cast<LottieText*>(parent->children.first());
-        if (text->type == LottieObject::Text) {
+        auto child = parent->children.first();
+        if (child->type == LottieObject::Text) {
+            auto text = static_cast<LottieText*>(child);
             auto& doc = text->doc(0);
             if (doc.name) {
                 auto len = strlen(doc.name);
@@ -1670,7 +1671,8 @@ static void _attachFont(LottieComposition* comp, LottieLayer* parent)
             return;
         }
     }
-    parent->children.clear();  // make the invalid text layer empty to prevent an access during rendering.
+    // Make the invalid text layer empty to prevent an access during rendering.
+    parent->clear();
 }
 
 static bool _buildComposition(LottieComposition* comp, LottieRootLayer* parent)
--- a/src/loaders/lottie/tvgLottieModel.cpp
+++ b/src/loaders/lottie/tvgLottieModel.cpp
@@ -351,17 +351,23 @@ uint32_t LottieGradient::populate(ColorStop& color, size_t count)
 {
     if (!color.input) return 0;
 
-    uint32_t alphaCnt = (color.input->count - (count * 4)) / 2;
+    auto clast = count * 4;
+    if (color.input->count < clast) {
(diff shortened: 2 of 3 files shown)


src/loaders/lottie/tvgLottieBuilder.cpp
```diff
@@ -1714,18 +1714,17 @@ static bool _buildComposition(LottieComposition* comp, LottieRootLayer* parent)
...
 
     auto ctrl = layer->audio();
-    auto active = frameNo >= layer->inFrame && frameNo < layer->outFrame;
     auto volume = active ? ctrl->volume(frameNo, tween, exps) : 100.0f;
 
     // audio condition is changed
     if ((active != ctrl->prevActive) || (active && !tvg::equal(volume, ctrl->prevVolume))) {
         auto& src = static_cast<LottieAudio*>(layer->children.first())->src;
-        auto offset = active ? (layer->remap(comp, frameNo, exps) - layer->remap(comp, layer->inFrame, exps)) / comp->frameRate : 0.0f;
+        auto offset = active ? (frameNo - layer->inFrame) / comp->frameRate : 0.0f;
```
tinyjin: Audio's `sr` & `tm` is only read here. By removing `layer-remap` thie line, the time stretch/remapping are gone.

We can still reduce one remapping by this line, unless we do not support `sr`/`tm` from Audio spec.
```cpp
auto offset = active ? layer->remap(comp, frameNo, exps) / comp->frameRate : 0.0f;
```

tinyjin (commented):
Thanks. Please check comment

src/loaders/lottie/tvgLottieBuilder.cpp
hermet: @tinyjin, was there any specific issue that led you to apply this?

1. sr/tm to be applied only when going through a precomp, but this applies the remapping again.
2. If frameNo needs to be remapped here, then the active condition above would also be evaluated with the wrong frameNo. Thus, your logic is non-sense.

tinyjin: @hermet I meant Audio has `sr` while `tm` is only going through a precomp as you said. (Precomp handles both in `remap()`)

The Audio's `sr` is previously handled by calling `layer->remap`. By replacing whole line, Audio's time stretch effect has gone.

However still using `layer->remap` function would look ambigious, audio just needs to apply `sr`. 
This kind of line can be considerable:
```
auto offset = active ? (frameNo - layer->inFrame) / (layer->timeStretch * comp->frameRate) : 0.0f;
```

For real sample, you can check this. After applying the audio plays sound different.
[sr_stretch.json](https://github.com/user-attachments/files/30968372/sr_stretch.json)

Let me know if I said something wrong.

hermet: > @hermet I meant Audio has `sr` while `tm` is only going through a precomp as you said. (Precomp handles both in `remap()`)
> 
> The Audio's `sr` is previously handled by calling `layer->remap`. By replacing whole line, Audio's time stretch effect has gone.
> 
> However still using `layer->remap` function would look ambigious, audio just needs to apply `sr`. This kind of line can be considerable:
> 
> ```
> auto offset = active ? (frameNo - layer->inFrame) / (layer->timeStretch * comp->frameRate) : 0.0f;
> ```
> 
> For real sample, you can check this. After applying the audio plays sound different. [sr_stretch.json](https://github.com/user-attachments/files/30968372/sr_stretch.json)
> 
> Let me know if I said something wrong.

I think it would be better to make the behavior consistent across all layers and fix the current inconsistency.

The timeline basis should be identical across all layers. The same rules and policies must be applied consistently.
Howerver, the audio layer behaves differently from the general layers. That `active` is as well.

PR #4687, opened 2026-08-18
loaders/jpg: prevent buffer overflow

Nor-s:
issue: https://github.com/thorvg/thorvg/security/advisories/GHSA-v45v-rfh3-xmp4

### main

image

### fetch 

-  Returns failure without triggering the overflow.

### DIFF
--- a/src/loaders/external_jpg/tvgJpgLoader.cpp
+++ b/src/loaders/external_jpg/tvgJpgLoader.cpp
@@ -20,6 +20,7 @@
  * SOFTWARE.
  */
 
+#include <climits>
 #include <turbojpeg.h>
 #include "tvgJpgLoader.h"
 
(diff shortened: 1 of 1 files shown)


src/loaders/external_jpg/tvgJpgLoader.cpp
```diff
@@ -106,10 +107,16 @@ bool JpgLoader::read()
         cs = ColorSpace::ABGR8888;
     }
 
-    auto image = (unsigned char *)tjAlloc(static_cast<int>(w) * static_cast<int>(h) * tjPixelSize[format]);
+    auto width = static_cast<int>(w);
+    auto height = static_cast<int>(h);
+    auto pixelSize = tjPixelSize[format];
+    if (width > INT_MAX / height / pixelSize) return false;
```
hermet: Let's go lean code. 
It seems like you just need this one line exception but nothing else.

hermet (approved):

src/loaders/external_jpg/tvgJpgLoader.cpp
Nor-s: Updated. Thank you for the review.

PR #4694, opened 2026-08-22
lottie: support ellipse slots

theashraf:
image


src/loaders/lottie/tvgLottieParser.cpp
```diff
@@ -567,8 +567,8 @@ LottieEllipse* LottieParser::parseEllipse()
 
     while (auto key = nextObjectKey()) {
         if (parseCommon(ellipse, key)) continue;
-        else if (KEY_AS("p")) parseProperty(ellipse->position);
-        else if (KEY_AS("s")) parseProperty(ellipse->size);
+        else if (KEY_AS("p")) parseProperty(ellipse->position, ellipse);
```
hermet, line 570: This looks weird. As only a few cases have an object as a parameter, while many others do not (ie parseRects). 
If that's essential, Rather than handling this on a case-by-case basis, I think we need a more consistent approach.

hermet: \+ @tinyjin Please check a comment.

Issue #4426, opened 2026-06-01
gl_engine: blur shader optimization


hermet:
@shenwenjie As a separate note, gaussian looks can be computed in CPU level in advance, generating a LUT rather than computing every pixel in order to skip the gaussian function call every pixel.

At very least, computing the following values on every pixel seems inefficient. It may be worth exploring possible optimizations.

`(2.0 * sigma * sigma)`
`(sqrt(2.0 * 3.141592) * sigma)`

_Originally posted by @hermet in https://github.com/thorvg/thorvg/pull/4425#discussion_r3331426413_

Saususge:
I’m currently looking into this issue and analyzing the proposed optimization approach. I’ll share updates here with my findings.

hermet:
@Saususge are you working on this?

Saususge:
yeah sorry for delay, I am working on now. Have been busy but working on it

PR #4697, opened 2026-08-22
gl: optimize Gaussian blur weight calculation

Saususge:
Precompute the Gaussian exponent coefficient on the CPU and pass it to the GL blur shaders.

  This avoids repeated square-root and division operations for every blur sample. The normalization factor is omitted
  because it is canceled out by the weight normalization.

  Tested with GL unit tests and a 1024×1024 before/after comparison.
  Only 2 / 1,048,576 pixels differ, by one channel value.

  issue: #4426
  
  left: before / right: after
side-by-side

### DIFF
--- a/src/renderer/gpu_engine/gl/tvgGlEffect.cpp
+++ b/src/renderer/gpu_engine/gl/tvgGlEffect.cpp
@@ -38,6 +38,10 @@ struct GlGaussianBlur
     float scale{};
     float extend{};
     float quality{};
+    float invTwoSigmaSquared{};
+    float dummy0{};
+    float dummy1{};
+    float dummy2{};
 };
 
 static float _blurQuality(uint8_t quality, float extend)
@@ -69,8 +73,10 @@ void GlEffect::update(RenderEffectGaussianBlur* effect, const Matrix& transform)
     if (!blur) blur = tvg::malloc<GlGaussianBlur>(sizeof(GlGaussianBlur));
     blur->sigma = effect->sigma;
     blur->scale = std::sqrt(transform.e11 * transform.e11 + transform.e12 * transform.e12);
-    blur->extend = 2 * blur->sigma * blur->scale;
+    const auto sigma = blur->sigma * blur->scale;
+    blur->extend = 2 * sigma;
     blur->quality = _blurQuality(effect->quality, blur->extend);
+    blur->invTwoSigmaSquared = tvg::zero(sigma) ? 0.0f : -0.5f / (sigma * sigma);
     effect->rd = blur;
     effect->valid = (blur->extend > 0);
 }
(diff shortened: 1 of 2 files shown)


Saususge:
No significant difference was measured in this environment:
Mesa 26.0.8 llvmpipe (LLVM 21.1.8, 256 bits) / EGL 1.5 / OpenGL 4.5 Core / Intel i5-14500HX, 1024x1024, sigma 32, 100 frames x 6 interleaved trials with 20 warm-up frames, CPU wall-clock (high_resolution_clock) over update/draw/sync — baseline avg 46.2ms vs current avg 45.7ms, within run-to-run variance.

Nor-s:
No noticeable performance improvement was observed on a Mac with an
M5 Pro using the OpenGL backend across 10 runs.
- ThorVG Example, Lottie -
- ThorVG Example, EffectDropShadow +1% fps (mean)

src/renderer/gpu_engine/gl/tvgGlEffect.cpp
```diff
@@ -135,14 +137,16 @@ void GlEffect::update(RenderEffectDropShadow* effect, const Matrix& transform)
 
     dropShadow->sigma = effect->sigma;
     dropShadow->scale = scale;
+    const auto sigma = dropShadow->sigma * dropShadow->scale;
+    dropShadow->invTwoSigmaSquared = tvg::zero(sigma) ? 0.0f : -0.5f / (sigma * sigma);
```
wenjieshen, line 153: Hi @Saususge,

I agree with Copilot’s diagnosis, but clamping the denominator seems safer than its suggestion:
```cpp
const auto safeSigma = std::max(sigma, FLOAT_EPSILON);
invTwoSigmaSquared = -0.5f / (safeSigma * safeSigma);
```
This keeps the center tap at 1, makes other taps effectively 0, and avoids both box blur and division by zero.

wenjieshen:
This PR showed no measurable performance change on Apple M1 OpenGL.
- Release/O3 builds with assertions disabled.
- The examples disable VSync via SDL_GL_SetSwapInterval(0).

src/renderer/gpu_engine/gl/tvgGlEffect.cpp
Saususge: good!

hermet (approved):
If there were no regression, good to go.

src/renderer/gpu_engine/gl/tvgGlEffect.cpp
hermet: The best is to skip the blur effect when the sigma is zero, not exception handling. See 151. line.

hermet:
@Saususge I've approved this since it saves some binary size anyway, but let's hold off on it for now. I'd rather introduce https://github.com/thorvg/thorvg/pull/4708 with a dummy usage. Anyway thanks for your contribution.

hermet:
@Saususge Hi, could you please rebase this and update the PR to run it again? The 16-byte dummy addition might be necessary. Sorry for the inconvenience. Thanks.

Saususge:
@hermet Sure! I pushed with rebase.

src/renderer/gpu_engine/gl/tvgGlShaderSrc.cpp
```diff
@@ -1079,16 +1083,14 @@ layout(std140) uniform Viewport {
 in vec2 vUV;
 out vec4 FragColor;
 
-float gaussian(float x, float sigma) {
-    float exponent = -x * x / (2.0 * sigma * sigma);
-    return exp(exponent) / (sqrt(2.0 * 3.141592) * sigma);
+float gaussian(float x) {
+    return exp(x * x * uGaussian.invTwoSigmaSquared);
 }
```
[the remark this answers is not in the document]
Saususge: I accept this. It makes sense

hermet:
Thanks for your effort, but when I tested it multiple times, I didn't see any improvement in performance, neither binary size does't helpful. Let's close this. Thanks.

Issue #4700, opened 2026-08-24
Another GL_CHECK issue


xiaozhuai:
https://github.com/thorvg/thorvg/blob/3b5de7e3de68c13ad35a112642b4fbca8b1e9021/src/renderer/gpu_engine/gl/tvgGlStateCache.cpp#L56-L57

This issue was previously reported and fixed in #4527.

The root cause is that the syntax `#define GL_CHECK(stmt) stmt; assert(glGetError() == GL_NO_ERROR);` is very risky.
Even if we add curly braces now to fix it, we still cannot prevent the same problem from occurring again and again in the future.

I strongly recommend changing it to `#define GL_CHECK(stmt) do { stmt; assert(glGetError() == GL_NO_ERROR); } while(0)`.
The use of the `do {} while(0)` pattern has been proven to be a safer approach in many projects.

If the maintainer approves this change, I will submit a PR to resolve this issue at its root.

wenjieshen:
Hi @xiaozhuai, I apologize for the inconvenience. I believe we have sufficient time to validate the larger change now. Could you please submit a PR? Thank you for your contribution.

xiaozhuai:
> Could you please submit a PR? 

@wenjieshen Sure~
I'll submit it tomorrow.

hermet:
@wenjieshen otherwise, we can remove all GL_CHECK(). I can assume it provides very little practical help with debugging.

xiaozhuai:
@wenjieshen @hermet Please review #4704

wenjieshen:
> [@wenjieshen](https://github.com/wenjieshen) otherwise, we can remove all GL_CHECK(). I can assume it provides very little practical help with debugging.

Hi @hermet, the macro is useful, but wrapping nearly every GL call adds too much noise. I believe we should keep the utility and reduce its use to selected places where the check provides meaningful value. Thank you.

PR #4704, opened 2026-08-25
gl_engine: make GL_CHECK macro statement-safe

xiaozhuai:
* Wrap `GL_CHECK` in `do { ... } while(0)` so it behaves as a single statement in if/else and other control-flow contexts. It's purely a preprocessor/grammar trick to make a multi-statement macro expand into a single statement (safe inside if/else), with zero runtime cost.
* Include `<cassert>` for the debug assertion
* Move `blockIndex` declaration out of `GL_CHECK` to avoid scoping issues after the macro is wrapped
* Remove a outdated comment in `tvgGlRenderTask.cpp`

Related #4527 #4700

Note that this change will only narrow the scope.
Therefore, if it compiles, there is generally no problem with it.

The reason I also change `GL_CHECK` in non-debug is that a statement itself can also be a multi-statement.

### DIFF
--- a/src/renderer/gpu_engine/gl/tvgGl.h
+++ b/src/renderer/gpu_engine/gl/tvgGl.h
@@ -26,7 +26,7 @@
 #ifdef __EMSCRIPTEN__
     #include <GLES3/gl3.h>
     #include <emscripten/html5_webgl.h>
-    #define GL_CHECK(stmt) stmt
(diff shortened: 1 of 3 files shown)


xiaozhuai:
Hi @hermet 
Thank you for merging this PR and inviting me to join thorvg. I'd really appreciate it!
I will make more contributions in the future.

Issue #4557, opened 2026-07-14
wgpu: clear buffer spec doenst match


hermet:
Canvas buffer should be preserved if the canvas draw doenst have a background or with a draw(false) call.

## problem (wgpu)

<img width="206" height="177" alt="Image" src="https://github.com/user-attachments/assets/f80f1653-d076-47f2-91dd-1eb307869ce8" />

## expected (gl)

<img width="220" height="192" alt="Image" src="https://github.com/user-attachments/assets/3ae08737-a82e-4076-a679-91d78dacaade" />

## Sample
```
struct UserExample : tvgexam::Example
{
    tvg::Shape* rect = nullptr;
    uint32_t startTime = 0;
    bool first = true;

    static tvg::Matrix scaleAround(float scale)
    {
        const auto offset = 256.0f * (1.0f - scale);
        return {scale, 0.0f, offset, 0.0f, scale, offset, 0.0f, 0.0f, 1.0f};
    }

    bool content(tvg::Canvas* canvas, uint32_t w, uint32_t h) override
    {
        //Rectangle whose ONLY paint is a linear gradient stroke
        rect = tvg::Shape::gen();
        rect->appendRect(206, 206, 100, 100);

        auto gradient = tvg::LinearGradient::gen();
        gradient->linear(206, 256, 306, 256);
        const tvg::Fill::ColorStop stops[] = {
            {0.0f, 255, 0, 0, 255},
            {0.5f, 0, 255, 0, 255},
            {1.0f, 0, 0, 255, 255}
        };
        gradient->colorStops(stops, 3);
        rect->strokeWidth(10);
        rect->strokeFill(gradient);
        rect->strokeCap(tvg::StrokeCap::Round);
        rect->strokeJoin(tvg::StrokeJoin::Round);

        //FIRST render happens at scale 0 (the trigger)
        const auto transform = scaleAround(0.0f);
        rect->transform(transform);
        canvas->add(rect);
        return true;
    }


    bool update(tvg::Canvas* canvas, uint32_t elapsed) override
    {
        if (first) {
            first = false;
            startTime = elapsed;
        }

        elapsed -= startTime;
        const auto k = elapsed < 1000 ? elapsed * 0.001f : 1.0f;   //scale 0 -> 100% over 1s, then holds

        const auto transform = scaleAround(k);
        rect->transform(transform);

        canvas->update();

        return true;
    }
};
```

wenjieshen:
Hi @hermet ,

[...]

PR #4720, opened 2026-08-26
gpu_engine/wg: honor canvas clear requests 

hermet:
Track clear requests until synchronization and clear the destination attachment before blitting the rendered frame.

Preserve existing target contents when clearing is disabled and blend the new frame using premultiplied alpha.

issue: #4557


wenjieshen:
Related issues:
https://github.com/thorvg/thorvg/issues/4724
https://github.com/thorvg/thorvg/issues/4557

Issue #4726, opened 2026-08-26
gradient transform issue


thatoneham:
been using godot 4.6.2, found that a svg wont render correctly, checked thorvg view and found latest still has the issue.

currently
<img width="615" height="625" alt="Image" src="https://github.com/user-attachments/assets/d360e16e-a5cd-4308-a5c8-d8a05c20e09f" />

intended result

<img width="627" height="636" alt="Image" src="https://github.com/user-attachments/assets/a7011a29-3e30-426c-815c-12cacabd1829" />

found out the issue is the `gradientTransform` inside the `linearGradient`
```
<linearGradient gradientTransform="matrix(0.0, 0.0, 0.0, 0.0, 0.0, 0.0)" gradientUnits="userSpaceOnUse" id="gradient0" spreadMethod="pad" x1="-819.2" x2="819.2">
 ....
```
to
```
<linearGradient  gradientUnits="userSpaceOnUse" id="gradient0" spreadMethod="pad" x1="-819.2" x2="819.2">
  ....
```

here is the raw svg.
<img width="146" height="501" alt="Image" src="https://github.com/user-attachments/assets/39f41ce4-db85-4364-8f93-801c9444c140" />

hopefully this is helpful enough. second time creating a issue

PR #4727, opened 2026-08-27
svg: fix a shape lost by an undrawable gradient

JSUYA:
Drop the gradient when it has no stops or when its matrix cannot be inverted, so only that paint is skipped and the shape stays in the scene.

Related issue: https://github.com/thorvg/thorvg/issues/4726

### DIFF
--- a/src/loaders/svg/tvgSvgBuilder.cpp
+++ b/src/loaders/svg/tvgSvgBuilder.cpp
@@ -96,11 +96,16 @@ static LinearGradient* _applyLinearGradientProperty(SvgStyleGradient* g, const B
         else finalTransform = m;
         fillGrad->linear(g->linear.x1, g->linear.y1, g->linear.x2, g->linear.y2);
     }
+
+    Matrix inv;
+    if (g->stops.count == 0 || !inverse(&finalTransform, &inv)) {
+        delete(fillGrad);
+        return nullptr;
+    }
+
     fillGrad->spread(g->spread);
 
     //Update the stops
-    if (g->stops.count == 0) return fillGrad;
-
     stops = tvg::malloc<Fill::ColorStop>(g->stops.count * sizeof(Fill::ColorStop));
     auto prevOffset = 0.0f;
     for (uint32_t i = 0; i < g->stops.count; ++i) {
@@ -146,11 +151,16 @@ static RadialGradient* _applyRadialGradientProperty(SvgStyleGradient* g, const B
         else finalTransform = m;
         fillGrad->radial(g->radial.cx, g->radial.cy, g->radial.r, g->radial.fx, g->radial.fy, g->radial.fr);
     }
+
+    Matrix inv;
+    if (g->stops.count == 0 || !inverse(&finalTransform, &inv)) {
+        delete(fillGrad);
+        return nullptr;
+    }
+
     fillGrad->spread(g->spread);
 
     //Update the stops
-    if (g->stops.count == 0) return fillGrad;
-
     stops = tvg::malloc<Fill::ColorStop>(g->stops.count * sizeof(Fill::ColorStop));
     auto prevOffset = 0.0f;
     for (uint32_t i = 0; i < g->stops.count; ++i) {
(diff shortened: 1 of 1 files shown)


hermet:
@JSUYA does this the svg spec?

JSUYA:
> @JSUYA does this the svg spec?

As far as I know that the actual specification does not draw paint where an incorrect matrix is ​​applied.
I guess there might be more complex cases, so I intended to handle them with minimal modifications, but I will analyze the issue further and update the PR.

hermet:
> specification does not draw paint where an incorrect matrix is ​​applied.

That's quite controversial how do we define an "incorrect matrix" If the SVG itself is invalid, it's best for the user to fix their asset.

JSUYA:
> > specification does not draw paint where an incorrect matrix is ​​applied.
> 
> That's quite controversial how do we define an "incorrect matrix" If the SVG itself is invalid, it's best for the user to fix their asset.

In the issue, the problematic syntax is `matrix(0.0, 0.0, 0.0, 0.0, 0.0, 0.0)`, and the SVG Spec defines this as: "If a transform function causes the [current transformation matrix](https://www.w3.org/TR/css-transforms-1/#current-transformation-matrix) of an object to be non-invertible, the object and its content do not get displayed."
https://www.w3.org/TR/css-transforms-1/#transform-function-lists

Of course, I am not sure to what extent support for invalid SVGs should be provided, but since the handling of non-invertible matrices is covered in the spec, I believe this part can be supported.

hermet:
> > > specification does not draw paint where an incorrect matrix is ​​applied.
> > 
> > 
> > That's quite controversial how do we define an "incorrect matrix" If the SVG itself is invalid, it's best for the user to fix their asset.
> 
> In the issue, the problematic syntax is `matrix(0.0, 0.0, 0.0, 0.0, 0.0, 0.0)`, and the SVG Spec defines this as: "If a transform function causes the [current transformation matrix](https://www.w3.org/TR/css-transforms-1/#current-transformation-matrix) of an object to be non-invertible, the object and its content do not get displayed." https://www.w3.org/TR/css-transforms-1/#transform-function-lists
> 
> Of course, I am not sure to what extent support for invalid SVGs should be provided, but since the handling of non-invertible matrices is covered in the spec, I believe this part can be supported.

@JSUYA  In that case, thorvg wouldnt(shouldn't) draw any elements with that matrix because it's empty (zero scale). I think svg loader shouldn't handle it individually.

hermet (changes requested):
This looks an ad-hoc approach.

JSUYA:
> 
> @JSUYA In that case, thorvg wouldnt(shouldn't) draw any elements with that matrix because it's empty (zero scale). I think svg loader shouldn't handle it individually.

According to spec document, for a gradient, the spec drops only the paint layer that gradient produces, the shape's other layers (e.g. stroke) must still be drawn (CSS Transforms 1 §10, SVG2 §13.2 "no paint is rendered for that layer"). Chrome's reference behave the same way.

Currently the SW engine looks like dismiss the shape when the gradient is invalid, so fill and stroke both disappear. This PR only keeps the loader from passing such a gradient to the engine.

Please let me know if there are any parts where I researched incorrectly.

hermet:
> According to spec document, for a gradient, the spec drops only the paint layer that gradient produces,

@JSUYA Is it right, only for gradient? Could you please share the detail spec link? This was what I see:

image

JSUYA:
> > According to spec document, for a gradient, the spec drops only the paint layer that gradient produces,
> 
> @JSUYA Is it right, only for gradient? Could you please share the detail spec link? This was what I see:

This is my understanding:
- According to the CSS Transforms 1 §10 specification (which you also mentioned), “objects” looks not be drawn.
- However, it appears that the behavior is different in Chrome or other libraries.
- Once the reason is found, it is necessary to verify whether this condition can vary depending on the provisions of [SVG2 §13.2](https://www.w3.org/TR/SVG2/painting.html#SpecifyingPaint) that I mentioned.
- (I wanted to find the exact meaning that "objects" targets, but there is no document with an exact correspondence, and it seems to be defined semantically or by referencing other documents.)
- “If a paint server reference in a <paint> is invalid, and no fall-back value is given, no paint is rendered for that layer. This is changed from SVG 1.1 behavior where the document is in error if a paint server reference is invalid and there is no fallback color specified.”
- Based on Chrome's commit and WPT svg file, it looks that stroke elements should be drawn as in the current implementation.

  https://chromium-review.googlesource.com/c/chromium/src/+/5272857
```
A non-invertible gradient/patternTransform make the paintserver invalid

The gradient/patternTransform properties follow the rules of the CSS
transforms spec:

 "If a transform function causes the current transformation matrix of an
  object to be non-invertible, the object and its content do not get
  displayed." [1]

Check if the relevant transform is non-invertible and flag the paintserver
as invalid if it is.
```
(https://wpt.fyi/results/svg/pservers/reftests?label=experimental&label=master&aligned&q=gradient-transform)

- According to this commit, it appears to block only invalid gradients.

- In actual WPT test cases as well, when multiple fills(fallback sentence) are provided, only invalid gradients are ignored.
- (https://github.com/web-platform-tests/wpt/blob/master/svg/pservers/reftests/gradient-transform-01.svg)
- And, in the case below as well, Chrome will draw a green border box.
```
<linearGradient xmlns="http://www.w3.org/2000/svg" id="matrix0" gradientTransform="matrix(0, 0, 0, 0, 0, 0)">
    <stop offset="0" stop-color="yellow"/>
    <stop offset="1" stop-color="red"/>
  </linearGradient>
<rect xmlns="http://www.w3.org/2000/svg" x="50" width="50" height="100" fill="url(#matrix0)" stroke="green"/>
```
Therefore, overall, I thought that this applies only to gradients in our case, and I believe that the svg loader can block this.

Issue #4735, opened 2026-08-28
Pixel-Inspector Mismatch (v1.1.1 ~ main)


Nor-s:
| Target | First bad commit | Pull request | Direct good parent | First-bad diff ratio / pixels |
| --- | --- | --- | --- | ---: |
| `basic/lottie/heart_fill.json` | [`4a089b8b`](https://github.com/thorvg/thorvg/commit/4a089b8b3bc576c147846b7efc4f174fea7e15a6) | [#4711](https://github.com/thorvg/thorvg/pull/4711) | [`68df12e5`](https://github.com/thorvg/thorvg/commit/68df12e5523a9694e8ce19127e187a795e4cd5d2) | `0.0482125` / `146,412` |
| `lottie/18580.json` | [`4a089b8b`](https://github.com/thorvg/thorvg/commit/4a089b8b3bc576c147846b7efc4f174fea7e15a6) | [#4711](https://github.com/thorvg/thorvg/pull/4711) | [`68df12e5`](https://github.com/thorvg/thorvg/commit/68df12e5523a9694e8ce19127e187a795e4cd5d2) | `0.263494` / `104,086` |
| `svg/issue-4255-mix-blend-mode.svg` | [`4c79c84c`](https://github.com/thorvg/thorvg/commit/4c79c84cd155807ca791db33e8d663cd5f2458ae) | [#4719](https://github.com/thorvg/thorvg/pull/4719) | [`3473e541`](https://github.com/thorvg/thorvg/commit/3473e541f8c351a3759428383130adacb00fb259) | `0.279493` / `10,568` |
| `draw_test/blending` | [`4c79c84c`](https://github.com/thorvg/thorvg/commit/4c79c84cd155807ca791db33e8d663cd5f2458ae) | [#4719](https://github.com/thorvg/thorvg/pull/4719) | [`3473e541`](https://github.com/thorvg/thorvg/commit/3473e541f8c351a3759428383130adacb00fb259) | `0.201711` / `290,293` |
| `draw_test/issue_4190_nonseparable_blend` | [`4c79c84c`](https://github.com/thorvg/thorvg/commit/4c79c84cd155807ca791db33e8d663cd5f2458ae) | [#4719](https://github.com/thorvg/thorvg/pull/4719) | [`3473e541`](https://github.com/thorvg/thorvg/commit/3473e541f8c351a3759428383130adacb00fb259) | `0.169157` / `143,420` |

## 4711(WebGPU Texture Update Regression)

[...]

PR #4737, opened 2026-08-28
renderer: fixed a wgpu regression during main dev.

hermet:
Mark relocated paints for color updates because changing their parent can change inherited opacity. This prevents renderers from retaining stale effective opacity when paints are reused.

issue: #4735


Nor-s (approved):
The regression issue has been fixed with this patch. Thank you.

hermet:
https://github.com/thorvg/thorvg/pull/4740

PR #4738, opened 2026-08-28
meson: fix simd_type detection for aarch64 cpus

LonerDan:
`host_machine.cpu()` returns specific cpu model (for example `cortex-a72`), which would never match the condition. 
This replaces it with the `cpu_family()`, as is used in the conditions above, to match aarch64. 
Note: `cpu_family` values are defined [here](https://mesonbuild.com/Reference-tables.html#cpu-families)


hermet (approved):

hermet:
@LonerDan thanks!

PR #4746, opened 2026-08-30
svg: support rgba colors

hermet:
- Parse rgba() paint values and apply their alpha to fills and strokes.
- Store element opacity in the shared RGBA color representation.
- Reject malformed hex, rgb, and rgba color values.

fix: https://github.com/thorvg/thorvg.test-assets/blob/main/performance/paris-30k.svg


JSUYA (commented):
Merging opacity into color.a breaks CSS inheritance semantics: color is an inherited property while opacity is non-inherited. Since they now share one field, _styleInherit() (child->color = parent->color;, tvgSvgLoader.cpp:3088) overwrites each child's own
opacity with its parent's value.

Element opacity and paint alpha also compose differently (group-level vs per-paint), so the loader must keep them as separate values and combine them at build time.

```html
<svg xmlns="http://www.w3.org/2000/svg" width="320" height="100" viewBox="0 0 320 100">
  <rect width="320" height="100" fill="#ffffff"/>
  <!-- A (reference): fill-opacity 0.5 - not affected by this PR -->
  <rect x="10"  y="20" width="60" height="60" fill="#ff0000" fill-opacity="0.5"/>
  <!-- B: opacity 0.5 - must look identical to A -->
  <rect x="110" y="20" width="60" height="60" fill="#ff0000" opacity="0.5"/>
  <!-- C: color + opacity 0.5 - must also look identical to A -->
  <g color="#ff0000" opacity="0.5">
    <rect x="210" y="20" width="60" height="60" fill="currentColor"/>
  </g>
</svg>
```
chrome
Image

current
image

this pr
Screenshot from 2026-08-31 10-43-36

hermet:
> Merging opacity into color.a breaks CSS inheritance semantics: color is an inherited property while opacity is non-inherited. Since they now share one field, _styleInherit() (child->color = parent->color;, tvgSvgLoader.cpp:3088) overwrites each child's own opacity with its parent's value.
> 
> Element opacity and paint alpha also compose differently (group-level vs per-paint), so the loader must keep them as separate values and combine them at build time.
> 
> ```
@JSUYA updated, thanks.

Issue #4688, opened 2026-08-20
cpu: repeat-gradient anti-aliasing ignores transform


Nor-s:
<img width="800" height="auto" alt="Image" src="https://github.com/user-attachments/assets/600d6d23-b02f-4d3c-9631-4e0d30e8dd9e" />

1. A radius of 2048 scaled by `RadialGradient::transform()` with a factor of `0.0625`
2. A radius of 2048 scaled together with the shape by `Shape::transform()` with a factor of `0.0625`
3. A radius of 128 rendered directly without scaling (reference)

```cpp
struct UserExample : tvgexam::Example
{
    bool content(tvg::Canvas* canvas, uint32_t w, uint32_t h) override
    {
        auto bg = tvg::Shape::gen();
        bg->appendRect(0, 0, w, h);
        bg->fill(32, 32, 32);
        canvas->add(bg);

        tvg::Fill::ColorStop colors[4];
        colors[0] = {0.0f, 127, 39, 255, 255};
        colors[1] = {0.33f, 159, 112, 253, 255};
        colors[2] = {0.66f, 253, 191, 96, 255};
        colors[3] = {1.0f, 255, 137, 17, 255};

        // Only the fill coordinate space is scaled; the shape remains 512 x 512.
        auto left = tvg::Shape::gen();
        left->appendRect(0, 0, 512, 512);
        left->translate(16, 16);

        auto transformedFill = tvg::RadialGradient::gen();
        transformedFill->radial(256, 256, 2048, 256, 256, 0);
        transformedFill->transform({0.0625f, 0, 240, 0, 0.0625f, 240, 0, 0, 1});
        transformedFill->colorStops(colors, 4);
        transformedFill->spread(tvg::FillSpread::Repeat);
        left->fill(transformedFill);
        canvas->add(left);

        // The shape transform scales both its 8192 x 8192 geometry and radius 2048 fill.
        auto middle = tvg::Shape::gen();
        middle->appendRect(0, 0, 8192, 8192);
        middle->transform({0.0625f, 0, 544, 0, 0.0625f, 16, 0, 0, 1});

        auto shapeFill = tvg::RadialGradient::gen();
        shapeFill->radial(4096, 4096, 2048, 4096, 4096, 0);
        shapeFill->colorStops(colors, 4);
        shapeFill->spread(tvg::FillSpread::Repeat);
        middle->fill(shapeFill);
        canvas->add(middle);

        // Reference rendering uses the final dimensions without any transform.
[...]

PR #4750, opened 2026-08-31
cpu_engine: sync repeat-gradient anti-aliasing with transform

alpakaDurumi:
## 1. resolve root cause

`_estimateAAMargin()` computed the boundary anti-aliasing margin from the gradient's raw, untransformed radius or length. Once a transform was applied, the actual on-screen extent no longer matched that raw value, so the margin was estimated against the wrong scale.

Added `SwFill::extent`, the on-screen span of one full color table cycle, and made `_estimateAAMargin()` use it. `_prepareLinear()` derives it from the already inverse-transformed gradient direction, which stays exact under any invertible transform. `_prepareRadial()` takes the semi-minor axis of the ellipse the transform maps the gradient circle to, since that direction packs the table tightest and needs the widest margin.

image

## 2. sync LUT (`ctable`) with change of extent

Runtime extent changes require only the AA margin portion of the color table to be updated, without rebuilding the entire table.

A scale transform changes the extent, which shifts the AA margin range of the color table. `_updateColorTable()` only reacts to `RenderUpdateFlag::Gradient` or `RenderUpdateFlag::GradientStroke`, so the margin was left stale. `_updateAAMargin()` now runs when the extent of a repeat gradient changes and rebuilds the `MARGIN_MAX` entries at each end. The color stop walk is factored out into `_genColorTable()` with a `[begin, end)` range for that.

`_updateColorTable()` and `_updateAAMargin()` both compute the margin bounds from the color stop offsets in `_applyRepeatAA()`, since the partial rebuild has no full walk to take them from. The entry count of the first segment was also one short, as the walk reported its last index rather than the number of entries.

Also, the last entry of ctable is now pinned to the last color stop, which was previously guaranteed only when the spread was not Repeat.

> `MARGIN_MAX` holds the AA margin upper bound that `_adjustAAMargin()` previously hardcoded.

issue: #4688

### DIFF
--- a/src/renderer/cpu_engine/tvgSwCommon.h
+++ b/src/renderer/cpu_engine/tvgSwCommon.h
@@ -187,6 +187,7 @@ struct SwFill
     };
 
     uint32_t ctable[SW_COLOR_TABLE];
+    float extent;
     FillSpread spread;
 
     bool solid = false; //solid color fill with the last color from colorStops
--- a/src/renderer/cpu_engine/tvgSwFill.cpp
+++ b/src/renderer/cpu_engine/tvgSwFill.cpp
@@ -30,6 +30,7 @@
 #define RADIAL_A_THRESHOLD 0.0005f
 #define FIXPT_BITS 8
 #define FIXPT_SIZE (1<<FIXPT_BITS)
+#define MARGIN_MAX 40 // The maximum number of color table entries for the anti-aliasing (AA) margin at both ends.
 
 /*
  * quadratic equation with the following coefficients (rx and ry defined in the _calculateCoefficients()):
@@ -61,29 +62,20 @@ static void _calculateCoefficients(const SwFill* fill, uint32_t x, uint32_t y, f
 }
 
 
-static uint32_t _estimateAAMargin(const Fill* fdata)
+static uint32_t _estimateAAMargin(const SwFill* fill)
 {
     constexpr float marginScalingFactor = 800.0f;
-
-    if (fdata->type() == Type::RadialGradient) {
-        auto radius = CONST_RADIAL(fdata)->r;
-        return tvg::zero(radius) ? 0 : static_cast<uint32_t>(marginScalingFactor / radius);
-    } else {
-        auto grad = CONST_LINEAR(fdata);
-        auto len = length(grad->p1, grad->p2);
-        return tvg::zero(len) ? 0 : static_cast<uint32_t>(marginScalingFactor / len);
-    }
+    return tvg::zero(fill->extent) ? 0 : static_cast<uint32_t>(marginScalingFactor / fill->extent);
 }
 
 
-static void _adjustAAMargin(uint32_t& iMargin, uint32_t index)
+static void _adjustAAMargin(uint32_t& iMargin, uint32_t count)
 {
     constexpr float threshold = 0.1f;
-    constexpr uint32_t iMarginMax = 40;
 
-    auto iThreshold = static_cast<uint32_t>(index * threshold);
+    auto iThreshold = static_cast<uint32_t>(count * threshold);
     if (iMargin > iThreshold) iMargin = iThreshold;
-    if (iMargin > iMarginMax) iMargin = iMarginMax;
+    if (iMargin > MARGIN_MAX) iMargin = MARGIN_MAX;
 }
 
 
@@ -121,14 +113,25 @@ static void _applyAA(SwFill* fill, uint32_t begin, uint32_t end)
 }
 
 
-static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface* surface, uint8_t opacity)
+static void _applyRepeatAA(SwFill* fill, const Fill::ColorStop* colors, uint32_t cnt)
 {
-    if (fill->solid) return true;
+    if (cnt == 1) return;
 
-    const Fill::ColorStop* colors;
-    auto cnt = fdata->colorStops(&colors);
-    if (cnt == 0 || !colors) return false;
+    // Calculate count that satisfies (i + 0.5) / SW_COLOR_TABLE < offset
+    auto countBegin = static_cast<uint32_t>(max(0.0f, floor((colors + 1)->offset * SW_COLOR_TABLE - 0.5f) + 1.0f));
+    auto countEnd = static_cast<uint32_t>(max(0.0f, floor((1.0f - (colors + cnt - 2)->offset) * SW_COLOR_TABLE - 0.5f) + 1.0f));
 
+    auto iAABegin = _estimateAAMargin(fill);
+    _adjustAAMargin(iAABegin, countBegin);
+    auto iAAEnd = iAABegin;
+    _adjustAAMargin(iAAEnd, countEnd);
+    _applyAA(fill, iAABegin, iAAEnd);
+}
+
+
+// Generate the ctable entries over the [begin, end) range
+static void _genColorTable(SwFill* fill, const Fill::ColorStop* colors, uint32_t cnt, const SwSurface* surface, uint8_t opacity, uint32_t begin, uint32_t end)
+{
     auto pColors = colors;
 
     auto a = MULTIPLY(pColors->a, opacity);
@@ -139,28 +142,17 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface* 
     auto b = pColors->b;
     auto rgba = surface->join(r, g, b, a);
     auto inc = 1.0f / static_cast<float>(SW_COLOR_TABLE);
-    auto pos = 1.5f * inc;
-    uint32_t i = 0;
+    auto pos = (begin + 0.5f) * inc;
+    auto i = begin;
 
-    //If repeat is true, anti-aliasing must be applied between the last and the first colors.
-    auto repeat = fill->spread == FillSpread::Repeat;
-    uint32_t iAABegin = repeat ? _estimateAAMargin(fdata) : 0;
-    uint32_t iAAEnd = 0;
+    auto padColor = ALPHA_BLEND(rgba | 0xff000000, a);
 
-    fill->ctable[i++] = ALPHA_BLEND(rgba | 0xff000000, a);
-
-    while (pos <= pColors->offset) {
-        fill->ctable[i] = fill->ctable[i - 1];
-        ++i;
+    while (i < end && (i == 0 || pos <= pColors->offset)) {
+        fill->ctable[i++] = padColor;
         pos += inc;
     }
 
     for (uint32_t j = 0; j < cnt - 1; ++j) {
-        if (repeat && j == cnt - 2 && iAAEnd == 0) {
-            iAAEnd = iAABegin;
-            _adjustAAMargin(iAAEnd, SW_COLOR_TABLE - i);
-        }
-
         auto curr = colors + j;
         auto next = curr + 1;
         auto div = next->offset - curr->offset;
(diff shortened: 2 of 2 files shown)


hermet:
@Nor-s please confirm with the performance check **thorvg.benchmark (linear, radial + rotation)**. thanks.

Nor-s:
@alpakaDurumi 

It seems that the current approach does not work correctly with non-uniform scale. Could you please verify this?

image



<details> <summary> code </summary>

```cpp
const auto ANGLE_30 = std::acos(-1.0f) / 6.0f;
const auto COSINE_30 = std::cos(ANGLE_30);
const auto SINE_30 = std::sin(ANGLE_30);

constexpr auto PANEL_SIZE = 512.0f;
constexpr auto PADDING = 16.0f;

static tvg::Matrix _repeatTransform(float x, float y)
{
    constexpr auto SCALE_X = 1.0f / 16.0f;
    constexpr auto SCALE_Y = 3.0f / 4.0f;

    return {COSINE_30 * SCALE_X, -SINE_30 * SCALE_Y, x,
            SINE_30 * SCALE_X, COSINE_30 * SCALE_Y, y,
            0, 0, 1};
}

struct UserExample : tvgexam::Example
{
    {
        auto bg = tvg::Shape::gen();
        bg->appendRect(0, 0, w, h);
        bg->fill(32, 32, 32);
        canvas->add(bg);

        tvg::Fill::ColorStop colors[4];
        colors[0] = {0.0f, 127, 39, 255, 255};
        colors[1] = {0.33f, 159, 112, 253, 255};
        colors[2] = {0.66f, 253, 191, 96, 255};
        colors[3] = {1.0f, 255, 137, 17, 255};

        // Deliberately use M = T * R(30 degrees) * S(1/16, 3/4). Scaling local
        // x turns the 2048-unit Repeat period into 128px; the larger y scale
        // exposes AA that replaces this directional footprint with one scalar.
        // Rotation preserves the 128px period but makes each seam cross the
        // pixel grid diagonally, so missing AA is visible as stair steps.

        // Apply M to the fill only. The shape geometry remains 512px square.
        auto left = tvg::Shape::gen();
        left->appendRect(0, 0, 512, 512);
        left->translate(16, 16);

        auto transformedFill = tvg::LinearGradient::gen();
        transformedFill->linear(-1024, 0, 1024, 0);
        transformedFill->transform(_repeatTransform(256, 256));
        transformedFill->colorStops(colors, 4);
        transformedFill->spread(tvg::FillSpread::Repeat);
        left->fill(transformedFill);
        canvas->add(left);

        // Apply the same M to the shape, which transforms its geometry and fill.
        // The clip keeps only the middle 512px result visible.
        auto middle = tvg::Shape::gen();
        middle->appendRect(-8192, -8192, 16384, 16384);
        auto middleX = PADDING * 2.0f + PANEL_SIZE;
        auto middleY = PADDING;
        middle->transform(_repeatTransform(middleX + PANEL_SIZE * 0.5f, middleY + PANEL_SIZE * 0.5f));

        auto shapeFill = tvg::LinearGradient::gen();
        shapeFill->linear(-1024, 0, 1024, 0);
        shapeFill->colorStops(colors, 4);
        shapeFill->spread(tvg::FillSpread::Repeat);
        middle->fill(shapeFill);

        clip->appendRect(middleX, middleY, PANEL_SIZE, PANEL_SIZE);
        middle->clip(clip);
        canvas->add(middle);

        // Draw the equivalent final-space reference without a transform: a
        // 128px period at 30 degrees. All three AA seams must look identical.
        auto right = tvg::Shape::gen();
        right->appendRect(0, 0, 512, 512);
        right->translate(1072, 16);

        auto directFill = tvg::LinearGradient::gen();
        directFill->linear(256 - COSINE_30 * 64, 256 - SINE_30 * 64,
                           256 + COSINE_30 * 64, 256 + SINE_30 * 64);
        directFill->colorStops(colors, 4);
        directFill->spread(tvg::FillSpread::Repeat);
        right->fill(directFill);
        canvas->add(right);

        return true;
    }
};

{
    return tvgexam::main(new UserExample, argc, argv, false, 1600, 544, 0);
}

```

 </details>

Nor-s:
@hermet 

I tested the performance using thorvg.benchmark and observed a difference of less than 1%.  (with repeat)

However, when I configured thorvg.benchmark to use a size close to 1024 pixels(32x32) and 
set threads=0, performance decreased by approximately 5%.

Thank you.

alpakaDurumi:
@Nor-s Thanks for the report. I fixed it to work correctly with non-uniform scale, and removed the Performance section as it became outdated.

- `_prepareLinear()`: now reflects the gradient (amount of change) accurately.
- `_prepareRadial()`: now derives the semi-minor axis (the smaller singular value of the transform) and uses it as the extent.

Could you review it again? Thanks.

src/renderer/cpu_engine/tvgSwCommon.h
```diff
@@ -187,6 +188,11 @@ struct SwFill
     };
 
     uint32_t ctable[SW_COLOR_TABLE];
+    uint32_t marginCache[SW_MARGIN_MAX * 2];
```
Nor-s: Instead of caching the original edge entries, 
could we regenerate only the affected pre-AA LUT samples using the existing ColorStop-to-ColorTable generation logic?

alpakaDurumi: @Nor-s Thanks for the review. Instead of caching, the margin entries are now
regenerated using the color table generation logic. Could you review it again? Thank you.

src/renderer/cpu_engine/tvgSwCommon.h
```diff
@@ -187,6 +188,10 @@ struct SwFill
     };
 
     uint32_t ctable[SW_COLOR_TABLE];
+    uint32_t countSegmentBegin;
+    uint32_t countSegmentEnd;
```
Nor-s: Lastly, I think we should remove this cache. Since AA is currently applied only to repeat gradients, it would be better to keep the related state to a minimum.(countSegmentBegin/countSegmentEnd)

alpakaDurumi: @Nor-s Agreed, fixed that part as well. The margin bounds are now derived from the offsets directly. Thank you.

Nor-s (approved):
Image

| thorvg.benchmark | + %fps(vs main) |
|---|---:|
| linear/default | -1.24% |
| linear/rotation | +0.21% |
| radial/default | +1.50% |
| radial/rotation | -0.72% |

@hermet 
The thorvg.benchmark results show differences within 
the noise level, and anti-aliasing works correctly.

src/renderer/cpu_engine/tvgSwFill.cpp
```diff
@@ -773,15 +811,18 @@ void fillLinear(const SwSurface* surface, const SwFill* fill, uint32_t* dst, uin
 
 bool fillGenColorTable(SwFill* fill, const Fill* fdata, const Matrix& transform, SwSurface* surface, uint8_t opacity, bool ctable)
 {
+    bool extentChanged = false;
```
hermet: bool -> auto

src/renderer/cpu_engine/tvgSwFill.cpp
```diff
@@ -231,11 +249,19 @@ bool _prepareLinear(SwFill* fill, const LinearGradient* linear, const Matrix& pT
...
+        *extentChanged = true;
+        fill->extent = ext;
+    }
+
     return true;
 }
 
 
-bool _prepareRadial(SwFill* fill, const RadialGradient* radial, const Matrix& pTransform)
+bool _prepareRadial(SwFill* fill, const RadialGradient* radial, const Matrix& pTransform, bool* extentChanged)
```
hermet: bool* extentChanged -> bool& extentChanged

src/renderer/cpu_engine/tvgSwFill.cpp
```diff
@@ -182,25 +175,50 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface*
...
+    _genColorTable(fill, colors, cnt, surface, opacity, SW_COLOR_TABLE - MARGIN_MAX, SW_COLOR_TABLE);
+
+    _applyRepeatAA(fill, colors, cnt);
 
     return true;
 }
 
 
-bool _prepareLinear(SwFill* fill, const LinearGradient* linear, const Matrix& pTransform)
+bool _prepareLinear(SwFill* fill, const LinearGradient* linear, const Matrix& pTransform, bool* extentChanged)
```
hermet: bool* extentChanged -> bool& extentChanged

src/renderer/cpu_engine/tvgSwFill.cpp
```diff
@@ -182,25 +175,50 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface*
...
+    return true;
+}
+
+
+// Update only the ctable corresponding to the margin based on the updated extent
+static bool _updateAAMargin(SwFill* fill, const Fill* fdata, const SwSurface* surface, uint8_t opacity)
+{
+    const Fill::ColorStop* colors;
+    auto cnt = fdata->colorStops(&colors);
+    if (cnt == 0 || !colors) return false;
```
hermet: if (cnt == 0) reutrn false;

src/renderer/cpu_engine/tvgSwFill.cpp
```diff
@@ -182,25 +175,50 @@ static bool _updateColorTable(SwFill* fill, const Fill* fdata, const SwSurface*
...
+
+    _genColorTable(fill, colors, cnt, surface, opacity, 0, MARGIN_MAX);
+    _genColorTable(fill, colors, cnt, surface, opacity, SW_COLOR_TABLE - MARGIN_MAX, SW_COLOR_TABLE);
+
```
hermet: -- empty line.

src/renderer/cpu_engine/tvgSwFill.cpp
```diff
@@ -121,14 +113,26 @@ static void _applyAA(SwFill* fill, uint32_t begin, uint32_t end)
...
+static void _applyRepeatAA(SwFill* fill, const Fill::ColorStop* colors, uint32_t cnt)
 {
-    if (fill->solid) return true;
+    if (cnt == 1) return;
+
+    // Calculate count that satisfies (i + 0.5) / SW_COLOR_TABLE < offset
+    auto countBegin = static_cast<uint32_t>(max(0.0f, floor((colors + 1)->offset * SW_COLOR_TABLE - 0.5f) + 1.0f));
+    auto countEnd = static_cast<uint32_t>(max(0.0f, floor((1.0f - (colors + cnt - 2)->offset) * SW_COLOR_TABLE - 0.5f) + 1.0f));
+
+    uint32_t iAABegin = _estimateAAMargin(fill);
```
hermet: uint32_t -> auto

src/renderer/cpu_engine/tvgSwFill.cpp
```diff
@@ -121,14 +113,26 @@ static void _applyAA(SwFill* fill, uint32_t begin, uint32_t end)
...
+    _adjustAAMargin(iAABegin, countBegin);
+    uint32_t iAAEnd = iAABegin;
```
hermet: uint32_t -> auto

src/renderer/cpu_engine/tvgSwFill.cpp
```diff
@@ -121,14 +113,26 @@ static void _applyAA(SwFill* fill, uint32_t begin, uint32_t end)
...
+    _adjustAAMargin(iAAEnd, countEnd);
+
```
hermet: -- empty line

src/renderer/cpu_engine/tvgSwFill.cpp
```diff
@@ -30,6 +30,7 @@
 #define RADIAL_A_THRESHOLD 0.0005f
 #define FIXPT_BITS 8
 #define FIXPT_SIZE (1<<FIXPT_BITS)
+#define MARGIN_MAX 40
```
hermet: good to add a code comment MARGIN_MAX means here.

hermet (approved):
@alpakaDurumi thanks, please check minor nitpicks and squash 2 commits to 1.

alpakaDurumi:
@hermet Thank you for the thorough review. Squashed into a single commit.

Issue #4756, opened 2026-09-01
infra: Remove Redundant Build Tests


hermet:
Some of the simple build tests appear to duplicate coverage already provided by unit-tests.

Running redundant build tests increases CI time and resource usage without providing additional validation.

i.e) removal target:
<img width="1027" height="55" alt="Image" src="https://github.com/user-attachments/assets/356fe988-58ad-4a9a-80d0-f89777d1ca05" />

### Proposal

Review the current build tests and remove build tests when the same build configuration is already verified by other CI tests.

### Goal

Reduce redundant CI workloads and improve overall CI efficiency while maintaining sufficient build coverage.

Nor-s:
<img width="1609" height="219" alt="Image" src="https://github.com/user-attachments/assets/449cb766-26ce-40fb-849d-e67d3ef2f38e" />

We could also consider testing the CPU, WebGPU, and OpenGL backends in parallel using Valgrind.

hermet:
[...]

PR #4762, opened 2026-09-02
infra: Remove Redundant Build Tests

Nor-s:
### infra: streamline platform CI workflows

Remove obsolete submodule checkout options now that the repository
no longer contains Git submodules.

Drop redundant standalone build jobs on macOS, Ubuntu, and Windows.
Static configurations remain covered by platform build jobs, while
shared configurations continue to be exercised by unit tests. 

### infra: parallelize memory checks

Split CPU, GL, and WG unit tests into four parallel Valgrind
shards, and move ASAN tests to a dedicated Ubuntu job so both
checks can run concurrently.

Issue: #4756

### Result 

| Main (14m~16m) | Fetch (6m ~8m) |
|-|-|
|image | image|


hermet:
1. unify `unit-test` + `asan` -> `unit-test`
or
2. `asan (asan + valgrind)` -> `memcheck`

Nor-s:
@hermet 

Valgrind and ASan couldn't be integrated into a single binary:

* They cannot be run as a single binary.: **Valgrind crashes** when handling **ASan-generated data**.
* They need to run separately. (Slow)

One question is about the naming, since both use the same `unit_test` name.

Would `unit_test(asan)` and `unit_test(valgrind)` be better?

hermet:
> @hermet
> 
> Valgrind and ASan couldn't be integrated into a single binary:
> 
> * They cannot be run as a single binary.: **Valgrind crashes** when handling **ASan-generated data**.
> * They need to run separately. (Slow)
> 
> One question is about the naming, since both use the same `unit_test` name.
> 
> Would `unit_test(asan)` and `unit_test(valgrind)` be better?

@Nor-s sounds good.

hermet (approved):

PR #4775, opened 2026-09-07
cpu_engine: optimize rle for large outlines

Nor-s:
The rle rasterizer uses a fixed 16 KB cell pool per thread.
When an outline overflows it, rleRender() halves the band and
re-walks the whole outline once per resulting band,
so complex outlines were traversed hundreds of times per frame.

Grow the pool instead: double it on overflow (up to MAX_CELL_POOL_SIZE)
via realloc and retry the same band, falling back to band halving
only at the cap or when the allocation fails.

### Perf Review 

> [!NOTE]
> ` meson setup builddir --wipe  -Dthreads=false -Dloaders=all -Ddefault_library=shared`
> `.\builddir\src\Animation.exe -i .\res\lottie\monkey.json`

>
> `Windows11`, 
> `amd Ryzen 7600`, `64GB`, 
> `msvc 19.50.35728 "Microsoft (R) C/C++ Optimizing Compiler Version 19.50.35728 for x64`
> I couldn’t reproduce the improvement on macOS(M5 pro), wasm. 
> Although I observed a performance improvement in a headless environment, I did not see an FPS increase when rendering to a SDL window.

> For the other assets, I did not observe any significant difference. (big_donut.json, flying.json ... )

| thorvg.example | vs main (fps +%)| 
| - | - |
| Lottie | - |
| Animation/guitar | +15% |
| Animation/monkey  |  +30% | 
| Animation/yarn_loading | +40% |
| Animation/isometric | +140% |
| [Animation/test_20p.json](https://github.com/user-attachments/files/31923747/test_20p.json) | +450% |
| [Animation/test_40p.json](https://github.com/user-attachments/files/31923765/test_40p.json) | +608% |

> [!NOTE]
> thorvg.benchmark: MacOS, thorvg.benchmark main vs pr, threads=true

| thorvg.benchmark | vs main (fps +%) |
| ---|--- |
| stroke / rotation | +3.7 % |
| strokerect / rotation | +3.2 % |

### DIFF
--- a/src/renderer/cpu_engine/tvgSwRle.cpp
+++ b/src/renderer/cpu_engine/tvgSwRle.cpp
@@ -837,7 +837,19 @@ SwRle* rleRender(SwRle* rle, const SwOutline* outline, const RenderRegion& bbox,
             }
 
         reduce_bands:
-            /* render pool overflow: we will reduce the render band by half */
+            /* render pool overflow: grow the cell pool first (bounded), then retry the same band.
+               Halving the band instead re-walks the whole outline once per band. */
+            if (cellPool->size < MAX_CELL_POOL_SIZE) {
+                auto newSize = std::min(cellPool->size * 2, uint32_t(MAX_CELL_POOL_SIZE));
+                if (auto newBuffer = tvg::realloc<SwCell>(cellPool->buffer, newSize)) {
+                    cellPool->size = newSize;
+                    cellPool->buffer = newBuffer;
+                    rw.buffer = cellPool->buffer;
+                    rw.bufferSize = cellPool->size;
+                    continue;
+                }
+            }
+            /* pool is at its cap: reduce the render band by half */
             auto bottom = band->min;
             auto top = band->max;
             auto middle = bottom + ((top - bottom) >> 1);
--- a/src/renderer/cpu_engine/tvgSwCommon.h
+++ b/src/renderer/cpu_engine/tvgSwCommon.h
@@ -319,6 +319,7 @@ struct SwCompositor : RenderCompositor
 struct SwCellPool
 {
     #define DEFAULT_POOL_SIZE 16368
+    #define MAX_CELL_POOL_SIZE (DEFAULT_POOL_SIZE * 64)   //1 MB per thread, reached only by very complex outlines
 
     uint32_t size;
     SwCell* buffer;


hermet:
Item | Rate
-- | --
Monkey | +5%
Yarn Loading | --
Guitar | +5%
BigDonut | -20.92%
Flying | -4.27%
ManyStar | +2.36%
Samurai | -15.39%
Isometric | --
Guitar | +13.82%
test_20p | +808.75%
test_40p | +471.15%
Lottie | -4%

hermet:
@Nor-s could you please confirm the `Lottie` & `thorvg.test-assets/performance/big_donut` performance ?

Nor-s:
> @Nor-s could you please confirm the `Lottie` & `thorvg.test-assets/performance/big_donut` performance ?

The Lottie example goes through this path, but big_donut never hits it.
I could not reproduce the regression in my environment. 
Could you share test environment? (ex. os, compiler, threads)

hermet:
> > @Nor-s could you please confirm the `Lottie` & `thorvg.test-assets/performance/big_donut` performance ?
> 
> The Lottie example goes through this path, but big_donut never hits it.
> I could not reproduce the regression in my environment. 
> Could you share test environment? (ex. os, compiler, threads)

You dont need to. Please just confirm before/after in your machine. Thanks.

Nor-s:
@hermet 

Measured on macOS. Thank you.
(10000 tick, 3 time, mean)

| thorvg.example | vs before (fps +%) | 
| - | - |
| Lottie | +0.2 % |
| Animation/big_donut | -0.6 % |

src/renderer/cpu_engine/tvgSwRle.cpp
```diff
@@ -837,7 +837,19 @@ SwRle* rleRender(SwRle* rle, const SwOutline* outline, const RenderRegion& bbox,
             }
 
         reduce_bands:
-            /* render pool overflow: we will reduce the render band by half */
+            /* render pool overflow: grow the cell pool first (bounded), then retry the same band.
```
hermet, line 840: Regardless of whether this improves or hurts performance, If we've already determined the maximum memory allowed for the pool, it would be more efficient to allocate that memory upfront with that MAX rather than incremental way (why you do this?). Then, the real question seems to be what an appropriate maximum memory size should be.

hermet (changes requested):
The current maximum memory size could be problematic on resource-constrained devices. This needs to be addressed with a more logical approach.

hermet:
Reopen this, it is worth to investigate further.
issue: https://github.com/thorvg/thorvg/issues/4795

PR #4808, opened 2026-09-14
cpu_engine: treat single-stop gradients as solid

alpakaDurumi:
When a gradient has a single color stop, every color table entry is generated with the same color. Treating it as solid removes the need to build the table, and also improves performance in rasterization.

Added a check in `shapeGenFillColors()` to return early.

## Performance

The program below fills a 1024x1024 canvas with a single-stop linear gradient for 200 frames. On a Ryzen 7 9700X (Ubuntu 24.04.2, WSL2), `hyperfine -N -w 1 -r 10` shows an 81% fps increase.

<details>
<summary>code</summary>

```cpp
#include <thorvg.h>
#include <cstdlib>
#include <vector>

using namespace tvg;

{
    const uint32_t W = 1024, H = 1024;
    int frames = (argc > 1) ? atoi(argv[1]) : 200;

    Initializer::init();

    std::vector<uint32_t> buf(W * H);
    canvas->target(buf.data(), W, W, H, ColorSpace::ARGB8888);

    auto shape = Shape::gen();
    shape->appendRect(0, 0, W, H);
    canvas->add(shape);

    for (int i = 0; i < frames; ++i) {
        auto g = LinearGradient::gen();
        g->linear(0, 0, W, H);
        Fill::ColorStop stop = {0.0f, uint8_t(i), 120, 240, 190};
        g->colorStops(&stop, 1);
        shape->fill(g);
        canvas->update();
        canvas->draw(true);
        canvas->sync();
    }

    delete(canvas);
    Initializer::term();
    return 0;
}
```
</details>


src/renderer/cpu_engine/tvgSwShape.cpp
```diff
@@ -520,6 +520,13 @@ bool shapeGenFillColors(SwFill*& out, const Fill* fill, const Matrix& transform,
     } else if (ctable) {
         fillReset(out);
     }
+
+    // single color stop is treated as solid
+    if (fill->colorStops(nullptr) == 1) {
```
hermet, line 525: `fill->colorStops()`

https://github.com/thorvg/thorvg/pull/4809

hermet (approved):

hermet (approved):

PR #4819, opened 2026-09-16
renderer: hotfix clip/mask dangling parent pointers

hermet:
unset parent with all attached resources.
to prevent mask/clip targets indicate invalid parent pointers.


hermet:
\+ @ShinyChang

Issue #3121, opened 2025-01-13
API: Introduce a Path class


hermet:
This is a proposal to separate the Path feature from the Shape, enabling path instance reuse and providing a clearer user API. Meanwhile, here are some concerns regarding the potential downsides:

 - **Increased Complexity**: more class, more code and complexity.
 - **Potential Overhead**: memory fragmentation. 2x instances (shape/path) are required.
 - **Duplication Risks**: Some logic & APIs such as reference count are duplicated.

```
class Path
{    
    Result reset();

    //primitives
    Result moveTo(Point to);
    Result lineTo(Point to);
    Result curveTo(Point ctrl1, Point ctrl2, Point end);
    Result close();

    //preset
    Result append(const PathCommand* cmds, uint32_t cmdCnt, const Point* pts, uint32_t ptsCnt);
    Result appendCircle(Point center, Point radius);
    Result appendRect(Point pos, Point size, Point roundness);

    //TBD: Path combines (maybe after 1.0)
    Result add(Path* path);
    Result subtract(Path* path);
    Result intersect(Path* path);
    Result difference(Path* path);

    Result get(const PathCommand** cmds, uint32_t* cmdsCnt, const Point** pts, uint32_t* ptsCnt) const;

    Path* duplicate();
    Path* gen();

    //reference count
    uint32_t ref();
    uint32_t unref();
    uint32_t refCnt();
};
```
```
auto path = Path::gen();
...
Shape->path(path);
```
```
auto path = Shape->path();
//update the existing path
...
Shape->path(path);
```

SergeyLebedkin:
> Result intersect(Path* path);
> Result difference(Path* path);

Its looks like path merge and can be used for clipping/masking in general way for cross api

hermet:
> Its looks like path merge and can be used for clipping/masking in general way for cross 

@SergeyLebedkin yes, ultimately.

hermet:
This is the first prototype essential API. 
Implemented in `renderer/tvgPath.h/tvgPath.cpp`

```
Path API in thorvg.h
{
    // append the given Path data
    Result add(const PathCommand* cmds, uint32_t cmdCnt, const Point* pts, uint32_t ptsCnt);
    // return the current Path data
    Result get(const PathCommand*& cmds, uint32_t* cmdsCnt, const Point*& pts, uint32_t* ptsCnt) const;
    // reset the internal path data
    void reset();

    // Path Manipulation, apply the rhs to this Path instance. 
[...]

PR #4825, opened 2026-09-19
renderer: introduce Path with boolean operations

tinyjin:
Add an experimental Path class retaining commands and points independently of Shape's path with boolean operations such as union, subtraction, intersection and exclusion between two paths.

https://github.com/user-attachments/assets/7d0a4b1d-aebc-41b7-9e57-4bef0664c010

### Usage
```cpp
PathCommand cmds[] = {PathCommand::MoveTo, PathCommand::LineTo, PathCommand::LineTo, PathCommand::LineTo, PathCommand::Close};
Point pts1[] = {{100, 100}, {400, 100}, {400, 400}, {100, 400}};
Point pts2[] = {{250, 250}, {550, 250}, {550, 550}, {250, 550}};

auto a = Path::gen();
a->add(cmds, 5, pts1, 4);

auto b = Path::gen();
b->add(cmds, 5, pts2, 4);

a->subtract(*b);  //a = a - b

const PathCommand* outCmds;
const Point* outPts;
uint32_t cmdCnt, ptsCnt;
a->get(outCmds, &cmdCnt, outPts, &ptsCnt);

auto shape = Shape::gen();
shape->appendPath(outCmds, cmdCnt, outPts, ptsCnt);
shape->fill(0, 200, 255);
canvas->add(shape);

delete(a);
delete(b);
```

### Experimental APIs:
```cpp
+ Result Path::add(const PathCommand* cmds, uint32_t cmdCnt, const Point* pts, uint32_t ptsCnt)
+ Result Path::get(const PathCommand*& cmds, uint32_t* cmdsCnt, const Point*& pts, uint32_t* ptsCnt) const
+ void Path::reset()
+ Result Path::add(const Path& rhs)
+ Result Path::subtract(const Path& rhs)
+ Result Path::intersect(const Path& rhs)
+ Result Path::difference(const Path& rhs)
+ static Result Path::add(const Path& lhs, const Path& rhs, Path& out)
+ static Result Path::subtract(const Path& lhs, const Path& rhs, Path& out)
+ static Result Path::intersect(const Path& lhs, const Path& rhs, Path& out)
+ static Result Path::difference(const Path& lhs, const Path& rhs, Path& out)
+ static Path* Path::gen()
```

issue: https://github.com/thorvg/thorvg/issues/3121


hermet:
@tinyjin please use tvg formatter. thanks.

src/renderer/tvgPath.cpp
```diff
@@ -0,0 +1,1228 @@
...
+};
+
+
+struct Root
+{
+    float t, u; //lhs(t) = rhs(u)             
+};
+
+
+struct PathPool
```
hermet, line 131: Please add an appropriate comment whenever the code uses any non-obvious technique or special trick.

src/renderer/tvgPath.cpp
```diff
@@ -0,0 +1,1228 @@
...
+            ++block;
+            used = 0;
+        }
+        if (block >= blocks.count) blocks.push(tvg::malloc<uint8_t>(DEFAULT_POOL_SIZE));
+        auto ret = blocks[block] + used;
+        used += size;
+        return ret;
+    }
+
+    static PathPool& scratch()
```
hermet, line 154: Path Manipulation is not a primary feature, static memory pooler doesn't recommended unless we have clear benefits with benchmarking. You can make a separate commit instead.

hermet (changes requested):
@tinyjin You might wanna add proper comments for complex logic addression. Thanks.

Issue #30, opened 2020-09-10
cpu: NEON optimization


hermet:
Still there are a lot of rasterzing parts, not have SIMD operation.
We need a better fine-tuned software raster engine for ARM NEON devices.

hermet:
https://github.com/thorvg/thorvg/wiki/24'-Development-Roadmap

eunsikNota:
I'm interested in this. Can I work on it?

hermet:
@eunsikNota sure, thanks!

PR #4827, opened 2026-09-19
infra: detect CPU SIMD support automatically

hermet:
- Remove the manual Meson SIMD option.
- Probe AVX or NEON support in the CPU engine build.
- Apply the matching compiler flags and configuration defines.

Note that -Dsimd option is removed

issue: https://github.com/thorvg/thorvg/issues/29 https://github.com/thorvg/thorvg/issues/30


hermet:
will be enabled simd by default.

PR #4828, opened 2026-09-20
fix(capi): make C API declarations strict-prototypes compatible

suda-morris:
## Summary

- remove the ineffective top-level `const` from `tvg_picture_get_paint()`'s return type
- use `(void)` for `tvg_lottie_animation_expressions_supported()` so the declaration is a strict C prototype
- keep the declaration and definition in sync

These changes allow C consumers built with `-Werror -Wstrict-prototypes -Wignored-qualifiers` to include the ThorVG C API without requiring warning-suppression flags.

## Validation

- `gcc -fsyntax-only -Werror -Wstrict-prototypes -Wignored-qualifiers` against `thorvg_capi.h`
- `g++ -fsyntax-only -Werror -Wignored-qualifiers` against `thorvg_capi.h`
- `git diff --check`


src/bindings/capi/thorvg_capi.h
```diff
@@ -2163,7 +2163,7 @@ TVG_API Tvg_Result tvg_picture_get_origin(const Tvg_Paint picture, float* x, flo
  * @see tvg_accessor_generate_id()
  * @since 1.0
  */
-TVG_API const Tvg_Paint tvg_picture_get_paint(Tvg_Paint picture, uint32_t id);
+TVG_API Tvg_Paint tvg_picture_get_paint(Tvg_Paint picture, uint32_t id);
```
hermet, line 2166: This shouldn't be changed

hermet (changes requested):

src/bindings/capi/thorvg_capi.h
suda-morris: Tvg_Paint is a pointer typedef:

```c
  typedef struct _Tvg_Paint* Tvg_Paint;
```
Thus, `const Tvg_Paint` means `struct _Tvg_Paint* const`, not `const struct _Tvg_Paint*`. The qualifier applies to the returned pointer value, not to the opaque paint
object. As a top-level qualifier on a function return type, it has no meaningful effect and triggers `-Wignored-qualifiers`.
Is this qualifier being retained for API/source compatibility reasons? If so, we can keep the warning suppression downstream. Otherwise, would removing this
ineffective qualifier be acceptable?

hermet:
@suda-morris  If you need, please add yourself here in Contributor section. https://github.com/thorvg/thorvg/blob/main/CONTRIBUTORS.md
Thanks.

hermet (approved):

PR #4829, opened 2026-09-20
cpu_engine: upgrade x86 SIMD support to AVX2

hermet:
Require AVX2 when SIMD is enabled on x86, including MSVC builds.

Process eight pixels per iteration in alpha blending and premultiply/unpremultiply routines using 256-bit operations. Simplify channel scaling and match scalar blending rounding.

issue: #29


hermet:
No remarkable performance/binary diff.

PR #4830, opened 2026-09-20
cpu_engine: fix the ALPHA_BLEND inconsistency between scalar and NEON

alpakaDurumi:
`ALPHA_BLEND` approximates `c * a / 255` by dividing by 256, and NEON used `(c * a) >> 8` while the scalar has used `c * (a + 1) >> 8` since c825c2f.

NEON now uses the scalar approximation, so a blended pixel no longer depends on the build.

issue: #30


alpakaDurumi:
Narrowed to NEON. #4829 covers AVX.

Nor-s (approved):

hermet:
This is not worth introducing a performance regression just to align with behavior that is arguably less desirable. We'll revisit this if users report real-world issues.

Issue #3539, opened 2025-06-16
lottie: support multiple modifiers of the same type


mgrudzinska:
[sample_1.json](https://github.com/user-attachments/files/20765246/sample_1.json)
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/9a66da29-7023-4d38-a424-bce843c35b46" />

[sample_2.json](https://github.com/user-attachments/files/20765261/sample_2.json)
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/da7724e2-4e5e-4549-bee3-d39c20660e93" />

on #3515 samples trigger the error msg:  ```Decoration skipped to prevent cyclic chain with modifier:...```

hermet:
+ @tinyjin crashed.

```
operator() (__closure=<synthetic pointer>, n=0, pts=0x7fffec002280) at ../src/loaders/lottie/tvgLottieModifier.cpp:404
404                 area += cross(pts[i], pts[i + 1]);
(gdb) where
#0  operator() (__closure=<synthetic pointer>, n=0, pts=0x7fffec002280)
    at ../src/loaders/lottie/tvgLottieModifier.cpp:404
#1  LottieOffsetModifier::modify (this=this@entry=0x7fffec001f50, in=..., out=..., transform=transform@entry=0x0)
    at ../src/loaders/lottie/tvgLottieModifier.cpp:415
#2  0x00007ffff7f831a2 in LottieOffsetModifier::polystar (this=0x7fffec001f50, in=..., out=..., outerRoundness=0,
    hasRoundness=false) at ../src/loaders/lottie/tvgLottieModifier.cpp:462
#3  0x00007ffff7f5c45d in LottieBuilder::updatePolystar (this=this@entry=0x5555555b6fe0,
    parent=parent@entry=0x7fffec001190, child=child@entry=0x7fffec001640, frameNo=frameNo@entry=0, contexts=...,
    ctx=ctx@entry=0x7fffec001e70) at ../src/loaders/lottie/tvgLottieBuilder.cpp:758
#4  0x00007ffff7f6117a in LottieBuilder::updateChildren (this=this@entry=0x5555555b6fe0,
    parent=parent@entry=0x7fffec001190, frameNo=frameNo@entry=0, contexts=...)
    at ../src/loaders/lottie/tvgLottieBuilder.cpp:889
#5  0x00007ffff7f6178d in LottieBuilder::updateGroup (this=this@entry=0x5555555b6fe0,
[...]

PR #4841, opened 2026-09-21
lottie: fix modifier chain

tinyjin:
Modifiers were not applied in the valid order:
- Since roundness modifier doesn't handle LineTo, emitted line by a modifier was ignored, then the shape got broken.
- The modifier chain was reversed in nested group

Emit the lines of the modifiers as cubic segments, and remove decorate trick to assemble the chain in the declared order.

issue: https://github.com/thorvg/thorvg/issues/3539

### Fixed

- test sample: [sample_1 (1).json](https://github.com/user-attachments/files/32459127/sample_1.1.json)

| Patched | Expectation |
| --- | --- |
| CleanShot 2026-09-21 at 6 12 49 PM@2x | Comet 2026-09-21 18 15 18 |


src/loaders/lottie/tvgLottieModifier.cpp
```diff
@@ -408,6 +408,8 @@ RenderPath& LottieOffsetModifier::modify(const RenderPath& in, RenderPath& out,
     };
 
     auto& path = (next) ? RenderPath::scratch() : out;
+    if (in.pts.count == 0) return path;
```
hermet: -> `if (in.pts.empty())`

hermet (changes requested):
please separate fixes
- modifier chain correction
- roundness logic correction

Issue #4840, opened 2026-09-21
SVG ellipses with non-positive radii don't follow any specification


MewPurPur:
Ellipses are weird in SVG specifications, SVG 1.1, SVG Tiny 1.2, and SVG 2.0 all have different ideas of what to do with their edge cases, like when one radius is missing, if "auto" takes from the other radius, and how negative values are treated. At least, they are all consistent about how to treat the value 0.

ThorVG fails on negative values, not conforming to any specifications. For negative values, it uses the absolute value, but the specifications instead have the following ideas:
- SVG 1.1: This should error out
- SVG Tiny 1.2: This is unsupported, therefore assuming a value of 0, therefore should not render
- SVG 2 current draft: This should be ignored, therefore assuming a value of "auto" and copying the other radius

It also fails on 0. All specifications say this should stop rendering, but ThorVG still renders.

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16">
	<ellipse rx="6" ry="0" cx="8" cy="8" fill="red" stroke="blue" />
</svg>
```

<img width="1040" height="682" alt="Image" src="https://github.com/user-attachments/assets/2b4ad8b9-b4d0-4520-a9c6-aed528ce99e6" />

MewPurPur:
Seems circles should similarly not accept negative numbers.

hermet:
@MewPurPur @JSUYA the fix will be in v1.1.3. thanks.

PR #4847, opened 2026-09-22
svg: fix circle and ellipse radius handling

JSUYA:
Resolve missing and negative ellipse radii from the other axis when building paths, preserving the parsed node values. Keep explicit zero radii non-rendering and skip circles with non-positive radii.

issues: #4840


src/loaders/svg/tvgSvgLoader.cpp
```diff
@@ -1865,6 +1865,8 @@ static SvgNode* _createEllipseNode(SvgParserContext* ctx, SvgNode* parent, const
 
     if (!ctx->parser->node) return nullptr;
 
+    ctx->parser->node->node.ellipse.rx = ctx->parser->node->node.ellipse.ry = -1.0f;
```
hermet, line 1869: `// Unspecified radii default to 'auto(-1)'; resolved when building the path.`

src/loaders/svg/tvgSvgBuilder.cpp
```diff
@@ -209,6 +209,10 @@ static void _appendRect(Shape* shape, float x, float y, float w, float h, float
 
 static void _appendCircle(Shape* shape, float cx, float cy, float rx, float ry)
 {
+    if (rx < 0.0f) rx = ry;
```
hermet, line 213: `// Resolve unspecified or invalid radii from the other axis (SVG 2).`

hermet (approved):

hermet:
good to add those comments.

PR #4853, opened 2026-09-22
cpu_engine: size reduction with consolidate gradient rasterization

hermet:
Replace per-gradient raster templates with shared fillRaster entry points to reduce duplicated raster code. Cache the gradient type in SwFill and dispatch to linear, radial, or conic helpers for each span.

Group gradient preparation and sampling helpers by type, and localize repeat antialiasing and conic angle calculations.

size: -3kb


hermet:
LinearBench: --
RadialBench: -1%

